PR #155, opened 2021-09-30
Use Docker ENTRYPOINT instead of CMD

mterron:
Enables using spicedb from docker directly. 
```docker run quay.io/authzed/spicedb serve --grpc-preshared-key "somerandomkeyhere" --grpc-no-tls```

Docker's documentation:
The best use for ENTRYPOINT is to set the image’s main command, allowing that image to be run as though it was that command (and then use CMD as the default flags).


jzelinskie:
I actually got burnt by this when first writing the SpiceDB GitHub Action.
We didn't previously have SpiceDB as the entrypoint because we had multiple binaries in the image, but now we're consolidated into one.

PR #164, opened 2021-10-04
Ensure e2e doesn't time out when it would have succeeded

ecordell:
Because of the network delay added for the newenemy test, sometimes the number of required successful iterations exceeded the timeout. This change makes the timeout depend on the number of iterations picked.

I also adjusted the constants to try to improve the worst-case runtime, and marked the artifact upload step to not fail the job if it fails (since it has a surprisingly high failure rate).


josephschorr (approved):
LGTM

PR #165, opened 2021-10-04
fix TestReadBadZookieFlake

### LINKED ISSUE #101: TestReadBadZookie is flakey
```
--- FAIL: TestReadBadZookie (0.01s)
    acl_test.go:266: 
        	Error Trace:	acl_test.go:266
        	Error:      	Received unexpected error:
        	            	rpc error: code = OutOfRange desc = invalid zookie: revision has expired
        	Test:       	TestReadBadZookie
```

ecordell:
The flake is caused by GC pauses. I can't produce a failure with `GOGC=off`.

When the test gets unlucky w.r.t GC, the timestamps in the changelog can be behind `now-gcwindow` by 5-7ms. Bumping the GC window up a bit stabilizes the test so that `now-gcwindow` is less than the changelog timestamps even if there are GC pauses.

Prior to this change, I could get 1-5 failures / 1k runs. After the change, I can't reproduce the flake.

Fixes #101

### DIFF
--- a/internal/services/v0/acl_test.go
+++ b/internal/services/v0/acl_test.go
@@ -246,7 +246,7 @@ func TestRead(t *testing.T) {
 func TestReadBadZookie(t *testing.T) {
 	require := require.New(t)
 
-	client, stop, revision, _ := newACLServicer(require, 0, 10*time.Millisecond, 0)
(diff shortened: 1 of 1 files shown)


josephschorr (approved):
LGTM

PR #168, opened 2021-10-04
Garbage collection support for Postgres

josephschorr:


internal/datastore/postgres/postgres.go
```diff
@@ -134,7 +155,157 @@ type pgDatastore struct {
...
+	for {
+		select {
+		case <-pgd.disposed:
+			log.Info().Msg("shutting down postgres GC")
+			return
+		case <-time.After(pgd.gcInterval):
+			err := pgd.collectGarbage()
+			if err != nil {
+				log.Warn().Err(err).Msg("error when attempting to perform garbage collection")
+			} else {
```
ecordell, line 179: nit: this can be undented

internal/datastore/postgres/postgres.go
```diff
@@ -118,14 +126,27 @@ func NewPostgresDatastore(
...
 		dburl:                     url,
 		dbpool:                    dbpool,
 		watchBufferLength:         config.watchBufferLength,
 		revisionFuzzingTimedelta:  config.revisionFuzzingTimedelta,
 		gcWindowInverted:          -1 * config.gcWindow,
+		gcInterval:                config.gcInterval,
+		gcMaxOperationTime:        config.gcMaxOperationTime,
 		splitAtEstimatedQuerySize: config.splitAtEstimatedQuerySize,
-	}, nil
+		disposed:                  make(chan struct{}),
```
ecordell, line 138: since this is the only datastore than needs to GC anything what about injecting the channel when constructing it?

i.e.:
```go
func main() {
  done := make(chan struct{}) 
  ds := NewPostgresDatastore("url", WithDone(done))
  defer close(done)
}
```

that way you don't have to have `Dispose()` method in the interface (this approach can also close the channel when the process is killed, which I don't see happening right now)

internal/datastore/postgres/postgres.go
```diff
@@ -134,7 +155,157 @@ type pgDatastore struct {
...
+
+	// RelationTupleTransaction is not timezone aware
+	// Explicitly use UTC before using as a query arg
+	now = now.UTC()
+
+	return now, nil
+}
+
+func (pgd *pgDatastore) collectGarbage() error {
+	ctx, cancel := context.WithTimeout(context.Background(), pgd.gcMaxOperationTime)
```
ecordell, line 207: it seems important to know if GC is timing out frequently - do we want a prom metric for this?

internal/datastore/postgres/postgres.go
```diff
@@ -134,7 +155,157 @@ type pgDatastore struct {
...
+	relCount, err := pgd.batchDelete(ctx, tableTuple, sq.LtOrEq{colDeletedTxn: highest})
+	if err != nil {
+		return 0, 0, err
+	}
+
+	log.Trace().Uint64("highest_transaction_id", highest).Int64("relationships_deleted", relCount).Msg("deleted stale relationships")
+
+	// Delete all transaction rows with ID < the transaction ID. We don't delete the transaction
+	// itself to ensure there is always at least one transaction present.
+	transactionCount, err := pgd.batchDelete(ctx, tableTransaction, sq.Lt{colID: highest})
```
ecordell, line 274: `transactionCount` could be a useful metric to track as well.

internal/datastore/postgres/options.go
```diff
@@ -38,7 +43,9 @@ type Option func(*postgresOptions)
 
 func generateConfig(options []Option) (postgresOptions, error) {
 	computed := postgresOptions{
-		gcWindow:                  24 * time.Hour,
+		gcWindow:                  defaultGarbageCollectionWindow,
+		gcInterval:                defaultGarbageCollectionInterval,
```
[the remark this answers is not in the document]
josephschorr: I'm open to suggestions :) I think `gcMaxOperationTime` is pretty explanatory. As for `gcWindow` vs `gcInterval`, maybe `gcRunFrequency`?

internal/datastore/postgres/postgres.go
josephschorr: I was going to add metrics as a followup, but I can do so in this PR if you like

internal/datastore/postgres/postgres.go
ecordell: return an error if `WithDone` isn't given?

or we could have datastores take a context, and any datastore that needs to can listen for context.Done?

ecordell: `WithDone` would only be required by the postgres ds constructor in that case, though. And context still may be a better fit?

internal/datastore/postgres/options.go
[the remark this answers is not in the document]
ecordell: snapshotGC vs transactionGC?

internal/datastore/postgres/postgres.go
jzelinskie: Injecting a channel is very weird. I think we should just make the API similar to how we're managing the lifecycles of http and gRPC servers in the cobra runfuncs: a simple Close() or Shutdown() method that blocks (I've seen them take contexts sometimes too for configuring deadlines).

Inside the postgres implementation we should probably use errgroup and context cancellation rather than a chan of empty struct.

josephschorr: @jzelinskie That's what `Dispose` does (minus using a chan vs an errgroup). I preferred `Dispose` for the name as it might be also cleaning up data too (in the potential sqlite case if we ended up using a file)

internal/datastore/postgres/options.go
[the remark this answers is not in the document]
josephschorr: Its not those, though. `gcInterval` is how often is runs. `gcWindow` is the duration until things are marked as garbage and thereafter collectable

josephschorr:
Updated

internal/datastore/postgres/options.go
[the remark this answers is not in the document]
ecordell: from a user's perspecitve, `gcWindow` affects how old of a revision they're allowed to use to query (it's really not related to GC at all in the crdb datastore for example). It's a setting that directly affects end users

gcInterval and gcMaxOperationTime are more ops-facing flags, they affect how garbage is managed over time. 

maybe this is picking too many nits, they just seem like a different class of thing to me. I wouldn't block the PR on this, but I do think there's probably a clearer way to name this.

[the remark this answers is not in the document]
josephschorr: `gcWindow` only does that in CRDB and memdb because they have different means of collecting garbage (or not at all). After this change, it does directly affect postgres, because anything older than that window will (eventually) be deleted

internal/datastore/memdb/namespace.go
```diff
@@ -19,6 +19,10 @@ const (
 )
 
 func (mds *memdbDatastore) WriteNamespace(ctx context.Context, newConfig *v0.NamespaceDefinition) (datastore.Revision, error) {
+	if mds.db == nil {
+		panic("memdb datastore already closed")
```
jzelinskie, line 23: this is panic but the rest are errors, which do we want? I'd say panic -- that's what other things in the stdlib do (even a double Close() on resources in the stdlib panic)

internal/datastore/postgres/postgres.go
```diff
@@ -134,7 +178,177 @@ type pgDatastore struct {
...
+	cancelGc context.CancelFunc
+}
+
+func (pgd *pgDatastore) Close() error {
+	pgd.dbpool.Close()
+	pgd.cancelGc()
+
+	if pgd.gcGroup != nil {
+		err := pgd.gcGroup.Wait()
+		log.Info().Err(err).Msg("completed shutdown of postgres datastore")
```
jzelinskie, line 196: this should probably be log.Warn()

internal/datastore/postgres/postgres.go
```diff
@@ -118,14 +146,30 @@ func NewPostgresDatastore(
...
+		gcInterval:                config.gcInterval,
+		gcMaxOperationTime:        config.gcMaxOperationTime,
 		splitAtEstimatedQuerySize: config.splitAtEstimatedQuerySize,
-	}, nil
+		gcCtx:                     gcCtx,
+		cancelGc:                  cancelGc,
+	}
+
+	// Start a goroutine for garbage collection.
+	if datastore.gcInterval > 0*time.Minute {
```
jzelinskie, line 165: i think there's a time.IsZero we could use here `if !time.IsZero(datastore.gcInterval)`

internal/datastore/postgres/postgres.go
```diff
@@ -178,19 +182,32 @@ type pgDatastore struct {
...
-	disposed chan struct{}
-	wg       sync.WaitGroup
+	gcGroup  *errgroup.Group
+	gcCtx    context.Context
+	cancelGc context.CancelFunc
 }
 
-func (pgd *pgDatastore) runGarbageCollector() {
+func (pgd *pgDatastore) Close() error {
+	pgd.dbpool.Close()
```
ecordell, line 191: shouldn't the connections be closed after the gcgroup has finished stopping?

PR #169, opened 2021-10-05
k8s: init basic deployment

jzelinskie:
This is quite basic, but is documented as such and should be a reasonably starting place for people until we have time to flesh out something more mature.

It's hard to know where to draw the line for what dependencies the community would like to see maintained in the upstream example. For example, the Authzed deployment of SpiceDB depends on a some internal software and few Kubernetes Operators such as cert-manager and cockroach.

### DIFF
--- a/k8s/basic.yaml
+++ b/k8s/basic.yaml
@@ -1,0 +1,90 @@
+# This file contains the most basic configuration of SpiceDB in Kubernetes.
+#
+# It runs with the following:
+#   - a single node deployment
+#   - default ports (gRPC 50051, dashboard 8080)
+#   - no TLS
+#   - debug logging
+#   - in-memory datastore
+#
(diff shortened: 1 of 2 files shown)


k8s/basic.yaml
```diff
@@ -0,0 +1,90 @@
+# This file contains the most basic configuration of SpiceDB in Kubernetes.
+#
+# It runs with the following:
+#   - a single node deployment
+#   - default ports (gRPC 50051, dashboard 8080)
+#   - no TLS
+#   - debug logging
+#   - in-memory datastore
+#
+# To apply this confiugration execute the following:
```
ecordell, line 10:
```suggestion
# To apply this configuration execute the following:
```

ecordell (approved):
LGTM

Issue #13, opened 2021-08-17
Document Zookies


jzelinskie:
A sufficient document on Zookie would cover the following subjects:

- How they are different from a typical read snapshot token
- How they are used in Zanzibar
- How they are used in v0 API
- How they are used in v1 API
- How they are represented internally within SpiceDB

PR #174, opened 2021-10-05
Add documentation about ZedTokens/Zookies and consistency

josephschorr:
Fixes https://github.com/authzed/spicedb/issues/13


docs/consistency-options.md
```diff
@@ -0,0 +1,69 @@
+# Consistency Options
+
+SpiceDB's [V1 API] provides the ability to specify the *consistency* level at which API calls are made.
+
+Consistency is provided via the [Consistency block] on API calls.
+
+[V1 API]: https://buf.build/authzed/api/tree/main/authzed/api/v1
+[Consistency block]: https://buf.build/authzed/api/file/main/authzed/api/v1/permission_service.proto#L50
```
jzelinskie, line 8: That URL is guaranteed to change, you probably meant this one: https://buf.build/authzed/api/docs/main/authzed.api.v1#authzed.api.v1.Consistency

docs/consistency-options.md
```diff
@@ -0,0 +1,69 @@
+# Consistency Options
```
jzelinskie, line 1: Do we think it's possible that this could live as protobuf docstring documentation instead of a dedicated document? The documentation is almost already exactly this already.

docs/zedtokens-and-zookies.md
```diff
@@ -0,0 +1,187 @@
+# ZedTokens and Zookies
```
jzelinskie, line 1: Rather than the focus of this being documenting zed tokens / zookies, could we make this focused on "consistency and availability of spicedb" in general? Then we can answer questions like

- where does SpiceDB fit into the CAP theorem?
- what knobs for consistency are available to application developers?
- what happens if SpiceDB is unavailable?

and also cover zed tokens/zookies in a way that people will discover them when they aren't familiar with the terms or how they relate to the system in general

docs/consistency-options.md
josephschorr: Its not super discoverable right now, which is why I put it here

docs/zedtokens-and-zookies.md
josephschorr: IMO most people will not care initially about reading about the CAP theorem, but rather how to effectively use the system. The consistency knobs are described in the other consistency document.

If we want a section on CAP guarantees, we can add that, but it is likely too technical for people trying to integrate effectively initially.

PR #202, opened 2021-10-21
add support for decorators to schema langauge

ecordell:
Supports decorators of the forms:

```
@decoratorName
@decoratorName(opt1,opt2)
@decoratorName(opt1,(opt2a,opt2b))
```

This parses the decorators into the AST as a new edge type from the items they decorate (currently only supported within `definition` blocks, pending more discussion of how they will be used)

- [x] basic decorators
- [x] options
- [x] nested options
- [x] multiple decorators for the same statement
- [ ] strings or identifiers as option values

### DIFF
--- a/pkg/schemadsl/dslshape/dslshape.go
+++ b/pkg/schemadsl/dslshape/dslshape.go
@@ -17,6 +17,9 @@ const (
 	NodeTypeRelation   // A relation
 	NodeTypePermission // A permission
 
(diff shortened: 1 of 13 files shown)


josephschorr:
@alsbury yeah, makes perfect sense. The reason we kept the comment data in the parsed protos so was that we could use the entered descriptions in generated code down the road

josephschorr:
Closing in favor of work that will be done as part of composable schemas

PR #220, opened 2021-10-26
docs: remove all

jzelinskie:
We open-sourced our documentation website, which is where all future documentation will live and be maintained.


samkim (approved):
LGTM

PR #242, opened 2021-11-01
proxy: use buffered channels and only let one subrequest write a result

### LINKED ISSUE #235: Goroutine leak possibly in hedging middleware
It appears that there could be an issue with context cancelation and the goroutines being created for hedging read requests where they get parked indefinitely.

Output from an goroutine pprof profile:

```
Showing nodes accounting for 44362, 100% of 44364 total
Dropped 102 nodes (cum <= 221)
      flat  flat%   sum%        cum   cum%
     44362   100%   100%      44362   100%  runtime.gopark
         0     0%   100%       6888 15.53%  github.com/authzed/spicedb/internal/datastore/proxy.hedgingProxy.ReadNamespace.func1
         0     0%   100%      37158 83.76%  github.com/authzed/spicedb/internal/datastore/proxy.hedgingProxy.Revision.func1
         0     0%   100%      44260 99.77%  runtime.chansend
         0     0%   100%      44260 99.77%  runtime.chansend1
```

ecordell:
Fixes #235 

Once we’re in a “hedged” case, we have two goroutines running, and this select reading from both response channels for answers:
```go
select {
case <-responseReady:
  duration = timeSource.Since(originalStart)
case <-hedgedResponseReady:
  duration = timeSource.Since(hedgedStart)
}
```

Once we get an answer from one of the two, the hedger unblocks and cancels the context that is shared by both subrequest goroutines.

This is what a subrequest looks like for reference:

```go
func(ctx context.Context, responseReady chan<- struct{}) {
  rev, err = hp.delegate.Revision(ctx)
  responseReady <- struct{}{}
}
```
So if there are two of these running, the context is cancelled once the fastest one returns and the hedger func exits, and there’s nothing to read from the “slow” channel, so the subrequest goroutine hangs on send.

There are two parts to the fix:

- using buffered "ready" channels for hedger will keep the subrequest goroutines from blocking on send
- since both subrequests capture the parent return values and attempt to write to them, there's a potential data race. they're protected with `sync.Once` to avoid this problem

The tests were updated to detect the leak (and confirm it no longer happens)

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -213,7 +213,7 @@ func (cds *crdbDatastore) Revision(ctx context.Context) (datastore.Revision, err
 }
 
 func (cds *crdbDatastore) SyncRevision(ctx context.Context) (datastore.Revision, error) {
-	ctx, span := tracer.Start(ctx, "SyncRevision")
+	ctx, span := tracer.Start(datastore.SeparateContextWithTracing(ctx), "SyncRevision")
 	defer span.End()
 
 	tx, err := cds.conn.BeginTx(ctx, pgx.TxOptions{AccessMode: pgx.ReadOnly})
--- a/internal/datastore/proxy/hedging.go
+++ b/internal/datastore/proxy/hedging.go
@@ -60,7 +60,7 @@ func newHedger(
 	digests[0].Add(initialSlowRequestThreshold.Seconds(), float64(maxSampleCount)/2)
 
 	return func(ctx context.Context, req subrequest) {
-		responseReady := make(chan struct{})
+		responseReady := make(chan struct{}, 1)
 
 		digestLock.Lock()
 		slowRequestThresholdSeconds := digests[0].Quantile(quantile)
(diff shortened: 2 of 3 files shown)


internal/datastore/crdb/crdb.go
```diff
@@ -213,7 +213,7 @@ func (cds *crdbDatastore) Revision(ctx context.Context) (datastore.Revision, err
 }
 
 func (cds *crdbDatastore) SyncRevision(ctx context.Context) (datastore.Revision, error) {
-	ctx, span := tracer.Start(ctx, "SyncRevision")
+	ctx, span := tracer.Start(datastore.SeparateContextWithTracing(ctx), "SyncRevision")
```
ecordell, line 216: this seemed like an oversight - any reason we wouldn't want a separate context here?

internal/datastore/proxy/hedging.go
```diff
@@ -276,24 +292,20 @@ type hedgingTupleQuery struct {
...
-		doneLock.Lock()
-		defer doneLock.Unlock()
-
-		if alreadyReturned {
-			if tempErr == nil {
-				tempIterator.Close()
-			}
-		} else {
-			alreadyReturned = true
+		executed := false
```
jakedt, line 298: Maybe rename `executed` to `resultsUsed` or something? As it is it's a little tricky to reason about why they don't all end up with `executed = true`.

jakedt (changes requested):

internal/datastore/proxy/hedging.go
ecordell: used the name you suggested + added a comment about it

jakedt (approved):

PR #243, opened 2021-11-01
support https in download API

ecordell:
this brings the download API in line with the other http / grpc apis.

### DIFF
--- a/cmd/spicedb/developer.go
+++ b/cmd/spicedb/developer.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"net/http"
 	"os"
 	"os/signal"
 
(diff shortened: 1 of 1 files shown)


cmd/spicedb/developer.go
```diff
@@ -48,6 +48,9 @@ func registerDeveloperServiceCmd(rootCmd *cobra.Command) {
 	developerServiceCmd.Flags().String("s3-endpoint", "", "s3 endpoint for s3 share store")
 	developerServiceCmd.Flags().String("s3-region", "auto", "s3 region for s3 share store")
 	developerServiceCmd.Flags().String("download-addr", ":8443", "address to listen for download requests")
```
jzelinskie, line 50: This is actually unfortunate because I think this is the same as the default port for the gateway.
The gateway should probably run on this port instead, but I'm not sure if it makes sense to run a whole 'nother HTTP server just for downloading schemas like this.

jakedt: it's unlikely that these servers will ever run in the same binary for real though, right? In production we run developer service separately from the actual service. I think it's fine to define to http servers for the two different use cases. It will still work in the unlikely event that they get run together. We do have to untangle the ports though.

ecordell:
blocked on #254

jzelinskie (approved):

PR #259, opened 2021-11-04
Use buffered channels for lookup results

ecordell:
This prevents leaf goroutines from getting stuck on write to a channel nothing is reading from

(I haven't convinced myself that there are no other goroutine leaks in the lookup dispatch, but this is definitely a source of leaks)

### DIFF
--- a/internal/services/v1/permissions_test.go
+++ b/internal/services/v1/permissions_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/shopspring/decimal"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/test/bufconn"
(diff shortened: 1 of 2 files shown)


internal/services/v1/permissions_test.go
```diff
@@ -378,6 +379,7 @@ func TestLookupResources(t *testing.T) {
 				t.Run(fmt.Sprintf("%s::%s from %s:%s#%s", tc.objectType, tc.permission, tc.subject.Object.ObjectType, tc.subject.Object.ObjectId, tc.subject.OptionalRelation), func(t *testing.T) {
 					require := require.New(t)
 					client, stop, revision := newPermissionsServicer(require, delta, memdb.DisableGC, 0)
+					defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
```
ecordell, line 382: it looks weird to have this before `defer stop()`, but that lets `IgnoreCurrent` capture the running server's connections goroutines and check for leaks after `stop()`

josephschorr (approved):
LGTM

Issue #261, opened 2021-11-05
LookupSubjects API


josephschorr:
The [Lookup Watch API Proposal](https://github.com/authzed/spicedb/issues/207) includes the addition of the "reachability" APIs, which allow a caller to query the data-driven shape of the permissions graph.

One of the APIs proposed is `LookupSubjects` which would act as a filtered, *streaming* form of `ExpandPermission`, but across an entire object type:

```
message LookupSubjectsRequest {
    Consistency consistency = 1;

    ObjectReference resource = 2;
    string optional_permission = 3;

    string optional_subject_type = 4;
    string optional_subject_relation = 5;
}

message LookupSubjectsResponse {
    Relationship found_relationship = 1;
    ZedToken found_at = 2;
}
```

All fields on the request besides `consistency` and `resource` would be optional, in which case *all* subjects (of all kinds) would be find for the specified `resource`.

## Open Questions

1) Should the `LookupSubjectsResponse` contain the *path* of all relations/permissions that were traversed to reach a subject? This could be very useful in building permissions panels or auditing systems.
2) Should `optional_subject_type` (and relation) be repeated, to allow filtering to a set of allowed types, instead of a single type?
3) Should `optional_permission` be repeated, to allow filtering to a set of allowed permissions/relations?

jonwhitty:
> Should the LookupSubjectsResponse contain the path of all relations/permissions that were traversed to reach a subject? This could be very useful in building permissions panels or auditing systems.

I think the client should be able to selectively choose to return the path of all relations/permissions. I like the idea of using a FieldMask here to allow the client to specify to include it or not.

> Should optional_subject_type (and relation) be repeated, to allow filtering to a set of allowed types, instead of a single type?

I think so, yeah. This just gives the client more flexibility to be selective about what they want to lookup. This could be really useful if the client only cares about certain subject types or relations. It also doesn't restrict the ability to lookup _all_, we'll just want to make sure we account for that case. To me it makes sense that if you omit this repeated field then that indicates you want to lookup all subjects for any subject type (and relation).

> Should optional_permission be repeated, to allow filtering to a set of allowed permissions/relations?

[...]

PR #263, opened 2021-11-05
feat: add v1 Watch API implementation

jonwhitty:
The changes herein add support for the v1 Watch API. The implementation chooses to implement a wildcard approach which allows the user to omit the `ObjectType` list if they wish to watch all object types. This approach was recommended by @ecordell and myself in the development channel of Discord and has also been discussed as an approach in the LookupSubjects API as discussed in #261.

https://discord.com/channels/844600078504951838/900405749405089812/906247932490158082
https://discord.com/channels/844600078504951838/900405749405089812/906274500004438036

### DIFF
--- a/internal/services/v1/watch.go
+++ b/internal/services/v1/watch.go
@@ -1,0 +1,104 @@
+package v1
+
+import (
+	"errors"
+
+	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	grpcvalidate "github.com/grpc-ecosystem/go-grpc-middleware/validator"
+	"github.com/shopspring/decimal"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/authzed/spicedb/internal/datastore"
+	"github.com/authzed/spicedb/internal/services/shared"
+	"github.com/authzed/spicedb/pkg/tuple"
+	"github.com/authzed/spicedb/pkg/zedtoken"
+)
+
+type watchServer struct {
+	v1.UnimplementedWatchServiceServer
+	shared.WithStreamServiceSpecificInterceptor
+
+	ds datastore.Datastore
+}
+
+// NewWatchServer creates an instance of the watch server.
+func NewWatchServer(ds datastore.Datastore) v1.WatchServiceServer {
+	s := &watchServer{
+		ds: ds,
+		WithStreamServiceSpecificInterceptor: shared.WithStreamServiceSpecificInterceptor{
+			Stream: grpcvalidate.StreamServerInterceptor(),
+		},
+	}
+	return s
+}
+
+func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchServer) error {
+	objectTypesMap := make(map[string]struct{})
+	for _, objectType := range req.GetOptionalObjectTypes() {
+		objectTypesMap[objectType] = struct{}{}
+	}
+
+	var afterRevision decimal.Decimal
+	if req.OptionalStartCursor != nil && req.OptionalStartCursor.Token != "" {
+		decodedRevision, err := zedtoken.DecodeRevision(req.OptionalStartCursor)
+		if err != nil {
+			return status.Errorf(codes.InvalidArgument, "failed to decode start revision: %s", err)
+		}
+
+		afterRevision = decodedRevision
+	} else {
+		var err error
+		afterRevision, err = ws.ds.Revision(stream.Context())
+		if err != nil {
+			return status.Errorf(codes.Unavailable, "failed to start watch: %s", err)
+		}
+	}
+
+	updates, errchan := ws.ds.Watch(stream.Context(), afterRevision)
+	for {
+		select {
+		case update, ok := <-updates:
+			if ok {
+				filtered := filterUpdates(objectTypesMap, update.Changes)
+				if len(filtered) > 0 {
+					if err := stream.Send(&v1.WatchResponse{
+						Updates:        filtered,
+						ChangesThrough: zedtoken.NewFromRevision(update.Revision),
+					}); err != nil {
+						return status.Errorf(codes.Canceled, "watch canceled by user: %s", err)
+					}
+				}
+			}
+		case err := <-errchan:
+			switch {
+			case errors.As(err, &datastore.ErrWatchCanceled{}):
+				return status.Errorf(codes.Canceled, "watch canceled by user: %s", err)
+			case errors.As(err, &datastore.ErrWatchDisconnected{}):
+				return status.Errorf(codes.ResourceExhausted, "watch disconnected: %s", err)
+			default:
+				return status.Errorf(codes.Internal, "watch error: %s", err)
+			}
+		}
+	}
+}
+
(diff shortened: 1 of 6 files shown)


internal/datastore/datastore.go
```diff
@@ -14,8 +14,9 @@ const Ellipsis = "..."
 
 // RevisionChanges represents the changes in a single transaction.
 type RevisionChanges struct {
-	Revision Revision
-	Changes  []*v0.RelationTupleUpdate
+	Revision            Revision
+	Changes             []*v0.RelationTupleUpdate
+	RelationshipUpdates []*v1.RelationshipUpdate
```
jonwhitty, line 19: I favored adding this here instead of converting all `v0.RelationTupleUpdate` objects downstream because it makes downstream consumption easier. At one point when we deprecate v0 officially we can just drop the `Changes` field altogether and refactor the datastore implementations to omit it as well.

jonwhitty:
Still need to add tests, but this can start being reviewed.

internal/datastore/datastore.go
josephschorr: I have a slight concern about performance, but as you said, we can just deprecate v0 at some point and convert in there if need be

internal/services/v1/watch.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return s
+}
+
+func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchServer) error {
+	objectTypesMap := make(map[string]struct{})
+	for _, objectType := range req.GetObjectTypes() {
+		objectTypesMap[objectType] = struct{}{}
+	}
+	filter := objectTypeFilter{objectTypes: objectTypesMap}
```
josephschorr, line 41: Why the new filter type vs just sending the map?

internal/services/v1/watch.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+}
+
+type objectTypeFilter struct {
+	objectTypes map[string]struct{}
+}
+
+func (of objectTypeFilter) filterUpdates(candidates []*v1.RelationshipUpdate) []*v1.RelationshipUpdate {
+	if len(of.objectTypes) < 1 {
```
josephschorr, line 92: `len(...) == 0`

internal/services/v1/watch.go
jonwhitty: No particular reason other than it was similar to the other pattern used in v0's implementation. I changed it though. I like passing it in to a function better anyways.

internal/services/v1/watch.go
jonwhitty: 👍

jonwhitty:
@josephschorr is it common for this test to fail?

```
FAIL: TestReverseTupleQueryRequestHedging (0.01s)
```
I don't see that test fail locally ever and these additions didn't really impact that at all I don't think? 🤷

josephschorr:
> @josephschorr is it common for this test to fail?
> 
> ```
> FAIL: TestReverseTupleQueryRequestHedging (0.01s)
> ```
> 
> I don't see that test fail locally ever and these additions didn't really impact that at all I don't think? 🤷

I just forced a rerun; we shall see. It might be flaky.

internal/services/v1/watch_test.go
```diff
@@ -0,0 +1,206 @@
...
+				update(v1.RelationshipUpdate_OPERATION_CREATE, "document", "document1", "viewer", "user", "user1"),
+				update(v1.RelationshipUpdate_OPERATION_CREATE, "folder", "folder1", "viewer", "user", "user1"),
+			},
+			expectedUpdates: []*v1.RelationshipUpdate{
+				update(v1.RelationshipUpdate_OPERATION_CREATE, "document", "document1", "viewer", "user", "user1"),
+				update(v1.RelationshipUpdate_OPERATION_CREATE, "folder", "folder1", "viewer", "user", "user1"),
+			},
+		},
+		{
+			name:              "watch with objectType filter",
```
josephschorr, line 75: Please add a test with different update kinds as well (DELETE, TOUCH)

internal/services/v1/watch_test.go
```diff
@@ -0,0 +1,206 @@
...
+				OptionalStartCursor: cursor,
+			})
+			require.NoError(err)
+
+			if tc.expectedCode == codes.OK {
+
+				updatesChan := make(chan []*v1.RelationshipUpdate, len(tc.mutations))
+
+				go func() {
+					defer func() {
```
josephschorr, line 133: `defer close(updatesChan)`?

internal/services/v1/watch_test.go
```diff
@@ -0,0 +1,206 @@
...
+							panic(fmt.Errorf("timed out waiting for stream updates"))
+						default:
+							resp, err := stream.Recv()
+							if err != nil {
+								errStatus, ok := status.FromError(err)
+								if (ok && (errStatus.Code() == codes.Canceled || errStatus.Code() == codes.Unavailable)) || err == io.EOF {
+									break
+								}
+
+								panic(fmt.Errorf("received a stream read error: %v", err))
```
josephschorr, line 151: `require.Fail`?

jonwhitty: `require.Fail` inside of a goroutine that is not the main routine that the `testing.T` object is associated with causes unclear failures such as `panic: Fail in goroutine after <process> has completed`.

Would you prefer `require.Fail` here instead?

internal/services/v1/watch_test.go
jonwhitty: Ok 👍

internal/services/v1/watch_test.go
jonwhitty: Yeah, is it wrong to close the updatesChan after we're done writing to it?

internal/services/v1/watch_test.go
josephschorr: Ah... I guess this is okay then

internal/services/v1/watch_test.go
jonwhitty: Ah, sorry you're saying streamlining the defer. 👍

josephschorr: Probably not, no

jonwhitty: Fixed.

pkg/tuple/tuple.go
```diff
@@ -188,6 +188,18 @@ func ToFilter(tpl *v0.RelationTuple) *v1.RelationshipFilter {
 	}
 }
 
+// UpdatesToRelationshipUpdates converts a slice of RelationTupleUpdate into a
+// slice of RelationshipUpdate.
+func UpdatesToRelationshipUpdates(updates []*v0.RelationTupleUpdate) []*v1.RelationshipUpdate {
+	relationshipUpdates := make([]*v1.RelationshipUpdate, len(updates))
```
josephschorr, line 194: `make([]*v1.RelationshipUpdate, 0, len(updates))`

and then use `append` in the for loop

internal/datastore/datastore.go
jakedt, line 19: Since `RelationTupleUpdate` objects are trivially convertible to `RelationshipUpdate` objects we should pick one and only one for the datastore interface.

jonwhitty: Ok, sounds good. I've undone these changes in favor of doing the conversion in the Watch implementation instead.

josephschorr (approved):
LGTM

PR #268, opened 2021-11-09
Multi level caching and Lookup caching fixes

josephschorr:


internal/dispatch/caching/caching.go
```diff
@@ -182,22 +183,28 @@ func (cd *CachingDispatcher) DispatchExpand(ctx context.Context, req *v1.Dispatc
...
 			return cachedResult.result, nil
 		}
 	}
 
 	computed, err := cd.d.DispatchLookup(ctx, req)
 
-	// We only want to cache the result if there was no error
-	if err == nil {
+	// We only want to cache the result if there was no error and nothing was excluded.
+	if err == nil && len(computed.Metadata.LookupExcludedDirect) == 0 && len(computed.Metadata.LookupExcludedTtu) == 0 {
```
jakedt, line 200: Where do these actually get set?

jakedt (changes requested):

internal/dispatch/caching/caching.go
josephschorr: In the Lookup dispatcher whenever something is skipped:

- https://github.com/authzed/spicedb/pull/268/files#diff-e24eb2e567330120b1257156553fbb962c4bcaf4c44ed5366f2ffeb180af35c3R567
- https://github.com/authzed/spicedb/pull/268/files#diff-e24eb2e567330120b1257156553fbb962c4bcaf4c44ed5366f2ffeb180af35c3R342

They are then cleared when the direct dispatch or TTU has completed:
- https://github.com/authzed/spicedb/pull/268/files#diff-e24eb2e567330120b1257156553fbb962c4bcaf4c44ed5366f2ffeb180af35c3R44
- https://github.com/authzed/spicedb/pull/268/files#diff-e24eb2e567330120b1257156553fbb962c4bcaf4c44ed5366f2ffeb180af35c3R481

This ensures that only "fully" resolved lookups are cached

jakedt: Thanks github, `lookup.go` was collapsed, and even your links didn't expand it.

jakedt (approved):

jakedt (dismissed):

cmd/spicedb/serve.go
```diff
@@ -315,16 +315,18 @@ func serveRun(cmd *cobra.Command, args []string) {
 		redispatch = remote.NewClusterDispatcher(v1.NewDispatchServiceClient(conn))
 	}
 
-	cachingRedispatch, err := caching.NewCachingDispatcher(redispatch, nil, "dispatch_client")
+	cachingRedispatch, err := caching.NewCachingDispatcher(nil, "local_caching")
```
ecordell, line 318: this isn't really `local` caching, right? metrics from this dispatcher will be requests that go through the dispatch api client?

cmd/spicedb/serve.go
```diff
@@ -315,16 +315,18 @@ func serveRun(cmd *cobra.Command, args []string) {
...
 	if err != nil {
-		log.Fatal().Err(err).Msg("failed to initialize redispatcher cache")
+		log.Fatal().Err(err).Msg("failed to initialize dispatcher cache")
 	}
+	cachingRedispatch.SetDelegate(redispatch)
```
ecordell, line 322: I think I'm missing the importance of separating out `SetDelegate`

cmd/spicedb/serve.go
josephschorr: Changed back

cmd/spicedb/serve.go
jakedt: We need to separate it from the constructor because the delegate needs this item to be set as `it's` delegate

ecordell: where is that happening?

ecordell (commented):
server initialization LGTM

jakedt (approved):

PR #269, opened 2021-11-09
fix watch index in the memdb datastore implementation

jonwhitty:
@jakedt this fixes the memdb Watch implementation. With this change, tuples that are written after the revision provided to the memdb Watch API will show up properly.

The current implementation starts the watch over the lower range of keys starting at one commit ahead of the revision specified by the client. By the time Watch is invoked the transaction counter will be 1 greater than the last committed index, so nothing prior will have a Watch over it


jakedt:
This test fails because the IDs for memdb aren't unified between namespaces and relationships. We should rewrite the ID generation code for memdb to use a single monotonically increasing transaction source, like our other datastores.

jakedt:
Opened #270 to track.

PR #271, opened 2021-11-09
Add revision support to v1alpha1 schema API

josephschorr:
Allows for reading a combined revision from ReadSchema, and using the same revision as a precondition to WriteSchema to determine if any of the types have changed since the last read occurred


internal/services/v1alpha1/v1alpharevision.go
```diff
@@ -0,0 +1,75 @@
...
+		msg.NsRevisions = append(msg.NsRevisions, &internal.NamespaceAndRevision{
+			NamespaceName: nsName,
+			Revision:      revision.String(),
+		})
+	}
+
+	sort.Sort(byName(msg.NsRevisions))
+
+	// NOTE: We use a Deterministic marshaller here to ensure the produced revision is the same
+	// across calls. Also note that this is not *guarenteed* to be the same across versions of
```
jzelinskie, line 34: spelling -> guaranteed

this is also kinda scary although I'm not sure if makes too much of a difference for our use cases

jzelinskie (changes requested):

internal/services/v1alpha1/v1alpharevision.go
josephschorr: Yeah

josephschorr:
Updated

internal/services/v1alpha1/schema.go
```diff
@@ -55,18 +59,27 @@ func NewSchemaServer(ds datastore.Datastore, prefixRequired PrefixRequiredOption
 
 func (ss *schemaServiceServer) ReadSchema(ctx context.Context, in *v1alpha1.ReadSchemaRequest) (*v1alpha1.ReadSchemaResponse, error) {
 	var objectDefs []string
+	revisions := map[string]datastore.Revision{}
```
jzelinskie, line 62: this could also use `make(map[string]datastore.Revision, len(in.GetObjectDefinitionNames())`

internal/services/v1alpha1/schema_test.go
```diff
@@ -141,6 +145,10 @@ func TestSchemaWriteAndReadBack(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, requestedObjectDefNames, writeResp.GetObjectDefinitionsNames())
 
+	rev, err := decodeV1Alpha1Revision(writeResp.ComputedDefinitionsRevision)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(rev))
```
jzelinskie, line 150: I should've clarified, I meant to use Len everywhere not just in that one place

josephschorr:
Updated

jzelinskie (dismissed):

jzelinskie (approved):

PR #278, opened 2021-11-10
Move golangci-lint timeout into config

jzelinskie:


jzelinskie:
Looks like it just gets OOM killed, haha. Doesn't seem to make _that_ much of a perf difference anyways.

josephschorr (approved):
LGTM

PR #284, opened 2021-11-12
use consistent-hash load balancer with kubernetes resolver for dispatch

ecordell:
This PR adds a new flag `--dispatch-upstream-addr` and optional `--dispatch-upstream-ca-path`

If set, spicedb will use the new consistent hash loadbalancer (which is just the existing consistent hashring, wired into grpc-go) to pick endpoints for a dispatch request. This is an alternative to configuring a client to watch the a Dispatch API provider like servok.

This also adds support for using the `kubernetes` name resolver strategy provided by [kuberesolver](https://github.com/sercand/kuberesolver).
 
 Example:
 
 ```sh
spicedb serve --grpc-preshared-key "somerandomkeyhere" --dispatch-upstream-addr=kubernetes://spicedb.dev:50053
 ```


pkg/balancer/balancer.go
```diff
@@ -0,0 +1,95 @@
+package balancer
```
jakedt, line 1: Rename this file to what type of balancer it is specifically, maybe `hashring.go`?

cmd/spicedb/serve.go
```diff
@@ -106,6 +111,7 @@ func registerServeCmd(rootCmd *cobra.Command) {
 	// Flags for configuring dispatch behavior
 	serveCmd.Flags().Uint32("dispatch-max-depth", 50, "maximum recursion depth for nested calls")
 	cobrautil.RegisterGrpcServerFlags(serveCmd.Flags(), "dispatch-cluster", "dispatch", ":50053", false)
+	serveCmd.Flags().String("dispatch-upstream-addr", "", "upstream grpc address to dispatch to")
```
jakedt, line 114: I think we should just drop everything servok and see what the minimal set of flags would be.

jakedt (changes requested):

cmd/spicedb/serve.go
ecordell: did it in a separate commit in case we want to revert

ecordell:
@jakedt also bumped the replicationFactor to a `uint16`

internal/dispatch/remote/cluster.go
```diff
@@ -31,6 +38,11 @@ func (cr *clusterDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
 		return &v1.DispatchCheckResponse{Metadata: emptyMetadata}, err
 	}
 
+	requestKey, err := protoMarshal.Marshal(req)
```
jakedt, line 41: I think we should manually be creating the request keys, not just using the entire dispatched proto.

jakedt (changes requested):
There is more servok-related code that we can kill now, right?

internal/dispatch/remote/cluster.go
ecordell: Could you double check the fields I included? I left out depth and limit but didn't fully convince myself that was okay.

cmd/spicedb/serve.go
```diff
@@ -275,51 +278,30 @@ func serveRun(cmd *cobra.Command, args []string) {
...
-			resolverConfig = consistentbackend.NewEndpointResolverNoTLS(resolverAddr)
-		}
+
+	// grpc consistent loadbalancer redispatch configuration
+	dispatchAddr := cobrautil.MustGetStringExpanded(cmd, "dispatch-upstream-addr")
+	if len(dispatchAddr) > 0 {
+		log.Info().Str("upstream", dispatchAddr).Msg("configuring grpc consistent load balancer for redispatch")
 
 		peerPSK := cobrautil.MustGetStringExpanded(cmd, "grpc-preshared-key")
 		peerCertPath := cobrautil.MustGetStringExpanded(cmd, "dispatch-cluster-tls-cert-path")
```
jakedt, line 288: We actually want a different config flag for the client certificate, and we want to use the CA as the certificate.

jakedt (changes requested):

internal/dispatch/remote/cluster.go
jakedt: I didn't realize the current impl just marshall's the protos, I wonder if that's contributing to our hit rate being low: https://github.com/authzed/spicedb/blob/e5ebe46d37a6a23965f97042161df979361855ab/internal/dispatch/client/consistentbackend/client.go#L200

proto/buf-generate.sh
```diff
@@ -1,3 +1,2 @@
 #!/bin/sh
-buf generate -o internal/proto buf.build/authzed/servok:4027ca77d1b04ea2888b00348c3fd3d9 --template $@
 buf generate -o internal/proto proto/internal --template $@
```
jzelinskie, line 2: If we're getting rid of the servok proto generation, you should delete its protos and you can get rid of this script entirely and just make the first line in buf.gen.yaml

```yaml
#!/usr/bin/env -S buf generate -o internal/proto proto/internal --template
```

cmd/spicedb/serve.go
ecordell: I didn't see any client cert config in `main`, just the CA config. The credentials here are configured the same as here: https://github.com/authzed/spicedb/blob/d6dec1252db5065d29d726f3c157462f0ca40fd6/internal/dispatch/client/consistentbackend/options.go#L62

(but I removed the dns name flag, it looked unnecessary to me based on how the `ServerName` field is treated if missing in grpc ClientHandshake: https://github.com/grpc/grpc-go/blob/master/credentials/tls.go#L76-L83

ecordell: Added a new `--dispatch-cluster-upstream-ca-path` flag to do this

jakedt (approved):

PR #289, opened 2021-11-16
Add proper dispatch and cached dispatch tracking

josephschorr:
Unify handling of response metadata in dispatch, add testing for dispatch and depth required and add cached dispatch counting


internal/dispatch/caching/caching.go
```diff
@@ -164,8 +165,11 @@ func (cd *CachingDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
 
 	// We only want to cache the result if there was no error
 	if err == nil {
-		toCache := checkResultEntry{computed, computed.Metadata.DepthRequired}
-		toCache.result.Metadata.DispatchCount = 0
+		adjustedComputed := proto.Clone(computed).(*v1.DispatchCheckResponse)
+		adjustedComputed.Metadata.CachedDispatchCount = adjustedComputed.Metadata.DispatchCount
+		adjustedComputed.Metadata.DispatchCount = 0
+
+		toCache := checkResultEntry{adjustedComputed, computed.Metadata.DepthRequired}
```
jakedt, line 172: Do we want to just also include the depth required in the response metadata, since there is a limit?

internal/graph/expand.go
```diff
@@ -59,7 +60,7 @@ func (ce *ConcurrentExpander) expandDirect(
 	return func(ctx context.Context, resultChan chan<- ExpandResult) {
 		requestRevision, err := decimal.NewFromString(req.Metadata.AtRevision)
 		if err != nil {
-			resultChan <- expandResultError(NewExpansionFailureErr(err), 0, 0)
```
jakedt, line 62: Why do the `(1 ,0)` and `(0, 0)` responses both use `emptyMetadata`?

internal/dispatch/caching/caching.go
```diff
@@ -164,8 +165,11 @@ func (cd *CachingDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
 
 	// We only want to cache the result if there was no error
 	if err == nil {
-		toCache := checkResultEntry{computed, computed.Metadata.DepthRequired}
-		toCache.result.Metadata.DispatchCount = 0
+		adjustedComputed := proto.Clone(computed).(*v1.DispatchCheckResponse)
+		adjustedComputed.Metadata.CachedDispatchCount = adjustedComputed.Metadata.DispatchCount
+		adjustedComputed.Metadata.DispatchCount = 0
```
jakedt, line 170: Either here or in the cluster dispatcher we want to set the dispatch count for things that are server side cached to `1`, vs `0` if it was in the client-side cache.

jakedt (changes requested):

internal/dispatch/graph/check_test.go
```diff
@@ -197,6 +197,92 @@ func TestMaxDepth(t *testing.T) {
 	require.Equal(v1.DispatchCheckResponse_UNKNOWN, checkResult.Membership)
 }
 
+func TestCheckMetadata(t *testing.T) {
+	type expected struct {
+		relation              string
+		isMember              bool
+		expectedDispatchCount int
+		expectedDepthRequired int
```
jzelinskie, line 205: Can we come up with a test where expectedDepthRequired and expectedDispatchCount are different values

internal/dispatch/caching/caching.go
josephschorr: Sure, we can use that directly if you like

internal/dispatch/caching/caching.go
josephschorr: Probably best to be adjusted in the cluster dispatcher... although that would require cloning the message again, which kind of sucks.

internal/dispatch/graph/check_test.go
josephschorr: Yeah, I'll add something that dispatches on the same level

internal/dispatch/caching/caching.go
josephschorr: I switched to use the metadata's var, but kept the entries in case there is anything we want to add the cache entries. Let me know if I should just store the response message directly in the cache as a pointer (which might have its own concerns)

internal/graph/expand.go
josephschorr: Because the `(1, 0)` was wrong after my change: it was adding `1` to the dispatch count, without a dispatch. Now the counter is handled at the entrypoint of the dispatch for each graph

internal/dispatch/caching/caching.go
josephschorr: @jakedt let me know which you prefer

jakedt: Dispatcher would seem "right" to me, but if we want the responses to match for local dispatch vs remote dispatch, maybe the cache entries should just default to `1` always.

internal/dispatch/caching/caching.go
jakedt: I think keeping the entry layer of indirection is fine.

internal/dispatch/graph/check_test.go
jakedt: Something with a TTU with two values would be good.

josephschorr: Should add another one? I added one already

internal/dispatch/caching/caching.go
josephschorr: So just set it to `1` here for dispatched, and always return that?

jakedt: Does the rationale seem sound to you?

internal/graph/graph.go
```diff
@@ -74,3 +74,50 @@ func max(x, y uint32) uint32 {
 	}
 	return x
 }
+
+var emptyMetadata = &v1.ResponseMeta{}
+
+func updateRespMetadata(existing *v1.ResponseMeta, responseMetadata *v1.ResponseMeta) *v1.ResponseMeta {
```
jakedt, line 80: Should we rename this to combineResponseMetadata? I assumed it updated the first parameter in place. We could also make it take a list, collect up any that need to be combined, and combine them all lazily, that would create less ResponseMeta garbage for `any` and `all` calls.

internal/dispatch/caching/caching.go
josephschorr: I believe so: If the local dispatch hits the cache, it has dispatched technically once, so it probably should return `1` from there.

internal/graph/graph.go
josephschorr: The downside of taking a list is that it makes it harder to pass around; we could create an intermediate struct, add the responses into that, and then combine as a final step. I'm all for renaming this though, as it is somewhat vague

josephschorr: Updated the name

jakedt (approved):

Issue #290, opened 2021-11-17
CLI command for validating schemas


jzelinskie:
We've had some discussions on improving the workflow for testing schemas in CI/CD environments rather than human-oriented environments (e.g. [the playground](https://play.authzed.com)). A CLI like `spicedb validate <config.yaml>` could satisfy this.

Open questions:
- Does this belong in SpiceDB or zed?
- Should validation be done "server-side" or "client-side"? "Server-side" would guarantee that validation is done with version of SpiceDB used by the server, rather than the version of the tool

PR #292, opened 2021-11-18
introduce validate command

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/290

## What

The purpose of this command is to take a playground file and run the assertions and validations defined.

The rationale is that schema development happens in the playground, but once the YAML is downloaded, there is nothing developers can do other than loading it with testserve command, or uploading it back to the playground. This attempts to reuse and run the assertions and validations as test-suite outside of the the playground, and in a programmatic way rather than only interactively. Rather than duplicating the same tests in the client application, the playground tests become the canonical representation for the business rules defined in the schema.

Example:
1. developers introduce changes in schema via the playground
2. YAML file is downloaded and persisted in git repository
3. changes are pushed, PR is opened, CI runs spicedb validate, demonstrating changes are sound.

## Assumptions
- Introducing a new CLI command is cool, exposing new API in the go code requires more consideration
- Version 2 of the Playground file is not really API, so instead of updating the public structures, in parsed the file in two phases: one time with the public stuff, and one with the v2 fields
- I'm not sure I got right the versioning strategy y'all have with the API. It sounds like `v0` is like "it's public, but may be broken anytime". I assumed it's OK to expose methods reusing v0 types, but would definitely appreciate some guidance here  

## Features
- accepts multiple playground files as input
- process returns 0 if valid, non-zero if invalid
- errors by line and message are logged (e.g. can be surfaced in the GitHub PR)

## TODO
- Planning to add tests if this is the design seems sound

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 dist/
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 dist/
-.idea
--- a/cmd/spicedb/main.go
+++ b/cmd/spicedb/main.go
@@ -18,6 +18,7 @@ func main() {
 	registerHeadCmd(rootCmd)
 	registerDeveloperServiceCmd(rootCmd)
 	registerTestserverCmd(rootCmd)
+	registerValidateCmd(rootCmd)
 
 	_ = rootCmd.Execute()
 }
--- a/cmd/spicedb/validate.go
+++ b/cmd/spicedb/validate.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
+	v0svc "github.com/authzed/spicedb/internal/services/v0"
+	"github.com/authzed/spicedb/pkg/validationfile"
+	"github.com/jzelinskie/cobrautil"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+)
+
+func registerValidateCmd(rootCmd *cobra.Command) {
+	testserveCmd := &cobra.Command{
+		Use:               "validate",
+		Short:             "validate a playground YAML file",
+		Long:              "validate a playground YAML file.",
+		PersistentPreRunE: persistentPreRunE,
+		Run:               runValidate,
+	}
+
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
 dist/
+.idea
```
jzelinskie, line 2: I'd put this in your global ~/.gitignore -- I don't want to pollute our .gitignore with editor-specific configurations.

josephschorr:
> Version 2 of the Playground file is not really API, so instead of updating the public structures, in parsed the file in two phases: one time with the public stuff, and one with the v2 fields

We should turn this into a proper package function and make use of it here and in playground

> I'm not sure I got right the versioning strategy y'all have with the API. It sounds like v0 is like "it's public, but may be broken anytime". I assumed it's OK to expose methods reusing v0 types, but would definitely appreciate some guidance here

Using v0 types internally is fine, but I think we should extract out the work developer API does for validation into a validation `pkg`, and then have both the CLI command and the developer API invoke it

vroldanbet:
@josephschorr thanks for the feedback!

I explored your suggestions in https://github.com/authzed/spicedb/pull/292/commits/cdff9151b8b47a7f13ed8ec882d35b62f8a22473, and I must admit I struggled a bit to refactor the code. It seemed like having that as part of the `devServer` led to better encapsulation. I'm not convinced the refactor has led to a proper separation of responsibilities between the two packages. I found myself several times in package loops that hinted me such separation wasn't adequate.

Please let me know what you think. If you have suggestions/direction, I'm happy to rework it.

josephschorr:
@vroldanbet I see what you mean: I think the major dependency are the `v0` developer types we're using as inputs and outputs for the API. We likely need to change the validation stuff to define its own structs for input and output, and then have the developer context translate those to its own types. It'll be more verbose, but should break the dependency.

vroldanbet:
>@vroldanbet I see what you mean: I think the major dependency are the v0 developer types we're using as inputs and outputs for the API. We likely need to change the validation stuff to define its own structs for input and output, and then have the developer context translate those to its own types. It'll be more verbose, but should break the dependency.

@josephschorr I'm OK with that direction! 👍🏻 Can give it a go

josephschorr:
@vroldanbet Sounds good. Let me know when to take a look :)

josephschorr:
Closed in favor of https://github.com/authzed/zed/pull/86

PR #294, opened 2021-11-23
Properly calculate virtualnode ids for uint16 replicationFactor

ecordell:
Missed this when bumping the replication factor to a `uint16`.

### DIFF
--- a/pkg/consistent/hashring.go
+++ b/pkg/consistent/hashring.go
@@ -78,14 +78,14 @@ func (h *Hashring) Add(member Member) error {
 		nil,
 	}
 
-	// virtualNodeBuffer is a 9-byte array, where 8 bytes are the hash value of the
-	// member key, and the final byte is an offset of the virtual node itself. This
-	// value is then hashed to get the final hash value of the virtual node.
-	virtualNodeBuffer := make([]byte, 9)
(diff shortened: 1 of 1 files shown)


pkg/consistent/hashring.go
```diff
@@ -78,14 +78,14 @@ func (h *Hashring) Add(member Member) error {
...
 	}
 
-	// virtualNodeBuffer is a 9-byte array, where 8 bytes are the hash value of the
-	// member key, and the final byte is an offset of the virtual node itself. This
-	// value is then hashed to get the final hash value of the virtual node.
-	virtualNodeBuffer := make([]byte, 9)
+	// virtualNodeBuffer is a 10-byte array, where 8 bytes are the hash value of
+	// the member key, and the final 2 bytes are an offset of the virtual node
+	// itself. This value is then hashed to get the final hash value of the virtual node.
+	virtualNodeBuffer := make([]byte, 10)
```
josephschorr, line 84: Could we somehow calculate this based on a data type we define somewhere, so we can't accidentally forget again?

jakedt: Not at compile time, I don't think.

jakedt (approved):

PR #295, opened 2021-11-23
Add dispatch and cached dispatch counts to response trailer metadata and prometheus

josephschorr:
All dispatched API operations will now contain the count of dispatched calls and cached dispatched calls, for complexity tracking via the new responsemeta package

### DIFF
--- a/internal/middleware/usagemetrics/usagemetrics.go
+++ b/internal/middleware/usagemetrics/usagemetrics.go
@@ -1,0 +1,138 @@
+package usagemetrics
+
+import (
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/authzed/authzed-go/pkg/responsemeta"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
+	prometheus "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc"
+
+	dispatch "github.com/authzed/spicedb/internal/proto/dispatch/v1"
+)
+
+var dispatchBuckets = []float64{1, 5, 10, 25, 50, 100, 250}
+
+var dispatchedCountHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace: "spicedb",
+	Subsystem: "services",
+	Name:      "dispatched_count_histogram",
+	Help:      "dispatch count per api call distribution in seconds.",
+	Buckets:   dispatchBuckets,
+}, []string{"method"})
+
+var cachedCountHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace: "spicedb",
+	Subsystem: "services",
+	Name:      "cached_count_histogram",
+	Help:      "dispatches avoid by caching per api call in seconds.",
+	Buckets:   dispatchBuckets,
+}, []string{"method"})
+
+var dispatchedCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "services",
+	Name:      "dispatched_counter",
+	Help:      "dispatch counts.",
+}, []string{"method"})
+
+var cachedCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "services",
+	Name:      "cached_counter",
+	Help:      "dispatches avoid by caching.",
+}, []string{"method"})
+
+type reporter struct{}
+
+func (r *reporter) ServerReporter(ctx context.Context, callMeta interceptors.CallMeta) (interceptors.Reporter, context.Context) {
+	_, methodName := interceptors.SplitMethodName(callMeta.FullMethod())
+	ctx = ContextWithHandle(ctx)
+	return &serverReporter{ctx: ctx, methodName: methodName}, ctx
+}
+
+type serverReporter struct {
+	interceptors.NoopReporter
+	ctx        context.Context
+	methodName string
+}
+
+func (r *serverReporter) PostCall(err error, rpcDuration time.Duration) {
+	responseMeta := FromContext(r.ctx)
+	if responseMeta != nil {
+		err := annotateAndReportForMetadata(r.ctx, r.methodName, responseMeta)
+		if err != nil {
+			log.Ctx(r.ctx).Err(err).Msg("could not report metadata")
+		}
+	}
+}
+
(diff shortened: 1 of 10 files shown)


pkg/responsemeta/responsemeta.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
+	dispatch "github.com/authzed/spicedb/internal/proto/dispatch/v1"
+)
+
+// NewHandler creates a new response metadata handler for setting the response
+// metadata into the trailer of an API request.
+func NewHandler(ctx context.Context) *ResponseMetadataHandler {
```
jakedt, line 17: `NewUsageHandler` or `NewMeteringHandler`

pkg/responsemeta/responsemeta.go
```diff
@@ -0,0 +1,111 @@
...
+	return &ResponseMetadataHandler{
+		ctx:    ctx,
+		wasSet: false,
+	}
+}
+
+// ResponseMetadataHandler is a helper struct for tracking whether response
+// metadata was written to the trailer.
+type ResponseMetadataHandler struct {
```
jakedt, line 26: `UsageHandler` or `MeteringHandler`

pkg/responsemeta/responsemeta.go
```diff
@@ -0,0 +1,111 @@
...
+	CachedOperationsCount ResponseMetadataKey = "io.spicedb.respmeta.cachedoperationscount"
+)
+
+// setResponseMetadata sets the external response metadata, extracted from the dispatch
+// response metadata, on the given context.
+func setResponseMetadata(ctx context.Context, dispatchMetadata *dispatch.ResponseMeta) error {
+	trailer := metadata.Pairs(
+		// DispatchedOperationsCount
+		string(DispatchedOperationsCount),
+		fmt.Sprintf("%d", dispatchMetadata.DispatchCount),
```
jakedt, line 76: `strconv.Itoa`?

pkg/responsemeta/responsemeta.go
```diff
@@ -0,0 +1,111 @@
...
+		string(CachedOperationsCount),
+		"0",
+	)
+	return grpc.SetTrailer(ctx, trailer)
+}
+
+// GetIntResponseMetadata retrieves an integer value for the given key in the trailer
+// metadata of a SpiceDB API response.
+func GetIntResponseMetadata(trailer metadata.MD, key ResponseMetadataKey) (int, error) {
+	values := trailer.Get(string(key))
```
jakedt, line 102: If you can tell if they are set by the presence of this trailer, why the need to track whether they've been set separately through a bool? In fact, why do you need a class at all? Can't you just use the metadata associated with the context?

jakedt (changes requested):

pkg/responsemeta/responsemeta.go
josephschorr: I dislike both names because they aren't being used for either system here; I originally had `NewResponseMetaHandler`, but then we have `responemeta.NewResponseMetaHandler`, which I thought was too verbose

pkg/responsemeta/responsemeta.go
josephschorr: I tend to prefer using `fmt` for converting to strings, but I can switch if you like

pkg/responsemeta/responsemeta.go
josephschorr: We don't have access to the trailer when setting the data, only when reading in the client.

pkg/responsemeta/responsemeta.go
jakedt: https://gist.github.com/evalphobia/caee1602969a640a4530

josephschorr: Updated

josephschorr:
Added prom metrics as well

internal/services/v1/permissions.go
```diff
@@ -45,6 +50,8 @@ func (ps *permissionServer) CheckPermission(ctx context.Context, req *v1.CheckPe
 			Relation:  normalizeSubjectRelation(req.Subject),
 		},
 	})
+	metaHandler.SetFor(cr.Metadata)
+	shared.UpdateServiceMetrics("v1.CheckPermission", cr.Metadata)
```
jakedt, line 54: Should we be counting/returning these on read/write/delete operations as well? At least read?

pkg/responsemeta/responsemeta.go
jakedt: We decided that this either needed to be renamed or the fields made dynamic. I'm still in favor of it being renamed. `responsemeta.NewHandler` tells me nothing about what it's going to do.

jakedt (changes requested):

internal/services/v1/permissions.go
josephschorr: Right now, they'd all be zero, since there are no dispatches involved. Maybe a different set of metrics?

pkg/responsemeta/responsemeta.go
josephschorr: I'm open to other suggestions. `responsemeta.NewTrailerHandler`? `responsemeta.NewTrailerReporter`?

internal/middleware/usagemetrics/usagemetrics.go
```diff
@@ -0,0 +1,138 @@
+package usagemetrics
```
jakedt, line 1: This should be moved to `pkg` so that we can use it from other projects.

internal/middleware/usagemetrics/usagemetrics.go
```diff
@@ -0,0 +1,138 @@
...
+	"google.golang.org/grpc"
+
+	dispatch "github.com/authzed/spicedb/internal/proto/dispatch/v1"
+	"github.com/authzed/spicedb/pkg/responsemeta"
+)
+
+var dispatchBuckets = []float64{1, 5, 10, 25, 50, 100, 250}
+
+var dispatchedCountHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace: "spicedb",
```
jakedt, line 21: If this is made generic/importable, do these values make sense? should they be pluggable? is it enough for the instantiator to be able to supply another label value?

internal/services/v1/permissions.go
jakedt: We can handle this in a follow-up.

internal/middleware/usagemetrics/usagemetrics.go
jakedt: Actually I take this back, this isn't likely to be the method we would use.

jakedt (approved):

PR #297, opened 2021-11-24
Add follower read delay option

samkim:
Add a new serve option to specify a read delay to enable [crdb follower reads](https://www.cockroachlabs.com/docs/stable/follower-reads.html). 

Follower reads will trade off a bit of staleness for reduced read latencies and hopefully also extra reliability as reads can go to a replica. 

Note: the option is set to zero by default and the default crdb delay value is 4.8s.

### DIFF
--- a/cmd/spicedb/serve.go
+++ b/cmd/spicedb/serve.go
@@ -83,7 +83,7 @@ func registerServeCmd(rootCmd *cobra.Command) {
 	serveCmd.Flags().Duration("datastore-gc-max-operation-time", 1*time.Minute, "maximum amount of time a garbage collection pass can operate before timing out (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-revision-fuzzing-duration", 5*time.Second, "amount of time to advertize stale revisions")
 	// See crdb doc for info about follower reads and how it is configured: https://www.cockroachlabs.com/docs/stable/follower-reads.html
-	serveCmd.Flags().Duration("datastore-follower-read-delay-duration", 0*time.Second, "amount of time to subtract from non-sync revision timestamps to ensure they are sufficiently in the past to enable follower reads (cockroach driver only)")
+	serveCmd.Flags().Duration("datastore-follower-read-delay-duration", 4_800*time.Millisecond, "amount of time to subtract from non-sync revision timestamps to ensure they are sufficiently in the past to enable follower reads (cockroach driver only)")
 	serveCmd.Flags().String("datastore-query-split-size", common.DefaultSplitAtEstimatedQuerySize.String(), "estimated number of bytes at which a query is split when using a remote datastore")
 	serveCmd.Flags().StringSlice("datastore-bootstrap-files", []string{}, "bootstrap data yaml files to load")
 	serveCmd.Flags().Bool("datastore-bootstrap-overwrite", false, "overwrite any existing data with bootstrap data")
--- a/cmd/spicedb/serve.go
+++ b/cmd/spicedb/serve.go
@@ -82,7 +82,8 @@ func registerServeCmd(rootCmd *cobra.Command) {
 	serveCmd.Flags().Duration("datastore-gc-interval", 3*time.Minute, "amount of time between passes of garbage collection (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-gc-max-operation-time", 1*time.Minute, "maximum amount of time a garbage collection pass can operate before timing out (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-revision-fuzzing-duration", 5*time.Second, "amount of time to advertize stale revisions")
-	serveCmd.Flags().Duration("datastore-follower-read-delay-duration", 0*time.Second, "amount of time to use as a delay to enable follower reads (cockroach driver only)")
+	// See crdb doc for info about follower reads and how it is configured: https://www.cockroachlabs.com/docs/stable/follower-reads.html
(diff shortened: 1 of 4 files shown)


internal/datastore/crdb/crdb_test.go
```diff
@@ -78,6 +82,44 @@ func TestCRDBDatastore(t *testing.T) {
...
+func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
+	followerReadDelay := time.Duration(4.8 * float64(time.Second))
+	gcWindow := 100 * time.Second
+
+	quantizationDurations := []time.Duration{
+		0 * time.Second,
+		100 * time.Millisecond,
+	}
+	for _, quantization := range quantizationDurations {
+		t.Run(fmt.Sprintf("followReadDelayWithQuantization%s", quantization), func(t *testing.T) {
```
josephschorr, line 94: We tend to use underscores or spaces in generated test names

internal/datastore/crdb/crdb.go
```diff
@@ -195,7 +199,8 @@ func (cds *crdbDatastore) Revision(ctx context.Context) (datastore.Revision, err
 	}
 
 	// Round the revision down to the nearest quantization
-	crdbNow := nowHLC.IntPart()
+	// Apply a delay to enable historial reads
+	crdbNow := nowHLC.IntPart() - cds.followerReadDelayNanos
```
josephschorr, line 203: Can we link to the CRDB documentation on historical reads and how this works?

internal/datastore/crdb/crdb.go
```diff
@@ -195,7 +199,8 @@ func (cds *crdbDatastore) Revision(ctx context.Context) (datastore.Revision, err
...
 	quantized := crdbNow
 	if cds.quantizationNanos > 0 {
 		quantized -= (crdbNow % cds.quantizationNanos)
```
josephschorr, line 206: Is there a risk that the quantization here pushes the window just slightly outside of the follower read and, if so, should we adjust the follower read delay accordingly?

cmd/spicedb/serve.go
```diff
@@ -82,6 +82,7 @@ func registerServeCmd(rootCmd *cobra.Command) {
 	serveCmd.Flags().Duration("datastore-gc-interval", 3*time.Minute, "amount of time between passes of garbage collection (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-gc-max-operation-time", 1*time.Minute, "maximum amount of time a garbage collection pass can operate before timing out (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-revision-fuzzing-duration", 5*time.Second, "amount of time to advertize stale revisions")
+	serveCmd.Flags().Duration("datastore-follower-read-delay-duration", 0*time.Second, "amount of time to use as a delay to enable follower reads (cockroach driver only)")
```
josephschorr, line 85: "amount of time to subtract from non-sync revision timestamps to ensure they are sufficiently in the past to enable follower (historical) reads (cockroach driver only)"

josephschorr: We should also ideally link to documentation so people know how to configure this

internal/datastore/crdb/crdb.go
samkim: Any quantization will only add additional delay so the timestamp used for read will always be _at least_ the delay required. We could do some adjustment to reduce the overall amount of staleness by incorporating the quantization amount into the delay though but the difference should be minimal relative to the ~5s delay.

cmd/spicedb/serve.go
jakedt: We should probably also default it to 4.8s

samkim: It is only applied to the crdb datastore init so that's reasonable.

internal/datastore/crdb/crdb_test.go
```diff
@@ -109,5 +151,5 @@ func newTester(containerOpts *dockertest.RunOptions, creds string, portNum uint1
 		}
 	}
 
-	return &sqlTest{conn: conn, port: port, cleanup: cleanup, creds: creds}
+	return &sqlTest{conn: conn, port: port, cleanup: cleanup, creds: creds, followerReadDelay: 0 * time.Second}
```
jakedt, line 154: You could technically skip this as `0s` is the zero type for a `time.Duration`

cmd/spicedb/serve.go
```diff
@@ -82,6 +82,8 @@ func registerServeCmd(rootCmd *cobra.Command) {
 	serveCmd.Flags().Duration("datastore-gc-interval", 3*time.Minute, "amount of time between passes of garbage collection (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-gc-max-operation-time", 1*time.Minute, "maximum amount of time a garbage collection pass can operate before timing out (postgres driver only)")
 	serveCmd.Flags().Duration("datastore-revision-fuzzing-duration", 5*time.Second, "amount of time to advertize stale revisions")
+	// See crdb doc for info about follower reads and how it is configured: https://www.cockroachlabs.com/docs/stable/follower-reads.html
+	serveCmd.Flags().Duration("datastore-follower-read-delay-duration", time.Duration(4.8*float64(time.Second)), "amount of time to subtract from non-sync revision timestamps to ensure they are sufficiently in the past to enable follower reads (cockroach driver only)")
```
jakedt, line 86: You could make this `4_800 * time.Millisecond` without any other casts.

jakedt (changes requested):

jakedt (approved):

Issue #209, opened 2021-10-23
Test Flake: TestCRDBDatastore/TestRevisionFuzzing/fuzzing100ms


ecordell:
```
--- FAIL: TestCRDBDatastore (57.38s)
    --- FAIL: TestCRDBDatastore/TestRevisionFuzzing (0.78s)
        --- FAIL: TestCRDBDatastore/TestRevisionFuzzing/fuzzing100ms (0.40s)
            revisions.go:77: 
                	Error Trace:	revisions.go:77
                	Error:      	Should be true
                	Test:       	TestCRDBDatastore/TestRevisionFuzzing/fuzzing100ms
```
First seen in #208

ecordell:
Again: https://github.com/authzed/spicedb/runs/4472166539?check_suite_focus=true

josephschorr:
Haven't see this in a while

PR #298, opened 2021-11-29
Add retries with a newly acquired connection

samkim:
* Add tx reset for "ambiguous result errors" which includes connection closed errors. 
* Tx retries will attempt to reset the tx if unable to rollback

### DIFF
--- a/internal/datastore/crdb/tx.go
+++ b/internal/datastore/crdb/tx.go
@@ -149,31 +149,21 @@ func resetExecution(ctx context.Context, conn conn, tx *pgx.Tx, txOptions pgx.Tx
 }
 
 func retriable(ctx context.Context, err error) bool {
-	errorCode, err := sqlErrorCode(ctx, err)
-	if err != nil {
-		log.Ctx(ctx).Error().Err(err).Msg("error not retriable")
-		return false
-	}
-	return errorCode == crdbRetryErrCode
+	return sqlErrorCode(ctx, err) == crdbRetryErrCode
 }
 
 func resetable(ctx context.Context, err error) bool {
-	errorCode, err := sqlErrorCode(ctx, err)
-	if err != nil {
-		log.Ctx(ctx).Info().Err(err).Msg("error not resetable")
-		return false
-	}
-
 	// Ambiguous result error includes connection closed errors
 	// https://www.cockroachlabs.com/docs/stable/common-errors.html#result-is-ambiguous
-	return errorCode == crdbAmbiguousErrorCode
+	return sqlErrorCode(ctx, err) == crdbAmbiguousErrorCode
 }
 
-func sqlErrorCode(ctx context.Context, err error) (string, error) {
+func sqlErrorCode(ctx context.Context, err error) string {
 	var pgerr *pgconn.PgError
 	if !errors.As(err, &pgerr) {
-		return "", err
+		log.Ctx(ctx).Info().Err(err).Msg("couldn't determine a sqlstate error code")
+		return ""
 	}
 
-	return pgerr.SQLState(), nil
+	return pgerr.SQLState()
 }
--- a/internal/datastore/crdb/tx.go
+++ b/internal/datastore/crdb/tx.go
@@ -12,10 +12,16 @@ import (
 )
 
 const (
-	crdbRetryErrCode       = "40001"
+	// https://www.cockroachlabs.com/docs/stable/common-errors.html#restart-transaction
+	crdbRetryErrCode = "40001"
+	// https://www.cockroachlabs.com/docs/stable/common-errors.html#result-is-ambiguous
 	crdbAmbiguousErrorCode = "40003"
 	errUnableToRetry       = "failed to retry conflicted transaction: %w"
 	errReachedMaxRetry     = "maximum retries reached"
+
+	sqlRollback         = "ROLLBACK TO SAVEPOINT cockroach_restart"
+	sqlSavepoint        = "SAVEPOINT cockroach_restart"
+	sqlReleaseSavepoint = "RELEASE SAVEPOINT cockroach_restart"
 )
 
 var retryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
@@ -65,7 +71,7 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
 		_ = tx.Rollback(ctx)
 	}()
 
-	if _, err = tx.Exec(ctx, "SAVEPOINT cockroach_restart"); err != nil {
+	if _, err = tx.Exec(ctx, sqlSavepoint); err != nil {
 		return
 	}
 
@@ -85,16 +91,16 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
 
 		// RELEASE SAVEPOINT itself can fail, in which case the entire
 		// transaction needs to be retried
-		if _, err := tx.Exec(ctx, "RELEASE SAVEPOINT cockroach_restart"); err != nil {
+		if _, err := tx.Exec(ctx, sqlReleaseSavepoint); err != nil {
 			return err
 		}
 		return nil
 	}
 
-	for i = 0; i < maxRetries+1; i++ {
+	for i = 0; i <= maxRetries; i++ {
 		if err = releasedFn(tx); err != nil {
 			if retriable(ctx, err) {
-				if _, retryErr := tx.Exec(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); retryErr != nil {
+				if _, retryErr := tx.Exec(ctx, sqlRollback); retryErr != nil {
 					// Attempt to reset on failed retries
 					newTx, resetErr := resetExecution(ctx, conn, &tx, txOptions)
 					if resetErr != nil {
@@ -120,7 +126,7 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
 	return errors.New(errReachedMaxRetry)
 }
 
-// tx will be rolled back and a new tx will be started
+// resetExecution attempts to rollback the given tx, begins a new tx with a new connection, and creates a savepoint.
 func resetExecution(ctx context.Context, conn conn, tx *pgx.Tx, txOptions pgx.TxOptions) (newTx pgx.Tx, err error) {
 	err = (*tx).Rollback(ctx)
 	if err != nil {
--- a/internal/datastore/crdb/tx.go
+++ b/internal/datastore/crdb/tx.go
@@ -81,16 +81,18 @@ func executeWithResets(ctx context.Context, conn conn, txOptions pgx.TxOptions, 
 
 	var tx pgx.Tx
 	defer func() {
-		if err == nil {
-			commitErr := tx.Commit(ctx)
-			if commitErr == nil {
-				return
+		if tx != nil {
+			if err == nil {
+				commitErr := tx.Commit(ctx)
+				if commitErr == nil {
+					return
+				}
+				log.Err(commitErr).Msg("failed tx commit")
 			}
-			log.Err(commitErr).Msg("failed tx commit")
-		}
 
-		if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
-			log.Err(rollbackErr).Msg("error during tx rollback")
+			if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
+				log.Err(rollbackErr).Msg("error during tx rollback")
+			}
 		}
 	}()
 
@@ -111,7 +113,7 @@ func executeWithResets(ctx context.Context, conn conn, txOptions pgx.TxOptions, 
 		err = txErr
 
 		if resetable(ctx, err) {
-			log.Err(err).Msg("resettable error, reconnecting with new transaction")
+			log.Err(err).Msg("resettable error, will attempt to reset tx")
 			resets++
 			continue
 		}
@@ -159,6 +161,7 @@ func executeWithRetries(ctx context.Context, currentTx pgx.Tx, fn transactionFn,
 
 // resetExecution attempts to rollback the given tx, begins a new tx with a new connection, and creates a savepoint.
 func resetExecution(ctx context.Context, conn conn, tx pgx.Tx, txOptions pgx.TxOptions) (newTx pgx.Tx, err error) {
+	log.Info().Msg("attempting to initialize new tx")
 	if tx != nil {
 		err = tx.Rollback(ctx)
 		if err != nil {
--- a/internal/datastore/crdb/tx.go
+++ b/internal/datastore/crdb/tx.go
@@ -7,14 +7,16 @@ import (
 
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog/log"
 )
 
 const (
-	crdbRetryErrCode   = "40001"
-	errUnableToRetry   = "failed to retry conflicted transaction: %w"
-	errReachedMaxRetry = "maximum retries reached"
+	crdbRetryErrCode       = "40001"
+	crdbAmbiguousErrorCode = "40003"
+	errUnableToRetry       = "failed to retry conflicted transaction: %w"
+	errReachedMaxRetry     = "maximum retries reached"
 )
 
 var retryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
@@ -27,11 +29,7 @@ func init() {
 	prometheus.MustRegister(retryHistogram)
 }
 
-// conn is satisfied by both pgx.conn and pgxpool.Pool.
-type conn interface {
-	Begin(context.Context) (pgx.Tx, error)
-	BeginTx(context.Context, pgx.TxOptions) (pgx.Tx, error)
-}
+type conn = *pgxpool.Pool
 
 type transactionFn func(tx pgx.Tx) error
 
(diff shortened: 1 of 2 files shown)


internal/datastore/crdb/tx.go
```diff
@@ -85,13 +83,26 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
+				continue
+			} else if resetable(ctx, err) {
+				// Close connection and retry with newly acquired connection
+				tx.Rollback(ctx)
+				tx.Conn().Close(ctx)
+				newConn, acqErr := conn.Acquire(ctx)
+				if acqErr != nil {
+					return fmt.Errorf(errUnableToRetry, err)
+				}
+				tx, err = newConn.BeginTx(ctx, txOptions)
```
ecordell, line 99: since this is all the same setup as the beginning of the fn, maybe there's a nice way to structure `resets` as a wrapper around `retries`?

internal/datastore/crdb/tx.go
```diff
@@ -106,3 +117,14 @@ func retriable(ctx context.Context, err error) bool {
 	}
 	return pgerr.SQLState() == crdbRetryErrCode
 }
+
+func resetable(ctx context.Context, err error) bool {
```
ecordell, line 121: this is a nit, but we could probably use a `sqlState(context, error)` fn that just extracts the error code (that resetable and retriable can use)

internal/datastore/crdb/tx.go
samkim: Hmm yeah, let me try wrapping it and incorporating the above feedback about reset failed retries if possible

internal/datastore/crdb/tx.go
```diff
@@ -27,11 +29,7 @@ func init() {
 	prometheus.MustRegister(retryHistogram)
 }
 
-// conn is satisfied by both pgx.conn and pgxpool.Pool.
-type conn interface {
```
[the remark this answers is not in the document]
ecordell: Ah makes sense. Yeah, not much point if regular connections can't be reset.

internal/datastore/crdb/tx.go
```diff
@@ -12,9 +12,10 @@ import (
 )
 
 const (
-	crdbRetryErrCode   = "40001"
-	errUnableToRetry   = "failed to retry conflicted transaction: %w"
-	errReachedMaxRetry = "maximum retries reached"
+	crdbRetryErrCode       = "40001"
```
josephschorr, line 15: Can we link to some documentation on these codes?

internal/datastore/crdb/tx.go
```diff
@@ -83,26 +91,68 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
 		return nil
 	}
 
-	for i = 0; i < maxRetries; i++ {
+	for i = 0; i < maxRetries+1; i++ {
```
josephschorr, line 94: maybe just do `<=` if we're not using the index?

internal/datastore/crdb/tx.go
josephschorr, line 15: Can we link to some documentation on these codes?

internal/datastore/crdb/tx.go
```diff
@@ -83,26 +91,68 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
 		if err = releasedFn(tx); err != nil {
-			if !retriable(ctx, err) {
-				return err
+			if retriable(ctx, err) {
+				if _, retryErr := tx.Exec(ctx, "ROLLBACK TO SAVEPOINT cockroach_restart"); retryErr != nil {
```
josephschorr, line 97: Can we put the savepoint name into a const?

josephschorr, line 97: Can we put the savepoint name into a const?

josephschorr, line 97: maybe move this logic into a function and then just do `rtx, rerr := retry(...)` and assign the `rtx` to `tx` if the retry succeeded? It would then allow the reset to be its own thing as well.

internal/datastore/crdb/tx.go
```diff
@@ -83,26 +91,68 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
-			}
-			continue
+			return err
 		}
 		return nil
 	}
 	return errors.New(errReachedMaxRetry)
 }
 
+// tx will be rolled back and a new tx will be started
```
josephschorr, line 123: Please use doc comment format if commenting: `resetExecution will ....`

internal/datastore/crdb/tx.go
```diff
@@ -83,26 +91,68 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
+	return sqlErrorCode(ctx, err) == crdbAmbiguousErrorCode
+}
+
+func sqlErrorCode(ctx context.Context, err error) string {
 	var pgerr *pgconn.PgError
 	if !errors.As(err, &pgerr) {
-		log.Ctx(ctx).Error().Err(err).Msg("error not retriable")
-		return false
+		log.Ctx(ctx).Info().Err(err).Msg("couldn't determine a sqlstate error code")
+		return ""
```
josephschorr, line 154: All errors are strings?

internal/datastore/crdb/tx_test.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+const (
+	testUserNamespace = "test/user"
+)
+
+var testUserNS = namespace.Namespace(testUserNamespace)
+
+// Copied from crdb_test.go
+func (st sqlTest) newCRDB() (*crdbDatastore, error) {
```
josephschorr, line 29: Can we share it somehow?

internal/datastore/crdb/tx_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	newDBName := "db" + uniquePortion
+
+	_, err = st.conn.Exec(context.Background(), "CREATE DATABASE "+newDBName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create database: %w", err)
+	}
+
+	connectStr := fmt.Sprintf(
+		"postgres://%s@localhost:%s/%s?sslmode=disable",
```
josephschorr, line 43: why using postgres protocol vs crdb?

internal/datastore/crdb/tx_test.go
```diff
@@ -0,0 +1,163 @@
...
+			require.True(ok)
+
+			// WriteNamespace utilizes execute so we'll use it
+			revision, err := ds.WriteNamespace(ctx, testUserNS)
+			if tt.expectedError != nil {
+				require.Error(err)
+			} else {
+				require.NoError(err)
+			}
+			require.NotNil(revision)
```
josephschorr, line 157: Can we check that it wrote in the non-error case, and that it did not in the error case?

internal/datastore/crdb/tx.go
```diff
@@ -58,15 +75,37 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
 	}()
 
-	if _, err = tx.Exec(ctx, "SAVEPOINT cockroach_restart"); err != nil {
+	for retries = 0; retries < maxRetries; retries++ {
+		tx, err = resetExecution(ctx, conn, &tx, txOptions)
+		if err != nil {
+			resets++
+			continue
+		}
+		var attempts int
```
josephschorr, line 84: newline before this line

internal/datastore/crdb/tx.go
```diff
@@ -58,15 +75,37 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
+		attempts, err = executeWithRetries(ctx, &tx, fn, maxRetries-retries)
+		if err != nil {
```
josephschorr, line 86:
```
if err == nil {
  return
}
```

internal/datastore/crdb/tx.go
```diff
@@ -39,17 +54,19 @@ type executeTxRetryFunc func(context.Context, conn, pgx.TxOptions, transactionFn
...
+	}()
+
 	var tx pgx.Tx
-	tx, err = conn.BeginTx(ctx, txOptions)
-	if err != nil {
-		return err
-	}
 	defer func() {
 		if err == nil {
 			_ = tx.Commit(ctx)
```
josephschorr, line 72: What does `tx.Commit` return?

internal/datastore/crdb/tx.go
```diff
@@ -58,15 +75,37 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
+			resets++
+			continue
+		}
+
+		// non-resettable error
+		err = errors.New(errUnableToReset)
 		return
 	}
+	err = errors.New(errReachedMaxRetry)
+	return
```
josephschorr, line 103: no need for a `return` here

internal/datastore/crdb/tx.go
```diff
@@ -58,15 +75,37 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
+}
 
-	var i int
-	defer func() {
-		retryHistogram.Observe(float64(i))
-	}()
-
+// executeWithRetries executes the transaction fn and attempts to retry up to maxRetries.
+// adapted from https://github.com/cockroachdb/cockroach-go
```
josephschorr, line 107: Can you link the specific place?

internal/datastore/crdb/tx.go
```diff
@@ -77,32 +116,65 @@ func execute(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transac
...
-		return nil
+		return i, nil
 	}
-	return errors.New(errReachedMaxRetry)
+	return i, errors.New(errReachedMaxRetry)
+}
+
+// resetExecution attempts to rollback the given tx, begins a new tx with a new connection, and creates a savepoint.
+func resetExecution(ctx context.Context, conn conn, tx *pgx.Tx, txOptions pgx.TxOptions) (newTx pgx.Tx, err error) {
+	if *tx != nil {
```
josephschorr, line 144: is `pgx.Tx` itself a pointer type? If so, why are we taking in a pointer to the pointer?

internal/datastore/crdb/tx.go
```diff
@@ -37,19 +50,33 @@ type transactionFn func(tx pgx.Tx) error
 
 type executeTxRetryFunc func(context.Context, conn, pgx.TxOptions, transactionFn) error
 
+type RetryError struct {
```
josephschorr, line 53: Doc comment on exported type and any methods on it

internal/datastore/crdb/tx.go
josephschorr: We don't need to handle that error?

internal/datastore/crdb/tx_test.go
josephschorr: If we can, I'd like to share the util, even if it means returning two handles

internal/datastore/test/datastore.go
```diff
@@ -27,18 +27,18 @@ type DatastoreTester interface {
 
 // All runs all generic datastore tests on a DatastoreTester.
 func All(t *testing.T, tester DatastoreTester) {
-	t.Run("TestSimple", func(t *testing.T) { SimpleTest(t, tester) })
-	t.Run("TestRevisionFuzzing", func(t *testing.T) { RevisionFuzzingTest(t, tester) })
+	// t.Run("TestSimple", func(t *testing.T) { SimpleTest(t, tester) })
```
josephschorr, line 30: Fix please

internal/datastore/crdb/tx_test.go
samkim: Ok I can update and use the existing one

internal/datastore/crdb/tx.go
samkim: Commit will return an error either for the tx already being close which can be ignored or to rollback. Updated to call rollback on the appropriate error.

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,73 @@ type transactionFn func(tx pgx.Tx) error
...
 	var tx pgx.Tx
-	tx, err = conn.BeginTx(ctx, txOptions)
-	if err != nil {
-		return err
-	}
 	defer func() {
 		if err == nil {
-			_ = tx.Commit(ctx)
-			return
+			if commitErr := tx.Commit(ctx); commitErr == nil {
```
josephschorr, line 82: Log the commit error if not nil

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,73 @@ type transactionFn func(tx pgx.Tx) error
...
+				return
+			}
 		}
 		_ = tx.Rollback(ctx)
```
josephschorr, line 86: Log the rollback error if not nil

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,73 @@ type transactionFn func(tx pgx.Tx) error
...
 	}()
 
-	if _, err = tx.Exec(ctx, "SAVEPOINT cockroach_restart"); err != nil {
+	// n maxRetries can yield n+1 executions of the transaction fn
```
josephschorr, line 89: add `NOTE: `

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,73 @@ type transactionFn func(tx pgx.Tx) error
...
+	for retries = 0; retries <= maxRetries; retries++ {
+		tx, err = resetExecution(ctx, conn, tx, txOptions)
+		if err != nil {
+			log.Ctx(ctx).Err(err).Msg("error resetting transaction")
+			resets++
+			continue
+		}
+
+		var retryAttempts int
+		retryAttempts, err = executeWithRetries(ctx, tx, fn, maxRetries-retries)
```
josephschorr, line 99: Maybe use a newly named variable for err, and then you can use `:=` here?

samkim: Using the named return var name so that `err` can be used below

internal/datastore/test/datastore.go
samkim: doh

internal/datastore/crdb/tx.go
josephschorr: Yeah, I prefer in these cases to use a newly named error var, to remove the need for the `var`

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,80 @@ type transactionFn func(tx pgx.Tx) error
...
+
+		if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
+			log.Ctx(ctx).Err(err).Msg("error during tx rollback")
 		}
-		_ = tx.Rollback(ctx)
 	}()
 
-	if _, err = tx.Exec(ctx, "SAVEPOINT cockroach_restart"); err != nil {
+	// NOTE: n maxRetries can yield n+1 executions of the transaction fn
+	for retries = 0; retries <= maxRetries; retries++ {
```
ecordell, line 97: why n+1? why not do `retries < maxRetries`?

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,80 @@ type transactionFn func(tx pgx.Tx) error
...
+		tx, err = resetExecution(ctx, conn, tx, txOptions)
+		if err != nil {
+			log.Ctx(ctx).Err(err).Msg("error resetting transaction")
```
ecordell, line 100: these log calls that do `log.Ctx` look in the context to fetch a logger (and disables logging if it doesn't find one). I didn't think we were passing the logger around in the context here; are we?

See: https://github.com/rs/zerolog/blob/master/ctx.go#L41-L51

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,80 @@ type transactionFn func(tx pgx.Tx) error
...
-			_ = tx.Commit(ctx)
-			return
+			if commitErr := tx.Commit(ctx); commitErr == nil {
+				return
+			}
+			log.Ctx(ctx).Err(err).Msg("failed tx commit")
+		}
+
+		if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
+			log.Ctx(ctx).Err(err).Msg("error during tx rollback")
```
ecordell, line 92: should we log the `rollbackErr` (and `commitErr` above) too? not sure I've ever seen one so I don't know if they're useful or not

ecordell (approved):

josephschorr (approved):
LGTM

internal/datastore/crdb/tx.go
```diff
@@ -161,7 +161,8 @@ func executeWithRetries(ctx context.Context, currentTx pgx.Tx, fn transactionFn,
 func resetExecution(ctx context.Context, conn conn, tx pgx.Tx, txOptions pgx.TxOptions) (newTx pgx.Tx, err error) {
 	if tx != nil {
 		err = tx.Rollback(ctx)
-		if err != nil {
+		// May return ErrTxClosed if tx is already closed and can be ignored since we're resetting
+		if err != nil && !errors.Is(err, pgx.ErrTxClosed) {
```
ecordell, line 165: when would this already be closed? if we're calling `resetExecution` we shouldn't have called rollback or commit before, right?

samkim: Perhaps here? Initial error is retriable, rollback, rollback returns a resettable error https://github.com/authzed/spicedb/pull/298/files#diff-f86db8a3a74f2b860b9f814eeb6ca84aea022526eaefedf57120a5f1a32ffe80R148

Rollback is happening in the failure that inspired this change: https://github.com/authzed/spicedb/runs/5206162023?check_suite_focus=true#step:4:35313

ecordell: Retriable errors should "release" the savepoint, not actually rollback/close the transaction though. This makes me think there's a bug where we can close the transaction twice

internal/datastore/crdb/tx.go
```diff
@@ -37,36 +50,81 @@ type transactionFn func(tx pgx.Tx) error
...
 	var tx pgx.Tx
-	tx, err = conn.BeginTx(ctx, txOptions)
-	if err != nil {
-		return err
-	}
 	defer func() {
 		if err == nil {
-			_ = tx.Commit(ctx)
-			return
+			commitErr := tx.Commit(ctx)
```
ecordell, line 85: since the `BeginTx` has been moved after this, `tx` can be nil here now (it would be unlikely; every reset would have to fail, but it would be possible)

internal/datastore/crdb/tx.go
samkim: So reading through the integration test error again:
```
{"level":"info","module":"pgx","args":[],"commandTag":"U0FWRVBPSU5U","sql":"SAVEPOINT cockroach_restart","time":0.471605,"time":"2022-02-15T19:45:28Z","message":"Exec"}
{"level":"info","module":"pgx","args":[],"commandTag":"Uk9MTEJBQ0s=","sql":"ROLLBACK TO SAVEPOINT cockroach_restart","time":1.245013,"time":"2022-02-15T19:45:28Z","message":"Exec"}
{"level":"error","error":":  (SQLSTATE 40003)","time":"2022-02-15T19:45:28Z","message":"resettable error, reconnecting with new transaction"}
{"level":"info","module":"pgx","args":[],"commandTag":"Uk9MTEJBQ0s=","sql":"rollback","time":0.197202,"time":"2022-02-15T19:45:28Z","message":"Exec"}
{"level":"info","module":"pgx","time":"2022-02-15T19:45:28Z","message":"closed connection"}
```

I believe it's actually just the retry/reset loop completing so no last reset is actually attempted. I've updated the log statements to be more accurate but that leaves the question of ignoring `ErrTxClosed`. Should we leave it in since we're resetting the tx anyway or remove it to be explicit that we expect transactions to only be closed by this code once?

ecordell: @samkim looking at the error you linked, I think it was just a test flake. I saw the same one a while ago and reported it: https://github.com/authzed/spicedb/issues/209

I think it's fine to revert this line, I think it was fine as it was

samkim: Reverted the ErrTxClosed ignore

ecordell (approved):
LGTM

PR #318, opened 2021-12-03
Transaction Resets

ecordell:
Updated version of #298 for discussion; I haven't tested this yet.


ecordell:
(I accidentially pushed this onto #298 so closing)

PR #320, opened 2021-12-03
crdb: transaction resets

ecordell:
builds on top of #298


ecordell:
Closing in favor of #298 (only opened this when @samkim was out for a bit)

PR #321, opened 2021-12-05
internal/dispatch: extract combined dispatcher

jzelinskie:


internal/dispatch/combined/combined.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	redispatch := graph.NewDispatcher(cachingRedispatch, nsm, ds)
+
+	// If an upstream is specified, create a cluster dispatcher.
+	if opts.upstreamAddr != "" {
+		grpcOpts := []grpc.DialOption{
+			grpc.WithUnaryInterceptor(otelgrpc.UnaryClientInterceptor()),
+			grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy":"consistent-hashring"}`),
+		}
```
jzelinskie, line 79: maybe i could make these configurable via a function option `DefaultGrpcDialOpts()`

ecordell (approved):
LGTM

can't think of a better name than `combined`

PR #323, opened 2021-12-06
Injectable datastore

josephschorr:


internal/datastore/proxy/context.go
```diff
@@ -0,0 +1,119 @@
...
+func (cd ctxDatastore) ListNamespaces(ctx context.Context) ([]*v0.NamespaceDefinition, error) {
+	return DatastoreFromContext(ctx).ListNamespaces(ctx)
+}
+
+// Create a new type to prevent context collisions
+type datastoreKey string
+
+var datastoreCtxKey datastoreKey = "injected-datastore"
+
+type dsHandle struct{ ds datastore.Datastore }
```
jzelinskie, line 89: Why do we need a handle and not just the datastore itself?

internal/datastore/proxy/context.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return datastore
+}
+
+// SetInContext should be called in a gRPC middleware to set the datastore for the context.
+func SetInContext(ctx context.Context, ds datastore.Datastore) {
+	possibleHandle := ctx.Value(datastoreCtxKey)
+	if possibleHandle == nil {
+		return
+	}
```
jzelinskie, line 109: Is this desirable over just replacing it?

internal/datastore/proxy/context.go
josephschorr: Its a pattern we follow everywhere, in case we need to add additional stuff to the context

internal/datastore/proxy/context.go
josephschorr: What do you mean?

internal/datastore/proxy/context.go
jzelinskie: It's really easy to add it in the future when we need it rather than speculatively now. We don't do this in that many places and I consistently advocate against it.

internal/datastore/proxy/context.go
jzelinskie: The behavior of this function is currently "if the datastore has been set, bail early" rather than just overwriting the datastore regardless of whether it's been set or not

jzelinskie: totally misread this, sorry

internal/datastore/proxy/context.go
```diff
@@ -0,0 +1,119 @@
+package proxy
```
jzelinskie, line 1: Do we want this in the proxy package or did we want to make a new middleware package?

internal/datastore/proxy/context.go
josephschorr: Okay

internal/datastore/proxy/context.go
josephschorr: Open question; I just put it there for the draft

ecordell:
https://github.com/authzed/spicedb/pull/404 changed the service layer to fetch the datastore from the context, so this is no longer needed

PR #324, opened 2021-12-06
Log revision skew values

samkim:


internal/datastore/crdb/crdb.go
```diff
@@ -205,6 +205,7 @@ func (cds *crdbDatastore) Revision(ctx context.Context) (datastore.Revision, err
 	if cds.quantizationNanos > 0 {
 		quantized -= (crdbNow % cds.quantizationNanos)
 	}
+	log.Ctx(ctx).Info().Int64("readSkew", cds.followerReadDelayNanos).Int64("totalSkew", nowHLC.IntPart()-quantized).Msg("revision skews")
```
jzelinskie, line 208: Do we want this to be info or debug?

samkim: Thought we'd might want to see it by default but prob too noisy. Updated to debug.

Issue #207, opened 2021-10-22
Proposal: Lookup Watch API and Tiger Cache for fast ACL-aware filtering


josephschorr:
![Tiger Cache](https://user-images.githubusercontent.com/4073002/138512125-5e62f52e-49ab-4ae1-9524-361e5b089585.png)

# Lookup  Watch API and Caching Service ("Tiger cache")

## Background

The [Lookup Resources API](https://buf.build/authzed/api/docs/main/authzed.api.v1#LookupResources) provides a graph-driven permission-aware reverse lookup of all resources, of a particular type, that are accessible within a SpiceDB permissions system for a specified subject.

This API was implemented to solve the [ACL-aware indexing problem](https://authzed.com/blog/acl-filtering-in-authzed/), and does so in a solid manner for the common use case of listing all resources accessible to a subject.

However, the Lookup Resources API has two downsides:

1) it must compute the results from the SpiceDB graph in real time - while caching helps, this can be somewhat unperformant[^1] in systems with millions of relationships and *deeply* nested schemas
2) it provides no ability to perform any additional filtering beyond the type of resource for the target subject - often, applications will want to both filter based on permissions AND other criteria (entered queries, last accessed date/time, etc).

[^1]: Various tests show approximately ~100ms for a LookupResources call on a SpiceDB with 100-250K relationships, and a graph nesting of 3-5 deep.

To solve both of these uses, this issue proposes three new pieces of technology for SpiceDB:

1) a reachability resolution API to be used in concert with the existing [Watch API](https://buf.build/authzed/api/docs/main/authzed.api.v1#authzed.api.v1.WatchRequest)
2) a Lookup Resources Watch API Service
3) a Lookup Resources Caching Service (nicknamed the "Tiger" cache, to follow the naming scheme of the "Leopard" cache for nesting from the Zanzibar paper, and because of its proposed use of a [Roaring Bitmap](https://github.com/RoaringBitmap/roaring))

## Proposal

***NOTE:*** In order to effectively use [Roaring Bitmaps](https://github.com/RoaringBitmap/roaring) in the ways specified below, **ALL** resource IDs in the system **must** be 64-bit integers. We likely need to make it a requirement to use these new services, unless we can come up with an alternative way of storing and transmitting the data, in a compact and efficient form.

### Reachability Resolution APIs

The first new piece to be added to SpiceDB will be the "Reachability Resolution APIs".

[...]

PR #326, opened 2021-12-07
WIP feat: add preliminary LookupWatch API implementation to `v1alpha1`

jon-whit:
The changes here add a preliminary implementation of the LookupWatch API as proposed in issue #207. The LookupWatch API will provide clients with a stream of updates indicating changes in the accessibility of one or more kinds of resources.

### DIFF
--- a/internal/services/v1alpha1/lookupwatch.go
+++ b/internal/services/v1alpha1/lookupwatch.go
@@ -1,0 +1,204 @@
+package v1alpha1
+
+import (
+	"fmt"
+	"io"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	v1alpha1 "github.com/authzed/authzed-go/proto/authzed/api/v1alpha1"
+	"github.com/authzed/grpcutil"
+	"github.com/authzed/spicedb/internal/services/shared"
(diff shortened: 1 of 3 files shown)


internal/services/v1alpha1/lookupwatch.go
```diff
@@ -0,0 +1,204 @@
...
+						// 	}
+						// 	if err != nil {
+						// 		// handle error
+						// 	}
+
+						// 	subjects = append(subjects, resp.GetFoundRelationship().GetSubject())
+						// }
+					}()
+
+					for _, resource := range resources {
```
jon-whit, line 142: We could evaluate these checks concurrently. For example, something like:

```go
var wg sync.WaitGroup
for _, resource := range resources {
    wg.Add(1)
    go func() {
        defer wg.Done()

        for _, subject := range subjects {
            wg.Add(1)
            go func() {
                defer wg.Done()

                Check(resource, subject, ...)
            }()
        }
    }()
}

wg.Wait()
```

My only concern there is just the sheer number of concurrent requests that might be issued. We could throttle it like I did with the `limiter` approach above, but I wanted to get your guys' opinion.

josephschorr:
Stale, so closing due to the work resuming in https://github.com/authzed/spicedb/pull/776

PR #330, opened 2021-12-09
Fix data races and enable race detector in CI

ecordell:
There are three data races fixed in this PR (separate commits):

- lookup dispatch was writing back to the request object (which is shared among many goroutines), the write has been removed
- the lexer's state could be concurrently accessed during parsing, the state has been protected with a mutex and goroutine cleanup has been re-ordered
- The testserver creates and destroys memdbs, but it was possible for dispatch to continue using a memdb datastore after it had been closed. Memdb now has a mutex to protect against that.

With those fixed, the race detector can be enabled in CI to prevent future races.

### DIFF
--- a/internal/services/v0/devcontext.go
+++ b/internal/services/v0/devcontext.go
@@ -118,17 +118,20 @@ func newDevContext(ctx context.Context, requestContext *v0.RequestContext, ds da
 }
 
 func (dc *DevContext) dispose() {
-	datastore := dc.Datastore
-	if datastore != nil {
-		err := dc.NamespaceManager.Close()
-		if err != nil {
-			log.Ctx(dc.Ctx).Err(err).Msg("error when disposing of namespace manager in devcontext")
-		}
+	if err := dc.Dispatcher.Close(); err != nil {
+		log.Ctx(dc.Ctx).Err(err).Msg("error when disposing of dispatcher in devcontext")
+	}
+	if dc.Datastore == nil {
+		return
+	}
+	err := dc.NamespaceManager.Close()
+	if err != nil {
+		log.Ctx(dc.Ctx).Err(err).Msg("error when disposing of namespace manager in devcontext")
+	}
 
-		err = datastore.Close()
-		if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/schemadsl/lexer/lex.go
```diff
@@ -27,16 +28,46 @@ func createLexer(source input.InputSource, input string) *Lexer {
...
+	var state stateFn
+	for {
+		l.withRLock(func() {
+			state = l.state
+		})
+		if state == nil {
+			break
+		}
+		l.withLock(func() {
+			l.state = state(l)
```
josephschorr, line 46: This will lock during the entire lex; you probably want to call the `state(l)` outside of the lock

ecordell: Good call - updated

josephschorr (dismissed):
LGTM

ecordell:
There are a few issues I'm still seeing in `datastore/proxy`:

- `TestDatastoreRequestHedging`
- `TestTupleQueryRequestHedging`
- `TestReverseTupleQueryRequestHedging`
- `TestCommonTupleQueryRequestHedging`

They can be reproduced by enabling the race detector and running the tests for that package with a higher count (i.e. `-count=50`)

internal/services/v0/devcontext.go
```diff
@@ -118,17 +118,20 @@ func newDevContext(ctx context.Context, requestContext *v0.RequestContext, ds da
...
+	err := dc.NamespaceManager.Close()
+	if err != nil {
+		log.Ctx(dc.Ctx).Err(err).Msg("error when disposing of namespace manager in devcontext")
+	}
 
-		err = datastore.Close()
-		if err != nil {
-			log.Ctx(dc.Ctx).Err(err).Msg("error when disposing of datastore in devcontext")
-		}
+	err = dc.Datastore.Close()
```
josephschorr, line 132: Is `dc.Datastore` under a lock?

ecordell: datastore isn't actually used anywhere, just referenced on the devcontext

josephschorr (approved):
LGTM

PR #332, opened 2021-12-09
Versioned namespaces

jakedt:


jakedt:
Note: this make may performance somewhat less predictable, but fixes a real issue with the rollout of new schemas. Now the schema version that was in place as of the chose query revision will be the one that is used to calculate the answer uniformly across all nodes.

ecordell (dismissed):
LGTM

internal/datastore/memdb/memdb.go
```diff
@@ -406,3 +441,10 @@ func createNewTransaction(txn *memdb.Txn) (uint64, error) {
 
 	return newTransactionID, nil
 }
+
+func filterToLiveObjects(revision datastore.Revision) memdb.FilterFunc {
+	return func(hasLifetimeRaw interface{}) bool {
+		obj := hasLifetimeRaw.(hasLifetime)
+		return uint64(revision.IntPart()) < obj.getCreatedTxn() || uint64(revision.IntPart()) >= obj.getDeletedTxn()
```
josephschorr, line 448: Mind adding a comment here that the filter is inverted?

josephschorr (approved):
LGTM

PR #333, opened 2021-12-10
internal/dispatch: return cachingRedispatch

jzelinskie:
This was an autocomplete mistake and should've been the value returned.


ecordell (approved):
LGTM

Issue #329, opened 2021-12-09
SpiceDB server cannot parse comments in the `relationships` section


bryanhuhta:
The Playground supports putting comments into the Test Relationships section. However, the SpiceDB server cannot parse comments in the `relationships` section of a config file.

Using this [playground](https://play.authzed.com/s/Zbshi3GVS9Fc/schema), the exported config looks like:

```yaml
schema: >-
  definition github/user {}


  definition github/project {
      relation reader: github/user
      relation writer: github/user
      relation owner: github/user

      permission read = reader + writer
      permission write = writer + owner
      permission admin = owner
  }
relationships: >-
  // Add users to project

  github/project:pied_piper#owner@github/user:milburga

  github/project:pied_piper#reader@github/user:tarben

  github/project:pied_piper#writer@github/user:freyja
assertions:
  assertTrue: []
  assertFalse: []
validation: null
```

This config can be loaded by the SpiceDB server without an issue, but running `zed schema read` results in a this failure:

```
Error: rpc error: code = Unknown desc = unable to initialize upstream: failed to load config files: Error parsing relationship #0: // Add users to project
```

PR #335, opened 2021-12-10
fix: skip comments when loading test relationships

bryanhuhta:
Fixes https://github.com/authzed/spicedb/issues/329

From a brief test in the playground, comments in the Test Relationships are only of the format `// my comment` and not `/** my comment */`. This may need to be fact check though 😁

### DIFF
--- a/pkg/validationfile/loader.go
+++ b/pkg/validationfile/loader.go
@@ -94,7 +94,7 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*FullyParsed
 			lines := strings.Split(relationships, "\n")
 			for index, line := range lines {
 				trimmed := strings.TrimSpace(line)
-				if len(trimmed) == 0 {
+				if len(trimmed) == 0 || strings.HasPrefix(trimmed, "//") {
 					continue
 				}
 
--- a/pkg/validationfile/loader_test.go
+++ b/pkg/validationfile/loader_test.go
@@ -1,0 +1,50 @@
+package validationfile
+
+import (
+	"testing"
+
+	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
+	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/internal/datastore/memdb"
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+func TestPopulateFromFiles(t *testing.T) {
+	tests := []struct {
+		name      string
+		filePaths []string
+		want      []*v0.RelationTuple
+	}{
+		{
+			name:      "no comment",
+			filePaths: []string{"testdata/loader_no_comment.yaml"},
+			want: []*v0.RelationTuple{
+				tuple.Parse("example/project:pied_piper#owner@example/user:milburga"),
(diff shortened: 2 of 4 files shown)


josephschorr:
> From a brief test in the playground, comments in the Test Relationships are only of the format `// my comment` and not `/** my comment */`. This may need to be fact check though 😁

Yeah, that's correct :)

bryanhuhta:

pkg/validationfile/loader.go
```diff
@@ -94,7 +94,7 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*FullyParsed
 			lines := strings.Split(relationships, "\n")
 			for index, line := range lines {
 				trimmed := strings.TrimSpace(line)
-				if len(trimmed) == 0 {
+				if len(trimmed) == 0 || strings.HasPrefix(trimmed, "//") {
```
josephschorr, line 97: Mind adding a test into `loader_test.go`?

bryanhuhta:
When writing the tests, a few thoughts came to me:

- Is there any recommendation around using `testdata/` to store test data?
- I used `memdb.NewMemdbDatastore` to build a `datastore.Datastore` for my test. I'm not sure if there's an existing `datastore.Datastore` implementation specifically for being mocked in tests.

josephschorr:
> When writing the tests, a few thoughts came to me:
> 
> * Is there any recommendation around using `testdata/` to store test data?

We actually do something similar for the consistency tests: https://github.com/authzed/spicedb/blob/main/internal/services/consistency_test.go (it loads from a directory called `testconfigs`), so follow the style in there. If the test data is small enough, you can also inline.

> * I used `memdb.NewMemdbDatastore` to build a `datastore.Datastore` for my test. I'm not sure if there's an existing `datastore.Datastore` implementation specifically for being mocked in tests.

MemDB is the correct datastore to use. See https://github.com/authzed/spicedb/blob/main/internal/services/consistency_test.go#L64 for an example

pkg/validationfile/loader_test.go
```diff
@@ -0,0 +1,138 @@
...
+	tests := []struct {
+		name      string
+		filePaths []string
+		want      []*v0.RelationTuple
+	}{
+		{
+			name:      "no comment",
+			filePaths: []string{"testdata/loader_no_comment.yaml"},
+			want: []*v0.RelationTuple{
+				{
```
josephschorr, line 21: You can use `tuple.Parse` to make this more compact (not required, but makes it a bit more readable)

pkg/validationfile/testdata/loader_with_comment.yaml
```diff
@@ -0,0 +1,25 @@
...
+      permission write = writer + admin
+      permission admin = owner
+  }
+relationships: >-
+  // Some comment
+
+  example/project:pied_piper#owner@example/user:milburga
+
+  example/project:pied_piper#reader@example/user:tarben
+
```
josephschorr, line 20: Please add another comment here as well

josephschorr (dismissed):
LGTM

josephschorr (approved):
LGTM

PR #336, opened 2021-12-10
Implement support for the public proposal

josephschorr:
See: #1


internal/services/v0/validation.go
```diff
@@ -32,6 +44,17 @@ func validateTupleWrite(
 		return err
 	}
 
+	// Ensure wildcard writes have no subject relation.
+	if tpl.User.GetUserset().ObjectId == tuple.PublicWildcard {
+		if tpl.User.GetUserset().Relation != tuple.Ellipsis {
+			return invalidRelationError{
+				error:   fmt.Errorf("wildcard (public) relationships require a subject relation of `...` on %v", tpl.ObjectAndRelation),
```
jzelinskie, line 51: Just call it wildcard everywhere instead of "wildcard (public)".

jzelinskie (changes requested):
first pass is just nitpicks tbh
I need to stare at the type system changes a bit more, but beyond that the nsbuilder/validation/services all look good.

internal/datastore/common/validation.go
```diff
@@ -0,0 +1,28 @@
...
+func ValidateUpdatesToWrite(updates []*v1.RelationshipUpdate) error {
+	for _, update := range updates {
+		err := update.HandwrittenValidate()
+		if err != nil {
+			return err
+		}
+
+		if update.Relationship.Subject.Object.ObjectId == tuple.PublicWildcard && update.Relationship.Subject.OptionalRelation != "" {
+			return fmt.Errorf(
+				"Attempt to write a wildcard relationship (`%s`) with a non-empty relation. Please report this bug",
```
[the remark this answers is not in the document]
josephschorr: We could put it into a proxy, but if that doesn't get installed (for whatever reason), then bad data could get into the system. I wanted it to be done via each handler to ensure its done right at the point before we write the relationships. I also did not have it panic because, well, I don't want it to crash the entire binary

internal/graph/check.go
```diff
@@ -65,9 +66,15 @@ func (cc *ConcurrentChecker) dispatch(req ValidatedCheckRequest) ReduceableCheck
 	}
 }
 
+func onrEqualOrPublic(tpl, target *v0.ObjectAndRelation) bool {
```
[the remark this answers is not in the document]
josephschorr: Yeah, I can rename it. I just copied Evan's changes over wholesale :)

internal/datastore/common/validation.go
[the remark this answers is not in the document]
jzelinskie: It's more likely that someone forgets they need to call this in their implementation of a data store than we miss adding it to the gRPC server. We already have absolutely required gRPC middleware -- either we trust it or we shouldn't bother having it.

internal/services/v1/relationships.go
```diff
@@ -169,6 +182,16 @@ func (ps *permissionServer) WriteRelationships(ctx context.Context, req *v1.Writ
...
 
+		// Ensure wildcard writes have no subject relation.
+		if update.Relationship.Subject.Object.ObjectId == tuple.PublicWildcard {
+			if update.Relationship.Subject.OptionalRelation != "" {
+				return nil, status.Errorf(
+					codes.InvalidArgument,
+					"wildcard (public) relationships require an empty subject relation",
+				)
+			}
+		}
```
jzelinskie, line 193: Is this something we should have in the `HandwrittenValidate()` in authzed-go?

josephschorr: We could, although it does raise a question of when we should be items in there vs here.

internal/datastore/common/validation.go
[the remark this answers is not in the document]
josephschorr: True, but I always tend to err on the side of extra checks where possible. Ideally, this would be a debug and testing-only check, but Go makes that less than fun to do

internal/services/v1/relationships.go
jzelinskie: This is purely an invariant that is self contained within the contents of the request, thus I think it should be in a validation step, rather than in a handler where our logic lives.

josephschorr: Okay. This will require another change to authzed-go, though

internal/datastore/common/validation.go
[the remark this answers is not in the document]
jakedt: We already have a validating datastore proxy for debug and testing checks. Put it there?

internal/services/v1/relationships.go
jzelinskie: Yep, I think that's the right way to do it.

jzelinskie: Handled in https://github.com/authzed/authzed-go/pull/47

jzelinskie (approved):

PR #340, opened 2021-12-14
Backport some datastore changes from datastore-v2

### LINKED ISSUE #192: Rename Datastore Revision() method
A new name should more accurately represent the intention of the method.

CRDB datastores uses a quantized time as the revision, while Postgres/Memory datastores use random fuzzing to determine their revision

jakedt:
Fixes #192


internal/datastore/proxy/mapping.go
```diff
@@ -179,18 +180,52 @@ func (mp mappingProxy) DeleteNamespace(ctx context.Context, nsName string) (data
...
+		}
+
+		subFilter = &v1.SubjectFilter{
+			SubjectType:       subResourceType,
+			OptionalSubjectId: filter.OptionalSubjectFilter.OptionalSubjectId,
+			OptionalRelation:  filter.OptionalSubjectFilter.OptionalRelation,
+		}
+	}
+
+	queryOpts := options.NewQueryOptionsWithOptions(opts...)
```
josephschorr, line 208: Can we add a check in here that the *only* two options used are Limit and/or Usersets? That way, if we add another new option down the road, we don't forget to map it

internal/datastore/test/tuples.go
```diff
@@ -89,64 +90,75 @@ func SimpleTest(t *testing.T, tester DatastoreTester) {
...
-					iter, err := query.Execute(ctx)
-					require.NoError(err)
-					tRequire.VerifyIteratorResults(iter, tupleToFind)
-				}
+				iter, err := ds.QueryTuples(ctx, &v1.RelationshipFilter{
+					ResourceType:       tupleToFind.ObjectAndRelation.Namespace,
+					OptionalResourceId: tupleToFind.ObjectAndRelation.ObjectId,
+				}, lastRevision)
+				require.NoError(err)
+				tRequire.VerifyIteratorResults(iter, tupleToFind)
```
josephschorr, line 96: Can we add some negative tests here, where the tuple is not found due to the extra options given?

internal/graph/lookup.go
```diff
@@ -260,10 +262,17 @@ func (cl *ConcurrentLookup) lookupDirect(ctx context.Context, req ValidatedLooku
...
+				limit := uint64(req.Limit)
+				it, err := cl.ds.QueryTuples(
+					ctx,
+					&v1_proto.RelationshipFilter{
+						ResourceType:     req.ObjectRelation.Namespace,
+						OptionalRelation: req.ObjectRelation.Relation,
+					},
+					req.Revision,
+					options.SetUsersets(result.Resp.ResolvedOnrs),
+					options.WithLimit(&limit),
```
josephschorr, line 280: `WithLimit` taking in a pointer makes sense to define no limit, but the requirement to `&theIntValue` isn't great... could we have a `WithLimit(uint64)` and `WithNoLimit`, or would that break the optgen?

internal/datastore/options/options.go
```diff
@@ -4,14 +4,26 @@ import (
...
 
 // QueryOptions are the options that can affect the results of a normal forward query.
 type QueryOptions struct {
 	Limit    *uint64
 	Usersets []*v0.ObjectAndRelation
 }
 
+// ReverseQueryOptions are the options that can affect the results of a reverse query.
+type ReverseQueryOptions struct {
+	ReverseLimit *uint64
```
josephschorr, line 17: Why `ReverseLimit`? Just to make sure the options don't collide?

internal/datastore/proxy/mapping.go
```diff
@@ -228,23 +228,46 @@ func (mp mappingProxy) QueryTuples(
...
-func (mp mappingProxy) ReverseQueryTuplesFromSubjectRelation(subjectNamespace, subjectRelation string, revision datastore.Revision) datastore.ReverseTupleQuery {
-	translatedNamespace, err := mp.mapper.Encode(subjectNamespace)
-	return mappingReverseTupleQuery{
-		mp.delegate.ReverseQueryTuplesFromSubjectRelation(translatedNamespace, subjectRelation, revision),
-		mp.mapper,
-		err,
+	translatedOptions := []options.ReverseQueryOptionsOption{
+		options.WithReverseLimit(queryOpts.ReverseLimit),
 	}
+	if queryOpts.ResRelation != nil {
```
josephschorr, line 247: Same check here to ensure we aren't use any other options than those handled already

internal/services/shared/schema.go
```diff
@@ -38,7 +38,9 @@ func EnsureNoRelationshipsExist(ctx context.Context, ds datastore.Datastore, nam
 		return err
 	}
 
-	qy, qyErr = ds.ReverseQueryTuplesFromSubjectNamespace(namespaceName, syncRevision).Limit(1).Execute(ctx)
+	qy, qyErr = ds.ReverseQueryTuples(ctx, &v1.SubjectFilter{
```
josephschorr, line 41: Do we want to rename `syncRevision` to `headRevision` to match the other commit?

internal/datastore/options/options.go
jakedt: Yeah, optgen would have generated two `WithLimit` methods if they were both called `Limit`.

internal/graph/lookup.go
jakedt: optgen is not magic, it sees a type and it builds a `WithX` function using the same type. @ecordell correct me if I'm wrong.

internal/datastore/proxy/mapping.go
jakedt: ideas on how we would do that?

josephschorr: Have optgen generate some sort of count of the number of options possible and then panic here if it mismatches? Alternatively, optgen could generate a unique "type" or constant name encompassing the options available, and then we reference it here, ensuring that if a new options is added, that reference no longer type checks?

internal/datastore/datastore.go
```diff
@@ -70,65 +73,26 @@ type Datastore interface {
...
-	ReverseQueryTuplesFromSubject(subject *v0.ObjectAndRelation, revision Revision) ReverseTupleQuery
-
-	// ReverseQueryTuplesFromSubjectRelation creates a builder for reading tuples
-	// from a subject relation onward from the datastore.
-	ReverseQueryTuplesFromSubjectRelation(subjectNamespace, subjectRelation string, revision Revision) ReverseTupleQuery
-
-	// ReverseQueryTuplesFromSubjectNamespace creates a builder for reading
-	// tuples from a subject namespace onward from the datastore.
-	ReverseQueryTuplesFromSubjectNamespace(subjectNamespace string, revision Revision) ReverseTupleQuery
+	QueryTuples(
```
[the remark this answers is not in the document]
jakedt: we could also theoretically combine them by just making a version of relationshipfilter that allowed the resource type float freely...

josephschorr (approved):
LGTM

PR #349, opened 2021-12-22
e2e: fill schema with many namespaces to span ranges

ecordell:
This PR restores the e2e test suite after #332 broke them. Now that namespaces are read at specific revisions on tuple read/write, it's more difficult to reason about which cockroach nodes will be used for a given operation. This PR accounts for the additional namespace reads in the test.

What the test used to do:
- write a schema
- fill with tuples so that tuples span across ranges
- slow time on a node
- create tuples until we see the second write hit the slow node, which causes the check to return an incorrect value
- (enable the mitigations and confirm it works)

What the test does now:
- writes many schemas so that schemas span ranges
- find a schema that has a leader on the slow node. this ensures that when the namespaces are read when writing to the slow node that the slow node doesn't need to communicate to other nodes (which would sync their timestamps)
- fill just the schema we identified with tuples so that the tuples for that schema span multiple ranges
- the rest of the test is mostly the same, though the network delay was removed in favor of a sleep after the first write, which seems to be more reliable


e2e/spice/spicedb.go
```diff
@@ -76,6 +76,7 @@ func (s *Node) Start(ctx context.Context, logprefix string, args ...string) erro
 		"./spicedb",
 		"serve",
 		"--log-level=debug",
+		"--datastore-request-hedging=false",
```
jzelinskie, line 79: does this need an option func?

internal/namespace/manager.go
```diff
@@ -29,7 +29,7 @@ type Manager interface {
 	// ReadNamespaceAndTypes reads a namespace definition, version, and type system and returns it if found.
 	ReadNamespaceAndTypes(ctx context.Context, nsName string, revision decimal.Decimal) (*v0.NamespaceDefinition, *NamespaceTypeSystem, error)
 
-	// Closes the namespace manager, disposing of any resources.
+	// Close closes the namespace manager, disposing of any resources.
```
jzelinskie, line 32: Are we going to enforce linting rules on the e2e code? Because it's a separate project almost all the tooling doesn't run on this code.

e2e/newenemy/newenemy_test.go
```diff
@@ -270,37 +317,78 @@ func fill(require *require.Assertions, client v0.ACLServiceClient, fillerCount,
...
+	return nil
+}
+
+// prefixForNode finds a prefix with a leader the node id
+func prefixForNode(ctx context.Context, conn *pgx.Conn, data []SchemaData, node int) (prefix string) {
+	for prefix == "" {
+		// get a random prefix
+		d := data[rand.Intn(len(data))]
+		candidate := d.Prefixes[rand.Intn(len(d.Prefixes))]
+		fmt.Println("checking prefix", candidate)
```
jzelinskie, line 346: Should we be using a logger instead of everywhere in this tool?

e2e/spice/spicedb.go
ecordell: I've only been adding option funcs for things that need to vary between test runs; so I would add one if we start using this to run other e2e tests where we want request hedging

internal/namespace/manager.go
ecordell: this is actually in the main module, not e2e

but yeah we should probably run the linters over this too

e2e/newenemy/newenemy_test.go
ecordell: right now there's a mix of stdout and t.log, so this isn't making things any worse, but yes I think it should be more consistent

e2e/spice/spicedb.go
ecordell: Added one for this

e2e/newenemy/newenemy_test.go
ecordell: everything goes through t.Log now

internal/namespace/manager.go
jzelinskie: >this is actually in the main module, not e2e

Yep, the fact that this is a lint fix just reminded me that the other files are not being checked.

jzelinskie (approved):

PR #352, opened 2021-12-25
Fix linter for many packages

jzelinskie:
- The only packages that are ignored by revive stutter in a way that I'm not really comfortable changing yet, because I feel like it'd cascade into renaming almost all the types/functions in those packages.
- I tried to use every linter in golanglint-ci that was reasonably easy adopt
- I vendored gofumpt/stringer similar to how we vendor optgen in `tools.go`
- I implemented a `pkg/testutil` that includes an equality function that treats `[]T(nil)` and `[]T{}` as equal
- The names of some of the metrics did not follow Prometheus idioms and so they were renamed (`promlint` will enforce this in the future)

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,17 +15,12 @@ jobs:
       - uses: "actions/setup-go@v2"
         with:
           go-version: "^1.17"
-      - name: "Install linting tools"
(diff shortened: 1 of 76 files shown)


josephschorr (approved):
LGTM

PR #355, opened 2021-12-28
pkg/testutil: ensure types in RequireEqualEmptyNil

jzelinskie:
Previously we were not checking that the values were the same type, only types that had lengths.
This now ensures expected and actual are the same types.


josephschorr (approved):
LGTM

PR #376, opened 2022-01-12
Add a config object for spicedb servers, control graceful stop of all services

ecordell:
this simplifies using spicedb as a library while retaining the default
validation logic provided by the cli

This also improves spicedb startup/shutdown to properly clean up all servers.


pkg/cmd/server.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/authzed/spicedb/internal/auth"
+	"github.com/authzed/spicedb/internal/middleware/servicespecific"
+	logmw "github.com/authzed/spicedb/pkg/middleware/logging"
+	"github.com/authzed/spicedb/pkg/middleware/requestid"
+)
+
+type ServerOption func(*ServerConfig)
+
+//go:generate go run github.com/ecordell/optgen -output zz_generated.server_options.go . ServerConfig
```
jzelinskie, line 35: We can follow up afterwards to make some of these fields unexported.

pkg/cmd/serve/serve.go
```diff
@@ -3,80 +3,75 @@ package serve
...
 	cmdutil "github.com/authzed/spicedb/pkg/cmd"
-	logmw "github.com/authzed/spicedb/pkg/middleware/logging"
-	"github.com/authzed/spicedb/pkg/middleware/requestid"
 	"github.com/authzed/spicedb/pkg/validationfile"
 )
 
-func RegisterServeFlags(cmd *cobra.Command, dsConfig *cmdutil.DatastoreConfig) {
+const PresharedKeyFlag = "grpc-preshared-key"
+
+func RegisterServeFlags(cmd *cobra.Command, config *cmdutil.ServerConfig) {
```
jzelinskie, line 14: We discussed a couple things with regards to registering flags:
- cobrautil could have the ability to bind to variables and spicedb would lightly wrap that
- we could have a function that generates []Option by iterating over the flagset

pkg/cmd/serve/serve.go
```diff
@@ -161,40 +148,25 @@ func serveRun(ctx context.Context, cmd *cobra.Command, args []string, datastoreO
...
-	middleware := grpc.ChainUnaryInterceptor(
-		requestid.UnaryServerInterceptor(requestid.GenerateIfMissing(true)),
-		logmw.UnaryServerInterceptor(logmw.ExtractMetadataField("x-request-id", "requestID")),
-		grpclog.UnaryServerInterceptor(grpczerolog.InterceptorLogger(log.Logger)),
-		otelgrpc.UnaryServerInterceptor(),
-		grpcauth.UnaryServerInterceptor(auth.RequirePresharedKey(token)),
-		grpcprom.UnaryServerInterceptor,
-		servicespecific.UnaryServerInterceptor,
-	)
+	if len(config.UnaryMiddleware) == 0 || len(config.StreamingMiddleware) == 0 {
```
jzelinskie: What if you really don't want any middleware?

pkg/cmd/serve/serve.go
```diff
@@ -161,40 +148,25 @@ func serveRun(ctx context.Context, cmd *cobra.Command, args []string, datastoreO
...
-	streamMiddleware := grpc.ChainStreamInterceptor(
-		requestid.StreamServerInterceptor(requestid.GenerateIfMissing(true)),
-		logmw.StreamServerInterceptor(logmw.ExtractMetadataField("x-request-id", "requestID")),
-		grpclog.StreamServerInterceptor(grpczerolog.InterceptorLogger(log.Logger)),
-		otelgrpc.StreamServerInterceptor(),
-		grpcauth.StreamServerInterceptor(auth.RequirePresharedKey(token)),
-		grpcprom.StreamServerInterceptor,
-		servicespecific.StreamServerInterceptor,
-	)
+	// TODO: should we validate that there is a psk-check middleware if a psk is set?
```
jzelinskie: This sounds like a good thing to have before shipping this.

pkg/cmd/serve/serve.go
ecordell: You can wipe them out if you want to:

```go
server, _ := config.Complete()
server.SetMiddleware(nil, nil)
```

pkg/cmd/serve/serve.go
ecordell: I wasn't totally sure this was a good idea, similar to your other comment ("What if you really don't want any middleware?") 

For example, if we want to provide other authn options in the future, the psk might be superseded by some other auth middleware.

pkg/cmd/serve/serve.go
ecordell: Will update

pkg/cmd/serve/serve.go
jzelinskie: I think at that point in time we'd just have to change the checking code.

`WithAuthentication(presharedkeyMiddleware(key))` might be a way forward and we can change the strategies passed to WithAuthentication in the future.

jzelinskie (approved):

PR #379, opened 2022-01-13
Set bounds on the parsed line number int in developer API

josephschorr:
This ensures there is no overflow from the YAML line number when reported in the developer API

Fixes https://github.com/authzed/spicedb/security/code-scanning/1


jzelinskie:
This fix was included in #378.

PR #380, opened 2022-01-13
Attempt to avoid failed crdb range splits in e2e

ecordell:
https://github.com/cockroachdb/cockroach/issues/63260 suggests that this setting will avoid the issue


josephschorr (approved):
LGTM

PR #381, opened 2022-01-13
Better usage metrics on non-permissions endpoints

jakedt:
Originally I was going to write empty usage metrics on all error conditions, which is what prompted the `errgroup` work, to try to reduce the number of top level return calls. I ended up putting it in the middleware to make sure it still writes the trailers if other middlewares fail the request pre-emptively. We can either keep or drop the `errgroup` changes.

### DIFF
--- a/internal/services/v0/acl.go
+++ b/internal/services/v0/acl.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
 	v1_api "github.com/authzed/authzed-go/proto/authzed/api/v1"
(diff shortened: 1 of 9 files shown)


internal/services/v0/acl.go
```diff
@@ -94,6 +95,10 @@ func (as *aclServer) Write(ctx context.Context, req *v0.WriteRequest) (*v0.Write
 		mutations = append(mutations, tuple.UpdateToRelationshipUpdate(mut))
 	}
 
+	usagemetrics.SetInContext(ctx, &v1.ResponseMeta{
+		DispatchCount: uint32(len(preconditions)) + 1,
```
josephschorr, line 99: Mind adding a comment on why its `+ 1`?

internal/services/v1/relationships.go
```diff
@@ -146,93 +152,108 @@ func (ps *permissionServer) WriteRelationships(ctx context.Context, req *v1.Writ
 		return nil, rewritePermissionsError(ctx, err)
 	}
 
+	errG, groupCtx := errgroup.WithContext(ctx)
 	for _, precond := range req.OptionalPreconditions {
-		if err := ps.checkFilterNamespaces(ctx, precond.Filter, readRevision); err != nil {
-			return nil, rewritePermissionsError(ctx, err)
-		}
+		precond := precond
```
josephschorr, line 157: Please add a comment on why this is necessary to prevent from changing

internal/services/v1/relationships.go
```diff
@@ -253,6 +274,10 @@ func (ps *permissionServer) DeleteRelationships(ctx context.Context, req *v1.Del
 		return nil, rewritePermissionsError(ctx, err)
 	}
 
+	usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
+		DispatchCount: uint32(len(req.OptionalPreconditions)) + 1,
```
josephschorr, line 278: Ditto comment here on `+ 1`

internal/services/v1/schema.go
```diff
@@ -60,6 +62,10 @@ func (ss *schemaServer) ReadSchema(ctx context.Context, in *v1.ReadSchemaRequest
 		objectDefs = append(objectDefs, objectDef)
 	}
 
+	usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
+		DispatchCount: uint32(len(nsDefs)),
```
josephschorr, line 66: I wonder if this should be `+ 1`, since if their schema is empty, we are still doing *something*

internal/services/v1/schema.go
```diff
@@ -140,6 +146,10 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
 		names = append(names, nsdef.Name)
 	}
 
+	usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
+		DispatchCount: uint32(len(nsdefs)),
```
josephschorr, line 150: Ditto

internal/services/v1alpha1/schema.go
```diff
@@ -78,6 +82,10 @@ func (ss *schemaServiceServer) ReadSchema(ctx context.Context, in *v1alpha1.Read
 		objectDefs = append(objectDefs, objectDef)
 	}
 
+	usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
+		DispatchCount: uint32(numRequested),
```
josephschorr, line 86: Ditto

internal/services/v1/schema.go
jakedt: I didn't think so, it's easier to document that it's the number of schemas rather than `max(numSchemas, 1)`

jakedt: Like, if we fail the request, that's also counting as zero since we didn't really do any work on their behalf...

jakedt:
Updated

internal/services/v1/schema.go
josephschorr: Well, we are still making the request itself, but yeah

josephschorr (approved):
LGTM

PR #391, opened 2022-01-27
add universal consistency middleware

ecordell:
Depends on #392


internal/services/v1alpha1/schema_test.go
```diff
@@ -28,8 +28,27 @@ func TestSchemaReadNoPrefix(t *testing.T) {
 	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC, 0)
 	require.NoError(t, err)
 
-	srv := NewSchemaServer(ds, PrefixRequired)
-	_, err = srv.ReadSchema(context.Background(), &v1alpha1.ReadSchemaRequest{
+	// test relies on middleware
```
jakedt, line 31: There are far too many copies of code that looks just like this in this PR. Why?

pkg/testutil/require.go
```diff
@@ -3,30 +3,22 @@
...
+	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/require"
 )
 
 // RequireEqualEmptyNil is a version of require.Equal, but considers nil
 // slices/maps to be equal to empty slices/maps.
 func RequireEqualEmptyNil(t *testing.T, expected, actual interface{}, msgAndArgs ...interface{}) {
-	expectedVal := reflect.ValueOf(expected)
```
jakedt, line 15: These changes should probably also be in a separate commit.

internal/services/v0/acl.go
```diff
@@ -258,20 +239,13 @@ func (as *aclServer) Read(ctx context.Context, req *v0.ReadRequest) (*v0.ReadRes
 }
 
 func (as *aclServer) Check(ctx context.Context, req *v0.CheckRequest) (*v0.CheckResponse, error) {
-	atRevision, err := as.pickBestRevision(ctx, req.AtRevision)
```
jakedt, line 261: It's a little unclear why `Check` and `ContentChangeCheck` will behave differently at first glance. I wonder if we could improve the ergonomics at the call site of `MustRevisionFromContext`.

jakedt (changes requested):

internal/services/v0/acl.go
ecordell: Is it worth doing much more than a comment here, since it's the v0 api? (I think the consistency block logic is much clearer in v1).

internal/services/v1alpha1/schema_test.go
ecordell: I was planning to use to buffconn PR to clean these up later, but I'll do it now so it's more reviewable.

internal/services/v0/acl.go
ecordell: (added a comment to ContentChangeCheck)

pkg/testutil/require.go
ecordell: I didn't track down exactly why this started happening, but after adding the consistency middleware some of the tests that assert with `RequireEqualEmptyNil` had some deeply nested types be unequal (because one side was nil and the other was an empty slice). I thought it was easier to make the test helper (which already collapsed those cases for top-level values) work for nested types too (and also print out more helpful error messages when they fail) instead of touching every test expectation.

TL;DR this is pretty coupled to the other changes in this commit

jakedt: That's fine, but can't you just put it in a commit before you introduce the universal consistency? universal consistency has this as a pre-requisite IIUC.

cmd/spicedb/servetesting_integration_test.go
```diff
@@ -159,13 +160,14 @@ func newTester(t *testing.T, containerOpts *dockertest.RunOptions) (*spicedbHand
 		}
 	}
 
+	conn, err := grpc.Dial(fmt.Sprintf("localhost:%s", port),
```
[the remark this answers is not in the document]
jakedt: Did you move this to a separate PR or something?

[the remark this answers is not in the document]
ecordell: I just removed it since it was unrelated. I'll have to touch this file when I update the test server anyway (so the linter will complain that I'm using a deprecated api and I'll have to change it there)

internal/services/v0/namespace_test.go
```diff
@@ -111,29 +90,30 @@ func TestNamespaceChanged(t *testing.T) {
...
-			ns.Namespace(
-				"folder",
-				ns.Relation("anotherrel",
-					nil,
-					ns.AllowedRelation("folder", "..."),
-				),
-			),
-			"cannot delete relation `viewer` in definition `folder`, as a relationship references it",
-		},
+		// TODO: review this test
```
jakedt, line 93: Review it for what? Was it reviewed? Is it just failing?

jakedt (changes requested):

internal/services/v0/namespace_test.go
ecordell: Fixed - the tuple being written wasn't valid under the namespace config; the previous test wasn't type-checking

PR #392, opened 2022-01-27
support buffconn for grpc server config

ecordell:
this makes it much easier to use the spicedb server config in tests


pkg/cmd/server/server.go
```diff
@@ -271,3 +284,10 @@ func (c *completedServerConfig) Run(ctx context.Context) error {
 
 	return nil
 }
+
+type GrpcMetadataCredentials map[string]string
```
jzelinskie, line 288: Is this being used?

pkg/cmd/util/util.go
```diff
@@ -78,29 +96,90 @@ func (c *GRPCServerConfig) Complete(level zerolog.Level, svcRegistrationFn func(
...
 	}, nil
 }
 
+func (c *GRPCServerConfig) listenerAndDialer() (l net.Listener, dial DialFunc, err error) {
+	if c.Network == "bufconn" {
+		bl := bufconn.Listen(c.BufferSize)
+		dial = func(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
+			opts = append([]grpc.DialOption{grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
+				return bl.DialContext(ctx)
+			})}, opts...)
```
jzelinskie, line 115: does this need to be in this order? can you instead do

```go
opts = append(opts, grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
	return bl.DialContext(ctx)
})
```

pkg/cmd/util/util.go
```diff
@@ -78,29 +96,90 @@ func (c *GRPCServerConfig) Complete(level zerolog.Level, svcRegistrationFn func(
...
+			return grpc.DialContext(ctx, "bufnet", opts...)
+		}
+		l = bl
+		return
```
jzelinskie, line 119: I was a little confused that this mixed and match returning named params with explicit returns

pkg/cmd/util/util.go
```diff
@@ -78,29 +96,90 @@ func (c *GRPCServerConfig) Complete(level zerolog.Level, svcRegistrationFn func(
...
 				Str("prefix", c.flagPrefix).Msg("grpc server stopped listening")
 		},
 		stopFunc: srv.GracefulStop,
-		enabled:  c.Enabled,
+		creds:    clientCreds,
 	}, nil
 }
 
+func (c *GRPCServerConfig) listenerAndDialer() (l net.Listener, dial DialFunc, err error) {
+	if c.Network == "bufconn" {
```
jzelinskie, line 110: should we call this bufnet?

pkg/cmd/util/util.go
```diff
@@ -78,29 +96,90 @@ func (c *GRPCServerConfig) Complete(level zerolog.Level, svcRegistrationFn func(
...
+	case c.TLSCertPath == "" && c.TLSKeyPath == "":
+		log.Warn().Str("prefix", c.flagPrefix).Msg("grpc server serving plaintext")
+	case c.TLSCertPath != "" && c.TLSKeyPath != "":
+		creds, err := credentials.NewServerTLSFromFile(c.TLSCertPath, c.TLSKeyPath)
+		if err != nil {
+			return nil, err
+		}
+		return []grpc.ServerOption{grpc.Creds(creds)}, nil
+	}
+	return nil, nil
```
jzelinskie, line 142: maybe pull this into the default case

pkg/cmd/util/util.go
```diff
@@ -78,29 +96,90 @@ func (c *GRPCServerConfig) Complete(level zerolog.Level, svcRegistrationFn func(
...
+		} else {
+			pool, err = x509.SystemCertPool()
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		return credentials.NewTLS(&tls.Config{RootCAs: pool, MinVersion: tls.VersionTLS12}), nil
+	}
+	return nil, nil
```
jzelinskie, line 163: default again

jzelinskie (changes requested):

pkg/cmd/server/server.go
ecordell: good catch, I missed that when I was cleaning up

pkg/cmd/util/util.go
ecordell: The functions that are defined inside the `listenerAndDialer` use explicit returns, the method `listenerAndDialer` itself uses named params. Do you think they should match?

pkg/cmd/util/util.go
ecordell: sure - added a constant as well

pkg/cmd/util/util.go
jzelinskie: I think everything should be explicit returns because it's harder to read. I can treat the `if c.Network == BufferedNetwork` conditional as standalone scope-wise if not for the named return arguments.

ecordell: updated

jzelinskie (approved):

PR #393, opened 2022-01-30
Retrieve datastore from context

ecordell:
Includes #391, #392

This switches services to use the datastore middleware to get the datastore, rather than passing it in during construction. It also provides a new namespace manager that will fetch the datastore from context. The plan is to use this to simplify the testserver implementation.

Includes #323; the namespace manager uses this to delegate to the datastore in the context


ecordell:
I decided it made more sense to include these in #404 - it makes the other PR a little larger, but it's easier to see how this changes will be used.

PR #406, opened 2022-02-04
Move the bulk of the dev API impl into its own package

josephschorr:
This will allow it to be called by other tooling, in addition to the developer API


pkg/development/devcontext.go
```diff
@@ -0,0 +1,303 @@
...
+			Line:    line,
+			Column:  column,
+			Context: context,
+		}, nil
+	}
+
+	return nil, rewriteACLError(ctx, dispatchError)
+}
+
+func rewriteACLError(ctx context.Context, err error) error {
```
ecordell, line 267: this is copied from `internal/services/v0`, right?

should this just be public and shared?

josephschorr: Yes, and I don't think so. Once v0 is gone, it'll only be needed in here, so I decided to just copy it

ecordell (approved):

PR #408, opened 2022-02-07
Add more context to schema parse errors

josephschorr:


internal/services/v0/developer_test.go
```diff
@@ -98,11 +98,12 @@ func TestEditCheck(t *testing.T) {
 			[]*v0.RelationTuple{},
 			[]*v0.RelationTuple{},
 			&v0.DeveloperError{
-				Message: "parse error in `schema`, line 3, column 4: Expected identifier, found token TokenTypeRightBrace",
+				Message: "Expected identifier, found token TokenTypeRightBrace",
```
ecordell, line 101: would this message be more readable with `}` instead of `TokenTypeRightBrace`?

pkg/schemadsl/compiler/compiler.go
```diff
@@ -94,9 +106,11 @@ func toContextError(errMessage string, node *dslNode, mapper input.PositionMappe
 	}
 
 	return ErrorWithContext{
-		error:       fmt.Errorf("parse error in %s: %s", formattedRange, errMessage),
-		SourceRange: sourceRange,
-		Source:      input.Source(source),
+		error:            fmt.Errorf("parse error in %s: %s", formattedRange, errMessage),
```
ecordell, line 109: Isn't the thing you're doing here with `error`/`BaseErrorMessage` basically what you get with golang's error wrapping?

internal/services/v0/developer_test.go
josephschorr: That's actually what I'm passing along now out of band from the error message itself. We could change the error message as well, but it would require a reverse map of the token types (which, I suppose would be fairly easy to do), but may appear odd for non-ASCII runes

pkg/schemadsl/compiler/compiler.go
josephschorr: Yes, but I wanted it to be explicit from the compiler's error type

ecordell: But whenever you wanted to get the underlying error, you can just `errors.As(err, &BaseError{})` to get it back out. I'm not clear on the advantage to doing it outside the stdlib?

internal/services/v0/developer_test.go
ecordell: This could be a follow up, but it would it make sense to not keep a reverse map, but instead keep an "error message map"? then we can write the best representation for any given token.

josephschorr: Yeah, we could do that too

pkg/schemadsl/compiler/compiler.go
josephschorr: Well, we'd have to declare another error type to do that. Do you prefer that approach?

josephschorr: Done

internal/services/v0/developer_test.go
josephschorr: Let's do that as a followup

Issue #153, opened 2021-09-30
Implement a MySQL or Vitess datastore


jzelinskie:
Creating this issue to gauge interest

derekperkins:
Very interested, specifically to run on Vitess

josephschorr:
Fixed by https://github.com/authzed/spicedb/pull/525, which introduces a beta MySQL store

derekperkins:
🎉

PR #411, opened 2022-02-09
add migrations and scaffolding for mysql datastore implementation

chriskirkland:
This is the first PR in a series which will add a new `datastore.Datastore` implementation for MySQL.  We're breaking these changes up into smaller logical chunks to make them more digestible by maintainers 🤞🏼 

Changes include:
* Scaffolding a `NewMysqlDatastore` implementation, which currently returns an `unimplemented` error for all methods required to satisfy the `datastore.Datastore` interface
* New migrations for the MySQL datastore which correctly configured the backing tables (very similar to the existing `postgres` implementation)
* Add `mysql` as a supported `datastore-engine` for `spicedb migrate`
* Integration tests to verify the correctness of the migrations

Notably, the new datastore tests are configured to create a single `mysql` container for all integration tests.  Each `setupDatabase` invocation (per-test) will create a unique database within that container. This avoids the additional container runtime startup cost, which speeds up test execution considerably (more on this in subsequent PRs).  This is a deviation from the datastore testing models for postgres/crdb which create a unique container-instance per test.

Related issue https://github.com/authzed/spicedb/issues/153


chriskirkland:

internal/datastore/mysql/migrations/driver.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if updatedCount != 1 {
+		return fmt.Errorf("writing version update affected %d rows, should be 1", updatedCount)
+	}
+
+	return nil
+}
+
+func (mysql *MysqlDriver) Dispose() {
+	defer common.LogOnError(context.Background(), mysql.db.Close)
```
ecordell, line 90: Why the defer here?

internal/datastore/common/errors.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+
+	"github.com/rs/zerolog/log"
+)
+
+// LogOnError executes the function and logs the error.
+// Useful to avoid silently ignoring errors in defer statements
+func LogOnError(ctx context.Context, f func() error) {
+	if err := f(); err != nil {
+		log.Ctx(ctx).Error().Err(err)
```
ecordell, line 13: this is pulling the logger from the context, but in both call sites we're calling `LogOnError` with an empty context. Since we don't set the [DefaultContextLogger](https://github.com/rs/zerolog/blob/c0c2e11fc3cd04ae28d856789cf58ffd1666bc3f/globals.go#L106), this is likely swallowing the error rather than logging it

internal/datastore/mysql/datastore.go
```diff
@@ -0,0 +1,114 @@
...
+	return nil, errUnimplemented
+}
+
+// IsReady returns whether the datastore is ready to accept data. Datastores that require
+// database schema creation will return false until the migrations have been run to create
+// the necessary tables.
+func (mds *mysqlDatastore) IsReady(ctx context.Context) (bool, error) {
+	return false, errUnimplemented
+}
+
```
ecordell, line 110: There's one more datastore method now; [`NamespaceCacheKey`](https://github.com/authzed/spicedb/blob/7c81b9c100aaddfe14781b125a627f4f745e1d1a/internal/datastore/datastore.go#L69). I know there's no implementation in this PR yet but it should be the same as postgres/crdb when we get there.

internal/datastore/mysql/migrations/driver.go
```diff
@@ -0,0 +1,91 @@
...
+
+/* scheme: The transport protocol to use. Use mysqlx for X Protocol connections and mysql for classic
+   MySQL protocol connections. If no protocol is specified, the server attempts to guess the protocol.
+   Connectors that support DNS SRV can use the mysqlx+srv scheme (see Connections Using DNS SRV Records). */
+// schema: The default database for the connection. If no database is specified, the connection has no default database.
+
+// NewMysqlDriver creates a new driver with active connections to the database specified.
+func NewMysqlDriver(url string) (*MysqlDriver, error) {
+	// TODO: we're currently using a DSN here, not a URI
+	dbConfig, err := sqlDriver.ParseDSN(url)
```
ecordell, line 35: need to double check, but I believe the postgres and crdb drivers take a DSN or a URI, not one or the other.

internal/datastore/mysql/migrations/zz_migration.0002_namespace_tables.go
```diff
@@ -0,0 +1,50 @@
+package migrations
+
+// namespace max size: https://buf.build/authzed/api/file/main/authzed/api/v0/core.proto#L29
```
ecordell, line 3: it looks like the initial schema is missing some indexes that the pg driver has:

https://github.com/authzed/spicedb/blob/7c81b9c100aaddfe14781b125a627f4f745e1d1a/internal/datastore/postgres/migrations/add_reverse_index.go#L4-L5
https://github.com/authzed/spicedb/blob/7c81b9c100aaddfe14781b125a627f4f745e1d1a/internal/datastore/postgres/migrations/add_transaction_timestamp_index.go#L4

is there a reason to leave those out at this stage?

internal/datastore/mysql/migrations/zz_migration.0001_initial_schema.go
```diff
@@ -0,0 +1,18 @@
+package migrations
+
+const createMysqlMigrationVersion = `CREATE TABLE mysql_migration_version (
```
ecordell, line 3: The pg driver uses migration tables that are compatible with alembic. I don't see a big reason to maintain that compatibility but figured I'd highlight the difference just in case.

ecordell (changes requested):
@chriskirkland this is a great start, thanks!

I did a first pass and left a few comments, and I think @josephschorr is planning to take a look as well

internal/datastore/mysql/migrations/driver.go
chriskirkland: Yes, I see that for `crdb` and `postgres`.  Unfortunately, it looks like the Go sql driver [only supports DSN](https://pkg.go.dev/github.com/go-sql-driver/mysql#MySQLDriver.Open) for opening connections.  I'm not sure how to resolve this at the moment.

We could certainly document the limitation for the `mysql` datastore/driver.

internal/datastore/common/errors.go
chriskirkland: Ah, that's a good call out!  I mentioned this is part of a larger set of changes that we've got in our internal fork.  Most other call sites in the change _do_ pass a context with an inject logger, but these two happen to not.

I don't think it's appropriate to set a `DefaultContextLogger` within the `mysql` or `migrations` package here as it would modify logging for the whole application. I've opted to inject the [default `log.Logger`](https://pkg.go.dev/github.com/rs/zerolog/log#pkg-variables) into the context in these two call spots.  [c6ab86b](https://github.com/authzed/spicedb/pull/411/commits/c6ab86bd42cbde03526837cab83bd50cf7fbe6a3)

Thoughts?

internal/datastore/mysql/migrations/driver.go
chriskirkland: Ah, this was an oversight. Removed [c6ab86b](https://github.com/authzed/spicedb/pull/411/commits/c6ab86bd42cbde03526837cab83bd50cf7fbe6a3)

pkg/cmd/migrate.go
```diff
@@ -106,3 +89,46 @@ func headRevisionRun(cmd *cobra.Command, args []string) error {
 	fmt.Println(headRevision)
 	return nil
 }
+
+func datastoreManagerAndDriver(datastoreEngine, dbURL string) (*migrate.Manager, migrate.Driver, error) {
+	var migrationDriver migrate.Driver
+	var migrationManager *migrate.Manager
+	var err error
+
+	if datastoreEngine == "cockroachdb" {
```
josephschorr, line 98: A `switch` might be easier to read here

internal/datastore/mysql/datastore.go
ecordell: ^ this is the cause of the current CI failures

internal/datastore/common/errors.go
ecordell: That seems reasonable for now 👍 

We should probably (separately) track an effort to make the use of a context logger consistent throughout the codebase

internal/datastore/mysql/migrations/driver.go
ecordell: I think it's not the end of the world if we can only support DSN here - spanner, for example, doesn't support anything remotely like a standard db uri, so there will be some variability regardless. But URI connection strings are common for mysql and its variants, so it would be nice if we could support it too.

PR #420, opened 2022-02-15
Fix version command

bryanhuhta:
`spicedb version` doesn't work as of https://github.com/authzed/spicedb/commit/b928d39562c0a46c840b17d60b5491ba478eb0c2. This should fix it by creating the correct command.


ecordell (approved):
LGTM

Thanks for noticing and fixing!

PR #423, opened 2022-02-18
Add a non-caching namespace manager

ecordell:
this allows certain operations to deal with non-persisted namespaces
with the same interface as persisted namespaces

### DIFF
--- a/internal/namespace/caching_test.go
+++ b/internal/namespace/caching_test.go
@@ -3,9 +3,9 @@ package namespace
 import (
 	"context"
 	"testing"
-	"time"
 
 	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
+	"github.com/dgraph-io/ristretto"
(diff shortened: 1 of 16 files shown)


internal/namespace/caching_test.go
```diff
@@ -62,3 +62,25 @@ func TestDisjointCacheKeys(t *testing.T) {
...
+
+	def, err := cache.ReadNamespace(ctx, "test/user", rev)
+	require.NoError(t, err)
+	require.Equal(t, "test/user", def.Name)
+
+	defB, err := cache.ReadNamespace(ctx, "test/user", rev)
+	require.NoError(t, err)
+	require.Equal(t, "test/user", defB.Name)
+
+	require.Equal(t, def, defB)
```
josephschorr, line 85: Add another write and make sure it immediately updates

josephschorr (approved):
LGTM

PR #426, opened 2022-02-22
internal/datastore: singlefight revision updates

jzelinskie:
This PR extracts the quantization logic used in the CRDB datastore into its own type, that could be used with other datastores. It also adds singleflight to any calls to HeadRevision so that there's no thundering herd to get timestamps from the database when the quantization window slides.


ecordell (approved):
LGTM

PR #428, opened 2022-02-24
Add command line flags for setting the sizes of caches

### LINKED ISSUE #424: Have cache sizes be configurable
Currently, they are hard coded, but as people run SpiceDB on different node sizes and types, it would be good to make them configurable (and maybe even semi-automatic if possible as an option)

josephschorr:
Fixes #424

### DIFF
--- a/internal/dispatch/caching/caching.go
+++ b/internal/dispatch/caching/caching.go
@@ -6,6 +6,7 @@ import (
 	"unsafe"
 
 	"github.com/dgraph-io/ristretto"
+	"github.com/dustin/go-humanize"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog/log"
 	"google.golang.org/protobuf/proto"
@@ -61,6 +62,8 @@ func NewCachingDispatcher(
 			BufferItems: 64,      // number of keys per Get buffer.
 			Metrics:     true,    // collect metrics.
 		}
+	} else {
+		log.Info().Int64("numCounters", cacheConfig.NumCounters).Str("maxCost", humanize.Bytes(uint64(cacheConfig.MaxCost))).Msg("configured caching dispatcher")
 	}
 
 	cache, err := ristretto.NewCache(cacheConfig)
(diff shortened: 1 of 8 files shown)


internal/dispatch/caching/caching.go
```diff
@@ -6,6 +6,7 @@ import (
 	"unsafe"
 
 	"github.com/dgraph-io/ristretto"
+	"github.com/inhies/go-bytesize"
```
jzelinskie, line 9: You might want to check out https://github.com/dustin/go-humanize for this dependency instead. It's been around longer, has way more users, and supports more than just bytes. It might be useful in other places, too.

josephschorr: Thank you! I thought I had seen a lib that did this already... changed

internal/dispatch/caching/caching.go
```diff
@@ -61,6 +62,8 @@ func NewCachingDispatcher(
 			BufferItems: 64,      // number of keys per Get buffer.
 			Metrics:     true,    // collect metrics.
 		}
+	} else {
```
ecordell, line 65: should we just always log this even if it's default?

internal/namespace/caching.go
```diff
@@ -42,6 +44,8 @@ func NewCachingNamespaceManager(
 			MaxCost:     1 << 24, // maximum cost of cache (16MB).
 			BufferItems: 64,      // number of keys per Get buffer.
 		}
+	} else {
+		log.Info().Int64("numCounters", cacheConfig.NumCounters).Str("maxCost", humanize.Bytes(uint64(cacheConfig.MaxCost))).Msg("configured caching namespace manager")
```
ecordell, line 48: same question here - always log?

internal/dispatch/caching/caching.go
josephschorr: It was making the testing output longer

ecordell (approved):
LGTM

PR #438, opened 2022-03-01
configure dispatch for tests

ecordell:
This adds a test that can fire a large set of requests at the same time to see how spicedb handles the load (likely should make it a benchmark)

This sets up spicedb with a custom resolver that will return `bufconn` connections to the dispatch api, meaning that we can easily test full cluster behavior in "unit" tests. 

This also addresses a long-standing comment in the consistency tests (to run with dispatch) and runs the dispatch tests in parallel.

### DIFF
--- a/internal/datastore/common/revisions.go
+++ b/internal/datastore/common/revisions.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -20,15 +21,40 @@ type RemoteNowFunction func(context.Context) (datastore.Revision, error)
 // RemoteClockRevisions handles revision calculation for datastores that provide
 // their own clocks.
 type RemoteClockRevisions struct {
-	QuantizationNanos      int64
-	GCWindowNanos          int64
-	FollowerReadDelayNanos int64
-	MaxRevisionStaleness   time.Duration
-	NowFunc                RemoteNowFunction
+	quantizationNanos      int64
+	gcWindowNanos          int64
+	followerReadDelayNanos int64
+	maxRevisionStaleness   time.Duration
+	nowFunc                RemoteNowFunction
 
-	updateGroup           singleflight.Group
-	lastQuantizedRevision decimal.Decimal
-	revisionValidThrough  time.Time
+	// these values are read and set by multiple consumers, they're protected
+	// by atomic load/store
+	lastQuantizedRevision *safeRevision
+	revisionValidThrough  *safeTime
+
+	// the updategroup consolidates concurrent requests to the database into 1
+	updateGroup singleflight.Group
+}
+
+// NewRemoteClockRevisions returns a RemoteClockRevisions for the given configuration
+func NewRemoteClockRevisions(quantizationNanos, gcWindowNanos, followerReadDelayNanos int64, maxRevisionStaleness time.Duration) *RemoteClockRevisions {
+	rev := safeRevision{}
+	rev.set(decimal.Zero)
+	t := safeTime{}
+	t.set(time.Time{})
+	return &RemoteClockRevisions{
+		quantizationNanos:      quantizationNanos,
+		gcWindowNanos:          gcWindowNanos,
+		followerReadDelayNanos: followerReadDelayNanos,
+		maxRevisionStaleness:   maxRevisionStaleness,
+		lastQuantizedRevision:  &rev,
+		revisionValidThrough:   &t,
+	}
+}
+
+// SetNowFunc sets the function used to determine the head revision
+func (rcr *RemoteClockRevisions) SetNowFunc(nowFunc RemoteNowFunction) {
+	rcr.nowFunc = nowFunc
 }
 
 // OptimizedRevision picks a revision that is valid for the request's
@@ -38,15 +64,16 @@ func (rcr *RemoteClockRevisions) OptimizedRevision(ctx context.Context) (datasto
 	defer span.End()
 
 	localNow := time.Now()
-	if localNow.Before(rcr.revisionValidThrough) {
-		log.Debug().Time("now", localNow).Time("valid", rcr.revisionValidThrough).Msg("returning cached revision")
-		return rcr.lastQuantizedRevision, nil
+	revisionValidThrough := rcr.revisionValidThrough.get()
+	if localNow.Before(revisionValidThrough) {
+		log.Debug().Time("now", localNow).Time("valid", revisionValidThrough).Msg("returning cached revision")
+		return rcr.lastQuantizedRevision.get(), nil
 	}
 
 	lastQuantizedRevision, err, _ := rcr.updateGroup.Do("", func() (interface{}, error) {
-		log.Debug().Time("now", localNow).Time("valid", rcr.revisionValidThrough).Msg("computing new revision")
+		log.Debug().Time("now", localNow).Time("valid", revisionValidThrough).Msg("computing new revision")
 
-		nowHLC, err := rcr.NowFunc(ctx)
+		nowHLC, err := rcr.nowFunc(ctx)
 		if err != nil {
 			return datastore.NoRevision, err
 		}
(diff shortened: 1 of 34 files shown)


internal/datastore/common/revisions.go
```diff
@@ -38,12 +41,17 @@ func (rcr *RemoteClockRevisions) OptimizedRevision(ctx context.Context) (datasto
...
 		return rcr.lastQuantizedRevision, nil
 	}
+	rcr.RUnlock()
 
 	lastQuantizedRevision, err, _ := rcr.updateGroup.Do("", func() (interface{}, error) {
+		rcr.Lock()
+		defer rcr.Unlock()
 		log.Debug().Time("now", localNow).Time("valid", rcr.revisionValidThrough).Msg("computing new revision")
 
 		nowHLC, err := rcr.NowFunc(ctx)
```
josephschorr, line 57: Does this mean the DB access is under the lock? If so, we can't have that occur.

internal/datastore/common/revisions.go
```diff
@@ -20,6 +21,8 @@ type RemoteNowFunction func(context.Context) (datastore.Revision, error)
 // RemoteClockRevisions handles revision calculation for datastores that provide
 // their own clocks.
 type RemoteClockRevisions struct {
+	sync.RWMutex
```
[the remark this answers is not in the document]
josephschorr: Could we use a CAS?

internal/dispatch/combined/combined.go
```diff
@@ -125,9 +125,12 @@ func NewDispatcher(nsm namespace.Manager, options ...Option) (dispatch.Dispatche
 // NewClusterDispatcher takes a caching redispatcher (such as one created by
 // NewDispatcher) and returns a cluster dispatcher suitable for use as the
 // dispatcher for the dispatch grpc server.
-func NewClusterDispatcher(cachingRedispatch dispatch.Dispatcher, nsm namespace.Manager, config *ristretto.Config) (dispatch.Dispatcher, error) {
+func NewClusterDispatcher(cachingRedispatch dispatch.Dispatcher, nsm namespace.Manager, prometheusSubsystem string, config *ristretto.Config) (dispatch.Dispatcher, error) {
```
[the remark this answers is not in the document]
josephschorr: Maybe we should change to a ClusterDispatcher config struct now?

internal/logging/interceptor.go
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+// Package logging is a copy of https://github.com/grpc-ecosystem/go-grpc-middleware/tree/v2/providers/zerolog
```
josephschorr, line 4: Did you file an issue with them as well?

internal/testserver/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+var testResolverbuilder = &SafeManualResolverBuilder{}
+
+func init() {
+	// register hashring balancer
+	balancer.Register(hashbalancer.NewConsistentHashringBuilder(xxhash.Sum64, 20, 1))
+
+	// register a manual resolver that we can feed addresses during tests
+	resolver.Register(testResolverbuilder)
```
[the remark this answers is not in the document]
josephschorr: Add a comment?

internal/testserver/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+		fmt.Println("ERROR UPDATING STATE", err) // shouldn't happen, log
+	}
+}
+
+// Close implements the resolver.Resolver interface
+func (r *SafeManualResolver) Close() {}
+
+// TestClusterWithDispatch creates a cluster with `size` nodes
+// The cluster has a real dispatch stack that uses bufconn grpc connections
+func TestClusterWithDispatch(t testing.TB, size int, ds datastore.Datastore) ([]*grpc.ClientConn, func()) {
```
josephschorr, line 133: Use a uint for size

internal/testserver/cluster.go
```diff
@@ -0,0 +1,224 @@
...
+	// each cluster gets a unique prefix since grpc resolution is process-global
+	prefix := getPrefix(t)
+
+	// make placeholder resolved addresses, 1 per node
+	addresses := make([]resolver.Address, 0, size)
+	for i := 0; i < size; i++ {
+		addresses = append(addresses, resolver.Address{
+			Addr:       prefix + "_" + strconv.Itoa(i),
```
josephschorr, line 141: fmt.Stringf

pkg/cmd/server/server.go
```diff
@@ -193,6 +203,7 @@ func (c *Config) Complete() (RunnableServer, error) {
 				v1SchemaServiceOption,
 			)
 		},
+		// grpc.NumStreamWorkers(uint32(runtime.NumCPU())),
```
josephschorr, line 206: Remove?

internal/testserver/cluster.go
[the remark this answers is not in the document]
ecordell: there's lots of comments on the object that implements it?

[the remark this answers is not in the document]
ecordell: added more detail here though

internal/datastore/common/revisions.go
ecordell: refactored so that this is not an issue anymore

internal/datastore/common/revisions.go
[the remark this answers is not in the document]
ecordell: I pushed an implementation that uses atomic Load/Store, reasoning that we don't really care what the value was before the new one, just that the new one is set (so we don't need the C in CAS). We're singlelfight fetching the new value so there will only ever be one goroutine updating the value at a time already.

I could see an argument for a full CAS from a sanity checking / debug logging perspective, but lmk what you think.

internal/logging/interceptor.go
ecordell: https://github.com/grpc-ecosystem/go-grpc-middleware/pull/487

Also removed the mutex; made copies of the logger instead

internal/datastore/spanner/spanner_test.go
```diff
@@ -1,3 +1,4 @@
+// go:build ci
```
josephschorr, line 1: How is this different from the line below it?

internal/datastore/common/revisions.go
```diff
@@ -2,6 +2,7 @@ package common
```
josephschorr, line 2: Can we add a test for this module that is run with the race detector in CI?

internal/dispatch/combined/combined.go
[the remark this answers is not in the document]
josephschorr: ^

internal/datastore/spanner/spanner_test.go
ecordell: they're the same, but 1.17 `go fmt` enforces both lines during the transition from `+build` to `go:build`

https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

internal/services/perf_test.go
```diff
@@ -0,0 +1,74 @@
...
+	testdatastore "github.com/authzed/spicedb/internal/testserver/datastore"
+	"github.com/authzed/spicedb/internal/testserver/datastore/config"
+	dsconfig "github.com/authzed/spicedb/pkg/cmd/datastore"
+	"github.com/authzed/spicedb/pkg/tuple"
+	"github.com/authzed/spicedb/pkg/zedtoken"
+)
+
+func TestBurst(t *testing.T) {
+	blacklist := []string{
+		"spanner", // spanner emulator doesn't support parallel transactions
```
josephschorr, line 29: Can we have the ID be pulled from the spanner package or would that create a weird dependency?

internal/testserver/datastore/datastore.go
```diff
@@ -0,0 +1,47 @@
...
+	return f(t, initFunc)
+}
+
+// NewTestDatastoreBuilder returns a TestDatastoreBuilder for a given engine
+// this is primarily useful for tests against all datastores; most other tests
+// can directly build the datastore they want
+func NewTestDatastoreBuilder(t testing.TB, engine string) TestDatastoreBuilder {
+	switch engine {
+	case "memory":
+		return TestDatastoreBuilderFunc(func(t testing.TB, initFunc InitFunc) datastore.Datastore {
```
josephschorr, line 36: why is memory special?

internal/testserver/cluster.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// Register a manual resolver.Builder  that we can feed addresses for tests
+	// Registration is not thread safe, so we register a single resolver.Builder
+	// to handle all clusters, rather than registering a unique resolver.Builder
+	// per cluster.
+	resolver.Register(testResolverBuilder)
+}
+
+// SafeManualResolverBuilder is a resolver builder
+type SafeManualResolverBuilder struct {
```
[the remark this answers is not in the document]
ecordell: it's similar to the manual resolver builder: https://pkg.go.dev/google.golang.org/grpc@v1.45.0/resolver/manual

but is thread safe

internal/testserver/datastore/datastore.go
ecordell: there's no external resource (docker image) that the "builder" needs to track, unlike with all of the other datastores.

internal/dispatch/cluster/cluster.go
```diff
@@ -0,0 +1,54 @@
...
+func CacheConfig(config *ristretto.Config) Option {
+	return func(state *optionState) {
+		state.cacheConfig = config
+	}
+}
+
+// NewClusterDispatcher takes a caching redispatcher (such as one created by
+// combined.NewDispatcher) and returns a cluster dispatcher suitable for use as
+// the dispatcher for the dispatch grpc server.
+func NewClusterDispatcher(cachingRedispatch dispatch.Dispatcher, nsm namespace.Manager, options ...Option) (dispatch.Dispatcher, error) {
```
josephschorr, line 37: Should the internal dispatcher be called `cachingDispatcher`? It doesn't *have* to be cached, right?

internal/testserver/cluster.go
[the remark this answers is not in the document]
josephschorr: Add that to the comment?

internal/testserver/cluster.go
```diff
@@ -0,0 +1,227 @@
...
+	resolvers sync.Map
+	addrs     sync.Map
+}
+
+func (b *SafeManualResolverBuilder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
+	if target.URL.Scheme != "test" {
```
josephschorr, line 74: Put into a const?

internal/dispatch/cluster/cluster.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// CacheConfig sets the configuration for the local dispatcher's cache.
+func CacheConfig(config *ristretto.Config) Option {
+	return func(state *optionState) {
+		state.cacheConfig = config
+	}
+}
+
+// NewClusterDispatcher takes a caching redispatcher (such as one created by
```
josephschorr, line 34: Fix comment

josephschorr (approved):
LGTM

PR #468, opened 2022-03-07
Remove Clone call on metadata filtering on namespaces

josephschorr:
Should reduce memory usage a bit and make loading namespaces a bit faster


ecordell (approved):
LGTM

Issue #470, opened 2022-03-08
LookupResources does not resolve all objects reachable


NickyHeuperman:
Given schema
`definition folder {
relation parent: folder
relation reader : user | group#member
relation organization : organization
permission organization_member = organization->member
permission read= (reader + parent->read) & organization_member

}
definition group{
relation user: user
relation group: group#member
permission member = user + group->user
}
definition user{}
definition organization{
relation reader: user | group#member
relation member: user | group#member
permission read = reader
}`

With relations

`folder:bottom#parent@folder:one_up
folder:one_up#parent@folder:top
group:1#user@user:1337
folder:top#reader@group:1#member
organization:1#member@user:1337
folder:top#organization@organization:1
folder:bottom#organization@organization:1
folder:one_up#organization@organization:1`

Results in the user with id 1337 having read permission on all three folders (as verified by checkpermission api)
LookupResources with Permission=read,ResourceObjectType="folder",SubjectReference.Object.ObjectType="user" and SubjectReference.Object.ObjectId however only returns two results: folders with ids top and one_up, bottom is missing.

The issue seems to lie with the intersection on organization_member irrespective of it being a synthetic relation (on the contrary, changing the intersection to & organization->member makes lookupresources only return folder with id top

PR #471, opened 2022-03-08
lookup: fall back to a slow path (list all + check) when necessary

ecordell:
Fixes https://github.com/authzed/spicedb/issues/470

### DIFF
--- a/internal/dispatch/graph/graph.go
+++ b/internal/dispatch/graph/graph.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 
 	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/rs/zerolog/log"
 	"github.com/shopspring/decimal"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -15,6 +17,7 @@ import (
 	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/internal/graph"
 	"github.com/authzed/spicedb/internal/namespace"
+	graphwalk "github.com/authzed/spicedb/pkg/graph"
 	v1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
@@ -22,6 +25,11 @@ import (
 const errDispatch = "error dispatching request: %w"
 
 var tracer = otel.Tracer("spicedb/internal/dispatch/local")
+
+var slowLookupCounter = prometheus.NewCounter(prometheus.CounterOpts{
+	Name: "spicedb_dispatch_slow_lookup_total",
+	Help: "count of how many times the slow lookup path is taken",
+})
 
 // NewLocalOnlyDispatcher creates a dispatcher that consults with the graph to formulate a response.
 func NewLocalOnlyDispatcher(
@@ -31,7 +39,7 @@ func NewLocalOnlyDispatcher(
 
 	d.checker = graph.NewConcurrentChecker(d, nsm)
 	d.expander = graph.NewConcurrentExpander(d, nsm)
-	d.lookupHandler = graph.NewConcurrentLookup(d, nsm)
+	d.lookupHandler = graph.NewConcurrentLookup(d, d, nsm)
 
 	return d
 }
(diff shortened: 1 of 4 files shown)


internal/dispatch/graph/graph.go
```diff
@@ -184,6 +186,46 @@ func (ld *localDispatcher) DispatchLookup(ctx context.Context, req *v1.DispatchL
 		Revision:              revision,
 	}
 
+	relation, err := ld.loadRelation(ctx, req.ObjectRelation.Namespace, req.ObjectRelation.Relation, revision)
+	if err != nil {
+		return &v1.DispatchLookupResponse{Metadata: emptyMetadata, ResolvedOnrs: []*v0.ObjectAndRelation{}}, nil
+	}
+
+	// check top-level rewrite
+	slowPath := false
```
josephschorr, line 195: Extract into `requiresSlowPath` function?

internal/graph/lookup.go
```diff
@@ -69,6 +73,98 @@ func (cl *ConcurrentLookup) Lookup(ctx context.Context, req ValidatedLookupReque
...
+		objects.Add(&v0.ObjectAndRelation{
+			Namespace: tpl.ObjectAndRelation.Namespace,
+			ObjectId:  tpl.ObjectAndRelation.ObjectId,
+			Relation:  req.ObjectRelation.Relation,
+		})
+	}
+
+	cancelCtx, checkCancel := context.WithCancel(ctx)
+	defer checkCancel()
+	g, checkCtx := errgroup.WithContext(cancelCtx)
```
josephschorr, line 106: newline before this line

internal/graph/lookup.go
```diff
@@ -69,6 +73,115 @@ func (cl *ConcurrentLookup) Lookup(ctx context.Context, req ValidatedLookupReque
...
+
+	g, checkCtx := errgroup.WithContext(cancelCtx)
+
+	type onrWithMeta struct {
+		onr  *v0.ObjectAndRelation
+		meta *v1.ResponseMeta
+	}
+	allowedONRs := make(chan onrWithMeta)
+
+	sem := make(chan struct{}, MaxConcurrentSlowLookupChecks)
```
jakedt, line 115: Can we add some metrics so we know how often this is happening, and what proportion of them are succeeding vs failing?

jzelinskie, line 115: This seems like a good place to use https://pkg.go.dev/golang.org/x/sync/semaphore and not just a channel.

internal/graph/lookup.go
```diff
@@ -69,6 +73,115 @@ func (cl *ConcurrentLookup) Lookup(ctx context.Context, req ValidatedLookupReque
...
+	if it.Err() != nil {
+		resp := lookupResultError(req, err, emptyMetadata)
+		return resp.Resp, resp.Err
+	}
+
+	var groupErr error
+	go func() {
+		groupErr = g.Wait()
+		close(allowedONRs)
+	}()
```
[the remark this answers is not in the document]
ecordell: closing `allowedONRs` is the stop condition for the loop that reads from `allowedONRs`, and I want to be reading from the channel while writes are still happening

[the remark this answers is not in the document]
ecordell: I came around on this and put the consumer in a goroutine in the errgroup as well

internal/dispatch/graph/graph.go
```diff
@@ -183,10 +191,55 @@ func (ld *localDispatcher) DispatchLookup(ctx context.Context, req *v1.DispatchL
...
+		return ld.lookupHandler.LookupViaChecks(ctx, validatedReq)
+	}
 
 	return ld.lookupHandler.Lookup(ctx, validatedReq)
 }
 
+func (ld *localDispatcher) requiresLookupViaChecks(relation *v0.Relation) bool {
+	// TODO: refactor walker so that we don't have to make two separate checks
+	// check top-level rewrite
+	slowPath := false
```
jzelinskie, line 211: We can probably just return early instead of using a var

internal/graph/lookup.go
```diff
@@ -69,6 +74,116 @@ func (cl *ConcurrentLookup) Lookup(ctx context.Context, req ValidatedLookupReque
 	return resolved.Resp, resolved.Err
 }
 
+// LookupViaChecks performs a slow-path lookup request with the provided request and context.
+// it performs a lookup by finding all instances of the requested object and then
```
jzelinskie, line 78: capitalize "it"

internal/dispatch/graph/graph.go
```diff
@@ -183,10 +191,55 @@ func (ld *localDispatcher) DispatchLookup(ctx context.Context, req *v1.DispatchL
...
+	rw := relation.GetUsersetRewrite()
+	if rw != nil {
```
jzelinskie, line 213: this can be done in one line

jzelinskie (approved):

jzelinskie (approved):

PR #472, opened 2022-03-09
Consolidate crdb tx retry and reset

samkim:
- Remove crdb tx savepoint usage
- Update remaining crdb tx to use datatstore execute

### DIFF
--- a/internal/datastore/crdb/namespace.go
+++ b/internal/datastore/crdb/namespace.go
@@ -79,21 +79,23 @@ func (cds *crdbDatastore) ReadNamespace(
 ) (*core.NamespaceDefinition, datastore.Revision, error) {
 	ctx = datastore.SeparateContextWithTracing(ctx)
 
-	tx, err := cds.conn.BeginTx(ctx, pgx.TxOptions{AccessMode: pgx.ReadOnly})
-	if err != nil {
-		return nil, datastore.NoRevision, fmt.Errorf(errUnableToReadConfig, err)
-	}
-	defer tx.Rollback(ctx)
-
-	if err := prepareTransaction(ctx, tx, revision); err != nil {
-		return nil, datastore.NoRevision, fmt.Errorf(errUnableToReadConfig, err)
-	}
-
-	config, timestamp, err := loadNamespace(ctx, tx, nsName)
-	if err != nil {
-		if errors.As(err, &datastore.ErrNamespaceNotFound{}) {
-			return nil, datastore.NoRevision, err
+	var config *core.NamespaceDefinition
+	var timestamp time.Time
+	if err := cds.execute(ctx, cds.conn, pgx.TxOptions{AccessMode: pgx.ReadOnly}, func(tx pgx.Tx) error {
+		var err error
+		if err := prepareTransaction(ctx, tx, revision); err != nil {
+			return fmt.Errorf(errUnableToReadConfig, err)
 		}
+
+		config, timestamp, err = loadNamespace(ctx, tx, nsName)
+		if err != nil {
+			if errors.As(err, &datastore.ErrNamespaceNotFound{}) {
+				return err
+			}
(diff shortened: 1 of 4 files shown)


internal/datastore/crdb/tx.go
```diff
@@ -184,9 +185,11 @@ func retriable(ctx context.Context, err error) bool {
 }
 
 func resetable(ctx context.Context, err error) bool {
-	// Ambiguous result error includes connection closed errors
-	// https://www.cockroachlabs.com/docs/stable/common-errors.html#result-is-ambiguous
-	return sqlErrorCode(ctx, err) == crdbAmbiguousErrorCode
+	// Unexpected EOF can be returned during high clock skew between nodes
+	return errors.Is(err, io.ErrUnexpectedEOF) ||
```
josephschorr, line 189: Is there any way we can be slightly more specific on the error, even if it does mean string matching?

josephschorr: Is this for high clock skew or for the connection being dropped?

internal/datastore/crdb/namespace.go
```diff
@@ -79,25 +79,25 @@ func (cds *crdbDatastore) ReadNamespace(
...
-	}
-	defer tx.Rollback(ctx)
-
-	if err := prepareTransaction(ctx, tx, revision); err != nil {
-		return nil, datastore.NoRevision, fmt.Errorf(errUnableToReadConfig, err)
-	}
+	var err error
+	var config *core.NamespaceDefinition
+	var timestamp time.Time
+	err = cds.execute(ctx, cds.conn, pgx.TxOptions{AccessMode: pgx.ReadOnly}, func(tx pgx.Tx) error {
```
josephschorr, line 85: So this wraps the reset handling for ReadNamespace now too?

internal/datastore/crdb/tx.go
```diff
@@ -78,10 +51,9 @@ func executeWithMaxRetries(max int) executeTxRetryFunc {
 
 // executeWithResets executes transactionFn and resets the tx when ambiguous crdb errors are encountered.
 func executeWithResets(ctx context.Context, conn conn, txOptions pgx.TxOptions, fn transactionFn, maxRetries int) (err error) {
-	var retries, resets int
+	var retries int
```
josephschorr, line 54: Shouldn't this be resets now?

josephschorr: Whichever, they just should match

internal/datastore/crdb/tx.go
```diff
@@ -129,42 +96,7 @@ func executeWithResets(ctx context.Context, conn conn, txOptions pgx.TxOptions,
...
-		}
-		return i, nil
-	}
-	return i, errors.New(errReachedMaxRetry)
-}
-
-// resetExecution attempts to rollback the given tx, begins a new tx with a new connection, and creates a savepoint.
+// resetExecution attempts to rollback the given tx and begins a new tx with a new connection.
 func resetExecution(ctx context.Context, conn conn, tx pgx.Tx, txOptions pgx.TxOptions) (newTx pgx.Tx, err error) {
 	log.Info().Msg("attempting to initialize new tx")
```
ecordell, line 101: I know this here before, but now that it's getting called on revision and namespace reads, shouldn't this be a higher log level than `Info`?

internal/datastore/crdb/tx.go
```diff
@@ -102,24 +74,19 @@ func executeWithResets(ctx context.Context, conn conn, txOptions pgx.TxOptions,
 	}()
```
ecordell, line 74: can't comment on the defer block directly, but this would unwind some of the nesting:
```go
	defer func() {
		if tx == nil {
                    return
                }
                // the rest
         }
```

ecordell (commented):
A couple of small things but looks good otherwise

PR #477, opened 2022-03-11
add prefixes to lookup metrics

ecordell:


internal/dispatch/graph/graph.go
```diff
@@ -198,13 +199,21 @@ func (ld *localDispatcher) DispatchLookup(ctx context.Context, req *v1.DispatchL
...
 		log.Warn().Msg("slow path dispatch lookup")
-		slowLookupCounter.Inc()
+		slowLookupCounter.WithLabelValues(prefix(validatedReq.Subject.Namespace)).Inc()
 		return ld.lookupHandler.LookupViaChecks(ctx, validatedReq)
 	}
 
 	return ld.lookupHandler.Lookup(ctx, validatedReq)
 }
 
+func prefix(namespace string) (prefix string) {
```
jakedt, line 209: We must have some code for this somewhere...

internal/dispatch/graph/graph.go
```diff
@@ -198,13 +199,21 @@ func (ld *localDispatcher) DispatchLookup(ctx context.Context, req *v1.DispatchL
...
+	parts := strings.Split(namespace, "/")
```
jakedt, line 210: `SplitN`

jakedt (changes requested):

internal/dispatch/graph/graph.go
ecordell: I didn't see anything that assumed the namespace string had already been validated like it is here

jakedt (approved):

PR #480, opened 2022-03-16
protect prom metric registration with a lock

ecordell:
instead of making it a global

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/authzed/grpcutil"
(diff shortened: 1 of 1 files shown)


jzelinskie:
This might be a good place for sync.Once

pkg/cmd/server/server.go
```diff
@@ -364,3 +361,18 @@ func (c *completedServerConfig) Run(ctx context.Context) error {
...
+// Normally only one server will be constructed at a time, but the protection is
+// helpful for test scenarios where multiple servers may start in parallel.
+func enableGRPCHistogram() {
+	// EnableHandlingTimeHistogram is not thread safe and only needs to happen
+	// once
+	promOnce.Do(func() {
+		grpcprom.EnableHandlingTimeHistogram(grpcprom.WithHistogramBuckets(
+			[]float64{.006, .010, .018, .024, .032, .042, .056, .075, .100, .178, .316, .562, 1.000},
+		))
+	})
```
jzelinskie, line 377: I think it's more confusing to not just inline this, rather than putting it in a function.

pkg/cmd/server/server.go
```diff
@@ -364,3 +361,18 @@ func (c *completedServerConfig) Run(ctx context.Context) error {
 
 	return nil
 }
+
+var promOnce sync.Once
+
+// enableGRPCHistogram protects prom metric registration with a sync.Once.
```
josephschorr, line 367: I'd say something like: `enableGRPCHistogram` enables the standard time history for gRPC requests, ensuring that it is only enabled once

josephschorr (approved):
LGTM

PR #481, opened 2022-03-16
Add additional error context onto schema errors

### LINKED ISSUE #447: Have type system errors raise with proper context

josephschorr:
Adds source position information onto the compiled namespace definitions and then uses it in the type system to return errors with more context

Fixes #447


proto/internal/core/v1/core.proto
```diff
@@ -211,4 +219,10 @@ message ComputedUserset {
     pattern : "^[a-z][a-z0-9_]{1,62}[a-z0-9]$",
     max_bytes : 64,
   } ];
-}
\ No newline at end of file
+  SourcePosition source_position = 3;
+}
+
+message SourcePosition {
+  uint64 zero_indexed_line_number = 1;
```
samkim, line 226: Why are these zero based but the struct and errors in common 1 based?

internal/namespace/typesystem.go
```diff
@@ -94,13 +95,32 @@ func BuildNamespaceTypeSystemForDefs(nsDef *core.NamespaceDefinition, allDefs []
 	})
 }
 
+func newErrorWithSource(withSource nspkg.WithSourcePosition, sourceCodeString string, message string, args ...interface{}) error {
+	sourcePosition := withSource.GetSourcePosition()
+	if sourcePosition != nil {
+		return commonerrors.NewErrorWithSource(
+			fmt.Errorf(message, args...),
+			sourceCodeString,
+			sourcePosition.ZeroIndexedLineNumber+1,     // +1 to make 1-indexed
```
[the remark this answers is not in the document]
josephschorr: That the first line is `1` and the first character is `1`

proto/internal/core/v1/core.proto
josephschorr: Because I prefer 0-indexed in general, but we used 1-indexed in developer API because proto doesn't have optional fields. Here, however, it is in a message which itself is optional, so I figured I could use 0 like I wanted

internal/services/v0/developer_test.go
```diff
@@ -130,18 +130,18 @@ func TestEditCheck(t *testing.T) {
 			
 			definition resource {
 				relation writer: user
-				permission writer = writer
+				    permission writer = writer
```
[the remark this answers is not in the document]
josephschorr: Was actually done on purpose to give a different column index on the error message, but I adjusted to make less odd looking as bit

pkg/commonerrors/common.go
```diff
@@ -0,0 +1,47 @@
...
+	ColumnPosition uint64
+}
+
+// InnerError returns the inner, wrapped error.
+func (ews *ErrorWithSource) InnerError() error {
+	return ews.error
+}
+
+// NewErrorWithSource creates and returns a new ErrorWithSource.
+func NewErrorWithSource(err error, sourceCodeString string, lineNumber uint64, columnPosition uint64) *ErrorWithSource {
```
samkim, line 36: Note in method doc or change line number and column position param names to include that they are 1-based to make it clear to the caller? All other line/col fields are named and/or documented so might as well for this as well.

samkim (approved):
LGTM

PR #494, opened 2022-03-22
Add support for `nil` in schema

josephschorr:
A `nil` expression under a permission is considered an empty set, and returns no permission


internal/graph/check.go
```diff
@@ -150,6 +150,10 @@ func (cc *ConcurrentChecker) checkSetOperation(ctx context.Context, req Validate
 			requests = append(requests, cc.checkUsersetRewrite(ctx, req, child.UsersetRewrite))
 		case *core.SetOperation_Child_TupleToUserset:
 			requests = append(requests, cc.checkTupleToUserset(ctx, req, child.TupleToUserset))
+		case *core.SetOperation_Child_XNil:
+			requests = append(requests, notMember())
+		default:
+			return checkError(errors.New("unknown set operation child in check"))
```
jakedt, line 156: Can you get the operation name in the error somehow? All errors should have some data that helps debug it.

pkg/namespace/builder.go
```diff
@@ -127,12 +127,20 @@ func setOperation(firstChild *core.SetOperation_Child, rest []*core.SetOperation
 }
 
 // This creates a child for a set operation that references only direct usersets with the parent relation.
+// DEPRECATED
```
jakedt, line 130: `Deprecated:`? https://go.dev/blog/godoc

jakedt (changes requested):
do we need to make it something that's not a valid relation/permission name? Do we need to audit existing schemas for nil? What happens if I attempt to define a relation named nil?

internal/graph/check.go
josephschorr: Yeah, I wasn't sure what to print out before, but `%T` seems to provide sufficient info without being too verbose

pkg/namespace/builder.go
josephschorr: Changed to a TODO to remove, otherwise it breaks lint

jakedt (approved):

PR #507, opened 2022-03-31
.github: fix passing of secrets to shared actions

jzelinskie:


jzelinskie:
It's failing because it's operating on the PR target, so it needs to be merged before it'll function.

jakedt (approved):

PR #514, opened 2022-04-01
update straggler dependencies

jakedt:


jzelinskie (changes requested):
Can you add

```yaml
- uses: "ludeeus/action-shellcheck@1.1.0"
  with:
    scandir: "./hack"
```

to here? https://github.com/authzed/spicedb/blob/main/.github/workflows/lint.yaml#L33

jakedt:
Done.

jzelinskie (approved):

PR #515, opened 2022-04-06
Telemetry stats

jakedt:


internal/telemetry/metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	ds                datastore.Datastore
+	infoDesc          *prometheus.Desc
+	objectDefsDesc    *prometheus.Desc
+	relationshipsDesc *prometheus.Desc
+	dispatchedDesc    *prometheus.Desc
+}
+
+func (c *collector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.infoDesc
+	ch <- c.objectDefsDesc
```
jzelinskie, line 103: Missing the other descs here

internal/telemetry/metrics.go
```diff
@@ -88,21 +104,59 @@ func (c *collector) Describe(ch chan<- *prometheus.Desc) {
...
-	}
+	ch <- prometheus.MustNewConstMetric(c.infoDesc, prometheus.GaugeValue, 1)
+	ch <- prometheus.MustNewConstMetric(c.objectDefsDesc, prometheus.GaugeValue, float64(len(dsStats.ObjectTypeStatistics)))
+	ch <- prometheus.MustNewConstMetric(c.relationshipsDesc, prometheus.GaugeValue, float64(dsStats.EstimatedRelationshipCount))
 
-	return float64(len(namespaces))
+	dispatchedCountMetrics := make(chan prometheus.Metric)
+	g := errgroup.Group{}
+	g.Go(func() error {
+		for {
```
jzelinskie, line 122: I think this is one of those rare cases where you can actually range over the channel

internal/telemetry/metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	dto "github.com/prometheus/client_model/go"
+	"github.com/rs/zerolog/log"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/authzed/spicedb/internal/datastore"
+	"github.com/authzed/spicedb/internal/middleware/usagemetrics"
+)
+
+var (
+	Registry                  = prometheus.NewRegistry()
```
jzelinskie, line 21: I don't think we need to export this

jzelinskie (changes requested):
only some small things

internal/telemetry/metrics.go
jakedt: Neat, I didn't even know this was a thing!

internal/telemetry/metrics.go
jakedt: I use it to export the telemetry with a different handler on `/telemetry`

jakedt: Added a method to just create a handler. Having the registry exported could invite abuse and attempts at data exfiltration.

internal/telemetry/reporter.go
```diff
@@ -0,0 +1,226 @@
...
+		for _, metric := range fam.GetMetric() {
+			var labels []*prompb.Label
+			for _, label := range metric.GetLabel() {
+				labels = append(labels, &prompb.Label{
+					Name:  label.GetName(),
+					Value: label.GetValue(),
+				})
+			}
+
+			switch *fam.Type {
```
ecordell, line 99: I think you can just use https://pkg.go.dev/github.com/prometheus/common/expfmt#ExtractSamples to avoid this, but I didn't test it

internal/telemetry/metrics.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+type collector struct {
+	ds                datastore.Datastore
+	infoDesc          *prometheus.Desc
+	objectDefsDesc    *prometheus.Desc
+	relationshipsDesc *prometheus.Desc
+	dispatchedDesc    *prometheus.Desc
+}
+
```
ecordell, line 102: I assume this is satisfying some prometheus interface, maybe add an interface assertion?

ecordell (changes requested):

internal/telemetry/reporter.go
jzelinskie: Looked into it and this looks usable.

internal/telemetry/metrics.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := g.Wait(); err != nil {
+		log.Error().Err(err).Msg("error collecting metrics")
+	}
+}
+
+// GetHTTPHandler returns a prometheus handler for the metrics being exported
+// by this collector. These metrics should NOT be scraped at a frequent
+// interval because some may be expensive to compute.
+func GetHTTPHandler() http.Handler {
+	return promhttp.HandlerFor(registry, promhttp.HandlerOpts{})
```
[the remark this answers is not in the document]
jakedt: Well, I wanted it for debugging, but it's also useful to let the user easily inspect the payload we are going to send ourselves, and if one were disabling our telemetry they could scrape it in a "standard" prometheus setup.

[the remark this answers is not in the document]
ecordell: that seems pretty reasonable 👍

internal/telemetry/reporter.go
```diff
@@ -0,0 +1,226 @@
...
+	if _, err := url.Parse(endpoint); err != nil {
+		return fmt.Errorf("invalid telemetry endpoint: %w", err)
+	}
+	return nil
+}
+
+// ReportForever reports telemetry to the endpoint until the context is
+// canceled while warning on failure.
+func ReportForever(ctx context.Context, endpoint string) error {
+	if err := ValidateEndpoint(endpoint); err != nil {
```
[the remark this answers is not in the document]
jakedt: I didn't write this, but it's defensive and it's literally a few nanoseconds to re-validate.

internal/telemetry/reporter.go
```diff
@@ -0,0 +1,226 @@
...
+	for {
+		select {
+		case <-ticker:
+			nextPush := backoffInterval.InitialInterval
+			if err := discoverAndWriteMetrics(ctx, endpoint); err != nil {
+				log.Warn().
+					Err(err).
+					Str("endpoint", endpoint).
+					Msg("failed to push telemetry metric")
+				nextPush = backoffInterval.NextBackOff()
```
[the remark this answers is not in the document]
jakedt: I don't understand the question.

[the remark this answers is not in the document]
ecordell: The normal interval is 1hr, can we exponentially backoff until the interval is 2hrs?

Don't we want a base frequency of 1hr regardless of backoff?

internal/telemetry/reporter.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	DefaultEndpoint = "https://telemetry.authzed.com"
+	MetricPrefix    = "spicedb_telemetry_"
+	Interval        = time.Hour
+)
+
+func writeTimeSeries(ctx context.Context, client *http.Client, endpoint string, ts []*prompb.TimeSeries) error {
```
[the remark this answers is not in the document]
jakedt: Is there a reason we wouldn't want to use this actual client?

internal/telemetry/reporter.go
```diff
@@ -0,0 +1,226 @@
...
+	"net/url"
+	"strconv"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/prompb"
```
jzelinskie, line 23: We can drop importing all of prometheus if we use this:

`go.buf.build/protocolbuffers/go/prometheus/prometheus`

ecordell (dismissed):
LGTM

internal/telemetry/reporter.go
```diff
@@ -0,0 +1,232 @@
...
+				MinVersion: tls.VersionTLS12,
+			},
+		}
+
+		client.Transport = t
+	}
+
+	return func(ctx context.Context) error {
+		// Smear the startup delay out over 10% of the reporting interval
+		startupDelay := time.Duration(rand.Int63n(int64(interval.Seconds()/10))) * time.Second
```
ecordell, line 168: Since that could be up to 10m after startup, maybe we should attempt to write on shutdown if we haven't written at all yet and the server has at least successfully started up. And actually, maybe we want separate telemetry on startup error rates. 

I realize this is a fairly involved change for an uncommon edge case, maybe we just make an issue and do it later.

ecordell (approved):
LGTM

just had one thing we might want to think about for the future

PR #517, opened 2022-04-10
Implement a reachability graph

josephschorr:
The reachability graph is a helper system for computing the entrypoint relations/permissions for a subject of a specific type when trying to *reach* a target relation/permission. This will form the basis of the reachable resources dispatch and API, and hopefully be used to significantly narrow down the working set for lookup in the future.


vroldanbet (dismissed):
Changes to MySQL datastore tests look good - sorry for the oversight!

internal/namespace/reachabilitygraphbuilder.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	graph := &core.ReachabilityGraph{
+		EntrypointsBySubjectType:     map[string]*core.ReachabilityEntrypoints{},
+		EntrypointsBySubjectRelation: map[string]*core.ReachabilityEntrypoints{},
+	}
+
+	usersetRewrite := targetRelation.GetUsersetRewrite()
+	if usersetRewrite != nil {
+		return graph, computeRewriteReachability(graph, usersetRewrite, core.ReachabilityEntrypoint_DIRECT_OPERATION_RESULT, targetRelation, ts, option)
```
ecordell, line 33: I'm confused why we would be passing in `core.ReachabilityEntrypoint_DIRECT_OPERATION_RESULT` here when the whole point of `computeRewriteReachability` is to determine what type of reachability this rewrite has.

internal/namespace/reachabilitygraphbuilder.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func computeRewriteReachability(graph *core.ReachabilityGraph, rewrite *core.UsersetRewrite, operationResultState core.ReachabilityEntrypoint_EntrypointResultStatus, targetRelation *core.Relation, ts *NamespaceTypeSystem, option reachabilityOption) error {
+	switch rw := rewrite.RewriteOperation.(type) {
+	case *core.UsersetRewrite_Union:
+		return computeRewriteOpReachability(rw.Union.Child, operationResultState, graph, targetRelation, ts, option)
+
+	case *core.UsersetRewrite_Intersection:
+		// If optimized mode is set, only return the first child of the intersection.
+		if option == reachabilityOptimized {
```
ecordell, line 48: I thought we would be storing the full reachability graph (since it's static per schema revision) and then pruning the branches at runtime based on heuristics.

These seems to be optimizing for calculating a pruned reachability graph rather than pruning an existing graph.

proto/internal/core/v1/core.proto
```diff
@@ -114,50 +130,221 @@ message Metadata {
...
+ *     relation viewer: user
+ *     relation owner: user
+ *     permission view = viewer + owner + org->admin
+ *   }
+ * ```
+ *
+ * The reachability graph for `viewer` and the other relations will have entrypoints for each
+ * subject type found for those relations.
+ * 
+ * The reachability graph for the `view` relation will have three entrypoints, representing:
```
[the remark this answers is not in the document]
josephschorr: Conditional is an implementation of the Go side structure, not the protos

internal/namespace/oppath.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+/**
+ * decorateRelationOpPaths decorates all SetOperations found within the given relation's rewrite
+ * (if any) with a path indicating each operation's position in the tree of operations.
```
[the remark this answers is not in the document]
josephschorr: > Do we get any benefit from re-using the BDD representations here?

I don't believe so; this is just providing a unique ID for each operation so we can jump into the middle if necessary.

> It might be easy to just ask BDD "are there any negated variables" to determine whether the branch is "conditional"?

Yes, but its not simply if there is one, but whether its reachable.

internal/namespace/oppath.go
```diff
@@ -0,0 +1,52 @@
...
+	case *core.UsersetRewrite_Intersection:
+		return decorateRelationRewriteChildrenWithPath(rw.Intersection, parentPath)
+	case *core.UsersetRewrite_Exclusion:
+		return decorateRelationRewriteChildrenWithPath(rw.Exclusion, parentPath)
+	default:
+		return fmt.Errorf("unknown type of rewrite operation in rewrite path annotator: %T", rw)
+	}
+}
+
+func decorateRelationRewriteChildrenWithPath(so *core.SetOperation, parentPath []uint32) error {
```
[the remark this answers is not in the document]
josephschorr: Right now, yes. It will be used eventually though

internal/namespace/reachabilitygraphbuilder.go
josephschorr: Because until we see an intersection or exclusion, it is assumed to be a direct result. Only once we've seen one (or more) of those operations, do we need to make the entire thing conditional

internal/namespace/reachabilitygraphbuilder.go
josephschorr: That's how I originally had it, but pruning became a bit of a mess, so I decided to change to a simpler model where we just compute it in real time. I figure we use real time computation until it no longer scales.

internal/namespace/reachabilitygraph.go
```diff
@@ -0,0 +1,156 @@
...
+// reachability graph, starting at the given subject type and walking to the given resource type.
+//
+// The optimized set will skip branches on intersections and exclusions in an attempt to minimize
+// the number of entrypoints.
+func (rg *ReachabilityGraph) OptimizedEntrypointsForSubjectToResource(
+	ctx context.Context,
+	subjectType *core.RelationReference,
+	resourceType *core.RelationReference,
+) ([]ReachabilityEntrypoint, error) {
+	return rg.entrypointsForSubjectToResource(ctx, subjectType, resourceType, reachabilityOptimized)
```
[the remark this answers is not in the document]
josephschorr: Optimized right now is only doing the pruning, it is not concerned with which branch is better. To do so will require more exploratory work in my opinion, so I wanted to do the simplest thing now until such time as we find a good solution for a stats based solution. For example, a branch may be better because its *intermediate* object sets are smaller, even if the leaf is larger, so we definitely need to have a discussion about how to pick what is "best"

josephschorr:
Open items:

1) Figuring out why the consistency tests occasionally show the following error (likely something being closed early on the reachable resources stream):

```
rpc error: code = Unknown desc = error in reachablility: rpc error: code = Internal desc = transport: transport: the stream is done or WriteHeader was already called
```

2) Adding proper dispatch count tracking (and testing for it) to reachable resources
3) Adding caching for reachable resources

josephschorr:
Open items:

1) Figuring out why the consistency tests occasionally show the following error (likely something being closed early on the reachable resources stream):

internal/graph/lookup.go
```diff
@@ -43,864 +38,97 @@ type ValidatedLookupRequest struct {
...
 	}
 
-	cancelCtx, checkCancel := context.WithCancel(ctx)
-	defer checkCancel()
+	(func() {
+		ls.mu.Lock()
+		defer ls.mu.Unlock()
 
-	g, checkCtx := errgroup.WithContext(cancelCtx)
+		ls.dispatchCount += result.Metadata.DispatchCount
```
ecordell, line 66: these could be outside of the mutex as `sync/atomic` operations instead?

internal/graph/parallelchecker.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			pc.g.Go(func() error {
+				defer sem.Release(1)
+				res, err := pc.c.DispatchCheck(pc.checkCtx, req)
+				if err != nil {
+					return err
+				}
+
+				(func() {
```
ecordell, line 95: you can drop the outer parens

internal/graph/parallelchecker.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (pc *ParallelChecker) CachedDispatchCount() uint32 {
+	return pc.cachedDispatchCount
+}
+
+func (pc *ParallelChecker) DepthRequired() uint32 {
+	return pc.depthRequired
+}
+
+func (pc *ParallelChecker) addResultUnderLock(resource *core.ObjectAndRelation) {
```
ecordell, line 57: I think you can interpret this name to mean the opposite of what you actually mean; I will always have to come check the implementation. 

Maybe just drop these helpers?

internal/graph/parallelchecker.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+type ParallelChecker struct {
+	toCheck       chan *v1.DispatchCheckRequest
+	c             dispatch.Check
+	g             *errgroup.Group
+	checkCtx      context.Context
+	subject       *core.ObjectAndRelation
+	maxConcurrent uint8
```
ecordell, line 23: why not just int64 since that's what the semaphore lib expects?

internal/dispatch/stream.go
```diff
@@ -0,0 +1,115 @@
...
+	"context"
+	"sync"
+
+	grpc "google.golang.org/grpc"
+)
+
+// Stream defines the interface generically matching a streaming dispatch response.
+type Stream[T any] interface {
+	// Publish publishes the result to the stream.
+	Publish(T) error
```
ecordell, line 13: Why not `Send` here too to match grpc / channel naming?

internal/graph/lookup.go
```diff
@@ -43,864 +38,97 @@ type ValidatedLookupRequest struct {
...
 
-			resultChan <- lookupResult(req, objects.AsSlice(), emptyMetadata)
-		})
-	}
+	// Start the checker.
+	checker.Start()
 
-	// Dispatch a check for the subject wildcard, if allowed.
-	isWildcardAllowed, err := typeSystem.IsAllowedPublicNamespace(req.ObjectRelation.Relation, req.Subject.Namespace)
+	// TODO(jschorr): Collect dispatch count and depth required.
```
ecordell, line 98: What's the plan for reporting this? Add the count to the lookup count and take the max depth?

internal/graph/lookup.go
```diff
@@ -43,864 +38,97 @@ type ValidatedLookupRequest struct {
...
-		))
-	}
-
-	// Dispatch a check for the target ONR directly, if it is allowed on the start relation.
-	isDirectAllowed, err := typeSystem.IsAllowedDirectRelation(req.ObjectRelation.Relation, req.Subject.Namespace, req.Subject.Relation)
-	if err != nil {
-		return returnResult(lookupResultError(req, err, emptyMetadata))
+func (cl *ConcurrentLookup) LookupViaReachability(ctx context.Context, req ValidatedLookupRequest) (*v1.DispatchLookupResponse, error) {
+	if req.Subject.ObjectId == tuple.PublicWildcard {
+		resp := lookupResultError(req, NewErrInvalidArgument(errors.New("cannot perform lookup on wildcard")), emptyMetadata)
```
ecordell, line 85: If we made this a streaming API w/ paging, maybe this is not so unreasonable?

internal/graph/parallelchecker.go
```diff
@@ -0,0 +1,120 @@
...
+		if err := sem.Acquire(pc.checkCtx, int64(pc.maxConcurrent)); err != nil {
+			return err
+		}
+		return nil
+	})
+}
+
+func (pc *ParallelChecker) Finish() (*tuple.ONRSet, error) {
+	close(pc.toCheck)
+	if err := pc.g.Wait(); err != nil {
```
ecordell, line 115: Might be nicer to make this `Wait()` to match errgroup and friends?

Also maybe a comment that you should only call this once there can be no more calls to `QueueCheck`?

internal/graph/parallelchecker.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+type ParallelChecker struct {
+	toCheck       chan *v1.DispatchCheckRequest
+	c             dispatch.Check
+	g             *errgroup.Group
+	checkCtx      context.Context
+	subject       *core.ObjectAndRelation
+	maxConcurrent uint8
+	results       *tuple.ONRSet
```
ecordell, line 24: since the result set is grow-only, a `sync.Map` should work well here

internal/graph/reachableresources.go
```diff
@@ -0,0 +1,309 @@
...
+								AtRevision:     req.Revision.String(),
+								DepthRemaining: req.Metadata.DepthRemaining - 1,
+							},
+						},
+					))
+				}
+			}
+
+		default:
+			panic(fmt.Sprintf("Unknown kind of entrypoint: %v", entrypoint.EntrypointKind()))
```
ecordell, line 189: so if we add a new kind of entrypoint, we can't do a rolling restart?

internal/graph/lookup.go
```diff
@@ -24,15 +19,15 @@ import (
 const MaxConcurrentSlowLookupChecks = 10
 
 // NewConcurrentLookup creates and instance of ConcurrentLookup.
-func NewConcurrentLookup(d dispatch.Lookup, c dispatch.Check, nsm namespace.Manager) *ConcurrentLookup {
-	return &ConcurrentLookup{d: d, c: c, nsm: nsm}
+func NewConcurrentLookup(c dispatch.Check, r dispatch.ReachableResources, nsm namespace.Manager) *ConcurrentLookup {
```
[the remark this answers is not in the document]
josephschorr: We broke this out before in case we wanted to have different impls

internal/graph/lookup.go
```diff
@@ -43,864 +38,97 @@ type ValidatedLookupRequest struct {
...
-			Relation:  req.ObjectRelation.Relation,
-		})
+func (ls *collectingStream) Publish(result *v1.DispatchReachableResourcesResponse) error {
+	if result == nil {
+		panic("Got nil result")
 	}
 
-	cancelCtx, checkCancel := context.WithCancel(ctx)
-	defer checkCancel()
+	(func() {
```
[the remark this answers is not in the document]
josephschorr: Oh really? good to know

internal/graph/parallelchecker.go
josephschorr: I was hoping to make it clear it needs to be under the lock... any ideas on how to name it?

internal/dispatch/stream.go
josephschorr: I *explicitly* named it differently to prevent sending a gRPC stream (which is not concurrent safe) into this interface (which all callers expect to be safe to use concurrently)

internal/graph/lookup.go
josephschorr: Maybe, but I think for now its fine to keep this as-is

internal/graph/reachableresources.go
josephschorr: Not until the new code is deployed; this is to ensure we don't accidentally miss handling a kind of entrypoint and therefore miss loading the data on older code; it is better IMO to panic and fail the API call than to return incomplete data

internal/graph/parallelchecker.go
josephschorr: Since we need the lock for the stats update anyway, I figure we just keep it and do both changes

josephschorr:
Updated

ecordell (approved):
LGTM

I wasn't able to come up with any edge cases that need addressing, aside from the ones you've fixed by deduplicating in the `ParallelChecker`

Maybe make a tracking issue for caching the graph?

PR #519, opened 2022-04-11
introduces mysql migration driver

vroldanbet:
This is an initial contribution of the MySQL Datastore implementation. It tries to decompose the submission into smaller PRs to help with the review.

# What

Implements SpiceDB migration logic to support MySQL as backing datastore

# Context

GitHub runs mysql schema migrations at scale using skeema (https://github.com/skeema/skeema) and gh-ost (https://github.com/github/gh-ost). The SpiceDB migration driver for MySQL is designed in a way it is compatible with migration workloads relying on those tools.

The main difference is how the MySQL migration driver maintains the metadata needed to know which version is currently running. The PSQL migration driver follows a pattern similar to what's proposed by SQLAlchemy's Alembic, with a table that keeps track of the migration run. Because Skeema does not support data migrations, and seeding the database
could be considered a data migration, we decided to encode the metadata into the database schema (DML) so we can expose the same information but only relying on schema migrations. As such the migration_version table
is empty, and maintains the running version in a table column. This turns every migration into a pure DML operations, which Skeema understands well.

The way we run these then is by executing `spicedb migrate` in development DB, and letting skeema detect the changes to the development database. These are then persisted, and deployed using our own internal migration automation build on top of skeema and gh-ost.

The other distinction with the PSQL migration driver is the support for table prefixes. This is necessary to align with our own internal conventions.

# Contributors

These series of PRs are the result of the work from various current and former hubbers which I've added as co-authors, in alphabetic order:

@ams11
@chriskirkland
@hagould
@johnpreed
@sbryant

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/fatih/color v1.13.0
 	github.com/fatih/structs v1.1.0
 	github.com/go-co-op/gocron v1.13.0
-	github.com/go-sql-driver/mysql v1.6.0
+	github.com/go-sql-driver/mysql v1.5.0
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/snappy v0.0.4
 	github.com/google/go-cmp v0.5.7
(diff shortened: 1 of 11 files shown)


vroldanbet:
This is now merged as part of https://github.com/authzed/spicedb/pull/525
Closing!

PR #526, opened 2022-04-12
k8s: add RBAC and flesh out example

### LINKED ISSUE #499: Basic k8s example doesn't specify RBAC required to use kuberesolver dispatch
Need to add the following:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: spicedb
  namespace: default
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  namespace: default
  name: watch-service
rules:
- apiGroups: [""] # "" indicates the core API group
  resources: ["endpoints"]
  verbs: ["get", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: spicedb-watch-service
  namespace: default
subjects:
- kind: ServiceAccount
  name: spicedb
roleRef:
  kind: Role
  name: watch-service
  apiGroup: rbac.authorization.k8s.io
```

jzelinskie:
Fixes #499


k8s/example.yaml
```diff
@@ -0,0 +1,182 @@
...
+kind: "Role"
+metadata:
+  name: "watch-service"
+  labels:
+    app.kubernetes.io/name: "spicedb"
+    app.kubernetes.io/version: "example"
+rules:
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    verbs: ["get", "watch"]
```
ecordell, line 167: you should need `list` too?

jzelinskie: not according to the kuberesolver readme

ecordell: oh, super weird that it directly queries instead of using a client-go informer (which would've used list+watch)

ecordell (approved):
LGTM

PR #532, opened 2022-04-14
mysql: wire up the mysql datastore engine to the CLI

sbryant:
This PR adds the mysql datastore options to the CLI.

- add datastore-engine mysql
- `datastore-mysql-table-prefix` support
- `spicedb migrate` support

### DIFF
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -12,6 +12,7 @@ import (
 	"github.com/authzed/spicedb/internal/datastore"
 	"github.com/authzed/spicedb/internal/datastore/crdb"
 	"github.com/authzed/spicedb/internal/datastore/memdb"
+	"github.com/authzed/spicedb/internal/datastore/mysql"
 	"github.com/authzed/spicedb/internal/datastore/postgres"
 	"github.com/authzed/spicedb/internal/datastore/proxy"
 	"github.com/authzed/spicedb/internal/datastore/spanner"
@@ -25,6 +26,7 @@ const (
 	PostgresEngine  = "postgres"
 	CockroachEngine = "cockroachdb"
 	SpannerEngine   = "spanner"
+	MySQLEngine     = "mysql"
 )
 
 var BuilderForEngine = map[string]engineBuilderFunc{
@@ -32,6 +34,7 @@ var BuilderForEngine = map[string]engineBuilderFunc{
 	PostgresEngine:  newPostgresDatastore,
 	MemoryEngine:    newMemoryDatstore,
 	SpannerEngine:   newSpannerDatastore,
+	MySQLEngine:     newMySQLDatastore,
 }
 
 //go:generate go run github.com/ecordell/optgen -output zz_generated.options.go . Config
@@ -76,6 +79,9 @@ type Config struct {
 	// Internal
 	WatchBufferLength      uint16
 	EnableDatastoreMetrics bool
+
+	// MySQL
+	TablePrefix string
 }
 
 // RegisterDatastoreFlags adds datastore flags to a cobra command
@@ -105,6 +111,7 @@ func RegisterDatastoreFlags(cmd *cobra.Command, opts *Config) {
 	cmd.Flags().StringVar(&opts.OverlapStrategy, "datastore-tx-overlap-strategy", "static", `strategy to generate transaction overlap keys ("prefix", "static", "insecure") (cockroach driver only)`)
 	cmd.Flags().StringVar(&opts.OverlapKey, "datastore-tx-overlap-key", "key", "static key to touch when writing to ensure transactions overlap (only used if --datastore-tx-overlap-strategy=static is set; cockroach driver only)")
 	cmd.Flags().StringVar(&opts.SpannerCredentialsFile, "datastore-spanner-credentials", "", "path to service account key credentials file with access to the cloud spanner instance")
+	cmd.Flags().StringVar(&opts.TablePrefix, "datastore-mysql-table-prefix", "", "prefix to add to the name of all SpiceDB database tables")
 }
 
 func DefaultDatastoreConfig() *Config {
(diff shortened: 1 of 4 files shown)


pkg/cmd/datastore/datastore.go
```diff
@@ -76,6 +79,9 @@ type Config struct {
 	// Internal
 	WatchBufferLength      uint16
 	EnableDatastoreMetrics bool
+
+	// MySQL
```
josephschorr, line 83: Put above the `Internal` section

pkg/cmd/migrate.go
```diff
@@ -9,16 +9,18 @@ import (
...
+	mysqlmigrations "github.com/authzed/spicedb/internal/datastore/mysql/migrations"
 	"github.com/authzed/spicedb/internal/datastore/postgres/migrations"
 	spannermigrations "github.com/authzed/spicedb/internal/datastore/spanner/migrations"
 	"github.com/authzed/spicedb/pkg/cmd/server"
 	"github.com/authzed/spicedb/pkg/migrate"
 )
 
 func RegisterMigrateFlags(cmd *cobra.Command) {
-	cmd.Flags().String("datastore-engine", "memory", `type of datastore to initialize ("memory", "postgres", "cockroachdb")`)
+	cmd.Flags().String("datastore-engine", "memory", `type of datastore to initialize ("memory", "postgres", "cockroachdb", "mysql")`)
```
josephschorr, line 20: mind adding `spanner` in here as well?

pkg/cmd/migrate.go
```diff
@@ -113,6 +128,8 @@ func HeadRevision(engine string) (string, error) {
 		return crdbmigrations.CRDBMigrations.HeadRevision()
 	case "postgres":
 		return migrations.DatabaseMigrations.HeadRevision()
+	case "mysql":
```
josephschorr, line 131: mind adding spanner in here as well?

sbryant: Sure thing

pkg/cmd/migrate.go
sbryant: :+1:

pkg/cmd/datastore/datastore.go
```diff
@@ -236,6 +243,21 @@ func newSpannerDatastore(opts Config) (datastore.Datastore, error) {
...
+	mysqlOpts := []mysql.Option{
+		mysql.GCInterval(opts.GCInterval),
+		mysql.GCWindow(opts.GCWindow),
+		mysql.GCInterval(opts.GCInterval),
+		mysql.ConnMaxIdleTime(opts.MaxIdleTime),
+		mysql.ConnMaxLifetime(opts.MaxLifetime),
+		mysql.MaxOpenConns(opts.MaxOpenConns),
+		mysql.RevisionFuzzingTimedelta(opts.RevisionQuantization),
+		mysql.TablePrefix(opts.TablePrefix),
+		mysql.EnablePrometheusStats(),
```
vroldanbet, line 256: @sbryant I ended up enabling by default the prometheus stats, I couldn't figure out how to wire it up with `EnableDatastoreMetrics`, [just like the PSQL driver recently did](https://github.com/vroldanbet/spicedb/blob/0134a627bc5827c7e9929c1cf7e58c2e4289dd58/pkg/cmd/datastore/datastore.go#L222-L224).

@ecordell since you introduced that internal flag, is that something that can be passed as an ENV variable via CLI? I couldn't figure out why it wasn't being parsed.

sbryant: The way I read to cobrautil setup is that a flag still needs to be defined for the env var approach to work? https://github.com/jzelinskie/cobrautil/blob/main/cobrautil.go#L38

vroldanbet (dismissed):
LGTM, we can follow up on the prometheus flag. Deferring to @authzed/spicedb-maintainers

pkg/cmd/datastore/datastore.go
ecordell: Correct - right now that option is not bound to a cli flag. If you bind `EnableDatastoreMetrics` to a flag in `RegisterDatastoreFlags` then you'll be able to set it via CLI or env vars

sbryant: @ecordell how does that flag even work then? How are metrics ever turned on?

vroldanbet: @authzed test ping

ecordell: @sbryant It wasn't being used anywhere, it was a flag we could turn on during tests if we wanted to (via code)

sbryant: > @sbryant It wasn't being used anywhere, it was a flag we could turn on during tests if we wanted to (via code)

Ahh okay! Thanks for the explanation

vroldanbet: @ecordell as it stands, it means that postgresql prometheus metrics are off by default, and there is no way to turn them on via the CLI? Is that intentional? They used to be _on_ by default so that would be a regression for any customers running PSQL

pkg/cmd/migrate.go
```diff
@@ -86,7 +101,7 @@ func migrateRun(cmd *cobra.Command, args []string) error {
 }
 
 func RegisterHeadFlags(cmd *cobra.Command) {
-	cmd.Flags().String("datastore-engine", "postgres", "type of datastore to initialize (e.g. postgres, cockroachdb, memory")
+	cmd.Flags().String("datastore-engine", "postgres", "type of datastore to initialize (e.g. postgres, cockroachdb, memory, mysql")
```
josephschorr, line 104: Still missing `spanner`

sbryant: Thanks, missed this!

sbryant:
@josephschorr i believe this is ready to go, is there anything else missing?

pkg/cmd/datastore/datastore.go
josephschorr: Sounds like this may need to be addressed, but a followup should do

josephschorr (approved):
LGTM

pkg/cmd/datastore/datastore.go
ecordell: @vroldanbet You're right, they were on by default. I've filed https://github.com/authzed/spicedb/issues/549 to follow up and change the default

sbryant:
The failures appear to be unrelated to this PR. I can't seem to force an retry workflow either :(

josephschorr:
I've rerun them

pkg/cmd/migrate.go
```diff
@@ -67,6 +69,19 @@ func migrateRun(cmd *cobra.Command, args []string) error {
...
+		var err error
+		tablePrefix, err := cmd.Flags().GetString("datastore-mysql-table-prefix")
+		if err != nil {
+			log.Fatal().Msg(fmt.Sprintf("unable to get table prefix: %s", err))
+		}
+
+		migrationDriver, err = mysqlmigrations.NewMySQLDriverFromDSN(dbURL, tablePrefix)
+		if err != nil {
+			log.Fatal().Err(err).Msg("unable to create migration driver")
+		}
```
vroldanbet, line 84: @sbryant I believe there is a bug here: `manager` is not being set to the MySQL migration manager.

vroldanbet (commented):
found a bug, let's fix on a follow up!

pkg/cmd/migrate.go
vroldanbet: follow up in https://github.com/authzed/spicedb/pull/550

PR #533, opened 2022-04-14
Switch MySQL tests to explicitly specify amd64

josephschorr:
Necessary because the mysql:5 image does not have an arm64 build


vroldanbet (approved):
LGTM!

Issue #288, opened 2021-11-15
Support Multiple Preshared Keys


ecordell:
The `--grpc-preshared-key` flag currently takes a single string as a value.

This should instead support multiple values (i.e. with a cobra `StringSlice`) so that users can rotate keys easily.

relcodedev:
I see it is used with multiple --grpc-preshared-key at command line.  Is there a way to use environment variables also?

josephschorr:
You can specify the different keys using a comma between them

relcodedev:
[like]  Edgar Robles reacted to your message:
________________________________

PR #537, opened 2022-04-18
Add support for multiple preshared keys

josephschorr:
Fixes https://github.com/authzed/spicedb/issues/288
Fixes AUTHZ-500

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -51,6 +51,11 @@ jobs:
         with:
           go-version: "^1.17"
       - uses: "authzed/actions/docker-build@main"
+        with:
+          push: false
+          tags: "authzed/spicedb:ci"
+          buildx: false
+          qemu: false
       - uses: "authzed/actions/go-test@main"
         with:
           working_directory: "cmd/spicedb"
--- a/cmd/spicedb/serve_integration_test.go
+++ b/cmd/spicedb/serve_integration_test.go
@@ -25,7 +25,7 @@ func TestServe(t *testing.T) {
 	tester, err := newTester(t,
 		&dockertest.RunOptions{
 			Repository:   "authzed/spicedb",
-			Tag:          "latest",
+			Tag:          "ci",
 			Cmd:          []string{"serve", "--log-level", "debug", "--grpc-preshared-key", "firstkey", "--grpc-preshared-key", "secondkey"},
 			ExposedPorts: []string{"50051/tcp"},
 		},
--- a/cmd/spicedb/serve_integration_test.go
+++ b/cmd/spicedb/serve_integration_test.go
@@ -1,0 +1,73 @@
+//go:build docker
+// +build docker
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/authzed/grpcutil"
+	"github.com/ory/dockertest/v3"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/status"
+)
+
+func TestServe(t *testing.T) {
+	requireParent := require.New(t)
+
+	tester, err := newTester(t,
+		&dockertest.RunOptions{
+			Repository:   "authzed/spicedb",
+			Tag:          "latest",
+			Cmd:          []string{"serve", "--log-level", "debug", "--grpc-preshared-key", "firstkey", "--grpc-preshared-key", "secondkey"},
+			ExposedPorts: []string{"50051/tcp"},
+		},
+		"firstkey",
+	)
+	requireParent.NoError(err)
+	defer tester.cleanup()
+
+	for key, expectedWorks := range map[string]bool{
+		"":           false,
+		"firstkey":   true,
+		"secondkey":  true,
+		"anotherkey": false,
+	} {
+		t.Run(key, func(t *testing.T) {
+			require := require.New(t)
+
+			opts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
+			if key != "" {
+				opts = append(opts, grpcutil.WithInsecureBearerToken(key))
+			}
+			conn, err := grpc.Dial(fmt.Sprintf("localhost:%s", tester.port), opts...)
+
+			require.NoError(err)
+			defer conn.Close()
+
+			resp, err := healthpb.NewHealthClient(conn).Check(context.Background(), &healthpb.HealthCheckRequest{Service: "authzed.api.v1.SchemaService"})
+			require.NoError(err)
+			require.Equal(healthpb.HealthCheckResponse_SERVING, resp.GetStatus())
+
+			client := v1.NewSchemaServiceClient(conn)
+			_, err = client.WriteSchema(context.Background(), &v1.WriteSchemaRequest{
+				Schema: `definition user {}`,
+			})
+
+			if expectedWorks {
+				require.NoError(err)
+			} else {
+				s, ok := status.FromError(err)
+				require.True(ok)
+				require.Equal(codes.Unknown, s.Code())
+			}
+		})
+	}
+}
--- a/cmd/spicedb/serve_integration_test.go
+++ b/cmd/spicedb/serve_integration_test.go
@@ -66,7 +66,12 @@ func TestServe(t *testing.T) {
 			} else {
 				s, ok := status.FromError(err)
 				require.True(ok)
-				require.Equal(codes.Unknown, s.Code())
+
+				if key == "" {
+					require.Equal(codes.Unauthenticated, s.Code())
+				} else {
+					require.Equal(codes.PermissionDenied, s.Code())
+				}
 			}
 		})
 	}
--- a/cmd/spicedb/servetesting_integration_test.go
+++ b/cmd/spicedb/servetesting_integration_test.go
@@ -189,8 +189,13 @@ func newTester(t *testing.T, containerOpts *dockertest.RunOptions, token string)
 			}
 			`,
 		})
+
+		if err != nil {
+			s, ok := status.FromError(err)
+			require.True(t, !ok || s.Code() == codes.Unavailable, fmt.Sprintf("Found unexpected error: %v", err))
+		}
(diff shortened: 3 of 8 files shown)


pkg/cmd/server/server.go
```diff
@@ -339,7 +352,7 @@ type completedServerConfig struct {
 
 	unaryMiddleware     []grpc.UnaryServerInterceptor
 	streamingMiddleware []grpc.StreamServerInterceptor
-	presharedKey        string
+	presharedKey        []string
```
jzelinskie, line 355: presharedKey => presharedKeys

pkg/cmd/server/server.go
```diff
@@ -103,7 +103,11 @@ func (c *Config) Complete() (RunnableServer, error) {
 	}
 
 	if c.GRPCAuthFunc == nil {
-		log.Trace().Int("preshared-key-length", len(c.PresharedKey)).Msg("using gRPC auth with preshared key")
+		log.Trace().Int("preshared-keys", len(c.PresharedKey)).Msg("using gRPC auth with preshared key(s)")
```
jzelinskie, line 106: preshared-keys-count

pkg/cmd/server/server.go
```diff
@@ -141,10 +145,15 @@ func (c *Config) Complete() (RunnableServer, error) {
 			return nil, fmt.Errorf("failed to create dispatcher: %w", cerr)
 		}
 
+		presharedKey := ""
```
jzelinskie, line 148: presharedKey  => dispatchKey

pkg/cmd/server/server.go
```diff
@@ -141,10 +145,15 @@ func (c *Config) Complete() (RunnableServer, error) {
...
+		if len(c.PresharedKey) > 0 {
```
jzelinskie, line 149: Why check this? Shouldn't it error before here if `len(c.PresharedKeys) == 0`?

pkg/cmd/server/server.go
```diff
@@ -103,7 +103,11 @@ func (c *Config) Complete() (RunnableServer, error) {
 	}
```
jzelinskie, line 103: here we should check that each preshared key is not the empty string

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,86 @@
...
+				)
+			}
+			require.NoError(err)
+			defer conn.Close()
+
+			resp, err := healthpb.NewHealthClient(conn).Check(context.Background(), &healthpb.HealthCheckRequest{Service: "authzed.api.v1alpha1.SchemaService"})
+			require.NoError(err)
+			require.Equal(healthpb.HealthCheckResponse_SERVING, resp.GetStatus())
+
+			client := v1alpha1.NewSchemaServiceClient(conn)
```
jzelinskie, line 63: why not use v1?

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,86 @@
...
+			var conn *grpc.ClientConn
+			var err error
+			if key == "" {
+				conn, err = grpc.Dial(fmt.Sprintf("localhost:%s", tester.port), grpc.WithTransportCredentials(insecure.NewCredentials()))
+			} else {
+				conn, err = grpc.Dial(fmt.Sprintf("localhost:%s", tester.port),
+					grpc.WithInsecure(),
+					grpcutil.WithInsecureBearerToken(key),
+				)
+			}
```
jzelinskie, line 55:
```go
opts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
if key != "" {
    opts = append(opts, grpcutil.WithInsecureBearerToken(key))
}
conn, err := grpc.Dial(fmt.Sprintf("localhost:%s", tester.port), opts...)
```

jzelinskie (changes requested):

pkg/cmd/server/server.go
josephschorr: Not if no keys were given; remember: preshared keys are ignored if a custom auth function is specified

cmd/spicedb/serve_integration_test.go
josephschorr: Copy paste

.github/workflows/build-test.yaml
```diff
@@ -50,7 +50,10 @@ jobs:
       - uses: "actions/setup-go@v3"
         with:
           go-version: "^1.17"
-      - uses: "authzed/actions/docker-build@main"
+      - uses: "docker/build-push-action@v1"
```
jzelinskie, line 53: probably need to rename this job

`image-build` => `docker-tests`
`Build Container Image` => `Build & Test Container Image`

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func TestServe(t *testing.T) {
+	requireParent := require.New(t)
+
+	tester, err := newTester(t,
+		&dockertest.RunOptions{
+			Repository:   "authzed/spicedb",
+			Tag:          "currentpr",
```
jzelinskie, line 28: dev

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+	defer tester.cleanup()
+
+	for key, expectedWorks := range map[string]bool{
+		"":           false,
+		"firstkey":   true,
+		"secondkey":  true,
+		"anotherkey": false,
+	} {
+		t.Run(key, func(t *testing.T) {
+			require := require.New(t)
```
jzelinskie, line 44: Not required, but fwiw I prefer to just use require directly and not shadow the import.

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+			opts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
+			if key != "" {
+				opts = append(opts, grpcutil.WithInsecureBearerToken(key))
+			}
+			conn, err := grpc.Dial(fmt.Sprintf("localhost:%s", tester.port), opts...)
+
```
jzelinskie, line 51: why the new line here?

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+			require.NoError(err)
+			defer conn.Close()
+
+			resp, err := healthpb.NewHealthClient(conn).Check(context.Background(), &healthpb.HealthCheckRequest{Service: "authzed.api.v1alpha1.SchemaService"})
```
jzelinskie, line 55: This is against v1alpha1, but then you use v1

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+				}
+				`,
+			})
+
+			if expectedWorks {
+				require.NoError(err)
+			} else {
+				s, ok := status.FromError(err)
+				require.True(ok)
+				require.Equal(codes.Unknown, s.Code())
```
jzelinskie, line 78: Shouldn't this be `codes.Unauthenticated`?

cmd/spicedb/servetesting_integration_test.go
```diff
@@ -29,7 +29,7 @@ func TestTestServer(t *testing.T) {
 	tester, err := newTester(t,
 		&dockertest.RunOptions{
 			Repository:   "authzed/spicedb",
-			Tag:          "latest",
+			Tag:          "currentpr",
```
jzelinskie, line 32: dev

cmd/spicedb/serve_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+			defer conn.Close()
+
+			resp, err := healthpb.NewHealthClient(conn).Check(context.Background(), &healthpb.HealthCheckRequest{Service: "authzed.api.v1alpha1.SchemaService"})
+			require.NoError(err)
+			require.Equal(healthpb.HealthCheckResponse_SERVING, resp.GetStatus())
+
+			client := v1.NewSchemaServiceClient(conn)
+			_, err = client.WriteSchema(context.Background(), &v1.WriteSchemaRequest{
+				Schema: `
+				definition user {}
```
jzelinskie, line 62: Why not use a simpler schema? It's not really relevant to the test

pkg/cmd/server/server.go
jzelinskie: Oh, gotcha.

I wonder if it makes more sense to hoist this up to where all of the other preshared key logic is.

jzelinskie (changes requested):

.github/workflows/build-test.yaml
```diff
@@ -50,7 +50,10 @@ jobs:
...
+        with:
+          push: false
+          tags: "currentpr"
```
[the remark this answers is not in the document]
josephschorr: `dev` has a risk of overloading with a, well, development image. I wanted to be explicitly different from all other contexts

.github/workflows/build-test.yaml
josephschorr: If we do, it'll break the CI submit checks, so I left it as is (unless we should change those too)

cmd/spicedb/serve_integration_test.go
josephschorr: I do too, but I was keeping to the style in the existing integration tests

cmd/spicedb/serve_integration_test.go
josephschorr: Currently, we don't return that code if the key does not match

cmd/spicedb/serve_integration_test.go
josephschorr: Its mostly to make sure the service is running at all, but I can change it

cmd/spicedb/serve_integration_test.go
josephschorr: Copy-pasted it; I can make it simpler, but it wasn't a big deal to change either way

.github/workflows/build-test.yaml
jzelinskie: It's really easy to change them. I'd rather spend two seconds in the GitHub settings than live with names and descriptions of things that don't make sense.

cmd/spicedb/serve_integration_test.go
jzelinskie: that seems... like a bug

jzelinskie (approved):

PR #525, opened 2022-04-12
introduces mysql datastore

vroldanbet:
Depends on https://github.com/authzed/spicedb/pull/519

in its current state the MySQL datastore duplicates a
significant portion of the PSQL datastore's logic. Given
the fast pace of breaking changes in the internal implementation,
we considered code should be upstreamed ASAP and start addressing
duplication and gaps as more-approachable and smaller follow ups.

The fundamental differences with the PSQL implementation are:
- usage of Go's databases/sql API
- queries need to be generated on Datastore initialization
  in order to handle customizable prefix functionality
- transaction seed in runtime: the implementation tries
  hard to avoid data migrations. Instead of handling the
  seeding of "transaction zero" as part of "spicedb migrate",
  we introduced it in IsReady(). This will be revisited in a follow-up,
  to explore moving it to the datastore struct initialization.
  - SQL query changes: some queries were tweaked to make it compatible
  with Vitess
- IsReady(): introduces changes to avoid downtime during migrations.
  The implementation relies on a new function "IsHeadCompatible()", which
  has a rather simplistic heuristic: compatible if is HEAD or the version
  it HEAD replaced. This is probably the most controversial part of the change
  given it introduces a new method as part of the API.
- Stats cluster ID is not implemented - we didn't see an obvious need to introduce
  the complexity needed to handle it, but if controversial can be handled in a follow up
- only supports a subset of the CLI options. Those will be handled as part of a follow up
- does not emit metrics for garbage-collection - also to be handled in a follow up
- exposes connection pool metrics via dlmiddlecote/sqlstats

This commit does not wire the datastore into the spicedb CLI.

Issue #482, opened 2022-03-16
fail to start spice serve command if Datastore isn't ready


vroldanbet:
# Ask

`spicedb serve` should terminate with code non-zero if `Datastore.IsReady()` is false on bootstrap

# Context

During the development of a `Datastore` implementation based on MySQL we identified missing seeding that would prevent the `Datastore.IsReady()` from returning `true`, signaling that the `Datastore` is ready to serve operations.

Our surprise was that the deployment of SpiceDB, which runs `spicedb serve` succeeded, leaving the cluster "Healthy" from a kubernetes perspective, but otherwise completely unable to serve API calls because `Datastore.IsReady()` would prevent it. In a production scenario this would have caused an incident. From the perspective of the rollout of a new Kube `Deployment`, we would have expected it to never become healthy.

The proposal is to make sure `spicedb serve` terminate during the bootstrap sequence if `Datastore.IsReady()` is false.

jzelinskie:
I think there could be a bit more discussion about how to fix this.

Another possibility is fleshing out the gRPC health endpoint to make this call so that Pods configured with proper healthiness/readiness probes won't route traffic to these instances until it's all green.

Maybe we could add a boolean `--datastore-exit-if-unready` so that simple deployments don't need scripts around calling a readiness API.

vroldanbet:
@jzelinskie from a conceptual perspective of an `IsReady()` function, I'm OK with that 👍🏻  although I'd be wary of recommending using that for liveness probes (see https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html). Fortunately something folks are not forced to adopt it they don't need to.

The challenge I see is that the way `Datastore.IsReady()` is currently implemented in practice makes it unlikely a deployment will become healthy within a reasonable amount of time: they check whether the database is running at the HEAD migration. From my perspective that's akin to an **incorrect system configuration** (e.g. if you pass an incorrect flag value to `spicedb serve`, you return with exit code non-zero, right?), hence my suggestion to terminate the process. I'm not sure I can see a spicedb not having a migration run being in the realm of "a transient state the system will eventually recover from" 😕. 

Perhaps the problem here is that `IsReady()` is not a good fit to check for HEAD migrations?

PR #539, opened 2022-04-19
moves seeding to the initialization of the datastore

vroldanbet:
Addresses one of the follow ups identified in https://github.com/authzed/spicedb/pull/525

This moves the seeding responsibility out of `IsReady` and makes it side-effect free. It's conceptually unexpected that `IsReady` can have side effects, and beyond that, having the mutation as part of the datastore initialization gives us some nice properties:

- the application fails to start if seeding didn't succeed. Lack of seeding is not something the application can recover from,
and we wouldn't want to have a SpiceDB pod running as healthy when the datastore isn't seeded. If seeding failed due to a transient, the pod will get another chance as its restarted by kubernetes when non-zero error codes are returned.
- `IsReady` is not invoked right after SpiceDB starts, so SpiceDB can start even though the datastore is not ready, leading to a situation where the service is healthy but unseeded until IsReady is invoked. This was reported in authzed/spicedb#482 but there hasn't been traction on the proposal, so by moving seeding to datastore initialization we can make sure the application will fail to start.


jakedt:
What is the plan for incremental seeding? For example, if a datastore were already seeded with the zeroth transaction and a new migration, e.g. the one proposed in #546 needed to do more "initial" DML statements, how would these get picked up? Can we add separate DML statements to migrations which get aggregated into a "seeded-through-migration" value which parallels the DDL statements? Maybe the migration executor could have both a `migrate` and a `seed` function?

vroldanbet:
>What is the plan for incremental seeding? For example, if a datastore were already seeded with the zeroth transaction and a new migration, e.g. the one proposed in https://github.com/authzed/spicedb/pull/546 needed to do more "initial" DML statements, how would these get picked up?

Hey @jakedt, that's a great question. TL;DR I don't _think_ it's needed. Read further for _the nitty gritty details_:

The way I think about the seeding is that is closely related to / needed by changes in the codebase, so deploying a new version of the codebase is warranted anyway, and so I haven't seen the need to run the seeding at any other moment during of the lifecycle of the spicedb processes. So in short, I don't _think_ it's needed. Could you think of scenarios where it is? Doing this in initialization is somewhat the same because it makes sure that the application does not successfully startup unless the required migration and seeding is in place. This is also related to https://github.com/authzed/spicedb/issues/482

A more fundamental question https://github.com/authzed/spicedb/pull/546 opens up, and that I keep going back to, is the _chicken and egg problem_ (we discussed about this recently in Discord). As far as I can tell, we can identify at least 2 classes of migrations:
- one that require codebase changes to make sure the application is able to work with both "before and after" migration states (e.g. dropping column, changes to column type...)
- migrations that are by nature backward compatible (e.g. adding an index, adding new table) and does not require the application to be "ready for it".

https://github.com/authzed/spicedb/pull/546 falls under the latter - it's adding a new table, and the currently running SpiceDB would remain unaffected by it. Then you deploy the application, and seeding happens. Great! 🎉 . Unfortunately due to how `Datastore.IsReady()` is implemented (it checks the system is running at `HEAD`), running such migration would invariably lead to `IsReady() == false`, because suddenly the codebase finds its running with a migration version _it does not know about_ (a.k.a future migrations). I'll admit I don't have a clear picture yet of _when_ is `IsReady` invoked. As far as I could tell from us ending with migrations that led it to be not ready, the application was still responding to `Check` requests.

I addressed the reverse scenario with `migrate.Manager.IsHeadCompatible()` which helps the datastore run at head, or head-1. That works well for "deploy, then migrate". However, that still does not solve the problem where you _need_ to run the migration first because the code _needs it_. You could change the codebase to handle both "with and without it the required changes", but it may be tricky / inconvenient to handle in the codebase 😢 So I think we could revisit the checks in `IsReady` to not report `not ready` if using a future migration:
- one option is to define an environment variable where customers can configure "migrations soon to be run". You redeploy the application, tell spicedb "there is a new future migration version coming, don't worry about it". This a low effort change (famous last words) and could be encapsulated in `migrate.Manager.IsHeadCompatible()`.
- another option is to relax `IsReady` and say "future migrations are expected". That would be fine if it not were by the fact that then we loose the information of "which migrations have been run". We could change the implementation of the versions table to keep such a list, so when a new "future migration" comes, the system says "it's fine as long as migration X was run".

>Can we add separate DML statements to migrations which get aggregated into a "seeded-through-migration" value which parallels the DDL statements? Maybe the migration executor could have both a migrate and a seed function?

I'm not sure I fully understood what you mentioned there. I thought about separating schema-migrations and data-migrations in `spicedb migrate` before landing onto this design. `spicedb migrate` could have subcommands to separate DML and DDL. That would certainly remove the need for these seeding functions on startup, at the expense of more operational burden. My preference was the system reconciling its own state, because regardless _there is still the need to deploy a version that uses the changes introduced by migration anyway_. But I don't discard scenarios where the seeding may not be trivial or practical on application bootstrap _at all_ (e.g. normalize `relation_tuple` table to avoid duplicating namespace string representation that would lead to a _large_ data migration) and that would warrant an offline data-migration process to be run.

jakedt (approved):

PR #540, opened 2022-04-19
Have the Docker-image based test suite run solely those tests

josephschorr:


.github/workflows/build-test.yaml
```diff
@@ -52,7 +52,8 @@ jobs:
           go-version: "^1.17"
       - uses: "authzed/actions/docker-build@main"
       - uses: "authzed/actions/go-test@main"
-        with:
+        working-directory: "cmd/spicedb"
```
jzelinskie, line 55: Unfortunately, `working-directory` doesn't work with `use`.
All of our actions do support the following though:
```yaml
with:
  working_directory: "path"
```

jzelinskie (approved):

Issue #538, opened 2022-04-19
spicedb CLI returns error code 0 regardless of application errors


vroldanbet:
# What

SpiceDB `spicedb` CLI should return non-zero error codes when a failure happens

# Why

The application should signal why did the process terminate through error codes. This particularly important for container schedulers like kubernetes that make use of the error code to [determine pod lifecycle state](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) (e.g. pods that terminate with non-zero error code are considered in `Failed` state). If a pod terminates with a zero error code (state `Succeeded`) no restarts will take place.

vroldanbet:
@jzelinskie yeah I had already tested the same snippet locally and seemed to do what I needed:
- SIGTERM'ing the process returned error code zero
- Any error returned in the execution of the command would return non-zero

A quick look into the cobra repository (https://github.com/spf13/cobra/issues/914) invites to think it should be sufficient to do what you posted above, with some friction around when to show the command list.

PR #541, opened 2022-04-19
makes cli application return non-zero error code on errors

vroldanbet:
fixes https://github.com/authzed/spicedb/issues/538

changes the application to return non-zero error codes when an `error` is returned by the root command. 

I attempted to make the output of the CLI _"pretty"_:

- if the error is a parsing error, command usage will be printed
- if the error is an application error, the error is logged using `zerolog` so that it shows up nicely just like any other application log message

If we wanted the flag values to be recognized as a "parsing/cli error" (e.g. unrecognized datastore) we could signal it with a specific error, but we would need to adjust every part of the application that initializes spicedb to return an error deemed to be considered "cli error". I considered that to be out of scope of this PR.

### example: unrecognized flag: error code 1, prints command usage
 
Screen Shot 2022-04-19 at 19 39 22

### example: incorrect flag value: error code 1, does not print command usage

Screen Shot 2022-04-19 at 19 25 20

### example: SIGTERM: error code 0

Screen Shot 2022-04-19 at 19 29 03

### DIFF
--- a/cmd/spicedb/main.go
+++ b/cmd/spicedb/main.go
@@ -1,11 +1,15 @@
 package main
 
 import (
+	"errors"
 	"math/rand"
+	"os"
 	"time"
 
 	"github.com/cespare/xxhash"
+	"github.com/rs/zerolog/log"
 	"github.com/sercand/kuberesolver/v3"
+	"github.com/spf13/cobra"
 	"google.golang.org/grpc/balancer"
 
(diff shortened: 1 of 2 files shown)


jzelinskie (approved):
I think this could also be done with a new error type that wraps the flag error then using `errors.As` at the `Execute()` callsite.

I don't think it really makes much of a difference, though.
LGTM

PR #544, opened 2022-04-19
Reduce the namespace manager interface to the essentials

josephschorr:
All other methods are moved into utility methods, as their implementations are shared


josephschorr:
Closed in favor #581

PR #547, opened 2022-04-20
Ensure preconditions on tuple writes in Postgres and MySQL honor deletions

### LINKED ISSUE #339: Preconditions on the postgres datastore do not consider relationship liveliness
An example log line from checking preconditions that does not take into account the relationship's `deleted_transaction` status:

```{"level":"info","module":"pgx","args":["user","user_id","project"],"pid":45594,"rowCount":1,"sql":"SELECT id FROM relation_tuple WHERE namespace = $1 AND object_id = $2 AND userset_namespace = $3 LIMIT 1","time":1.402475,"time":"2021-12-14T16:16:40Z","message":"Query"}```

### PR DESCRIPTION
**Pull request by josephschorr**
Fixes #339

### DIFF
--- a/internal/datastore/mysql/tuple.go
+++ b/internal/datastore/mysql/tuple.go
@@ -17,7 +17,7 @@ import (
 )
 
 // TODO (@vroldanbet) dupe from postgres datastore - need to refactor
-func (mds *Datastore) selectQueryForFilter(filter *v1.RelationshipFilter) sq.SelectBuilder {
+func (mds *Datastore) selectQueryForFilter(filter *v1.RelationshipFilter, revision datastore.Revision) sq.SelectBuilder {
 	query := mds.QueryTupleExistsQuery.Where(sq.Eq{colNamespace: filter.ResourceType})
 
 	if filter.OptionalResourceId != "" {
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

**Review by vroldanbet — DISMISSED** [review]
Changes to mysql datastore LGTM

File: internal/datastore/postgres/tuple.go [hunk 1]
```diff
@@ -56,6 +56,8 @@ func selectQueryForFilter(filter *v1.RelationshipFilter) sq.SelectBuilder {
 		}
 	}
 
+	query = query.Where(sq.Eq{colDeletedTxn: liveDeletedTxnID})
```
jakedt, line 59: The caller of this method should use `filterToLivingObjects`.

internal/datastore/mysql/tuple.go
```diff
@@ -37,6 +37,8 @@ func (mds *Datastore) selectQueryForFilter(filter *v1.RelationshipFilter) sq.Sel
 		}
 	}
 
+	query = query.Where(sq.Eq{colDeletedTxn: liveDeletedTxnID})
```
jakedt, line 40: The caller of this method should use `filterToLivingObjects`.

jakedt (changes requested):

josephschorr:
Closed in favor of #581

PR #550, opened 2022-04-21
sets mysql manager singleton

vroldanbet:
Follow up to https://github.com/authzed/spicedb/pull/532

# What

fixes bug where the MySQL migration manager singleton was not selected when using `spicedb migrate` for the MySQL datastore

# Notes

`spicedb migrate` command has not tests whatsoever and I wonder if its worth adding some very basic smoke tests?


josephschorr (approved):
LGTM

PR #551, opened 2022-04-21
Add integration testing for the migrate command

josephschorr:
Runs each datastore's backing engine or service in a container, and then creates another container with SpiceDB to perform the migration


internal/testserver/datastore/datastore.go
```diff
@@ -3,22 +3,38 @@ package datastore
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/authzed/spicedb/internal/datastore"
 )
 
 // InitFunc initialize a datastore instance from a uri that has been
 // generated from a TestDatastoreBuilder
```
ecordell, line 12:
```suggestion
// InitFunc initializes a datastore instance from a uri that has been
// generated from a RunningEngineForTest
```

internal/testserver/datastore/datastore.go
```diff
@@ -3,22 +3,38 @@ package datastore
...
+// requires env vars set to use from an external source (like a container).
+type RunningEngineForTestWithEnvVars interface {
+	RunningEngineForTest
+
+	// ExternalEnvVars are the environment variables to be set.
+	ExternalEnvVars() []string
+}
+
 // TestDatastoreBuilderFunc allows a function to implement TestDatastoreBuilder
 type TestDatastoreBuilderFunc func(t testing.TB, initFunc InitFunc) datastore.Datastore
```
ecordell, line 39: I would rename these too to match

internal/datastore/mysql/datastore.go
```diff
@@ -64,6 +66,10 @@ var (
 	sb = sq.StatementBuilder.PlaceholderFormat(sq.Question)
 )
 
+func init() {
+	datastore.Engines = append(datastore.Engines, Engine)
```
ecordell, line 70: 👍

internal/testserver/datastore/datastore.go
josephschorr: Removed

ecordell (approved):
LGTM

PR #555, opened 2022-04-22
Postgres optimized revision caching

jakedt:


internal/datastore/common/revisions/optimized.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// this value is read and set by multiple consumers, it's protected
+	// by atomic load/store
+	lastQuantizedRevision *atomicRevision
+
+	// the updategroup consolidates concurrent requests to the database into 1
+	updateGroup singleflight.Group
+}
+
+type validRevision struct {
```
ecordell, line 87: 👍 this is a small change but I like it

internal/datastore/postgres/postgres.go
```diff
@@ -188,7 +189,13 @@ func NewPostgresDatastore(
 		config.gcWindow.Seconds(),
 	)
 
+	maxRevisionStaleness := time.Duration(float64(config.revisionQuantization.Nanoseconds())*
+		config.maxRevisionStalenessPercent) * time.Nanosecond
```
ecordell, line 193: isn't duration already in nanoseconds?

internal/datastore/postgres/revisions.go
```diff
@@ -23,7 +24,8 @@ const (
 	SELECT COALESCE(
 		(SELECT MIN(%[1]s) FROM %[2]s WHERE %[3]s >= TO_TIMESTAMP(FLOOR(EXTRACT(EPOCH FROM NOW() AT TIME ZONE 'utc') * 1000000000 / %[4]d) * %[4]d / 1000000000)),
 		(SELECT MAX(%[1]s) FROM %[2]s)
-	);`
+	),
+	%[4]d - CAST(EXTRACT(EPOCH FROM NOW() AT TIME ZONE 'utc') * 1000000000 as bigint) %% %[4]d;`
```
ecordell, line 28: update the comment to note that it's also returning validForNanos now?

internal/datastore/postgres/postgres.go
jakedt: Yes, this just makes it explicit. They must want you to do this otherwise `time.Nanosecond` wouldn't exist!

jakedt:
Updated

ecordell (approved):
LGTM

ecordell (changes requested):
unlgtm, the e2e failure looks potentially real

ecordell (dismissed):
LGTM

internal/datastore/common/revisions/optimized.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+var tracer = otel.Tracer("spicedb/internal/datastore/common/revisions")
+
+// OptimizedRevisionFunction instructs the datastore to compute its own current
+// optimized revision given the specific quantization, and return for how long
+// it will remain valid.
+type OptimizedRevisionFunction func(context.Context) (rev datastore.Revision, validFor time.Duration, err error)
+
+// NewRemoteClockRevisions returns a RemoteClockRevisions for the given configuration
```
josephschorr, line 25: Fix comment

internal/datastore/common/revisions/optimized.go
```diff
@@ -0,0 +1,103 @@
...
+	lastRevision := cor.lastQuantizedRevision.get()
+	if localNow.Before(lastRevision.validThrough) {
+		log.Debug().Time("now", localNow).Time("valid", lastRevision.validThrough).Msg("returning cached revision")
+		return lastRevision.revision, nil
+	}
+
+	lastQuantizedRevision, err, _ := cor.updateGroup.Do("", func() (interface{}, error) {
+		log.Debug().Time("now", localNow).Time("valid", lastRevision.validThrough).Msg("computing new revision")
+
+		optimized, validFor, err := cor.optimizedFunc(ctx)
```
josephschorr, line 56: Add a panic here if it is not set? its slightly nicer than just an NPE

internal/datastore/common/revisions/optimized.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to compute optimized revision: %w", err)
+		}
+
+		rvt := localNow.
+			Add(validFor).
+			Add(cor.maxRevisionStaleness)
```
josephschorr, line 63: Won't adding these push it above the max staleness?

internal/datastore/common/revisions/optimized.go
```diff
@@ -0,0 +1,103 @@
...
+		log.Debug().Time("now", localNow).Time("valid", rvt).Stringer("validFor", validFor).Msg("setting valid through")
+
+		return optimized, nil
+	})
+
+	return lastQuantizedRevision.(decimal.Decimal), err
+}
+
+// CachedOptimizedRevisions does caching and deduplication for requests for optimized revisions.
+type CachedOptimizedRevisions struct {
```
josephschorr, line 74: Move above the function defs for this type?

internal/datastore/common/revisions/optimized.go
jakedt: At the cost of requiring an extra conditional every time.

internal/datastore/common/revisions/optimized.go
jakedt: My definition for `staleness` is "time after which the new revision would have been selected". So in this case we allow for it to be used slightly longer than it's strictly valid for, but which gives us a minimum valid lifetime for any revision we ask the database to compute. E.g you can't get back a revision that's only good for 1 nanosecond.

internal/datastore/common/revisions/optimized.go
jakedt: I don't know. All of the fields are private, it's probably less important than the API to interact with it.

internal/datastore/common/revisions/optimized.go
josephschorr: Yes, but I tend to think having panics when its an *expected* state makes sense; it would mean they forgot to call the assignment method, and that's certainly an expected bug

josephschorr: I wish we had debug-only checks

internal/datastore/common/revisions/optimized.go
josephschorr: But why not just do `max(validFor, maxRevisionStaleness)` then?

internal/datastore/common/revisions/optimized.go
ecordell: I've seen this done with build tags in https://github.com/dlsniper/debugger and in some etcd fault injection library that I can't find now. I assume the compiler removes no-op functions but I haven't checked.

ecordell (approved):
LGTM

PR #557, opened 2022-04-22
update to a version of rudd that doesn't race

ecordell:
see: https://github.com/authzed/spicedb/runs/6131744750?check_suite_focus=true
and: https://github.com/dalzilio/rudd/pull/1


jzelinskie:
Can we fork these into our own GitHub namespace?

josephschorr (approved):
LGTM

PR #563, opened 2022-04-25
Add a custom analyzers package for custom lint checks

josephschorr:


tools/analyzers/nilvaluecheck/nilvaluecheck.go
```diff
@@ -0,0 +1,130 @@
...
+			inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
+
+			nodeFilter := []ast.Node{
+				(*ast.ReturnStmt)(nil),
+				(*ast.DeclStmt)(nil),
+			}
+
+			typePaths := strings.Split(*disalllowedPaths, ",")
+			hasTypePath := func(path string) bool {
+				for _, typePath := range typePaths {
```
jzelinskie, line 41: This looks like it's also stringz.SliceContains

tools/analyzers/nilvaluecheck/nilvaluecheck.go
```diff
@@ -0,0 +1,130 @@
...
+						if foundType == nil {
+							return false
+						}
+
+						resultType := foundType.String()
+						if hasTypePath(resultType) {
+							// Check for a nil result.
+							returnedValue := s.Results[index]
+							if ident, ok := returnedValue.(*ast.Ident); ok {
+								if ident.Name == "nil" {
```
jzelinskie, line 91:
```go
if ident, ok := returnedValue.(*ast.Ident); ok && ident.Name == "nil" {
```

tools/analyzers/nilvaluecheck/nilvaluecheck.go
```diff
@@ -0,0 +1,130 @@
...
+				}
+				return false
+			})
+
+			return nil, nil
+		},
+		Requires: []*analysis.Analyzer{inspect.Analyzer},
+		Flags:    *flagSet,
+	}
+}
```
jzelinskie, line 130: Can we add tests for this?

tools/analyzers/nilvaluecheck/nilvaluecheck.go
```diff
@@ -0,0 +1,130 @@
...
+							return false
+						}
+
+						foundType := pass.TypesInfo.TypeOf(valueSpec.Type)
+						if foundType == nil {
+							return false
+						}
+
+						varType := foundType.String()
+						if len(valueSpec.Values) == 0 && hasTypePath(varType) {
```
jzelinskie, line 112: looks like this catches `var thing *myType`, but not `var thing *mytype = nil` or `var thing *mytype = otherNilVariable`

jzelinskie (changes requested):

tools/analyzers/nilvaluecheck/nilvaluecheck.go
josephschorr: Correct. `var thing *mytype = nil` should be caught by existing lint and to do the `otherNilVariable` would require real flow analysis

tools/analyzers/nilvaluecheck/nilvaluecheck.go
josephschorr: Yeah, I can add some unit tests based on local parsing

jzelinskie (approved):

PR #564, opened 2022-04-26
Add a check on startup for the last released version of SpiceDB

josephschorr:
Can be disabled with the `skip-release-check` flag/env var


pkg/releases/releases.go
```diff
@@ -0,0 +1,79 @@
...
+	// ViewURL is the URL at which the release can be viewed.
+	ViewURL string
+}
+
+// GetLatestRelease returns the latest release of SpiceDB, as reported by the GitHub API.
+func GetLatestRelease(ctx context.Context) (*Release, error) {
+	client := github.NewClient(nil)
+	var current *Release
+
+	// Load the releases from GitHub in a paginated fashion (to not overwhelm the API).
```
vroldanbet, line 37: Did you consider using the API specific to get the latest release instead of iterating over all releases?

https://docs.github.com/en/rest/releases/releases#get-the-latest-release

josephschorr: My concern there was if we published a "fix" release; say we're at 1.6.0 and we publish 1.5.7 as a fix afterwards for some reason. Will GitHub now report 1.5.7 as the "latest" release?

pkg/releases/releases_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReleases(t *testing.T) {
+	release, err := GetLatestRelease(context.Background())
```
vroldanbet, line 14: I think this test is the _least interesting bit to test_ but also puts it at the mercy of GitHub's availability and thus flakes. Plus is not being a great citizen by having all those CI builds hit GitHub.

I think it would the interesting bits to test IMHO are in `CheckAndLogRunE`, since it's the core logic of this functionality. You could mock `GetLatestRelease` and test the various scenarios on how the spicedb warns the operator about updates.

vroldanbet: overlooked that bit! thank you!

josephschorr: Updated

vroldanbet (approved):
This looks great, thanks for addressing my comments! 🙇🏻

PR #568, opened 2022-04-27
pkg/cmd: catch nil registry initialization

### LINKED ISSUE #567: Bad telemetry initialization shouldn't panic
https://github.com/authzed/spicedb/blob/02e2d53d4424050a65a68afa0be419f0570774b3/pkg/cmd/server/server.go#L287-L290

If this fails, `registry` is nil but we don't exit. Telemetry will later panic attempting to `Gather` from a nil registry unless telemetry is explicitly disabled via the CLI:

```
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x61be34]

    goroutine 219 [running]:
    github.com/prometheus/client_golang/prometheus.(*Registry).Gather(0x0)
    	/go/pkg/mod/github.com/prometheus/client_golang@v1.12.1/prometheus/registry.go:419 +0xc4
    github.com/authzed/spicedb/internal/telemetry.discoverTimeseries(0x40004f8600?)
    	/go/src/spicedb/internal/telemetry/reporter.go:89 +0x28
```

jzelinskie:
Fixes #567

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -289,15 +289,10 @@ func (c *Config) Complete() (RunnableServer, error) {
 		log.Warn().Err(err).Msg("unable to initialize telemetry collector")
 	}
 
-	metricsServer, err := c.MetricsAPI.Complete(zerolog.InfoLevel, MetricsHandler(registry))
-	if err != nil {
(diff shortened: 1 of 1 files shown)


ecordell (approved):
LGTM

PR #569, opened 2022-04-28
Fix `datastore-engine` help text

bryanhuhta:
The `datastore-engine` flag doesn't include MySQL as an option, even though it is legal to select. This PR adds it to the example list.

Also, instead of hard-coding all the possible datastores, I sorted the keys in `BuilderForEngine` and displayed them. This way, any new datastore will automatically get added to the options list and there won't be any accidental drift.

Before:

```
--datastore-engine string    type of datastore to initialize ("memory", "postgres", "cockroachdb", "spanner") (default "memory")
```

After:

```
--datastore-engine string    type of datastore to initialize ("cockroachdb", "memory", "mysql", "postgres", "spanner") (default "memory")
```

### DIFF
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


josephschorr (approved):
LGTM

PR #571, opened 2022-04-28
Switch to using Engines for the engine parameter

josephschorr:


pkg/cmd/datastore/datastore.go
```diff
@@ -89,10 +89,7 @@ type Config struct {
 
 // RegisterDatastoreFlags adds datastore flags to a cobra command
 func RegisterDatastoreFlags(cmd *cobra.Command, opts *Config) {
-	engines := []string{}
-	for engine := range BuilderForEngine {
-		engines = append(engines, fmt.Sprintf("%q", engine))
-	}
+	engines := append([]string{}, datastore.Engines...)
 	sort.Strings(engines) // Sort for stability.
```
ecordell, line 93:
```suggestion
	sort.Strings(datastore.Engines) // Sort for stability.
```

I don't think we really care if the original list gets sorted?

ecordell (commented):
I think we also want this for the `migrate` and `head` commands?

pkg/cmd/datastore/datastore.go
josephschorr: I'd rather not change it, as its defined to be "static", for lack of a better term.

josephschorr:
@ecordell Updated

internal/datastore/datastore.go
```diff
@@ -13,6 +14,13 @@ import (
 
 var Engines = []string{}
 
+// SortedEngineIDs returns the full set of engine IDs, sorted.
+func SortedEngineIDs() []string {
+	engines := append([]string{}, Engines...)
```
bryanhuhta, line 19: This will drop the quotes around each engine. Previously it was

> "memory", "postgres" ...

But now it will be

> memory, postgres ...

Nothing blocking, but it will look different from other flags that list valid options.

josephschorr: Fixed

ecordell (approved):
LGTM

PR #572, opened 2022-04-28
Add middleware to return the server version when requested, unless disabled

josephschorr:


pkg/cmd/serve.go
```diff
@@ -67,6 +67,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) {
 
 	// Flags for configuring API behavior
 	cmd.Flags().BoolVar(&config.DisableV1SchemaAPI, "disable-v1-schema-api", false, "disables the V1 schema API")
+	cmd.Flags().BoolVar(&config.DisableVersionResponse, "disable-version-response", false, "disables version response support in the API")
 
 	// Flags for misc services
 	util.RegisterHTTPServerFlags(cmd.Flags(), &config.DashboardAPI, "dashboard", "dashboard", ":8080", true)
```
prashant3286, line 73: why are we using two `dashboard` key inside a flag

josephschorr: One is the flag prefix, the other is the service name

jzelinskie (dismissed):

PR #574, opened 2022-04-29
Add spanner emulator env var detection

samkim:
Detect prefixed spanner emulator env var config and set the actual env var.

### DIFF
--- a/internal/datastore/spanner/migrations/driver.go
+++ b/internal/datastore/spanner/migrations/driver.go
@@ -7,6 +7,7 @@ import (
 	"cloud.google.com/go/spanner"
 	admin "cloud.google.com/go/spanner/admin/database/apiv1"
 	"github.com/rs/zerolog/log"
+	"github.com/spf13/viper"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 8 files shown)


internal/datastore/spanner/migrations/driver.go
```diff
@@ -84,3 +88,15 @@ func (smd SpannerMigrationDriver) Close() error {
 }
 
 var _ migrate.Driver = SpannerMigrationDriver{}
+
+func detectEmulatorSetting() (string, bool) {
+	for _, env := range os.Environ() {
```
ecordell, line 93: viper knows the env prefix we use - we pass it around as `programName` and configure viper with it:

https://github.com/authzed/spicedb/blob/7bd2af59981f91152009461135b5bb6fd7f304fd/pkg/cmd/server/defaults.go#L53

I'm not overly familiar with viper but it seems like there should be a way to directly ask for this value instead of manually flipping through the environment variables.

Does something like `viper.Get("SPANNER_EMULATOR_HOST")` work? Maybe @jzelinskie has some suggestions?

internal/datastore/spanner/spanner.go
```diff
@@ -58,6 +59,11 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore,
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
 
+	setting, found := detectEmulatorSetting()
```
ecordell, line 62: couldn't this be another `Option` on the spanner datastore (so that we don't have to read env vars here), and then re-use the implementation from the migrate command for the serve command?

ecordell (dismissed):
LGTM

even though this adds a new flag that's only used for tests, I like that it's more discoverable than undocumented env vars

josephschorr (approved):

PR #581, opened 2022-05-02
Datastore v2

### LINKED ISSUE #118: Datastore interface should have more fine grained control over transactions
Without exposing a better API that can enable calls to span across the same transaction, API handlers cannot make perfect guarantees about the consistency of the data they are saving.

A strawman API would look something like this, using a closure to get any data in and out:

```go
// within an API handler
err := datastore.Transact(func(ctx context.Context, tx datastore.Tx) error {
    query := tx.QueryTuples(...)
    // do stuff with query
    tx.DeleteRelationships(...)
})
```

jakedt:
Fixes #118


jakedt:
I will look into resolving any E2E test problems, but I don't think it should hold up the review.

jzelinskie (changes requested):

internal/datastore/crdb/README.md
```diff
@@ -0,0 +1,10 @@
+# CockroachDB Datastore
+
+CockroachDB is a Spanner-like datastore supporting global, immediate consistency, with the mantra "no stale reads."
+The CockroachDB implementation should be used when your SpiceDB service runs in multiple geographic regions, and Google's Cloud Spanner is unavailable (e.g. AWS, Azure, bare metal.)
+
+## Implementation Caveats
+
+In order to prevent the new-enemy problem, we need to make related transactions overlap.
+We do this by choosing a common database key and writing to that key with all relationships that may overlap.
+This tradeoff is cataloged in our blog post [The One Crucial Difference Between Spanner and CockroachDB](https://authzed.com/blog/prevent-newenemy-cockroachdb/).
```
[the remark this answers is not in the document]
jakedt: Will GitHub pick that up as a mini landing page for the DS when you navigate there?

internal/datastore/common/sql.go
```diff
@@ -207,41 +205,36 @@ func (tqs TupleQuerySplitter) SplitAndExecuteQuery(
...
-type TransactionPreparer func(ctx context.Context, tx pgx.Tx, revision datastore.Revision) error
+// TransactionCleanupFunc is a function that should be executed when the caller of
+// TransactionFactory is done with the transaction.
+type TransactionCleanupFunc func(context.Context) error
 
-func NewPGXExecutor(conn *pgxpool.Pool, prepTxn TransactionPreparer) ExecuteQueryFunc {
-	return func(ctx context.Context, revision datastore.Revision, sql string, args []interface{}) ([]*core.RelationTuple, error) {
+// TransactionFactory returns a transaction, cleanup function, and any errors that may have
+// occurred when building the transaction.
+type TransactionFactory func(context.Context) (pgx.Tx, TransactionCleanupFunc, error)
```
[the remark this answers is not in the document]
jakedt: Tx is good, we can't call it "rollback" because in the read-write transaction case, no cleanup is actually required an this is a no-op.

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +34,199 @@ func NewMemdbDatastore(
...
-			errors.New("gc window must be larger than quantization interval"),
-		)
+		return nil, errors.New("gc window must be larger than quantization interval")
 	}
 
-	db, err := memdb.NewMemDB(schema)
-	if err != nil {
-		return nil, fmt.Errorf(errUnableToInstantiateTuplestore, err)
+	if revisionQuantization <= 1 {
+		revisionQuantization = 1
```
jzelinskie, line 43: Having something like this in the datastore package removes decimal.Decimal from being directly in datastore implementations.

```go
func QuantizationFromDuration(d, gcWindow time.Duration) Revision {
	if d > gcWindow {
		return nil, fmt.Errorf(
			errUnableToInstantiate,
			fmt.Errorf("cluster gc window is less than requested gc window %d < %d",
				clusterTTLNanos,
				gcWindowNanos,
			),
		)
	}

	if d <= 1 {
		d = 1
	}

	return decimal.NewFromInt(d.Nanoseconds())
}
```

internal/datastore/crdb/crdb.go
```diff
@@ -159,11 +154,95 @@ type crdbDatastore struct {
...
+	createTxFunc := func(ctx context.Context) (pgx.Tx, common.TransactionCleanupFunc, error) {
+		tx, err := cds.pool.BeginTx(ctx, pgx.TxOptions{AccessMode: pgx.ReadOnly})
+		if err != nil {
+			return nil, nil, err
+		}
+
+		setTxTime := fmt.Sprintf(querySetTransactionTime, rev)
+		if _, err := tx.Exec(ctx, setTxTime); err != nil {
+			if err := tx.Rollback(ctx); err != nil {
+				log.Warn().Err(err).Msg("error rolling back transaction")
```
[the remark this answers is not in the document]
jakedt: It's not during creating, but I made it `error rolling back transaction`

internal/datastore/crdb/reader.go
```diff
@@ -0,0 +1,240 @@
...
+
+	"github.com/authzed/spicedb/internal/datastore/common"
+	"github.com/authzed/spicedb/internal/datastore/options"
+	"github.com/authzed/spicedb/pkg/datastore"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+const (
+	errUnableToReadConfig     = "unable to read namespace config: %w"
+	errUnableToListNamespaces = "unable to list namespaces: %w"
```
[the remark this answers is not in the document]
jakedt: Let's do deduplication later

internal/datastore/crdb/reader.go
```diff
@@ -0,0 +1,240 @@
...
+		qBuilder = qBuilder.FilterToResourceID(filter.OptionalResourceId)
+	}
+
+	if filter.OptionalRelation != "" {
+		qBuilder = qBuilder.FilterToRelation(filter.OptionalRelation)
+	}
+
+	if filter.OptionalSubjectFilter != nil {
+		qBuilder = qBuilder.FilterToSubjectFilter(filter.OptionalSubjectFilter)
+	}
```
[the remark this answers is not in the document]
jakedt: Let's hold off on deduplication

internal/datastore/memdb/memdb.go
jakedt: Let's hold off on simplification, the PR in question doesn't introduce decimal.Decimal to the implementations

internal/datastore/crdb/reader.go
```diff
@@ -0,0 +1,240 @@
...
+	keyer         overlapKeyer
+	overlapKeySet keySet
+	execute       executeTxRetryFunc
+}
+
+func (cr *crdbReader) ReadNamespace(
+	ctx context.Context,
+	nsName string,
+) (*core.NamespaceDefinition, datastore.Revision, error) {
+	ctx = datastore.SeparateContextWithTracing(ctx)
```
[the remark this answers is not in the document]
jakedt: It would be interesting to know when this is even required. We added this because any time we canceled the context upstream, it would cancel the context downstream and kill the PG connection. I'm not even sure if this is required for some of the datastores...

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +34,199 @@ func NewMemdbDatastore(
...
+
+				tx = mdb.db.Txn(true)
+				tx.TrackChanges()
+				mdb.activeWriteTxn = tx
+			})
+
+			return tx, err
+		}
+
+		newRevision := revisionFromTimestamp(time.Now().UTC())
```
jzelinskie, line 145: this is defined in 3 different datastores, probably should pull it out

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +34,199 @@ func NewMemdbDatastore(
...
-	if err := txn.Insert(tableTransaction, newChangelogEntry); err != nil {
-		return 0, err
+	return &memdbReader{noopTryLocker{}, txSrc, snapshotRevision, nil}
+}
+
+func (mdb *memdbDatastore) ReadWriteTx(
+	ctx context.Context,
+	f func(context.Context, datastore.ReadWriteTransaction) error,
+) (datastore.Revision, error) {
+	for i := 0; i < 10; i++ {
```
jzelinskie, line 123: should make 10 a constant

internal/datastore/crdb/README.md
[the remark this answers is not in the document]
jzelinskie: I guess the thought process is that if you're writing code, you're using an editor with godoc. Users will likely just be reading the "selecting a datastore" doc on docs.authzed.com

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +34,199 @@ func NewMemdbDatastore(
...
-		return 0, err
+	return &memdbReader{noopTryLocker{}, txSrc, snapshotRevision, nil}
+}
+
+func (mdb *memdbDatastore) ReadWriteTx(
+	ctx context.Context,
+	f datastore.TxUserFunc,
+) (datastore.Revision, error) {
+	for i := 0; i < 10; i++ {
+		var tx *memdb.Txn
```
jzelinskie, line 124: This variable seems unnecessary if it's available the same thing as `mdb.activeWriteTxn`. The locking semantics will be more clear if you always refer to this field instead of the local var.

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +34,199 @@ func NewMemdbDatastore(
...
+				revisionNanos: newRevision.IntPart(),
+				changes:       newChanges,
+			}
+			if err := tx.Insert(tableChangelog, change); err != nil {
+				return datastore.NoRevision, fmt.Errorf("error writing changelog: %w", err)
+			}
+
+			tx.Commit()
+		}
+		mdb.activeWriteTxn = nil
```
jzelinskie, line 206: newChanges can be defined in the conditional below.

The whole thing could be made a little more clear with a refactor:

```go
if mdb.activeWriteTxn != nil {
	insertChanges(tx, newRevision)
	mdb.activeWriteTxn.Commit()
}
mdb.activeWriteTxn = nil
```

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +34,199 @@ func NewMemdbDatastore(
...
+	mdb.RLock()
+	defer mdb.RUnlock()
+
+	return len(mdb.revisions) > 0, nil
 }
+
+func (mdb *memdbDatastore) Close() error {
+	mdb.Lock()
+	defer mdb.Unlock()
+	mdb.db = nil
```
jzelinskie, line 228: probably also want to drop the revisions here

internal/datastore/crdb/readwrite.go
```diff
@@ -0,0 +1,285 @@
...
+		case v1.RelationshipUpdate_OPERATION_DELETE:
+			rwt.relCountChange--
+			sql, args, err := queryDeleteTuples.Where(exactRelationshipClause(rel)).ToSql()
+			if err != nil {
+				return fmt.Errorf(errUnableToWriteRelationships, err)
+			}
+
+			if _, err := rwt.tx.Exec(ctx, sql, args...); err != nil {
+				return fmt.Errorf(errUnableToWriteRelationships, err)
+			}
```
jzelinskie, line 132: Why not try to perform these deletes all together rather than one at a time?

internal/datastore/memdb/readonly.go
```diff
@@ -0,0 +1,353 @@
...
+
+type memdbReader struct {
+	TryLocker
+	txSource txFactory
+	revision datastore.Revision
+	initErr  error
+}
+
+// QueryRelationships reads relationships starting from the resource side.
+func (rot *memdbReader) QueryRelationships(
```
jzelinskie, line 28: what is rot? i assumed read only transaction, but that's not accurate as this is a reader which has a tx factory and isn't an active transaction itself.

internal/datastore/memdb/readonly.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	var nsDefs []*core.NamespaceDefinition
+
+	it, err := tx.LowerBound(tableNamespace, indexID)
+	if err != nil {
+		return nil, err
+	}
+
+	for {
```
jzelinskie, line 208: big fan of the break in the for loop clauses

```go
	for foundRaw := it.Next(); foundRaw != nil; foundRaw = it.Next() {
		found := foundRaw.(*namespace)
		var loaded core.NamespaceDefinition
		if err := proto.Unmarshal(found.configBytes, &loaded); err != nil {
			return nil, err
		}
		nsDefs = append(nsDefs, &loaded)
	}
```

internal/datastore/memdb/readwrite.go
```diff
@@ -0,0 +1,210 @@
...
+		case v1.RelationshipUpdate_OPERATION_TOUCH:
+			if err := tx.Insert(tableRelationship, rel); err != nil {
+				return fmt.Errorf("error inserting relationship: %w", err)
+			}
+		case v1.RelationshipUpdate_OPERATION_DELETE:
+			if existing != nil {
+				if err := tx.Delete(tableRelationship, existing); err != nil {
+					return fmt.Errorf("error deleting relationship: %w", err)
+				}
+			}
```
jzelinskie, line 79: We don't fail if you try to delete something that doesn't exist? I guess that makes sense based on the SQL expressions for other datastores.

internal/datastore/mysql/datastore.go
```diff
@@ -39,14 +40,14 @@ const (
...
-	errUnableToInstantiate  = "unable to instantiate datastore: %w"
-	errUnableToQueryTuples  = "unable to query tuples: %w"
-	errUnableToWriteTuples  = "unable to write tuples: %w"
-	errUnableToDeleteTuples = "unable to delete tuples: %w"
-	liveDeletedTxnID        = uint64(9223372036854775807)
-	batchDeleteSize         = 1000
-	noLastInsertID          = 0
-	seedingTimeout          = 10 * time.Second
+	errUnableToInstantiate = "unable to instantiate datastore: %w"
+	liveDeletedTxnID       = uint64(9223372036854775807)
```
jzelinskie, line 44:
```go
liveDeletedTxnID = uint64(math.MaxInt64)
```

internal/datastore/postgres/gc.go
```diff
@@ -0,0 +1,145 @@
...
+			log.Info().Msg("shutting down garbage collection worker for postgres driver")
+			return pgd.gcCtx.Err()
+
+		case <-time.After(pgd.gcInterval):
+			err := pgd.collectGarbage()
+			if err != nil {
+				log.Warn().Err(err).Msg("error when attempting to perform garbage collection")
+			} else {
+				log.Debug().Msg("garbage collection completed for postgres")
+			}
```
jzelinskie, line 30: might actually want to copy that backoff v4 exponential backoff pattern here for gc

internal/datastore/memdb/memdb.go
jakedt: It's actually subtly different from that. Having the local var lets us distinguish between whether we allocated the transaction or someone else did in the error cases. On line 152 we wouldn't know if we had to abort the `tx`.

jakedt: It's actually subtly different from that. Having the local var lets us distinguish between whether we allocated the transaction or someone else did in the error cases. On line 152 we wouldn't know if we had to abort the `tx`.

internal/datastore/postgres/gc.go
jakedt: Open an issue. This PR didn't change GC at all.

internal/datastore/spanner/readonly.go
```diff
@@ -0,0 +1,219 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/authzed/spicedb/internal/datastore/common"
+	"github.com/authzed/spicedb/internal/datastore/options"
+	"github.com/authzed/spicedb/pkg/datastore"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// The underlying Spanner shared read transaction interface is not exposed, so we re-create
```
josephschorr, line 19: Can we link to the interface's definition?

internal/datastore/spanner/readonly.go
```diff
@@ -0,0 +1,219 @@
...
+var queryTuples = sql.Select(
+	colNamespace,
+	colObjectID,
+	colRelation,
+	colUsersetNamespace,
+	colUsersetObjectID,
+	colUsersetRelation,
+).From(tableRelationship)
+
+var schema = common.SchemaInformation{
```
josephschorr, line 210: Why is this in `readonly.go`?

internal/datastore/spanner/spanner.go
```diff
@@ -102,6 +95,41 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore,
 	return ds, nil
 }
 
+func (sd spannerDatastore) SnapshotReader(revision datastore.Revision) datastore.Reader {
+	txSource := func() readTX {
+		return sd.client.Single().WithTimestampBound(spanner.ReadTimestamp(timestampFromRevision(revision)))
+	}
+	querySplitter := common.TupleQuerySplitter{
+		Executor:         queryExecutor(txSource),
+		UsersetBatchSize: 100,
```
josephschorr, line 104: Move to a const?

internal/datastore/spanner/spanner.go
```diff
@@ -102,6 +95,41 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore,
...
+	fn datastore.TxUserFunc,
+) (datastore.Revision, error) {
+	ts, err := sd.client.ReadWriteTransaction(ctx, func(ctx context.Context, spannerRWT *spanner.ReadWriteTransaction) error {
+		txSource := func() readTX {
+			return spannerRWT
+		}
+
+		querySplitter := common.TupleQuerySplitter{
+			Executor:         queryExecutor(txSource),
+			UsersetBatchSize: 100,
```
josephschorr, line 121: And use it here

internal/datastore/mysql/datastore.go
```diff
@@ -39,14 +41,14 @@ const (
...
-	batchDeleteSize         = 1000
-	noLastInsertID          = 0
-	seedingTimeout          = 10 * time.Second
+	errUnableToInstantiate = "unable to instantiate datastore: %w"
+	liveDeletedTxnID       = uint64(math.MaxInt64)
+	batchDeleteSize        = 1000
+	noLastInsertID         = 0
+	seedingTimeout         = 10 * time.Second
+
+	errMysqlLockWaitTimeout = 1205
```
josephschorr, line 50: Can we link to these error codes in the MySQL docs?

internal/datastore/mysql/datastore.go
```diff
@@ -196,7 +196,107 @@ func NewMySQLDatastore(uri string, options ...Option) (*Datastore, error) {
...
+		createTxFunc,
+		querySplitter,
+		buildLivingObjectFilterForRevision(rev),
+	}
+}
+
+func noCleanup() error { return nil }
+
+// TODO (@vroldanbet) dupe from postgres datastore - need to refactor
+// ReadWriteTx tarts a read/write transaction, which will be committed if no error is
```
josephschorr, line 226: starts

internal/datastore/mysql/datastore.go
```diff
@@ -196,7 +196,107 @@ func NewMySQLDatastore(uri string, options ...Option) (*Datastore, error) {
...
+// returned and rolled back if an error is returned.
+func (mds *Datastore) ReadWriteTx(
+	ctx context.Context,
+	fn datastore.TxUserFunc,
+) (datastore.Revision, error) {
+	var err error
+	for i := uint8(0); i <= mds.maxRetries; i++ {
+		var newTxnID uint64
+		if err = BeginTxFunc(ctx, mds.db, &sql.TxOptions{Isolation: sql.LevelSerializable}, func(tx *sql.Tx) error {
+			_, err := tx.ExecContext(ctx, "SET innodb_lock_wait_timeout=1")
```
josephschorr, line 236: Does this set the timeout on the lock to 1s? If so, please document that and link to the docs for it

internal/datastore/mysql/datastore.go
```diff
@@ -196,7 +196,107 @@ func NewMySQLDatastore(uri string, options ...Option) (*Datastore, error) {
...
+			}
+
+			querySplitter := common.TupleQuerySplitter{
+				Executor:         newMySQLExecutor(tx),
+				UsersetBatchSize: mds.usersetBatchSize,
+			}
+
+			rwt := &mysqlReadWriteTXN{
+				&mysqlReader{
+					mds.QueryBuilder,
```
josephschorr, line 257: Use named properties here, since we have more than 2 or 3

internal/datastore/mysql/datastore.go
```diff
@@ -196,7 +196,107 @@ func NewMySQLDatastore(uri string, options ...Option) (*Datastore, error) {
...
+
+			return datastore.NoRevision, err
+		}
+
+		return revisionFromTransaction(newTxnID), nil
+	}
+	return datastore.NoRevision, fmt.Errorf("max retries exceeded: %w", err)
+}
+
+func errorRetryable(err error) bool {
```
josephschorr, line 285: nit: I prefer `isErrorRetryable`, although I know many people dislike that naming convention

internal/datastore/mysql/datastore_test.go
```diff
@@ -150,13 +150,12 @@ func GarbageCollectionTest(t *testing.T, ds datastore.Datastore) {
...
 	// Write basic namespaces.
-	_, err = ds.WriteNamespace(ctx, namespace.Namespace(
-		"resource",
-		namespace.Relation("reader", nil),
-	))
-	req.NoError(err)
-
-	writtenAt, err := ds.WriteNamespace(ctx, namespace.Namespace("user"))
+	writtenAt, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+		return rwt.WriteNamespaces(namespace.Namespace(
```
josephschorr, line 154: Put a newline after `WriteNamespaces(` to make it more clear that you're writing two namespaces

internal/datastore/mysql/datastore_test.go
```diff
@@ -311,13 +298,12 @@ func GarbageCollectionByTimeTest(t *testing.T, ds datastore.Datastore) {
...
 	// Write basic namespaces.
-	_, err = ds.WriteNamespace(ctx, namespace.Namespace(
-		"resource",
-		namespace.Relation("reader", nil),
-	))
-	req.NoError(err)
-
-	_, err = ds.WriteNamespace(ctx, namespace.Namespace("user"))
+	_, err = ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+		return rwt.WriteNamespaces(namespace.Namespace(
```
josephschorr, line 302: ditto

internal/datastore/mysql/datastore_test.go
```diff
@@ -340,14 +326,12 @@ func GarbageCollectionByTimeTest(t *testing.T, ds datastore.Datastore) {
 	}
 	relationship := tuple.ToRelationship(tpl)
 
-	relLastWriteAt, err := ds.WriteTuples(
-		ctx,
-		nil,
-		[]*v1.RelationshipUpdate{{
+	relLastWriteAt, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
```
josephschorr, line 329: Maybe we should have a datastore util function for writing a set of rels, without the syntax around opening the transaction, then calling WriteRelationships?

internal/datastore/mysql/readwrite.go
```diff
@@ -0,0 +1,288 @@
...
+
+	selectForUpdateQuery := rwt.QueryTupleIdsQuery
+
+	clauses := sq.Or{}
+
+	// Process the actual updates
+	for _, mut := range mutations {
+		rel := mut.Relationship
+
+		// Implementation for TOUCH deviates from PostgreSQL datastore to prevent a deadlock in MySQL
```
josephschorr, line 56: @vroldanbet Can you explain more context here on how this prevents a deadlock? We might want to extend the comment here

internal/datastore/mysql/readwrite.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (rwt *mysqlReadWriteTXN) WriteNamespaces(newNamespaces ...*core.NamespaceDefinition) error {
+	// TODO (@vroldanbet) dupe from postgres datastore - need to refactor
+	ctx := datastore.SeparateContextWithTracing(rwt.ctx)
+
+	ctx, span := tracer.Start(ctx, "WriteNamespaces")
+	defer span.End()
+
+	// span.SetAttributes(common.ObjNamespaceNameKey.String(newNamespace.Name))
```
josephschorr, line 197: Why is this commented out?

internal/datastore/mysql/tx.go
```diff
@@ -0,0 +1,29 @@
...
+
+// BeginTXFunc is a polyfill for database/sql which does a closure style transaction lifecycle.
+// The underlying transaction is aborted if the supplied function returns an error.
+// The underlying transaction is committed if the supplied function returns nil.
+func BeginTxFunc(ctx context.Context, db *sql.DB, txOptions *sql.TxOptions, f func(*sql.Tx) error) error {
+	tx, err := db.BeginTx(ctx, txOptions)
+	if err != nil {
+		return err
+	}
+	defer migrations.LogOnError(ctx, tx.Rollback)
```
josephschorr, line 18: Can we move this into the core mysql package? Its odd to rely on the migrations package

pkg/datastore/datastore.go
```diff
@@ -43,18 +42,57 @@ type RevisionChanges struct {
 	Changes  []*core.RelationTupleUpdate
 }
 
+type Reader interface {
+	// QueryTuples reads relationships starting from the resource side.
```
josephschorr, line 46: Fix doc comment

josephschorr, line 46: Add comma after `relationships`

pkg/datastore/datastore.go
```diff
@@ -43,18 +42,57 @@ type RevisionChanges struct {
...
+	QueryRelationships(
+		ctx context.Context,
+		filter *v1.RelationshipFilter,
+		options ...options.QueryOptionsOption,
+	) (RelationshipIterator, error)
+
+	// ReverseQueryRelationships reads relationships starting from the subject.
```
josephschorr, line 53: ditto

pkg/datastore/datastore.go
```diff
@@ -43,18 +42,57 @@ type RevisionChanges struct {
...
+	ReverseQueryRelationships(
+		ctx context.Context,
+		subjectFilter *v1.SubjectFilter,
+		options ...options.ReverseQueryOptionsOption,
+	) (RelationshipIterator, error)
+
+	// ReadNamespace reads a namespace definition and version and returns it, and the revision at
```
josephschorr, line 60: ReadNamespace reads a namespace definition, its version, and its last updated revision

internal/datastore/mysql/readwrite.go
vroldanbet: I'm going to defer to @bryanhuhta as they worked on the actual solution back in the day. For what's worth, the changes decreased the likelihood of the dead-lock, but it's still happening when doing massive concurrent imports.

internal/middleware/consistency/consistency_test.go
```diff
@@ -16,180 +16,169 @@ import (
...
-	"github.com/authzed/spicedb/internal/datastore/memdb"
+	"github.com/authzed/spicedb/internal/datastore/proxy/proxy_test"
 	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/zedtoken"
 	"github.com/authzed/spicedb/pkg/zookie"
 )
 
+var (
+	zero      = decimal.NewFromInt(0)
```
josephschorr, line 27: Are these from the mock?

internal/services/consistency_test.go
```diff
@@ -64,26 +64,27 @@ func TestConsistency(t *testing.T) {
 				t.Run(path.Base(filePath), func(t *testing.T) {
 					for _, dispatcherKind := range []string{"local", "caching"} {
 						t.Run(dispatcherKind, func(t *testing.T) {
-							t.Parallel()
+							// t.Parallel()
```
josephschorr, line 67: Why?

internal/services/v0/acl.go
```diff
@@ -17,7 +17,6 @@ import (
 	"google.golang.org/grpc/codes"
```
josephschorr, line 17: Should we just drop V0?

internal/services/v0/namespace.go
```diff
@@ -43,119 +42,118 @@ func NewNamespaceServer() v0.NamespaceServiceServer {
...
-				err = shared.ErrorIfTupleIteratorReturnsTuples(
-					ctx,
-					qy,
-					qyErr,
-					"cannot delete relation `%s` in definition `%s`, as a relationship exists under it", delta.RelationName, config.Name)
-				if err != nil {
-					return nil, rewriteNamespaceError(ctx, err)
-				}
+			// Ensure that the updated namespace does not break the existing tuple data.
+			existing, _, err := rwt.ReadNamespace(ctx, config.Name)
```
josephschorr, line 68: I don't think we need this anymore, since we're inside a transaction. We might be able to just grab the namespace from the validated type system, which has it in memory

internal/services/v0/namespace.go
```diff
@@ -184,62 +182,67 @@ func (nss *nsServer) ReadConfig(ctx context.Context, req *v0.ReadConfigRequest)
...
-		err = shared.ErrorIfTupleIteratorReturnsTuples(
-			ctx,
-			qy,
-			qyErr,
-			"cannot delete definition `%s`, as a relationship exists under it", nsName)
-		if err != nil {
-			return nil, rewriteNamespaceError(ctx, err)
-		}
+			// Check for relationships under the namespace.
+			qy, qyErr := rwt.QueryRelationships(
```
josephschorr, line 197: Now that I think about it, we might want to reverse these checks: put the `ReadNamespace` below the queries: after all, if there does exist a relationship with that namespace, then we don't need to read it. We only need to read it if there are none

internal/services/v1alpha1/schema_test.go
```diff
@@ -182,36 +182,36 @@ func TestSchemaDeleteRelation(t *testing.T) {
...
-		Schema: `definition example/user {}
-		
-			definition example/document {}`,
-	})
-	require.Nil(t, err)
-
-	rev, err := nspkg.DecodeV1Alpha1Revision(writeResp.ComputedDefinitionsRevision)
-	require.NoError(t, err)
-	require.Len(t, rev, 2)
+	// // Delete the relationship.
```
josephschorr, line 192: Why is this commented out?

internal/datastore/postgres/reader.go
```diff
@@ -0,0 +1,206 @@
...
+	return nsDefs, err
+}
+
+func loadAllNamespaces(ctx context.Context, tx pgx.Tx, query sq.SelectBuilder) ([]*core.NamespaceDefinition, error) {
+	sql, args, err := query.ToSql()
+	if err != nil {
+		return nil, err
+	}
+
+	var nsDefs []*core.NamespaceDefinition
```
jzelinskie, line 177: move this to right above `for rows.Next()` so it's clear that's what the loop is doing

internal/datastore/postgres/reader.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, datastore.NoRevision, err
+	case err == nil:
+		return loaded, version, nil
+	default:
+		return nil, datastore.NoRevision, fmt.Errorf(errUnableToReadConfig, err)
+	}
+}
+
+func loadNamespace(ctx context.Context, namespace string, tx pgx.Tx, baseQuery sq.SelectBuilder) (*core.NamespaceDefinition, datastore.Revision, error) {
+	ctx = datastore.SeparateContextWithTracing(ctx)
```
jzelinskie, line 125: you can inline this call to SeparateContextWithTracing in the tracer.Start()

internal/datastore/postgres/readwrite.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	if bulkWriteHasValues {
+		sql, args, err := bulkWrite.ToSql()
+		if err != nil {
+			return fmt.Errorf(errUnableToWriteRelationships, err)
+		}
+
+		_, err = rwt.tx.Exec(ctx, sql, args...)
+		if err != nil {
```
jzelinskie, line 101: if _, err := rwt.tx.Exec(ctx, sql, args...); err != nil {

internal/datastore/postgres/readwrite.go
```diff
@@ -0,0 +1,258 @@
...
+		return fmt.Errorf(errUnableToDeleteRelationships, err)
+	}
+
+	return nil
+}
+
+func (rwt *pgReadWriteTXN) WriteNamespaces(newConfigs ...*core.NamespaceDefinition) error {
+	ctx := datastore.SeparateContextWithTracing(rwt.ctx)
+
+	ctx, span := tracer.Start(ctx, "WriteNamespaces")
```
jzelinskie, line 158: can inline this context like the rest

internal/datastore/postgres/readwrite.go
```diff
@@ -0,0 +1,258 @@
...
+	defer span.End()
+
+	// span.SetAttributes(common.ObjNamespaceNameKey.String(newConfig.Name))
```
jzelinskie, line 161: delete this or leave a comment describing it

internal/datastore/postgres/readwrite.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		return fmt.Errorf(errUnableToWriteConfig, err)
+	}
+	span.AddEvent("Namespace config written")
+
+	return nil
+}
+
+func (rwt *pgReadWriteTXN) DeleteNamespace(nsName string) error {
+	ctx := datastore.SeparateContextWithTracing(rwt.ctx)
```
jzelinskie, line 205: probably want to start a span here, too

internal/datastore/spanner/readonly.go
jakedt: Because it's only used by the schema query filterer in `readonly.go`.

internal/datastore/mysql/datastore.go
jakedt: Yes it does. The default is 50s, which means you could have interleaved write transactions just sitting there for 50s waiting for what amounts to a deadlock. @vroldanbet is it appropriate to override this setting at the session/transaction layer? Can we even override it per transaction?

internal/datastore/mysql/datastore.go
jakedt: I much prefer doing it this way because I can't forget to initialize things. It has saved my bacon countless times.

internal/datastore/mysql/datastore_test.go
jakedt: It's really only an issue for tests, and I already moved them all over. I don't think we should hide the way it works under the covers.

internal/datastore/mysql/readwrite.go
jakedt: Actually maybe we want to allow it to deadlock, that should be faster to detect and retry than the current lock wait timeout that we're doing.

internal/datastore/mysql/readwrite.go
jakedt: Oh, because I didn't know what to do when there were more than one namespace. Suggestions?

internal/datastore/mysql/tx.go
jakedt: This PR doesn't change anything about that. If you want to move it, file an issue or open another PR.

internal/datastore/proxy/caching.go
```diff
@@ -0,0 +1,152 @@
...
+
+	loaded := loadedRaw.(*cacheEntry)
+
+	return loaded.def, loaded.updated, loaded.notFound
+}
+
+type nsCachingRWT struct {
+	sync.Mutex
+	datastore.ReadWriteTransaction
+	namespaceCache map[string]cacheEntry
```
jzelinskie, line 117: This is actually a good use case for `sync.Map`.

internal/datastore/proxy/caching.go
```diff
@@ -0,0 +1,152 @@
...
+) (*core.NamespaceDefinition, datastore.Revision, error) {
+	// Check the nsCache.
+	nsRevisionKey := fmt.Sprintf("%s@%s", nsName, r.rev)
+
+	loadedRaw, found := r.p.c.Get(nsRevisionKey)
+	if !found {
+		// We couldn't use the cached entry, load one
+		var err error
+		loadedRaw, err, _ = r.p.readNsGroup.Do(nsRevisionKey, func() (interface{}, error) {
+			loaded, updatedRev, err := r.Reader.ReadNamespace(ctx, nsName)
```
jzelinskie, line 87: This needs to be `r.Datastore.SnapshotReader(rev).ReadNamespace(ctx, nsName)`

internal/middleware/consistency/consistency_test.go
jakedt: We load them into the mock.

internal/services/v0/namespace.go
jakedt: I don't want to spend any time on v0 if I don't have to. Is it correct as written?

internal/services/v0/namespace.go
jakedt: Ditto.

internal/datastore/postgres/reader.go
jakedt: Pre-allocate it to what? We have no idea how many rows we're going to get back.

internal/datastore/postgres/readwrite.go
jakedt: Just like with MySQL, I don't know what to set it to now that the method is vectorized. Suggestions?

internal/datastore/postgres/readwrite.go
jakedt: Added it. I would like to hold off on any unnecessary refactoring.

internal/datastore/proxy/caching_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	ds, err := NewCachingDatastoreProxy(dsMock, nil)
+	require.NoError(err)
+
+	rev, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+		_, updatedA, err := rwt.ReadNamespace(ctx, nsA)
+		require.NoError(err)
+		require.Equal(zero.IntPart(), updatedA.IntPart())
+
+		// This should not call out the mock RWT again
```
jzelinskie, line 103: There's an AssertNumberOfCalls() method for mocks that you could use to prove this comment

jakedt: I don't need to. The mock is set with `.Once()` and it will panic if it _does_ call out again.

jzelinskie (approved):
🥹

internal/datastore/mysql/readwrite.go
bryanhuhta: Allowing the deadlocks to happen could be faster, I haven't really measured the impact of a deadlock detection and rollback.

> Can you explain more context here on how this prevents a deadlock? We might want to extend the comment here

The TL;DR of the situation is MySQL takes greedy row locks when trying to update a row that doesn't exist, whereas Postgres doesn't have the same behavior.

The longer version: With the Postgres implementation, there are basically two database actions that happen in a transaction during a `TOUCH` operation (assuming a single tuple):

- Update the most recent tuple record as "deleted"
- Insert a new tuple record as "newest"

Using that same logic in MySQL, two locks would need to be gathered:

- **next-key lock** to block writes on the row (multiple transactions can hold this at once)
- **insert intention gap lock** to insert a new record (multiple transactions can hold this, as long as they aren't inserting into the same gap)

This is fine if the tuple to marked as "deleted" already exists. However, if it doesn't, then the next-key lock will lock rows between the last row and the `supremum` row. This creates a situation where transaction `T1` and `T2` are trying to perform a `TOUCH` on two records that don't exist:

```
*
| T1 acquires next-key lock (locks rows 1 to supremum)
| T2 acquires next-key lock (locks rows 1 to supremum) 
| T1 acquires gap lock (locks row of next possible id, which is between 1 and supremum)
| T2 acquires gap lock (locks row of T1 + 1, which is between 1 and supremum)
!! Deadlock, T1 needs to insert into a gap that T2 holds a lock on (via next-key lock)
!! and T2 needs to insert into a gap that T1 holds a lock on (via next-key lock). T2 rolls back.
```

In our case, the fix was to only update rows that exist in the database. That stops a transaction from acquiring a next-key lock up to `supremum`. We still have deadlocks in certain situations that I haven't quite been able to figure out yet.

For a better explanation of how this deadlock occurs, there's a 2 million year old MySQL bug report: https://bugs.mysql.com/bug.php?id=25847.

jakedt: > there's a 2 million year old MySQL bug report

😂 

The bug mentions using `select for update`, and the MySQL code calls the query `selectForUpdateQuery`, but I don't actually see where any queries are actually using the `for update` modifier. Could this be related to your ongoing deadlocks?

I would like to work together to get real data on whether it's faster/better to just let the deadlock happen and retry the transaction. This is more or less the entire justification for doing the datastore v2 work now. We have similar client-side retry requirements across all of our datastores.

internal/datastore/mysql/datastore.go
jakedt: I've refactored this to move it into an option, PTAL.

vroldanbet: @jakedt I fail to see practical reasons to wait 50s for a row-lock, so this would seem sensible to me. The [docs](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_lock_wait_timeout) indicate this can be set at both global and session level. If what you intend is to set it only for that transaction, I'm not sure that's exactly what will happen. The session lasts for as long as the connection is around, which is pooled, and governed by the max lifetime established. So I suspect that state will be set even after that transaction completes, can you check that?

Even if that state _leaked_ outside of the transaction, I think it would be _fine_, but it may make things difficult to troubleshoot in the future if the state only emerges when a specific operation takes place in the backend.

jakedt: I pushed an update this morning which centralizes it on the connection factory. Now all of the sessions will get it applied as soon as the connection opens. The tests verify that it has the intended effect.

vroldanbet: That's great and should address the problem 👍🏻 But at that point it may be worth recommending it as a global. It could be done perhaps in a migration, or through documentation? I guess since a part of me is concerned about doing that in the factory because it leads to _more time waiting for a connection to be opened and more overhead_, in light of all the contention we are observing on getting connections from the pool.

internal/services/v0/namespace.go
josephschorr: Yeah, its just extra work for now, but that's not any different than now

internal/datastore/common/sql.go
```diff
@@ -207,41 +205,36 @@ func (tqs TupleQuerySplitter) SplitAndExecuteQuery(
...
+// TransactionFactory is done with the transaction.
+type TxCleanupFunc func(context.Context)
 
-func NewPGXExecutor(conn *pgxpool.Pool, prepTxn TransactionPreparer) ExecuteQueryFunc {
-	return func(ctx context.Context, revision datastore.Revision, sql string, args []interface{}) ([]*core.RelationTuple, error) {
+// TxFactory returns a transaction, cleanup function, and any errors that may have
+// occurred when building the transaction.
+type TxFactory func(context.Context) (pgx.Tx, TxCleanupFunc, error)
+
+func NewPGXExecutor(txSource TxFactory) ExecuteQueryFunc {
```
josephschorr, line 224: Doc comment?

internal/datastore/crdb/crdb.go
```diff
@@ -159,11 +154,101 @@ type crdbDatastore struct {
...
+func (cds *crdbDatastore) SnapshotReader(rev datastore.Revision) datastore.Reader {
+	createTxFunc := func(ctx context.Context) (pgx.Tx, common.TxCleanupFunc, error) {
+		tx, err := cds.pool.BeginTx(ctx, pgx.TxOptions{AccessMode: pgx.ReadOnly})
+		if err != nil {
+			return nil, nil, err
+		}
+
+		cleanup := func(ctx context.Context) {
+			if err := tx.Rollback(ctx); err != nil {
+				log.Ctx(ctx).Err(err).Msg("error rolling back transaction")
```
josephschorr, line 173: Add additional context to the error message, since it is the same message as below, and once line numbers move, it'll be harder to pin down the exact location if we see the error occur

internal/datastore/crdb/readwrite.go
```diff
@@ -0,0 +1,285 @@
...
+	ctx, span := tracer.Start(datastore.SeparateContextWithTracing(rwt.ctx), "WriteTuples")
+	defer span.End()
+
+	bulkWrite := queryWriteTuple
+	var bulkWriteCount int64
+
+	bulkTouch := queryTouchTuple
+	var bulkTouchCount int64
+
+	// Process the actual updates	var tupleCountChange int64
```
josephschorr, line 94: `var tupleCountChange int64`?

internal/datastore/crdb/readwrite.go
```diff
@@ -0,0 +1,285 @@
...
+		colUsersetNamespace: r.Subject.Object.ObjectType,
+		colUsersetObjectID:  r.Subject.Object.ObjectId,
+		colUsersetRelation:  stringz.DefaultEmpty(r.Subject.OptionalRelation, datastore.Ellipsis),
+	}
+}
+
+func (rwt *crdbReadWriteTXN) DeleteRelationships(filter *v1.RelationshipFilter) error {
+	ctx, span := tracer.Start(datastore.SeparateContextWithTracing(rwt.ctx), "DeleteRelationships")
+	defer span.End()
+	// Add clauses for the ResourceFilter
```
josephschorr, line 175: newline before this line

internal/datastore/crdb/readwrite.go
```diff
@@ -0,0 +1,285 @@
...
+	query := queryWriteNamespace
+
+	for _, newConfig := range newConfigs {
+		rwt.addOverlapKey(newConfig.Name)
+
+		serialized, err := proto.Marshal(newConfig)
+		if err != nil {
+			return fmt.Errorf(errUnableToWriteConfig, err)
+		}
+		query = query.Values(newConfig.Name, serialized)
```
josephschorr, line 228: We might have a risk of hitting query limits here too

internal/datastore/crdb/crdb.go
jakedt: i mean they will have different underlying errors.

internal/datastore/crdb/readwrite.go
jakedt: Yeah, there is a hard upper bound on how big our schema can get now, but it's unlikely to hit this limit without first hitting the overall schema limit.

internal/datastore/memdb/memdb.go
```diff
@@ -348,93 +35,208 @@ func NewMemdbDatastore(
...
-	lastChangeRaw, err := txn.Last(tableTransaction, indexID)
-	if err != nil {
-		return 0, err
+	if err := mdb.checkRevisionLocal(revision); err != nil {
+		return &memdbReader{nil, nil, datastore.NoRevision, err}
 	}
 
-	if lastChangeRaw != nil {
-		newTransactionID = lastChangeRaw.(*transaction).id + 1
+	revIndex := sort.Search(len(mdb.revisions), func(i int) bool {
```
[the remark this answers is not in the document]
jakedt: No

internal/datastore/mysql/connection.go
```diff
@@ -74,3 +74,50 @@ func instrumentConnector(c driver.Connector) (driver.Connector, error) {
...
+	defer span.End()
+
+	conn, err := s.conn.Connect(ctx)
+	if err != nil {
+		span.RecordError(err)
+		log.Ctx(ctx).Error().Err(err).Msg("failed to open db connection")
+		return nil, fmt.Errorf("failed to open connection to db: %w", err)
+	}
+
+	// The go mysql driver implements the ExecerContext interface, assert that here.
```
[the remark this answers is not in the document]
jakedt: I mean you can see it on the next line. What do you want here?

internal/datastore/memdb/schema.go
```diff
@@ -0,0 +1,206 @@
...
+		r.namespace,
+		r.resourceID,
+		r.relation,
+		r.subjectNamespace,
+		r.subjectObjectID,
+		r.subjectRelation,
+	))
+}
+
+func (r relationship) Relationship() *v1.Relationship {
```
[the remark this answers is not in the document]
jakedt: No, this is going from the flat struct that we use to represent a row in the memdb to the relationship.

internal/datastore/mysql/connection.go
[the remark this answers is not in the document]
josephschorr: Just seems a weird name, but its fine

internal/datastore/memdb/schema.go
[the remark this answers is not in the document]
josephschorr: We do convert to a core rel below, so we could in theory chain that way, but not sure it is worth it

internal/datastore/mysql/connection.go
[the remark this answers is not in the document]
jakedt: I didn't name it, it's part of the go sql interface.

josephschorr (approved):
LGTM

PR #527, opened 2022-04-13
Quantize revisions for memdb, postgres datastores

jakedt:

PR #582, opened 2022-05-02
Implement revision quantization for MySQL

bryanhuhta:
This PR implements the cached revision quantization optimization for the MySQL driver. We attempted to stay as true to the Postgres implementation as possible.

See also:
- https://github.com/authzed/spicedb/pull/527
- https://github.com/authzed/spicedb/pull/555


internal/datastore/mysql/revisions.go
```diff
@@ -0,0 +1,178 @@
...
+
+// TODO (@vroldanbet) dupe from postgres datastore - need to refactor
+func revisionFromTransaction(txID uint64) datastore.Revision {
+	return decimal.NewFromInt(int64(txID))
+}
+
+// TODO (@vroldanbet) dupe from postgres datastore - need to refactor
+func transactionFromRevision(revision datastore.Revision) uint64 {
+	return uint64(revision.IntPart())
+}
```
ecordell, line 178:
```suggestion
func revisionFromTransaction(txID uint64) datastore.Revision {
	return decimal.NewFromBigInt(new(big.Int).SetUint64(txID), 0)
}

func transactionFromRevision(revision datastore.Revision) uint64 {
	return revision.BigInt().Uint64()
}
```

I think something like this will work, but I haven't tested it. Note that this does differ from the postgres implementation; postgres's `BIGSERIAL` is a signed int, but the revision field in the mysql schema is `BIGINT UNSIGNED`.

internal/datastore/mysql/revisions.go
```diff
@@ -0,0 +1,178 @@
...
+	// database seeding in runtime, instead of through migrate command
+	ctx, span := tracer.Start(ctx, "HeadRevision")
+	defer span.End()
+
+	revision, err := mds.loadRevision(ctx)
+	if err != nil {
+		return datastore.NoRevision, err
+	}
+	if revision == 0 {
+		return datastore.NoRevision, nil
```
ecordell, line 75: Why return `NoRevision` here vs. just returning the `0` revision?

internal/datastore/mysql/datastore_test.go
```diff
@@ -495,6 +498,102 @@ func ChunkedGarbageCollectionTest(t *testing.T, ds datastore.Datastore) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.testName, func(t *testing.T) {
+			require := require.New(t)
+			ctx, cancel := context.WithTimeout(context.Background(), 900*time.Second)
+			defer cancel()
+
+			ds := testdatastore.RunMySQLForTesting(t, "").NewDatastore(t, func(engine, uri string) datastore.Datastore {
```
ecordell, line 546: `RunMySQLForTesting` can be pulled out of the loop, and each test just makes it's own logical datastore with `NewDatastore`

This might also help lower that 900s timeout?

ecordell (changes requested):
Looking good! 

I had a few comments about some (unlikely) overflows, but this does appear to quantize the same way as postgres now 👍

internal/datastore/mysql/revisions.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("createNewTransaction: %w", err)
+	}
+
+	result, err := tx.ExecContext(ctx, createQuery)
+	if err != nil {
+		return 0, fmt.Errorf("createNewTransaction: %w", err)
+	}
+
+	lastInsertID, err := result.LastInsertId()
```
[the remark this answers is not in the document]
bryanhuhta: I did some digging into the MySQL driver implementation of `LastInsertId()` ([source](https://github.com/go-sql-driver/mysql/blob/ad9fa14acdcf7d0533e7fbe58728f3d216213ade/connection.go#L316-L322)):

```go
// mysqlConn struct
type mysqlConn struct {
	// ... some stuff
	affectedRows     uint64
	insertId         uint64
	// more stuff ...
}
```

```go
// Exec implementation
// ... some stuff
err := mc.exec(query)
if err == nil {
	return &mysqlResult{
		affectedRows: int64(mc.affectedRows),
		insertId:     int64(mc.insertId),
	}, err
}
// more stuff ...
```

Internally, the MySQL driver uses a `uint64` to store `insertedId`. It will convert that to an `int64` to return from `LastInsertId()`. This lead me to wonder if there would be overflow issues here. So I tested in the Go playground: https://go.dev/play/p/m6qQJtcotAe. As the playground shows, since the original value was a `uint64`, we can convert the `int64` to a `uint64` without a problem.

internal/datastore/mysql/revisions.go
bryanhuhta: `loadRevision()` can return a revision of `0` with no error indicating the database hasn't yet been seeded.

https://github.com/authzed/spicedb/blob/b33fe794326461cd95cc77f3c106974818e83946/internal/datastore/mysql/datastore.go#L643-L654

Returning `0` here would imply the database was seeded with the initial value of `0`, which would be incorrect. Instead, we return `NoRevision` and no error to signal there isn't a revision but also no error.

internal/datastore/mysql/revisions.go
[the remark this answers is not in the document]
ecordell: Awesome, thanks for digging in 👍

internal/datastore/mysql/revisions.go
ecordell: I haven't looked at the seeding part much, would it make sense to return an error in that case?

Anyway that's not really an issue with this PR so I'll resolve this.

internal/datastore/mysql/revisions.go
```diff
@@ -0,0 +1,178 @@
...
+	// slightly changed to support no revisions at all, needed for runtime seeding of first transaction
+	ctx, span := tracer.Start(ctx, "loadRevision")
+	defer span.End()
+
+	query, args, err := mds.GetLastRevision.ToSql()
+	if err != nil {
+		return 0, fmt.Errorf(errRevision, err)
+	}
+
+	var revision sql.NullInt64
```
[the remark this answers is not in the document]
ecordell: your playground convinced me that this change is unnecessary, and golang does the right thing in a uint64->int64->uint64 conversion

internal/datastore/mysql/revisions.go
bryanhuhta: Changed and added tests in https://github.com/authzed/spicedb/pull/582/commits/b93e5c083e7794e241c01f2a5426cf61bc112c6d

ecordell (approved):
LGTM

internal/datastore/mysql/revisions.go
bryanhuhta: Yeah, that probably would make sense. The `isSeeded` function is the one checking for no error and `NoRevision` to trigger seeding

https://github.com/authzed/spicedb/blob/a86dccf77489314d6cc3ec1fe1f722330865ff06/internal/datastore/mysql/datastore.go#L454-L470

If we refactor this to return an error, `isSeeded` will have to inspect the error and check if it's equal to `ErrNoRevisionSeed` or some such.

PR #592, opened 2022-05-03
pg: set timezone to utc for revision selection

ecordell:
I noticed this when reviewing #582

It looks like this does make a difference if the db has a timezone set: https://dbfiddle.uk/?rdbms=postgres_14&fiddle=a63a65dd0d6b02c771a2ad2a1d4ffa88


josephschorr (approved):
LGTM

Issue #595, opened 2022-05-07
Migration functions should have a context.Context parameter


jakedt:
Also, we should probably be able to execute the migration and the part that writes the version in a single transaction.

Issue #347, opened 2021-12-21
Migrations engine should pass a context and be configurable with a timeout


jakedt:
Right now the operations in the migration engine can take arbitrarily long. The engine should pass a context to operations, operations should respect context cancelation, and the engine should use a configurable timeout for the operations.

PR #596, opened 2022-05-09
Propagate context in migrations

nbarbey:
This is to allow timeouts in migration operations.

This is related to issue #595 and issue #347.

### DIFF
--- a/internal/testserver/datastore/mysql.go
+++ b/internal/testserver/datastore/mysql.go
@@ -1,6 +1,7 @@
 package datastore
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"testing"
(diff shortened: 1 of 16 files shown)


pkg/migrate/migrate.go
```diff
@@ -30,7 +31,7 @@ type Driver interface {
 	Version() (string, error)
```
josephschorr, line 31: Can you also add context to the `Version()` call as well?

nbarbey: Done !

josephschorr:
@nbarbey Please rebase on main

josephschorr:
@nbarbey Mind rebasing one more time?

internal/testserver/datastore/mysql.go
```diff
@@ -109,9 +110,9 @@ func (mb *mysqlTester) NewDatabase(t testing.TB) string {
 func (mb *mysqlTester) runMigrate(t testing.TB, dsn string) {
 	driver, err := migrations.NewMySQLDriverFromDSN(dsn, mb.options.Prefix)
 	require.NoError(t, err, "failed to create migration driver: %s", err)
-	err = migrations.Manager.Run(driver, migrate.Head, migrate.LiveRun)
+	err = migrations.Manager.Run(context.Background(), driver, migrate.Head, migrate.LiveRun)
```
josephschorr, line 113: I think the duplication here might be in error

nbarbey: Indeed, thanks ! This must have slipped through my attention while resolving the rebase conflicts ...

josephschorr (approved):
LGTM

PR #598, opened 2022-05-11
Follow-up: add metrics for gc in mysql datastore

christroger:
Follow-up for the MySQL Datastore implementation.

In this PR I add the missing garbage collector metrics to the MySQL datastore implementation.
There is one change I made in regard to the gauges, where I added a `success` Tag so we are able to monitor (outside of the logs) if the garbage collection is successful (The Postgres implementation does not have this additional Tag)


christroger:

PR #599, opened 2022-05-11
Follow-up adding CLI options to mysql datastore

christroger:
This is a Follow-up PR from the MySQL Datastore implementation.

It updates the `SplitQueryCount` because that is what is being fetched from `cobra`.

The following CLI options are now supported by the MySQL Datastore:

- `SplitAtUsersetCount`
- `GCMaxOperationTime`

Authors: @bryanhuhta and @cjs


christroger:
Closing for fixing the co-authoring in the commit message

PR #600, opened 2022-05-11
Follow-up adding CLI options to mysql datastore

christroger:
This is a Follow-up PR from the MySQL Datastore implementation.

It updates the `SplitQueryCount` because that is what is being fetched from `cobra`.

The following CLI options are now supported by the MySQL Datastore:

- `SplitAtUsersetCount`
- `GCMaxOperationTime`

Authors: @bryanhuhta and @cjs


vroldanbet (changes requested):
Hi @christroger! 👋🏻 👋🏻 

Thanks for opening this PR, it seems like even though the plumbing for the missing CLI options is there, it hasn't been actually wired up in the MySQL datastore factory method. Would you mind tackling that in this PR? 🙏🏻 

https://github.com/authzed/spicedb/blob/main/pkg/cmd/datastore/datastore.go#L260-L274

vroldanbet (approved):

josephschorr:
@christroger Mind rebasing?

PR #605, opened 2022-05-17
service-discovery: Added ZooKeeper based service discovery

gergof:
I have implemented an alternative service discovery that can be used without kubernetes. It uses [Apache ZooKeeper](https://zookeeper.apache.org/). It also contains the code necessary to work inside AWS ECS containers (it can get the IP from the task and instance metadata endpoint), but it falls back to the IP of the first public network interface. The address defined in `dispatch-cluster-addr` takes precedence in any case.

I will use this in our deployment on ECS. The SRV record method was not reliable so I made a custom resolver that uses ZooKeeper to discover the peers, since we were already using ZooKeeper for some of our existing services.

This is the first time I'm coding in Go, so I hope I didn't mess up anything.

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/jackc/pgx/v4 v4.16.0
 	github.com/stretchr/testify v1.7.1
 	golang.org/x/tools v0.1.10
-	google.golang.org/grpc v1.46.0
+	google.golang.org/grpc v1.46.2
 	mvdan.cc/gofumpt v0.3.1
 )
 
--- a/e2e/go.sum
+++ b/e2e/go.sum
@@ -614,8 +614,9 @@ google.golang.org/grpc v1.31.0/go.mod h1:N36X2cJ7JwdamYAgDz+s+rVMFjt3numwzf/HckM
 google.golang.org/grpc v1.33.1/go.mod h1:fr5YgcSWrqhRRxogOsw7RzIpsmvOZ6IcH4kBYTpR3n0=
 google.golang.org/grpc v1.36.0/go.mod h1:qjiiYl8FncCW8feJPdyg3v6XW24KsRHe+dy9BAGRRjU=
 google.golang.org/grpc v1.40.0/go.mod h1:ogyxbiOoUXAkP+4+xa6PZSE9DZgIHtSpzjDTB9KAK34=
-google.golang.org/grpc v1.46.0 h1:oCjezcn6g6A75TGoKYBPgKmVBLexhYLM6MebdrPApP8=
 google.golang.org/grpc v1.46.0/go.mod h1:vN9eftEi1UMyUsIF80+uQXhHjbXYbm0uXoFCACuMGWk=
+google.golang.org/grpc v1.46.2 h1:u+MLGgVf7vRdjEYZ8wDFhAVNmhkbJ5hmrA1LMWK1CAQ=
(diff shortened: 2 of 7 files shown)


gergof:
_I have read the CLA Document and I hereby sign the CLA_

gergof:
recheck

gergof:
Yes, of course.
I have changed the license to Apache 2.0 and updated the module.

ecordell:
I think it's a little unusual that the behavior in the resolver includes ECS metadata discovery and self-registration in zookeeper. I would be concerned about the behavior if the spicedb were ever stopped non-gracefully and failed to de-register.

What other options have you considered for ECS? Is there anything that we could include to automate it a bit more? The other resolver we have for kube "just works" if you deploy in a kubernetes cluster with appropriate permissions.

For example, I see that [App Mesh](https://docs.aws.amazon.com/app-mesh/latest/userguide/getting-started-ecs.html) makes the endpoints available via the xDS apis that envoy uses, and grpc-go already supports [using xDS to resolve endpoints](https://github.com/grpc/grpc-go/blob/87eb5b7502493f758e76c4d09430c0049a81a557/examples/features/xds/README.md). Would something like this work for you? If so we could enable xds resolution, which I think would be useful in more cases.

gergof:
> I would be concerned about the behavior if the spicedb were ever stopped non-gracefully and failed to de-register.

The znode created in zookeeper is an [ephemeral node](https://zookeeper.apache.org/doc/r3.4.13/zookeeperProgrammers.html#Ephemeral+Nodes). This means that as soon as the session closes, the node is deleted, so the service is unregistered. Calling .Unregister() gracefully closes the session, but it will also time out in 5 seconds (or a value defined using the `zookeeper-resolver-timeout` param). So in worst case 5 seconds after crashing the other nodes will be notified.

> What other options have you considered for ECS? Is there anything that we could include to automate it a bit more? The other resolver we have for kube "just works" if you deploy in a kubernetes cluster with appropriate permissions.

First I hoped that the ECS metadata endpoint can be used to retrieve the list of the running tasks. But it does not offer such functionality. It doesn't even specify the IP address if the task doesn't have it's own network interface (it uses the host or bridge networking modes). I came up with a solution that uses the task metadata endpoint to get the cluster name and the task ARN, uses these to make a request to the ECS.DescribeTasks API to get the service name and uses the ECS.ListTasks API to get the rest of the tasks from the same service. Then we can run ECS.DescribeTasks for the retrieved ARNs to get the IPs and port mappings. If the task definition uses the bridge or host networking mode, then we need to use the container instance ARN to retrieve the instance ID of the underlaying EC2 instance and use the EC2 api to get its IP address.

This solution has a lot of roundtrips, but it's actually very solid. The problem with it is that we can not detect changes in real time.

But AppMesh looks really promising. Somehow I didn't even hear about this service until now. It would solve the problem without requiring an external service (zookeeper in this case), and it can be easily configured. If you enable xDS resolution I would willingly test it and write a step-by-step guide to use it.

ecordell:
Thanks for the explanation; I wasn't familiar with ephemeral nodes in zookeeper. I think that's a reasonable option to keep on the backburner if AppMesh doesn't work.

I haven't used AppMesh before (I've only taken a look at the docs), but we can definitely try enabling xds support in spicedb and you can see if it works for you. 

If you'd like to make a PR adding xds support we'd be happy to review it, otherwise I should be able to get to it on Monday.

ecordell:
@gergof I made https://github.com/authzed/spicedb/pull/612 if you want to try building spicedb and testing with appmesh

PR #607, opened 2022-05-18
Update NewEnemy test for datastore v2

ecordell:
The updated test passed multiple times in another branch (though against a slightly older version of dsv2 branch) but I don't have enough data to confidently say it has an acceptable flake rate.

Temporarily removed the schema-based newenemy test. It would be nice to restore it if possible, but the larger transactions in dsv2 may make that more difficult (it was certainly more difficult for the data-based newenemy test). The protections we have in place are the same for both cases, so schema newenemy test only serves to verify implementation bugs, unlike the data newenemy test which verifies SpiceDB's approach to dealing with timestamp drift in CRDB.

### DIFF
--- a/e2e/newenemy/newenemy_test.go
+++ b/e2e/newenemy/newenemy_test.go
@@ -14,7 +14,6 @@ import (
 	"text/template"
 	"time"
 
-	v0 "github.com/authzed/authzed-go/proto/authzed/api/v0"
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 	"github.com/authzed/authzed-go/proto/authzed/api/v1alpha1"
 	"github.com/jackc/pgtype"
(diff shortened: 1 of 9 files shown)


e2e/newenemy/newenemy_test.go
```diff
@@ -64,10 +88,11 @@ const (
 var (
 	maxIterations = flag.Int("max-iterations", 1000, "iteration cap for statistic-based tests (0 for no limit)")
 
-	schemaTpl       = template.Must(template.New("schema").Parse(schemaText))
-	schemaAllowTpl  = template.Must(template.New("schema_allow").Parse(schemaAllowAllText))
-	objIdGenerator  = generator.NewUniqueGenerator(objIDRegex)
-	prefixGenerator = generator.NewUniqueGenerator(namespacePrefixRegex)
+	schemaTpl         = template.Must(template.New("schema").Parse(schemaText))
+	schemaOriginalTpl = template.Must(template.New("schema_orig").Parse(schemaOriginalText))
```
jakedt, line 92: Is this being used still?

jakedt (dismissed):

jakedt (approved):

PR #611, opened 2022-05-20
prevent crdb from crashing in e2e tests

ecordell:
#607 updated the e2e tests to work against #581 

But after that change, there seems to be some likelihood that cockroach will crash in the middle of the test and the step will time out.

We're doing a lot of "bad" things with cockroach in a resource constrained environment, so I'm using this PR to test out settings that keep CRDB running without breaking the test in other ways.

### DIFF
--- a/e2e/cockroach/cockroach.go
+++ b/e2e/cockroach/cockroach.go
@@ -23,6 +23,7 @@ type Node struct {
 	Httpaddr  string
 	ID        string
 	MaxOffset time.Duration
+	Cancel    context.CancelFunc
 	// only available after Start()
 	pid  int
 	conn *pgx.Conn
@@ -37,7 +38,7 @@ func (c *Node) Start(ctx context.Context) error {
 	cmd := []string{
 		"./cockroach",
 		"start",
-		"--store=node" + c.ID,
+		"--store=type=mem,size=640MiB",
 		"--logtostderr",
 		"--insecure",
 		"--listen-addr=" + c.Addr,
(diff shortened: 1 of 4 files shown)


e2e/cockroach/cockroach.go
```diff
@@ -46,10 +47,20 @@ func (c *Node) Start(ctx context.Context) error {
 		"--max-offset=" + c.MaxOffset.String(),
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	c.Cancel = cancel
 	c.pid, err = e2e.GoRun(ctx, logfile, logfile, cmd...)
 	return err
 }
 
+func (c *Node) Stop() error {
```
ecordell, line 56: changed to match the spicedb interface

e2e/newenemy/newenemy_test.go
```diff
@@ -276,11 +297,19 @@ func iterationsForHighConfidence(samples []int) (iterations int) {
...
 
+var goodNs *NamespaceNames
+
 func checkDataNoNewEnemy(ctx context.Context, t testing.TB, slowNodeID int, crdb cockroach.Cluster, spicedb spice.Cluster, maxAttempts int, exitEarly bool) (bool, int) {
-	prefix := namespacesForNode(ctx, t, crdb[1].Conn(), spicedb[0].Client().V1Alpha1().Schema(), slowNodeID)
-	t.Log("filling with data to span multiple ranges for prefix", prefix)
 	objIDGenerator := generator.NewUniqueGenerator(objIDRegex)
-	fill(ctx, t, spicedb[0].Client().V1().Permissions(), prefix, objIDGenerator, 100, 100)
+	var prefix NamespaceNames
+	if goodNs == nil {
```
ecordell, line 305: if we get reversed timestamps in a namespace, it seems to much more likely to get subsequent reversed timestamps with the same namespaces, so we just keep using a good one until it doesn't work anymore

e2e/cockroach/cockroach.go
```diff
@@ -37,7 +38,7 @@ func (c *Node) Start(ctx context.Context) error {
 	cmd := []string{
 		"./cockroach",
 		"start",
-		"--store=node" + c.ID,
+		"--store=type=mem,size=640MiB",
```
ecordell, line 41: this seems to have had the largest effect; I haven't seen a failure since changing this

jakedt (approved):

PR #612, opened 2022-05-21
support xDS as a dispatch resolver option

ecordell:
I haven't confirmed this with an xDS setup, but the client-side parsing seems to correctly accept `xds://` addresses and look for the xDS bootstrap file


josephschorr (approved):
LGTM

PR #614, opened 2022-05-24
Move the cache implementation behind an interface

josephschorr:
To allow for different implementations based on platform

### DIFF
--- a/pkg/cache/cache_ristretto.go
+++ b/pkg/cache/cache_ristretto.go
@@ -1,0 +1,32 @@
+//go:build !wasm
+// +build !wasm
+
+package cache
+
+import (
+	"github.com/dgraph-io/ristretto"
+)
+
+// NewCache creates a new ristretto cache from the given config.
+func NewCache(config *Config) (Cache, error) {
+	cache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters: config.NumCounters,
+		MaxCost:     config.MaxCost,
+		BufferItems: config.BufferItems,
+		Metrics:     config.Metrics,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return wrapped{cache}, nil
+}
+
+type wrapped struct {
+	*ristretto.Cache
+}
+
+func (w wrapped) GetMetrics() Metrics {
+	return w.Cache.Metrics
+}
+
+var _ Cache = &wrapped{}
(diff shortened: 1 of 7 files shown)


pkg/cmd/server/cacheconfig.go
```diff
@@ -24,7 +25,7 @@ const (
 )
 
 // Complete converts the cache config into a ristretto cache config.
```
jakedt, line 27: Fix the comment.

pkg/cache/cache_ristretto.go
```diff
@@ -0,0 +1,30 @@
+//go:build !wasm
+// +build !wasm
+
+package cache
+
+import (
+	"github.com/dgraph-io/ristretto"
+)
+
+// NewCache creates a new ristretto cache from the given config.
```
jakedt, line 10: Should we stick to the "return structs accept interfaces" mantra here?

pkg/cache/cache_ristretto.go
```diff
@@ -0,0 +1,30 @@
...
+	return wrapped{cache}, nil
+}
+
+type wrapped struct {
+	*ristretto.Cache
+}
+
+func (w wrapped) GetMetrics() Metrics {
+	return w.Cache.Metrics
+}
```
jakedt, line 30: Add an assertion that `wrapped` implements the interface.

jakedt (changes requested):

pkg/cache/cache_ristretto.go
josephschorr: We can, but it does mean that `NewCache`'s signature will change based on the platform, which may cause issues in certain compilation targets

josephschorr:
Updated

pkg/cache/cache_ristretto.go
jakedt: I don't get it. In this case it's always going to be a ristretto cache coming out of `cache_ristretto`. The method should probably be called `NewRistrettoCache`, and it will be totally missing on the wasm arch. If there were a wasm compatible version it would be called `NewHashmapCache` or something, and its version of `wrapped` would be specific to that cache impl.

jakedt (changes requested):

pkg/cache/cache_ristretto.go
josephschorr: > If there were a wasm compatible version it would be called NewHashmapCache or something

Actually, no. The whole point is that it is `NewCache` regardless of platform, so that when it is invoked outside of the package, the caller does not need to know on which platform the code is operating. Otherwise, it would fail to compile.

PR #615, opened 2022-05-24
Add dispatch ready, health check integration

samkim:
* Add `Ready` to dispatcher interface
* Check dispatch ready status when setting grpc server health status


internal/services/server.go
```diff
@@ -62,4 +69,35 @@ func RegisterGrpcServices(
...
+	go checkDispatcherServicesReady(dispatch, healthSrv, v1.PermissionsService_ServiceDesc.ServiceName, OverallServerHealthCheckKey)
+}
+
+// checkDispatcherServicesReady waits for the dispatcher to become ready before setting the health check status for the given services
+func checkDispatcherServicesReady(dispatch dispatch.Dispatcher, healthSrv *grpcutil.AuthlessHealthServer, services ...string) {
+	backoffInterval := backoff.NewExponentialBackOff()
+	// Run immediately for the initial check
+	ticker := time.After(0)
+
+	for {
```
ecordell, line 82: there's also https://pkg.go.dev/google.golang.org/grpc#ClientConn.WaitForStateChange which you could use to detect immediately when there's a change in connection state?

josephschorr: +1 to both points

internal/dispatch/remote/cluster.go
```diff
@@ -122,6 +124,12 @@ func (cr *clusterDispatcher) Close() error {
 }
 
 // Always verify that we implement the interfaces
```
ecordell, line 126:
```suggestion
// Ready returns whether the underlying dispatch connection is available
```

ecordell (approved):
LGTM

one comment nit

josephschorr (approved):
LGTM

josephschorr (approved):

PR #618, opened 2022-05-26
Fully remove support for the V0 ACL and Schema API

josephschorr:
The V0 API was deprecated on March 14, 2022. This PR removes all support for it, leaving only the developer API which happens to still live in the V0 namespace. A followup PR will remove the api endpoints themselves from the github.com/authzed/api repository


internal/testserver/server.go
```diff
@@ -68,5 +68,5 @@ func NewTestServer(require *require.Assertions,
 			require.NoError(conn.Close())
 		}
 		cancel()
-	}, revision
+	}, ds, revision
```
ecordell, line 71: is this only for the deletenamespace calls?

pkg/schemadsl/generator/generator_test.go
```diff
@@ -92,7 +92,9 @@ func TestGenerator(t *testing.T) {
 			"legacy relation",
 			namespace.Namespace("foos/test",
```
ecordell, line 93: why not just delete the test case?

internal/services/v1alpha1/schema_test.go
```diff
@@ -288,8 +258,14 @@ func TestSchemaReadDeleteAndFailWrite(t *testing.T) {
 	require.NoError(t, err)
 
 	// Issue a delete out of band for the namespace.
-	_, err = v0client.DeleteConfigs(context.Background(), &v0.DeleteConfigsRequest{
-		Namespaces: requestedObjectDefNames,
+	_, err = ds.ReadWriteTx(context.Background(), func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
```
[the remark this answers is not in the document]
josephschorr: The `out of band` is important: we need to be able to delete the namespaces without it going through the normal flow

internal/testserver/server.go
josephschorr: Yes, but other callers might need it too

pkg/schemadsl/generator/generator_test.go
josephschorr: Because it is still valid for those who might still have schema-less namespaces: we should still convert (so they can fix the problem), and that conversion should still be tested.

ecordell (approved):
LGTM

PR #638, opened 2022-06-02
Skip checking and redispatch in reachability when seeing duplicates

josephschorr:
Note that this just skips within the current dispatch, not globally (that should be handled by cache)

### DIFF
--- a/internal/graph/reachableresources.go
+++ b/internal/graph/reachableresources.go
@@ -3,6 +3,7 @@ package graph
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/scylladb/go-set/strset"
 	"github.com/shopspring/decimal"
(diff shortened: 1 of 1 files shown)


internal/graph/reachableresources.go
```diff
@@ -36,11 +37,22 @@ type ValidatedReachableResourcesRequest struct {
 	Revision decimal.Decimal
 }
 
+type syncONRSet struct {
+	items sync.Map
+}
+
+func (s *syncONRSet) Add(onr *core.ObjectAndRelation) bool {
+	key := tuple.StringONR(onr)
+	_, existed := s.items.LoadOrStore(key, true)
```
jzelinskie, line 46: I think the empty struct instead of `true` trick to save memory should work for sync.Map, too.

josephschorr: Looks like `nil` works too... maybe I should just use that?

jzelinskie: if the implementation of sync.Map uses a `map[any]any`, the go compiler will optimize `struct{}{}` to consume 0 bytes vs nil which is probably going to be sizeof(pointer)

josephschorr: Updated

PR #646, opened 2022-06-07
propagate context to migration functions

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/347

# What

Introduces `context.Context` argument into the signature of the migration function

# Why

So that long running migrations can be timed out via `spicedb migrate` CLI command.

# Considerations

- The new flag is named `migration-timeout` - open to suggestions
- No timeout is defined by default to retain original behaviour
- Reflection is being used to for the migration function. An abstraction could be explored but decided to punt on it to keep the PR moderately sized
- `context.Context` argument was added to `migrate.Driver` as the PSQL/CRDB driver expect it. This would API breakage.

## Example

example CLI output

### DIFF
--- a/pkg/cmd/migrate.go
+++ b/pkg/cmd/migrate.go
@@ -25,7 +25,7 @@ func RegisterMigrateFlags(cmd *cobra.Command) {
 	cmd.Flags().String("datastore-spanner-credentials", "", "path to service account key credentials file with access to the cloud spanner instance")
 	cmd.Flags().String("datastore-spanner-emulator-host", "", "URI of spanner emulator instance used for development and testing (e.g. localhost:9010)")
 	cmd.Flags().String("datastore-mysql-table-prefix", "", "prefix to add to the name of all mysql database tables")
-	cmd.Flags().Duration("migration-timeout", 0*time.Second, "defines a timeout for the execution of the migration, disabled by default")
+	cmd.Flags().Duration("migration-timeout", 1*time.Hour, "defines a timeout for the execution of the migration, set to 1 hour by default")
 }
 
 func NewMigrateCommand(programName string) *cobra.Command {
@@ -100,12 +100,8 @@ func migrateRun(cmd *cobra.Command, args []string) error {
 	targetRevision := args[0]
 
 	log.Info().Str("targetRevision", targetRevision).Msg("running migrations")
-	ctx := cmd.Context()
-	if timeout > 0 {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(cmd.Context(), timeout)
-		defer cancel()
-	}
+	ctx, cancel := context.WithTimeout(cmd.Context(), timeout)
+	defer cancel()
 	if err := manager.Run(ctx, migrationDriver, targetRevision, migrate.LiveRun); err != nil {
 		log.Fatal().Err(err).Msg("unable to complete requested migrations")
 	}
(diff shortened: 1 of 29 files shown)


pkg/migrate/migrate.go
```diff
@@ -124,7 +124,7 @@ func (m *Manager) Run(ctx context.Context, driver Driver, throughRevision string
 
 			log.Info().Str("from", migrationToRun.replaces).Str("to", migrationToRun.version).Msg("migrating")
 
-			in := []reflect.Value{reflect.ValueOf(driver)}
+			in := []reflect.Value{reflect.ValueOf(ctx), reflect.ValueOf(driver)}
```
josephschorr, line 127: Now that we have generics, we could in theory just make this a generic `func` and remove the use of reflection

pkg/cmd/migrate.go
```diff
@@ -97,11 +100,17 @@ func migrateRun(cmd *cobra.Command, args []string) error {
 	targetRevision := args[0]
 
 	log.Info().Str("targetRevision", targetRevision).Msg("running migrations")
-	if err := manager.Run(context.Background(), migrationDriver, targetRevision, migrate.LiveRun); err != nil {
+	ctx := cmd.Context()
+	if timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithDeadline(cmd.Context(), time.Now().Add(timeout))
```
jakedt, line 106: There is a `context.WithTimeout` that does this add already.

pkg/cmd/migrate.go
```diff
@@ -24,6 +25,7 @@ func RegisterMigrateFlags(cmd *cobra.Command) {
 	cmd.Flags().String("datastore-spanner-credentials", "", "path to service account key credentials file with access to the cloud spanner instance")
 	cmd.Flags().String("datastore-spanner-emulator-host", "", "URI of spanner emulator instance used for development and testing (e.g. localhost:9010)")
 	cmd.Flags().String("datastore-mysql-table-prefix", "", "prefix to add to the name of all mysql database tables")
+	cmd.Flags().Duration("migration-timeout", 0*time.Second, "defines a timeout for the execution of the migration, disabled by default")
```
jakedt, line 28: Yeah just set it to something super long, like 1h or something.

jakedt (changes requested):
It would be super cool if we could get rid of the reflection by using generics like Joey suggests. This was always the most brittle part.

pkg/cmd/migrate.go
```diff
@@ -97,11 +100,17 @@ func migrateRun(cmd *cobra.Command, args []string) error {
 	targetRevision := args[0]
 
 	log.Info().Str("targetRevision", targetRevision).Msg("running migrations")
-	if err := manager.Run(context.Background(), migrationDriver, targetRevision, migrate.LiveRun); err != nil {
+	ctx := cmd.Context()
+	if timeout > 0 {
```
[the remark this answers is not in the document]
vroldanbet: Sounds like Jake is also onboard - I'll go with the 1h default 👍🏻

pkg/migrate/migrate.go
vroldanbet: That's a great idea and the perfect excuse for me to start learning the syntax 😅

pkg/cmd/migrate.go
vroldanbet: oh! how convenient +1

pkg/migrate/migrate.go
```diff
@@ -2,9 +2,7 @@ package migrate
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"reflect"
```
jakedt, line 7: 👍

jakedt (approved):

Issue #648, opened 2022-06-08
run migration operations transactionally


vroldanbet:
Surfaced as feedback from https://github.com/authzed/spicedb/pull/646

The execution of the various migration operations and changes to the migration metadata tables should all happen transactionally. Currently they happen in different transactions, which means that any failure would leave the database in an intermediate state.

PR #651, opened 2022-06-10
introduce support to run migrations transactionally

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/648

# What

makes it possible to run all the database operations needed for a migration transactionally

# Why

The `migrate.Manager` was responsible to invoke `driver.WriteVersion`. However when this was invoked by the `migrate.Manager` it happened in its own transaction. The process could fail before the migration version is written to the database, leaving the system in an inconsistent state that would require manual intervention.

# How

The `Driver.WriteVersion` interface method is deprecated in favour of `Conn()` and the interfaces is parameterized with generics. This avoids the need to introduce an abstraction to connection pools for each `MigrationFunc` of each datastore, which get the full access to the connection handler API. This is particularly important given how different each datastore driver is.

The `migrate.Manager` is no longer responsible for invoking `WriteVersion`: the `MigrationFunc`. This is relevant given that each migration is different, and some cannot be executed transactionally. Some examples:
- Spanner: DML and DDL operations cannot run transactionally
- CRDB: SET CLUSTER SETTINGS cannot be run inside a transaction
- PostgreSQL: CREATE INDEX CONCURRENTLY cannot run inside a transaction block (SQLSTATE 25001)

This is not as dramatic as it may seem: the `migrate.Manager`: the driver is ultimately the one that knows how to write the version to the database, and the manager was just imposing a contract to orchestrate it. We can ensure the same effect by making the manager ensure that after a `MigrationFunc` is invoked the datastore is indeed reporting the expected version through the `Version` method - this was introduced in https://github.com/authzed/spicedb/pull/651/commits/06bf4d0182f64869ecd8f2788e17c87008969c18. We bake the cake and eat it.

# Notes

Goland is _struggling_ with generics. You may see some errors popping up in places where generics are used. Don't mind those, they actually compile.


vroldanbet:
Closed in favour of https://github.com/authzed/spicedb/pull/655

PR #652, opened 2022-06-10
Cleanup the core messages now that v0 is gone

josephschorr:
- Renames object_and_relation on the RelationTuple to resource_and_relation
- Renames user on the RelationTuple to subject
- Removes the user_oneof, as it is no longer needed
- Changes DirectUsersets into DirectSubjects
- Various small test fixes
- Remove unneeded conversion methods and change to manually written for the remaining


proto/internal/core/v1/core.proto
```diff
@@ -7,23 +7,11 @@ import "google/protobuf/any.proto";
...
-  /** object_and_relation is the resource for the tuple */
-  ObjectAndRelation object_and_relation = 1 [ (validate.rules).message.required = true ];
-
-  /** user is the subject for the tuple */
-  User user = 2 [ (validate.rules).message.required = true ];
+  /** resource_and_relation is the resource for the tuple */
+  ObjectAndRelation resource_and_relation = 1 [ (validate.rules).message.required = true ];
+
+  /** subject is the subject for the tuple */
+  ObjectAndRelation subject = 2 [ (validate.rules).message.required = true ];
```
vroldanbet, line 14: This would seem like a deliberate API breakage, and presumably we are OK with that? Would this cause downtime to clients as the server with the new proto definition is rolled out?

internal/datastore/common/sql.go
```diff
@@ -247,21 +247,16 @@ func NewPGXExecutor(txSource TxFactory) ExecuteQueryFunc {
...
 		for rows.Next() {
 			nextTuple := &core.RelationTuple{
-				ObjectAndRelation: &core.ObjectAndRelation{},
-				User: &core.User{
-					UserOneof: &core.User_Userset{
-						Userset: &core.ObjectAndRelation{},
-					},
-				},
+				ResourceAndRelation: &core.ObjectAndRelation{},
+				Subject:             &core.ObjectAndRelation{},
```
vroldanbet, line 251: If a `Subject` does also have a `Relation`, which isn't it named `SubjectAndRelation`, or perhaps why not naming `ResourceAndRelation` just `Relation`?

vroldanbet (commented):
This is a nice cleanup, thank you! ✨ Left some comments about naming.

What's not clear to me is the impact to the proto backward compatibility. Because the type of a field is changed it would look like a breaking change - which seems corroborated by a CI error. What is the plan around that? How can we help folks transition to the newer version without downtime?

internal/datastore/crdb/watch.go
```diff
@@ -113,19 +113,15 @@ func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Rev
 
 			oneChange := &core.RelationTupleUpdate{
```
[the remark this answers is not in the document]
josephschorr: We do have some helper methods defined in various places, so we could in theory grab one, but I like that everything is explicitly named here

internal/graph/expand.go
```diff
@@ -59,23 +59,23 @@ func (ce *ConcurrentExpander) expandDirect(
...
 		if err != nil {
 			resultChan <- expandResultError(NewExpansionFailureErr(err), emptyMetadata)
 			return
 		}
 		defer it.Close()
 
-		var foundNonTerminalUsersets []*core.User
-		var foundTerminalUsersets []*core.User
+		var foundNonTerminalUsersets []*core.ObjectAndRelation
+		var foundTerminalUsersets []*core.ObjectAndRelation
```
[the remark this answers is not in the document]
josephschorr: The datastores still do, so I'm unsure whether we should change it (we probably should)

proto/internal/core/v1/core.proto
josephschorr: Referenced below

internal/datastore/common/sql.go
josephschorr: The real change would be to break the relation out from both, but that is a major refactor, so I left it using ObjectAndRelation. Since I did so, I wanted to be clear that the relation for the operation is in the first ONR

vroldanbet:
>We don't store these protos; they are only used for internal dispatching, so the only real issue with compatibility is during deployment, where dispatches might (briefly) fail. We could back-compat support for that case, or just say "yes, this will have an issue for a short period of time"

As discussed in standup: understood! I somehow thought those protos where part of the public API.

The problem you described perhaps warrants a different conversation on how can we make rolling upgrades that are isolated from each other? Pperhaps `kuberesolver` could be improved to differentiate also on deployments. Depending on whoch the rolling upgrade strategy is configured, it could lead to multiple hashring reconfigurations and thus inefficient usage of the caches during the rollout.

PR #655, opened 2022-06-17
introduce support to run migrations atomically

jakedt:
Closes https://github.com/authzed/spicedb/issues/648

# What

makes it possible to run all the database operations needed for a migration atomically

# Why

The `migrate.Manager` was responsible to invoke `driver.WriteVersion`. However when this was invoked by the `migrate.Manager` it happened in its own transaction. The process could fail before the migration version is written to the database, leaving the system in an inconsistent state that would require manual intervention.

# How

A new method `Conn()` is added to the `Driver` interface and parameterized with generics. This avoids the need to introduce an abstraction to connection pools for each `MigrationFunc` of each datastore, which get the full access to the connection handler API. This is particularly important given how different each datastore driver is.

Migrations will be composed of a part executed in a transaction, another where its provided with the database connection handler. The `migrate.Manager` will execute first the part of the migration that cannot be run in a transaction, and then the function that is to be run in a transaction, alongside a call to  `WriteVersion` to make sure both are execute atomically. The separation is relevant given that some operations cannot be run in a transaction. Some examples:
- Spanner: DML and DDL operations cannot run in a transaction
- CRDB: SET CLUSTER SETTINGS cannot be run inside a transaction
- PostgreSQL: CREATE INDEX CONCURRENTLY cannot run inside a transaction block (SQLSTATE 25001)

# Notes

Goland is _struggling_ with generics. You may see some errors popping up in places where generics are used. If that's your case, update to the latest EAP version.


internal/datastore/crdb/migrations/driver.go
```diff
@@ -60,10 +62,24 @@ func (apd *CRDBDriver) Version(ctx context.Context) (string, error) {
...
-// WriteVersion overwrites the value stored to track the version of the
-// database schema.
-func (apd *CRDBDriver) WriteVersion(ctx context.Context, version, replaced string) error {
-	result, err := apd.db.Exec(ctx, queryWriteVersion, version, replaced)
+// Conn returns the underlying pgx.Conn instance for this driver
+func (apd *CRDBDriver) Conn() *pgx.Conn {
+	return apd.db
+}
+
+func (apd *CRDBDriver) RunTx(ctx context.Context, f migrate.TxMigrationFunc[pgx.Tx]) error {
```
josephschorr, line 70: Add doc comments for all exported functions

internal/datastore/crdb/migrations/manager.go
```diff
@@ -1,6 +1,12 @@
 package migrations
 
-import "github.com/authzed/spicedb/pkg/migrate"
+import (
+	"github.com/authzed/spicedb/pkg/migrate"
+
+	"github.com/jackc/pgx/v4"
+)
+
+var noNonatomicMigration migrate.MigrationFunc[*pgx.Conn]
```
josephschorr, line 9: Maybe `emptyMigration` or `noMigration`? The current name is more descriptive but is harder to parse

internal/datastore/mysql/migrations/driver.go
```diff
@@ -98,16 +100,52 @@ func (driver *MySQLDriver) Version(ctx context.Context) (string, error) {
...
+		ctx,
+		driver.db,
+		&sql.TxOptions{Isolation: sql.LevelSerializable},
+		func(tx *sql.Tx) error {
+			return f(ctx, TxWrapper{tx, driver.tables})
+		},
+	)
+}
+
+// BeginTxFunc is a polyfill for database/sql which does a closure style transaction lifecycle.
```
josephschorr, line 118: `which does` -> `which implements`

internal/datastore/mysql/migrations/driver.go
```diff
@@ -98,16 +100,52 @@ func (driver *MySQLDriver) Version(ctx context.Context) (string, error) {
...
+// The underlying transaction is aborted if the supplied function returns an error.
+// The underlying transaction is committed if the supplied function returns nil.
+func BeginTxFunc(ctx context.Context, db *sql.DB, txOptions *sql.TxOptions, f func(*sql.Tx) error) error {
+	tx, err := db.BeginTx(ctx, txOptions)
+	if err != nil {
+		return err
+	}
+	defer LogOnError(ctx, tx.Rollback)
```
josephschorr, line 126: Is it okay to call Rollback after Commit has completed?

internal/datastore/mysql/migrations/manager.go
```diff
@@ -10,21 +11,33 @@ import (
...
+
 	migrationNameRe = regexp.MustCompile(migrationNamePattern)
 
 	// Manager is the singleton migration manager instance for MySQL
-	Manager = migrate.NewManager[*MySQLDriver]()
+	Manager = migrate.NewManager[*MySQLDriver, Wrapper, TxWrapper]()
 )
 
-type mySQLMigrationFunc = migrate.MigrationFunc[*MySQLDriver]
+// Wrapper makes it possible to forward the state needed for MySQL MigrationFunc to run
```
josephschorr, line 22: Can you clarify what is meant by "forward the state"?

internal/datastore/mysql/migrations/migrations_test.go
```diff
@@ -12,8 +12,8 @@ import (
 
 func TestMySQLMigrationsWithUnsupportedPrefix(t *testing.T) {
 	req := require.New(t)
-	err := registerMigration("888", "", func(ctx context.Context, d *MySQLDriver) error {
+	err := registerMigration("888", "", func(ctx context.Context, d Wrapper) error {
 		return nil
-	})
+	}, nil)
```
josephschorr, line 17: maybe an `emptyTxMigration` here?

pkg/migrate/migrate.go
```diff
@@ -20,49 +20,67 @@ var (
...
+	// Tx returns a transaction for to be used by one or more TxMigrationFunc
+	RunTx(context.Context, TxMigrationFunc[T]) error
 
 	// Close frees up any resources in use by the driver.
 	Close(ctx context.Context) error
 }
 
-type migration[T Driver] struct {
+// MigrationFunc is a function that executes in the context of a specific database connection handler.
+// The version string represents the version of this migration to run, and replaced
```
josephschorr, line 45: Update doc comment here and below

pkg/migrate/migrate_test.go
```diff
@@ -14,28 +14,40 @@ func (*fakeDriver) Version(ctx context.Context) (string, error) {
...
 	ctx, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(1*time.Millisecond))
-	m := NewManager[Driver]()
+	m := NewManager[Driver[fakeConnPool, fakeTx], fakeConnPool, fakeTx]()
 
-	err := m.Register("1", "", func(ctx context.Context, driver Driver) error {
+	err := m.Register("1", "", func(ctx context.Context, conn fakeConnPool) error {
 		cancelFunc()
 		return nil
-	})
+	}, nil)
```
josephschorr, line 45: ditto on maybe `noTxMigration`?

pkg/migrate/migrate_test.go
```diff
@@ -150,34 +162,43 @@ func TestIsHeadCompatible(t *testing.T) {
...
 		req.Equal(tc.expectError, err != nil, err)
 	}
 }
 
-var noMigrations = map[string]migration[Driver]{}
+func TestManagerEnsureVersionIsWritten(t *testing.T) {
+	req := require.New(t)
+	m := NewManager[Driver[fakeConnPool, fakeTx], fakeConnPool, fakeTx]()
+	err := m.Register("0", "", nil, nil)
+	req.NoError(err)
```
josephschorr, line 176: Can we add a check here that reads the version and compares?

internal/datastore/crdb/migrations/manager.go
jakedt: I'm using typed nil to make sure it's clear at the call site which type of migration is being skipped.

internal/datastore/mysql/migrations/driver.go
jakedt: Yes.

internal/datastore/crdb/migrations/driver.go
jakedt: This is not required by the linter, and is actually noisy if you're implementing an interface.

internal/datastore/crdb/migrations/manager.go
josephschorr: Yep, I'm just commenting on the name

internal/datastore/mysql/migrations/driver.go
```diff
@@ -98,16 +100,52 @@ func (driver *MySQLDriver) Version(ctx context.Context) (string, error) {
...
 
+func (driver *MySQLDriver) Conn() Wrapper {
+	return Wrapper{db: driver.db, tables: driver.tables}
+}
+
+func (driver *MySQLDriver) RunTx(ctx context.Context, f migrate.TxMigrationFunc[TxWrapper]) error {
+	return BeginTxFunc(
+		ctx,
+		driver.db,
+		&sql.TxOptions{Isolation: sql.LevelSerializable},
```
vroldanbet, line 111: Just calling out that this was _recently added_ in this PR. When `nil` is provided, the default isolation level is used.

vroldanbet (approved):
The changes look good, thanks for refining it with the separation of with/without transaction and keeping the responsibility to run `WriteVersion` in the `migrate.Manager` ✨

PR #656, opened 2022-06-21
watch TLS certs for changes

ecordell:


vroldanbet (dismissed):
Changes look good to me! Wondering if it would be feasible to add a test.

vroldanbet (approved):
Thanks for adding the tests!

PR #658, opened 2022-06-22
Return an error instead of calling panic() when debug.ReadBuildInfo() is unavailable

lmt-swallow:
This PR changes the behaviour of `internal/telemetry/metrics.go` to avoid `panic()` when debug.ReadBuildInfo() is unavailable as same as `pkg/releases/versions.go`.

As a background, I'm using Bazel and `rules_go` to build and integrate SpiceDB for some of my private projects. However, the bazel rule is not currently supporting `debug.BuildInfo`(https://github.com/bazelbuild/rules_go/issues/3090) as of now. Just avoiding the `panic()` at `internal/telemetry/metrics.go` helps!


lmt-swallow:

lmt-swallow:
recheck

josephschorr (approved):
LGTM

PR #663, opened 2022-06-24
Add a WebAssembly interface for invoking the SpiceDB dev package

josephschorr:
This will allow browser-based tooling to directly perform development operations, such as validation and testing


pkg/development/wasm/main.go
```diff
@@ -0,0 +1,70 @@
...
+func derr(derrs *development.DeveloperErrors) interface{} {
+	marshalled, err := json.Marshal(derrs)
+	if err != nil {
+		return rerr(err.Error())
+	}
+
+	return js.ValueOf([]interface{}{string(marshalled), nil})
+}
+
+func runDevelopmentOperations(this js.Value, args []js.Value) interface{} {
```
samkim, line 32: Document the expected shape of `args` and perhaps the supported operations of the 2nd arg since this is kind of like an interface between the js and the wasm side. Or a link to the example `wasm.html` could suffice since it has an example request json.

samkim (approved):
LGTM

PR #673, opened 2022-07-03
Garbage collect namespaces

### LINKED ISSUE #660: Add GC to the namespace_config table
Currently, when a docker image is started with a bootstrap file and the flag _--datastore-bootstrap-overwrite=true_, the _namespace_config_ table expands with every restart of the docker image. The deleted rows are not cleanup up by the GC, like it does for the _relation_tuple_ and _relation_tuple_transaction_ tables.  The GC should also clean the _namespace_config_ table.

jzelinskie:
Fixes #660

Most of this change is clean-up.

- Converts gc metrics into counters
- Introduces a type for the amounts collected
- Hoists logging to one central defer
- Slight refactor to separate concerns between the various gc methods
- Shares GC logic in a common package
- Adds an `id` column to namespace tables for MySQL and Postgres datastores ⚠️ **WARNING DB MIGRATION**

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -18,25 +18,25 @@ var (
 		Buckets:   []float64{0.01, 0.1, 0.5, 1, 5, 10, 25, 60, 120},
 	})
 
-	gcRelationshipsClearedCounter = prometheus.NewCounter(prometheus.CounterOpts{
+	gcRelationshipsCounter = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: "spicedb",
 		Subsystem: "datastore",
 		Name:      "gc_relationships_total",
-		Help:      "The number of relationships cleared by the datastore garbage collection.",
+		Help:      "The number of stale relationships deleted by the datastore garbage collection.",
 	})
 
-	gcTransactionsClearedCounter = prometheus.NewCounter(prometheus.CounterOpts{
+	gcTransactionsCounter = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: "spicedb",
 		Subsystem: "datastore",
 		Name:      "gc_transactions_total",
-		Help:      "The number of transactions cleared by the datastore garbage collection.",
+		Help:      "The number of stale transactions deleted by the datastore garbage collection.",
 	})
 
-	gcNamespacesClearedCounter = prometheus.NewCounter(prometheus.CounterOpts{
+	gcNamespacesCounter = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: "spicedb",
 		Subsystem: "datastore",
 		Name:      "gc_namespaces_total",
-		Help:      "The number of namespaces cleared by the datastore garbage collection.",
+		Help:      "The number of stale namespaces deleted by the datastore garbage collection.",
 	})
 )
 
@@ -45,9 +45,9 @@ var (
 func RegisterGCMetrics() error {
 	for _, metric := range []prometheus.Collector{
 		gcDurationHistogram,
-		gcRelationshipsClearedCounter,
-		gcTransactionsClearedCounter,
-		gcNamespacesClearedCounter,
+		gcRelationshipsCounter,
+		gcTransactionsCounter,
+		gcNamespacesCounter,
 	} {
 		if err := prometheus.Register(metric); err != nil {
 			return err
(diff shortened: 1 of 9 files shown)


internal/datastore/common/gc.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+
+	return nil
+}
+
+// GarbageCollector represents any datastore that supports external garbage
+// collection.
+type GarbageCollector interface {
+	IsReady(context.Context) (bool, error)
```
josephschorr, line 63: Add doc comments on interface exported methods

internal/datastore/common/gc.go
```diff
@@ -0,0 +1,156 @@
...
+	Now(context.Context) (time.Time, error)
+	TxIDBefore(context.Context, time.Time) (uint64, error)
+	DeleteBeforeTx(context.Context, uint64) (GarbageCollectedAmounts, error)
+}
+
+// GarbageCollectedAmounts is the collection of amounts of deletions from
+// an individual run of the garbage collector.
+type GarbageCollectedAmounts struct {
```
josephschorr, line 71: `GarbageCollectionCounts`?

internal/datastore/common/gc.go
```diff
@@ -0,0 +1,156 @@
...
+	Relationships int64
```
josephschorr, line 72: `RemovedRelationships`, etc?

internal/datastore/common/gc.go
```diff
@@ -0,0 +1,156 @@
...
+// performs garbage collection on the provided interval.
+func StartGarbageCollector(ctx context.Context, gc GarbageCollector, interval, window, timeout time.Duration) error {
+	log.Ctx(ctx).Info().
+		Dur("interval", interval).
+		Msg("datastore garbage collection worker started")
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Ctx(ctx).Info().
```
josephschorr, line 94: Formatting here and the other log lines is odd

internal/datastore/mysql/datastore_test.go
```diff
@@ -166,9 +166,9 @@ func GarbageCollectionTest(t *testing.T, ds datastore.Datastore) {
 	// Run GC at the transaction and ensure no relationships are removed.
 	mds := ds.(*Datastore)
 
-	relsDeleted, _, err := mds.collectGarbageForTransaction(ctx, uint64(writtenAt.IntPart()))
-	req.Zero(relsDeleted)
+	amounts, err := mds.DeleteBeforeTx(ctx, uint64(writtenAt.IntPart()))
```
josephschorr, line 169: Add a namespace test?

internal/datastore/mysql/migrations/zz_migration.0003_add_ns_config_id.go
```diff
@@ -0,0 +1,26 @@
+package migrations
+
+import "fmt"
+
+func dropNSConfigPK(t *tables) string {
+	return fmt.Sprintf(
+		`ALTER TABLE %s DROP PRIMARY KEY;`,
```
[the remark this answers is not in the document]
jzelinskie: The id gives us a field to select for the bulk deletion code.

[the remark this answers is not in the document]
josephschorr: I know; I'm asking what happens to the index from the current key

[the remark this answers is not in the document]
jzelinskie: I see what you're asking -- it drops the index, but the columns are left untouched. I just tested both postgres and mysql commands in containers on the CLI to confirm everything functions how we expect.

internal/datastore/common/gc.go
jzelinskie: the callsites all now read `removed.Relationships`, which I think is a good middle-ground for verbosity

internal/datastore/common/gc.go
jzelinskie: yeah, it's because the messages are long but the context and level are always short

internal/datastore/common/gc.go
```diff
@@ -97,7 +100,13 @@ func StartGarbageCollector(ctx context.Context, gc GarbageCollector, interval, w
 
 		case <-time.After(interval):
 			err := collect(gc, window, timeout)
-			if err != nil {
+			switch {
+			case err == nil:
+				break
+			case errors.As(err, &datastore.ErrReadOnly{}):
```
josephschorr, line 106: Shouldn't this shutdown the GC worker?

jzelinskie: Yeah, I wasn't sure if we should shutdown the worker or just continue to no-op.
I figured running it like this means that the lifecycle/cleanup stays consistent regardless of read-only.

josephschorr: It does, but at the same time it means a lot of extra debug logs... so I think we should just shut down the process entirely

jzelinskie: I could also just make it trace() instead of debug

josephschorr (approved):
LGTM

PR #691, opened 2022-07-08
Implement support for batch reachability

josephschorr:
This change drastically reduces the number of goroutines and SQL statements made during reachability by batching up subsets of subjects of the same type into single dispatched call(s).

Benchmarks:
```
name                                                        old time/op  new time/op 

Services/basic_lookup_of_view_for_a_user/cockroachdb/v1-10  28.6ms ± 0%  25.0ms ± 0%
Services/basic_lookup_of_view_for_a_user/mysql/v1-10        9.39ms ± 0%  7.44ms ± 0%
Services/basic_lookup_of_view_for_a_user/postgres/v1-10     6.94ms ± 0%  6.69ms ± 0%  
Services/basic_lookup_of_view_for_a_user/spanner/v1-10      9.65ms ± 0%  7.19ms ± 0% 
Services/recursively_through_groups/cockroachdb/v1-10       93.1ms ± 0%  73.0ms ± 0% 
Services/recursively_through_groups/mysql/v1-10             26.6ms ± 0%  23.8ms ± 0%
Services/recursively_through_groups/postgres/v1-10          19.4ms ± 0%  17.3ms ± 0% 
Services/recursively_through_groups/spanner/v1-10           37.0ms ± 0%  34.0ms ± 0%
Services/recursively_through_wide_groups/cockroachdb/v1-10   1.80s ± 0%   0.33s ± 0%
Services/recursively_through_wide_groups/mysql/v1-10         663ms ± 0%   109ms ± 0%
Services/recursively_through_wide_groups/postgres/v1-10      395ms ± 0%    60ms ± 0%
Services/recursively_through_wide_groups/spanner/v1-10       11.1s ± 0%    1.7s ± 0% 
```

> **Note**
> The benchmarks were generated by running the various datastores in containers on a local M1, which likely explains why Postgres is so much faster: it is running under native `arm`. MySQL, CRDB, and Spanner are `amd64` images and therefore likely much slower than a real-world installation

> **Note**
> Numbers have been updated after discussed changes


internal/datastore/memdb/readonly.go
```diff
@@ -92,31 +93,11 @@ func (r *memdbReader) ReverseQueryRelationships(
 
 	queryOpts := options.NewReverseQueryOptionsWithOptions(opts...)
 
-	var bestIterator memdb.ResultIterator
-	if queryOpts.ResRelation != nil {
-		bestIterator, err = tx.Get(
-			tableRelationship,
-			indexSubjectAndResourceRelation,
```
jakedt, line 99: Can we drop some of these indices if they're no longer being selected?

internal/datastore/common/sql.go
```diff
@@ -104,6 +104,59 @@ func (sqf SchemaQueryFilterer) FilterToRelation(relation string) SchemaQueryFilt
 	return sqf
 }
 
+// FilterWithSubjectsFilter returns a new SchemaQueryFilterer that is limited to resources with
+// subjects that match the specified filter.
+func (sqf SchemaQueryFilterer) FilterWithSubjectsFilter(filter datastore.SubjectsFilter) SchemaQueryFilterer {
```
jakedt, line 109: Add a unit test for this method.

internal/graph/reachableresources.go
```diff
@@ -196,26 +147,40 @@ func (crr *ConcurrentReachableResources) lookupRelationEntrypoint(ctx context.Co
 		return err
 	}
 
-	isDirectAllowed, err := relTypeSystem.IsAllowedDirectRelation(relationReference.Relation, req.Subject.Namespace, req.Subject.Relation)
+	// Build the list of subjects to lookup based on the type information available.
+	isDirectAllowed, err := relTypeSystem.IsAllowedDirectRelation(relationReference.Relation, req.SubjectRelation.Namespace, req.SubjectRelation.Relation)
```
jakedt, line 151: Should probably break this line up.

internal/graph/reachableresources.go
```diff
@@ -226,31 +191,114 @@ func (crr *ConcurrentReachableResources) lookupRelationEntrypoint(ctx context.Co
...
+		})
+	})
+
+	return nil
+}
+
+func (crr *ConcurrentReachableResources) chunkedRedispatch(it datastore.RelationshipIterator, handler func(resourceIdsFound []string) error) error {
+	chunkIndex := 0
+
+	for {
```
jakedt, line 208: On first glance I immediately thought this would exceed the length of `dispatchChunkSizes` before I saw the compensation code in the loop body. Can this be rewritten maybe to use a monotonically increasing `chunkIndex` and then use a size function or something?

internal/graph/reachableresources.go
```diff
@@ -226,31 +191,114 @@ func (crr *ConcurrentReachableResources) lookupRelationEntrypoint(ctx context.Co
...
+		chunkSize := dispatchChunkSizes[chunkIndex]
+		resourceIdsFound := make([]string, 0, chunkSize)
+		for i := 0; i < chunkSize; i++ {
+			tpl := it.Next()
+			if tpl == nil {
```
jakedt, line 213: I think you still need to check the error case here.

jakedt (changes requested):

internal/dispatch/caching/caching.go
```diff
@@ -78,9 +78,15 @@ func NewCachingDispatcher(
 		log.Info().Int64("numCounters", cacheConfig.NumCounters).Str("maxCost", humanize.Bytes(uint64(cacheConfig.MaxCost))).Msg("configured caching dispatcher")
 	}
 
-	cache, err := cache.NewCache(cacheConfig)
-	if err != nil {
-		return nil, fmt.Errorf(errCachingInitialization, err)
+	var built cache.Cache
```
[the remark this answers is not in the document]
josephschorr: I had to rename it because it conflicted with the `cache` package name. Changed to `cacheImpl`

pkg/datastore/test/tuples.go
```diff
@@ -694,3 +694,11 @@ func ConcurrentWriteSerializationTest(t *testing.T, tester DatastoreTester) {
 	require.NoError(g.Wait())
 	require.Less(time.Since(startTime), 10*time.Second)
 }
+
+func onrToSubjectsFilter(onr *core.ObjectAndRelation) datastore.SubjectsFilter {
+	return datastore.SubjectsFilter{
+		Namespace:  onr.Namespace,
+		SubjectIds: []string{onr.ObjectId},
```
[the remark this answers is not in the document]
josephschorr: Would you prefer me to basically `oneof` the multiple subjects and multiple relations, or simply document that it is an `OR` instead of an `AND`?

internal/datastore/common/sql.go
josephschorr: Where do we have the unit tests for these methods defined? I don't see one for `FilterToSubjectFilter` either

josephschorr:
Updated

internal/datastore/common/sql.go
jakedt: I don't know if you're just being cheeky, but we don't have any unit tests for these helper sql methods yet. We should add them when they're purely functional though. They would go in `sql_test.go`.

pkg/datastore/test/tuples.go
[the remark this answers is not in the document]
jakedt: Is a oneof in the spirit of how it's intended to be used? That sounds pretty good to me if that's an option.

internal/datastore/common/sql.go
josephschorr: No, I was just making sure I didn't miss them. I'll add a bunch.

pkg/datastore/test/tuples.go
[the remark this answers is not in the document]
josephschorr: It is basically an AND between sets of relations and subject IDs: if you specify multiple subject IDs *and* multiple relations, then it'll find any relationships with any one of those relations, and any relationships with any one of those subject IDs. I could reduce the interface to allowing multiple Subject IDs *or* multiple Relations, but not both to remove that case.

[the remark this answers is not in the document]
josephschorr: Changed to use basically a builder pattern

internal/datastore/common/sql.go
josephschorr: Added

josephschorr:
Updated

internal/services/benchmark_test.go
```diff
@@ -0,0 +1,109 @@
...
+	title    string
+	fileName string
+	runner   runner
+}
+
+func BenchmarkServices(b *testing.B) {
+	bts := []benchmarkTest{
+		{
+			"basic lookup of view for a user",
+			"basicrbac.yaml",
```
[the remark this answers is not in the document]
jakedt: did this happen?

internal/dispatch/caching/caching.go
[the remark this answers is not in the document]
jakedt: how about `cacheInst`? `cacheImpl` seems to mean code to me.

internal/graph/reachableresources.go
jakedt: Now it can be re-written as a for loop.

internal/dispatch/dispatch.go
```diff
@@ -121,5 +122,13 @@ func ExpandRequestToKey(req *v1.DispatchExpandRequest) string {
...
 // ReachableResourcesRequestToKey converts a reachable resources request into a cache key
 func ReachableResourcesRequestToKey(req *v1.DispatchReachableResourcesRequest) string {
-	return fmt.Sprintf("%s//%s#%s@%s@%s", reachableResourcesPrefix, req.ObjectRelation.Namespace, req.ObjectRelation.Relation, tuple.StringONR(req.Subject), req.Metadata.AtRevision)
+	return fmt.Sprintf("%s//%s#%s@%s#%s:[%s]@%s",
+		reachableResourcesPrefix,
+		req.ResourceRelation.Namespace,
+		req.ResourceRelation.Relation,
+		req.SubjectRelation.Namespace,
+		req.SubjectRelation.Relation,
+		strings.Join(req.SubjectIds, ","),
```
jakedt, line 131: At standup I proposed changing this to a hash + count, but now I'm thinking we should probably just make sure that the cache memory usage takes into account the fully serialized cache key length. Correctness is more important than performance, and we can't rely on a fast-enough hash to be collision-free.

josephschorr: Yeah, the cache size estimator is adding the ID lengths together, so it should be (roughly) correct in its estimate

internal/graph/reachableresources.go
josephschorr: I'm not a huge fan of using the syntax for loops that break using internal conditions, but I changed it and added a comment to indicate that

josephschorr:
Updated

jakedt (dismissed):

jakedt (approved):

PR #697, opened 2022-07-18
Prevent release version errors from blocking SpiceDB startup

### LINKED ISSUE #695: SpiceDB release check failure shouldn't block
If the GitHub api returns a 403 rate limiting error, spicedb should skip the version check (or try again later), rather than blocking.

josephschorr:
Instead, we log a warning, as these errors could be due to external API conditions, such as GitHub 403ing

Fixes #695


ecordell (approved):
LGTM

If we want to in the future we could make this retry with backoff in the background, but this solves the immediate issue

jzelinskie (approved):

PR #701, opened 2022-07-19
Add a fallback into the MySQL driver for rel count

josephschorr:
We use the count(*) if the information schema has not yet been updated with the estimated row count


internal/datastore/mysql/query_builder.go
```diff
@@ -96,6 +98,12 @@ func queryTuples(tableTuple string) sq.SelectBuilder {
 	).From(tableTuple)
 }
 
+func countTuples(tableTuple string) sq.SelectBuilder {
```
jzelinskie, line 101: technically this will count any table, haha

jzelinskie (approved):

PR #703, opened 2022-07-19
Add ability to trace a check request

### LINKED ISSUE #696: Add ability to trace a check request
Add the ability for a check request's full set of dispatched requests to be traced.

This would likely involve a header sent to enable the trace, which would then be sent in response in the footer, and interpreted by tooling.

Features:
- if the permission was OK, the path that gave access to the resource. Having multiple paths would be OK too
- if the permission was Not OK, all paths that have been explored and why they failed

josephschorr:
This adds debugging information returned which traces a check request, but only if the specified header is given on the API call

Fixes #696


josephschorr:
Waiting on changes in API to be merged and pulled into authzed-go

proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -14,12 +14,19 @@ service DispatchService {
...
+  }
+
   ResolverMeta metadata = 1 [ (validate.rules).message.required = true ];
 
   core.v1.ObjectAndRelation resource_and_relation = 2
       [ (validate.rules).message.required = true ];
   core.v1.ObjectAndRelation subject = 3
       [ (validate.rules).message.required = true ];
+
+  DebugSetting debug = 4;
```
ecordell, line 29: should this be in resolvermeta?

internal/dispatch/graph/check_test.go
```diff
@@ -216,16 +216,16 @@ func TestCheckMetadata(t *testing.T) {
 				ONR("user", "product_manager", graph.Ellipsis),
 				[]expected{
 					{"owner", true, 1, 1},
-					{"edit", true, 2, 2},
-					{"view", true, 3, 3},
+					{"edit", true, 3, 2},
```
ecordell, line 219: Why did the dispatch count increase for requests that didn't have debug data?

ecordell (commented):
Are you going to add some tests showing the debug data?

josephschorr:
> Are you going to add some tests showing the debug data?

Yes, once I have pulled in the real API changes

proto/internal/dispatch/v1/dispatch.proto
josephschorr: No, otherwise we'd have to clear it explicitly from there which made things a bit more complicated when cloning

internal/dispatch/graph/check_test.go
josephschorr: I fixed the use of the response metadata in the `union` call in Check: it now uses the combined metadata, rather than just the subrequest one

josephschorr:
> Are you going to add some tests showing the debug data?

Yes, once I have pulled in the real API changes

internal/graph/check.go
```diff
@@ -62,6 +63,28 @@ func (cc *ConcurrentChecker) Check(ctx context.Context, req ValidatedCheckReques
 
 	resolved := union(ctx, []ReduceableCheckFunc{directFunc})
 	resolved.Resp.Metadata = addCallToResponseMetadata(resolved.Resp.Metadata)
+
+	// If debugging is enabled, decorate the metadata with debug information.
+	if req.Debug == v1.DispatchCheckRequest_ENABLE_DEBUGGING {
```
ecordell, line 68: Most of these debug checks could be:

```go
if !debug { 
  return <already constructed normal answer>
}
<compute debug answer>
return <debug answer>
```

josephschorr:
Updated

ecordell (approved):
LGTM

ecordell (approved):

PR #704, opened 2022-07-19
Add a dispatch test that verifies delete preconditions across datastores

josephschorr:
This PR adds a small test for running delete relationships with preconditions against all datastore implementations

### DIFF
--- a/internal/services/dispatch_test.go
+++ b/internal/services/dispatch_test.go
@@ -149,7 +149,6 @@ func TestDispatchIntegration(t *testing.T) {
 				require.Equal(t, v1.CheckPermissionResponse_PERMISSIONSHIP_NO_PERMISSION, cresp.Permissionship)
 			},
 		},
-
 		{
 			"unknown relation test",
 			`definition user {}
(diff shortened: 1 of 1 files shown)


internal/services/dispatch_test.go
```diff
@@ -190,6 +189,107 @@ func TestDispatchIntegration(t *testing.T) {
...
+					Updates: []*v1.RelationshipUpdate{
+						{
+							Operation:    v1.RelationshipUpdate_OPERATION_CREATE,
+							Relationship: tuple.MustToRelationship(tuple.MustParse("resource:someresource#viewer@user:sarah")),
+						},
+					},
+				})
+				require.NoError(t, werr)
+
+				// Ensure the delete now works.
```
jakedt, line 269: Can you add a `MUST_NOT_MATCH` after the write to verify that that also fails now.

jakedt (changes requested):

internal/services/dispatch_test.go
josephschorr: Done

josephschorr:
Updated

jakedt (approved):

PR #705, opened 2022-07-20
wip: test crdb global tables

ecordell:
In theory, global tables have the properties we need for new enemy prevention without "overlap keys".

This PR switches the test suite over to using global tables to see how they perform.

### DIFF
--- a/e2e/cockroach/cockroach.go
+++ b/e2e/cockroach/cockroach.go
@@ -23,6 +23,7 @@ type Node struct {
 	Httpaddr  string
 	ID        string
 	MaxOffset time.Duration
+	Region    string
 	Cancel    context.CancelFunc
 	// only available after Start()
 	pid  int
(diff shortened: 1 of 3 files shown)


ecordell:
```
rpc error: code = Unknown desc = maximum attempts reached (11/11): ERROR: restart transaction: TransactionRetryWithProtoRefreshError: TransactionRetryError: retry txn (RETRY_SERIALIZABLE - failed preemptive refresh): "sql txn" meta={id=d579987f key=/Table/55/1 pri=0.00268114 epo=0 ts=1658785861.125644489,2? min=1658785851.074435084,0 seq=2000} lock=true stat=PENDING rts=1658785851.074435084,0 wto=false gul=1658785856.074435084,0 (SQLSTATE 40001)
```

This error occurs during test setup - the tests may just need adjusting, but since overlap keys work fine (and since, global tables are a CRDB enteprise-only feature), I don't think it's worth pursuing more at the moment.

PR #707, opened 2022-07-25
remove rangefeed config from crdb migrations

ecordell:
`rangefeed` is a cluster-wide setting in cockroachdb that requires `admin` permission to change, and depending on how cockroach is running, it may not be user-configurable at all. It also affects the entire cockroach cluster (not just the logical database for spicedb) so there is a bit of an impedance mismatch by having the migrations set the flag.

For example, in cockroach serverless, attempting to enable rangefeeds gives this message:

```
{"level":"fatal","error":"error executing migration function: ERROR: cluster setting 'kv.rangefeed.enabled' is currently overridden by the operator (SQLSTATE XXUUU)","time":"2022-07-27T11:02:04-07:00","message":"unable to complete requested migrations"}
```
(thanks to @arashpayan for reporting this!)

This PR removes the rangefeed setting from the migrations. It adds a new `Features` method to the datastore interface which reports which features are enabled/disabled for a given datastore - the only one right now is `Watch`, and the only implementation that can disable it is cockroach, but we will likely find other uses for probing a datastore for settings in the future.

If the underlying cockroach doesn't support rangefeeds, the watch API is disabled, and a warning is printed on startup:
```json
{"level":"warn","reason":"Range feeds must be enabled in CockroachDB to support the Watch API. A CockroachDB admin can set kv.rangefeed.enabled = true to enable range feeds.","time":"2022-08-01T12:58:39Z","message":"watch api disabled; underlying datastore does not support it"}
```

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -52,6 +52,7 @@ const (
 	errUnableToInstantiate = "unable to instantiate datastore: %w"
 	errRevision            = "unable to find revision: %w"
 
+	queryShowRangefeed      = "SHOW CLUSTER SETTING kv.rangefeed.enabled;"
 	querySelectNow          = "SELECT cluster_logical_timestamp()"
 	queryShowZoneConfig     = "SHOW ZONE CONFIGURATION FOR RANGE default;"
 	querySetTransactionTime = "SET TRANSACTION AS OF SYSTEM TIME %s"
@@ -305,6 +306,18 @@ func (cds *crdbDatastore) HeadRevision(ctx context.Context) (datastore.Revision,
 	})
 
 	return hlcNow, err
+}
+
+func (cds *crdbDatastore) Features(ctx context.Context) (*datastore.Features, error) {
+	var features datastore.Features
+	err := cds.pool.QueryRow(ctx, queryShowRangefeed).Scan(&features.Watch.Enabled)
+	if err != nil {
+		return nil, err
+	}
+	if !features.Watch.Enabled {
+		features.Watch.Reason = "Range feeds must be enabled in CockroachDB to support the Watch API. A CockroachDB admin can set kv.rangefeed.enabled = true to enable range feeds."
+	}
+	return &features, nil
 }
 
 func readCRDBNow(ctx context.Context, tx pgx.Tx) (decimal.Decimal, error) {
--- a/internal/datastore/proxy/readonly.go
+++ b/internal/datastore/proxy/readonly.go
@@ -9,47 +9,15 @@ import (
 var errReadOnly = datastore.NewReadonlyErr()
 
 type roDatastore struct {
-	delegate datastore.Datastore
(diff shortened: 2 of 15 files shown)


pkg/cmd/server/server.go
```diff
@@ -207,6 +207,13 @@ func (c *Config) Complete() (RunnableServer, error) {
 		return nil, fmt.Errorf("failed to create dispatch gRPC server: %w", err)
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	datastoreFeatures, err := ds.Features(ctx)
```
josephschorr, line 212: newline before this line

pkg/cmd/server/server.go
```diff
@@ -217,6 +224,12 @@ func (c *Config) Complete() (RunnableServer, error) {
 		v1SchemaServiceOption = services.V1SchemaServiceDisabled
 	}
 
+	watchServiceOption := services.WatchServiceEnabled
+	if !datastoreFeatures.CanWatch {
+		log.Warn().Msg("watch api disabled; underlying datastore does not support it")
```
josephschorr, line 229: I wonder if we can at least print out a link to how to enable it? Ideally, we'd provide the command to run to enable it

ecordell: I debated adding a `CanWatchMessage string` to the `Features` object so that the datastore could return more info about what to do, but it seemed like overkill since this is something we can easily fill with docs

josephschorr: Maybe have the Features object contain a helper text field, and we put a link to the docs in there?

internal/datastore/crdb/crdb.go
```diff
@@ -307,6 +308,17 @@ func (cds *crdbDatastore) HeadRevision(ctx context.Context) (datastore.Revision,
 	return hlcNow, err
 }
 
+func (cds *crdbDatastore) Features(ctx context.Context) (*datastore.Features, error) {
+	var features datastore.Features
+	err := cds.pool.BeginTxFunc(ctx, pgx.TxOptions{AccessMode: pgx.ReadOnly}, func(tx pgx.Tx) error {
```
jakedt, line 313: I don't think you need to use BeginTxFunc here. You can just use `QueryRow` on the pool and it will do the right thing.

internal/datastore/proxy/readonly.go
```diff
@@ -9,47 +9,15 @@ import (
 var errReadOnly = datastore.NewReadonlyErr()
 
 type roDatastore struct {
-	delegate datastore.Datastore
```
jakedt, line 12: Cool.

jakedt (changes requested):

pkg/cmd/server/server.go
ecordell: Added a per-feature message field that we can add help text to

jakedt (approved):

PR #708, opened 2022-07-25
Further fixes to memdb to not panic post-close

josephschorr:
We were still creating a situation where the db could be nil, so we add a check for that


internal/datastore/memdb/memdb.go
```diff
@@ -108,8 +112,13 @@ func (mdb *memdbDatastore) SnapshotReader(revision datastore.Revision) datastore
 		revIndex = len(mdb.revisions) - 1
```
ecordell, line 112: shouldn't this also check that `len(mdb.revisions) > 1` so that revIndex can't go negative?

internal/datastore/memdb/memdb.go
```diff
@@ -108,8 +112,13 @@ func (mdb *memdbDatastore) SnapshotReader(revision datastore.Revision) datastore
...
 	}
 
-	snapshotRevision := mdb.revisions[revIndex].revision
-	roTxn := mdb.revisions[revIndex].db.Txn(false)
+	rev := mdb.revisions[revIndex]
+	if rev.db == nil {
```
ecordell, line 116: an alternative would be to have a `closed` bool on the datastore, which would could check at the top instead (since this and `.Close()` are both protected by the mutex)

no strong preference though, if you prefer this method.

internal/datastore/memdb/memdb.go
josephschorr: We check the number of revisions above

internal/datastore/memdb/memdb.go
josephschorr: I'd likely still want the `nil` check just to be on the safe side, but I can add a `closed` bool if you like

ecordell (approved):
LGTM

PR #710, opened 2022-07-25
Have WASM development package return updated validation YAML

josephschorr:
Necessary for tooling that allows for automatic regeneration of validation YAML


samkim (approved):
LGTM

PR #711, opened 2022-07-25
Add a flag for disabling stat writes

ecordell:
Writes to the stats table were making the e2e test flow less predictable.


josephschorr (approved):
LGTM

PR #713, opened 2022-07-26
Concurrency limit

jakedt:
This PR adds a concurrency limit to each step of graph resolution in order to limit the number of goroutines that are spawned at once. It does not add an overall global goroutine limit.

### DIFF
--- a/internal/dispatch/graph/check_test.go
+++ b/internal/dispatch/graph/check_test.go
@@ -169,7 +169,7 @@ func TestMaxDepth(t *testing.T) {
 		},
 	}}
 
-	ctx := datastoremw.ContextWithHandle(context.Background())
+	ctx := log.Logger.WithContext(datastoremw.ContextWithHandle(context.Background()))
 	require.NoError(datastoremw.SetInContext(ctx, ds))
 
 	revision, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
@@ -299,7 +299,7 @@ func newLocalDispatcher(require *require.Assertions) (context.Context, dispatch.
 	cachingDispatcher.SetDelegate(dispatch)
 	require.NoError(err)
 
-	ctx := datastoremw.ContextWithHandle(context.Background())
+	ctx := log.Logger.WithContext(datastoremw.ContextWithHandle(context.Background()))
 	require.NoError(datastoremw.SetInContext(ctx, ds))
 
 	return ctx, cachingDispatcher, revision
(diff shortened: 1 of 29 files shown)


internal/dispatch/graph/check_test.go
```diff
@@ -178,7 +178,7 @@ func TestMaxDepth(t *testing.T) {
 	require.NoError(err)
 	require.True(revision.GreaterThan(decimal.Zero))
 
-	dispatch := NewLocalOnlyDispatcher()
+	dispatch := NewLocalOnlyDispatcher(10)
```
josephschorr, line 181: I wonder if we should have a `New` function that sets a default

internal/graph/check.go
```diff
@@ -444,3 +463,38 @@ func combineResponseMetadata(existing *v1.ResponseMeta, responseMetadata *v1.Res
...
+	requests []ReduceableCheckFunc,
+	resultChan chan<- CheckResult,
+	concurrencyLimit uint16,
+) func() {
+	sem := make(chan struct{}, concurrencyLimit)
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
+	dispatcher:
```
josephschorr, line 478: Why the decision to use this vs the new limiting errorgroup?

internal/dispatch/graph/graph.go
```diff
@@ -24,23 +24,23 @@ const errDispatch = "error dispatching request: %w"
...
 	d.reachableResourcesHandler = graph.NewConcurrentReachableResources(d)
 
 	return d
 }
 
 // NewDispatcher creates a dispatcher that consults with the graph and redispatches subproblems to
 // the provided redispatcher.
-func NewDispatcher(redispatcher dispatch.Dispatcher) dispatch.Dispatcher {
-	checker := graph.NewConcurrentChecker(redispatcher)
+func NewDispatcher(redispatcher dispatch.Dispatcher, concurrencyLimit uint16) dispatch.Dispatcher {
```
josephschorr, line 40: Is there a dispatch test of some kind we can add to verify the limit works as intended?

internal/dispatch/graph/check_test.go
jakedt: There is a default for all non-test call sites.

internal/graph/check.go
jakedt: The errgroup approach was causing too much churn to the interfaces. We can port over to that one by one, but coupling sweeping interface changes was making that PR more risky than it had to be.

internal/dispatch/graph/graph.go
jakedt: I could definitely test the `dispatchAllAsync` with a unit test. Would that work? Hard to test it from the dispatch layer because the behavior is supposed to be identical/opaque.

josephschorr: Yeah, that's good

jakedt: Added

josephschorr (approved):
LGTM

josephschorr (approved):
LGTM

PR #714, opened 2022-07-28
Add ability to enable the REST gateway for the test server

### LINKED ISSUE #706: Add REST API option to the serve-testing command
**Problem**: The serve-testing command does not provide an option to start with the REST API available. This leaves teams that use the REST API instead of gRPC who want to run tests against spicedb to rely on spinning up a full-featured instance of spicedb as part of the pipeline and when doing so they lose access to the features of the serve-testing command that are excellent for testing (ex: being able to run tests in parallel)

**Suggestion**: add the following flags to the serve-testing command: `--http-enabled`, and `--http-addr` and enable the REST API when the `--http-enabled` flag is set to `true` (default to false)

josephschorr:
`serve-testing` can now have the REST gateway enabled for the normal and read-only endpoints via `--http-enabled` and `--readonly-http-enabled`, which will serve (by default) on ports `:8081` and `:8082`, respectively

Fixes #706

### DIFF
--- a/pkg/cmd/testserver/testserver.go
+++ b/pkg/cmd/testserver/testserver.go
@@ -2,6 +2,7 @@ package testserver
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
@@ -9,6 +10,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/authzed/spicedb/internal/dispatch/graph"
+	"github.com/authzed/spicedb/internal/gateway"
 	consistencymw "github.com/authzed/spicedb/internal/middleware/consistency"
 	dispatchmw "github.com/authzed/spicedb/internal/middleware/dispatcher"
 	"github.com/authzed/spicedb/internal/middleware/pertoken"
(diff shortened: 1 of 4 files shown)


pkg/cmd/testserver/testserver.go
```diff
@@ -96,17 +100,51 @@ func (c *Config) Complete() (RunnableTestServer, error) {
 		return nil, err
 	}
 
+	gatewayHandler, err := gateway.NewHandler(context.TODO(), c.GRPCServer.Address, c.GRPCServer.TLSCertPath)
```
jzelinskie, line 103: `context.Background()` unless we want to leave `TODO` here to signal that at some point we need to thread a context into this function

jzelinskie (approved):
Only a question, otherwise LGTM

pkg/cmd/testserver/testserver.go
josephschorr: Ah; I copied this directly from the existing serve implementation. @ecordell any idea why we have a `TODO` there as well?

ecordell: And I also was just matching what was there [originally](https://github.com/jzelinskie/spicedb/blob/e9b164af2de444cfc945b752b27fc0fbbcb36d06/cmd/spicedb/serve.go#L388) before I moved initialization into the cmd package 😄 

I think it would need to get refactored a good bit, ideally nothing that required a context would happen until you call `Run(ctx)`

PR #722, opened 2022-08-01
switch crdb watch feature detection to only require CONTROLCHANGEFEED

ecordell:
This is a followup to #707, which removed the need for admin permission for CRDB migrations.

#707 switched to using a `SHOW CLUSTER SETTING` command to determine if rangefeeds are enabled.

This PR switches the check to attempt to create a changefeed, similar to what the `Watch` API does. This requires the `CONTROLCHANGEFEED` permission in cockroach, which matches the requirements the `Watch` API already has.

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -2,6 +2,7 @@ package crdb
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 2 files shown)


josephschorr (approved):
LGTM

PR #728, opened 2022-08-01
Fix bounds on check test

josephschorr:
Sometimes we get the other branches, and it is more than 5 dispatches


ecordell (approved):
LGTM

ecordell (approved):
LGTM

PR #734, opened 2022-08-02
Skip checking of permissions for relationships in WriteSchema

### LINKED ISSUE #733: v1.WriteSchema relationship sanity check should skip permissions
To ensure that data remains valid across schema migrations, `v1.WriteSchema` [performs a sanity check](https://github.com/authzed/spicedb/blob/efa218eb0b9b13af6f0910d5eed0846fc16ff04c/internal/services/shared/schema.go#L53).

An optimization that is missing for this sanity check is to skip permissions as they are computed and cannot contain any data themselves. This optimization would vastly improve the performance of large schema writes on production systems and make schema migrations that only modify permissions nearly instantaneous. 

Big thanks to @williamdclt for the report the surfaced this issue.

josephschorr:
This PR changes the namespace diff logic to treat relations and permissions distinctly, thus ensuring that WriteSchema will only check for relationships if a *relation* is removed, and not a permission as well. This was a legacy of the time when relations could contain both data and expressions.

Fixes #733

### DIFF
--- a/internal/namespace/diff.go
+++ b/internal/namespace/diff.go
@@ -7,7 +7,9 @@ import (
 	"github.com/scylladb/go-set/strset"
 
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	iv1 "github.com/authzed/spicedb/pkg/proto/impl/v1"
 
+	nspkg "github.com/authzed/spicedb/pkg/namespace"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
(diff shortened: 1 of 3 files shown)


internal/namespace/diff.go
```diff
@@ -149,17 +178,47 @@ func DiffNamespaces(existing *core.NamespaceDefinition, updated *core.NamespaceD
...
+	}
+
+	for _, shared := range strset.Intersection(existingPermNames, updatedPermNames).List() {
+		existingPerm := existingPerms[shared]
+		updatedPerm := updatedPerms[shared]
+
+		// Compare implementations.
+		m := &jsonpb.Marshaler{}
+		existingRewriteJSON, _ := m.MarshalToString(existingPerm.UsersetRewrite)
+		updatedRewriteJSON, _ := m.MarshalToString(updatedPerm.UsersetRewrite)
```
jzelinskie, line 202: This might need to panic instead of ignoring the error

jzelinskie (approved):

PR #737, opened 2022-08-07
hack: mv install-tools into hack dir

jzelinskie:
This is a really simple PR to move the `install-tools.sh` script into the hack directory.

I also ran shfmt, shellcheck, and checkbashisms on the scripts.


josephschorr (approved):
LGTM

PR #740, opened 2022-08-10
fix panic on cached optimized revisions

vroldanbet:
# What

This PR fixes a panic that sometimes triggered when database connection errors happened during the computation of optimized revisions through the singleflight wrapper.

# Why

Any error happening in the optimized revision function executed through the singleflight group was unhandled, which led `lastQuantizedRevision.(decimal.Decimal)` type assertion to panic:

https://github.com/authzed/spicedb/blob/ae4552ed89f0561f71893da2feeb7feb1e767e71/internal/datastore/common/revisions/optimized.go#L72

```
panic: interface conversion: interface {} is nil, not decimal.Decimal
```

# How

By adding good old `if err != nil ...` 😄 

# Notes

There are some `gofumpt` lint errors caused with go 1.19, so had to fix then here 🙏🏻

### DIFF
--- a/internal/datastore/common/revisions/optimized.go
+++ b/internal/datastore/common/revisions/optimized.go
@@ -68,7 +68,9 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
 
 		return optimized, nil
 	})
-
+	if err != nil {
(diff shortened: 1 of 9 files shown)


ecordell (approved):
LGTM

PR #744, opened 2022-08-11
more resilient CRDB datastore on connection draining

vroldanbet:
I introduced some changes to help with connection draining:

- Log when datastore detect errors and retries
- Retry on `unexpected EOF` - all operations are idempotent
- Retry when CRDB is draining and does not accept connections
- Propagate spicedb CLI value for health-check-interval parameter.
- Do retry on errors in `Datastore.HeadRevision`

# A look at connection draining with the pgx driver

In order to drain connections properly we need to understand how that's handled by the PGX driver.

- connections that are committed or rolledback are released back to the pool. The `Release()` function will check for idle and max lifetime, so it will close any connections that have exceeded those values.
- the other way to get connections closed is via the health-check. This is a goroutine that is executed on a regular interval. PGX **tries** to get available connections from the pool to check if they have exceeded those parameters. This is important, because if a connection is always locked by the application, it means the goroutine will never be able to check max lifetime, and therefore can exceed that value until it's returned back to the pool

The TLDR is:

- make sure to always end statements with commit/rollback connections, or they will never be released to the pool, and never closed even if they have exceeded max lifetime
- the health-check should be set to a value close the the max lifetime, or less, in order to increase the odds the deadline is honored
- depending on the workload, it's entirely possible connections may exceed significantly their lifetime if they are held very often by the application.
- the health-check may put pressure on the connection pool. The more frequent, the more those connections will spend doing unproductive health checks outside of the pool.

### DIFF
--- a/internal/datastore/crdb/options.go
+++ b/internal/datastore/crdb/options.go
@@ -6,10 +6,11 @@ import (
 )
 
 type crdbOptions struct {
-	connMaxIdleTime *time.Duration
-	connMaxLifetime *time.Duration
-	minOpenConns    *int
-	maxOpenConns    *int
+	connMaxIdleTime         *time.Duration
+	connMaxLifetime         *time.Duration
+	connHealthCheckInterval *time.Duration
+	minOpenConns            *int
+	maxOpenConns            *int
 
 	watchBufferLength           uint16
 	revisionQuantization        time.Duration
@@ -82,6 +83,24 @@ func generateConfig(options []Option) (crdbOptions, error) {
 func SplitAtUsersetCount(splitAtUsersetCount uint16) Option {
 	return func(po *crdbOptions) {
 		po.splitAtUsersetCount = splitAtUsersetCount
+	}
+}
+
+// ConnHealthCheckInterval is the frequency at which both idle and max lifetime connections
+// are checked, and also the frequency at which minimum number of connections is
+// checked. This happens asynchronously.
+//
+// This is not the only approach to evaluate those: connection idle/max lifetime
+// is also checked when connections are released to the pool.
+//
+// There is no guarantee connections won't last longer than their specified idle/max lifetime. It's largely
+// dependent on the health-check goroutine being able to pull them from the connection pool. The health-check
+// may not be able to clean up those connections is they are held by the application very frequently.
+//
+// This value defaults to 30s.
+func ConnHealthCheckInterval(interval time.Duration) Option {
+	return func(po *crdbOptions) {
+		po.connHealthCheckInterval = &interval
 	}
 }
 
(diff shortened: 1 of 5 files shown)


internal/datastore/crdb/options.go
```diff
@@ -85,6 +86,24 @@ func SplitAtUsersetCount(splitAtUsersetCount uint16) Option {
 	}
 }
 
+// ConnHealthCheckInterval is the frequency at which both idle and max lifetime connections
+// are checked, and also the frequency at which minimum number of connections is
```
josephschorr, line 90: at which the minimum

internal/datastore/crdb/options.go
```diff
@@ -85,6 +86,24 @@ func SplitAtUsersetCount(splitAtUsersetCount uint16) Option {
...
+// checked. This happens asynchronously.
+//
+// This is not the only approach to evaluate those: connection idle/max lifetime
```
josephschorr, line 93: to evaluate these counts

internal/datastore/crdb/options.go
```diff
@@ -85,6 +86,24 @@ func SplitAtUsersetCount(splitAtUsersetCount uint16) Option {
...
+// is also checked when connections are released to the pool.
+//
+// There is no guarantee connections won't last longer than their specified idle/max lifetime. It's largely
+// dependent on the health-check goroutine being able to pull them from the connection pool. The health-check
+// may not be able to clean up those connections is they are held by the application very frequently.
```
josephschorr, line 98: is -> if

internal/datastore/crdb/tx.go
```diff
@@ -67,7 +69,7 @@ func executeWithResets(ctx context.Context, fn innerFunc, maxRetries uint8) (err
 	}
 
 	// The last error was resettable but we're out of retries
-	return fmt.Errorf(errReachedMaxAttempts, retries, maxRetries+1, err)
+	return fmt.Errorf(errReachedMaxRetries, retries, maxRetries+1, err)
 }
 
 func resettable(ctx context.Context, err error) bool {
```
josephschorr, line 75: Can we add some sort of unit test for this method?

vroldanbet: @josephschorr the method seems fairly tested indirectly [through table tests invoking](https://github.com/authzed/spicedb/blob/ad65811afa4efeb2ca157cb6a876141163f57f28/internal/datastore/crdb/tx_test.go#L43) `executeWithResets`. Just run code coverage and showed  100% code coverage.

With that said, I can still add unit tests if you feel strongly about it 👍🏻

josephschorr: Nope, that's fine

josephschorr (approved):
LGTM

PR #745, opened 2022-08-11
bump crdb in tests to 22.1.5

ecordell:


samkim:
Stage's crdb operator will need to be upgraded to support v22.1.5 and the spicedb operator e2e test cockroach version can be updated too

vroldanbet (approved):

PR #747, opened 2022-08-16
replace Improper random of math/rand to crypto/rand

CoolerVoid:
More information on the context cwe.mitre.org/data/definitions/330.html

### DIFF
--- a/pkg/balancer/hashring.go
+++ b/pkg/balancer/hashring.go
@@ -1,7 +1,7 @@
 package balancer
 
 import (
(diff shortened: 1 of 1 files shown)


CoolerVoid:
I have read the CLA Document, and I hereby sign the CLA.

jakedt:
We explicitly chose `math/rand` for its performance. Every time the backend list changes the hashring needs to be recomputed. Is there a reason you think there is a security implication to this particular use of `math/rand`?

PR #760, opened 2022-08-22
Begin implementation of a library for processing caveat expressions

### LINKED ISSUE #759: Library for evaluating caveats expressions
Tracking issue: https://github.com/authzed/spicedb/issues/386

Develop a general purpose library for evaluating caveats expressions and providing an answer

josephschorr:
This library implements parsing, compilation, type checking, evaluation and custom types via CEL

Fixes #759


vroldanbet (commented):
it would seem like test coverage could be improved? Some files are not even tested.

Screenshot 2022-08-29 at 18 46 17

josephschorr:
@vroldanbet that doesn't seem correct... the `eval_test.go` definitely runs the code in `eval.go`

josephschorr:
@vroldanbet Yeah, that makes a lot more sense, especially since most of `types.go` is scaffolding anyway

pkg/caveats/env.go
```diff
@@ -0,0 +1,71 @@
...
+		return fmt.Errorf("variable `%s` already exists", name)
+	}
+
+	e.variables[name] = varType.celType
+	return nil
+}
+
+// asCelEnvironment converts the exported Environment into an internal CEL environment.
+func (e *Environment) asCelEnvironment() (*cel.Env, error) {
+	opts := make([]cel.EnvOption, 0, len(e.variables)+1)
```
vroldanbet, line 55: nit: this avoids having the code make the assumption there is just one custom type, and adds 2 extra of capacity for type adapter and default utc timezone. Spares 2 allocations.
```suggestion
		opts := make([]cel.EnvOption, 0, len(e.variables) + len(customtypes.CustomTypes) + 2)
```

pkg/caveats/customtypes/customtypes.go
```diff
@@ -0,0 +1,11 @@
+package customtypes
+
+import (
+	"github.com/google/cel-go/cel"
+)
+
+var CustomTypes = map[string][]cel.EnvOption{}
+
+func registerCustomType(name string, opts ...cel.EnvOption) {
```
vroldanbet, line 9: exposing this would allow folks to inject their own types, but it also means exposing CEL as part of the API. I think exploring how to abstract this and export it in a follow up would be a nice addition.

pkg/caveats/eval.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/interpreter"
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// EvaluationConfig is configuration given to an EvaluateCaveatWithConfig call.
+type EvaluationConfig struct {
+	// MaxCost is the max cost of the caveat to be executed.
+	MaxCost uint64
```
vroldanbet, line 16: this seems _a bit specific to CEL_. But I guess we could say "alternative caveat languages" do not support it.

vroldanbet (commented):
Library looks great! I left some comments on potential improvements, and took the liberty to push some extra tests

pkg/caveats/types.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// TypeKeywords returns all keywords associated with types.
+func TypeKeywords() []string {
+	keywords := append(BasicTypeKeywords(), listTypeKeyword, mapTypeKeyword)
+	for customTypeName := range customtypes.CustomTypes {
+		keywords = append(keywords, strings.ToLower(customTypeName))
+	}
+	return keywords
+}
```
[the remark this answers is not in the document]
josephschorr: These will be used by the schema parser to determine the keywords necessary for use in the schema for types

pkg/caveats/customtypes/customtypes.go
josephschorr: Agreed.

pkg/caveats/eval.go
josephschorr: Yep, that's the idea

vroldanbet (approved):
LGTM!

Issue #757, opened 2022-08-22
Add caveats to the datastore API and implementations


vroldanbet:
Tracking issue: https://github.com/authzed/spicedb/issues/386

The datastore API and implementations will need to be made aware of caveats, and store/retrieve them on relationship operations.

PR #761, opened 2022-08-23
introduce feature flags middleware

vroldanbet:
part of https://github.com/authzed/spicedb/issues/757

introduces a `--feature-flag` argument to `spicedb serve` that allows specifying a number of enabled feature flags.
Those values will then get injected into the gRPC server context via a middleware that will make it available to the request path, so that any component can guard new logic behind a feature flag for a safer rollout of new functionality.


vroldanbet:
we are not going to move forward with feature flags for the time being

PR #770, opened 2022-08-25
Fully implement the LookupSubjects API

josephschorr:
This change also fixes a number of issues discovered while testing:

1) Dispatch for lookup subjects has to track wildcard exclusions. For this a new SubjectSet is implemented and tested
2) LookupSubjects was added to the consistency test suite
3) TrackingSubjectSet was changed to be based on the same logic as SubjectSet and tests updated+improved
4) Added additional consistency tests for wildcards and other case


internal/graph/lookupsubjects.go
```diff
@@ -342,47 +351,53 @@ type lookupSubjectsReducer interface {
...
+		for _, result := range collector.Results() {
+			metadata = combineResponseMetadata(metadata, result.Metadata)
+			results.UnionWith(result.FoundSubjects)
+		}
+
+		if index == 0 {
+			foundSubjects = results
+		} else {
+			foundSubjects.UnionWithSet(results)
+		}
```
jzelinskie, line 390: Does this make a huge perf difference that you shouldn't just always Union and preallocate `foundSubjects := util.NewSubjectSet()` before the loop?

internal/membership/foundsubject.go
```diff
@@ -22,13 +22,23 @@ type FoundSubject struct {
...
-	excludedSubjects *tuple.ONRSet
+	excludedSubjectIds []string
 
 	// relations are the relations under which the subject lives that informed the locating
 	// of this subject for the root ONR.
 	relationships *tuple.ONRSet
 }
 
+// GetSubjectId is named to match the Subject interface for the BaseSubjectSet.
+//nolint:all
```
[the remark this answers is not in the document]
josephschorr: Because otherwise it complains that it should be `GetSubjectID`, which doesn't match the generated proto code

internal/graph/lookupsubjects.go
josephschorr: Nope; this was mostly a copy-paste thing. I can change it

josephschorr:
Updated

install-tools.sh
```diff
@@ -0,0 +1,2 @@
+tools=$(go list -f '{{range .Imports}}{{.}} {{end}}' tools.go)
```
jzelinskie, line 1: This exists already, it's just in the hack directory

jzelinskie (changes requested):

install-tools.sh
josephschorr: Right! Removing...

josephschorr: Removed

jzelinskie (approved):

PR #771, opened 2022-08-29
removes v1.RelationshipUpdate from datastore.ReadWriteTransaction

vroldanbet:
Part of https://github.com/authzed/spicedb/issues/757

# What

This refactor replaces `v1.RelationshipUpdate` with `core.RelationTupleUpdate` in the `datastore.ReadWriteTransaction` interface.

# Why

The goal is to help decouple the datastore from the public interfaces and use internal data structures to help support implementation details, like those needed for Caveats. This makes sure we don't need to push implementation details to the public interface just because it's needed in the datastore. Particularly, this will be used to keep Caveat as strings in the public interfaces, but bytes in the internal interfaces, as we will store them compiled in the datastore.

# How

Changing the datastore interfaces causes a large and boring changeset. I found myself doing repetitive work during the refactor that could have been avoid with some DRY, so I took the opportunity to do it in this PR.

I added a temporal workaround to `TestDeleteRelationship` because I found datastores aren't able to cope with duplicate tuples. This didn't surface earlier because most tests called `ReadWriteTransaction.WriteRelationships` multiple times with single-element slices, instead of having all tuples in the slice. I didn't want this changeset to blow up in scope, so I made sure the original unintentional behaviour remains.

Finally, `ValidateUpdatesToWrite` was removed from datastore/common package because it was only used in testfixtures, and it called HandwrittenValidate. This is unavailable in core, so I converted the core tuple to a v1 relationship just to call HandwrittenValidate. This should be fine because this is just used for tests and not in the critical path.


pkg/datastore/datastore.go
```diff
@@ -171,7 +171,7 @@ type ReadWriteTransaction interface {
 	Reader
 
 	// WriteRelationships takes a list of tuple mutations and applies them to the datastore.
-	WriteRelationships(mutations []*v1.RelationshipUpdate) error
+	WriteRelationships(mutations ...*core.RelationTupleUpdate) error
```
jakedt, line 174: This change really didn't improve readability or usability at all. There isn't a single call site in the PR that doesn't call this using a slice with the splat `...` operator. Please revert it.

pkg/datastore/test/tuples.go
```diff
@@ -440,14 +396,15 @@ func DeleteRelationshipsTest(t *testing.T, tester DatastoreTester) {
 
 			tRequire := testfixtures.TupleChecker{Require: require, DS: ds}
 
+			// FIXME temporarily store tuples in multiple calls to ReadWriteTransaction since no Datastore
```
jakedt, line 399: We use TODO in other places in our codebase.

jakedt (changes requested):

internal/datastore/common/helpers.go
```diff
@@ -0,0 +1,26 @@
+package common
+
+import (
+	"context"
+
+	"github.com/authzed/spicedb/pkg/datastore"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+func WriteTuples(ctx context.Context, ds datastore.Datastore, op core.RelationTupleUpdate_Operation, tuples ...*core.RelationTuple) (datastore.Revision, error) {
```
josephschorr, line 10: Doc comments on all exported methods

josephschorr, line 10: Technically I think this should be `UpdateTuplesInDatastore`, because the update op is given

internal/datastore/crdb/readwrite.go
```diff
@@ -93,34 +93,34 @@ func (rwt *crdbReadWriteTXN) WriteRelationships(mutations []*v1.RelationshipUpda
...
-				rel.Relation,
-				rel.Subject.Object.ObjectType,
-				rel.Subject.Object.ObjectId,
-				stringz.DefaultEmpty(rel.Subject.OptionalRelation, datastore.Ellipsis),
+				rel.ResourceAndRelation.Namespace,
+				rel.ResourceAndRelation.ObjectId,
+				rel.ResourceAndRelation.Relation,
+				rel.Subject.Namespace,
+				rel.Subject.ObjectId,
+				stringz.DefaultEmpty(rel.Subject.Relation, datastore.Ellipsis),
```
josephschorr, line 109: `rel.Subject.Relation` should never be empty, since it is a `core` type: instead, we should validate it is never given to the datastore empty

pkg/datastore/datastore.go
vroldanbet: here is my rationale for the change: in the test codebase there is _a lot_ of snippets of code list this:

```go
return rwt.WriteRelationships([]*v1.RelationshipUpdate{{
			Operation:    v1.RelationshipUpdate_OPERATION_CREATE,
			Relationship: relationship,
		}})
```

Granted, that became not so much of a big deal once I introduced `common.WriteTuples` which DRY things even further.

There is already another method in the interface. If you feel strongly for this I can change it.

internal/datastore/crdb/readwrite.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/771/commits/c38d5a48c8670afefed0afd045c2764fddc5e112

pkg/datastore/datastore.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/771/commits/f709966b64c80a46f44ef75cfe0dcfa8ddfb31c0

jakedt (approved):

PR #773, opened 2022-08-29
introduce caveats in datastore API

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/757

# What

Introduces `Caveat` in `core.RelationTuple` and adjusts all datastores to be able to read/write it

# Why

Caveats are additional environmental constraints of a tuple, and hence the first step is to be able to store it in the `Datastore` interface.

# How

This implementation does not make any assumption on how the actual caveats are implemented. However, we assume that somewhere in the gRPC controllers the `core.RelationTuple.Caveat` field will be set, and will be the result of compiling the textual representation of a Caveat language.

An initial limit of 4KB has been established and is enforced via proto validation. Each datastore uses a different datatype, but only spanner allows defining a limit, as far as I could tell.

I identified `WriteRelationships` wasn't validating the corev1 protos, so I also modified all implementations to call `Validate` on the proto value.


internal/datastore/spanner/migrations/zz_migration.0003_add_caveat.go
```diff
@@ -0,0 +1,31 @@
+package migrations
+
+import (
+	"context"
+
+	"google.golang.org/genproto/googleapis/spanner/admin/database/v1"
+)
+
+const (
+	addCaveatToRelationTuple     = `ALTER TABLE relation_tuple ADD COLUMN caveat BYTES(4096)`
```
josephschorr, line 10: The `4096` here is smaller than the other datastore migrations; we might need to document this somewhere

proto/internal/core/v1/core.proto
```diff
@@ -12,6 +12,8 @@ message RelationTuple {
 
   /** subject is the subject for the tuple */
   ObjectAndRelation subject = 2 [ (validate.rules).message.required = true ];
+
+  bytes caveat = 3 [(validate.rules).bytes = {min_len: 0, max_len: 4096}];
```
josephschorr, line 16: Add a comment

internal/datastore/spanner/migrations/zz_migration.0003_add_caveat.go
vroldanbet: this is defined in the proto: https://github.com/authzed/spicedb/pull/773/files#diff-daf9cb335c0ae98207245de93841668f2a4bf8025073e2cdf0af2cdf7589a18eR16 but you are right other datastores are not defining the limit 👍🏻 

do you have any thoughts on what a reasonable starting limit for this could look like?

pkg/datastore/test/caveat.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/authzed/spicedb/internal/testfixtures"
+	"github.com/authzed/spicedb/pkg/datastore"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/tuple"
+
+	"github.com/stretchr/testify/require"
+)
+
+// SimpleCaveatTest makes sure a caveated tuples is stored and retrieved from the datastore correctly
+func SimpleCaveatTest(t *testing.T, tester DatastoreTester) {
```
jakedt, line 16: Can you make this a table driven test that exercises nil, empty, small (current), and large caveat values?

jakedt (changes requested):

proto/internal/core/v1/core.proto
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/773/commits/16db8c396bfac1613471584d50b7bb3d28cd456b

internal/datastore/spanner/migrations/zz_migration.0003_add_caveat.go
vroldanbet: I've documented the size limit of each datastore in https://github.com/authzed/spicedb/pull/773/commits/aa8b50e52f054c8a92ac359b1eb92378a8fff72f.
The proto validation plugin should make sure no more than 4KB are allowed, but I'll add a test to validate it

pkg/datastore/test/caveat.go
vroldanbet: @jakedt good catch, addressed in https://github.com/authzed/spicedb/pull/773/commits/9e7b75c429ce83769520e6443e19bbdda2ba82fe.

Also identified datastores weren't validating protos, as the assumption was gRPC controllers would do this, but does not hold true anymore as we now use `core` for `WriteRelationships`. Added test in https://github.com/authzed/spicedb/pull/773/commits/e93451c38cd69e592614fefb9000d84c2939be1a

vroldanbet:
closing to explore yet another approach

Issue #765, opened 2022-08-25
add database connection pool metrics for PostgreSQL/CRDB


vroldanbet:
We currently have no means to gather metrics from the connection pool in PostgreSQL/CRDB datastores.
The MySQL datastore has some prior art on how to do this.

PR #774, opened 2022-08-30
add datastore metrics to CockroachDB datastore

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/765

- ports postgres' support for prometheus stats of connection pools into CRDB
- refactor `datastore/common` so that shared PGX code moves to `datastore/postgres/common`


internal/datastore/postgres/common/common.go
```diff
@@ -0,0 +1,74 @@
+package common
```
jzelinskie, line 1: could name this file pgx.go

jzelinskie (dismissed):

internal/datastore/postgres/common/common.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/774/commits/6981da212673943af881e315906a72f325ac8922

jzelinskie (approved):

PR #775, opened 2022-08-30
Add configurable limits for write and delete relationship APIs

josephschorr:
This sets a new (configurable) limit of 1000 updates and preconditions for WriteRelationships and DeleteRelationships

### DIFF
--- a/internal/services/v1/relationships.go
+++ b/internal/services/v1/relationships.go
@@ -28,14 +28,35 @@ import (
 	"github.com/authzed/spicedb/pkg/zedtoken"
 )
 
+// PermissionsServerConfig is configuration for the permissions server.
+type PermissionsServerConfig struct {
+	// MaxUpdatesPerWrite holds the maximum number of updates allowed per
+	// WriteRelationships call.
+	MaxUpdatesPerWrite uint16
+
+	// MaxPreconditionsCount holds the maximum number of preconditions allowed
+	// on a WriteRelationships or DeleteRelationships call.
+	MaxPreconditionsCount uint16
+
+	// MaximumAPIDepth is the default/starting depth remaining for API calls made
+	// to the permissions server.
+	MaximumAPIDepth uint32
+}
+
 // NewPermissionsServer creates a PermissionsServiceServer instance.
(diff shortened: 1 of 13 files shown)


internal/services/v1/relationships.go
```diff
@@ -28,14 +28,27 @@ import (
...
+	// MaxPreconditionsCount holds the maximum number of preconditions allowed
+	// on a WriteRelationships or DeleteRelationships call.
+	MaxPreconditionsCount uint16
+}
+
 // NewPermissionsServer creates a PermissionsServiceServer instance.
 func NewPermissionsServer(
 	dispatch dispatch.Dispatcher,
 	defaultDepth uint32,
+	config PermissionsServerConfig,
```
ecordell, line 46: Looking at the existing code, the servers all just take in a list of their "config" so that the Server object itself holds all the options.

Why not do that here? Or if we're going have config objects, why not put all of the config in them?

pkg/cmd/server/server.go
```diff
@@ -235,6 +238,11 @@ func (c *Config) Complete() (RunnableServer, error) {
 		c.UnaryMiddleware, c.StreamingMiddleware = DefaultMiddleware(log.Logger, c.GRPCAuthFunc, !c.DisableVersionResponse, dispatcher, ds)
 	}
 
+	permSysConfig := v1svc.PermissionsServerConfig{
+		MaxPreconditionsCount: defaultIfZero(c.MaximumPreconditionCount, 1000),
```
ecordell, line 242: You shouldn't need the `defaultIfZero` in this case - the cli flag bindings set the default values for you.

internal/services/v1/relationships.go
josephschorr: I'm fine with moving the defaultDepth and other non-dispatch values into the config, if you prefer

pkg/cmd/server/server.go
josephschorr: They do, but if I manually create the config they'll start at 0, so I want to be absolutely sure

ecordell (approved):
LGTM

PR #778, opened 2022-09-01
newly added lookup subject metrics weren't being unregistered

vroldanbet:
fixes bug introduced in https://github.com/authzed/spicedb/pull/736

This makes the caching dispatcher to unregister newly added metrics. It caused some of our internal tests to fail with duplicate metrics being registered


ecordell (approved):
LGTM

PR #798, opened 2022-09-06
map pgx info logging level to debug

vroldanbet:
This PR reduces noise in SpiceDB's logs for PostgreSQL / CRDB datastores, and reduces performance overhead caused by the logging of very large statements, like for example `WriteRelationships` of thousands of elements. We observed the server would sometimes fail with `context deadline exceeded` when PGX statement logging was enabled.

It does it by turning every pgx `info` event into a `debug` event.

Screenshot 2022-09-05 at 18 03 04
Screenshot 2022-09-05 at 18 03 17

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -10,7 +10,6 @@ import (
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/jackc/pgx/v4"
-	"github.com/jackc/pgx/v4/log/zerologadapter"
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


jakedt:
I was thinking it might be interesting for this to re-write PGX logs to a different level. For example request logging should be transposed from INFO -> DEBUG. If we put this change in as is, it will require a code change to get request logs out of PGX, and those have proven useful for debugging in the past.

vroldanbet:
@jakedt good suggestion! I've updated the implementation to introduce an adapter function that maps pgx `info` events as `debug`.

`info` level now hides those:
Screenshot 2022-09-07 at 11 58 20

While `debug` shows them: 
Screenshot 2022-09-07 at 11 57 41

PR #799, opened 2022-09-06
remove dupe validation in WriteRelationships

vroldanbet:
I noticed in flamegraphs that validation was taking a chunk of the request time on large WriteRelationshipsRequest payloads, and that it was happening two times:
- one through the validation middleware
- one in `permissionServer.WriteRelationships` to detect dupes

Screenshot 2022-09-05 at 17 56 10

This was a regression from https://github.com/authzed/spicedb/pull/796

The commit improves the performance in various ways:
- removes the duplicated validation
- reduces CPU time to generate a v1.Relationship string in 85%
- reduces allocations to generate a v1.Relationship string in 100% (no allocations)

Output from `benchstat` from the new `StringRelationship` function. FWIW this is not going to have a dramatic impact on the request path, but given I was removing the duped validation, I decided to have a closer look for any low hanging fruit.

```
name                   old time/op    new time/op    delta
StringRelationship-16     516ns ± 2%      77ns ± 2%   -85.08%  (p=0.008 n=5+5)

name                   old alloc/op   new alloc/op   delta
StringRelationship-16      192B ± 0%        0B       -100.00%  (p=0.008 n=5+5)

name                   old allocs/op  new allocs/op  delta
StringRelationship-16      11.0 ± 0%       0.0       -100.00%  (p=0.008 n=5+5)
```

## Bonus: disable server logging during test

I also disabled server logs during tests in  https://github.com/authzed/spicedb/commit/62f1401a0075db3beceffcf3f4337916a986b643
Makes tests slightly faster with `-json` and with a less cluttered output

### DIFF
--- a/pkg/tuple/relationship.go
+++ b/pkg/tuple/relationship.go
@@ -1,0 +1,26 @@
+package tuple
+
+import (
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+)
+
+// StringObjectRef marshals a *v1.ObjectReference into a string.
+func StringObjectRef(ref *v1.ObjectReference) string {
+	return ref.ObjectType + ":" + ref.ObjectId
+}
+
+// StringSubjectRef marshals a *v1.SubjectReference into a string.
+func StringSubjectRef(ref *v1.SubjectReference) string {
+	if ref.OptionalRelation != "" {
+		return ref.Object.ObjectType + ":" + ref.Object.ObjectId + "#" + ref.OptionalRelation
+	}
+	return ref.Object.ObjectType + ":" + ref.Object.ObjectId
+}
+
+// StringRelationship converts a v1.Relationship to a string.
+func StringRelationship(rel *v1.Relationship) string {
+	if rel == nil || rel.Resource == nil || rel.Subject == nil {
+		return ""
+	}
+	return StringObjectRef(rel.Resource) + "#" + rel.Relation + "@" + StringSubjectRef(rel.Subject)
+}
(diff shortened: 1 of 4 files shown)


internal/datastore/common/revisions/optimized_test.go
```diff
@@ -31,6 +32,7 @@ var (
 )
 
 func TestOptimizedRevisionCache(t *testing.T) {
+	zerolog.SetGlobalLevel(zerolog.Disabled)
```
jakedt, line 35: This has always bothered me about zerolog. Is there no way to control the zerolog level without calling a global singleton? It leads to things like: "if you run these tests in this order you get no logs, if you run only one of the tests you get logs"

jakedt (changes requested):

pkg/tuple/relationship.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// StringObjectRef marshals a *v1.ObjectReference into a string.
+func StringObjectRef(ref *v1.ObjectReference) string {
+	return ref.ObjectType + ":" + ref.ObjectId
+}
+
+// StringSubjectRef marshals a *v1.SubjectReference into a string.
+func StringSubjectRef(ref *v1.SubjectReference) string {
+	if ref.OptionalRelation != "" {
```
josephschorr, line 14: Why the change from builder here?

vroldanbet: It's _faster_ than `fmt.Sprintf`, that's how I got it 85% down.

josephschorr: Right, but is it faster than the `string.Builder`?

internal/datastore/common/revisions/optimized_test.go
vroldanbet: Each logger instance can have its own logging level, for sure. I can give it another spin in a different PR so it does not block this one

pkg/tuple/relationship.go
vroldanbet: yep!

jakedt (approved):

PR #807, opened 2022-09-08
caveat in MemDB datastore redux

vroldanbet:
https://github.com/authzed/spicedb/issues/757

This PR introduces datastore interface to do CRUD operations with Caveats. We choose to start with MemDB because it does not require migrations, it's not meant to be used in production, and thus allows us to start building the graph evaluation logic without requiring modifications to all datastores. Once we've nailed its usage, we can go back to add support in the rest of datastores.

The main design choices are:
- Caveats have its own lifecycle independent from `Namespaces`
- They get a globally unique identifier so that we can continue referencing them even though name may change
- Name is a user-friendly way to reference a caveat. It will be used in the schema language to give caveats a name the schema developer can reference in their `definition`
- The datastore does not know anything about how caveats are implemented, and are simply stored as a sequence of bytes
- Tuples will reference caveats by its `ID`.
- Tuples may include a `caveat context`, which will be passed along to the caveat evaluator. Think input arguments of a function.
- The caveat package has evolved to serialize the caveat name with the proto representation. This is necessary for type checking in the schema.
- New interfaces are added to make datastores support caveats. They include the various CRUD operations. Writes are upserts, and caveats can be looked up by ID or name.
- Two caveats with the same name are considered the same caveat. There cannot be two caveats with distinct IDs and same Name


internal/datastore/memdb/caveat.go
```diff
@@ -0,0 +1,97 @@
...
+			caveatType: coreCaveat.Type,
+		}
+		if err := tx.Insert(tableCaveats, &c); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (rwt *memdbReadWriteTx) DeleteCaveat(caveats []*core.Caveat) error {
```
josephschorr, line 69: DeleteCaveats

internal/datastore/memdb/readwrite.go
```diff
@@ -87,6 +96,37 @@ func (rwt *memdbReadWriteTx) write(tx *memdb.Txn, mutations ...*core.RelationTup
 	return nil
 }
 
+func (rwt *memdbReadWriteTx) prepareCaveat(tx *memdb.Txn, mutation *core.RelationTupleUpdate) (*caveatReference, error) {
+	var cr *caveatReference
+	if mutation.Tuple.Caveat != nil {
```
josephschorr, line 101: So, my plan for this:

1. Every caveat written into the system gets a unique `uint64` ID
2. When the schema is compiled, a name -> ID map is written alongside the type information
3. When we perform type checking in the `WriteRelationships` call, we'll need to perform a lookup by name against the schema, which will give us back the ID
4. The caveat *ID* should be placed into the `RelationTupleUpdate` and then we can just use it directly here without further queries

internal/datastore/memdb/readwrite.go
```diff
@@ -35,13 +35,18 @@ func (rwt *memdbReadWriteTx) WriteRelationships(mutations []*core.RelationTupleU
...
+			return err
+		}
 		rel := &relationship{
 			mutation.Tuple.ResourceAndRelation.Namespace,
 			mutation.Tuple.ResourceAndRelation.ObjectId,
 			mutation.Tuple.ResourceAndRelation.Relation,
 			mutation.Tuple.Subject.Namespace,
 			mutation.Tuple.Subject.ObjectId,
 			mutation.Tuple.Subject.Relation,
+			cr,
```
josephschorr, line 49: This will need to store the context information written with the tuple as well; we'll need to add that the the `RelationTupleUpdate`

pkg/caveats/compile.go
```diff
@@ -48,6 +55,15 @@ type CompilationErrors struct {
 	issues *cel.Issues
 }
 
+func CompileCaveatWithName(env *Environment, exprString, name string) (*CompiledCaveat, error) {
+	c, err := CompileCaveat(env, exprString)
+	if err != nil {
+		return nil, err
+	}
+	c.name = name
```
josephschorr, line 63: We should probably store the name in the compiled caveat too

internal/datastore/memdb/readwrite.go
vroldanbet: @josephschorr that's already in place?

pkg/caveats/compile.go
vroldanbet: Yeah I didn't know how to proceed with the `pkg/caveats` bits so left it intentionally incomplete. Will look into that 👍🏻

internal/datastore/memdb/caveat.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (r *memdbReader) readCaveat(tx *memdb.Txn, name string) (datastore.CaveatIterator, error) {
+	it, err := tx.Get(tableCaveats, indexID, name)
+	if err != nil {
+		return nil, err
+	}
+	return &memdbCaveatIterator{it: it}, nil
+}
+
+func (rwt *memdbReadWriteTx) WriteCaveat(caveats []*core.Caveat) error {
```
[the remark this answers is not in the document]
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/807/commits/e9b09fb78f317316ed56808d8601cefbf111e490

internal/datastore/memdb/caveat.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/807/commits/e9b09fb78f317316ed56808d8601cefbf111e490

internal/datastore/memdb/readwrite.go
josephschorr: No, we'll need to decide how to store it.

internal/datastore/memdb/readwrite.go
vroldanbet: I've added caveat ID and lookup methods in https://github.com/authzed/spicedb/pull/807/commits/d2d7ce075cb08cd260011d79057aed85674df9f9

internal/datastore/memdb/caveat.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (c *caveat) CoreCaveat() *core.Caveat {
+	return &core.Caveat{
+		Name:       c.name,
+		Expression: c.expression,
+		Type:       c.caveatType,
+	}
+}
+
+func (r *memdbReader) ReadCaveat(name string) (datastore.CaveatIterator, error) {
```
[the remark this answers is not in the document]
vroldanbet: introduced here as string https://github.com/authzed/spicedb/pull/807/commits/d2d7ce075cb08cd260011d79057aed85674df9f9
then pivoted to uint64 https://github.com/authzed/spicedb/pull/807/commits/cc251ceb2995ef1cb0580796489ef8af39f90588

pkg/caveats/compile.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/807/commits/015507d6de0d01742c7e1ccb68e1093ff5979ed0

internal/datastore/memdb/readwrite.go
```diff
@@ -87,6 +97,25 @@ func (rwt *memdbReadWriteTx) write(tx *memdb.Txn, mutations ...*core.RelationTup
 	return nil
 }
 
+func (rwt *memdbReadWriteTx) prepareCaveat(tx *memdb.Txn, mutation *core.RelationTupleUpdate) (*caveatReference, error) {
+	var cr *caveatReference
+	if mutation.Tuple.Caveat != nil {
+		id := datastore.CaveatID(mutation.Tuple.Caveat.CaveatId)
```
josephschorr, line 103: As discussed, I think this can be done by the calling API layer

internal/datastore/memdb/schema.go
```diff
@@ -76,7 +97,11 @@ func (r relationship) Relationship() *v1.Relationship {
 	}
 }
 
-func (r relationship) RelationTuple() *core.RelationTuple {
+func (r relationship) RelationTuple() (*core.RelationTuple, error) {
+	cr, err := r.caveat.CaveatReference()
```
josephschorr, line 101: I wonder if it would be better to "parse" the caveat at the caller site; most relationships loaded won't necessarily need the fully struct-ed context, and constructing it here is going to allocate a lot more times that will be necessary. Instead, maybe we just RelationTuple store the raw form, and add a helper to retrieve it (which can return an error at *that* time)?

internal/datastore/memdb/readwrite.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/807/commits/307d92b61805024e0ea4231308b0db2b127ea375

internal/datastore/memdb/schema.go
vroldanbet: I'm not sure I fully grasp the idea. This method is used in a handful of places where it would seem like the exact representation of the `core.RelationTuple` is needed.
- It's used to [insert into the changelog table](https://github.com/vroldanbet/spicedb/blob/dff6f68808862f4fca561c3f3491921b6d763cac/internal/datastore/memdb/memdb.go#L197-L216) that powers `Watch` API, and hence we need the context returned back to the client. It's true that we could store the memdb-internal struct instead of the proto and avoid the internal-to-core conversion, but for some reason memdb is storing full blown proto values for the `changelog` but not for the `relationshipTuple` table. I could address that in a follow up, but seems a problem that was already in place before this PR
- it's used to implement the `RelationshipIterator`, which uses `core.RelationTuple`. Once again we would be returning the full Tuple representation in order to send it back to the client
- It's used in the `WriteRelationship` [path](https://github.com/vroldanbet/spicedb/blob/77b11ff10f71381c8db27bb1071b197a0ac00988/internal/datastore/memdb/readwrite.go#L70) to return a human readable error on duplicates. Not sure it's worth optimizing the error path.
- It's used to to generate the mutations needed [when a namespace is deleted](https://github.com/vroldanbet/spicedb/blob/77b11ff10f71381c8db27bb1071b197a0ac00988/internal/datastore/memdb/readwrite.go#L130). This is certainly an inefficient conversion given the system is [about to turns those again into their memdb-internal representation](https://github.com/vroldanbet/spicedb/blob/77b11ff10f71381c8db27bb1071b197a0ac00988/internal/datastore/memdb/readwrite.go#L38-L46) to insert then. This is a relatively easy refactor if you feel strongly about it. It also led me realize that we are not inserting in the changelog those delete tuples caused by a namespace deletion? Seems like a bug?

Overall it wasn't clear to me if optimization is a focus on the memdb implementation. I was under the impression it's meant to serve as reference implementation + testing environments. But happy to do any potential adjustments wherever it makes sense!

internal/datastore/memdb/caveat.go
```diff
@@ -0,0 +1,118 @@
...
+
+const tableCaveats = "caveats"
+
+type caveat struct {
+	id         datastore.CaveatID
+	name       string
+	expression []byte
+}
+
+func (c *caveat) CoreCaveat() *core.Caveat {
```
josephschorr, line 22: `CoreCaveat` is a big vague of a name... maybe `Unwrap()`?

internal/datastore/memdb/caveat.go
```diff
@@ -0,0 +1,118 @@
...
+	ids := make([]datastore.CaveatID, 0, len(caveats))
+	for _, coreCaveat := range caveats {
+		id := datastore.CaveatID(time.Now().UnixNano())
+		c := caveat{
+			id:         id,
+			name:       coreCaveat.Name,
+			expression: coreCaveat.Expression,
+		}
+		// TODO(vroldanbet) why does go-memdb not honor unique index name?
+		found, err := rwt.readCaveatByName(tx, coreCaveat.Name)
```
josephschorr, line 95: @jakedt any idea why this might be?

internal/datastore/memdb/caveat.go
```diff
@@ -0,0 +1,118 @@
...
+	defer rwt.Unlock()
+	tx, err := rwt.txSource()
+	if err != nil {
+		return nil, err
+	}
+	return rwt.writeCaveat(tx, caveats)
+}
+
+func (rwt *memdbReadWriteTx) writeCaveat(tx *memdb.Txn, caveats []*core.Caveat) ([]datastore.CaveatID, error) {
+	ids := make([]datastore.CaveatID, 0, len(caveats))
```
josephschorr, line 86: Are we handling revisions for caveats in here? They will need to be revisioned the same as namespaces, which I don't think are revisioned in memdb? @jakedt am I correct?

internal/datastore/memdb/caveat_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Fails to write dupes in the same transaction
+	_, _, err = writeCaveat(ctx, ds, coreCaveat, coreCaveat)
+	req.Error(err)
+
+	// Succeeds writing a caveat
+	rev, ID, err := writeCaveat(ctx, ds, coreCaveat)
+	req.NoError(err)
+
+	// fails to write caveat with the same name in different tx
```
josephschorr, line 37: Mind making all comments start with a capital letter?

internal/datastore/memdb/caveat_test.go
```diff
@@ -0,0 +1,142 @@
...
+	ctx := context.Background()
+
+	ds, err := NewMemdbDatastore(0, 1*time.Hour, 1*time.Hour)
+	req.NoError(err)
+	sds, _ := testfixtures.StandardDatastoreWithSchema(ds, req)
+
+	// store caveat, write caveated tuple and read back same value
+	coreCaveat := createCoreCaveat(t)
+	_, cavID, err := writeCaveat(ctx, ds, coreCaveat)
+	req.NoError(err)
```
josephschorr, line 73: newlines after the require statements

internal/datastore/memdb/schema.go
josephschorr: Its more the need to raise an error in the iterator, but I guess that's fine

pkg/caveats/compile.go
```diff
@@ -9,13 +9,22 @@ import (
...
+
 // CompiledCaveat is a compiled form of a caveat.
 type CompiledCaveat struct {
 	// env is the environment under which the CEL program was compiled.
 	celEnv *cel.Env
 
 	// ast is the AST form of the CEL program.
 	ast *cel.Ast
+
+	name string
```
josephschorr, line 22: doc comment

pkg/caveats/compile.go
```diff
@@ -9,13 +9,22 @@ import (
 	impl "github.com/authzed/spicedb/pkg/proto/impl/v1"
 )
 
+const anonymousCaveat = ""
```
josephschorr, line 12: doc comment

pkg/datastore/caveat.go
```diff
@@ -0,0 +1,26 @@
+package datastore
+
+import (
+	"errors"
+
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// ErrCaveatNotFound is the error returned when a caveat is not found by either ID or name
+var ErrCaveatNotFound = errors.New("caveat not found")
```
josephschorr, line 10: This should be a structured error, different for name vs ID, and include the name or ID in both the struct AND in the message

pkg/datastore/caveat.go
```diff
@@ -0,0 +1,26 @@
...
+
+// CaveatID serves as the system identifier for a persisted Caveat
+type CaveatID uint64
+
+// CaveatReader offers read operations for caveats
+type CaveatReader interface {
+	ReadCaveatByName(name string) (*core.Caveat, error)
```
josephschorr, line 17: Doc comments on the interface methods

proto/internal/core/v1/core.proto
```diff
@@ -12,6 +13,28 @@ message RelationTuple {
 
   /** subject is the subject for the tuple */
   ObjectAndRelation subject = 2 [ (validate.rules).message.required = true ];
+
+  /** referencing a caveat that must be enforced over the tuple **/
+  ContextualizedCaveat caveat = 3 [ (validate.rules).message.required = false ];
```
josephschorr, line 18: `caveat is a ...`

proto/internal/core/v1/core.proto
```diff
@@ -12,6 +13,28 @@ message RelationTuple {
...
+}
+
+message ContextualizedCaveat {
```
josephschorr, line 21: doc comments on the messages

proto/internal/core/v1/core.proto
```diff
@@ -12,6 +13,28 @@ message RelationTuple {
...
+  /** caveat_id is the reference to the internal id of a stored caveat **/
+  uint64 caveat_id = 1 [(validate.rules).uint64.gt = 0];
+
+  /** predefined are arguments used as input during caveat evaluation with a predefined value **/
```
josephschorr, line 25: `context is ...`

proto/internal/core/v1/core.proto
```diff
@@ -12,6 +13,28 @@ message RelationTuple {
...
+}
+
+message Caveat {
+  /** name represents the globally-unique identifier of the caveat **/
+  string name = 1 [ (validate.rules).string = {
+    pattern : "^(([a-zA-Z0-9_][a-zA-Z0-9/_|-]{0,127})|\\*)$",
+    max_bytes : 128,
+  } ];
+
+  /** expression is the byte representation of a caveat's logic */
```
josephschorr, line 36: I wonder if we should just make this `DecodedCaveat` and move it in here...

internal/datastore/memdb/caveat.go
vroldanbet: I followed the same convention used for `relationship` which has a method to create create a `v1.Relationship` named `Relationship`. I'm OK with `Unwrap` 👍🏻

internal/datastore/memdb/caveat.go
vroldanbet: As far as I can tell namespaces are not revisioned in MemDB

proto/internal/core/v1/core.proto
vroldanbet: honestly it's not clear to me what is the design rationale behind the `impl` proto package. It feels like both would overlap in responsibility.

pkg/datastore/caveat.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/807/commits/3ff40205f50abd9bfbc68c76ecd4ae7b3e1565b8

internal/datastore/memdb/caveat.go
jakedt: Honor in what way?

internal/datastore/memdb/caveat.go
jakedt: Everything is revisioned in memdb by keeping snapshots of the database after each mutation.

jakedt: https://github.com/authzed/spicedb/blob/main/internal/datastore/memdb/memdb.go#L224

internal/datastore/memdb/caveat.go
jakedt: https://github.com/hashicorp/go-memdb/issues/7

pkg/datastore/caveat.go
```diff
@@ -14,13 +14,18 @@ type CaveatID uint64
...
 	ReadCaveatByID(ID CaveatID) (*core.Caveat, error)
 }
 
 // CaveatStorer offers both read and write operations for Caveats
 type CaveatStorer interface {
 	CaveatReader
+	// WriteCaveats stores the provided caveats, and returns the assigned IDs
+	// Each element o the returning slice corresponds by possition to the input slice
 	WriteCaveats([]*core.Caveat) ([]CaveatID, error)
+	// DeleteCaveats deletes the provided caveats
```
josephschorr, line 26: newline before this comment and the one above

proto/internal/core/v1/core.proto
```diff
@@ -14,15 +14,19 @@ message RelationTuple {
   /** subject is the subject for the tuple */
   ObjectAndRelation subject = 2 [ (validate.rules).message.required = true ];
 
-  /** referencing a caveat that must be enforced over the tuple **/
+  /** caveat is a reference to a the caveat that must be enforced over the tuple **/
   ContextualizedCaveat caveat = 3 [ (validate.rules).message.required = false ];
 }
 
+/**
+* ContextualizedCaveat represents a reference to a caveat used to by caveated tuples.
```
josephschorr, line 22: space before the `*`'s

internal/datastore/memdb/caveat.go
josephschorr: @vroldanbet please update your comment to point to that issue

internal/datastore/mysql/readwrite.go
```diff
@@ -57,6 +57,9 @@ func (rwt *mysqlReadWriteTXN) WriteRelationships(mutations []*core.RelationTuple
 	// Process the actual updates
 	for _, mut := range mutations {
 		tpl := mut.Tuple
+		if tpl.Caveat != nil {
+			panic("caveats are not supported in MySQL datastore")
+		}
 		// Implementation for TOUCH deviates from PostgreSQL datastore to prevent a deadlock in MySQL
```
josephschorr, line 63: newline before this line

internal/datastore/crdb/readwrite.go
```diff
@@ -95,6 +95,10 @@ func (rwt *crdbReadWriteTXN) WriteRelationships(mutations []*core.RelationTupleU
 	// Process the actual updates
 	for _, mutation := range mutations {
 		rel := mutation.Tuple
+		if rel.Caveat != nil {
+			panic("caveats are not supported in CRDB datastore")
```
josephschorr, line 99: `not currently supported` here and below

internal/datastore/memdb/caveat.go
vroldanbet: ah, my bad! I was somehow looking for some explicit app level MVCC. MemDB handles this for us, thanks for the pointer Jake!

internal/datastore/memdb/caveat.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/807/commits/d104b717c7cbaf8e68247b265dabfd543f815c83

internal/datastore/memdb/caveat.go
vroldanbet: added a test to demonstrate snapshot reads for caveats: https://github.com/authzed/spicedb/pull/807/commits/b22266d54f0506b68c4a8290c98eeae7f5c7a71b

in order to support this I had to turn the caveat-write operation into an upsert

josephschorr (approved):
LGTM

PR #811, opened 2022-09-09
Disable GC in datastore drivers when in read-only mode

### LINKED ISSUE #738: SpiceDB tries to garbage-collect when `--datastore-readonly="true"`
I have some instances of SpiceDB using `--datastore-readonly="true"`. I am seeing a constant stream of errors "error when attempting to perform garbage collection" due to the transaction being readonly.

SpiceDB probably shouldn't attempt garbage collection when the datastore is in readonly.

image

josephschorr:
Fixes #738

### DIFF
--- a/internal/datastore/spanner/spanner.go
+++ b/internal/datastore/spanner/spanner.go
@@ -98,12 +98,16 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore, 
 	}
 	ds.RemoteClockRevisions.SetNowFunc(ds.HeadRevision)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	if err := ds.runGC(ctx); err != nil {
-		cancel()
-		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	if config.gcInterval > 0*time.Minute && config.gcEnabled {
+		ctx, cancel := context.WithCancel(context.Background())
+		if err := ds.runGC(ctx); err != nil {
+			cancel()
+			return nil, fmt.Errorf(errUnableToInstantiate, err)
+		}
+		ds.stopGC = cancel
+	} else {
+		log.Warn().Msg("datastore garbage collection disabled")
 	}
-	ds.stopGC = cancel
 
 	return ds, nil
 }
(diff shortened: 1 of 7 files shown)


internal/datastore/spanner/spanner.go
```diff
@@ -98,12 +98,16 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore,
 	}
 	ds.RemoteClockRevisions.SetNowFunc(ds.HeadRevision)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	if err := ds.runGC(ctx); err != nil {
-		cancel()
-		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	if config.gcInterval > 0*time.Minute && config.gcEnabled {
```
vroldanbet, line 101: Every GC callsite has the same condition `if datastore.gcInterval > 0*time.Minute && config.gcEnabled`. The `GCEnabled` configuration is currently set to `mysql.GCEnabled(!opts.ReadOnly)`. Why don't we make it instead `mysql.GCEnabled(!opts.ReadOnly && opts.gcInterval > 0*time.Minute)` so that these callsites only need to do `if config.gcEnabled`?

vroldanbet (commented):
I'd suggest refactoring the condition that enables GC so that callsites don't need to deal with the `GCInterval` implementation detail.

I also wonder in which circumstances do we want to allow  `GCInterval` to be set to zero, which effectively disables GC. Other than disabling it in read-only mode, in which other scenarios is disabling GC a good idea?

internal/datastore/spanner/spanner.go
josephschorr: Because nothing precludes a caller from also setting the interval manually to 0 time, and we should either raise an error or disable GC in that case (which is, in either case, a check logic)

vroldanbet: yeah that's what I mean: we can assume `0` as interval should be checked earlier on process start and that it's not a valid value.

josephschorr: I tend to still want to have the check within the handlers, to be safe

PR #824, opened 2022-09-16
Use MarshalVT() and UnmarshalVT() everywhere

jzelinskie:
This drops the few remaining places where we used reflection-based protobuf encoding/decoding.

I experimented with adding pooling for namespaces, but lifetimes get tricky with our caching layer. The caching layer should already be getting us most of the benefit of having a pool already.


josephschorr (approved):
LGTM

PR #825, opened 2022-09-17
internal/datastore: use proto.Marshal

jzelinskie:
MarshalVT does not handle empty fields the same.

The following marshals two different ways:

&pb.AllowedRelation{
  RelationOrWildcard: &pb.AllowedRelation_PublicWildcard_{},
}

protobuf: []byte{0x22, 0x0}
vtprotobuf: []byte{}


josephschorr (approved):
LGTM

PR #826, opened 2022-09-17
Add a test for writing and reading back a serialized namespace in the datastores

josephschorr:
This test ensures that the marshal and unmarshal code work together properly for namespaces across all datastores

This is important because we store the namespaces as serialized proto bytes in the datastore namespace tables

### DIFF
--- a/pkg/datastore/test/namespace.go
+++ b/pkg/datastore/test/namespace.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"testing"
 
+	"google.golang.org/protobuf/proto"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/testing/protocmp"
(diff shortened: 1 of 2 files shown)


pkg/datastore/test/namespace.go
```diff
@@ -175,3 +180,49 @@ func EmptyNamespaceDeleteTest(t *testing.T, tester DatastoreTester) {
...
+	schemaString := `definition document {
+	relation viewer: user | user:*
+	relation editor: user | group#member
+	relation parent: organization
+	permission edit = editor
+	permission view = viewer + edit + parent->view
+	permission other = viewer - edit
+	permission intersect = viewer & edit
+	permission with_nil = ((viewer - edit) & parent->view) & nil
+}`
```
jzelinskie, line 198: this could be table-driven with a few different schemas

josephschorr: Could be, but this schema basically encompasses the entire set of operations we currently support. Once we extend schema further, I'd be fine with making this table driven

jzelinskie (approved):

PR #833, opened 2022-09-21
Add building of WASM developer binary to releases

samkim:
Add a workflow that builds SpiceDB as wasm and commits it to `dist`

### DIFF
--- a/.github/workflows/wasm.yaml
+++ b/.github/workflows/wasm.yaml
@@ -1,0 +1,27 @@
+---
+name: "Build WASM"
+on: # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - "main"
+    paths:
+      - "pkg/development/wasm/**"
+      - ".github/workflows/wasm.yaml"
+jobs:
+  build:
+    name: "Build WASM"
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: "actions/checkout@v3"
+      - uses: "actions/setup-go@v3"
+        with:
+          go-version: "~1.19.1"
+      - run: |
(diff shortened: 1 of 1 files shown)


jzelinskie:
Checking binaries into a git repo makes it super slow. There's git-lfs or git-annex which I is intended to track large files in git, but store them on S3. GitHub actions also has an artifact upload action that we could potentially use (build logs and stuff are often stored here).

.github/workflows/wasm.yaml
```diff
@@ -0,0 +1,27 @@
+---
+name: "Build WASM"
+on: # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - "main"
+    paths:
+      - "pkg/development/wasm/**"
```
josephschorr, line 8: This path isn't sufficient, since if we make a development package change, we need to rebuild as well. We likely need to have any paths, except the `dist` directory

samkim:
> Checking binaries into a git repo makes it super slow. There's git-lfs or git-annex which I is intended to track large files in git, but store them on S3. GitHub actions also has an artifact upload action that we could potentially use (build logs and stuff are often stored here).

The end goal for committing the binary is to have it automatically available for cdn proxies like jsdelivr. The wasm file is fairly large so it definitely will add latency when interacting with the repo. If we think it's not worth the tradeoff, we could also just build the wasm file for releases and store as release artifacts.

samkim:
@josephschorr updated the build output name and did a test run of the workflow https://github.com/samkim/spicedb/releases/tag/v1.12.0-wasm

josephschorr (approved):
LGTM

PR #837, opened 2022-09-21
Fix performance for large schema writes in V1Alpha1

josephschorr:
This will hopefully reduce the flakiness of the E2E test suite, which writes large batches of definitions via the V1Alpha1 WriteSchema call

### DIFF
--- a/internal/datastore/crdb/reader.go
+++ b/internal/datastore/crdb/reader.go
@@ -112,6 +112,37 @@ func (cr *crdbReader) ListNamespaces(ctx context.Context) ([]*core.NamespaceDefi
 	return nsDefs, nil
 }
 
+func (cr *crdbReader) LookupNamespaces(ctx context.Context, nsNames []string) ([]*core.NamespaceDefinition, error) {
+	if len(nsNames) == 0 {
+		return nil, nil
+	}
+
+	ctx = datastore.SeparateContextWithTracing(ctx)
+
+	var nsDefs []*core.NamespaceDefinition
+	if err := cr.execute(ctx, func(ctx context.Context) error {
+		tx, txCleanup, err := cr.txSource(ctx)
+		if err != nil {
+			return err
+		}
+		defer txCleanup(ctx)
+
+		nsDefs, err = lookupNamespaces(ctx, tx, nsNames)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf(errUnableToListNamespaces, err)
+	}
+
+	for _, nsDef := range nsDefs {
+		cr.addOverlapKey(nsDef.Name)
+	}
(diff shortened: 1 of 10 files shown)


internal/testfixtures/validating.go
```diff
@@ -64,6 +64,25 @@ func (vsr validatingSnapshotReader) ListNamespaces(
 	return read, err
 }
 
+func (vsr validatingSnapshotReader) LookupNamespaces(
+	ctx context.Context,
+	nsNames []string,
+) ([]*core.NamespaceDefinition, error) {
+	read, err := vsr.delegate.LookupNamespaces(ctx, nsNames)
+	if err != nil {
+		return read, err
```
vroldanbet, line 73: nit: may confuse the reader, even though `read` should be `nil`
```suggestion
		return nil, err
```

internal/testfixtures/validating.go
```diff
@@ -64,6 +64,25 @@ func (vsr validatingSnapshotReader) ListNamespaces(
...
+	}
+
+	for _, nsDef := range read {
+		err := nsDef.Validate()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return read, err
```
vroldanbet, line 83:
```suggestion
	return read, nil
```

internal/datastore/crdb/reader.go
```diff
@@ -112,6 +112,37 @@ func (cr *crdbReader) ListNamespaces(ctx context.Context) ([]*core.NamespaceDefi
...
+		}
+
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf(errUnableToListNamespaces, err)
+	}
+
+	for _, nsDef := range nsDefs {
+		cr.addOverlapKey(nsDef.Name)
+	}
```
vroldanbet, line 142: for very large namespaces (presumably the ones causing the flakiness we are attempting to fix in this PR) this would lead to _many_ "touches" in the transactions table:

https://github.com/authzed/spicedb/blob/ffdf5210c7a97ae50b3f3d06070c04c4e7ea9e15/internal/datastore/crdb/crdb.go#L260-L264

@ecordell is there are reason we don't batch all those query operations instead of looping over the keyset?

vroldanbet (approved):
LGTM. Left 2 nit suggestions, and a potential follow up to how we handle overlapping keys in CRDB

vroldanbet:
ugh sorry I didn't notice auto-merged was enabled 😭

internal/datastore/crdb/reader.go
ecordell: No real reason other than almost all requests should have only 1 overlap key

vroldanbet: Awesome, I think it would be relatively easy to turn that into a batch pgx request

Issue #836, opened 2022-09-21
add caveats in the WriteRelationships/ReadRelationships API


vroldanbet:
Tracking: https://github.com/authzed/spicedb/issues/386

We should adjust the `WriteRelationships` and `ReadRelationships` API to make it possible to store caveated tuples

Issue #386, opened 2022-01-25
Proposal: Caveated relationships


jakedt:
# Goal

Let people handle their abac-y style needs in a Zanzibar first way.

# Use Cases

- Time
  - “I want to bound the amount of time that a user has this role”
  - “The cleaning staff should only be allowed to access the office between 5pm and 10pm on weekdays”
- IP Address
  - “You can only access this site if you’re not in N.K.”
- Money
  - “This user is an approver if the order value is < $50”

# Proposal

Allow small fragments of policy to be associated with individual relationships in a new field called “caveats”. As we attempt to evaluate permissions these pieces of policy will be combined and surfaced as immutable caveats that apply to the result sets as they are collected. Before the result is returned to the user, a final policy is assembled and evaluated against user-supplied attributes, and a final decision is made.

Because the caveats are immutable and apply to the sub-problem, they can be cached at every level of the decision making process.

# Examples

### Schema and Relationships

```
definition user {}
 
definition document {
   relation writer: user
   relation reader: user
 
   permission edit = writer
   permission view = reader + edit
}
```

```
document:firstdoc#writer@user:tom[day($date) == Monday]
document:firstdoc#reader@user:tom
```

### Resolution

- check::document:firstdoc#view@user:tom
  - check::document:firstdoc#reader@user:tom
    - <- true
  - check::document:firstdoc#edit@user:tom
    - check::document:firstdoc#writer@user:tom
      - <- true[day($date) == Monday]
    - <- true[day($date) == Monday]
  - <- true <- Union(true, true[day($date) == Monday])
- check::document:firstdoc#edit@user:tom
  - check::document:firstdoc#edit@user:tom
    - check::document:firstdoc#writer@user:tom
      - <- true[day($date) == Monday]
    - <- true[day($date) == Monday]
  - <- true[day($date) == Monday]

### Schema and Relationships

```
definition user {}
 
definition document {
   relation banned: user
   relation reader: user
 
   permission view = reader - banned
}
```

```
document:firstdoc#banned@user:tom[day($date) == Monday]
document:firstdoc#reader@user:tom
document:firstdoc#reader@user:jerry
```

### Resolution

- check::document:firstdoc#view@user:tom
  - check::document:firstdoc#reader@user:tom
    - <- true
  - check::document:firstdoc#banned@user:tom
    - <- true[day($date) == Monday]
  - <- true[!(day($date) == Monday)] <- Exclusion(true, true[day($date) == Monday])
[...]

PR #838, opened 2022-09-22
introduce caveat support in WriteRelationships/ReadRelationships

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/836
Depends on https://github.com/authzed/api/pull/38
Depends on https://github.com/authzed/authzed-go/pull/71

## What

Introduces support for writing caveated relationships via `WriteRelationships` API

## How

The duality `caveat_id`/`caveat_name` was problematic to handle between the controller layer and the datastore layer, so it was adjusted.

The usage of ID introduced a situation where a `core.RelationTuple` cannot be converted to a `v1.Relationship` without a database lookup, because database tuples reference caveats via ID, but we return named caveats to the client via `v1.Relationship`. It would be a departure from what happens today in service controllers: v1-to-core would introduce I/O operations.

☝🏻  is certainly not a deal breaker we couldn't addres, but the question here is that it's not clear the usage of ID buys us anything at this point. The rationale behind them was being able to support "anonymous caveats". An anonymous caveat signature defined in place in a relation in the schema. Because they don't have a name, we need a way to identify them.

However, it's not clear anonymous caveats is needed at this point, and that we should work on it when the need arises.

Note: because caveat is not yet exposed via `Schema` API, I resourced to writing it directly via the datastore, using type assertions.  

## Note on failing buf build

I intentionally broke `core` proto definition. These haven't been yet released and are not being used so we can ignore that error:

```
Error: Field "1" on message "ContextualizedCaveat" changed type from "uint64" to "string". See https://developers.google.com/protocol-buffers/docs/proto3#updating for wire compatibility rules.
Error: buf found 1 breaking changes.
```


internal/datastore/memdb/caveat.go
```diff
@@ -36,31 +32,8 @@ func (r *memdbReader) ReadCaveatByName(name string) (*core.Caveat, error) {
...
-	if found == nil {
-		return nil, datastore.NewCaveatIDNotFoundErr(ID)
-	}
-	c := found.(*caveat)
-	return c.Unwrap(), nil
-}
-
 func (r *memdbReader) readCaveatByName(tx *memdb.Txn, name string) (*caveat, error) {
-	found, err := tx.First(tableCaveats, indexName, name)
+	found, err := tx.First(tableCaveats, indexID, name)
```
josephschorr, line 36: shouldn't the index be `indexName`?

internal/services/v1/relationships_test.go
```diff
@@ -331,6 +336,56 @@ func TestWriteRelationships(t *testing.T) {
...
+	})
+	req.NoError(err)
+
+	// should succeed
+	resp, err := client.WriteRelationships(context.Background(), writeReq)
+	req.NoError(err)
+
+	// read relationship back
+	relRead := readFirst(req, client, resp.WrittenAt, relWritten)
+	foundDiff := cmp.Diff(relWritten, relRead, protocmp.Transform())
```
josephschorr, line 385: Could also use `proto.Equal`

vroldanbet: That's not the convention used in the codebase. It uses `indexID` because it's the "ID" column for `hashicorp/go-memdb`, which is required. You can think of it as the "PK" constraint

internal/datastore/memdb/caveat.go
vroldanbet: That's not the convention used in the codebase. It uses `indexID` because it's the "ID" column for `hashicorp/go-memdb`, which is required. You can think of it as the "PK" constraint

josephschorr: That's not confusing at all! /s

internal/services/v1/relationships_test.go
```diff
@@ -331,6 +335,58 @@ func TestWriteRelationships(t *testing.T) {
...
+		}},
+	}
+
+	// Should fail due to non-existing caveat
+	ctx := context.Background()
+	_, err = client.WriteRelationships(ctx, writeReq)
+	grpcutil.RequireStatus(t, codes.FailedPrecondition, err)
+	errStatus, ok := status.FromError(err)
+	req.True(ok, "could not fetch status from grpc error")
+	req.Equal("the provided caveat 'test' does not exist", errStatus.Message())
```
josephschorr, line 368: This error message should indicate the relationship as well in its message

josephschorr (approved):
LGTM

PR #840, opened 2022-09-22
Fix chunking util to never call for an empty chunk

josephschorr:
An empty chunk was being returned when the chunk size and overall size hit the same border. This could hit a panic in the new chunked check dispatch code


ecordell (approved):
LGTM

vroldanbet (approved):

jzelinskie:
and this is why you only ever have one implementation of chunking folks!

josephschorr:
> and this is why you only ever have one implementation of chunking folks!

We do! It just wasn't handling this case :P

PR #843, opened 2022-09-24
Batch check for lookup

josephschorr:
Changes the lookup resources' `parallelChecker` to use batch checking for non-union results.

This should significantly improve the performance of larger data sets that need to be checked, as instead of spinning up a single check for each returned result (up to some maximum number of concurrent checks), a batch check is issued with many results to be checked concurrently (if available)

Performance comparison of a `LookupResources` call which finds 500 results that need to be checked due to an intersection at the top-level:

```
name                                                 old time/op    new time/op    delta
Services/lookup_with_intersection/cockroachdb/v1-10     3.13s          0.12s        
Services/lookup_with_intersection/mysql/v1-10           1.16s          0.07s        
Services/lookup_with_intersection/postgres/v1-10        709ms           30ms        
Services/lookup_with_intersection/spanner/v1-10         42.6s           2.8s        

name                                                 old alloc/op   new alloc/op   delta
Services/lookup_with_intersection/cockroachdb/v1-10     128MB           10MB        
Services/lookup_with_intersection/mysql/v1-10           126MB           15MB        
Services/lookup_with_intersection/postgres/v1-10        135MB           13MB        
Services/lookup_with_intersection/spanner/v1-10         235MB           16MB        

name                                                 old allocs/op  new allocs/op  delta
Services/lookup_with_intersection/cockroachdb/v1-10     2.01M          0.09M        
Services/lookup_with_intersection/mysql/v1-10           2.04M          0.09M        
Services/lookup_with_intersection/postgres/v1-10        2.10M          0.09M        
Services/lookup_with_intersection/spanner/v1-10         2.81M          0.13M        
```

> **Note**
> The benchmarks were generated by running the various datastores in containers on a local M1, which likely explains why Postgres is so much faster: it is running under native `arm`. MySQL, CRDB, and Spanner are `amd64` images and therefore likely much slower than a real-world installation


internal/graph/parallelchecker.go
```diff
@@ -34,101 +35,139 @@ type ParallelChecker struct {
...
-			Namespace: resource.Namespace,
-			Relation:  resource.Relation,
-		},
-		ResourceIds: []string{resource.ObjectId},
-		Subject:     pc.subject,
-	}
+	pc.toCheck <- resourceID
 }
 
 // Start starts the parallel checks over those items added via QueueCheck.
```
vroldanbet, line 117: `QueueToCheck`

internal/graph/parallelchecker.go
```diff
@@ -142,11 +181,11 @@ func (pc *ParallelChecker) Start() {
 // Wait waits for the parallel checker to finish performing all of its
 // checks and returns the set of resources that checked, along with whether an
 // error occurred. Once called, no new items can be added via QueueCheck.
```
vroldanbet, line 183: `QueueToCheck`

internal/graph/lookup.go
```diff
@@ -94,7 +85,7 @@ func (cl *ConcurrentLookup) LookupViaReachability(ctx context.Context, req Valid
 	cancelCtx, checkCancel := context.WithCancel(ctx)
 	defer checkCancel()
 
-	checker := NewParallelChecker(cancelCtx, cl.c, req.Subject, cl.concurrencyLimit)
+	checker := newParallelChecker(cancelCtx, cl.c, req, cl.concurrencyLimit)
 	stream := &collectingStream{checker, req, cancelCtx, 0, 0, 0, sync.Mutex{}}
 
 	// Start the checker.
```
vroldanbet, line 91: To add in the comment below: clarify that `DispatchReachableResources` is blocking until the all elements are streamed (and thus there is no race with `Wait` closing the channel)

vroldanbet (approved):
This is good stuff Joey! LGTM! Left a few comments

internal/graph/lookup.go
josephschorr: Did so and have the parallel checker cancel the context if it hits the requested number of results

josephschorr: Did so and have the parallel checker cancel the context if it hits the requested number of results

Issue #753, opened 2022-08-18
revisit logs and levels


vroldanbet:
The spicedb log output is fairly verbose at the info level. We should revisit our log output and improve the operator experience:
- all requests are reported at `info` level
- some grpc operations with specific codes are being reported at level `warn` (example `FailedPrecondition` when someone tries to interact with a schema type that does not exist).

jzelinskie:
Another user requested that health checking specifically be a higher verbosity than other requests. Perhaps it makes sense it only have that be `debug`.

vroldanbet:
@jzelinskie we can easily remove logging the health endpoint, `grpclog` allows customizing that via `WithDecider` which we already have set up. Unfortunately what I cannot do is change the level for a specific log entry. Another option is "log at info only if code is != `OK`"

If none of those seem like a reasonable compromise, and if I'm right the module does not give us the customization we need, the option would be to add a custom gRPC interceptor around it.

JoaoSouMoreira:
As a user I'd be happy with either option above, our issue is that the health endpoint log is incredibly noisy and obfuscates any other logs coming out of the serve command.
As a sidenote, would it be difficult to set up a way for users to specify a log filter regex based on the`grpc.service` property? That way users are a bit more free with what they choose to log or not.

gunpuz:
Hi guys, any update?

There seem to be too many logs out there with information level
For example Health probes and usual permission check requests. 
These could be debug level logs.

```JSON
{"level":"info","requestID":"0a4957f3bdb2d9956610a31b7523f1da","protocol":"grpc","grpc.component":"server","grpc.service":"grpc.health.v1.Health","grpc.method":"Check","grpc.method_type":"unary","peer.address":"[::1]:43620","grpc.start_time":"2023-10-09T15:02:06Z","grpc.request.deadline":"2023-10-09T15:02:07Z","grpc.code":"OK","grpc.time_ms":"0.14","time":"2023-10-09T15:02:06Z","message":"finished call"}
[...]

PR #844, opened 2022-09-26
streamline logging

vroldanbet:
Part of https://github.com/authzed/spicedb/issues/753
Built on top of https://github.com/jzelinskie/cobrautil/pull/19

# What

Adjustments to homogenize and streamline how logging is used in SpiceDB.

# How

Not all logging messages have been revisited. That will be tackled in a follow up. This PR sets the foundation.

- migrates to cobrautil v2
- makes it possible to stop relying on zerolog global logger, and thus be able to have our own default logger (disabled) 
- configures logger to be non-blocking
- configures same logger for OTel library
- logging is now disabled by default in our tests (no more need to do `zerolog.SetGlobalLevel(zerolog.Disabled)` in each test)
- `zerolog.Ctx()` is used in the codebase frequently, but because [we don't set the default logger for contet-based logging](https://github.com/rs/zerolog/blob/master/globals.go#L108-L110). We are now setting `zerolog.DefaultContextLogger` whenever our own global logger is set.
- [add missing Close on http downloader server and log error](https://github.com/authzed/spicedb/pull/844/files#diff-cd758728458ab57fbaf0644ea41b2c62aeb417a789098b6cf3bd20316b95030cR110-R112)

The changeset is relatively simple, it's just that lots of imports had to be adjusted. Files to look into are:
- https://github.com/authzed/spicedb/pull/844/files#diff-a47342e8d77d583a1cdd9b90f604b523ddfc98b1135231779f8b8d3de2de32cc
- https://github.com/authzed/spicedb/pull/844/files#diff-bd140be825bc795319d8e428c19dd9c6c726167777d27b346718f52b801a368b
- https://github.com/authzed/spicedb/pull/844/files#diff-df0635588ca5d6a577487faacc185c3838a3fe2b321660c36372541797c54c74

### Example startup at `INFO` level

image

### Example startup at `DEBUG` level

image


internal/logging/logger.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+
+	"github.com/rs/zerolog"
+)
+
+var Logger zerolog.Logger
+
+func init() {
+	nop := zerolog.Nop()
+	SetGlobalLogger(nop)
```
jzelinskie, line 14: why not `SetGlobalLogger(zerolog.Nop())`?

internal/logging/logger.go
```diff
@@ -0,0 +1,72 @@
...
+	return Logger.Fatal()
+}
+
+func WithLevel(level zerolog.Level) *zerolog.Event {
+	return Logger.WithLevel(level)
+}
+
+func Log() *zerolog.Event {
+	return Logger.Log()
+}
```
jzelinskie, line 60: You can actually make these all one line each and gofmt won't complain:

```go
func With() zerolog.Context { return Logger.With() }
func Err(err error) *zerolog.Event { return Logger.Err(err) }
func Trace() *zerolog.Event { return Logger.Trace() }
func Debug() *zerolog.Event { return Logger.Debug() }
func Info() *zerolog.Event { return Logger.Info() }
func Warn() *zerolog.Event { return Logger.Warn() }
func Error() *zerolog.Event { return Logger.Error() }
func Fatal() *zerolog.Event { return Logger.Fatal() }
func WithLevel(level zerolog.Level) *zerolog.Event { return Logger.WithLevel(level) }
func Log() *zerolog.Event { return Logger.Log() }
```

pkg/cmd/root.go
```diff
@@ -1,16 +1,17 @@
...
 
 	"github.com/authzed/spicedb/pkg/cmd/server"
 	"github.com/authzed/spicedb/pkg/releases"
 )
 
 func RegisterRootFlags(cmd *cobra.Command) {
-	cobrautil.RegisterZeroLogFlags(cmd.PersistentFlags(), "log")
-	cobrautil.RegisterOpenTelemetryFlags(cmd.PersistentFlags(), "otel", cmd.Use)
+	cobrazerolog.New().RegisterFlags(cmd.PersistentFlags())
+	cobraotel.New("spicedb").RegisterFlags(cmd.PersistentFlags())
```
jzelinskie, line 14: I think we might prefer to keep this cmd.Use so that the binary name is not hard coded

jzelinskie (changes requested):
Only minor nitpicks

pkg/cmd/root.go
vroldanbet: ah! good catch!

internal/logging/logger.go
vroldanbet: leftover from it being a pointer. Instafix!

jzelinskie (approved):

Issue #846, opened 2022-09-28
Add caveats to the dispatch layer


vroldanbet:
Tracking: https://github.com/authzed/spicedb/issues/386

Now that we have caveats at the datastore layer, we can start integrating it in the dispatch layer

vroldanbet:
Fixed via https://github.com/authzed/spicedb/pull/865

PR #849, opened 2022-09-28
Start work for dispatch for caveats by adding MembershipSet

josephschorr:
The MembershipSet helper type provides helper methods for easily combining the results of downstream dispatch checks to determine the membership of resources for which determined results (MEMBERSHIP) or caveated results (CAVEAT_MEMBERSHIP) apply, along with the caveat expressions and their context(s)

First part of #846


internal/graph/membershipset.go
```diff
@@ -0,0 +1,254 @@
...
+		existing, ok := ms.membersByID[resourceID]
+		if !ok {
+			continue
+		}
+
+		ms.membersByID[resourceID] = caveatAnd(existing, details.Expression)
+	}
+
+	hasDeterminedMember := false
+	for resourceID, expr := range ms.membersByID {
```
vroldanbet, line 103: what if we move this block up, just handling the deletions of the items that do not exist in the input map at the beginning of this method, and do the  loop over `resultsMap`: we can handle the `hasDeterminedMember` logic in the same if block, so you can do optimizations like not calling `caveatAnd`, just like you did in `addMember`. As it stands, one has to bear with the cognitive load that `caveatAnd` has some optimizations that would equal to "noop" assignment to the map.

```go
func (ms *MembershipSet) IntersectWith(resultsMap CheckResultsMap) {
	for resourceID, _ := range ms.membersByID {
		if _, ok := resultsMap[resourceID]; !ok {
			delete(ms.membersByID, resourceID)
		}
	}

	ms.hasDeterminedMember = false
	for resourceID, details := range resultsMap {
		existing, ok := ms.membersByID[resourceID]
		if !ok {
			continue
		}
		if existing == nil && details.Expression == nil {
			ms.hasDeterminedMember = true
			continue
		}

		ms.membersByID[resourceID] = caveatAnd(existing, details.Expression)
	}
}
```

internal/graph/membershipset.go
```diff
@@ -0,0 +1,254 @@
...
+		OperationOrCaveat: &v1.CaveatExpression_Operation{
+			Operation: &v1.CaveatOperation{
+				Op:       v1.CaveatOperation_OR,
+				Children: []*v1.CaveatExpression{first, second},
+			},
+		},
+	}
+}
+
+func caveatAnd(first *v1.CaveatExpression, second *v1.CaveatExpression) *v1.CaveatExpression {
```
vroldanbet, line 206: I guess that while doing those logical operations at CEL level is plausible, it would make things more complicated to test as we need to introspect the resulting CEL. Another consideration is we would need to load the caveat from database earlier, we may be able to optimize further if we batch load those once we need to resolve the chain of caveats. But if we need to load the tuples, we may as well join it with the caveats tuple so that wouldn't be a concern.

Did you consider handle the logical operations at CEL level, and if so, why not?

internal/graph/membershipset.go
```diff
@@ -0,0 +1,254 @@
...
+		ms.membersByID[resourceID] = caveatSub(existing, details.Expression)
+	}
+
+	ms.hasDeterminedMember = false
+	for _, expr := range ms.membersByID {
+		if expr == nil {
+			ms.hasDeterminedMember = true
+			break
+		}
+	}
```
vroldanbet, line 143: We can avoid doing two loops i we iterate on the `membersByID` first:

```go
	ms.hasDeterminedMember = false
	for resourceID, expression := range ms.membersByID {
		if details, ok := resultsMap[resourceID]; ok {
			// If the incoming member has no caveat, then this removal is absolute.
			if details.Expression == nil {
				delete(ms.membersByID, resourceID)
				continue
			}

			// Otherwise, the caveat expression gets combined with an intersection of the inversion
			// of the expression.
			ms.membersByID[resourceID] = caveatSub(expression, details.Expression)
		} else {
			if expression == nil {
				ms.hasDeterminedMember = true
			}
		}
	}
```

vroldanbet (commented):
This is looking good, I left some questions. I'd like to see clarified _when_ in the dispatch chain are caveats evaluated. The code here would seem to suggest we want to accumulate caveated membership first and then evaluate, instead of evaluating at the leafs.

internal/graph/membershipset.go
```diff
@@ -0,0 +1,254 @@
...
+	if caveat == nil {
+		return nil
+	}
+
+	return &v1.CaveatExpression{
+		OperationOrCaveat: &v1.CaveatExpression_Caveat{
+			Caveat: caveat,
+		},
+	}
+}
```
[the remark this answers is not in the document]
josephschorr: `CaveatExpression` is only defined within the dispatch layer, so it should remain there IMO

internal/graph/membershipset.go
```diff
@@ -0,0 +1,254 @@
...
+
+func caveatOr(first *v1.CaveatExpression, second *v1.CaveatExpression) *v1.CaveatExpression {
+	if first == nil {
+		return second
+	}
+
+	if second == nil {
+		return first
+	}
```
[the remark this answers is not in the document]
josephschorr: I feel better having it applied in all three of the combination methods; I can add a test if necessary

internal/graph/membershipset.go
josephschorr: Its a tradeoff: loading the caveats in the relationship loader and combining them via CEL means we might get some early termination (depending on the data already found for the relationship), but it does mean carrying around a lot more data if the expressions are larger. It also makes testing of the MembershipSet much harder and binds us directly to CEL. We may very well change over to it, but this approach seemed simpler for the initial pass.

josephschorr:
Updated

vroldanbet (approved):

PR #852, opened 2022-09-29
Improves Postgres write performance

jakedt:
Previously Postgres was limited to 1 concurrent write request globally. With this change we leverage some of the transaction isolation primitives introduced in Postgres 13 to extend the effective repeatable read lifetime beyond the transaction's lifetime. With this we can leverage internal Postgres transaction IDs and snapshots to greatly improve the performance of our MVCC implementation.

**Contains a 4-phase migration!**

### DIFF
--- a/internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
+++ b/internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
@@ -1,0 +1,90 @@
+package migrations
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
+)
+
+const (
+	addTransactionXIDColumns = `
+		ALTER TABLE relation_tuple_transaction
+			ADD COLUMN xid xid8 NOT NULL DEFAULT (pg_current_xact_id()),
+			ADD COLUMN snapshot pg_snapshot`
+
+	addTupleXIDColumns = `
+		ALTER TABLE relation_tuple
+			ADD COLUMN created_xid xid8,
+			ADD COLUMN deleted_xid xid8 NOT NULL DEFAULT ('9223372036854775807');`
+
+	addNamespaceXIDColumns = `
+		ALTER TABLE namespace_config
+			ADD COLUMN created_xid xid8,
+			ADD COLUMN deleted_xid xid8 NOT NULL DEFAULT ('9223372036854775807');`
+
+	addTransactionDefault = `
+		ALTER TABLE relation_tuple_transaction
+			ALTER COLUMN snapshot SET DEFAULT (pg_current_snapshot());`
+
+	addRelationTupleDefault = `
+		ALTER TABLE relation_tuple
+			ALTER COLUMN created_xid SET DEFAULT (pg_current_xact_id());`
+
+	addNamepsaceDefault = `
+		ALTER TABLE namespace_config
+			ALTER COLUMN created_xid SET DEFAULT (pg_current_xact_id());`
+
+	backfillTransactionXids = `
+		UPDATE relation_tuple_transaction SET xid = id::text::xid8, snapshot = CONCAT(id, ':', id, ':')::pg_snapshot
+			WHERE snapshot IS NULL;`
+
+	backfillTupleXIDColumns = `
+		UPDATE relation_tuple
+			SET deleted_xid = deleted_transaction::text::xid8,
+			created_xid = created_transaction::text::xid8
+			WHERE created_xid IS NULL;`
+
+	backfillNamespaceXIDColumns = `
+		UPDATE namespace_config
+			SET deleted_xid = deleted_transaction::text::xid8,
+			created_xid = created_transaction::text::xid8
+			WHERE created_xid IS NULL;`
+
+	addTransactionSnapshotNotNull = `
+		ALTER TABLE relation_tuple_transaction ALTER COLUMN snapshot SET NOT NULL;`
+
+	addTupleCreatedNotNull = `
+		ALTER TABLE relation_tuple ALTER COLUMN created_xid SET NOT NULL;`
+
+	addNamespaceCreatedtNotNull = `
+		ALTER TABLE namespace_config ALTER COLUMN created_xid SET NOT NULL;`
+)
+
+func init() {
+	if err := DatabaseMigrations.Register("add-xid-columns", "add-ns-config-id",
+		noNonatomicMigration,
+		func(ctx context.Context, tx pgx.Tx) error {
+			for _, stmt := range []string{
+				addTransactionXIDColumns,
+				addTupleXIDColumns,
+				addNamespaceXIDColumns,
+				addTransactionDefault,
+				addRelationTupleDefault,
+				addNamepsaceDefault,
+				backfillTransactionXids,
+				backfillTupleXIDColumns,
+				backfillNamespaceXIDColumns,
+				addTransactionSnapshotNotNull,
+				addTupleCreatedNotNull,
+				addNamespaceCreatedtNotNull,
+			} {
+				if _, err := tx.Exec(ctx, stmt); err != nil {
+					return err
+				}
+			}
+
+			return nil
+		}); err != nil {
+		panic("failed to register migration: " + err.Error())
+	}
+}
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -106,9 +106,25 @@ func NewPostgresDatastore(
 
 	configurePool(config, pgxConfig)
 
-	dbpool, err := pgxpool.ConnectConfig(context.Background(), pgxConfig)
+	initializationContext, cancelInit := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancelInit()
+
+	dbpool, err := pgxpool.ConnectConfig(initializationContext, pgxConfig)
 	if err != nil {
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	}
+
+	// Verify that the server supports commit timestamps
+	var trackTSOn string
+	if err := dbpool.
+		QueryRow(initializationContext, "SHOW track_commit_timestamp;").
+		Scan(&trackTSOn); err != nil {
+		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	}
+
+	watchEnabled := trackTSOn == "on"
+	if !watchEnabled {
+		log.Warn().Msg("watch API disabled, postgres must be run with track_commit_timestamp=on")
 	}
 
 	if config.enablePrometheusStats {
@@ -160,6 +176,7 @@ func NewPostgresDatastore(
 		gcTimeout:               config.gcMaxOperationTime,
 		analyzeBeforeStatistics: config.analyzeBeforeStatistics,
 		usersetBatchSize:        config.splitAtUsersetCount,
+		watchEnabled:            watchEnabled,
 		gcCtx:                   gcCtx,
 		cancelGc:                cancelGc,
 		readTxOptions:           pgx.TxOptions{IsoLevel: pgx.RepeatableRead, AccessMode: pgx.ReadOnly},
@@ -222,6 +239,7 @@ type pgDatastore struct {
 	analyzeBeforeStatistics bool
 	readTxOptions           pgx.TxOptions
 	maxRetries              uint8
+	watchEnabled            bool
 
 	gcGroup  *errgroup.Group
 	gcCtx    context.Context
@@ -355,7 +373,7 @@ func (pgd *pgDatastore) IsReady(ctx context.Context) (bool, error) {
 }
 
 func (pgd *pgDatastore) Features(ctx context.Context) (*datastore.Features, error) {
-	return &datastore.Features{Watch: datastore.Feature{Enabled: true}}, nil
+	return &datastore.Features{Watch: datastore.Feature{Enabled: pgd.watchEnabled}}, nil
 }
 
 func buildLivingObjectFilterForRevision(revision datastore.Revision) queryFilterer {
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -291,7 +291,7 @@ func (pgd *pgDatastore) ReadWriteTx(
 ) (datastore.Revision, error) {
 	var err error
 	for i := uint8(0); i <= pgd.maxRetries; i++ {
-		var newXID XID8
+		var newXID xid8
 		err = pgd.dbpool.BeginTxFunc(ctx, pgx.TxOptions{IsoLevel: pgx.Serializable}, func(tx pgx.Tx) error {
 			var err error
 			newXID, err = createNewTransaction(ctx, tx)
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -40,8 +40,6 @@ const (
 	colTimestamp        = "timestamp"
 	colNamespace        = "namespace"
 	colConfig           = "serialized_config"
-	colCreatedTxn       = "created_transaction"
-	colDeletedTxn       = "deleted_transaction"
 	colCreatedXid       = "created_xid"
 	colDeletedXid       = "deleted_xid"
 	colSnapshot         = "snapshot"
@@ -53,7 +51,7 @@ const (
 
 	errUnableToInstantiate = "unable to instantiate datastore: %w"
 
-	createTxn = "INSERT INTO relation_tuple_transaction DEFAULT VALUES RETURNING id, xid"
+	createTxn = "INSERT INTO relation_tuple_transaction DEFAULT VALUES RETURNING xid"
 
 	// The parameters to this format string are:
 	// 1: the created_xid or deleted_xid column name
@@ -73,8 +71,7 @@ const (
 	pgSerializationFailure      = "40001"
 	pgUniqueConstraintViolation = "23505"
 
-	livingTupleConstraintOld = "uq_relation_tuple_living"
-	livingTupleConstraint    = "uq_relation_tuple_living_xid"
+	livingTupleConstraint = "uq_relation_tuple_living_xid"
 )
 
 func init() {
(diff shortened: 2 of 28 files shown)


internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
```diff
@@ -0,0 +1,46 @@
...
+
+const addTransactionXIDColumns = `
+	ALTER TABLE relation_tuple_transaction
+		ADD COLUMN xid xid8 NOT NULL DEFAULT (pg_current_xact_id()),
+		ADD COLUMN snapshot pg_snapshot NOT NULL DEFAULT (pg_current_snapshot())`
+
+const addTupleXIDColumns = `
+	ALTER TABLE relation_tuple
+		ADD COLUMN created_xid xid8 NOT NULL DEFAULT (pg_current_xact_id()),
+		ADD COLUMN deleted_xid xid8 NOT NULL DEFAULT ('9223372036854775807');`
```
vroldanbet, line 17: I'm curious, why creating those when they are available as metadata at the row level that can be queried? And why using `deleted_xid` as the max value when PSQL stores it as `0` when it's not yet deleted?

internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
```diff
@@ -0,0 +1,46 @@
...
+
+const addNamespaceXIDColumns = `
+	ALTER TABLE namespace_config
+		ADD COLUMN created_xid xid8 NOT NULL DEFAULT (pg_current_xact_id()),
+		ADD COLUMN deleted_xid xid8 NOT NULL DEFAULT ('9223372036854775807');`
+
+const backfillTransactionXids = `
```
vroldanbet, line 24: do we understand the impact of this operation in a very large dataset?
How can we guarantee this does not cause an availability issue in the database while executed?
Why only backfilling the transactions table and not theother tables?
When `SELECT`ing, is the default not used on rows that already existed before the migration?

internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
jakedt: `xmin` is of type `xid` which is only 32 bits, the `pg_currext_xact_id()` is and `xid8` which is 64-bits and isn't subject to the transaction rollover problem. It's also the data-type that's needed for us with the `pg_snapshot` functions. You can cast an `xid8` to an `xid`, for example with the `pg_xact_commit_timestamp` function but not vice-versa.

Originally we used the max int64 because the conditionals were simpler, e.g. `<=`, but since they can be out of order we have to compare with the actual value so the utility isn't as obvious. It's still an `exceptional` value either way. The only reason I can see to keep it "very high" is because I believe that `pg_visible_in_snapshot` may not special case 0 and it will return true when the `xmin` gets higher.

internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
jakedt: > do we understand the impact of this operation in a very large dataset?
The transaction table is being constantly garbage collected, so this table shouldn't be super huge. I haven't measured the impact on a very large database though.

> How can we guarantee this does not cause an availability issue in the database while executed?
We don't currently.

> Why only backfilling the transactions table and not theother tables?
Transaction table needs to have unique values before we can make it the primary key. The idea here is that after you enable the "writes to both" phase, you wait at least one garbage collection period before moving on so that everything that might be selected on becomes fresh enough due to the garbage collector. These backfilled rows can't be used as revisions because they don't have an accurate snapshot value.

> When SELECTing, is the default not used on rows that already existed before the migration?
We will be using values that are new enough after waiting the garbage collection period on the actual relationship rows. E.g. all of the alive ones will be considered alive as of the time of migration.

josephschorr: Maybe we should do it in chunks?

internal/datastore/postgres/xid8.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+type pguint64 struct {
+	Uint   uint64
+	Status pgtype.Status
+}
+
+var errUndefined = errors.New("cannot encode status undefined")
+
+type XID8 pguint64
```
josephschorr, line 23: Mind adding a doc comment explaining this here (or make it private to the package)?

internal/testserver/datastore/postgres.go
```diff
@@ -32,7 +32,7 @@ func RunPostgresForTesting(t testing.TB, bridgeNetworkName string) RunningEngine
 	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
 		Name:         name,
 		Repository:   "postgres",
-		Tag:          "10.20",
+		Tag:          "13.8",
```
josephschorr, line 35: I think it might be worth moving this version into a const in the postgres datastore itself

internal/datastore/postgres/postgres.go
```diff
@@ -106,11 +106,22 @@ func NewPostgresDatastore(
...
 	if err != nil {
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
 
+	// Verify that the server supports commit timestamps
+	var trackTSOn string
+	if err := dbpool.
+		QueryRow(initializationContext, "SHOW track_commit_timestamp;").
+		Scan(&trackTSOn); err != nil {
+		return nil, fmt.Errorf(errUnableToInstantiate, err)
```
josephschorr, line 123: I think we need a more descriptive error here if this feature is not enabled, ideally one indicating the exact documentation on how to enable it

josephschorr: Do we want to skip this check if watch is disabled?

internal/datastore/postgres/postgres.go
```diff
@@ -54,6 +55,8 @@ const (
 
 	createTxn = "INSERT INTO relation_tuple_transaction DEFAULT VALUES RETURNING id, xid"
 
+	snapshotAlive = "pg_visible_in_snapshot(%s, (SELECT %s FROM %s WHERE %s = %s)) = %s"
```
josephschorr, line 56: Might be worth documenting here the parameters expected in this expansion

internal/datastore/postgres/postgres.go
```diff
@@ -74,7 +78,7 @@ func init() {
 var (
 	psql = sq.StatementBuilder.PlaceholderFormat(sq.Dollar)
 
-	getRevision = psql.Select("MAX(id)").From(tableTransaction)
+	getRevision = psql.Select("MAX(xid::text::bigint)").From(tableTransaction)
```
josephschorr, line 78: Should this use a const name for the col name? I know we didn't use it before, but maybe we should now?

internal/datastore/postgres/postgres.go
```diff
@@ -373,16 +377,43 @@ func (pgd *pgDatastore) Features(ctx context.Context) (*datastore.Features, erro
...
-			})
+		txID := transactionFromRevision(revision)
+
+		createdBeforeTXN := sq.Expr(fmt.Sprintf(
+			snapshotAlive,
+			colCreatedXid,
+			colSnapshot,
+			tableTransaction,
+			colXID,
+			sq.Placeholders(1),
```
josephschorr, line 383: What does the `1` mean in `Placeholders`?

internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
jakedt: I considered it, but we don't want to overwrite any XIDs that are actually generated by the system. The other option would be to select out the XID for the transaction that is going to be applied as the default, and then update only the rows that have that as their XID, but that requires a "go migration" and since this table is likely small for all users I didn't think it was worth the effort.

internal/datastore/postgres/postgres.go
jakedt: 1 placeholder. Basically `?`, if you said 3 it would be `?, ?, ?`

internal/datastore/postgres/postgres.go
jakedt: The datastore has no idea if the watch server is enabled.

internal/testserver/datastore/postgres.go
jakedt: I think the min required version (13.0) and the test version (latest 13.x) aren't necessarily the same? I can see postgres advertising its min required version.

internal/datastore/postgres/postgres.go
josephschorr: Maybe we should thread that in then?

internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
josephschorr: Let's verify that GC is working and then this should be fine

internal/datastore/postgres/xid8.go
```diff
@@ -0,0 +1,206 @@
...
+type pguint64 struct {
+	Uint   uint64
+	Status pgtype.Status
+}
+
+var errUndefined = errors.New("cannot encode status undefined")
+
+// xid8 represents the xid8 postgres type, which is a 64-bit transaction ID that is *NOT*
+// subject to transaction wraparound limitations.
+type xid8 pguint64
```
josephschorr, line 25: Can we link this to the docs?

internal/datastore/postgres/postgres.go
```diff
@@ -378,16 +388,41 @@ func (pgd *pgDatastore) Features(ctx context.Context) (*datastore.Features, erro
...
 func buildLivingObjectFilterForRevision(revision datastore.Revision) queryFilterer {
 	return func(original sq.SelectBuilder) sq.SelectBuilder {
-		return original.Where(sq.LtOrEq{colCreatedTxn: transactionFromRevision(revision)}).
-			Where(sq.Or{
-				sq.Eq{colDeletedTxn: liveDeletedTxnID},
-				sq.Gt{colDeletedTxn: revision},
-			})
+		txID := transactionFromRevision(revision)
+
+		createdBeforeTXN := sq.Expr(fmt.Sprintf(
```
vroldanbet, line 388: do we need to `Sprintf` the arguments of the expression that do not change during the lifetime of the process each time? `colSnapshot`, `tableTransaction` and `colXID`?

vroldanbet (approved):
It took me a bit to process all the changes commit by commit, but I think this looks pretty solid to me. I think the fundamental question to clarify is what kind of impact these migrations can have in a production PSQL primary, and describing what the the various manual steps OSS users have to take to migrate their PSQL datastores to the this version are.

Great job @jakedt this is good stuff 👏🏻 👏🏻 👏🏻

internal/datastore/postgres/migrations/zz_migration.0009_add_xid8_columns.go
jakedt: GC is working.

internal/datastore/postgres/watch.go
```diff
@@ -3,32 +3,45 @@ package postgres
...
-	colUsersetRelation,
-	colCreatedTxn,
-	colDeletedTxn,
-).From(tableTuple)
-
-func (pgd *pgDatastore) Watch(ctx context.Context, afterRevision datastore.Revision) (<-chan *datastore.RevisionChanges, <-chan error) {
+var (
+	newRevisionsQuery = fmt.Sprintf(`
+	SELECT %[1]s from %[2]s
+	WHERE pg_xact_commit_timestamp(%[1]s::xid) > (
```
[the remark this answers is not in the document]
jakedt: I added a comment. They cannot provide an incorrect timestamp just by overflowing. Postgres considers the most recent `max_int` (~2bln) xids to be in the past and the next `max_int` xids to be in the future, meaning that as we move into wraparound space the existing ones will still be valid.

internal/datastore/postgres/watch.go
```diff
@@ -3,32 +3,45 @@ package postgres
...
+		SELECT pg_xact_commit_timestamp(%[1]s::xid) FROM relation_tuple_transaction where xid = $1
+	) AND xid < pg_snapshot_xmin(pg_current_snapshot())
```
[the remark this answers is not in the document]
jakedt: This prevents pop-in of transactions that have yet to get committed but are lower than some others that are already committed.

internal/datastore/postgres/revisions.go
```diff
@@ -80,56 +80,60 @@ func (pgd *pgDatastore) CheckRevision(ctx context.Context, revision datastore.Re
...
 
-func revisionFromTransaction(txID uint64) datastore.Revision {
-	return decimal.NewFromInt(int64(txID))
+func revisionFromTransaction(txID xid8) datastore.Revision {
+	return decimal.NewFromInt(int64(txID.Uint))
 }
 
-func transactionFromRevision(revision datastore.Revision) uint64 {
-	return uint64(revision.IntPart())
+func transactionFromRevision(revision datastore.Revision) xid8 {
```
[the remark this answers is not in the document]
jakedt: Yeah, they could pass a negative number and it would cause spicedb to panic. I need to think about this one for a minute.

internal/datastore/postgres/migrations/zz_migration.0010_backfill_xids.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
+)
+
+const backfillTupleXIDColumns = `
+	UPDATE relation_tuple
+		SET deleted_xid = deleted_transaction::text::xid8
```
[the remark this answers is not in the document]
jakedt: `deleted_transaction` is one of the manually assigned IDs from a bigserial column. None will be higher than the highest 64 bit integer.

Yes, we may have to revisit the backfill, right now it's all done in SQL.

internal/datastore/postgres/migrations/zz_migration.0011_add_xid_indices.go
```diff
@@ -0,0 +1,43 @@
+package migrations
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
+)
+
+var addXIDIndices = []string{
```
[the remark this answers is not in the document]
jakedt: You can't create indices concurrently in a transaction.

internal/datastore/postgres/migrations/zz_migration.0011_add_xid_indices.go
```diff
@@ -0,0 +1,43 @@
...
+	`CREATE UNIQUE INDEX CONCURRENTLY ix_namespace_config_pk
+		ON namespace_config (namespace, created_xid, deleted_xid);`,
+	`CREATE UNIQUE INDEX CONCURRENTLY ix_namespace_config_living
+		ON namespace_config (namespace, deleted_xid);`,
+	`CREATE UNIQUE INDEX CONCURRENTLY ix_relation_tuple_pk
+		ON relation_tuple (namespace, object_id, relation, userset_namespace, userset_object_id,
+						   userset_relation, created_xid, deleted_xid);`,
+	`CREATE UNIQUE INDEX CONCURRENTLY ix_relation_tuple_living
+		ON relation_tuple (namespace, object_id, relation, userset_namespace, userset_object_id,
+						   userset_relation, deleted_xid);`,
```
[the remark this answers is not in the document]
jakedt: Better than pausing everything...

internal/datastore/postgres/migrations/zz_migration.0011_add_xid_indices.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
+)
+
+var addXIDIndices = []string{
+	// Replace the indices that are inherent from having a primary key constraint
+	"CREATE UNIQUE INDEX CONCURRENTLY ix_rttx_oldpk ON relation_tuple_transaction (id)",
```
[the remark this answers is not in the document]
jakedt: We only have it through the primary key constraint. When that is dropped, the index gets dropped too. We don't have a named index on the primary key.

internal/datastore/postgres/migrations/zz_migration.0012_add_xid_constraints.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v4"
+)
+
+const (
+	getNSConfigPkeyName = `
```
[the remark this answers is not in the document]
jakedt: It's because one of the migrations added a constraint without a name:
https://github.com/authzed/spicedb/blob/main/internal/datastore/postgres/migrations/zz_migration.0008_add_ns_config_id.go#L11

I didn't want to assume that it would always follow a naming convention.

internal/datastore/postgres/postgres.go
jakedt: I thought about that, but it would have required double encoding in a format string, which.... yeesh.

internal/datastore/postgres/migrations/zz_migration.0011_add_xid_indices.go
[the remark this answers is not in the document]
vroldanbet: just what I thought, thanks for confirming 👍🏻

internal/datastore/postgres/postgres.go
vroldanbet: I guess it's not a big deal, just those tiny things add up. As I did profiling, putting I/O wait aside, generating SQL with `squirrel` was a visible chunk of the CPU time.

internal/datastore/postgres/migrations/zz_migration.0012_add_xid_constraints.go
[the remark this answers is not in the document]
vroldanbet: ah, missed that one 👍🏻

internal/datastore/postgres/revisions.go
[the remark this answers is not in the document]
jakedt: Ok, fixed.

vroldanbet (approved):

josephschorr (approved):
LGTM

vroldanbet (approved):

josephschorr (approved):

PR #853, opened 2022-09-29
Add command to export migrations

ecordell:
This PR adds a new command, `spicedb migrations`, that will export a representation of the migrations for a given datastore.

Initially, only Postgres is supported, and for now only SQL files are output.

```sh
$ spicedb migrations --datastore-engine=postgres
$ tree migrations
migrations
├── 1664918821_DDL_expand_1eaeba4b8a73.sql
├── 1664918822_DDL_expand_add_reverse_index.sql
├── 1664918823_DML_contract_delete_all_but_newest_ns.sql
├── 1664918824_DDL_expand_add_unique_living_ns.sql
├── 1664918825_DDL_expand_add_transaction_timestamp_index.sql
├── 1664918826_DDL_contract_change_transaction_timestamp_default.sql
├── 1664918827_DDL_expand_add_gc_index_internal.sql
├── 1664918828_DML_expand_add_gc_index.sql
├── 1664918829_DDL_expand_add_unique_datastore_id.sql
├── 1664918830_DDL_expand_add_ns_config_id.sql
├── 1664918831_DDL_expand_add_xid_columns.sql
├── 1664918832_DDL_contract_add_xid_indices_0.sql
├── 1664918833_DDL_contract_add_xid_indices_1.sql
├── 1664918834_DDL_contract_add_xid_indices_2.sql
├── 1664918835_DDL_contract_add_xid_indices_3.sql
├── 1664918836_DDL_contract_add_xid_indices_4.sql
├── 1664918837_DDL_contract_add_xid_indices_5.sql
├── 1664918838_DDL_contract_add_xid_indices_6.sql
├── 1664918839_DML_expand_add_xid_indices.sql
├── 1664918840_DDL_contract_drop_ns_config_pk_2.sql
├── 1664918841_DDL_contract_add_xid_constraints_0.sql
├── 1664918842_DDL_contract_add_xid_constraints_1.sql
├── 1664918843_DDL_contract_add_xid_constraints_2.sql
├── 1664918844_DML_expand_add_xid_constraints.sql
└── 1664918845_DDL_contract_drop_bigserial_ids.sql
```

The filenames ensure ordering via epoch prefixes, and indicate at a glance what affect the migration will have (DDL/DML, expand/contract). Note that the epoch prefixes are not stable between runs of `spicedb migrations` - looking for feedback for whether that should be changed.

The existing `spicedb migrate` commands work exactly as they did before for users that aren't using external migration tools.

### DIFF
--- a/internal/datastore/postgres/migrations/manager.go
+++ b/internal/datastore/postgres/migrations/manager.go
@@ -1,15 +1,111 @@
 package migrations
 
 import (
+	"context"
+	"fmt"
+	"strings"
+
+	"golang.org/x/exp/slices"
+
 	"github.com/authzed/spicedb/pkg/migrate"
 
 	"github.com/jackc/pgx/v4"
 )
 
-var (
-	noNonatomicMigration migrate.MigrationFunc[*pgx.Conn]
-	noTxMigration        migrate.TxMigrationFunc[pgx.Tx]
-)
-
 // DatabaseMigrations implements a migration manager for the Postgres Driver.
 var DatabaseMigrations = migrate.NewManager[*AlembicPostgresDriver, *pgx.Conn, pgx.Tx]()
+
+type PGMigrationSet []*PostgresMigration
+
+var PostgresMigrations PGMigrationSet = make([]*PostgresMigration, 0)
+
+func (p PGMigrationSet) Sort() {
+	slices.SortFunc(p, func(a *PostgresMigration, b *PostgresMigration) bool {
+		return b.replaces == a.version
+	})
+}
+
+func RegisterPGMigration(p *PostgresMigration) {
+	if p.migrationType == "" {
+		panic("migration missing type")
+	}
+	if p.migrationSafety == "" {
+		panic("migration missing safety")
+	}
+	PostgresMigrations = append(PostgresMigrations, p)
+	if err := DatabaseMigrations.Register(p.version, p.replaces, p.expected, p.Run, nil); err != nil {
+		panic("failed to register migration: " + err.Error())
+	}
+}
+
+type migrationType string
+
+var (
+	DDL migrationType = "DDL"
+	DML migrationType = "DML"
+)
+
+type migrationSafety string
+
+var (
+	expand   migrationSafety = "expand"
+	contract migrationSafety = "contract"
+)
+
+type PostgresMigration struct {
+	version         string
+	replaces        string
+	expected        string
+	command         strings.Builder
+	suffix          string
+	migrationType   migrationType
+	migrationSafety migrationSafety
+}
+
+func (p *PostgresMigration) Run(ctx context.Context, conn *pgx.Conn) error {
+	_, err := conn.Exec(ctx, p.command.String())
+	return err
+}
+
+func (p *PostgresMigration) Begin() {
+	p.command.WriteString("BEGIN;\n")
+}
+
+func (p *PostgresMigration) Commit() {
+	p.command.WriteString("COMMIT;")
+}
+
+func (p *PostgresMigration) WriteVersion() {
+	if p.version == "" {
+		panic("cannot write empty version")
+	}
+	p.command.WriteString(fmt.Sprintf("UPDATE alembic_version SET version_num='%[1]s' WHERE version_num='%[2]s';\n", p.version, p.replaces))
+}
+
+func (p *PostgresMigration) WriteVersionOverride(previous string) {
+	if p.version == "" {
+		panic("cannot write empty version")
+	}
+	p.command.WriteString(fmt.Sprintf("UPDATE alembic_version SET version_num='%[1]s' WHERE version_num='%[2]s';\n", p.version, previous))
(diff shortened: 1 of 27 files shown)


josephschorr:
I think we need to have the migration SQL update the revision used for tracking; otherwise SpiceDB won't know where it is in a multi-step migration

internal/datastore/postgres/migrations/manager.go
```diff
@@ -1,15 +1,99 @@
...
 )
 
+const (
+	dir = "migrations"
+
+	createVersionTable = `CREATE TABLE alembic_version (
+    		version_num VARCHAR NOT NULL
+		);
+		INSERT INTO alembic_version (version_num) VALUES ('');`
+)
```
vroldanbet, line 22: why is this bit not handled also into its own SQL file?

internal/datastore/postgres/migrations/manager.go
```diff
@@ -1,15 +1,99 @@
...
+		}
+		previous := version
+		version = strings.TrimSuffix(migration.Name(), ".sql")
+		version = strings.TrimSuffix(migration.Name(), ".sh")
+		version = version[11:] // trim epoch
+		version = strings.TrimPrefix(version, "DDL_")
+		version = strings.TrimPrefix(version, "DML_")
+		version = strings.TrimPrefix(version, "go_")
+		version = strings.TrimPrefix(version, "expand_")
+		version = strings.TrimPrefix(version, "contract_")
```
vroldanbet, line 60: can you elaborate a bit on the semantics behind these file name conventions?

vroldanbet (commented):
@ecordell thanks for prototyping this out there. I think it's important to describe in the PR what kind of use-case we are trying to solve (e.g. is this something geared towards users that use a migration framework like liquibase, and if so, how can folks operating spicedb consume these?).

As far as I can tell, nothing in this new approach prevents us to mix DDL and DML in the same `sql` file. It's the sole responsibility of spicedb mantainers to make sure those are not mixed. Mixing DDL and DML is problematic for some other schema management tools / frameworks like https://github.com/skeema/skeema.

I think a good exercise would be to find "data management pipelines" out there and see if this design helps integrating with them. For example, it occurs to me that Planetscale could be a good target to exercise, as it has its own schema migration flavor. I speak without knowledge of what the Planetscale end-user schema migration pipeline looks like, but perhaps for it would help to have a SQL dump as well (compared to the "SQL transition" here proposed) to make it easier to have SpiceDB operator integrate the migration with Planetscale. I know this would also help with Skeema-based pipelines.

internal/datastore/postgres/migrations/manager.go
ecordell: If you're running migrations with an external tool, that external tool almost certainly has its own tracking table (and we can't just put all `alembic_version` modifications in the migration files, since they will be DML and a goal is to have some DDL-only files).

I was thinking of `spicedb migrate` as "just another tool" that can consume the migration files to run them; the `alembic_version` table is what it uses for bookkeeping.

ecordell:
> SpiceDB won't know where it is in a multi-step migration

SpiceDB currently doesn't know where it is in a multi-step migration either; I was hoping to decouple that requirement from this one if possible.

> As far as I can tell, nothing in this new approach prevents us to mix DDL and DML in the same sql file. It's the sole responsibility of spicedb mantainers to make sure those are not mixed.

Correct - we would need to be careful about that, or find/write some linters for it.

> I think a good exercise would be to find "data management pipelines" out there and see if this design helps integrating with them.

This is a good idea! I'll look into some options.

I want to get at least 1 external tool in a test to exercise this workflow in CI.

internal/datastore/postgres/migrations/manager.go
ecordell: resolving since this is not relevant to the current implementation

internal/datastore/postgres/migrations/zz_migration.0011_add_xid_constraints.go
```diff
@@ -1,21 +1,9 @@
...
-const (
-	getNSConfigPkeyName = `
-	SELECT constraint_name FROM information_schema.table_constraints
-		WHERE table_schema = 'public'
-      	AND table_name = 'namespace_config'
-      	AND constraint_type = 'PRIMARY KEY';`
-
-	dropNSConfigIDPkey = "ALTER TABLE namespace_config DROP CONSTRAINT %s;"
-)
+const dropNSConfigIDPkey = `ALTER TABLE namespace_config DROP CONSTRAINT namespace_config_pkey;
```
ecordell, line 5: @jakedt I removed the query for the primary key constraint in favor of hardcoding it. What are the scenarios we this constraint name is different? In my tests it was always this value.

ecordell:
We seem to be moving in the direction of making `spicedb migrate` smarter (vs. trying to support external tools that handle migrations). Closing this since we've paused on the idea of exposing the migrations directly like this.

PR #855, opened 2022-09-29
Switch check dispatch to use the new MembershipSet

josephschorr:
Part 2 of #846 

Following this change, caveats should be implicitly tracked by the dispatcher, even though they are not yet used anywhere


internal/services/v1/permissions.go
```diff
@@ -105,9 +105,14 @@ func (ps *permissionServer) CheckPermission(ctx context.Context, req *v1.CheckPe
 		return nil, rewritePermissionsError(ctx, err)
 	}
 
+	// TODO(jschorr): Support caveats here
 	permissionship := v1.CheckPermissionResponse_PERMISSIONSHIP_NO_PERMISSION
-	if found, ok := cr.ResultsByResourceId[req.Resource.ObjectId]; ok && found.Membership == dispatch.DispatchCheckResponse_MEMBER {
-		permissionship = v1.CheckPermissionResponse_PERMISSIONSHIP_HAS_PERMISSION
+	if found, ok := cr.ResultsByResourceId[req.Resource.ObjectId]; ok {
+		if found.Membership == dispatch.ResourceCheckResult_MEMBER {
```
vroldanbet, line 111: aside from `CAVEATED_MEMBER` it also includes `NOT_MEMBER` and `UNKNOWN`, yet the code assumes that if it's not member, then it's a caveated member. Wouldn't this lead to panicking on `NOT_MEMBER`, or mislead us if `UNKNOWN` happened due to a bug?

josephschorr: We don't include the entry in the `ResultsByResourceId` if it is `NOT_MEMBER`, and `UNKNOWN` should never be hit (if it is, it is a bug)

vroldanbet: huh, I see, so when is `NOT_MEMBER` used?

josephschorr: Its used in the dispatcher internally, but not placed into the results set for efficiency reasons

internal/graph/check.go
```diff
@@ -134,9 +129,9 @@ func (cc *ConcurrentChecker) checkInternal(ctx context.Context, req ValidatedChe
 	//
 	// If the filtering results in no further resource IDs to check, or a result is found and a single
 	// result is allowed, we terminate early.
-	foundResourceIds, filteredResourcesIds := filterForFoundMemberResource(req.ResourceRelation, req.ResourceIds, req.Subject)
-	if len(foundResourceIds) > 0 && req.DispatchCheckRequest.ResultsSetting == v1.DispatchCheckRequest_ALLOW_SINGLE_RESULT {
-		return checkResultsForResourceIds(foundResourceIds, emptyMetadata)
+	membershipSet, filteredResourcesIds := filterForFoundMemberResource(req.ResourceRelation, req.ResourceIds, req.Subject)
+	if membershipSet.HasDeterminedMember() && req.DispatchCheckRequest.ResultsSetting == v1.DispatchCheckRequest_ALLOW_SINGLE_RESULT {
```
vroldanbet, line 133: do we need to handle caveated members here?

josephschorr: They can't be: the incoming resource cannot have a caveat, and the subject cannot have a caveat: this is a *direct* comparison if we reach a point where the check is occurring for the same resource and subject, ala `check group:something#member group:something#member`

internal/graph/check.go
```diff
@@ -193,26 +188,29 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
...
 
+		// If the subject of the relationship matches the target subject, then we've found
+		// a result.
 		if onrEqualOrWildcard(tpl.Subject, crc.parentReq.Subject) {
-			foundResourceIds = append(foundResourceIds, tpl.ResourceAndRelation.ObjectId)
-			if crc.resultsSetting == v1.DispatchCheckRequest_ALLOW_SINGLE_RESULT {
-				return checkResultsForResourceIds(foundResourceIds, emptyMetadata)
+			foundResources.AddDirectMember(tpl.ResourceAndRelation.ObjectId, tpl.Caveat)
+			if crc.resultsSetting == v1.DispatchCheckRequest_ALLOW_SINGLE_RESULT && foundResources.HasDeterminedMember() {
+				return checkResultsForMembership(foundResources, emptyMetadata)
```
vroldanbet, line 205: It may be faster to resolve the caveat here at the leaf if `HasDeterminedMember == false` than dispatching.

One possible optimization would be to hedge the dispatch against the caveat evaluation:
- concurrently start evaluating all caveated members
- concurrently send dispatch request
- if any of the relationships with caveats resolve to `MEMBER` cancel the dispatched request.

I don't think we need to optimize this right now but we should keep a list of potential optimizations to start working on once we have the first end to end functional example working.

internal/graph/check.go
```diff
@@ -245,33 +243,36 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
...
 }
 
-func mapResourceIds(result CheckResult, resourceType *core.RelationReference, resourceIDsBySubjectID *util.MultiMap[string, string]) CheckResult {
+func mapFoundResources(result CheckResult, resourceType *core.RelationReference, relationshipsBySubjectONR *util.MultiMap[string, *core.RelationTuple]) CheckResult {
 	// Map any resources found to the parent resource IDs.
-	mappedResourceIds := []string{}
+	membershipSet := NewMembershipSet()
 	for foundResourceID, result := range result.Resp.ResultsByResourceId {
-		if result.Membership != v1.DispatchCheckResponse_MEMBER {
+		if result.Membership != v1.ResourceCheckResult_MEMBER {
```
vroldanbet, line 256: Wouldn't the dispatched response come back with potentially caveated results? Wouldn't this be discarding those?

vroldanbet: Ah, I see, the caveat is being added here, because the tuple was loaded in this dispatching node.

internal/graph/check.go
josephschorr: We can't resolve the caveat here without passing the full context into the DispatchCheckRequest, which means that the cache key would need to include the full context, thus making cache reuse much less likely here.

We also cannot return immediately because if we do not have a determined result, we cannot know for certain that this will short circuit

internal/graph/check.go
josephschorr: Yes, good catch.

vroldanbet: Would it make sense to add some tests that exercise dispatching caveated tuples? Perhaps mocking the response. It felt like this PR lacked testing around the new conditions. I run checks_test.go and the debugger didn't stop in any of the paths affected by the PR.

josephschorr: It explicitly lacks those tests: this PR is to ensure that the *existing* check behavior works with the new membership set. All other considerations are out of scope until the *next* PR, which will be exercising the caveats once we have a test harness built for it (which is what I will be doing next)

internal/graph/check.go
vroldanbet: That is certainly an important consideration. While early to say, "I have the sneaking suspicion" that operations like lookup* with a large corpus of caveated tuples would become CPU bound due to the execution of potential complex expressions that has to be carried out for the ingress SpiceDB node. That single node becomes the bottleneck for the streaming API. In contrast, evaluating in the owning node of the ring means we are can continue to scale by trowing more compute at the problem. 

So it may be a worthwhile investment to include the context in the dispatch request and it in the cache, at the expense of higher memory requirements and potential less hit rate if it avoids costly dispatches. And I guess we only need to do this if caveated tuples where found?

internal/graph/check.go
vroldanbet: Alright, fair, let's do that!

vroldanbet (approved):

PR #857, opened 2022-09-30
docker: switch to chainguard base images

jzelinskie:
Distroless.dev migrated to these domains

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN apk update && apk add --no-cache git
 COPY . .
 RUN CGO_ENABLED=0 go build -v ./cmd/spicedb/
 
-FROM distroless.dev/static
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN apk update && apk add --no-cache git
 COPY . .
 RUN CGO_ENABLED=0 go build -v ./cmd/spicedb/
 
-FROM distroless.dev/static
+FROM cgr.dev/chainguard/static:latest
```
vroldanbet, line 7: do they not offer a specific release version yet?

vroldanbet (approved):
LGTM, we probably want to also use this in `zed`

jzelinskie:
This is actually a follow-up after making the same PR for zed :P

PR #858, opened 2022-09-30
Implement structured errors for all user facing errors

### LINKED ISSUE #816: Machine readable precondition failure information
Currently when a precondition fails I see a GRPC response with [FAILED_PRECONDITION](https://grpc.github.io/grpc-java/javadoc/io/grpc/Status.html#FAILED_PRECONDITION), received into my application as a `io.grpc.StatusRuntimeException`:
```
Caused by: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: failed precondition: relation/permission `foo` not found under definition `site`
```
On my project we'd like to be able to use machine readable data to synthesize higher level (schema-aware) error messages on our output, without parsing the GRPC `Description` string returned by [StatusRuntimeException.getDescription()](https://grpc.github.io/grpc-java/javadoc/io/grpc/Status.html#getDescription--). We were wondering if the metadata key/value pairs could be used to parameterize the response with a precondition type, and the specific parameters to the precondition that failed. In the above we'd want to know that the precondition type was checking for the existence of a relation/permission, that the parameter for the check was "foo" and the definition it was expected in was "site". Maybe something like the following:
```
precondition_type: RELATION_EXISTENCE
relation_permission_id: foo
object_definition: site
```
This level of request and response would likely be easy to implement, but the current errors do not appear (and I might be mistaken on this) to be able to always unambiguously indicate which precondition failed in the case that multiple preconditions have been sent, or multiple sent and internal preconditions fail, possibly referring to the same definitions, objects,  subjects, etc. So part of this request might involve enhancing internal error propagation to enable the returned metadata to always indicate, in a machine-readable way, exactly which instance of a precondition failed.

josephschorr:
Fixes #816

### DIFF
--- a/internal/namespace/errors.go
+++ b/internal/namespace/errors.go
@@ -1,6 +1,7 @@
 package namespace
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -16,13 +17,20 @@ type ErrNamespaceNotFound struct {
 }
 
 // NotFoundNamespaceName is the name of the namespace not found.
-func (enf ErrNamespaceNotFound) NotFoundNamespaceName() string {
-	return enf.namespaceName
+func (err ErrNamespaceNotFound) NotFoundNamespaceName() string {
+	return err.namespaceName
 }
 
 // MarshalZerologObject implements zerolog object marshalling.
-func (enf ErrNamespaceNotFound) MarshalZerologObject(e *zerolog.Event) {
-	e.Str("error", enf.Error()).Str("namespace", enf.namespaceName)
+func (err ErrNamespaceNotFound) MarshalZerologObject(e *zerolog.Event) {
+	e.Err(err).Str("namespace", err.namespaceName)
+}
+
+// DetailsMetadata returns the metadata for details for this error.
+func (err ErrNamespaceNotFound) DetailsMetadata() map[string]string {
+	return map[string]string{
+		"definition_name": err.namespaceName,
+	}
 }
 
 // ErrRelationNotFound occurs when a relation was not found under a namespace.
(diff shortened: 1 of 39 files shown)


pkg/errors/withstatus.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Domain is the domain used for all errors.
+const Domain = "authzed.com"
+
+// WithCodeAndDetails returns a gRPC status message containing the error's message, the given
+// status code and any supplied details.
+func WithCodeAndDetails(err error, code codes.Code, details ...protoiface.MessageV1) *status.Status {
+	status, err := status.New(code, err.Error()).WithDetails(details...)
+	if err != nil {
+		panic(err)
```
jzelinskie, line 22: I'd give an explanation as to why we're panicking here.

I looked into the code and saw it's if you pass an OK code.

pkg/errors/withstatus.go
```diff
@@ -0,0 +1,58 @@
...
+
+func WithCodeAndReason(err error, code codes.Code, reason v1.ErrorReason) error {
+	metadata := map[string]string{}
+
+	var hasMetadata HasMetadata
+	if ok := errors.As(err, &hasMetadata); ok {
+		metadata = hasMetadata.DetailsMetadata()
+	}
+
+	status, err := status.New(code, err.Error()).WithDetails(ForReason(reason, metadata))
```
jzelinskie, line 44: Why handle the error here instead of calling the function you just defined above?

pkg/errors/testutil.go
```diff
@@ -0,0 +1,28 @@
...
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/status"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+)
+
+// RequireReason asserts that an error is a gRPC error and returns the expected
+// reason in the ErrorInfo.
+// TODO(jschorr): Move into grpcutil.
+func RequireReason(t *testing.T, reason v1.ErrorReason, err error, expectedMetadataKeys ...string) {
```
jzelinskie, line 16: I know you're just moving this, but prefer `testing.TB` over `*testing.T`

internal/services/v1alpha1/errors.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/authzed/spicedb/internal/namespace"
+	"github.com/authzed/spicedb/internal/services/shared"
+	"github.com/authzed/spicedb/internal/sharederrors"
+	"github.com/authzed/spicedb/pkg/datastore"
+	cerror "github.com/authzed/spicedb/pkg/errors"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
+)
+
+func rewriteError(ctx context.Context, err error) error {
```
jzelinskie, line 20: If everything implements the proper interface for producing metadata, I wonder if it's better to be explicit at the call-sites as to how we're converting. That way we can distinguish when two different calls can return the same error, but have different desired responses.

jzelinskie (changes requested):
For the most part, no major issues. I had some organizational questions, too, though.

pkg/errors/common.go
```diff
@@ -1,4 +1,4 @@
-package commonerrors
+package errors
```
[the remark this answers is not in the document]
josephschorr: Well, I was trying to match the `errors.go` we have in every other package, but I'm open to suggestions on naming. Its not *solely* for grpc though

internal/services/v1alpha1/errors.go
josephschorr: `rewriteError` is per package, so it shouldn't need to be differentiated (yet anyway)

pkg/errors/common.go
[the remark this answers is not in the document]
josephschorr: Changed to `spiceerrors`

internal/namespace/errors.go
```diff
@@ -193,6 +261,29 @@ func NewPermissionsCycleErr(nsName string, permissionNames []string) error {
 	}
 }
 
+// wrapAsTypeError wraps another error in a type error.
+func wrapAsTypeError(wrapped error) error {
+	if wrapped == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
josephschorr: I tried changing it to a pointer and it broke the unwrapping, so I left it returning an error

josephschorr:
Updated

internal/namespace/errors.go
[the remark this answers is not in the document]
jzelinskie: interesting, ok

internal/services/v1alpha1/errors.go
jzelinskie: The only issue I see with this approach is it'll cause developers to just thoughtlessly do `rewriteError(ctx, err)` and not think about what actually needs to happen.

jzelinskie (approved):

PR #864, opened 2022-10-03
introduce a caveat evaluator dispatcher

vroldanbet:
WIP


vroldanbet:
Closing as dispatcher layer is not the right level of abstraction to evaluate caveats. Tests will be reused.

PR #888, opened 2022-10-07
disable async logging

vroldanbet:
async logging was recently introduced in https://github.com/authzed/spicedb/pull/844

unfortunately nothing is calling `Close()` on `diode.Writer`, which means buffer is not being flushed on process termination and causes log messages to get lost.

I propose disabling this until we come up with a solution that guarantees no log messages are lost on process termination.

Note: I also upgraded to go 1.19.2 to fix some vulns detected by `govulncheck`


ecordell (approved):
LGTM

Issue #884, opened 2022-10-04
support caveats in postgres datastore


vroldanbet:
Tracking: https://github.com/authzed/spicedb/issues/386
Part of https://github.com/authzed/spicedb/issues/835

Add support in `postgres` datastore to support caveats

PR #890, opened 2022-10-07
add caveat support to postgres datastore

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/884
Some improvements in the context of https://github.com/authzed/spicedb/issues/753

## What

introduces caveat support in postgres datastore

## How

- refactors `memdb` caveat test and turns it into a datastore test. Datastores not implementing the interface are skipped
- adds migration to add caveat table for PSQL, and contextualized caveat to `relational_tuple` table
- makes psql TX implements `CaveatStorer`, following same strategy for MVCC
- adds context to `ReadCaveatByName` to follow same signature as `ReadNamespace`
- adjusts `memdb` to avoid upserts on duplicate caveats, and align with PSQL
- adds missing test exercising `Watch API` with caveated relationships
- adds missing test exercising `DeleteCaveats` and implements method in `memdb`


internal/datastore/postgres/postgres_test.go
```diff
@@ -550,209 +543,6 @@ func QuantizedRevisionTest(t *testing.T, b testdatastore.RunningEngineForTest) {
 	}
 }
 
-func XIDMigrationAssumptionsTest(t *testing.T, b testdatastore.RunningEngineForTest) {
```
vroldanbet, line 553: @jakedt would love to hear your thoughts on this. I've removed this test. I'll just quote what I wrote in the commit log:

>unfortunately this test cannot be held around as the postgresql datastore evolves. It's the only test that runs code at a specific set of revisions, and therefore any changes to the code would have to maintain backward compatibility w.r.t to those revisions.

>In this particular example `QueryRelationships` failed to find the new caveat columns. It would require us to maintain both execution paths: relationship query with and without caveats.

pkg/cmd/migrate.go
```diff
@@ -50,18 +50,18 @@ func migrateRun(cmd *cobra.Command, args []string) error {
 		var err error
 		migrationDriver, err := crdbmigrations.NewCRDBDriver(dbURL)
 		if err != nil {
-			log.Fatal().Err(err).Msg("unable to create migration driver")
```
vroldanbet, line 53: move away from `log.Fatal` as it calls `os.Exit` and thus fails to unwind the call stack. `main.go` already does this if an `error` is returned by the cobra command.

pkg/migrate/migrate.go
```diff
@@ -115,6 +116,9 @@ func (m *Manager[D, C, T]) Run(ctx context.Context, driver D, throughRevision st
 	if err != nil {
 		return fmt.Errorf("unable to compute migration list: %w", err)
 	}
+	if len(toRun) == 0 {
+		log.Info().Str("targetRevision", requestedRevision).Msg("server already at requested revision")
```
vroldanbet, line 120: re-running migrate didn't provide any obvious log output indicating already running at requested revision

internal/datastore/postgres/postgres_test.go
jakedt: We can't remove this test until a revision of SpiceDB is cut that includes the 4-phase migrations and we have migrated all of our managed environments successfully. The specific set of revisions are actually created manually, so there is nothing dangerous about that particular part of the migration. We haven't cut a release of this yet, so there is still time to slot your migration in ahead of the 4-phase migrations if you want. That should eliminate the issues you saw.

jakedt (changes requested):

internal/datastore/postgres/postgres_test.go
vroldanbet: @jakedt I've inserted the caveat migration right before the `xid` migrations and the removed test is back and 🟢

internal/datastore/common/sql.go
```diff
@@ -18,6 +18,9 @@ import (
 )
 
 var (
+	// ObjCaveatNameKey is a tracing attribute representing a caveat name
+	ObjCaveatNameKey = attribute.Key("authzed.com/spicedb/sql/objCaveatName")
```
josephschorr, line 22: Should probably just be `caveatName` since it doesn't apply to the resource or the subject, but rather the relationship as a whole

internal/datastore/memdb/caveat.go
```diff
@@ -68,14 +69,19 @@ func (rwt *memdbReadWriteTx) WriteCaveats(caveats []*core.Caveat) error {
 }
 
 func (rwt *memdbReadWriteTx) writeCaveat(tx *memdb.Txn, caveats []*core.Caveat) error {
+	caveatNames := make(map[string]struct{}, len(caveats))
```
josephschorr, line 72: You could use a `util.NewSet[string]` here

internal/datastore/postgres/common/pgx.go
```diff
@@ -53,18 +54,46 @@ func queryTuples(ctx context.Context, sql string, args []any, span trace.Span, t
...
-			&nextTuple.ResourceAndRelation.Relation,
-			&nextTuple.Subject.Namespace,
-			&nextTuple.Subject.ObjectId,
-			&nextTuple.Subject.Relation,
-		)
+		var caveatName sql.NullString
+		var caveatCtx map[string]any
+		var err error
+		// TODO(vroldanbet) deprecate bool flag once CRDB also supports caveat
+		if caveatEnabled {
```
josephschorr, line 61: We should always be able to read the data, even if the caveats are disabled, as the column still exists, right?

internal/datastore/postgres/readwrite.go
```diff
@@ -94,22 +98,26 @@ func (rwt *pgReadWriteTXN) WriteRelationships(mutations []*core.RelationTupleUpd
...
+			}
 			valuesToWrite := []interface{}{
 				tpl.ResourceAndRelation.Namespace,
 				tpl.ResourceAndRelation.ObjectId,
 				tpl.ResourceAndRelation.Relation,
 				tpl.Subject.Namespace,
 				tpl.Subject.ObjectId,
 				tpl.Subject.Relation,
+				caveatName,
+				caveatContext,
```
josephschorr, line 120: You're writing a `map` here? Shouldn't it be the bytes of the encoded context StructPb?

Please add a test for writing some context and reading it back... I think this is broken

pkg/datastore/test/caveat.go
```diff
@@ -1,65 +1,79 @@
...
 	req.NoError(err)
 
-	// Dupes in same transaction are treated as upserts
-	coreCaveat := createCoreCaveat(t)
+	skipIfNotCaveatStorer(t, ds)
+
+	// Dupes in same transaction are fail to be written
 	ctx := context.Background()
+	coreCaveat := createCoreCaveat(t)
 	_, err = writeCaveats(ctx, ds, coreCaveat, coreCaveat)
```
josephschorr, line 33: Write some non-empty context here and verify on the read that it is the same map, with the same values

internal/datastore/postgres/common/pgx.go
vroldanbet: no because the column does not exist in CRDB yet

internal/datastore/postgres/readwrite.go
vroldanbet: this is intentional. The PGX driver knows how to map a `map` into a `JSONB`. Remember, we are using JSONB so we can query the context later on without parsing it.

pkg/datastore/test/caveat.go
vroldanbet: There is a test for this already: `WriteCaveatedRelationshipTest`

internal/datastore/postgres/common/pgx.go
josephschorr: d'oh! ok

internal/datastore/postgres/readwrite.go
josephschorr: Oh frick... can you add a comment here explaining that? it was totally non-obvious to me

pkg/datastore/test/caveat.go
josephschorr: Please add a table driven test with different context maps\

vroldanbet: thoughts on doing it in my follow up?

internal/datastore/postgres/readwrite.go
vroldanbet: Sure can!

josephschorr (approved):
LGTM from my side

josephschorr (approved):
LGTM

PR #891, opened 2022-10-07
Incremental migrations

jakedt:
Migration instructions:

1. `spicedb migrate add-xid-columns`
2. Rolling update spicedb with flag `--datastore-migration-phase write-both-read-old`
3. `spicedb migrate add-xid-constraints` - performs a backfill and may take a long time
4. Rolling update spicedb with flag `--datastore-migration-phase write-both-read-new`
5. `spicedb migrate drop-id-constraints`
6.  Rolling update spicedb without a `--datastore-migration-phase` flag
7. `spicedb migrate head`


williamdclt (commented):
If I understand the migration instructions correctly, we'd need to run the migrations manually rather than automatically? Would that mean disabling the k8s post-upgrade hook temporarily?

Suggestion: instead of starting SpiceDB with different flags, should that be different spicedb versions? ie something like:

- Upgrade to v1.14.0
	- migrations ran by post-upgrade hook (migration 9 and 10 but _without_ the backfilling in 10)
	- 1.14.0 always does "write both read old"
- Manually run `spicedb run-job backfill-xid`
	- `run-job` is something I made up. Effectively a migration to be ran manually.
- Upgrade to v1.15.0
	- Migration 11 ran by post-upgrade hook
- Upgrade to v1.16.0
	- 1.16.0 always does "write both read new"
	- Check that the v1.15.0 migrations have been ran, refuse to start otherwise
- Upgrade to v1.17.0
	- Run migrations 12/13
	- 1.16.0 always does "write new read new" ("old" doesn't exist anymore)
	
Maybe it can be consolidated to 3-phase rather than 4, I haven't overly thought about it

internal/datastore/postgres/migrations/zz_migration.0010_backfill_xid_add_indices.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/rs/zerolog/log"
+)
+
+const batchSize = 1000
```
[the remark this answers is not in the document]
jakedt: I'm just going to make it a command line parameter to the `migrate` command, defaulting to 1000.

jakedt:
@williamdclt Updated, PTAL

pkg/migrate/context.go
```diff
@@ -0,0 +1,11 @@
+package migrate
+
+// MigrationVariable contains constants that can be used as context keys that might
+// be relevant in a number of different migration scenarios.
+type MigrationVariable int
```
josephschorr, line 5: do we want to make these ints vs strings? I have concerns there might be accidental overlap with something we don't inject

jakedt: Not possible, the equality operator the context uses considers the type:
https://go.dev/play/p/16BfmcC0d9X

josephschorr (approved):
LGTM

Issue #639, opened 2022-06-02
Better Caching Cost & Density


benCoomes:
# Improve Cache Density and Cost Estimate

Hi Authzed folks - apologies in advance for this wall of text. 🙂

I noticed a few weeks ago that the cache cost functions are not accurate if the cost represents bytes (which I believe it does). For example, the cost of a `checkResultEntry` is set to just 8 bytes, the cost of that struct when empty. But that cost doesn't include the memory pointed to by `checkResultEntry.response`, which could be much more.

As I worked to improve the cache cost functions, I found a way to fit 2x more cache items into the same amount of memory: instead of caching the Go structs, cache the protobuf-marshaled bytes. 

The improved cache cost functions help keep the physical memory used by the cache much closer to the configured max cost. 

**I'd be happy to open some PRs for these changes, but wanted to post my findings here and see which of the changes you'd like (if any).**

## Cache Density
I experimented with storing the marshaled bytes of protobuf messages rather than the Go objects directly. 

There are two main advantages to this:
* Calculating the cost of a `[]byte` is quite simple. Most importantly, the cost function does not need to change as the protobuf message changes: protobuf takes care of those details.
* Second, the cache can store more items per MB of space used. In one test (below), the cache fit 212% more items per MB! However, later tests with more accurate cost functions improved cache density by a more modest 50-70%. All tests were on a single local instance of spicedb, so a load test at scale is warranted.

Below are the results for two tests run on a single spicedb instance serving check requests. Total profiled space is for the whole application, while cache profiled space includes just the stacks related to caching. In this test, the cost function was still poor, but it does show that using marshaled bytes significantly improves cache density.
| test | total profiled space | cache profiled space | cache calculated cost | key count | keys/ cache profiled MB |
| --- | --- | --- | --- | --- | --- |
| protobuf structs | 69.16 MB | 54.85 MB | 32 MB | 142,857 | 2,605 |
| marshaled []byte | 77.02 MB | 61.0 MB | 30.1 MB | 337,311 | 5,529 |

Of course, marshaling isn't free. However, existing code already calls `proto.Clone()` on every cache write, and as that is replaced with the call to `proto.Marshal()`, the relative cost may not be significant. Still, a test to check impact on CPU during a load test is warranted. 

## Cache Cost Function
Now, the long story. 

### Background
[...]

PR #900, opened 2022-10-14
internal: store serialized protos in caches

jzelinskie:
By trading off some CPU, the cache items can be stored as byte slices with very accurate size cost estimates.

This should somewhat close the loop on #639.

### DIFF
--- a/internal/datastore/proxy/caching.go
+++ b/internal/datastore/proxy/caching.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"sync"
 	"testing"
+	"unsafe"
 
 	"github.com/dustin/go-humanize"
 	"github.com/shopspring/decimal"
@@ -84,8 +85,14 @@ func (r *nsCachingReader) ReadNamespace(
 				return nil, err
 			}
 
-			entry := &cacheEntry{loaded, updatedRev, err}
-			r.p.c.Set(nsRevisionKey, entry, int64(loaded.SizeVT()))
+			marshalledNsDef, err := loaded.MarshalVT()
+			if err != nil {
+				// Propagate this error to the caller
+				return nil, err
+			}
+
+			entry := &cacheEntry{marshalledNsDef, updatedRev, err}
+			r.p.c.Set(nsRevisionKey, entry, entry.Size())
 
 			// We have to call wait here or else Ristretto may not have the key
 			// available to a subsequent caller.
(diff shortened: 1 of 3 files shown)


internal/datastore/proxy/caching.go
```diff
@@ -99,41 +106,58 @@ func (r *nsCachingReader) ReadNamespace(
 	}
 
 	loaded := loadedRaw.(*cacheEntry)
-	return loaded.def, loaded.updated, loaded.notFound
+
+	var def core.NamespaceDefinition
+	err := def.UnmarshalVT(loaded.marshalledNsDef)
```
josephschorr, line 111: In the case where the namespace was not found, what will have been marshaled in for us to unmarshal here?

internal/dispatch/caching/caching.go
```diff
@@ -246,25 +231,25 @@ func (cd *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckRe
...
-				return cachedResult.response, nil
-			}
+		var response v1.DispatchCheckResponse
+		if err := response.UnmarshalVT(cachedResultRaw.([]byte)); err != nil {
+			return &v1.DispatchCheckResponse{Metadata: &v1.ResponseMeta{}}, err
+		}
 
+		if req.Metadata.DepthRemaining >= response.Metadata.DepthRequired {
+			cd.checkFromCacheCounter.Inc()
 			// If debugging is requested, clone and add the req and the response to the trace.
```
josephschorr, line 241: Fix comment now that we're not cloning

internal/datastore/proxy/caching.go
jzelinskie: A NamespaceDefinition with all zero values for the fields. While it's not necessarily the most memory efficient (as I suppose using `nil` would be), it's the easiest to reason about, which I think is the more important thing for this codepath.

josephschorr: Okay, mind adding a comment?

internal/datastore/proxy/caching_test.go
```diff
@@ -25,6 +26,19 @@ const (
...
 )
 
+// TestNilUnmarshal asserts that if we get a nil NamespaceDefinition from a
+// datastore implementation, the process of inserting it into the cache and
+// back does not break anything.
+func TestNilUnmarshal(t *testing.T) {
+	nsDef := (*core.NamespaceDefinition)(nil)
+	marshalled, err := nsDef.MarshalVT()
+	require.Nil(t, err)
+	var newDef *core.NamespaceDefinition
```
josephschorr, line 36: newline before this line

Might also be worth adding another test for a "not found" entry, if we don't have one already

internal/datastore/proxy/caching.go
jzelinskie: I instead added a test because it actually does do `nil` and an empty slice. Not sure why I thought it did the zero value.

internal/datastore/proxy/caching_test.go
jzelinskie: Not sure how to go about that with the least setup. I'm willing to add it if you can point me in the right direction.

josephschorr: Let's do it as a follow up

josephschorr (approved):
LGTM

PR #904, opened 2022-10-17
fix calls to logger that are not being sent

vroldanbet:
Part of https://github.com/authzed/spicedb/issues/753

# What

Fixes logger calls that aren't actually being logged

# How

These calls to are not being actually sent (which is usually usually done with `Msg()` or `Send()`)


jakedt:
@josephschorr another opportunity for a custom linter, if one doesn't already exist.

jakedt (approved):

PR #906, opened 2022-10-17
more follow ups to caveats in datastore

vroldanbet:
- [x] adds exhaustive testing of CREATE/TOUCH semantics
- [x] add variadic caveat list to `ListCaveats`
- [x] make `WriteCaveat` serialize the whole `CaveatDefinition` proto instead of just the serialized bytes

### DIFF
--- a/internal/datastore/postgres/caveat.go
+++ b/internal/datastore/postgres/caveat.go
@@ -18,7 +18,7 @@ import (
 var (
 	writeCaveat            = psql.Insert(tableCaveat).Columns(colCaveatName, colCaveatDefinition)
 	writeCaveatDeprecated  = psql.Insert(tableCaveat).Columns(colCaveatName, colCaveatDefinition, colCreatedTxnDeprecated)
-	listCaveat             = psql.Select(colCaveatDefinition).From(tableCaveat)
+	listCaveat             = psql.Select(colCaveatDefinition).From(tableCaveat).OrderBy(colCaveatName)
 	readCaveat             = psql.Select(colCaveatDefinition, colCreatedXid).From(tableCaveat)
 	readCaveatDeprecated   = psql.Select(colCaveatDefinition, colCreatedTxnDeprecated).From(tableCaveat)
 	deleteCaveat           = psql.Update(tableCaveat).Where(sq.Eq{colDeletedXid: liveDeletedTxnID})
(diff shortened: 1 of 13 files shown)


pkg/datastore/caveat.go
```diff
@@ -11,8 +11,9 @@ type CaveatReader interface {
 	// ReadCaveatByName returns a caveat with the provided name
 	ReadCaveatByName(ctx context.Context, name string) (*core.CaveatDefinition, Revision, error)
 
-	// ListCaveats returns all caveats stored in the system
-	ListCaveats(ctx context.Context) ([]*core.CaveatDefinition, error)
+	// ListCaveats returns all caveats stored in the system. If caveatNames are provided
+	// the result will be filtered to the provided caveat names
+	ListCaveats(ctx context.Context, caveatNames ...string) ([]*core.CaveatDefinition, error)
```
josephschorr, line 16: name the parameter `optionalCaveatNames` or `caveatNamesForFiltering` or somesuch

internal/datastore/postgres/caveat.go
```diff
@@ -46,26 +46,32 @@ func (r *pgReader) ReadCaveatByName(ctx context.Context, name string) (*core.Cav
...
-	ctx, span := tracer.Start(ctx, "ListCaveats")
+func (r *pgReader) ListCaveats(ctx context.Context, caveatNames ...string) ([]*core.CaveatDefinition, error) {
+	ctx, span := tracer.Start(ctx, "ListCaveats", trace.WithAttributes(
+		attribute.StringSlice("names", caveatNames),
+	))
 	defer span.End()
 
-	filteredListCaveat := r.filterer(listCaveat)
+	caveatsWithNames := listCaveat
+	if len(caveatNames) > 0 {
```
josephschorr, line 70: I wonder if we should `nil` check here and if not nil and empty, just return nothing. @jzelinskie Thoughts?

josephschorr: Checked: it is empty by default, so this is fine

josephschorr: We'll just need to be extra careful with the call sites

josephschorr (approved):
LGTM

josephschorr (dismissed):
LGTM

pkg/datastore/test/caveat.go
```diff
@@ -81,29 +99,62 @@ func WriteCaveatedRelationshipTest(t *testing.T, tester DatastoreTester) {
...
+	req.NoError(err)
+	assertTupleCorrectlyStored(req, ds, rev, tpl)
+
+	// TOUCH can store caveat in relationship with no caveat
+	tpl.Caveat = caveatContext
+	rev, err = common.WriteTuples(ctx, sds, core.RelationTupleUpdate_TOUCH, tpl)
+	req.NoError(err)
+	assertTupleCorrectlyStored(req, ds, rev, tpl)
 
 	// Caveated tuple can reference non-existing caveat - controller layer is responsible for validation
```
josephschorr, line 140: Just thought of one more test for here: write a caveated relationship and ensure it can be deleted without the caveat?

josephschorr (approved):
LGTM

PR #908, opened 2022-10-18
Make logging package public

ecordell:
This also removes the manually-vendored grpc logging middleware, now that the fix is merged and in a tagged release.


vroldanbet (commented):
👋🏻 Hey Evan, what's the motivation behind making this package public? I believe it exposes zerolog API, which makes it then part of our API and difficult to switch logging implementation if the time ever comes.

PR #909, opened 2022-10-18
Fix stack overflow on error marshalling for zerolog

ecordell:
Prior to this, it was possible for an error that implemented `MarshalZeroLogObject` to infinitely recurse and cause a stack overflow.

I didn't track down when this was introduced, but my guess is a change in the serialization priority in a zerolog update. Since `MarshallZeroLogObject()` would call `.Err(err)`, zerolog would attempt to use the custom marshalling in the `Err()`(instead of as an `error`) and recurse indefinitely.

I also bumped grpc-go-middleware to a release that includes a fix we've been manually vendoring. When I did that, I noticed we were using a mix of v1 and v2, so I bumped everything to v2 (which is still RC, but we were using large chunks of it already).


vroldanbet (dismissed):
Changes look good to me. Would it make sense to open a follow up issue that adds a linter? It seems like it would be very easy to reintroduce those again.

pkg/datastore/errors.go
```diff
@@ -19,7 +19,7 @@ func (err ErrNamespaceNotFound) NotFoundNamespaceName() string {
 
 // MarshalZerologObject implements zerolog object marshalling.
 func (err ErrNamespaceNotFound) MarshalZerologObject(e *zerolog.Event) {
```
jzelinskie, line 21: I actually think if we used `e` as the receiver, this would've been unlikely to happen. Maybe we should move towards that, too.

vroldanbet (approved):

PR #912, opened 2022-10-18
Tech Debt cleanup: move LogOnError into common datastore package

### LINKED ISSUE #609: Move LogOnError from migrations package in MySQL into the main package
Its a nice utility and not only used by migrations anymore

josephschorr:
Fixes #609

### DIFF
--- a/internal/datastore/common/logging.go
+++ b/internal/datastore/common/logging.go
@@ -1,0 +1,15 @@
+package common
+
+import (
+	"context"
+
+	log "github.com/authzed/spicedb/internal/logging"
+)
+
+// LogOnError executes the function and logs the error.
+// Useful to avoid silently ignoring errors in defer statements
+func LogOnError(ctx context.Context, f func() error) {
+	if err := f(); err != nil {
+		log.Ctx(ctx).Err(err).Msg("datastore error")
+	}
+}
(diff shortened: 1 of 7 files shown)


internal/datastore/common/logging.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+
+	log "github.com/authzed/spicedb/internal/logging"
+)
+
+// LogOnError executes the function and logs the error.
+// Useful to avoid silently ignoring errors in defer statements
+func LogOnError(ctx context.Context, f func() error) {
+	if err := f(); err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("datastore error")
```
vroldanbet, line 13: `Error()` here is redundant. From zerolog API:

```go
// Err starts a new message with error level with err as a field if not nil or
// with info level if err is nil.
//
// You must call Msg on the returned event in order to send the event.
func (l *Logger) Err(err error) *Event {
	if err != nil {
		return l.Error().Err(err)
	}

	return l.Info()
}
```

vroldanbet (approved):
LGTM, just a small nit

internal/datastore/common/logging.go
josephschorr: That's what I get for copy-pasting :)

josephschorr: Fixed

vroldanbet (approved):

PR #913, opened 2022-10-18
Elide updates of namespaces which have not changed at all

### LINKED ISSUE #902: Investigate eliding updates of namespaces unchanged in schema write
If a schema is written with namespaces that have not changed at all, investigate whether they need to be written

This can occur if the *same* schema is written by tooling

josephschorr:
Should reduce some of the write work when updating larger schemas and produces less garbage

Fixes #902

### DIFF
--- a/internal/services/v1/schema.go
+++ b/internal/services/v1/schema.go
@@ -157,13 +157,25 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
 		}
 
 		// For each definition, perform a diff and ensure the changes will not result in any
-		// relationships left without associated schema.
+		// breaking changes.
+		objectDefsWithChanges := make([]*core.NamespaceDefinition, 0, len(compiled.ObjectDefinitions))
 		for _, nsdef := range compiled.ObjectDefinitions {
-			if err := shared.SanityCheckExistingRelationships(ctx, rwt, nsdef, existingObjectDefMap); err != nil {
+			diff, err := shared.SanityCheckNamespaceChanges(ctx, rwt, nsdef, existingObjectDefMap)
+			if err != nil {
 				return err
 			}
+
+			if len(diff.Deltas()) > 0 {
+				objectDefsWithChanges = append(objectDefsWithChanges, nsdef)
+			}
 		}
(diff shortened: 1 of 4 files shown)


vroldanbet:
@josephschorr please address merge conflicts to proceed with review 🙏🏻

internal/services/v1/schema.go
```diff
@@ -123,13 +123,19 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
...
+			if err != nil {
 				return err
 			}
+
+			if len(diff.Deltas()) > 0 {
+				nsDefsWithChanges = append(nsDefsWithChanges, nsdef)
+			}
 		}
-		log.Ctx(ctx).Trace().Interface("namespaceDefinitions", nsdefs).Msg("validated namespace definitions")
+		log.Ctx(ctx).Trace().Interface("namespaceDefinitions", nsdefs).Interface("namespaceDefinitionsWithChanges", nsDefsWithChanges).Msg("validated namespace definitions")
```
vroldanbet, line 138: wouldn't that be a potentially large payload for a trace?

internal/services/v1/schema.go
```diff
@@ -143,13 +149,15 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
...
+				return err
+			}
 		}
 
 		usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
-			DispatchCount: uint32(len(nsdefs)),
+			DispatchCount: uint32(len(nsDefsWithChanges)),
 		})
 
 		// Delete the removed namespaces.
```
vroldanbet, line 163: shouldn't the various statements next to this be part of the if statement above

vroldanbet (commented):
Looks good, just few questions

internal/services/v1/schema.go
josephschorr: Yes, but it is a trace, so better to include the information than not IMO. Trace should only ever be on if you're debugging something

internal/services/v1/schema.go
josephschorr: No; we still need to set the dispatch count (which could be 0)

vroldanbet (approved):

vroldanbet (approved):

PR #920, opened 2022-10-20
proto: implement dispatch PreMarshaler interface

jzelinskie:
This interface enables DispatchResponses to cache their encoded forms, so that it can be written directly to the wire later without re-encoding.


internal/dispatch/caching/caching.go
```diff
@@ -405,10 +419,14 @@ func (cd *Dispatcher) DispatchLookupSubjects(req *v1.DispatchLookupSubjectsReque
 	if cachedResultRaw, found := cd.c.Get(requestKey); found {
 		cd.lookupSubjectsFromCacheCounter.Inc()
 		for _, slice := range cachedResultRaw.([][]byte) {
+			// slice, response := slice, v1.DispatchLookupSubjectsResponse{}
+			// response.Marshaled = slice // Stash the marshaled form for re-use
```
jzelinskie, line 423: I have this commented out for the moment until we figure out how to avoid a panic when 		`dispatchpkg.AddResponseMetadata(respMetadata, result.Metadata)` gets called.

pkg/proto/dispatch/v1/01_codec.go
```diff
@@ -62,6 +74,12 @@ func (vtprotoCodec) Unmarshal(data []byte, v any) error {
...
 }
 
-func init() {
-	encoding.RegisterCodec(vtprotoCodec{})
-}
+func init() { encoding.RegisterCodec(vtprotoCodec{}) }
+
+// Implement the preMarshaler interface
+
+func (r *DispatchCheckResponse) PreMarshaledBytes() []byte              { return r.Marshaled }
```
josephschorr, line 81: Why not just call the field `premarshaled_bytes` and then have the interface be `func GetPremarshaledBytes()`? that way, you don't need to manually implement it

internal/dispatch/caching/caching.go
jzelinskie: Any idea how I should go about this @josephschorr

pkg/testcontainers/testcontainers.go
```diff
@@ -0,0 +1,62 @@
+package testcontainers
```
jzelinskie, line 1: oops I didn't mean to check this in

pkg/proto/dispatch/v1/01_codec.go
jzelinskie: This was a super good point. I totally forgot about the getters that get generated.

jzelinskie:
I'm not convinced that this is worth the hackiness/complexity just to avoid re-encoding in a few scenarios.

PR #921, opened 2022-10-20
introduce caveat support in CockroachDB

vroldanbet:
WIP


internal/datastore/crdb/caveat.go
```diff
@@ -2,24 +2,183 @@ package crdb
...
+	}
+	return rwt.executeWithOverlappingKeys(ctx, func(ctx context.Context, tx pgx.Tx) ([]string, error) {
+		if _, err := tx.Exec(ctx, sql, args...); err != nil {
+			return nil, fmt.Errorf("unable to delete caveats: %w", err)
+		}
+		return names, nil
+	})
+}
+
+func (cr *crdbReader) executeWithOverlappingKeys(ctx context.Context, f func(ctx context.Context, tx pgx.Tx) ([]string, error)) error {
```
ecordell, line 162: this is allocating an extra slice to track overlap keys, why not just loop over the names similar to how we do for namespaces?

there will almost always be only 1 key (and the "keyset" only stores it once), so it's not much help to store it N times. The default is even simpler - we discard the keys entirely (just write to a single static key instead).

internal/datastore/crdb/migrations/zz_migration.0004_add_caveats.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+
+	"github.com/jackc/pgx/v4"
+)
+
+const (
+	createCaveatTable = `CREATE TABLE caveat (
+		name VARCHAR NOT NULL,
+		definition BYTEA NOT NULL,
+    	timestamp TIMESTAMP WITHOUT TIME ZONE DEFAULT now() NOT NULL,
```
ecordell, line 13: nit: looks like a small spacing issue here?

internal/datastore/crdb/watch.go
```diff
@@ -105,12 +112,24 @@ func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Rev
...
-			revision, err := decimal.NewFromString(changeDetails.Updated)
+			revision, err := decimal.NewFromString(details.Updated)
 			if err != nil {
 				errs <- fmt.Errorf("malformed update timestamp: %w", err)
 				return
 			}
 
+			var caveat *core.ContextualizedCaveat
+			if details.After != nil && details.After.CaveatName != "" {
+				strct, err := structpb.NewStruct(details.After.CaveatContext)
```
ecordell, line 123: Isn't it possible to have a caveat with a name, but no context? Does NewStruct like being handed `nil`?

I think this same logic is in a couple of places

internal/datastore/crdb/migrations/zz_migration.0004_add_caveats.go
```diff
@@ -0,0 +1,34 @@
...
+		CONSTRAINT pk_caveat_v1 PRIMARY KEY (name)
+	);`
+
+	addRelationshipCaveatContext = `ALTER TABLE relation_tuple
+		ADD COLUMN caveat_name VARCHAR,
```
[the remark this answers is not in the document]
vroldanbet: No because one may have non caveated relationships

internal/datastore/crdb/watch.go
vroldanbet: yes, that's the case, `NewStruct` does handle well nil map. I've extended the watch test to exercise:
- caveat and nil context
- caveat and zero-value context
- caveat and context with values

I'll look into refactoring this

internal/datastore/crdb/caveat.go
vroldanbet: That is correct, it's a tradeoff I deliberately took, although I'm not 100% sold given the overhead. The goal was to spare us a bunch of repeated lines of code. This the `executeWithOverlappingKeys` could be used in other methods of the readerwriter implementation.

I'm not sure I follow what you mean the keyset stores it just once. Isn't each caveat name added to the keyset to compute the overlapping?
- `WriteCaveats` accepts a slice of caveats. Each name is an overlapping key.
- `DeleteCaveats` also accepts a slice of caveats, same here

internal/datastore/crdb/watch.go
vroldanbet: added a `common.ContextualizedCaveatFrom` method

internal/datastore/crdb/caveat.go
ecordell: > I'm not sure I follow what you mean the keyset stores it just once. Isn't each caveat name added to the keyset to compute the overlapping?

There's two types keyset: `static` and `prefix`.

If it's a static keyset, then `addOverlapKey` is a no-op, we always touch the same key.

If it's a prefix keyset, then `addOverlapKey` extracts the prefix (i.e. `mynamespace/<whatever>`) and only stores the prefix.

In theory you can make requests with multiple prefixes, in practice it's always just 1 prefix.

By collecting up the keys like this, you'll build a big list in memory only to throw it away (because you just needed this single string prefix, not an array of all of the caveat names)

vroldanbet: @ecordell I've updated the code to handle overlapping key one level above and avoid the extra allocation

>In theory you can make requests with multiple prefixes, in practice it's always just 1 prefix.

I'd like to understand this better. Given `WriteCaveats` will be handled by API `WriteSchema`, wouldn't it be possible to send a schema write request with caveats / namespaces with different prefixes? If so, doesn't that require us to loop over all namespaces and caveats to make sure we don't miss any overlapping key?

ecordell: > If so, doesn't that require us to loop over all namespaces and caveats to make sure we don't miss any overlapping key?

Yes, and the "keyer" is for tracking all prefixes from the request. But most deployments will either not use prefixes at all (authzed dedicated, spicedb users) or will have a single prefix enforced a layer up (authzed serverless). And in SpiceDB deployments that do use prefixes, the default "overlap strategy" is `static` (meaning all namespaces write to a single static key).

If you use the "prefix" overlap strategy, then prefixes in SpiceDB w/ CRDB are really "new enemy protection domains" rather than just for organization. I don't think we have any telemetry on it, but I doubt there are many (if any) SpiceDB users who have bothered - it's only really useful if you are trying to enforce multitenancy on top of SpiceDB and/or you're hitting QPS limits with `static` (and can untangle the namespaces in some reasonable way / don't care about newenemy problem)

internal/datastore/crdb/caveat.go
```diff
@@ -2,24 +2,189 @@ package crdb
...
+
+		for rows.Next() {
+			var defBytes []byte
+			var overlappingKey string
+			err = rows.Scan(&overlappingKey, &defBytes)
+			if err != nil {
+				return err
+			}
+			allDefinitionBytes = append(allDefinitionBytes, defBytes)
+			overlappingKeys = append(overlappingKeys, overlappingKey)
```
ecordell, line 101:
```suggestion
			cr.addOverlapKey(val)
```
this bit is still double-tracking the keys

vroldanbet: @ecordell whoops! Fixed!

internal/datastore/crdb/caveat.go
vroldanbet: Thanks for the context, helps a lot! Intentionally using prefix separation to reduce the bottleneck caused by serialization errors could indeed be an option for certain SpiceDB deployments, but still to be seen in real-life (beyond Authzed Serverless 😄 ). In any case, my take away is that, while it is relatively rare, we still need to make sure we account for all potential prefixes in an incoming request.

ecordell (approved):
LGTM

PR #922, opened 2022-10-20
fix more instances of zerolog marshall recursion

vroldanbet:
Follow up to https://github.com/authzed/spicedb/pull/909

Fixes more instances I found of potential stack overflow via zerolog marshalling recursion


ecordell (approved):
LGTM

Nice catch - I thought I grepped for these but apparently missed a few.

PR #924, opened 2022-10-20
add missing caveat test, update caveat example

vroldanbet:
- test delete caveated relationship with `RelationTupleUpdate_DELETE`. The caveat is not part of the primary key of the relationship and so it's ignored when it comes to delete operations
- updates application caveats use-case with "recursing subset" behaviour
- removes occurences of `zerolog/log` global logger in favour of SpiceDB's own global logger

### DIFF
--- a/pkg/caveats/types/map.go
+++ b/pkg/caveats/types/map.go
@@ -1,0 +1,45 @@
+package types
+
+import (
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+)
+
+func init() {
+	registerMethodOnDefinedType(cel.MapType(cel.StringType, cel.AnyType),
+		"isSubtreeOf",
+		[]*cel.Type{cel.MapType(cel.StringType, cel.AnyType)},
+		cel.BoolType,
+		func(arg ...ref.Val) ref.Val {
+			map0 := arg[0].Value().(map[string]any)
+			map1 := arg[1].Value().(map[string]any)
+			return types.Bool(subtree(map0, map1))
+		},
+	)
+}
+
+func subtree(map0 map[string]any, map1 map[string]any) bool {
+	for k, v := range map0 {
+		val, ok := map1[k]
+		if !ok {
+			return false
+		}
+		nestedMap0, ok := v.(map[string]any)
+		if ok {
+			nestedMap1, ok := val.(map[string]any)
+			if !ok {
+				return false
+			}
+			nestedResult := subtree(nestedMap0, nestedMap1)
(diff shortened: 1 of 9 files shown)


pkg/caveats/types/map.go
```diff
@@ -0,0 +1,46 @@
...
+		cel.MemberOverload("subtree_maps",
+			[]*cel.Type{cel.MapType(cel.StringType, cel.AnyType), cel.MapType(cel.StringType, cel.AnyType)},
+			cel.BoolType,
+			cel.FunctionBinding(func(arg ...ref.Val) ref.Val {
+				map0 := arg[0].Value().(map[string]any)
+				map1 := arg[1].Value().(map[string]any)
+				return types.Bool(subtree(map0, map1))
+			})))
+
+	CustomTypes["__map_subtree"] = []cel.EnvOption{subtreeFunction}
```
josephschorr, line 20: This is a bit of a hack... maybe we add a `types.RegisterMethodOnDefinedType` or somesuch, have it take in the function name and its member type, and produce this tree? It could even take in the args and return type.

pkg/datastore/test/caveat.go
```diff
@@ -146,6 +146,17 @@ func WriteCaveatedRelationshipTest(t *testing.T, tester DatastoreTester) {
 	req.NoError(err)
 	assertTupleCorrectlyStored(req, ds, rev, tpl)
 
+	// RelationTupleUpdate_DELETE ignores caveat part of the request
```
josephschorr, line 149: I still think we should make it *error* if the caveat is provided on the delete. Thoughts?

pkg/caveats/types/map.go
vroldanbet: heh, I actually went that path originally 😅  It makes sense to me! These things would allow folks to extend SpiceDB caveat functions, same they can already add their own custom types

pkg/datastore/test/caveat.go
vroldanbet: well, what happens if the original relationship has a caveat, and the user is just sending it _as is_. You'd be forcing them to do extra work, and clearly from the client perspective they are sending "the exact same tuple we just stored". I'm not sold on this idea.

josephschorr: True... but I could also see people being confused by sending the non-caveated rel for a caveated one and then being confused also

vroldanbet: what would confuse them? the rel would be deleted, which is what they would want in the first place. I can understand it feels off to send data that is not used at all as part of the method, but it seems like the least bad of the two options to me.

josephschorr (approved):
LGTM

PR #929, opened 2022-10-21
implement caveat support for spanner

### LINKED ISSUE #917: support caveats in Spanner datastore
Part of https://github.com/authzed/spicedb/issues/835

Add support in `spanner` datastore to support caveats

ecordell:
Fixes #917

### DIFF
--- a/internal/datastore/spanner/caveat.go
+++ b/internal/datastore/spanner/caveat.go
@@ -3,23 +3,134 @@ package spanner
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"cloud.google.com/go/spanner"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc/codes"
+
+	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/pkg/datastore"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 )
 
 func (sr spannerReader) ReadCaveatByName(ctx context.Context, name string) (*core.CaveatDefinition, datastore.Revision, error) {
-	return nil, datastore.NoRevision, fmt.Errorf("unimplemented caveat support in datastore")
+	ctx, span := tracer.Start(ctx, "ReadCaveatByName", trace.WithAttributes(attribute.String("name", name)))
+	defer span.End()
+
+	caveatKey := spanner.Key{name}
+	row, err := sr.txSource().ReadRow(ctx, tableCaveat, caveatKey, []string{colCaveatDefinition, colCaveatTS})
+	if err != nil {
+		if spanner.ErrCode(err) == codes.NotFound {
+			return nil, datastore.NoRevision, datastore.NewCaveatNameNotFoundErr(name)
+		}
+		return nil, datastore.NoRevision, fmt.Errorf(errUnableToReadCaveat, err)
+	}
+	var serialized []byte
+	var updated time.Time
+	if err := row.Columns(&serialized, &updated); err != nil {
+		return nil, datastore.NoRevision, fmt.Errorf(errUnableToReadCaveat, err)
+	}
+
+	loaded := &core.CaveatDefinition{}
+	if err := loaded.UnmarshalVT(serialized); err != nil {
+		return nil, datastore.NoRevision, err
+	}
+	return loaded, revisionFromTimestamp(updated), nil
 }
 
 func (sr spannerReader) ListCaveats(ctx context.Context, caveatNames ...string) ([]*core.CaveatDefinition, error) {
-	return nil, fmt.Errorf("unimplemented caveat support in datastore")
+	ctx, span := tracer.Start(ctx, "ListCaveats", trace.WithAttributes(attribute.StringSlice("names", caveatNames)))
+	defer span.End()
+
+	keyset := spanner.AllKeys()
+	if len(caveatNames) > 0 {
+		keys := make([]spanner.Key, 0, len(caveatNames))
+		for _, n := range caveatNames {
+			keys = append(keys, spanner.Key{n})
+		}
+		keyset = spanner.KeySetFromKeys(keys...)
+	}
+	iter := sr.txSource().Read(
+		ctx,
+		tableCaveat,
+		keyset,
+		[]string{colCaveatDefinition},
+	)
+
+	var caveats []*core.CaveatDefinition
+	if err := iter.Do(func(row *spanner.Row) error {
+		var serialized []byte
+		if err := row.Columns(&serialized); err != nil {
+			return err
+		}
+
+		loaded := &core.CaveatDefinition{}
+		if err := loaded.UnmarshalVT(serialized); err != nil {
+			return err
+		}
+		caveats = append(caveats, loaded)
+
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf(errUnableToListCaveats, err)
+	}
+
+	return caveats, nil
 }
 
 func (rwt spannerReadWriteTXN) WriteCaveats(caveats []*core.CaveatDefinition) error {
-	return fmt.Errorf("unimplemented caveat support in datastore")
+	_, span := tracer.Start(rwt.ctx, "WriteCaveats")
+	defer span.End()
+
+	names := map[string]struct{}{}
+	mutations := make([]*spanner.Mutation, 0, len(caveats))
+	for _, caveat := range caveats {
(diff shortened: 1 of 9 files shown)


internal/datastore/spanner/watch.go
```diff
@@ -102,10 +106,22 @@ func (sd spannerDatastore) loadChanges(
 			&tpl.Subject.Namespace,
 			&tpl.Subject.ObjectId,
 			&tpl.Subject.Relation,
+			&caveatName,
+			&caveatCtx,
 		)
 		if err != nil {
 			return err
 		}
+		if caveatName.Valid && caveatCtx.Valid {
```
vroldanbet, line 115: hrm, why checking `caveatCtx.Valid` though? It's entirely possible to have a caveated relationship with a null context. May be worth checking if our tests exercise getting updates for both relationships with caveat but null context and caveat with non-null context

internal/datastore/spanner/watch.go
```diff
@@ -102,10 +106,22 @@ func (sd spannerDatastore) loadChanges(
...
+			caveatStruct, err := structpb.NewStruct(caveatCtx.Value.(map[string]interface{}))
```
vroldanbet, line 116: Is the `JSON` type from Spanner a map[string]any under the hood, so we know this type assertion won't panic?

also super nit 😄 
```suggestion
			caveatStruct, err := structpb.NewStruct(caveatCtx.Value.(map[string]any))
```

internal/datastore/spanner/caveat.go
```diff
@@ -3,23 +3,122 @@ package spanner
...
+
+	keyset := spanner.AllKeys()
+	if len(caveatNames) > 0 {
+		keys := make([]spanner.Key, 0, len(caveatNames))
+		for _, n := range caveatNames {
+			keys = append(keys, spanner.Key{n})
+		}
+		keyset = spanner.KeySetFromKeys(keys...)
+	}
+	iter := sr.txSource().Read(
```
vroldanbet, line 54: Is Spanner ordering by any key implicitly? Would it be possible for the output to have a non-deterministic order?

internal/datastore/spanner/caveat.go
```diff
@@ -3,23 +3,122 @@ package spanner
...
+
+	return caveats, nil
 }
 
 func (rwt spannerReadWriteTXN) WriteCaveats(caveats []*core.CaveatDefinition) error {
-	return fmt.Errorf("unimplemented caveat support in datastore")
+	_, span := tracer.Start(rwt.ctx, "WriteCaveats")
+	defer span.End()
+
+	names := map[string]struct{}{}
```
vroldanbet, line 86: why not relying on the database to impose key uniqueness instead? I can understand it being faster given it does not have to hit the DB, but we still need to rely on the database anyway given an overlapping transaction could attempt to store the same caveat?

internal/datastore/spanner/reader.go
```diff
@@ -86,14 +87,28 @@ func queryExecutor(txSource txFactory) common.ExecuteQueryFunc {
...
 				&nextTuple.ResourceAndRelation.Namespace,
 				&nextTuple.ResourceAndRelation.ObjectId,
 				&nextTuple.ResourceAndRelation.Relation,
 				&nextTuple.Subject.Namespace,
 				&nextTuple.Subject.ObjectId,
 				&nextTuple.Subject.Relation,
+				&caveatName,
+				&caveatCtx,
 			)
+			if caveatName.Valid && caveatCtx.Valid {
```
vroldanbet, line 102: Same here, why enforcing caveat context needs to be non-null? I'm surprised no test fails, we may have a gap.

internal/datastore/spanner/readwrite.go
```diff
@@ -177,7 +186,16 @@ func deleteWithFilter(ctx context.Context, rwt *spanner.ReadWriteTransaction, fi
 
 func upsertVals(r *core.RelationTuple) []interface{} {
 	key := keyFromRelationship(r)
-	return append(key, spanner.CommitTimestamp)
+	key = append(key, spanner.CommitTimestamp)
+	if r.Caveat != nil && len(r.Caveat.CaveatName) > 0 {
+		key = append(key, r.Caveat.CaveatName)
+		if r.Caveat.Context != nil {
```
vroldanbet, line 192: shouldn't we appending a `nil` context if it's empty? we seem to be doing it in the else block below?

internal/datastore/spanner/caveat.go
josephschorr: Is order implied in this interface though?

internal/datastore/spanner/migrations/zz_migration.0003_add_caveats.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/genproto/googleapis/spanner/admin/database/v1"
+)
+
+const (
+	createCaveatTable = `CREATE TABLE caveat (
+    	name STRING(1024),
```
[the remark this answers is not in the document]
josephschorr: We should add the limits into the core definitions as well; I'll prep a PR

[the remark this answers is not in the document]
josephschorr: Just checked: its limit to 128 bytes in the core protos

internal/datastore/spanner/reader.go
josephschorr: Isn't the context nil when it is unspecified?

internal/datastore/spanner/readwrite.go
```diff
@@ -128,7 +126,9 @@ func deleteWithFilter(ctx context.Context, rwt *spanner.ReadWriteTransaction, fi
 	rel := core.RelationTuple{
 		ResourceAndRelation: &core.ObjectAndRelation{},
 		Subject:             &core.ObjectAndRelation{},
+		Caveat:              &core.ContextualizedCaveat{},
```
josephschorr, line 129: Ideally the caveat starts as `nil`

internal/datastore/spanner/migrations/zz_migration.0003_add_caveats.go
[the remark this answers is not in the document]
vroldanbet: a limit [already exists](https://github.com/authzed/spicedb/blob/376dbfd3385c73f33545c479daff2c4c9e4b0d03/proto/internal/core/v1/core.proto#L44) for the definition in `core.proto`. That's where the 4KB limit comes from. I was bringing up if we should also add those limits in the SQL migrations. I'd argue that if we have limits in the API, and we have the need to increase it, we can avoid a migration.

internal/datastore/spanner/migrations/zz_migration.0003_add_caveats.go
```diff
@@ -0,0 +1,46 @@
...
+    	definition BYTES(MAX) NOT NULL,
+    	timestamp TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true)
+	) PRIMARY KEY (name)`
+
+	addRelationshipCaveatName = `ALTER TABLE relation_tuple
```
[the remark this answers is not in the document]
ecordell: I couldn't find any syntax that would let me add multiple columns in one statement, but these all get sent in 1 DDL request to spanner.

internal/datastore/spanner/migrations/zz_migration.0003_add_caveats.go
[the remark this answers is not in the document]
ecordell: > I'd argue that if we have limits in the API, and we have the need to increase it, we can avoid a migration.

This seems reasonable to me. From [what I can tell](https://stackoverflow.com/questions/45964937/performance-difference-for-stringmax) there's no performance penalty to make these `STRING(MAX)` so doing that is probably the most flexible.

internal/datastore/spanner/caveat.go
ecordell: Not sure why I did this - removed it.

vroldanbet: Idempotency FTW 😄

internal/datastore/spanner/caveat.go
ecordell: I didn't think order mattered here?

vroldanbet: Well, I guess it makes API consumption easier - I think I added it in the first place for some tests, I'd need to revisit if that requirement still holds. CRDB/PSQL implementation do certainly have the order by clause.

vroldanbet (approved):

PR #930, opened 2022-10-23
Encode xmin in postgres revisions to respect zedtoken order

jakedt:


internal/datastore/postgres/revisions.go
```diff
@@ -27,12 +30,17 @@ const (
 	//   %[2] Relationship tuple transaction table
 	//   %[3] Name of timestamp column
 	//   %[4] Quantization period (in nanoseconds)
+	//   %[5] Name of snapshot column
```
ecordell, line 33: `%[5]` isn't used in the query? did you mean for the column name to be parameterizable?

internal/datastore/postgres/revisions.go
```diff
@@ -27,12 +30,17 @@ const (
 	//   %[2] Relationship tuple transaction table
```
ecordell, line 30: in the query description, might be worth adding some info about the `xmin`. It's really detailed right now, but doesn't mention the xmin at all.

internal/datastore/postgres/gc.go
```diff
@@ -71,16 +71,18 @@ func (pgd *pgDatastore) TxIDBefore(ctx context.Context, before time.Time) (datas
...
 func (pgd *pgDatastore) DeleteBeforeTx(ctx context.Context, txID datastore.Revision) (removed common.DeletionCounts, err error) {
+	revision := txID.(postgresRevision)
+
 	// Delete any relationship rows with deleted_transaction <= the transaction ID.
 	removed.Relationships, err = pgd.batchDelete(
 		ctx,
 		tableTuple,
 		relationTuplePKCols,
-		sq.LtOrEq{colDeletedXid: transactionFromRevision(txID)},
+		sq.LtOrEq{colDeletedXid: revision.tx},
```
ecordell, line 85: do we need to query against the xmins too?

I don't know that an application should be reasoning about liveness at the GC threshold, but I think this means you could send in a revision (close to the GC threshold) twice in a row and get different answers.

Ideally what you'd want is just one answer followed by an "revision out of bounds" error, right?

ecordell (commented):
looks good - one real question and a couple of small things, and I still want to run it and test it in the same setting we saw the reversals before just to double check

internal/datastore/memdb/readonly.go
```diff
@@ -18,7 +19,7 @@ type txFactory func() (*memdb.Txn, error)
 type memdbReader struct {
 	TryLocker
 	txSource      txFactory
-	revision      datastore.Revision
+	revision      decimal.Decimal
```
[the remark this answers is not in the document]
jakedt: This is actually vestigial, removing it.

ecordell (dismissed):
LGTM

Tested and didn't see the issues we were seeing before 👍

ecordell (dismissed):

ecordell (approved):
LGTM

PR #932, opened 2022-10-24
Update BaseSubjectSet to support caveat expressions

josephschorr:
This is the first (massive) step in supporting caveats in LookupSubjects, as this implements all the bookkeeping and tracking associated with each subject added to the subject set

First part of #931


internal/util/basesubjectset.go
```diff
@@ -3,210 +3,230 @@ package util
...
+// to the base implementation.
+type Subject[T any] interface {
 	// GetSubjectId returns the ID of the subject. For wildcards, this should be `*`.
 	GetSubjectId() string
 
-	// GetExcludedSubjectIds returns the list of subject IDs excluded. Should only have values
-	// for wildcards.
-	GetExcludedSubjectIds() []string
+	// GetConditionalExpression returns the caveat expression for this subject, if it is conditional.
+	GetConditionalExpression() *v1.CaveatExpression
```
ecordell, line 18: nit: why not just call this `GetCaveatExpression`?

internal/util/basesubjectset.go
```diff
@@ -3,210 +3,230 @@ package util
...
+// a subject ID, its (optional) conditional expression, any excluded subjects, and any sources
+// for bookkeeping. The sources are those other subjects that were combined to create the current
+// subject.
+type constructor[T Subject[T]] func(subjectID string, conditionalExpression *v1.CaveatExpression, excludedSubjects []T, sources ...T) T
+
 // Add adds the found subject to the set. This is equivalent to a Union operation between the
-// existing set of subjects and a set containing the single subject.
+// existing set of subjects and a set containing the single subject, but modifies the set
+// *in place*.
 func (bss BaseSubjectSet[T]) Add(foundSubject T) bool {
```
ecordell, line 58: I haven't gotten to the callsites yet, but it's not obvious to me why `Add` should return a bool or why it's helpful to know if the set had a wildcard before the add.

internal/util/basesubjectset.go
```diff
@@ -3,210 +3,230 @@ package util
...
 	}
+
+	bss.wildcard.subtractConcrete(foundSubject)
+	bss.concrete.subtractConcrete(foundSubject)
 }
 
 // SubtractAll subtracts the other set of subjects from this set of subtracts, modifying this
-// set in place.
+// set *in place*.
 func (bss BaseSubjectSet[T]) SubtractAll(other BaseSubjectSet[T]) {
```
ecordell, line 99: Why not just have `Subtract(sets ...BaseSubjectSet[T])` to cover `Subtract` and `SubtractAll`?

internal/util/basesubjectset.go
```diff
@@ -224,32 +244,581 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+	}
+
+	// Subtraction of a concrete type removes the entry from the concrete list
+	// *unless* the subtraction is conditional, in which case the conditional is updated
+	// to remove the element when it is true.
+	//
+	// Examples:
+	//  {user:sarah} - {user:tom} => {user:sarah}
+	//  {user:tom} - {user:tom} => {}
+	//  {user:tom[c1]} - {user:tom} => {user:tom[c1]}
```
ecordell, line 416: this is surprising

I would read this as: `maybe tom - definitely tom = maybe tom` when intuitively I would think it should be `maybe tom - definitely tom = {}`

internal/util/basesubjectset.go
```diff
@@ -3,210 +3,230 @@ package util
...
-	if wildcard, ok := bss.values[tuple.PublicWildcard]; ok {
-		exclusions := NewSet[string](wildcard.GetExcludedSubjectIds()...)
-		exclusions.Add(foundSubject.GetSubjectId())
-		bss.values[tuple.PublicWildcard] = bss.constructor(tuple.PublicWildcard, exclusions.AsSlice(), wildcard)
+	if len(foundSubject.GetExcludedSubjects()) > 0 {
+		panic("found excluded subjects for non-wildcard")
 	}
+
+	bss.wildcard.subtractConcrete(foundSubject)
+	bss.concrete.subtractConcrete(foundSubject)
```
ecordell, line 94: I think this double-bookkeeping is weird but I don't have an alternate suggestion at the moment.

Every operation actually modifies these two underlying sets, and I'll have to figure out how they get combined elsewhere.

internal/util/basesubjectset.go
```diff
@@ -224,32 +244,581 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+	for _, excludedSubject := range wildcard.GetExcludedSubjects() {
+		exclusions[excludedSubject.GetSubjectId()] = excludedSubject
+	}
+	return exclusions
+}
+
+// combine combines two caveat expressions into one. If one of the expressions is empty, then
+// the entire expression is *short circuited*, and a nil is returned. Otherwise, the expressions
+// are union'ed together.
+func combine(first *v1.CaveatExpression, second *v1.CaveatExpression) *v1.CaveatExpression {
```
ecordell, line 755: a little confusing that this is called by `union*` methods, when there's a `union` with slightly different behavior below.

internal/util/basesubjectset.go
```diff
@@ -224,32 +244,581 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+			Operation: &v1.CaveatOperation{
+				Op:       v1.CaveatOperation_OR,
+				Children: []*v1.CaveatExpression{first, second},
+			},
+		},
+	}
+}
+
+// union unions together two caveat expressions. If one expression is nil, the other is returned.
+func union(first *v1.CaveatExpression, second *v1.CaveatExpression) *v1.CaveatExpression {
```
ecordell, line 771: maybe just `or`?

ecordell (commented):
did an initial pass with questions, definitely haven't read the implementation and tests in detail yet

proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -161,7 +161,8 @@ message DispatchLookupSubjectsRequest {
 
 message FoundSubject {
   string subject_id = 1;
-  repeated string excluded_subject_ids = 2;
+  CaveatExpression conditional_expression = 2;
```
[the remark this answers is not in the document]
josephschorr: It only really matters if we intend back compat and to do so we'd have to fill in the excluded IDs as well. We could do that, but it would require keeping around the extra logic

internal/membership/foundsubject.go
```diff
@@ -58,13 +66,14 @@ func (fs FoundSubject) WildcardType() (string, bool) {
...
-			excludedSubjects = append(excludedSubjects, &core.ObjectAndRelation{
-				Namespace: fs.subject.Namespace,
-				ObjectId:  excludedID,
-				Relation:  fs.subject.Relation,
-			})
+		excludedSubjects := make([]*core.ObjectAndRelation, 0, len(fs.excludedSubjects))
+		for _, excludedSubject := range fs.excludedSubjects {
+			// TODO(jschorr): Fix once we add caveats support to debug tooling
+			if excludedSubject.conditionalExpression != nil {
+				panic("not yet supported")
```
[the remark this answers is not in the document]
josephschorr: Yep! That's intended; I want to do something here, even if it is just returning "(unsupported)", but I don't know exactly what yet, so I added in panics

internal/util/basesubjectset.go
josephschorr: I don't mind changing it

internal/util/basesubjectset.go
josephschorr: standard set semantics: did this subject already exist when I added it

internal/util/basesubjectset.go
josephschorr: `Subtract` takes in a single element, not a set and to have a combined method, we'd have to convert from one or the other

internal/util/basesubjectset.go
josephschorr: Good point... that's definitely not right. I'll fix it

internal/util/basesubjectset.go
josephschorr: The combination is just taking the values in each; they each track their own state independently.

internal/util/basesubjectset.go
josephschorr: Because in the case where we have a subject without a conditional expression at all, we need to treat the entire expression as short circuited. Its not a union: its a short circuiting union

ecordell: I understand what `combine` is doing differently from `union`, I'm just pointing out naming ambiguity

josephschorr: Oh, agreed. Any ideas for a better/good name?

vroldanbet (commented):
Not done with the review, but have some comments

internal/util/basesubjectset.go
josephschorr: Changed

internal/util/basesubjectset.go
josephschorr: Renamed

josephschorr:
Updated

internal/util/basesubjectset.go
```diff
@@ -224,32 +243,611 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+	updatedExclusions := make([]T, 0, len(existing.GetExcludedSubjects()))
+	for _, existingExclusion := range existing.GetExcludedSubjects() {
+		if existingExclusion.GetSubjectId() == concrete.GetSubjectId() {
+			// If the conditional on the concrete is empty, then the concrete is always present, so
+			// we remove the exclusion entirely.
+			if concrete.GetCaveatExpression() == nil {
+				continue
+			}
+
+			// Otherwise, the conditional expression for the new exclusion is the existing expression ||
```
vroldanbet, line 557: code coverage shows this is not being tested. This triggers it, but fails

```go
	{
			"add caveated sub to wildcard with caveated exclusion",
			[]*v1.FoundSubject{
				cwc(nil, csub("foo", caveatexpr("testcaveat"))),
			},
			csub("foo", caveatexpr("testcaveat")),
			true,
			[]*v1.FoundSubject{
				cwc(nil, csub("foo", caveatexpr("testcaveat"))),
			},
		},
```

vroldanbet (commented):
Had final look, this is some complex stuff and I really appreciate how thorough you were with tests and comments. Could only find one test-test that seems to be missing, touches a portion of code that was untested, and is not giving the result (I, at least) expected

Once that's addressed this LGTM

internal/util/basesubjectset.go
josephschorr: I added this test case (and a few others), but had to adjust the expected results to match

josephschorr:
Updated

vroldanbet (approved):

ecordell (dismissed):
LGTM

I'm convinced this works 👍 

Added some extra tests to convince myself in this branch, for future reference: https://github.com/josephschorr/spicedb/compare/josephschorr:dc79c72...ecordell:3b6fbc6

ecordell (approved):

josephschorr:
Updated and rewritten as discussed, and added additional tests

vroldanbet (approved):

internal/util/basesubjectset.go
```diff
@@ -224,32 +184,651 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+	//	{* - {user:tom}} + {* - {user:sarah}} => {*}
+	//	{* - {user:tom, user:sarah}} + {* - {user:sarah}} => {* - {user:sarah}}
+	//	{*}[c1] + {*} => {*}
+	//	{*}[c1] + {*}[c2] => {*}[c1 || c2]
+
+	// NOTE: since we're only using concretes here, it is safe to reuse the BaseSubjectSet itself.
+	exisingConcreteExclusions := NewBaseSubjectSet(constructor)
+	for _, excludedSubject := range existingWildcard.GetExcludedSubjects() {
+		if excludedSubject.GetSubjectId() == tuple.PublicWildcard {
+			panic("wildcards are not allowed in exclusions")
```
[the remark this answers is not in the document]
josephschorr: No. This is simply a sanity check

internal/util/basesubjectset.go
```diff
@@ -3,210 +3,170 @@ package util
...
-// existing set of subjects and a set containing the single subject.
-func (bss BaseSubjectSet[T]) Add(foundSubject T) bool {
-	existing, ok := bss.values[foundSubject.GetSubjectId()]
-	if !ok {
-		bss.values[foundSubject.GetSubjectId()] = foundSubject
-	}
+// existing set of subjects and a set containing the single subject, but modifies the set
+// *in place*.
+func (bss BaseSubjectSet[T]) Add(foundSubject T) {
+	switch foundSubject.GetSubjectId() == tuple.PublicWildcard {
```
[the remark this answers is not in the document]
josephschorr: I originally anticipated having more cases. I'll switch it (no pun intended)

internal/util/basesubjectset.go
```diff
@@ -224,32 +184,651 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+
+type handle[T any] struct {
+	value *T
+}
+
+func newHandle[T any]() *handle[T] {
+	return &handle[T]{}
+}
+
+func (h *handle[T]) getOrNil() *T {
```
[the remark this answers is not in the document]
josephschorr: To make it absolutely clear it can return the value or nil

internal/util/basesubjectset.go
```diff
@@ -224,32 +184,651 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+		return &existingConcrete
+	}
+
+	// The conditional of the exclusion is that of the exclusion itself OR the caveatInverted case of
+	// the wildcard, which would mean the wildcard itself does not apply.
+	exclusionConditional := caveatOr(caveatInvert(wildcardToRemove.GetCaveatExpression()), exclusion.GetCaveatExpression())
+
+	// If the whole exclusion is not conditional, nothing more to do.
+	if exclusionConditional == nil {
+		return &existingConcrete
```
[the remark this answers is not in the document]
josephschorr: Unreachable since the exclusion's caveat must be non-nil from above. Removed the code

internal/util/basesubjectset.go
```diff
@@ -224,32 +184,651 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+}
+
+// exclusionsMapFor creates a map of all the exclusions on a wildcard, by subject ID.
+func exclusionsMapFor[T Subject[T]](wildcard T) map[string]T {
+	exclusions := make(map[string]T, len(wildcard.GetExcludedSubjects()))
+	for _, excludedSubject := range wildcard.GetExcludedSubjects() {
+		exclusions[excludedSubject.GetSubjectId()] = excludedSubject
+	}
+	return exclusions
+}
```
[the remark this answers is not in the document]
josephschorr: It does mean carrying additional load over the wire, but its not that much larger... @ecordell thoughts?

internal/util/basesubjectset.go
```diff
@@ -3,210 +3,170 @@ package util
...
-		bss.Subtract(fs)
+	for _, otherSubject := range other.AsSlice() {
+		bss.Subtract(otherSubject)
 	}
 }
 
 // IntersectionDifference performs an intersection between this set and the other set, modifying
-// this set in place.
+// this set *in place*.
 func (bss BaseSubjectSet[T]) IntersectionDifference(other BaseSubjectSet[T]) {
```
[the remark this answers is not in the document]
josephschorr: `Intersection` implies it returns a new set

internal/util/basesubjectset.go
```diff
@@ -224,32 +184,651 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+	firstWildcard := *first
+	secondWildcard := *second
+
+	concreteExclusions := NewBaseSubjectSet(constructor)
+	for _, excludedSubject := range firstWildcard.GetExcludedSubjects() {
+		if excludedSubject.GetSubjectId() == tuple.PublicWildcard {
+			panic("wildcards are not allowed in exclusions")
+		}
+		concreteExclusions.Add(excludedSubject)
+	}
```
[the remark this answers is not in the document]
josephschorr: It is actually not exactly the same: in one call site it goes into the same set, in the other into different sets. If we switched excluded subjects to a map though, I can simplify this

internal/util/subjectset_test.go
```diff
@@ -356,19 +1013,1734 @@ func TestSubjectSetIntersection(t *testing.T) {
...
+		t.Run(fmt.Sprintf("%v", triple), func(t *testing.T) {
+			// A U (B U C) == (A U B) U C
+
+			A1, A2 := NewSubjectSet(), NewSubjectSet()
+			for _, l := range triple[0] {
+				A1.Add(l)
+				A2.Add(l)
+			}
+			B1, B2 := NewSubjectSet(), NewSubjectSet()
+			for _, l := range triple[0] {
```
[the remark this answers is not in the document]
josephschorr: @ecordell?

internal/util/subjectset_test.go
```diff
@@ -356,19 +1013,1734 @@ func TestSubjectSetIntersection(t *testing.T) {
...
+		t.Run(fmt.Sprintf("%v", triple), func(t *testing.T) {
+			// A ∩ (B ∩ C) == (A ∩ B) ∩ C
+
+			A1, A2 := NewSubjectSet(), NewSubjectSet()
+			for _, l := range triple[0] {
+				A1.Add(l)
+				A2.Add(l)
+			}
+			B1, B2 := NewSubjectSet(), NewSubjectSet()
+			for _, l := range triple[0] {
```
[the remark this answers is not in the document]
josephschorr: @ecordell?

internal/util/basesubjectset.go
```diff
@@ -224,32 +184,651 @@ func (bss BaseSubjectSet[T]) UnionWithSet(other BaseSubjectSet[T]) {
...
+			concrete,
+			wildcardToIntersect,
+			exclusion)
+		return &constructed
+
+	case isExcluded && wildcardToIntersect.GetCaveatExpression() == nil:
+		// If the concrete is excluded and the wildcard conditional is empty, then the concrete is never found.
+		// Example: {user:tom} & {* - user:tom} => {}
+		return nil
+	}
```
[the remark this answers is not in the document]
josephschorr: Removed because the branch right before it handles it

josephschorr:
Updated

internal/util/basesubjectset.go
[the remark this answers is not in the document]
vroldanbet: we could turn this into an assertion function and use it in other spots that could be subject to the same problem

vroldanbet (dismissed):

internal/util/subjectset_test.go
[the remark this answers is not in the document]
ecordell: you can't steal the code and not also steal the bugs! 😄 

yeah these should've been `[0]`, `[1]` and `[2]`, not all `[0]`

internal/util/basesubjectset.go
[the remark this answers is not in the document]
josephschorr: actually, if we changed to a map, we could probably just validate that the `*` is never in the key

internal/util/subjectset_test.go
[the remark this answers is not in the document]
ecordell: yep, the last one should be `triple[2]` as well

internal/util/basesubjectset.go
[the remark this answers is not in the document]
ecordell: do we rely on the order anywhere? changing to a map will lose the order

internal/util/subjectset_test.go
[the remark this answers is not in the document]
josephschorr: and I "borrowed" it :P

josephschorr:
Updated

vroldanbet (approved):

vroldanbet (approved):

ecordell (approved):
test changes LGTM

PR #939, opened 2022-10-25
postgres: add index to relation_tuple_transaction (timestamp)

ecordell:


ecordell:
already present: https://github.com/authzed/spicedb/blob/main/internal/datastore/postgres/migrations/zz_migration.0004_add_transaction_timestamp_index.go#L10

PR #940, opened 2022-10-25
datastore: DeleteNamespace => DeleteNamespaces

### LINKED ISSUE #935: Enhancement: Change DeleteNamespace in datastore interface to accept a slice
This would allow for deletion of multiple namespaces at once

jzelinskie:
This cleans up some of the tracing in the DeleteNamespace path, too.

Fixes #935

### DIFF
--- a/internal/datastore/crdb/readwrite.go
+++ b/internal/datastore/crdb/readwrite.go
@@ -250,18 +250,27 @@ func (rwt *crdbReadWriteTXN) WriteNamespaces(ctx context.Context, newConfigs ...
 	return nil
 }
 
-func (rwt *crdbReadWriteTXN) DeleteNamespace(ctx context.Context, nsName string) error {
-	_, timestamp, err := loadNamespace(ctx, rwt.tx, nsName)
-	if err != nil {
-		if errors.As(err, &datastore.ErrNamespaceNotFound{}) {
-			return err
+func (rwt *crdbReadWriteTXN) DeleteNamespaces(ctx context.Context, nsNames ...string) error {
+	// For each namespace, check they exist and collect predicates for the
+	// "WHERE" clause to delete the namespaces and associated tuples.
+	nsClauses := make([]sq.Sqlizer, 0, len(nsNames))
+	tplClauses := make([]sq.Sqlizer, 0, len(nsNames))
+	for _, nsName := range nsNames {
+		_, timestamp, err := loadNamespace(ctx, rwt.tx, nsName)
+		if err != nil {
+			if errors.As(err, &datastore.ErrNamespaceNotFound{}) {
+				return err
+			}
+			return fmt.Errorf(errUnableToDeleteConfig, err)
 		}
-		return fmt.Errorf(errUnableToDeleteConfig, err)
+
+		for _, nsName := range nsNames {
+			nsClauses = append(nsClauses, sq.Eq{colNamespace: nsName, colTimestamp: timestamp})
+			tplClauses = append(tplClauses, sq.Eq{colNamespace: nsName})
+		}
 	}
 
-	delSQL, delArgs, err := queryDeleteNamespace.
-		Where(sq.Eq{colNamespace: nsName, colTimestamp: timestamp}).
-		ToSql()
+	delSQL, delArgs, err := queryDeleteNamespace.Where(sq.Or(nsClauses)).ToSql()
 	if err != nil {
 		return fmt.Errorf(errUnableToDeleteConfig, err)
 	}
@@ -271,9 +280,7 @@ func (rwt *crdbReadWriteTXN) DeleteNamespace(ctx context.Context, nsName string)
 		return fmt.Errorf(errUnableToDeleteConfig, err)
 	}
 
-	deleteTupleSQL, deleteTupleArgs, err := queryDeleteTuples.
-		Where(sq.Eq{colNamespace: nsName}).
-		ToSql()
+	deleteTupleSQL, deleteTupleArgs, err := queryDeleteTuples.Where(sq.Or(tplClauses)).ToSql()
 	if err != nil {
 		return fmt.Errorf(errUnableToDeleteConfig, err)
 	}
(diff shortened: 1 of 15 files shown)


josephschorr:
We should also issue the deletes in a single SQL statement, where possible

internal/datastore/crdb/readwrite.go
```diff
@@ -267,8 +268,29 @@ func (rwt *crdbReadWriteTXN) WriteNamespaces(newConfigs ...*core.NamespaceDefini
 	return nil
 }
 
-func (rwt *crdbReadWriteTXN) DeleteNamespace(nsName string) error {
-	ctx := datastore.SeparateContextWithTracing(rwt.ctx)
+func (rwt *crdbReadWriteTXN) DeleteNamespaces(nsNames ...string) error {
+	ctx, span := tracer.Start(
+		datastore.SeparateContextWithTracing(rwt.ctx),
+		"DeleteNamespaces",
+		trace.WithAttributes(attribute.StringSlice("names", nsNames)),
```
vroldanbet, line 275: I've been using this sort of "standardized" attribute keys. I think it makes sense to this this to avoid unintended proliferation of keys. WDYT?

https://github.com/authzed/spicedb/blob/114455fcc6f263ef5853fe52be4658de9eb960ca/internal/datastore/common/sql.go#L22-L45

internal/datastore/crdb/readwrite.go
```diff
@@ -267,8 +268,29 @@ func (rwt *crdbReadWriteTXN) WriteNamespaces(newConfigs ...*core.NamespaceDefini
...
+	)
+	defer span.End()
+
+	for _, nsName := range nsNames {
```
vroldanbet, line 279: We want to do this in a single statement instead of looping over, but I'm OK leaving it that way right now since it's the case already, and this PR adjusts the interface to accept multiple namespaces. We should open a follow up issue to address this 👍🏻

internal/datastore/memdb/readwrite.go
```diff
@@ -163,7 +163,7 @@ func (rwt *memdbReadWriteTx) WriteNamespaces(newConfigs ...*core.NamespaceDefini
 	return nil
 }
 
-func (rwt *memdbReadWriteTx) DeleteNamespace(nsName string) error {
+func (rwt *memdbReadWriteTx) DeleteNamespaces(nsNames ...string) error {
 	rwt.lockOrPanic()
```
vroldanbet, line 167: we don't do tracing in memdb. I noticed this as I tested our tracing stack, and it would be nice to have it in place since it allows to very quickly test the tracing stack locally without spinning up a persistent datastore.

Let's open an issue for this?

internal/services/v1alpha1/schema_test.go
```diff
@@ -261,7 +261,7 @@ func TestSchemaReadDeleteAndFailWrite(t *testing.T) {
 	// Issue a delete out of band for the namespace.
 	_, err = ds.ReadWriteTx(context.Background(), func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
 		for _, nsName := range requestedObjectDefNames {
```
vroldanbet, line 263: Given the implementation would eventually be updated to do a single sql statement, we would be loosing that by looping over here. Let's pass `requestedObjectDefNames` instead of looping?

pkg/datastore/datastore.go
```diff
@@ -188,8 +188,8 @@ type ReadWriteTransaction interface {
 	// WriteNamespaces takes proto namespace definitions and persists them.
 	WriteNamespaces(newConfigs ...*core.NamespaceDefinition) error
 
-	// DeleteNamespace deletes a namespace and any associated tuples.
-	DeleteNamespace(nsName string) error
+	// DeleteNamespaces deletes namespaces including associated Relationships.
```
vroldanbet, line 191: TIL `DeleteNamespaces` also deletes relationships! It would be quite handy to expose this as an option in the `WriteSchema` API so that folks don't have to write client code that does the deletion of the relationships! @josephschorr ?

vroldanbet (commented):
Changes look good, I'm OK continuing to loop over deletes as that's the status quou anyway and is an incremental improvement by adjusting the datastore API. I left 2 remarks I think we could address:
- use standardized span attribute keys
- adjust one loop that could very easily be turned into a single `DeleteNamespaces` call

pkg/datastore/datastore.go
josephschorr: Folks having to explicitly empty namespaces is an enforced requirement on purpose: it ensures that someone cannot *accidentally* delete data by simply removing a definition from a schema. The deletion of relationships within the DeleteNamespace call is just an extra check to ensure that we are not left with dangling data

vroldanbet: yeah, I was just thinking of some sort of "do only if you know what you are doing" kind of flag. But yeah, I understand safety and doing the right thing by default comes first

internal/datastore/memdb/readwrite.go
jzelinskie: Fixes in #953

josephschorr (commented):
Can you add a datastore test case that deletes multiple namespaces and ensures their data is removed as well?

internal/datastore/postgres/readwrite.go
```diff
@@ -304,35 +304,46 @@ func (rwt *pgReadWriteTXN) DeleteNamespace(ctx context.Context, nsName string) e
...
 
-	_, createdAt, err := rwt.loadNamespace(ctx, nsName, rwt.tx, filterer)
-	switch {
-	case errors.As(err, &datastore.ErrNamespaceNotFound{}):
-		return err
-	case err == nil:
-		break
-	default:
-		return fmt.Errorf(errUnableToDeleteConfig, err)
+	nsClauses := make([]sq.Sqlizer, 0, len(nsNames))
```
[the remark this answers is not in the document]
josephschorr: I think that's okay, so long as they are purely client side?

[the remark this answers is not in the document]
jzelinskie: Yep

josephschorr (approved):
LGTM

PR #947, opened 2022-10-26
Breakout the integration test suite from unit tests

josephschorr:
The service integration tests (consistency, ops, perf) can take longer to run, so run them as a distinct test suite


.github/workflows/build-test.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: "Build & Test"
-on:  # yamllint disable-line rule:truthy
+on: # yamllint disable-line rule:truthy
```
jzelinskie, line 3: this should cause a warning from yamllint -- this is a regression

josephschorr: Weird; my YAML formatter is changing it

jzelinskie (approved):

PR #950, opened 2022-10-27
Faster CI Feedback Loop

vroldanbet:
A handful of improvements for a faster feedback loop and other miscellaneous changes

- move Analyzer Unit tests to Build / Test job
- rename "Development Package" to "WASM tests"
- faster: remove ineffective cmd/spicedb integration tests: the tests were being ignored due to wrong label, and would have not worked due to missing container image
- faster: build spicedb container before setting up go
- faster: parallel postgres tests (and fix to the race that made them fail)
- do not cancel datastore tests if one fails
- use `setup-go` module caching (disabled by default)

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -31,6 +31,8 @@ on:  # yamllint disable-line rule:truthy
       - "e2e/**"
       - "internal/**"
       - "proto/**"
+env:
+  GO_VERSION: "~1.19.2"
 jobs:
   build:
     name: "Build Binary"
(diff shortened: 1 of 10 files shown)


.github/workflows/build-test.yaml
```diff
@@ -68,40 +72,41 @@ jobs:
...
-        with:
-          tags: "ci,docker"
-          timeout: "5m"
-          working_directory: "cmd/spicedb"
+
   datastore:
     name: "Datastore Tests"
     runs-on: "ubuntu-latest"
     strategy:
+      fail-fast: false
```
jzelinskie, line 101: Why not fail fast?

vroldanbet: because then it cancels other datastore tests which could build independently and be cancelled. Then we would have to rebuild an additional job.

josephschorr (approved):
LGTM

PR #951, opened 2022-10-27
fixes caching of loop variable in the wrong place

vroldanbet:
this is introduced to prevent races, but I messed it up


josephschorr (approved):
LGTM

PR #952, opened 2022-10-27
internal/datastore: add observable proxy

jzelinskie:
This pulls out all tracing into one proxy to deduplicate some boilerplate around tracing the datastore.


vroldanbet (commented):
Was it deliberate to delete a bunch of event / attributes added in the controller method? Why not just fetching the span from the context and leaving those calls there?

jzelinskie:
>Was it deliberate to delete a bunch of event / attributes added in the controller method? Why not just fetching the span from the context and leaving those calls there?

I wanted to just make everything consistent for now. Is there any specific ones that you think are worth keeping?

vroldanbet (approved):
LGTM. Let's start by getting the fundamentals (method level granularity) right, and we can do a later pass adding what we feel is the most useful information in each method

ecordell:
Several of the wrappers omit `SeparateContextWithTracing` - is that intentional?

PR #953, opened 2022-10-27
Add datastore proxy for separating context

jzelinskie:
This removes the footguns for datastores that need to separate context before performing read-only actions.


internal/datastore/crdb/tx_test.go
```diff
@@ -117,7 +117,7 @@ func TestTxReset(t *testing.T) {
 			require := require.New(t)
 
 			ds := b.NewDatastore(t, func(engine, uri string) datastore.Datastore {
-				ds, err := NewCRDBDatastore(
+				ds, err := newCRDBDatastore(
```
vroldanbet, line 120: do we not want the tests with the proxy wrapper that we will be run in practice?

internal/datastore/proxy/caching.go
```diff
@@ -80,7 +80,7 @@ func (r *nsCachingReader) ReadNamespace(
 		loadedRaw, err, _ = r.p.readNsGroup.Do(nsRevisionKey, func() (any, error) {
 			// sever the context so that another branch doesn't cancel the
 			// single-flighted namespace read
-			loaded, updatedRev, err := r.Reader.ReadNamespace(datastore.SeparateContextWithTracing(ctx), nsName)
+			loaded, updatedRev, err := r.Reader.ReadNamespace(SeparateContextWithTracing(ctx), nsName)
```
vroldanbet, line 83: Why are we leaving this one here?

internal/datastore/proxy/context.go
```diff
@@ -0,0 +1,117 @@
...
+// This is useful for datastores that do not want to close connections when a
+// cancel or deadline occurs.
+func NewSeparatingContextDatastoreProxy(d datastore.Datastore) datastore.Datastore {
+	return &ctxProxy{d}
+}
+
+type ctxProxy struct{ delegate datastore.Datastore }
+
+func (p *ctxProxy) ReadWriteTx(ctx context.Context, f datastore.TxUserFunc) (datastore.Revision, error) {
+	return p.delegate.ReadWriteTx(ctx, f)
```
vroldanbet, line 42: maybe clarify we explicitly not want to sever context in operations that may involve writes?

vroldanbet (approved):
This is a nice improvement, thanks a ton! Left just some non-blocking nits, LGTM!

internal/datastore/proxy/caching.go
jzelinskie: We actually don't want to cancel this -- it explicitly has a test to assert this behavior.

internal/datastore/crdb/tx_test.go
jzelinskie: The tests use private fields on the datastore so we need to do this. No tests rely on cancellation.

PR #956, opened 2022-10-28
do not run CRDB migration in transaction

vroldanbet:
This is a recommendation from CRDB docs:

>Schema changes keep your data consistent at all times, but they do not run inside [transactions](https://www.cockroachlabs.com/docs/v22.1/transactions) in the general case. Making schema changes transactional would mean requiring a given schema change to propagate across all the nodes of a cluster. This would block all user-initiated transactions being run by your application, since the schema change would have to commit before any other transactions could make progress. This would prevent the cluster from servicing reads and writes during the schema change, requiring application downtime.

### DIFF
--- a/internal/datastore/crdb/migrations/manager.go
+++ b/internal/datastore/crdb/migrations/manager.go
@@ -6,7 +6,10 @@ import (
 	"github.com/jackc/pgx/v4"
 )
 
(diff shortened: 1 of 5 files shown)


ecordell (dismissed):
LGTM

ecordell (approved):

Issue #959, opened 2022-10-28
CI cache poisoning


vroldanbet:
Our CI jobs running tests rely on caching to skip the whole `go mod download` phase, which usually eats a lot of time. Caching was introduced recently in https://github.com/authzed/spicedb/pull/950, which can shove up to 2m in the best case.

Unfortunately, given we run multiple tests as individual jobs, and due to go's toolchain only downloading the modules needed as of recent versions, sometimes our CI my end up caching a version that has a small subset of modules, and as a consequence we lose the boost in CI time.

We should investigate why this happens, which job is doing this, and how can we make sure it does not write the local go empty module cache to CI cache.

vroldanbet:
The Analyzer CI build seems to be the culprit:

```
Post job cleanup.
/opt/hostedtoolcache/go/1.19.[2](https://github.com/authzed/spicedb/actions/runs/3347243555/jobs/5545015920#step:7:2)/x64/bin/go env GOMODCACHE
/opt/hostedtoolcache/go/1.19.2/x64/bin/go env GOCACHE
/home/runner/go/pkg/mod
/home/runner/.cache/go-build
/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/spicedb/spicedb --files-from manifest.txt
Cache Size: ~12 MB (12648049 B)
Cache saved successfully
Cache saved with the key: setup-go-Linux-go-~1.19.2-ee46[3](https://github.com/authzed/spicedb/actions/runs/3347243555/jobs/5545015920#step:7:3)255313e88769c6[4](https://github.com/authzed/spicedb/actions/runs/3347243555/jobs/5545015920#step:7:4)1f1b16cb6f[5](https://github.com/authzed/spicedb/actions/runs/3347243555/jobs/5545015920#step:7:5)4cac77c811a33f2a2f952[6](https://github.com/authzed/spicedb/actions/runs/3347243555/jobs/5545015920#step:7:6)9153a1315[7](https://github.com/authzed/spicedb/actions/runs/3347243555/jobs/5545015920#step:7:7)e
```

vroldanbet:
Reopening because we still have poisoning between builds doing `go build` and builds doing `go test`. It's a difference of 200MB of dependencies.

PR #960, opened 2022-10-28
specifies cache-dependency-path to prevent poisoning

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/959

These modules have very little dependencies, and end up "winning" in the race to cache, poisoning the cache.

By introducing the path we can have the Github Action disambiguate the cache key name by path.

From the docs:
>The action defaults to search for the dependency file - go.sum in the repository root, and uses its hash as a part of the cache key. Use cache-dependency-path input for cases when multiple dependency files are used, or they are located in different subdirectories.


josephschorr (approved):
LGTM

PR #961, opened 2022-10-28
prevent poisoning via build job

vroldanbet:
Close https://github.com/authzed/spicedb/issues/959 (for realz?)

- building only downloads the deps needed by non-test code, which is a smaller subset. We fix it by calling `go mod download` before `go build`, which will get all dependencies
- Protobuf build also poisoned the cache, disabled cache there since its a fast build
- WASM also poisoned the cache, disabled cache there since it's relatively fast

I checked the logs and I believe this time the problem should be circumented. We fare better by using the cache action explicitly instead of using `setup-go` built in functionality

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -43,6 +43,8 @@ jobs:
         with:
           go-version: "${{ env.GO_VERSION }}"
           cache: "true"
+      - name: "go mod download" # prevent CI cache poisoning by getting all deps
+        run: "go mod download"
       - uses: "authzed/actions/go-build@main"
 
   image-build:
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -198,6 +198,7 @@ jobs:
       - uses: "actions/setup-go@v3"
         with:
           go-version: "${{ env.GO_VERSION }}"
+          cache: "false" # do not cache to prevent cache poisoning
       - name: "Install Go Tools"
         run: "./hack/install-tools.sh"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yaml
```diff
@@ -182,7 +184,7 @@ jobs:
       - uses: "actions/setup-go@v3"
         with:
           go-version: "${{ env.GO_VERSION }}"
-          cache: "true"
+          cache: "false" # do not cache to prevent cache poisoning
```
josephschorr, line 187: Can you add the explicit cache false below too and document?

.github/workflows/build-test.yaml
```diff
@@ -43,6 +43,8 @@ jobs:
         with:
           go-version: "${{ env.GO_VERSION }}"
           cache: "true"
+      - name: "go mod download" # prevent CI cache poisoning by getting all deps
```
josephschorr, line 46: How, exactly, does this not poison the cache?

vroldanbet: go build omits all dependencies used for testing. Go mod download downloads all the modules. By calling it before go build, we force all modules to be downloaded.

.github/workflows/build-test.yaml
vroldanbet: done: https://github.com/authzed/spicedb/pull/961/commits/f0327e7c73813a5d53ee4b22843a1f164190b94e

josephschorr (approved):
LGTM

PR #962, opened 2022-10-29
Change experimental caveats flag to be handled at the service level

josephschorr:
Previously, we did so at the datastore level, but now that most datastores are getting support, we do so at the service level


ecordell (approved):
LGTM

PR #963, opened 2022-10-31
Debug API improvements

josephschorr:
Improvements to the debug API, including bug fixes and more accurate tracing


vroldanbet:
@josephschorr please rebase 🙏🏻

internal/graph/check.go
```diff
@@ -428,10 +436,6 @@ func (cc *ConcurrentChecker) checkTupleToUserset(ctx context.Context, crc curren
 				return childResult
 			}
 
-			if childResult.Err != nil {
-				return childResult
-			}
```
vroldanbet, line 433: funny 😄

internal/graph/computed/computecheck.go
```diff
@@ -11,14 +11,22 @@ import (
 	v1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
 )
 
+type DebugOption int
+
+const (
+	NoDebugging           DebugOption = 0
+	BasicDebuggingEnabled DebugOption = 1
+	TraceDebuggingEnabled DebugOption = 2
```
vroldanbet, line 19: Can we define what basic vs trace means so that users can figure out what the expectation is?

I almost feel like we want to have a set of multiple knobs like "debug_dispatch" which is additive, instead of having "pre-canned levels". But maybe it's just overkill and those levels work for most folks.

proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -19,6 +19,7 @@ message DispatchCheckRequest {
   enum DebugSetting {
     NO_DEBUG = 0;
     ENABLE_DEBUGGING = 1;
```
vroldanbet, line 21: maybe we want to rename this to BASIC debugging? it won't be a breaking change in the API.

pkg/util/chunking.go
```diff
@@ -1,7 +1,13 @@
 package util
 
+import "fmt"
+
 // ForEachChunk executes the given handler for each chunk of items in the slice.
 func ForEachChunk[T any](data []T, chunkSize uint64, handler func(items []T)) {
+	if chunkSize <= 0 {
+		panic(fmt.Sprintf("invalid chunk size: %d", chunkSize))
```
vroldanbet, line 8: I wonder if we need to panic. It's clearly an error, but can we recover from it by ignoring that value and setting it to 1, and logging the error?

I'm just growing wary of the panics.

internal/graph/computed/computecheck.go
josephschorr: I can add doc comments, yeah

pkg/util/chunking.go
josephschorr: Sure

josephschorr:
Updated

pkg/util/chunking.go
```diff
@@ -1,7 +1,18 @@
...
+import (
+	"fmt"
+
+	"github.com/authzed/spicedb/internal/logging"
+)
+
 // ForEachChunk executes the given handler for each chunk of items in the slice.
 func ForEachChunk[T any](data []T, chunkSize uint64, handler func(items []T)) {
+	if chunkSize == 0 {
+		logging.Err(fmt.Errorf("ForEachChunk was given an invalid chunk size: %d", chunkSize))
```
vroldanbet, line 12: hrm, I'm not sure this message will be `sent` because it's missing `Msg` call or `Send` call. Can you double check?

josephschorr: Lint caught it; fixed now

vroldanbet (dismissed):
LGTM!

vroldanbet (dismissed):

vroldanbet (approved):

PR #976, opened 2022-11-03
Remove support for the v1alpha1 API

### LINKED ISSUE #943: Deprecation Notice: v1alpha1 API
This is a formal notice that the [v1alpha1 API](https://buf.build/authzed/api/docs/main:authzed.api.v1alpha1) is being marked as deprecated for removal in the next released version of SpiceDB. The `ReadSchema` and `WriteSchema` methods have been replaced with their equivalent versions in [v1](https://buf.build/authzed/api/docs/main:authzed.api.v1), while `WatchResourcesService` is currently an alpha API that will be moved into an experimental package.

josephschorr:
Fixes #943


.github/labeler.yml
```diff
@@ -1,36 +1,35 @@
...
+  - 'DCO'
+  - 'LICENSE'
+  - 'README.md'
+'area/schema':
+  - 'pkg/schemadsl/**/*'
+'area/tooling':
+  - '**/*_test.go'
+  - '.github/**/*'
+  - '.*'
+  - 'Dockerfile*'
```
jzelinskie, line 35: Why this change?

internal/services/server.go
```diff
@@ -30,6 +28,10 @@ const (
 	// V1SchemaServiceEnabled indicates that the V1 schema service is enabled.
 	V1SchemaServiceEnabled SchemaServiceOption = 1
 
+	// V1SchemaServiceAdditiveOnly indicates that the V1 schema service is enabled in additive-only
+	// mode for testing.
+	V1SchemaServiceAdditiveOnly SchemaServiceOption = 2
```
jzelinskie, line 33: Can we switch this section to use `iota`?

internal/services/v1/schema.go
```diff
@@ -16,23 +14,22 @@ import (
...
-	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	dispatchv1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
 	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/schemadsl/generator"
 	"github.com/authzed/spicedb/pkg/schemadsl/input"
 )
 
 // NewSchemaServer creates a SchemaServiceServer instance.
-func NewSchemaServer(caveatsEnabled bool) v1.SchemaServiceServer {
+func NewSchemaServer(additiveOnly bool, caveatsEnabled bool) v1.SchemaServiceServer {
```
jzelinskie, line 26: `additiveOnly, caveatsEnabled bool`

pkg/cmd/serve.go
```diff
@@ -95,6 +95,11 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) {
 	cmd.Flags().Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	cmd.Flags().Uint16Var(&config.MaximumPreconditionCount, "update-relationships-max-preconditions-per-call", 1000, "maximum number of preconditions allowed for WriteRelationships and DeleteRelationships calls")
 
+	cmd.Flags().BoolVar(&config.V1SchemaAdditiveOnly, "for-testing-v1-schema-additive-only", false, "changes V1 schema API to be additive only")
+	if err := cmd.Flags().MarkHidden("for-testing-v1-schema-additive-only"); err != nil {
```
jzelinskie, line 99: I think this is slightly more descriptive:

`testing-only-schema-additive-writes`
"append new definitions to the existing schema, rather than overwriting it"

jzelinskie (changes requested):

internal/services/server.go
josephschorr: I thought general consensus lately was not to use it, lest you accidentally add one in between and change it?

josephschorr:
Updated

internal/services/server.go
jzelinskie: Ah, that's a pretty good argument. I've never seen it happen, though.

jzelinskie (approved):

PR #981, opened 2022-11-04
Add validation of relationships to file loader

josephschorr:
- Adds validation of relationships being used in file loader
- Removes support for deprecated fields in the YAML file format
- Changes some of the validation errors into properly typed errors

### DIFF
--- a/internal/relationships/validation.go
+++ b/internal/relationships/validation.go
@@ -1,0 +1,151 @@
+package relationships
+
+import (
+	"context"
+
+	"github.com/authzed/spicedb/internal/namespace"
+	"github.com/authzed/spicedb/pkg/caveats"
+	"github.com/authzed/spicedb/pkg/datastore"
+	ns "github.com/authzed/spicedb/pkg/namespace"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/tuple"
+	"github.com/authzed/spicedb/pkg/util"
+)
+
+// ValidateRelationshipUpdates performs validation on the given relationship updates, ensuring that
+// they can be applied against the datastore.
+func ValidateRelationshipUpdates(
+	ctx context.Context,
+	rwt datastore.ReadWriteTransaction,
+	updates []*core.RelationTupleUpdate,
+) error {
+	// Load caveats, if any.
+	var referencedCaveatMap map[string]*core.CaveatDefinition
+	referencedCaveatNamesWithContext := util.NewSet[string]()
+
+	for _, update := range updates {
+		if hasNonEmptyCaveatContext(update) {
+			referencedCaveatNamesWithContext.Add(update.Tuple.Caveat.CaveatName)
+		}
+	}
+
+	if !referencedCaveatNamesWithContext.IsEmpty() {
+		foundCaveats, err := rwt.ListCaveats(ctx, referencedCaveatNamesWithContext.AsSlice()...)
+		if err != nil {
+			return err
+		}
+
+		referencedCaveatMap = make(map[string]*core.CaveatDefinition, len(foundCaveats))
+		for _, caveatDef := range foundCaveats {
+			referencedCaveatMap[caveatDef.Name] = caveatDef
+		}
+	}
+
+	// TODO(jschorr): look into loading the type system once per type, rather than once per relationship
+	// Check each update.
+	for _, update := range updates {
+		// Validate the IDs of the resource and subject.
+		if err := tuple.ValidateResourceID(update.Tuple.ResourceAndRelation.ObjectId); err != nil {
+			return err
+		}
+
+		if err := tuple.ValidateSubjectID(update.Tuple.Subject.ObjectId); err != nil {
+			return err
+		}
+
+		// Ensure the namespace and relation for the resource and subject exist.
+		if err := namespace.CheckNamespaceAndRelation(
+			ctx,
+			update.Tuple.ResourceAndRelation.Namespace,
+			update.Tuple.ResourceAndRelation.Relation,
+			false,
+			rwt,
+		); err != nil {
+			return err
+		}
+
+		if err := namespace.CheckNamespaceAndRelation(
+			ctx,
+			update.Tuple.Subject.Namespace,
+			update.Tuple.Subject.Relation,
+			true,
+			rwt,
+		); err != nil {
+			return err
+		}
+
+		// Build the type system for the object type.
+		_, ts, err := namespace.ReadNamespaceAndTypes(
+			ctx,
+			update.Tuple.ResourceAndRelation.Namespace,
+			rwt,
+		)
+		if err != nil {
+			return err
+		}
+
+		// Validate that the relationship is not writing to a permission.
+		if ts.IsPermission(update.Tuple.ResourceAndRelation.Relation) {
+			return NewCannotWriteToPermissionError(update)
+		}
+
+		// Validate the subject against the allowed relation(s).
+		var relationToCheck *core.AllowedRelation
+		var caveat *core.AllowedCaveat
+
+		if update.Tuple.Caveat != nil {
+			caveat = ns.AllowedCaveat(update.Tuple.Caveat.CaveatName)
+		}
+
+		if update.Tuple.Subject.ObjectId == tuple.PublicWildcard {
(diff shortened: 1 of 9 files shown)


pkg/validationfile/loader_test.go
```diff
@@ -15,39 +12,68 @@ import (
...
+				"example/project:pied_piper#writer@example/user:freyja",
+				"example/project:pied_piper#owner@example/user:fred",
+				"example/project:pied_piper#reader@example/user:tom",
+				"example/project:pied_piper#writer@example/user:sarah",
 			},
+			expectedError: "",
+		},
+		{
+			name:      "multiple files",
+			filePaths: []string{"testdata/initial_schema_and_rels.yaml", "testdata/just_rels.yaml"},
```
ecordell, line 55: I think I mentioned this before, but what about when two have different schemas?

internal/relationships/validation.go
```diff
@@ -0,0 +1,151 @@
...
+		if err != nil {
+			return err
+		}
+
+		if isAllowed != namespace.AllowedRelationValid {
+			return NewInvalidSubjectTypeError(update, relationToCheck)
+		}
+
+		// Validate caveat and its context, if applicable.
+		// TODO(jschorr): once caveats are supported on all datastores, we should elide this check if the
```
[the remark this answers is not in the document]
josephschorr: Yes, but I wrote this PR before that was the case, and I'm not fixing caveat-related TODOs until a pass after everything is done

pkg/validationfile/loader_test.go
josephschorr: They'd be combined

josephschorr:
Rebased

internal/services/v1/relationships.go
```diff
@@ -162,31 +161,23 @@ func (ps *permissionServer) WriteRelationships(ctx context.Context, req *v1.Writ
...
-				),
+				NewDuplicateRelationshipErr(update),
 			)
 		}
 
-		// Only load the caveat if we need its type information for context checking.
-		if hasNonEmptyCaveatContext(update) {
-			if !ps.caveatsEnabled {
+		if !ps.caveatsEnabled {
+			if update.Relationship.OptionalCaveat != nil && update.Relationship.OptionalCaveat.CaveatName != "" {
```
vroldanbet, line 174: the previous logic seemed incorrect, I think? It checked if caveat was non-empty first instead of checking that the functionality was disabled first. The `hasNonEmptyCaveatContext ` method checked relationships had caveat *and context was not empty*, which means it would have allowed writing empty rels with empty context, despite the disablement flag?

The new logic seems correct 👍🏻

internal/services/v1/relationships.go
```diff
@@ -162,31 +161,23 @@ func (ps *permissionServer) WriteRelationships(ctx context.Context, req *v1.Writ
...
 				return nil, fmt.Errorf("caveats are currently not supported")
```
vroldanbet, line 175: shouldn't we create a new typed error here and rewrite it? This will end up in gRPC code `UNKNOWN`.

pkg/validationfile/loader.go
```diff
@@ -52,124 +51,96 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*PopulatedVa
...
 			return nil, datastore.NoRevision, fmt.Errorf("error when parsing config file %s: %w", filePath, err)
 		}
 
 		files = append(files, *parsed)
 
-		// Add schema-based namespace definitions.
+		// Disallow legacy sections.
+		if len(parsed.NamespaceConfigs) > 0 {
+			return nil, revision, fmt.Errorf("definitions must be specified in `schema`")
+		}
```
vroldanbet, line 77: Do you think folks out there could have validation files using legacy representation, and this will make it stop parsing? Do we have any recommendations for those folks to migrate to the new representation? (maybe a note in the release?

pkg/validationfile/loader.go
```diff
@@ -52,124 +51,96 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*PopulatedVa
...
 		}
 
-		wrevision, terr := ds.ReadWriteTx(ctx, func(rwt datastore.ReadWriteTransaction) error {
-			return rwt.WriteRelationships(ctx, updates)
-		})
-		if terr != nil {
-			return nil, datastore.NoRevision, fmt.Errorf("error when loading validation tuples from file %s: %w", filePath, terr)
+		err = relationships.ValidateRelationshipUpdates(ctx, rwt, updates)
+		if err != nil {
+			return err
```
vroldanbet, line 140: Isn't this something that should happen at the beginning of the transaction? You would have written caveats before validating it. Yes, eventually TX would rollback, but if feels like validation should happen first in order to fail fast?

pkg/validationfile/loader.go
```diff
@@ -52,124 +51,96 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*PopulatedVa
...
 func PopulateFromFilesContents(ds datastore.Datastore, filesContents map[string][]byte) (*PopulatedValidationFile, datastore.Revision, error) {
-	var revision datastore.Revision
-	var nsDefs []*core.NamespaceDefinition
+	var schema string
+	var objectDefs []*core.NamespaceDefinition
 	var caveatDefs []*core.CaveatDefinition
-	schema := ""
 	var tuples []*core.RelationTuple
+	var updates []*core.RelationTupleUpdate
+
```
vroldanbet, line 59:
```suggestion
```

pkg/validationfile/loader.go
```diff
@@ -52,124 +51,96 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*PopulatedVa
...
-			if nerr != nil {
-				return nil, revision, fmt.Errorf("error when parsing namespace config #%v from file %s: %w", index, filePath, nerr)
-			}
-			nsDefs = append(nsDefs, &nsDef)
+		// Parse relationships for updates.
+		for _, rel := range parsed.Relationships.Relationships {
+			tpl := tuple.MustFromRelationship(rel)
+			updates = append(updates, tuple.Touch(tpl))
+			tuples = append(tuples, tpl)
 		}
```
vroldanbet, line 100: This seems like a nice snippet to refactor into a helper at `pkg/tuple/tuple.go`.

pkg/validationfile/loader.go
```diff
@@ -52,124 +51,96 @@ func PopulateFromFiles(ds datastore.Datastore, filePaths []string) (*PopulatedVa
...
+	}
 
-		ctx := context.Background()
+	// Load the definitions and relationships into the datastore.
+	ctx := context.Background()
```
vroldanbet, line 104: Why is the loader not receiving a context? seems like an operation that could time out, get canceled...

vroldanbet (approved):
LGTM! left a handful of suggestions 🙏🏻

internal/services/v1/relationships.go
josephschorr: As discussed offline, the caveat itself would still have been checked below, but this was deemed safer too :)

internal/services/v1/relationships.go
josephschorr: Which is fine for now; this is temporary until such time until the flag is removed

pkg/validationfile/loader.go
josephschorr: I plan to add a note in the release and yes, that's why I expressly turned it into an error, although I'd be quite surprised if anyone is actually using the old format

pkg/validationfile/loader.go
josephschorr: I actually put it there expressly to make it more readable

pkg/validationfile/loader.go
josephschorr: Was never really necessary; I suppose we could add it

internal/services/v1/relationships.go
vroldanbet: 👍🏻

pkg/validationfile/loader.go
vroldanbet: 👍🏻

pkg/validationfile/loader.go
vroldanbet: 👍🏻

PR #992, opened 2022-11-09
datastore/postgres: remove the compensation code for migration phases

### LINKED ISSUE #993: Drop postgres multi-phase migrations code paths.

jakedt:
Closes #993


vroldanbet (commented):
LGTM, but:

It seems the `migrationPhase` CLI flag is still there. I assume that was deliberate so we don't break clients from 0.14.0, but I think we could clean up logged messages and remove the validation of the phase?

vroldanbet (approved):
Approving after chatting with @jakedt - the idea is that what's left of the 4-phase migration will be reused for any future 4-phase migration that may emerge.

PR #994, opened 2022-11-10
Fix for PG when schema is specified in the db url

jvassev:
This patch lets you put the datastore in any schema like this: postgres://user:pass@host/spicedb?**search_path=hello_world**


jvassev:

jakedt:
Hi @jvassev thanks for the PR. This change seems fine, but I'm curious what is your motivating factor to have more than one schema vs having more than one database?

jvassev:
Hi @jakedt 
In my case it was just less friction to reuse an existing database and live in a dedicated schema.
With a custom build i was able to pass the migration step and try out the demo content and `zed` !

vroldanbet (approved):

PR #997, opened 2022-11-11
Make sure ReadSchema returns caveats as well

josephschorr:


internal/services/v1/schema.go
```diff
@@ -51,22 +50,32 @@ func (ss *schemaServer) ReadSchema(ctx context.Context, in *v1.ReadSchemaRequest
 		return nil, rewriteError(ctx, err)
 	}
 
+	caveatDefs, err := ds.ListCaveats(ctx)
+	if err != nil {
+		return nil, rewriteError(ctx, err)
+	}
+
 	if len(nsDefs) == 0 {
```
vroldanbet, line 58: Do we have a test that prevents writing a schema with just caveats?

vroldanbet (approved):

Issue #1000, opened 2022-11-18
HTTP Gateway is not gracefully terminated


vroldanbet:
The HTTP Gateway opens new connections when `RunnableServer.Complete()` is called, and it does this with a `context.TODO()` call:

https://github.com/authzed/spicedb/blob/4190b0588f3a708b107e90666d9bbb270d127c84/pkg/cmd/server/server.go#L283

As a consequence, the connections that are created are not gracefully terminated since the cancelation signal never arrives 

https://github.com/authzed/spicedb/blob/14e65851381ab0b0b3860bcd6e28792e73134a83/internal/gateway/gateway.go#L53-L61

PR #1001, opened 2022-11-18
Http gateway graceful termination

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/1000
Also fixes a flakey test

As I was trying to make the cert test not flakey, I discovered with goleak that HTTP Gateway wasn't terminating gracefully.
This is due to the fact gRPC connections to upstream are created with a `context.TODO()` and thus cancellation wasn't being propagated.

The test was also modified. Sometimes it would hang indefinitely. This is because the gRPC connection created initially will retry indefinitely if the cert happened to expire (expiration is set to 1s after it's created). This would cause CI jobs to hang until timed out.

This makes it fail fast in case of a race. An actual proper fix to the race is not in scope of this PR, but at least now we will see how often that happens. Increasing the time it takes for expiration would reduce the likelihood.


internal/services/integrationtesting/cert_test.go
```diff
@@ -166,17 +170,15 @@ func TestCertRotation(t *testing.T) {
 	)
 
 	require.NoError(t, err)
-	t.Cleanup(func() {
+	defer func() {
```
josephschorr, line 173: why this change?

vroldanbet: because I added `<-wait` to make sure the `RunnableServer.Run()` call returned, which can only happen if cancellation is invoked. 
Same for `conn.Close`: if it was happening in `Cleanup` the call to goleak will happen first, and fail

internal/gateway/gateway.go
```diff
@@ -31,9 +31,9 @@ var histogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
 	Help:      "A histogram of the duration spent processing requests to the SpiceDB REST Gateway.",
 }, []string{"method"})
 
-// NewHandler creates an REST gateway HTTP Handler with the provided upstream
+// NewHandler creates an REST gateway HTTP ConnectionCloser with the provided upstream
```
josephschorr, line 34: `CloserHandler`

vroldanbet: whoops! fixed 👍🏻

josephschorr (approved):
LGTM

Issue #741, opened 2022-08-10
PostrgreSQL datastores do not timeout on initialization


vroldanbet:
If PostgreSQL datastores are initialized on an unreachable/unresponsive database backend, the initialization hangs indefinitely, and the spicedb process wont report becoming healthy.

We can see this is caused by a missing deadline:

https://github.com/authzed/spicedb/blob/26d13e56948492537e02501c8d018e0def5ad204/internal/datastore/crdb/crdb.go#L92-L95

vroldanbet:
This should be addressed for CRDB as of https://github.com/authzed/spicedb/pull/1042

josephschorr:
@vroldanbet is this complete as of https://github.com/authzed/spicedb/blob/main/internal/datastore/postgres/postgres.go#L160?

vroldanbet:
@josephschorr yup!

PR #1002, opened 2022-11-18
Add context and default timeout for validationfile loading

josephschorr:
Pipes context through to the validationfile loading and adds a default timeout when used for bootstrapping

### DIFF
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -58,6 +58,7 @@ type Config struct {
 	// Bootstrap
 	BootstrapFiles     []string
 	BootstrapOverwrite bool
+	BootstrapTimeout   time.Duration
 
 	// Hedging
 	RequestHedgingEnabled          bool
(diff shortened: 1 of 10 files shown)


pkg/cmd/datastore/datastore.go
```diff
@@ -106,6 +107,7 @@ func RegisterDatastoreFlags(cmd *cobra.Command, opts *Config) {
 	cmd.Flags().BoolVar(&opts.ReadOnly, "datastore-readonly", false, "set the service to read-only mode")
 	cmd.Flags().StringSliceVar(&opts.BootstrapFiles, "datastore-bootstrap-files", []string{}, "bootstrap data yaml files to load")
 	cmd.Flags().BoolVar(&opts.BootstrapOverwrite, "datastore-bootstrap-overwrite", false, "overwrite any existing data with bootstrap data")
+	cmd.Flags().DurationVar(&opts.RevisionQuantization, "datastore-bootstrap-timeout", 10*time.Second, "maximum duration before timeout for the bootstrap data to be written")
```
vroldanbet, line 110:
```suggestion
	cmd.Flags().DurationVar(&opts. BootstrapTimeout, "datastore-bootstrap-timeout", 10*time.Second, "maximum duration before timeout for the bootstrap data to be written")
```

vroldanbet: I think it should be possible to now fix https://github.com/authzed/spicedb/issues/741 with this change. We don't have to do it in this PR.

pkg/tuple/tuple.go
```diff
@@ -293,6 +293,16 @@ func MustFromRelationship(r *v1.Relationship) *core.RelationTuple {
 	return FromRelationship(r)
 }
 
+// MustFromRelationships converts a slice of Relationship's into a slice of RelationTuple's.
+func MustFromRelationships(rels []*v1.Relationship) []*core.RelationTuple {
```
vroldanbet, line 297: I think you meant to use this in `loader.go` right?

vroldanbet (changes requested):

pkg/tuple/tuple.go
josephschorr: Can't; we need to build updates in there too, so I just added the method

vroldanbet: What I actually suggested in my review was to create a method that took "relationship" and returned touch updates. You can have an argument in the method with the kind of update operation, so it can also be CREATE or DELETE.

josephschorr: We need to add it to two slices though, so that won't work either. As discussed offline, leaving as-is

vroldanbet (approved):

PR #1006, opened 2022-11-20
Add len checks to WriteCaveats before attempting to write nothing

### LINKED ISSUE #1005: Fail to bootstrap a cockroachdb-engine backed instance with yaml file
I'm having trouble bootstrapping my cockroachdb-backed spicedb with a `--datastore-bootstrap-files` yaml option. Getting a caveat-looking issue.

```sh
5:22AM ERR terminated with errors error="failed to create datastore: failed to load bootstrap files: unable to write new caveat revision: insert statements must have at least one set of values or select clause"
```

I'm pretty new to spicedb, but I have an inclination that when the [bootstrapping file](https://github.com/authzed/spicedb/blob/main/pkg/validationfile/loader.go#L106) loads the `caveatDefs` and then calls `rwt.WriteCaveats`, we hit an [empty](https://github.com/authzed/spicedb/blob/main/internal/datastore/crdb/caveat.go#L125) `write.ToSql()` .

To confirm I added a single caveat to my bootstrapping yaml and it worked without an issue.

To Reproduce:

- Start cockroach
  ` docker run --rm -d --name=roach -p 8080:8080 -p 26257:26257 cockroachdb/cockroach:latest start-single-node --insecure`
- Migrate head
  ` spicedb migrate head --datastore-conn-uri="postgresql://root@localhost:26257/defaultdb?sslmode=disable&max_conns=20&max_idle_conns=4" --datastore-engine=cockroachdb`
- Try bootstrapping 
  `spicedb serve --datastore-conn-uri="postgresql://root@localhost:26257/defaultdb?sslmode=disable&max_conns=20&max_idle_conns=4" --grpc-preshared-key "love" --datastore-engine=cockroachdb --datastore-bootstrap-files="./pkg/validationfile/testdata/initial_schema_and_rels.yaml”`

peterfoldes:
Closes #1005 

I have very little knowledge of Go, but I gave the fix a shot anyways. Feel free to advice me on how to do this better. Also tried adding some tests around `loader_test.go` to use the internal testservers to try out the different engines, but I failed to figure that out in the few hours I wanted to spend on this.

Only tested the solution for `crdb` following the same steps I described in the issue, with this the bootstrapping works neatly.

### DIFF
--- a/pkg/datastore/test/caveat.go
+++ b/pkg/datastore/test/caveat.go
@@ -29,8 +29,12 @@ func WriteReadDeleteCaveatTest(t *testing.T, tester DatastoreTester) {
 
 	skipIfNotCaveatStorer(t, ds)
 
-	// Dupes in same transaction are fail to be written
 	ctx := context.Background()
+	// Don't fail on writing empty caveat list
+	_, err = writeCaveats(ctx, ds)
+	req.NoError(err)
+
+	// Dupes in same transaction fail to be written
 	coreCaveat := createCoreCaveat(t)
 	coreCaveat.Name = "a"
 	_, err = writeCaveats(ctx, ds, coreCaveat, coreCaveat)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,6 +83,12 @@ In order to build and test the project, the [latest stable version of Go] and kn
 go test -v ./...
 ```
 
+To run integration tests (for example when testing datastores):
+
+```sh
+go test -v --failfast --tags 'ci,docker' -count=1 ./...
+```
+
 ### Adding dependencies
 
 This project does not use anything other than the standard [Go modules] toolchain for managing dependencies.
--- a/internal/datastore/crdb/caveat.go
+++ b/internal/datastore/crdb/caveat.go
@@ -109,6 +109,9 @@ func (cr *crdbReader) ListCaveats(ctx context.Context, caveatNames ...string) ([
 }
 
 func (rwt *crdbReadWriteTXN) WriteCaveats(ctx context.Context, caveats []*core.CaveatDefinition) error {
+	if len(caveats) == 0 {
+		return nil
+	}
 	write := writeCaveat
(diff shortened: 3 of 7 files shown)


peterfoldes:

recheck

peterfoldes:
recheck

josephschorr:
Hi @peterfoldes,

Thanks for the contribution! If you'd like to add a test, take a look at https://github.com/authzed/spicedb/blob/main/pkg/datastore/test/datastore.go#L51, which allows for adding tests that run against *all* datastores. Should be fairly simple to add a test that just tries to write an empty set of caveats.

pkg/datastore/test/caveat.go
```diff
@@ -29,8 +29,12 @@ func WriteReadDeleteCaveatTest(t *testing.T, tester DatastoreTester) {
 
 	skipIfNotCaveatStorer(t, ds)
 
-	// Dupes in same transaction are fail to be written
 	ctx := context.Background()
+	// Don't fail on writing empty caveat list
+	_, err = writeCaveatsArray(ctx, ds, []*core.CaveatDefinition{})
```
vroldanbet, line 34: I don't think we need this extra method, you can call `writeCaveats(ctx, ds)` without arguments. `nil` slice is an empty slice too.

peterfoldes: Yeah, I was just testing that out. Took some time getting used to go syntax, thanks for the help.

Also, let me add a note in contributors.md that to run integration test people can do something like
`go test -v --failfast --tags 'ci,docker' -count=1 ./...`. It took me a while to figure that out (and github actions for the W)

vroldanbet: Sounds good! I noted you removed the len check in spanner, is there any reason spanner does not need it?

peterfoldes: Not quite sure, but the test didn't fail there. Probably `BufferWrite` works differently on empty mutations

vroldanbet (approved):

PR #1008, opened 2022-11-21
Catch nil values for FoundSubjectsByResourceID map and return as errors

josephschorr:
Also ensure we never add nils in the first place

### DIFF
--- a/internal/datasets/subjectsetbyresourceid.go
+++ b/internal/datasets/subjectsetbyresourceid.go
@@ -1,6 +1,8 @@
 package datasets
 
 import (
+	"fmt"
+
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	v1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
 )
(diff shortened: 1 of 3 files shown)


internal/graph/lookupsubjects.go
```diff
@@ -460,7 +464,10 @@ func (lsu *lookupSubjectsUnion) CompletedChildOperations() error {
 
 		for _, result := range collector.Results() {
 			metadata = combineResponseMetadata(metadata, result.Metadata)
-			foundSubjects.UnionWith(result.FoundSubjectsByResourceId)
+			err := foundSubjects.UnionWith(result.FoundSubjectsByResourceId)
```
jakedt, line 467: You should use the `if err := thing(); err != nil` pattern here and at other callsites.

internal/graph/lookupsubjects.go
```diff
@@ -460,7 +464,10 @@ func (lsu *lookupSubjectsUnion) CompletedChildOperations() error {
...
+			if err != nil {
+				return err
```
jakedt, line 469: Wrap these errors with higher level context about _why_ you were trying to Union, so we can trace which callsite caused the problem.

internal/datasets/subjectsetbyresourceid.go
```diff
@@ -36,12 +42,22 @@ func (ssr SubjectSetByResourceID) AddFromRelationship(relationship *core.Relatio
 }
 
 // UnionWith unions the map's sets with the other map of sets provided.
-func (ssr SubjectSetByResourceID) UnionWith(other map[string]*v1.FoundSubjects) {
+func (ssr SubjectSetByResourceID) UnionWith(other map[string]*v1.FoundSubjects) error {
```
jakedt, line 45: Can you create a test which reproduces why this happens in real running-systems to verify the fix?

jakedt (changes requested):

josephschorr:
Updated

internal/datasets/subjectsetbyresourceid.go
```diff
@@ -18,30 +20,44 @@ type SubjectSetByResourceID struct {
...
 	for resourceID, subjects := range other {
+		if subjects == nil {
+			return fmt.Errorf("received nil FoundSubjects in other map of SubjectSetByResourceID's UnionWith for key %s", resourceID)
+		}
+
 		for _, subject := range subjects.FoundSubjects {
-			ssr.add(resourceID, subject)
+			err := ssr.add(resourceID, subject)
```
jakedt, line 53: `if err := ...; err != nil`

jakedt (approved):

PR #1009, opened 2022-11-21
Return a more descriptive error for watch when not enabled

josephschorr:
Requires https://github.com/authzed/docs/pull/80

### DIFF
--- a/internal/datastore/crdb/watch.go
+++ b/internal/datastore/crdb/watch.go
@@ -28,6 +28,17 @@ func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Rev
 	updates := make(chan *datastore.RevisionChanges, cds.watchBufferLength)
 	errs := make(chan error, 1)
 
+	features, err := cds.Features(ctx)
+	if err != nil {
+		errs <- err
+		return updates, errs
+	}
+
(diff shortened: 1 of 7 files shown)


internal/datastore/crdb/watch.go
```diff
@@ -28,6 +28,17 @@ func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Rev
...
 	errs := make(chan error, 1)
 
+	features, err := cds.Features(ctx)
+	if err != nil {
+		errs <- err
+		return updates, errs
+	}
+
+	if !features.Watch.Enabled {
+		errs <- datastore.NewWatchUnsupportedErr(fmt.Sprintf("%s. See https://docs.authzed.com/spicedb/enabling-watch-api#crdb", features.Watch.Reason))
```
jzelinskie, line 38: We might change this link

josephschorr: I thought of that; I do wonder if we should add some sort of short-link here instead, in case we need to change to where it points

josephschorr:
Updated

vroldanbet (approved):
LGTM

jzelinskie (approved):

PR #1010, opened 2022-11-23
Add configurable concurrency limits per dispatch type

josephschorr:
This allows for overrides on the concurrency limits for specific types of dispatch, as a basic form of QoS for different kinds of API calls

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -24,6 +24,7 @@ import (
 	"github.com/authzed/spicedb/internal/dispatch"
 	clusterdispatch "github.com/authzed/spicedb/internal/dispatch/cluster"
 	combineddispatch "github.com/authzed/spicedb/internal/dispatch/combined"
+	"github.com/authzed/spicedb/internal/dispatch/graph"
 	"github.com/authzed/spicedb/internal/gateway"
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/internal/services"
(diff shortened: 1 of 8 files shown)


pkg/cmd/server/server.go
```diff
@@ -64,14 +65,15 @@ type Config struct {
...
-	DispatchMaxDepth             uint32
-	DispatchConcurrencyLimit     uint16
-	DispatchUpstreamAddr         string
-	DispatchUpstreamCAPath       string
-	DispatchClientMetricsPrefix  string
-	DispatchClusterMetricsPrefix string
-	Dispatcher                   dispatch.Dispatcher
+	DispatchServer                  util.GRPCServerConfig
+	DispatchMaxDepth                uint32
+	OverallDispatchConcurrencyLimit uint16
```
vroldanbet, line 70: nit: `GlobalDispatchConcurrencyLimit`

pkg/cmd/server/server.go
```diff
@@ -159,6 +161,10 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 			dispatchPresharedKey = c.PresharedKey[0]
 		}
 
+		specificConcurrencyLimits := c.DispatchConcurrencyLimits
+		concurrencyLimits := specificConcurrencyLimits.WithOverallDefaultLimit(c.OverallDispatchConcurrencyLimit)
+		log.Debug().EmbedObject(concurrencyLimits).Msg("configured dispatch concurrency limits")
```
vroldanbet, line 166: hrm, I'm wondering if this is better suited for info level. It only happens once on bootstrap, and it's useful information to validate you are getting the expected values in production without bumping to debug level

josephschorr: Yeah, I went back and forth on this. @jzelinskie thoughts?

pkg/cmd/server/server.go
josephschorr: K

josephschorr:
Rebased

pkg/cmd/server/server.go
jzelinskie: I believe the others are printed as info, too. Unless we don't want folks to tinker with this flag, I'd say info is fine.

josephschorr:
Rebased and updated

vroldanbet (approved):

Issue #1011, opened 2022-11-23
go install github.com/authzed/spicedb fails


vroldanbet:
reported in discord by @ensonic

```
go install github.com/authzed/spicedb@v1.15.0
go: github.com/authzed/spicedb@v1.15.0 (in github.com/authzed/spicedb@v1.15.0):
    The go.mod file for the module providing named packages contains one or
    more replace directives. It must not contain directives that would cause
    it to be interpreted differently than if it were the main module.
```

jzelinskie:
We should be able to drop our replace directive now that the outcaste.io ristretto repository has merged our changes.

Outcaste is all of the ex-dgraph folks continuing to maintain ristretto as a community project.

PR #1012, opened 2022-11-23
move off ristretto fork

vroldanbet:
closes https://github.com/authzed/spicedb/issues/1011

Our submission got merged so we can move back to using upstream repo. This also fixes the inability to run `go install` due to the replacement directive.


josephschorr (approved):
LGTM

PR #1016, opened 2022-11-26
Early iterator closing in dispatch

josephschorr:
Add a lint check and fix usages of iterators to ensure they are closed as early as we can

### DIFF
--- a/tools/analyzers/closeafterusagecheck/closeafterusagecheck_test.go
+++ b/tools/analyzers/closeafterusagecheck/closeafterusagecheck_test.go
@@ -1,0 +1,15 @@
+package closeafterusagecheck
+
+import (
+	"testing"
+
+	"golang.org/x/tools/go/analysis/analysistest"
+)
+
+func TestAnalyzer(t *testing.T) {
+	analyzer := Analyzer()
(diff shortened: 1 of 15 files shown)


vroldanbet (changes requested):
Why is the analyzers binary being checked in?

josephschorr:
> Why is the analyzers binary being checked in?

Must have accidentally added it. Removed

tools/analyzers/closeafterusagecheck/closeafterusagecheck_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"testing"
+
+	"golang.org/x/tools/go/analysis/analysistest"
+)
+
+func TestAnalyzer(t *testing.T) {
+	analyzer := Analyzer()
+	analyzer.Flags.Set("must-be-closed-after-usage-types", "*notimmediatelyclosed.SomeIterator")
```
vroldanbet, line 11: unhandled error

vroldanbet (commented):
👋🏻 I fail to see how this help with DB connection connection, maybe I'm missing something. The only implementation of a relationship iterator used by production code is `sliceRelationshipIterator`, which does not hold any references to database connections or any state that would cause DB conns to be released.

vroldanbet (approved):
👍🏻 looks good to me

PR #1020, opened 2022-11-28
add caveat covering index

vroldanbet:
# What

tries to optimize relationship lookup for various datastores by making sure caveat name and context are part of a covering index. 

# How

this PR explores making sure every relationship query is served from the index (aka covering index). This was the case before, but it stopped the moment we added caveats to the table.

This functionality is nicely handled in all databases, except MySQL, which does not support "including a column in the covering index".

In MySQL case, we also had to shrink the `caveat_name` column because adding any new element to the index hit the 3072 bytes InnoDB limit. The migration won't make this a covering index, but:
- it will allow us to have an index for `ReadRelationships` with a `RelationshipFilter` that filters on caveat name
- it will allow us to select all columns but `column_context` and make decisions based on the existence of a caveat in the row.

### DIFF
--- a/internal/datastore/crdb/migrations/zz_migration.0005_add_caveat_covering_index.go
+++ b/internal/datastore/crdb/migrations/zz_migration.0005_add_caveat_covering_index.go
@@ -1,0 +1,29 @@
+package migrations
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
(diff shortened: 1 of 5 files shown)


vroldanbet:
superseded by https://github.com/authzed/spicedb/pull/1050

PR #1029, opened 2022-11-30
Fix revision checking in memdb to allow for past `now`

josephschorr:
We need to allow for revision later than `now` due to our recent fix for time granularity on macos

### DIFF
--- a/internal/datastore/memdb/revisions.go
+++ b/internal/datastore/memdb/revisions.go
@@ -38,7 +38,10 @@ func (mdb *memdbDatastore) newRevisionID() revision.Decimal {
 }
 
 func (mdb *memdbDatastore) HeadRevision(ctx context.Context) (datastore.Revision, error) {
-	head, err := mdb.headRevision(ctx)
+	mdb.RLock()
+	defer mdb.RUnlock()
+
+	head, err := mdb.headRevisionNoLock()
 	if err != nil {
 		return nil, err
 	}
@@ -46,10 +49,7 @@ func (mdb *memdbDatastore) HeadRevision(ctx context.Context) (datastore.Revision
 	return revision.NewFromDecimal(head), nil
 }
 
-func (mdb *memdbDatastore) headRevision(ctx context.Context) (decimal.Decimal, error) {
-	mdb.Lock()
-	defer mdb.Unlock()
-
+func (mdb *memdbDatastore) headRevisionNoLock() (decimal.Decimal, error) {
 	return mdb.revisions[len(mdb.revisions)-1].revision, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/datastore/memdb/revisions.go
```diff
@@ -70,6 +70,15 @@ func (mdb *memdbDatastore) checkRevisionLocal(revisionRaw revision.Decimal) erro
 	now := revisionFromTimestamp(time.Now().UTC())
 
 	if revisionRaw.GreaterThan(now) {
+		headRevision, err := mdb.headRevision()
```
jakedt, line 73: Hmm, I don't really like this approach. Wouldn't a better approach be to check that it's "< 1 big part more" than what was already passed back? This seems like it will generate a +2 snapshot when we don't need it...

jakedt (changes requested):

internal/datastore/memdb/revisions.go
josephschorr: Not sure I follow... its not generating a new revision, simply check that it is less than or equal to the current head revision

jakedt: Why not just always use the `mdb.HeadRevision` here?

josephschorr: It was returning a slightly different data type, if I recall correctly

josephschorr: Or do you mean why not just use `headRevision` always?

jakedt: Yeah I mean in this code block, if you're calling out to headRevision if it's > now, why not just always use the headrevision path.

josephschorr: I was trying to avoid hitting the lock condition unless absolutely necessary; this way, we only hit the lock on the check revision when necessary

josephschorr:
@jakedt Updated

internal/datastore/memdb/revisions.go
```diff
@@ -59,24 +59,42 @@ func (mdb *memdbDatastore) OptimizedRevision(ctx context.Context) (datastore.Rev
...
 	dr, ok := revisionRaw.(revision.Decimal)
 	if !ok {
 		return datastore.NewInvalidRevisionErr(revisionRaw, datastore.CouldNotDetermineRevision)
 	}
-	return mdb.checkRevisionLocal(dr)
+	return mdb.checkRevisionLocalNoLock(dr)
 }
 
-func (mdb *memdbDatastore) checkRevisionLocal(revisionRaw revision.Decimal) error {
+func (mdb *memdbDatastore) checkRevisionLocalNoLock(revisionRaw revision.Decimal) error {
```
jakedt, line 72: Maybe change this to checkRevisionLocalCallerLocks or something that tells the caller that the invariant must hold.

josephschorr: Changed to `checkRevisionLocalCallerMustLock` to be as explicit as possible

PR #1030, opened 2022-12-01
Adjust estimated query count metric to only count dispatch if it was necessary

josephschorr:
This is a more accurate estimate

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -214,12 +214,9 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 	hadDirectResult := false
 	hadDispatchedResult := false
 	defer (func() {
-		estimatedQueryCount := 0.0
-		if hadDirectResult {
-			estimatedQueryCount++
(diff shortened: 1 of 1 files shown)


internal/graph/check.go
```diff
@@ -214,12 +214,9 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
...
-			estimatedQueryCount++
-		}
-		if hadDispatchedResult {
-			estimatedQueryCount++
+		estimatedQueryCount := 1.0
+		if !hadDirectResult && hadDispatchedResult {
+			estimatedQueryCount = 2
 		}
 		estimatedDirectDispatchQueryHistogram.Observe(estimatedQueryCount)
 	})()
```
vroldanbet, line 222: do we need the wrapping parentheses?

josephschorr: Maybe not, but I always prefer it when using function definitions under defers

vroldanbet (approved):

PR #1041, opened 2022-12-06
Debug supporting caveats

josephschorr:
Updates the debug trace system to support caveats in the returned traces


internal/services/v1/permissions_test.go
```diff
@@ -1273,3 +1273,164 @@ func randString(length int) string {
 	}
 	return string(b)
 }
+
+func TestCheckPermissionWithCaveatedDebugInfoHasPermission(t *testing.T) {
```
vroldanbet, line 1277: I'd suggest moving this instead into a new `debug_test.go`

internal/services/v1/debug.go
```diff
@@ -0,0 +1,188 @@
...
+		subRelation := ct.Request.Subject.Relation
+		if subRelation == tuple.Ellipsis {
+			subRelation = ""
+		}
+
+		permissionship := v1.CheckDebugTrace_PERMISSIONSHIP_NO_PERMISSION
+		checkResult, ok := ct.Results[resourceID]
+		if ok {
+			if checkResult.Membership == dispatch.ResourceCheckResult_MEMBER {
+				permissionship = v1.CheckDebugTrace_PERMISSIONSHIP_HAS_PERMISSION
```
vroldanbet, line 80: The test does not exercise this codepath. I imagine a situation where, even though caveats were part of the graph, they where short-circuited. In that case `checkResult.Expression` would be `nil`. Perhaps could use a test.

internal/services/v1/debug.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	schema, _ := generator.GenerateSchema(defs)
+	converted, err := convertCheckTrace(ctx, caveatContext, debugInfo.Check, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	return &v1.DebugInformation{
+		Check:      converted[0],
```
vroldanbet, line 56: I assume `LookupResources` does not have debug support yet. In the meantime, we can assume `dispatch.CheckDebugTrace` is going to have only information for one checked resource, and then change `convertCheckTrace` to return one single  `v1.CheckDebugTrace` to make clear these are only used for `Check`-initiated dispatches.

Or we can make it more explicit in the method name `convertDispatchDebugInformation` that there is an expectation the dispatch debug information comes from a Check-initiated dispatch and thus has one single resource.

internal/services/v1/debug.go
```diff
@@ -0,0 +1,188 @@
...
+			var partialCaveatInfo *v1.PartialCaveatInfo
+			caveatResult := v1.CaveatEvalInfo_RESULT_FALSE
+			if computedResult.Value() {
+				caveatResult = v1.CaveatEvalInfo_RESULT_TRUE
+			} else if computedResult.IsPartial() {
+				caveatResult = v1.CaveatEvalInfo_RESULT_MISSING_SOME_CONTEXT
+				missingNames, _ := computedResult.MissingVarNames()
+				partialCaveatInfo = &v1.PartialCaveatInfo{
+					MissingRequiredContext: missingNames,
+				}
```
vroldanbet, line 103: untested

internal/services/v1/permissions_test.go
```diff
@@ -339,7 +339,7 @@ func TestCheckPermissionWithDebugInfo(t *testing.T) {
 		SchemaString: debugInfo.SchemaUsed,
 	}, &emptyDefaultPrefix)
 	require.NoError(err, "Invalid schema: %s", debugInfo.SchemaUsed)
-	require.Equal(3, len(compiled.OrderedDefinitions))
+	require.Equal(4, len(compiled.OrderedDefinitions))
```
vroldanbet, line 342: 👍🏻because the new implementation also takes caveats into consideration

internal/services/v1/permissions_test.go
josephschorr: Why? its part of the V1 API

internal/services/v1/debug.go
josephschorr: Which part isn't exercised?

internal/services/v1/debug.go
josephschorr: Renamed to `convertCheckDispatchDebugInformation`

internal/services/v1/debug.go
josephschorr: Added

internal/services/v1/permissions_test.go
josephschorr: Yep

josephschorr:
Rebased and updated

internal/services/v1/permissions_test.go
vroldanbet: yeah, I get it, but it's exercising code that lives in `v1/debug.go`

internal/services/v1/debug.go
vroldanbet: `CheckDebugTrace_PERMISSIONSHIP_HAS_PERMISSION`

vroldanbet (dismissed):

vroldanbet (approved):

PR #1042, opened 2022-12-06
datastore/crdb: upgrade to v22.2.0 to get arm support

jakedt:


internal/testserver/datastore/crdb.go
```diff
@@ -19,7 +19,11 @@ import (
 	"github.com/authzed/spicedb/pkg/secrets"
 )
 
-const enableRangefeeds = `SET CLUSTER SETTING kv.rangefeed.enabled = true;`
+const (
+	CRDBTestVersionTag = "v22.2.0"
```
josephschorr, line 23: Does this mean we'll lose support for v22.1 for watch then?

jakedt: Ugh, good catch. We'll have to add the `min_checkpoint_frequency` flag conditionally based on cockroach version.

`ERROR: invalid option "min_checkpoint_frequency" (SQLSTATE XXUUU).`

jakedt: Well that's going to have to wait.

jakedt: Updated

internal/datastore/crdb/crdb_test.go
```diff
@@ -166,7 +167,7 @@ func TestWatchFeatureDetection(t *testing.T) {
 	}
 }
 
-func newCRDBWithUser(t *testing.T, pool *dockertest.Pool) (adminConn *pgx.Conn, nonAdminConnURI string) {
+func newCRDBWithUser(t *testing.T, pool *dockertest.Pool) (adminConn *pgx.Conn, connURIFunc func(username, password string) string) {
```
josephschorr, line 170: I wonder if we should just return a map from username -> connection string, since otherwise the *caller* needs to know the username+password pair defined inside here, which seems an inversion

internal/datastore/crdb/crdb.go
```diff
@@ -82,11 +82,25 @@ func newCRDBDatastore(url string, options ...Option) (datastore.Datastore, error
...
+	initCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	pool, err := pgxpool.ConnectConfig(initCtx, poolConfig)
 	if err != nil {
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
 
+	var version crdbVersion
+	if err := queryServerVersion(initCtx, pool, &version); err != nil {
```
josephschorr, line 94: any reason we do the "pass-in-pointer" of this vs just returning a pointer of the version struct?

jakedt:
Updated

internal/datastore/crdb/crdb.go
josephschorr: Makes sense. I just find it a bit riskier since you could accidentally pass in `nil`, but since it is internal to CRDB package, its fine

josephschorr (approved):
LGTM

internal/datastore/crdb/crdb.go
```diff
@@ -82,11 +82,25 @@ func newCRDBDatastore(url string, options ...Option) (datastore.Datastore, error
 
 	configurePool(config, poolConfig)
 
-	pool, err := pgxpool.ConnectConfig(context.Background(), poolConfig)
+	initCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
vroldanbet, line 85: It seems this addresses https://github.com/authzed/spicedb/issues/741

PR #1048, opened 2022-12-09
Add prom metric for number of batch check dispatches

### LINKED ISSUE #851: add histogram metrics to batch check sizes
Batch checks were [recently introduced in spicedb](https://github.com/authzed/spicedb/pull/777). It would be useful to have prometheus histogram metrics around the sizes of the batch in order to help operators how to better optimize their schemas or spicedb code itself

josephschorr:
Fixes #851

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -27,8 +27,15 @@ var estimatedDirectDispatchQueryHistogram = prometheus.NewHistogram(prometheus.H
 	Buckets: []float64{1, 2},
 })
 
+var dispatchChunkCountHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
+	Name:    "spicedb_check_dispatch_chunk_count",
+	Help:    "number of chunks when dispatching in check",
+	Buckets: []float64{1, 2, 3, 5, 10, 25, 100, 250},
+})
+
 func init() {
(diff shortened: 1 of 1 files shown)


internal/graph/check.go
```diff
@@ -250,12 +257,15 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 	// Convert the subjects into batched requests.
 	toDispatch := make([]directDispatch, 0, subjectsToDispatch.Len())
 	subjectsToDispatch.ForEachType(func(rr *core.RelationReference, resourceIds []string) {
+		chunkCount := 0.0
```
vroldanbet, line 260: would it perhaps be easier to analyze the graphs if instead of "observing chunks" we observe the actual length of the chunk, so we don't have to do the mental math based on the default chunk size (which may as well become configurable)? Also with the current histogram bucket granularity, chances are everything falls within 1 or 2, which wouldn't offer a lot of information. If we decide to move to actual chunk length, perhaps we could use a different bucket size progression like `1, 5, 10, 20, 40, 80, 160, 320, 640, 1280...`

vroldanbet: after chatting with Joey, I've reconsidered my stand here, since it's more useful to see "how much we are dispatching" versus the size of the dispatched requests

vroldanbet (approved):

PR #1049, opened 2022-12-09
Move to golang 1.19.4

josephschorr:


.github/workflows/lint.yaml
```diff
@@ -1,14 +1,14 @@
 ---
 name: "Lint"
-on:  # yamllint disable-line rule:truthy
+on: # yamllint disable-line rule:truthy
```
jzelinskie, line 3: I think this breaks a lint rule

josephschorr: Darn autoformatting

vroldanbet (approved):

PR #1050, opened 2022-12-12
serve relationship queries from covering index in postgres and test it

vroldanbet:
introduces a test that demonstrates queries planner finds a covering index, and does not need to read from the main data table.

This PR reconciles two branches:
- @josephschorr's https://github.com/authzed/spicedb/commit/0a290aa26325e319f0768aec1ab5a5b14394976c
- https://github.com/authzed/spicedb/pull/1020

## TODO
- [x] test relationship read with caveat filter
- [x] assert indexes used in query
- [ ] identify other queries in `Check` critical path

### DIFF
--- a/internal/datastore/postgres/postgres_test.go
+++ b/internal/datastore/postgres/postgres_test.go
@@ -117,6 +117,10 @@ func TestPostgresDatastore(t *testing.T) {
 				WatchNotEnabledTest(t, b)
 			})
 
+			t.Run("IndexOnlyRead", func(t *testing.T) {
+				IndexOnlyReadTest(t, b)
+			})
+
 			if config.migrationPhase == "" {
 				t.Run("RevisionInversion", createDatastoreTest(
 					b,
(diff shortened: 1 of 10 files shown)


internal/datastore/postgres/postgres_test.go
```diff
@@ -739,3 +743,98 @@ func BenchmarkPostgresQuery(b *testing.B) {
...
+
+	defer iter.Close()
+
+	for tpl := iter.Next(); tpl != nil; tpl = iter.Next() {
+		require.Equal(testfixtures.DocumentNS.Name, tpl.ResourceAndRelation.Namespace)
+	}
+	require.NoError(iter.Err())
+
+	// Ensure that Index Only scans were used.
+	for stmt, explaination := range logger.explanations {
```
josephschorr, line 835: we likely need another test for the GC query to make sure it is also using the covering index

vroldanbet:
related https://github.com/authzed/spicedb/issues/1090

josephschorr:
Replaced by https://github.com/authzed/spicedb/pull/1118

Issue #21, opened 2021-08-19
Go panics should return a gRPC error


jzelinskie:
The [gRPC recovery middleware] can be added catch any panics and convert them into an gRPC internal error.

This requires a careful review to ensure that no resources are leaked as a result of continuing the program after a panic occurs.

[gRPC recovery middleware]: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/recovery

PR #1054, opened 2022-12-14
Add a linter for improper use of panics and fix all found instances

josephschorr:
A `panic` is now considered "improper" unless under the following conditions:

1) In a skipped file (tests, migrations, etc)
2) Is under a function starting with the prefix `must` or `Must`
3) Is under a `default` branch of a `switch` statement that is, itself, under a function that does not return an `error`

*All* other uses are considered improper and flagged by the linter


vroldanbet (commented):
something I've seen done in a bunch of go server frameworks is to create a middleware that rescues panics, with the goals of returning a 5XX to the client, and properly logging the error.
e.g. [gorilla handler](https://github.com/gorilla/handlers/blob/master/recovery.go). Not saying the process shouldn't terminate - it should - but at least we return to the user something, and make our lives easier to troubleshoot what happened.

josephschorr:
> something I've seen done in a bunch of go server frameworks is to create a middleware that rescues panics, with the goals of returning a 5XX to the client, and properly logging the error. e.g. [gorilla handler](https://github.com/gorilla/handlers/blob/master/recovery.go). Not saying the process shouldn't terminate - it should - but at least we return to the user something, and make our lives easier to troubleshoot what happened.

There is one: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/recovery

But as noted in the issue (see https://github.com/authzed/spicedb/issues/21), there are concerns about automatic recovery leaving the system in an unexpected state. So, rather than doing so, this PR ensures we minimize panics wherever possible in favor of proper errors, so we *know* the state of the world going out

pkg/cmd/serve.go
```diff
@@ -37,7 +37,7 @@ var (
 	}
 )
 
-func RegisterServeFlags(cmd *cobra.Command, config *server.Config) {
+func MustRegisterServeFlags(cmd *cobra.Command, config *server.Config) {
```
vroldanbet, line 40: I'd suggest avoiding panics in the CLI command:
- `main.go` is handling any error returned by cobra, we can customize error codes returned by the process
- panics do not unwind the stack and do not do any cleaning up, nor goroutines are terminated gracefully
- becomes difficult to log things in the CLI tool and make the output look pretty

pkg/cmd/server/server.go
```diff
@@ -171,7 +171,7 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 			log.Trace().Int(fmt.Sprintf("preshared-key-%d-length", index+1), len(presharedKey)).Msg("preshared key configured")
 		}
 
-		c.GRPCAuthFunc = auth.RequirePresharedKey(c.PresharedKey)
+		c.GRPCAuthFunc = auth.MustRequirePresharedKey(c.PresharedKey)
```
vroldanbet, line 174: same here: `Complete` has an error in the signature, so let's return an error

pkg/graph/walker.go
```diff
@@ -26,97 +25,43 @@ func WalkRewrite(rewrite *core.UsersetRewrite, handler WalkHandler) interface{}
...
-		}
-	}
-
-	return nil
-}
-
-type OperationPath = []uint32
-
-// FindOperation finds the operation at the given path in the rewrite, if any.
-func FindOperation[T any](rewrite *core.UsersetRewrite, opPath OperationPath) *T {
```
vroldanbet, line 69: dead code that was recursively calling each other?

pkg/schemadsl/generator/generator.go
```diff
@@ -91,7 +105,7 @@ func (sg *sourceGenerator) emitCaveat(caveat *core.CaveatDefinition) {
 
 		decoded, err := caveattypes.DecodeParameterType(caveat.ParameterTypes[paramName])
 		if err != nil {
-			panic("invalid parameter type on caveat")
+			return fmt.Errorf("invalid parameter type on caveat: %w", err)
```
vroldanbet, line 108: the assumption is we don'g use `MustBug` here because this is an input error

pkg/spiceerrors/bug.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Based on: https://stackoverflow.com/a/58945030
+func isInTests() bool {
+	for _, arg := range os.Args {
+		if strings.HasPrefix(arg, "-test.v=") {
+			return true
+		}
+	}
+	return false
+}
```
vroldanbet, line 17: Let's add a test `bug_test.go` file that checks that `isInTests` is returning true while in test, in case there is any regression with newer go versions

vroldanbet (commented):
Changes look good! I felt some questions

internal/services/v0/sharestore.go
```diff
@@ -268,6 +268,6 @@ func unmarshalShared(data []byte) (SharedDataV2, LookupStatus, error) {
 		}, LookupConverted, nil
 
 	default:
-		panic("Unsupported version")
+		return SharedDataV2{}, LookupError, fmt.Errorf("unsupported share version %s", v.Version)
```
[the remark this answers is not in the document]
josephschorr: Its user facing: if a user gives us an invalid version, it should be a normal error, not a bug. This should never have been a panic

pkg/graph/walker.go
josephschorr: Yes, I meant to use it for reachability analysis but then it got un-used and I never removed it. I figured I'd just remove it here since it isn't used aynmore

internal/namespace/typesystem.go
```diff
@@ -527,10 +532,6 @@ func SourceForAllowedRelation(allowedRelation *core.AllowedRelation) string {
 		return fmt.Sprintf("%s:*%s", allowedRelation.GetNamespace(), caveatStr)
 	}
 
-	if allowedRelation.GetRelation() == "" {
-		panic("invalid allowed relation: relation is empty for a non-wildcard")
-	}
-
```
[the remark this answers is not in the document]
josephschorr: Because relations cannot be empty on the core types and we check that elsewhere

internal/services/v1/schema.go
```diff
@@ -68,7 +68,10 @@ func (ss *schemaServer) ReadSchema(ctx context.Context, in *v1.ReadSchemaRequest
 		schemaDefinitions = append(schemaDefinitions, nsDef)
 	}
 
-	schemaText, _ := generator.GenerateSchema(schemaDefinitions)
+	schemaText, _, err := generator.GenerateSchema(schemaDefinitions)
```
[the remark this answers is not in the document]
josephschorr: The errors being returned are `spiceerrors` which means they should 500

josephschorr:
Updated and rebased

pkg/schemadsl/generator/generator.go
josephschorr: Yep

vroldanbet (dismissed):

pkg/cmd/datastore/datastore.go
```diff
@@ -91,8 +91,8 @@ type Config struct {
 	MigrationPhase string
 }
 
-// RegisterDatastoreFlags adds datastore flags to a cobra command
-func RegisterDatastoreFlags(cmd *cobra.Command, opts *Config) {
+// MustRegisterDatastoreFlags adds datastore flags to a cobra command
```
vroldanbet, line 94:
```suggestion
// RegisterDatastoreFlags adds datastore flags to a cobra command
```

vroldanbet (approved):

PR #1057, opened 2022-12-15
redesigns middlware options for RunnableServer

vroldanbet:
the methods to change middleware in `RunnableServer` were introduced because we needed a mechanism to be able to retain the default middleware and expand it.

This new design is explicit about it, by providing different operations to change the middleware layer: prepend, replace and append.

This allows us to remove the methods from `RunnableServer` and configure the final middleware chain through the `server.Config`, instead of having to do it in 2 different steps

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -90,8 +90,12 @@ type Config struct {
 	MetricsAPI   util.HTTPServerConfig
 
 	// Middleware for grpc
-	UnaryMiddleware     []grpc.UnaryServerInterceptor
-	StreamingMiddleware []grpc.StreamServerInterceptor
+	PrependUnaryMiddleware            []grpc.UnaryServerInterceptor
+	PrependStreamingMiddleware        []grpc.StreamServerInterceptor
+	ReplaceDefaultUnaryMiddleware     []grpc.UnaryServerInterceptor
+	ReplaceDefaultStreamingMiddleware []grpc.StreamServerInterceptor
+	AppendUnaryMiddleware             []grpc.UnaryServerInterceptor
+	AppendStreamingMiddleware         []grpc.StreamServerInterceptor
 
 	// Middleware for dispatch
 	DispatchUnaryMiddleware     []grpc.UnaryServerInterceptor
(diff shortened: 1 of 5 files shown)


pkg/cmd/server/server.go
```diff
@@ -345,6 +347,34 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 	}, nil
 }
 
+func (c *Config) buildMiddleware(dispatcher dispatch.Dispatcher, ds datastore.Datastore) ([]grpc.UnaryServerInterceptor, []grpc.StreamServerInterceptor) {
+	finalUnaryMiddleware, finalStreamingMiddleware := DefaultMiddleware(log.Logger, c.GRPCAuthFunc, !c.DisableVersionResponse, dispatcher, ds)
+	if len(c.ReplaceUnaryMiddleware) != 0 {
+		finalUnaryMiddleware = c.ReplaceUnaryMiddleware
+	}
+	if len(c.ReplaceStreamingMiddleware) != 0 {
```
josephschorr, line 355: So Replace goes first... but it doesn't actually replace any of the middleware besides the default?

pkg/cmd/server/server.go
```diff
@@ -345,6 +347,34 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 	}, nil
 }
 
+func (c *Config) buildMiddleware(dispatcher dispatch.Dispatcher, ds datastore.Datastore) ([]grpc.UnaryServerInterceptor, []grpc.StreamServerInterceptor) {
+	finalUnaryMiddleware, finalStreamingMiddleware := DefaultMiddleware(log.Logger, c.GRPCAuthFunc, !c.DisableVersionResponse, dispatcher, ds)
```
josephschorr, line 351: Calling this `final` seems odd since we're appending to it below

pkg/cmd/server/server.go
vroldanbet: Correct, it's a replacement for the defaults. Maybe I should make that more explicit in the name, like `ReplaceDefaultUnaryMiddleware` and `ReplaceDefaultStreamingMiddleware?

pkg/cmd/server/server.go
vroldanbet: I've renamed them to `output...`

vroldanbet:
@josephschorr I've pushed tests and addressed your comments, let me know what you think

vroldanbet:
Opened https://github.com/ecordell/optgen/issues/5 to improve the ergonomics of the generated `With` methods

pkg/cmd/server/server.go
```diff
@@ -90,8 +90,12 @@ type Config struct {
 	MetricsAPI   util.HTTPServerConfig
 
 	// Middleware for grpc
-	UnaryMiddleware     []grpc.UnaryServerInterceptor
-	StreamingMiddleware []grpc.StreamServerInterceptor
+	PrependUnaryMiddleware            []grpc.UnaryServerInterceptor
+	PrependStreamingMiddleware        []grpc.StreamServerInterceptor
+	ReplaceDefaultUnaryMiddleware     []grpc.UnaryServerInterceptor
```
josephschorr, line 95: I'm still not sold on the naming here and we *definitely* need comments on the ordering that applies

vroldanbet: I've attempted to come up with another design in https://github.com/authzed/spicedb/pull/1063

vroldanbet:
closed in favor of https://github.com/authzed/spicedb/pull/1063/

PR #1058, opened 2022-12-15
gomod: update cobraotel to support sample ratios

jzelinskie:
This adds a `--otel-sample-ratio` flag that defaults to 20% of traces being sampled.
This is following the best practices for production from the OpenTelemetry documentation.


josephschorr (approved):
LGTM

PR #1061, opened 2022-12-16
Fix concurrent access issue in reachable resources and add additional testing and a small perf improvement

josephschorr:


pkg/util/multimap.go
```diff
@@ -68,3 +87,46 @@ func (mm *MultiMap[T, Q]) Len() int {
 func (mm *MultiMap[T, Q]) Keys() []T {
 	return maps.Keys(mm.items)
 }
+
+// AsReadOnly returns a read-only *copy* of the mulitmap.
+func (mm *MultiMap[T, Q]) AsReadOnly() ReadOnlyMultimap[T, Q] {
+	return readOnlyMultimap[T, Q]{
+		maps.Clone(mm.items),
+	}
+}
```
vroldanbet, line 96: The moment we clone the multi-map items, we get a copy that won't be concurrently modified, so I wondered why the need for a read only version of `MultiMap`. I figured you wanted to make it compile-time explicit that the map is not meant to be modified - even though we were already fine with the clone.

vroldanbet (approved):

pkg/util/multimap.go
josephschorr: Correct. I wanted to prevent the bug from reappearing explicitly

PR #1062, opened 2022-12-16
improvements on context cancellation

vroldanbet:
I noticed some deficiencies on context cancellation while trying to address a flake:

## [fix leaking resources when `Complete()` errors](https://github.com/authzed/spicedb/commit/0ad3bf495c79ef8f95b2b742d0d5c196eab5fa63) 

if any part of `Complete()` fails, the method returns without calling `Close()` on all the resources that could have possibly been successfully initialized. This is problematic in testing when we want to start up servers and use cancelation signals to gracefully terminate.

the approach here taken is to collect all closers during the execution of `Complete()`, and add a defer statement that calls them if an error happened. The `completedServerConfig.closeFunc` also changes to call each one of the closers.

## [address leak on cancellation on `SignalContextWithGracePeriod`](https://github.com/authzed/spicedb/commit/186ee3154834d367d4288e38da8724d0182281cb) 

in scenarios where the context is cancelled the `SignalContextWithGracePeriod` was leaking goroutines because it wasn't listening to the parent context provided as an argument.

In the case of the second `NotifyContext`, this was deliberately left as is because cancellation would cause the grace period to be ignored.

## [caching proxy did not close ristretto](https://github.com/authzed/spicedb/commit/6b81f03a3d170f4ecd61d5ffc7a3267ca783c3bf) 

usage of goleak detected the caching proxy wasn't closing the ristretto cache instance.


pkg/cmd/server/server.go
```diff
@@ -196,13 +216,16 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
...
-			return nil, fmt.Errorf("failed to configure cluster dispatch: %w", cerr)
+		cdcc, err := c.ClusterDispatchCacheConfig.Complete()
+		if err != nil {
+			return nil, fmt.Errorf("failed to configure cluster dispatch: %w", err)
 		}
 		log.Info().EmbedObject(cdcc).Msg("configured cluster dispatch cache")
+		closers = append(closers, func() error {
+			cdcc.Close()
+			return nil
+		})
```
jzelinskie, line 227: may as well just use defer here since this Close() cannot fail

pkg/cmd/server/server.go
```diff
@@ -139,6 +153,7 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 			return nil, fmt.Errorf("failed to create datastore: %w", err)
 		}
 	}
+	closers = append(closers, ds.Close)
```
jzelinskie, line 156: You'd actually want to append in the opposite order (putting ds.Close as the first element and then closers as the tail) to emulate the stack-like behavior of `defer`

pkg/cmd/server/server.go
vroldanbet: If we used defer, we would close it when `Complete` returns, which we don't want. We want to close it _if an error happens_. That's what the first defer statement that I've added does.

pkg/cmd/server/server.go
vroldanbet: It's a good point, but I think it's not relevant because these `Close` methods are idempotent. I can still add it to simulate the defer unwinding.

josephschorr (approved):
LGTM

PR #1063, opened 2022-12-20
redesigns middlware options for RunnableServer v2

vroldanbet:
the methods to change middleware in `RunnableServer` were introduced because we needed a mechanism to be able to retain the default middleware and expand it.

This new design is explicit about it, by providing different operations to change the middleware layer: prepend, replace and append.

This allows us to remove the methods from `RunnableServer` and configure the final middleware chain through the `server.Config`, instead of having to do it in 2 different steps

### DIFF
--- a/pkg/cmd/server/defaults.go
+++ b/pkg/cmd/server/defaults.go
@@ -100,34 +100,92 @@ var defaultGRPCLogOptions = []grpclog.Option{
 	}),
 }
 
-func DefaultMiddleware(logger zerolog.Logger, authFunc grpcauth.AuthFunc, enableVersionResponse bool, dispatcher dispatch.Dispatcher, ds datastore.Datastore) ([]grpc.UnaryServerInterceptor, []grpc.StreamServerInterceptor) {
-	return []grpc.UnaryServerInterceptor{
-			requestid.UnaryServerInterceptor(requestid.GenerateIfMissing(true)),
-			logmw.UnaryServerInterceptor(logmw.ExtractMetadataField("x-request-id", "requestID")),
-			grpclog.UnaryServerInterceptor(grpczerolog.InterceptorLogger(logger), defaultGRPCLogOptions...),
-			otelgrpc.UnaryServerInterceptor(),
-			grpcauth.UnaryServerInterceptor(authFunc),
-			grpcprom.UnaryServerInterceptor,
-			dispatchmw.UnaryServerInterceptor(dispatcher),
-			datastoremw.UnaryServerInterceptor(ds),
-			consistencymw.UnaryServerInterceptor(),
-			servicespecific.UnaryServerInterceptor,
-			serverversion.UnaryServerInterceptor(enableVersionResponse),
-		}, []grpc.StreamServerInterceptor{
-			requestid.StreamServerInterceptor(requestid.GenerateIfMissing(true)),
-			logmw.StreamServerInterceptor(logmw.ExtractMetadataField("x-request-id", "requestID")),
-			grpclog.StreamServerInterceptor(grpczerolog.InterceptorLogger(logger), defaultGRPCLogOptions...),
-			otelgrpc.StreamServerInterceptor(),
-			grpcauth.StreamServerInterceptor(authFunc),
-			grpcprom.StreamServerInterceptor,
-			dispatchmw.StreamServerInterceptor(dispatcher),
-			datastoremw.StreamServerInterceptor(ds),
-			consistencymw.StreamServerInterceptor(),
-			servicespecific.StreamServerInterceptor,
-			serverversion.StreamServerInterceptor(enableVersionResponse),
-		}
+const (
+	DefaultMiddlewareRequestID = "requestid"
+	DefaultMiddlewareLog       = "log"
+	DefaultMiddlewareGRPCLog   = "grpclog"
+	DefaultMiddlewareOTelGRPC  = "otelgrpc"
+	DefaultMiddlewareGRPCAuth  = "grpcauth"
+	DefaultMiddlewareGRPCProm  = "grpcprom"
+
+	DefaultInternalMiddlewareDispatch       = "dispatch"
+	DefaultInternalMiddlewareDatastore      = "datastore"
+	DefaultInternalMiddlewareConsistency    = "consistency"
+	DefaultInternalMiddlewareServerSpecific = "servicespecific"
+	DefaultInternalMiddlewareServerVersion  = "serverversion"
+)
+
+// DefaultMiddleware generates the default middleware chain used for the public SpiceDB gRPC API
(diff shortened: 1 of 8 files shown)


pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+// MiddlewareChain describes an ordered sequence of middlewares that can be modified
+// with one or more MiddlewareModification. This struct is used to facilitate the
+// creation and modification of gRPC middleware chains
+type MiddlewareChain struct {
+	chain []ReferenceableMiddleware
+}
+
+// NewMiddlewareChain creates a new middleware chain given a zero or more named middlewares.
```
josephschorr, line 18: remove `a` before `zero`

pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,207 @@
...
+func NewMiddlewareChain(mw ...ReferenceableMiddleware) (MiddlewareChain, error) {
+	if err := validate(mw); err != nil {
+		return MiddlewareChain{}, err
+	}
+	return MiddlewareChain{chain: mw}, nil
+}
+
+// MiddlewareModification describes an operation to modify a MiddlewareChain
+type MiddlewareModification struct {
+	// Dependency is used to define with respect to which middleware an operation is performed
```
josephschorr, line 29: Maybe we should name it `DependencyMiddlewareName` to be explicit? Not sure its necessary

josephschorr: Should also comment this should be empty for replace all

pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,207 @@
...
+	Dependency string
+	// Operation describes the type of operation to be performed
+	Operation MiddlewareOperation
+	// Middlewares the named middlewares that will be part of this modification
```
josephschorr, line 33: Middlewares are the named...

pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,207 @@
...
+			return fmt.Errorf("found middleware with duplicate names in middleware modification: %s", mw.Name)
+		}
+	}
+	return nil
+}
+
+// ReferenceableMiddleware represents a middleware in a MiddlewareChain. Middlewares can
+// be referenceable in MiddlewareModification, for example "append after middleware abc".
+// Middlewares are referenced by name.
+// A named middleware can be also be unreferenceable, in which case can be left unnamed.
```
josephschorr, line 63: in which case the middleware can be left...

pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,207 @@
...
+	StreamingMiddleware grpc.StreamServerInterceptor
+}
+
+// MiddlewareOperation describes the type of operation that will be performed in a MiddlewareModification
+type MiddlewareOperation int
+
+const (
+	// OperationPrepend adds the middlewares right before the referenced dependency
+	OperationPrepend MiddlewareOperation = iota
+	// OperationReplace substitutes the referenced dependency with the middlewares of a modification
```
josephschorr, line 77: newlines between the comments and the previous operation types

josephschorr, line 77: Add that if the replacements slice is empty, this acts like a delete

pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,207 @@
...
+	if !existingNames.Has(mod.Dependency) {
+		return fmt.Errorf("referenced dependency does not exist on chain: %s", mod.Dependency)
+	}
+	// prevent appending/prepending a duplicate middleware
+	for _, middleware := range mod.Middlewares {
+		if existingNames.Has(middleware.Name) && mod.Dependency == middleware.Name && mod.Operation != OperationReplace {
+			return fmt.Errorf("modification will cause a duplicate in chain: %s", middleware.Name)
+		}
+	}
+	// prevent replacing an internal middleware
```
josephschorr, line 176: newline before this line

pkg/cmd/server/server_test.go
```diff
@@ -41,3 +43,51 @@ func TestServerGracefulTerminationOnError(t *testing.T) {
...
+					StreamingMiddleware: mockStreamInterceptor{val: errors.New("hi")}.streamIntercept,
+				},
+			},
+		},
+	}}
+	unary, streaming, err := c.buildMiddleware(nil)
+	require.NoError(t, err)
+	require.Len(t, unary, 1)
+	require.Len(t, streaming, 1)
+	val, _ := unary[0](context.Background(), nil, nil, nil)
```
josephschorr, line 64: I like to put newlines after sets of requires before the next "executing" statement

pkg/cmd/server/middleware.go
vroldanbet: this is actually stale. I changed the semantics here a bit: ReferenceableMiddleware can now be marked as `internal`. This only means you cannot replace that middleware, but can reference it for append or prepend. The goal is to mark middleware that shouldn't be removed at all.

pkg/cmd/server/middleware.go
vroldanbet: good point. I'll add a test for that as well

pkg/cmd/server/defaults.go
```diff
@@ -100,32 +100,88 @@ var defaultGRPCLogOptions = []grpclog.Option{
...
+	DefaultMiddlewareGRPCProm  = "grpcprom"
+
+	DefaultInternalMiddlewareDispatch       = "dispatch"
+	DefaultInternalMiddlewareDatastore      = "datastore"
+	DefaultInternalMiddlewareConsistency    = "consistency"
+	DefaultInternalMiddlewareServerSpecific = "servicespecific"
+	DefaultInternalMiddlewareServerVersion  = "serverversion"
+)
+
+func DefaultMiddleware(logger zerolog.Logger, authFunc grpcauth.AuthFunc, enableVersionResponse bool, dispatcher dispatch.Dispatcher, ds datastore.Datastore) *MiddlewareChain {
```
josephschorr, line 118: Add comment on exported member

pkg/cmd/server/middleware.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	return MiddlewareChain{chain: mw}, nil
+}
+
+// MiddlewareModification describes an operation to modify a MiddlewareChain
+type MiddlewareModification struct {
+	// DependencyMiddlewareName is used to define with respect to which middleware an operation is performed.
+	// Dependency is not required for ReplaceAll operation
+	DependencyMiddlewareName string
+	// Operation describes the type of operation to be performed
```
josephschorr, line 32: newlines before comments

josephschorr (approved):
LGTM

PR #1071, opened 2022-12-24
Add accessor in the dev package for V1 API

josephschorr:
Will be used for additional API tooling

### DIFF
--- a/pkg/development/devcontext.go
+++ b/pkg/development/devcontext.go
@@ -3,18 +3,25 @@ package development
 import (
 	"context"
 	"errors"
+	"net"
 	"time"
 
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
+	"google.golang.org/grpc/test/bufconn"
 
 	"github.com/authzed/spicedb/internal/datastore/memdb"
 	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/internal/dispatch/graph"
 	maingraph "github.com/authzed/spicedb/internal/graph"
 	log "github.com/authzed/spicedb/internal/logging"
+	"github.com/authzed/spicedb/internal/middleware/consistency"
 	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
 	"github.com/authzed/spicedb/internal/namespace"
+	v1svc "github.com/authzed/spicedb/internal/services/v1"
 	"github.com/authzed/spicedb/internal/sharederrors"
 	"github.com/authzed/spicedb/pkg/datastore"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/development/devcontext.go
```diff
@@ -102,6 +109,56 @@ func newDevContextWithDatastore(ctx context.Context, requestContext *devinterfac
...
+
+	s := grpc.NewServer(
+		grpc.ChainUnaryInterceptor(
+			datastoremw.UnaryServerInterceptor(dc.Datastore),
+			consistency.UnaryServerInterceptor(),
+		),
+		grpc.ChainStreamInterceptor(
+			datastoremw.StreamServerInterceptor(dc.Datastore),
+			consistency.StreamServerInterceptor(),
+		),
```
vroldanbet, line 128: Perhaps we could move this to `defaults.go` as a function like `RequiredMiddleware` that is then reused in `DefaultMiddleware`. These are the _required_ middleware for SpiceDB to actually properly work, but it's not explicit anywhere. In `defaults.go`, with recent changes introduced in https://github.com/authzed/spicedb/pull/1063, we tried to capture what are "optional middleware" and "required middleware" by marking a middleware as `internal`.

pkg/development/development_test.go
```diff
@@ -109,3 +110,33 @@ definition document {
 
 	require.Equal(t, "document:somedoc#viewer:\n- '[user:someuser[...]] is <document:somedoc#viewer>'\n", generated)
 }
+
+func TestDevContextV1Service(t *testing.T) {
+	defer goleak.VerifyNone(t, goleak.IgnoreTopFunction("github.com/golang/glog.(*loggingT).flushDaemon"), goleak.IgnoreCurrent())
```
vroldanbet, line 115: where is that glog goroutine coming from?

pkg/development/devcontext.go
josephschorr: I can't have this import any file that relies upon the other middlewares, otherwise it won't build under WASM. This was placed here explicitly to ensure we only need the bare minimum, while also not breaking the existing WASM build.

pkg/development/development_test.go
josephschorr: Its the goroutine used under the covers by zerolog to log things.

josephschorr:
Updated

pkg/development/devcontext.go
vroldanbet: let's then turn that into a method like `defaultWasmMiddleware` in this package? WDYT?

pkg/development/development_test.go
vroldanbet: that's weird. I've written tests that bring up a full blown server and I didn't have to add those ignores 😕

pkg/development/devcontext.go
josephschorr: Just for this one usage? Why?

pkg/development/development_test.go
josephschorr: They are only needed if you're tracking remaining goroutines via `goleak`

pkg/development/devcontext.go
vroldanbet: makes more explicit what middleware is needed to run a barebone SpiceDB in wasm?

It's also fine to leave it as is - my initial assumption was this was the required middleware for a normal spicedb deployment as well

pkg/development/development_test.go
vroldanbet: as I said, I have tests running with `goleak` that start up spicedb, and do not have that ignore, so I wondered what was the difference

vroldanbet (approved):

PR #1074, opened 2022-12-28
Add mention about PHP library to README

amsprost:
There's a PHP client library based on REST protocol (because of issues with generation from proto 2 & PHP). Since client libraries are official repos under authzed organisation, but because it can be beneficial for PHP developers to use this library in order being able to manage app permissions maybe can be useful to have a link related to this library in README. If you have better idea how to mention PHP library in this repo - please let me know, I'm open to suggestions

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
 SpiceDB is an open source database system for managing security-critical application permissions inspired by Google's [Zanzibar] paper.
 
 Developers create a schema that models their permissions requirements and use a [client library] to apply the schema to the database, insert data into the database, and query the data to efficiently check permissions in their applications.
+For PHP [this library](https://github.com/linkorb/spicedb-php) can be used.
 
 [client library]: https://github.com/orgs/authzed/repositories?q=client+library
 


amsprost:

vroldanbet (changes requested):
Hey @amsprost, thanks for the contribution! We are deliberately not exhaustive with client libraries in the spicedb repository. Instead we list those under the https://github.com/authzed/awesome-spicedb repo. We have there a PHP library listed already, we weren't aware of this one. What are the differences?

amsprost:
> Hey @amsprost, thanks for the contribution! We are deliberately not exhaustive with client libraries in the spicedb repository. Instead we list those under the https://github.com/authzed/awesome-spicedb repo. We have there a PHP library listed already, we weren't aware of this one. What are the differences?

Thanks @vroldanbet ! I wasn't aware of that other PHP library and that other repo. Perhaps, it can be mentioned more explicitly inside README.md on the main repo of SpiceDB as atm it shows only official libraries, so hard to get awareness of non-official libraries.

But as for other PHP library it's mentioned there that it's not production ready & there's no plans to fix that. Also that lib requires PHP 8.0, as mine can work on PHP 7.4. All other differences are minor as that lib is also based on REST API.

But thank you very much for feedback, I'm closing this PR & will open a new one to https://github.com/authzed/awesome-spicedb

vroldanbet:
@amsprost thank you for the feedback! I've opened https://github.com/authzed/spicedb/pull/1076 to link to awesome-spicedb repo instead, which should keep a more exhaustive list of client libraries. 

As for the PHP library, I think if we conclude https://github.com/linkorb/spicedb-php is better suited for PHP users, we would be happy to list it in awesome-spicedb. @jzelinskie could you perhaps have a look at https://github.com/linkorb/spicedb-php and see if it makes the cut to awesome-spicedb as replacement of the currently listed one?

PR #1076, opened 2022-12-29
reference libraries through awesome spicedb

vroldanbet:
We maintain a list of community-mantained client libraries, but are not linking it from the SpiceDB Repo.


josephschorr (approved):
LGTM

PR #1086, opened 2023-01-06
Add additional goroutine leak testing to Lookup* and fix possible deadlock in ReachableResources

josephschorr:


internal/graph/taskrunner.go
```diff
@@ -0,0 +1,134 @@
...
+				<-tr.sem
+				tr.lock.Unlock()
+				return
+			}
+
+			task = tr.tasks[0]
+			tr.tasks = tr.tasks[1:]
+			tr.lock.Unlock()
+
+			err := task(tr.ctx)
```
vroldanbet, line 116: What's not so great here is that we are not winding down the runner as fast as possible upon error/cancellation of any task because we are waiting for the currently running tasks to complete. If we have max concurrency of 10, and 10 tasks are running but one fails, we will be waiting for 9 potentially long-running tasks to complete.

We could perhaps we can turn `TaskRunner.Wait` into `TaskRunner.Done` and make it a channel, so that callsites can `select` over it and the `ctx.Done()` channel?

internal/graph/taskrunner.go
```diff
@@ -0,0 +1,134 @@
+package graph
+
+import (
+	"context"
+	"sync"
+)
+
+type token struct{}
```
vroldanbet, line 8: nit: maybe `workerReleased` instead?

internal/graph/taskrunner.go
```diff
@@ -0,0 +1,134 @@
...
+type TaskRunner struct {
+	// ctx holds the context given to the task runner and annotated with the cancel
+	// function.
+	ctx    context.Context
+	cancel func()
+
+	// err holds the error returned by any task, if any. If the context is canceled,
+	// this err will hold the cancelation error.
+	errOnce sync.Once
+	err     error
```
vroldanbet, line 21: nit: since you are a comment over the `err` field, make the comment on top of it
```suggestion
	// err holds the error returned by any task, if any. If the context is canceled,
	// this err will hold the cancelation error.
	err     error
	errOnce sync.Once
```

pkg/testutil/require.go
```diff
@@ -42,3 +43,27 @@ func RequireEqualEmptyNil(t *testing.T, expected, actual interface{}, msgAndArgs
...
+		runner(t)
+		ch <- true
+	})()
+
+	select {
+	case <-timer.C:
+		require.Failf(t, "timed out waiting for runner", "expected to complete in %v", timeout)
+
+	case <-ch:
+		timer.Stop()
```
vroldanbet, line 66: do we to call `timer.Stop()` if you've deferred it above?

internal/dispatch/graph/check_test.go
```diff
@@ -403,13 +403,13 @@ func TestCheckDebugging(t *testing.T) {
 	}
 }
 
-func newLocalDispatcher(t testing.TB) (context.Context, dispatch.Dispatcher, datastore.Revision) {
+func newLocalDispatcherWithConsistencyLimit(t testing.TB, consistencyLimit uint16) (context.Context, dispatch.Dispatcher, datastore.Revision) {
```
vroldanbet, line 406: I think you meant `concurrencyLimit`
```suggestion
func newLocalDispatcherWithConsistencyLimit(t testing.TB, concurrencyLimit uint16) (context.Context, dispatch.Dispatcher, datastore.Revision) {
```

internal/dispatch/graph/reachableresources_test.go
```diff
@@ -603,3 +609,109 @@ func TestCaveatedReachableResources(t *testing.T) {
 		})
 	}
 }
+
+func TestReachableResourcesWithConsistencyLimitOf1(t *testing.T) {
```
vroldanbet, line 613: consistency -> concurrency?

```suggestion
func TestReachableResourcesWithConcurencyLimitOf1(t *testing.T) {
```

internal/dispatch/graph/reachableresources_test.go
```diff
@@ -603,3 +609,109 @@ func TestCaveatedReachableResources(t *testing.T) {
...
+	defer goleak.VerifyNone(t, goleakIgnores...)
+
+	ctx, dispatcher, revision := newLocalDispatcherWithConsistencyLimit(t, 1)
```
vroldanbet, line 616: can we somehow assert that no more than one goroutine is spawn? There are already tests in `TaskRunner`, but would be nice to assert the dispatcher properly integrates it.

internal/dispatch/graph/reachableresources_test.go
```diff
@@ -603,3 +609,109 @@ func TestCaveatedReachableResources(t *testing.T) {
...
+	require.NoError(t, err)
+
+	for range stream.Results() {
+		// Break early
+		break
+	}
+	dispatcher.Close()
+}
+
+func TestReachableResourcesMultipleEntrypointEarlyCancel(t *testing.T) {
```
vroldanbet, line 642: My assumption is that the leak originates from cancellation not being propagated.

internal/dispatch/graph/reachableresources_test.go
vroldanbet, line 613: Does this test reproduce the "possible deadlock" as well? The PR body does not describe how the deadlock occurs, so I assume it has to do with unschedulable dispatch operations. How did that happen with the errgroup?

internal/dispatch/graph/reachableresources_test.go
```diff
@@ -603,3 +609,109 @@ func TestCaveatedReachableResources(t *testing.T) {
...
+				permission view = viewer + writer + admin + namespace->view + org->member
+			}
+		`,
+		testRels,
+		require.New(t),
+	)
+	dispatcher := NewLocalOnlyDispatcher(2)
+
+	ctx := log.Logger.WithContext(datastoremw.ContextWithHandle(context.Background()))
+	require.NoError(t, datastoremw.SetInContext(ctx, ds))
```
vroldanbet, line 691: this seems extraneous to the test at first glance. Are you doing it so the datastore operations can be canceled since the context is not propagated to them and leaves goroutines hanging around?

internal/graph/reachableresources.go
```diff
@@ -86,17 +84,13 @@ func (crr *ConcurrentReachableResources) ReachableResources(
 		return err
 	}
 
-	cancelCtx, checkCancel := context.WithCancel(ctx)
-	defer checkCancel()
-
-	g, subCtx := errgroup.WithContext(cancelCtx)
-	g.SetLimit(int(crr.concurrencyLimit))
```
vroldanbet, line 93: so from reading `errgroup` docs, `Go` function is blocking when the concurrency limit is hit. Is the assumption that the deadlock is occurring in the `lookupRelationEntrypoint`, `lookupTTUEntrypoint` and `chunkedRedispatch`?

internal/graph/taskrunner.go
josephschorr: The context itself is canceled, which means any runners should themselves be early terminating

internal/graph/taskrunner.go
josephschorr: I honestly don't know a great name for it; I just copied `token` from the errgroup impl

pkg/testutil/require.go
josephschorr: No, but I'll add it

josephschorr: No, but I like being explicit where possible

internal/dispatch/graph/check_test.go
josephschorr: Whoops. Fixed

internal/dispatch/graph/reachableresources_test.go
josephschorr: I'm not sure how we could reasonably do so. Any ideas?

internal/dispatch/graph/reachableresources_test.go
josephschorr: That and the possibility of deadlock causing things to just hang forever

internal/dispatch/graph/reachableresources_test.go
josephschorr: Yes, it was the first test I wrote to demonstrate it

internal/dispatch/graph/reachableresources_test.go
josephschorr: Its necessary to call the dispatcher; without it, it won't know the datastore to use

internal/graph/reachableresources.go
josephschorr: One of them, yes. The test with limit 1 was able to reproduce consistently.

josephschorr:
Updated

internal/graph/taskrunner.go
vroldanbet: isn't the `task` invocation here synchronous and blocking? indeed, if the `task` function reacts to cancellation ASAP (which they should) then it should be fine 👍🏻

pkg/testutil/require.go
vroldanbet: well it could have the effect of being confused

internal/dispatch/graph/reachableresources_test.go
vroldanbet: I have no clue honestly, was hoping there is some library out there that helps with that

internal/dispatch/graph/reachableresources_test.go
vroldanbet: oh whoops, you are right

internal/dispatch/graph/reachableresources_test.go
josephschorr: https://pkg.go.dev/runtime#NumGoroutine exists but we'd have to check exactly when it is running, which seems... problematic

vroldanbet (approved):

internal/graph/taskrunner.go
josephschorr: The task function reacting to the context cancelation is on the implementor of the tasks given; all our dispatching code should be reacting properly to it (and I indeed added an additional check into ReachableResources)

PR #1089, opened 2023-01-10
refactor datastore flags to make them reusable

vroldanbet:
This PR refactors `datastore.RegisterDatastoreFlags` to make them reusable in new scenarios, particularly when we may have a process that runs multiple SpiceDB with various datastores.

PR #1092, opened 2023-01-11
align datastore defaults

vroldanbet:
Built on top of https://github.com/authzed/spicedb/pull/1089

I noticed that `DefaultDatastoreConfig` didn't have the same config values as `RegisterDatastoreFlags`, so I addressed this in 2 commits:
- one that aligned `DefaultDatastoreConfig` to the values in `RegisterDatastoreFlags` and tested they were the same, to make sure I didn't miss any values and caused a potential problematic regression
- once they were aligned, I made `RegisterDatastoreFlags` reuse `DefaultDatastoreConfig`


josephschorr (approved):
LGTM

PR #1094, opened 2023-01-12
adds log.Ctx(ctx) calls (almost) everywhere

vroldanbet:
This is so that logs can be enriched with new tags via the context


josephschorr (approved):
LGTM

PR #1096, opened 2023-01-13
Cleanup lock handling in task runner

josephschorr:
Makes things easier to read and fixes a bug that could lead to a deadlock

### DIFF
--- a/internal/graph/taskrunner.go
+++ b/internal/graph/taskrunner.go
@@ -15,14 +15,13 @@ type TaskRunner struct {
 	ctx    context.Context
 	cancel func()
 
-	// err holds the error returned by any task, if any. If the context is canceled,
-	// this err will hold the cancelation error.
-	err     error
-	errOnce sync.Once
-
 	// sem is a chan of length `concurrencyLimit` used to ensure the task runner does
 	// not exceed the concurrencyLimit with spawned goroutines.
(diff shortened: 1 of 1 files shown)


internal/graph/taskrunner.go
```diff
@@ -50,13 +49,9 @@ func NewTaskRunner(ctx context.Context, concurrencyLimit uint16) *TaskRunner {
...
 // task handler function and immediately returns.
 func (tr *TaskRunner) Schedule(f TaskFunc) {
-	tr.wg.Add(1)
-
-	tr.lock.Lock()
-	tr.tasks = append(tr.tasks, f)
-	tr.lock.Unlock()
-
-	tr.spawnIfAvailable()
+	if tr.addTask(f) {
```
vroldanbet, line 52: why silently ignoring? `addTask` could return `error` instead and `Schedule` as well, hitting that we should not longer continue scheduling work

vroldanbet (commented):
just one question

internal/graph/taskrunner.go
josephschorr: The interface for `Schedule` is a "fire and forget". The error itself will be returned by the `Wait` command

vroldanbet (approved):

PR #1098, opened 2023-01-13
Optimize allocations by removing sprintf, using strings.Cut

jzelinskie:
An allocs profile showed that we're making lots of allocations from `fmt.Sprintf` (nearly half as many as allocations from allocation protos!). I removed many of the simple usages from the critical path to save allocations.

The second commit also adopts `strings.Cut` where it makes sense for further optimization.

### DIFF
--- a/internal/developmentmembership/trackingsubjectset.go
+++ b/internal/developmentmembership/trackingsubjectset.go
@@ -1,8 +1,6 @@
 package developmentmembership
 
 import (
-	"strings"
-
 	"github.com/authzed/spicedb/internal/datasets"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/tuple"
@@ -83,26 +81,19 @@ func (tss *TrackingSubjectSet) Add(subjectsAndResources ...FoundSubject) error {
 	return nil
 }
 
-func typeKey(ns, rel string) string { return ns + "#" + rel }
-
-func keyFor(fs FoundSubject) string {
-	return typeKey(fs.subject.Namespace, fs.subject.Relation)
-}
-
 func (tss *TrackingSubjectSet) getSetForKey(key string) datasets.BaseSubjectSet[FoundSubject] {
 	if existing, ok := tss.setByType[key]; ok {
 		return existing
 	}
 
-	parts := strings.Split(key, "#")
-
-	created := datasets.NewBaseSubjectSet[FoundSubject](
+	ns, rel := tuple.MustParseRelationTypeString(key)
+	created := datasets.NewBaseSubjectSet(
 		func(subjectID string, caveatExpression *core.CaveatExpression, excludedSubjects []FoundSubject, sources ...FoundSubject) FoundSubject {
 			fs := NewFoundSubject(&core.DirectSubject{
 				Subject: &core.ObjectAndRelation{
-					Namespace: parts[0],
+					Namespace: ns,
 					ObjectId:  subjectID,
-					Relation:  parts[1],
+					Relation:  rel,
 				},
 				CaveatExpression: caveatExpression,
 			})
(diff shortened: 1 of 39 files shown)


internal/developmentmembership/trackingsubjectset.go
```diff
@@ -84,24 +83,28 @@ func (tss *TrackingSubjectSet) Add(subjectsAndResources ...FoundSubject) error {
 	return nil
 }
 
-func keyFor(fs FoundSubject) string {
-	return fmt.Sprintf("%s#%s", fs.subject.Namespace, fs.subject.Relation)
+func typeKey(ns, rel string) string { return ns + "#" + rel }
```
josephschorr, line 86: we repeat this a lot now... maybe move into the `tuple` package and make public and unit tested?

internal/dispatch/keys/hasher_common.go
```diff
@@ -27,9 +27,7 @@ type hashableRelationReference struct {
 }
 
 func (hrr hashableRelationReference) AppendToHash(hasher hasherInterface) {
-	hasher.WriteString(hrr.Namespace)
-	hasher.WriteString("#")
-	hasher.WriteString(hrr.Relation)
+	hasher.WriteString(hrr.Namespace + "#" + hrr.Relation)
```
josephschorr, line 30: are we sure this better than multiple WriteString calls?

jzelinskie: Yep!

pkg/tuple/tuple.go
```diff
@@ -434,6 +435,16 @@ func UpdateFromRelationshipUpdate(update *v1.RelationshipUpdate) *core.RelationT
 	}
 }
 
+func RelationTypeString(namespace, relation string) string { return namespace + "#" + relation }
```
josephschorr, line 438: doc comments

internal/dispatch/graph/graph.go
```diff
@@ -154,28 +154,12 @@ func (ld *localDispatcher) lookupRelation(ctx context.Context, ns *core.Namespac
...
-
-func (rr stringableRelRef) String() string {
-	return fmt.Sprintf("%s::%s", rr.Namespace, rr.Relation)
-}
-
 // DispatchCheck implements dispatch.Check interface
 func (ld *localDispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckRequest) (*v1.DispatchCheckResponse, error) {
 	ctx, span := tracer.Start(ctx, "DispatchCheck", trace.WithAttributes(
-		attribute.Stringer("resource-type", stringableRelRef{req.ResourceRelation}),
+		attribute.String("resource-type", tuple.StringRR(req.ResourceRelation)),
```
josephschorr, line 160: Downside of this change is that we'll always be creating the string form of the resource relation

internal/dispatch/keys/hasher_common.go
```diff
@@ -111,14 +110,13 @@ func (hsv hashableStructValue) AppendToHash(hasher hasherInterface) {
 		hasher.WriteString("null")
 
 	case *structpb.Value_NumberValue:
-		hasher.WriteString(fmt.Sprintf("%f", t.NumberValue))
+		// AFAICT, this is how Sprintf-style formats float64s
```
josephschorr, line 113: So long as it produces a valid and unique hash, doesn't really matter the exact form

pkg/tuple/relationship.go
```diff
@@ -1,22 +1,24 @@
...
 
 import (
-	"fmt"
-
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 )
 
+// JoinObject joins the namespace and the objectId together into the standard
+// format.
+func JoinObjectRef(namespace, objectID string) string { return namespace + ":" + objectID }
```
josephschorr, line 9: Can you add some basic unit tests for this new methods?

internal/dispatch/graph/graph.go
jzelinskie: True, what if I made the Stringer struct generic, so we could at least reuse it other places?

```go
type CustomStringer[T any] struct {
  fn func(T) string
  T
}

func (cs *CustomStringer[T]) String() string { return cs.fn(cs.T) }
```

josephschorr: Sounds good to me

jzelinskie: On second thought, that'd put `fn` on the heap. I think since we always ultimately do create the string, doing what's in the PR is fine.

josephschorr (approved):
LGTM

Issue #1100, opened 2023-01-17
spicedb migrate does not output help text


easyCZ:
When `spicedb migrate` is invoked, no help text is provided.

This becomes even more confusing when I forget to specify the `revision` in the command. This outputs no information, and exists with status code 1.
```
spicedb migrate \
    --datastore-conn-uri='<URI>' \
    --datastore-engine="mysql" \
    --log-level="debug"
```

The fix here would be to validate the input arguments and fail with the help text, if the `revision` is not provided/valid.

vroldanbet:
Thanks for the report! I've opened https://github.com/authzed/spicedb/pull/1101 to address it

PR #1101, opened 2023-01-17
fixes positional argument errors failing silently

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/1100

unfortunately cobra positional arguments are not governed by the same logic as flag parsing, and as a consequence, we are not logging them out.

This is also compounded by the fact the global logger is not initialized until it's injected via
pkg/cmd/server/defaults.go/DefaultPreRunE. We recently moved to use our own global logger instead of ZeroLog's, and we've set it to Noop by default. As a consequence, the positional argument error was being printed by a noop logger.

ideally, we would invert the control via `main.go` and let `DefaultPreRunE` (which sets the global logger in the command stack) receive a logger instance, instead of it creating it and injecting it back.

In the meanwhile, and as a workaround, we set a default global logger so that errors before positional argument parsing are logged out. This default logger will be overwritten once the cobra `PreRunE` is executed, invoking `cobrautil.CommandStack`

### DIFF
--- a/cmd/spicedb/main.go
+++ b/cmd/spicedb/main.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/cespare/xxhash/v2"
+	"github.com/rs/zerolog"
 	"github.com/sercand/kuberesolver/v3"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc/balancer"
@@ -39,6 +40,8 @@ func main() {
 		hashringReplicationFactor,
 		backendsPerKey,
 	))
+
+	log.SetGlobalLogger(zerolog.New(os.Stdout))
 
 	// Create a root command
 	rootCmd := cmd.NewRootCommand("spicedb")


jzelinskie:
Why change it here and change the package default not to be noop?

vroldanbet:
@jzelinskie the default state was set deliberately to be noop because our tests generated logs and it made it very difficult to identify which tests failed, made GitHub Actions logs UI unresponsive or even sometimes the UI refused to show the logs due to the length.

The alternative is to set it in every test to noop, and we discussed that was a no-go and the very same reason we ended up engineering it this way.

jzelinskie:
Could make it set to `noop` in an `init()` in each test, too maybe.
It kinda sucks still, but at least it's at the package layer instead of each test.

vroldanbet:
>Could make it set to noop in an init() in each package, too maybe.
It kinda sucks still, but at least it's at the package layer instead of each test.

thing is, that logger is going to get overwritten as soon as the cobra command stack kicks in. We would be adding all those `init` function, which can be missed, just to have a logger for an insignificant portion of the lifecycle of the binary (start of main -> cobra command is run)

vroldanbet:
friendly ping @jzelinskie ☝🏻

jzelinskie (approved):

PR #1102, opened 2023-01-17
Add a distinct validation error type for schema write

josephschorr:
This allows us to distinguish between internal errors and validation errors

### DIFF
--- a/internal/services/shared/errors.go
+++ b/internal/services/shared/errors.go
@@ -1,6 +1,10 @@
 package shared
 
 import (
+	"errors"
+	"fmt"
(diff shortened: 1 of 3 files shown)


internal/services/shared/errors.go
```diff
@@ -1,6 +1,11 @@
 package shared
 
 import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
```
vroldanbet, line 7: please third party imports not being in the same blog

vroldanbet (approved):

PR #1108, opened 2023-01-19
pkg/singleflight: init, apply to remote dispatch

### LINKED ISSUE #20: Deduplicate dispatching within a short window
When the same read-only requests occur within a configurable window of time, a single dispatch can be performed to fetch their results.

A useful library for implementing this is the [golang.org/x/sync/singleflight][singleflight].

[singleflight]: https://pkg.go.dev/golang.org/x/sync/singleflight

jzelinskie:
Fixes #20


jzelinskie:
This is obviated by #1607 and the existence of [resenje.org/singleflight](https://pkg.go.dev/resenje.org/singleflight) which works with context propagation at the cost of an additional goroutine.

PR #1109, opened 2023-01-19
fixes problem with caveats not resolving protobuf types

vroldanbet:
as we were doing some experiments to do dynamic caveats, we discovered the CEL Environment with a CustomTypeAdapter was interfering with the way CEL resolves protobuf types reflectively.

It turns out the adapter we added was a relic from the past, and it wasn't necessary anymore


pkg/caveats/env.go
```diff
@@ -63,7 +63,6 @@ func (e *Environment) asCelEnvironment() (*cel.Env, error) {
 	opts := make([]cel.EnvOption, 0, len(e.variables)+len(types.CustomTypes)+2)
 
 	// Add the custom type adapter and functions.
```
josephschorr, line 65: Fix comment

josephschorr (changes requested):

josephschorr (approved):
LGTM

PR #1110, opened 2023-01-19
pkg/cache: default TTL of 2x quantization window

jzelinskie:
This should help reduce memory usage as it'll more eagerly expire keys that we know won't ever be queried again.

### DIFF
--- a/pkg/cache/cache_ristretto.go
+++ b/pkg/cache/cache_ristretto.go
@@ -4,6 +4,8 @@
 package cache
 
 import (
+	"time"
+
 	"github.com/outcaste-io/ristretto"
 	"github.com/outcaste-io/ristretto/z"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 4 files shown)


pkg/cache/cache_ristretto.go
```diff
@@ -29,14 +31,22 @@ func NewCache(config *Config) (Cache, error) {
...
 
 type wrapped struct {
-	config *Config
+	config     *Config
+	defaultTTL time.Duration
 	*ristretto.Cache
 }
 
+func (w wrapped) Set(key, entry any, cost int64) bool {
+	if w.defaultTTL == 0 {
```
josephschorr, line 44: What happens if the TTL is negative? Maybe we should catch that case here?

pkg/cmd/server/server.go
```diff
@@ -200,7 +200,7 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 
 	dispatcher := c.Dispatcher
 	if dispatcher == nil {
-		cc, err := c.DispatchCacheConfig.Complete()
+		cc, err := c.DispatchCacheConfig.WithTTL(2 * c.DatastoreConfig.RevisionQuantization).Complete()
```
josephschorr, line 203: move this into a variable and add a comment explaining why it is * 2

josephschorr (approved):
LGTM

PR #1115, opened 2023-01-23
Add consistency test for maps in caveats and better typed errors on caveat evaluation

josephschorr:


pkg/caveats/errors.go
```diff
@@ -7,6 +7,21 @@ import (
...
+}
+
+// MarshalZerologObject implements zerolog.LogObjectMarshaler
+func (err EvaluationErr) MarshalZerologObject(e *zerolog.Event) {
+	e.Err(err.error)
+}
+
+// DetailsMetadata returns the metadata for details for this error.
+func (err EvaluationErr) DetailsMetadata() map[string]string {
+	return map[string]string{}
```
vroldanbet, line 22: no need to allocate, just return `nil` for its zero value

```suggestion
	return nil
```

internal/caveats/errors.go
```diff
@@ -0,0 +1,107 @@
...
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+
+	"github.com/authzed/spicedb/pkg/caveats"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+// CaveatEvaluationErr is an error in evaluation of a caveat expression.
+type CaveatEvaluationErr struct {
```
vroldanbet, line 19: given the package name is `caveat`, the `Caveat` prefix in the error types are redundant.

`caveat.EvaluationErr` better than `caveat.CaveatEvaluationErr`

pkg/caveats/errors.go
```diff
@@ -7,6 +7,21 @@ import (
 	"github.com/rs/zerolog"
 )
 
+// EvaluationErr is an error in evaluation of a caveat expression.
+type EvaluationErr struct {
```
vroldanbet, line 11: Seems similar to the error in `internal/caveats`. Any chances we can move the one internal here and import it from `internal/caveats`?

internal/services/v1/permissions_test.go
```diff
@@ -865,6 +865,97 @@ func TestCheckWithCaveats(t *testing.T) {
...
+		},
+		{
+			"wrong value in map",
+			map[string]any{
+				"somemap": map[string]any{
+					"first":  42,
+					"second": "hello",
+				},
+			},
+			"evaluation error for caveat somecaveat: no such overload",
```
vroldanbet, line 929: any chances we can get more info in the error message about the type that caused the overload error?

pkg/caveats/errors.go
josephschorr: Nope; this was explicitly done (and hence the rename above) because it needs to add additional context to it

internal/services/v1/permissions_test.go
josephschorr: I wish... I haven't seen any way to do so. I plan to look again at some point, but I didn't find any obvious ways to extract more details.

internal/caveats/errors.go
josephschorr: Done

josephschorr:
Updated and rebased

pkg/caveats/errors.go
josephschorr: I'm intending to add context here at some point, and I'm not 100% sure our metadata code handles `nil`, so I'd rather just allocate here

vroldanbet (approved):

PR #1118, opened 2023-01-24
Add covering indexes to Postgres and associated testing

josephschorr:
We add missing indexes to ensure that relationship query operations and GC operations are done primarily from indexes in Postgres

This PR also adds tests to ensure that index-only scans are used for relationship queries, and that GC has the necessary indexes defined

Includes migrations to add these new indexes (concurrently)

Replaces https://github.com/authzed/spicedb/pull/1050


internal/datastore/postgres/options.go
```diff
@@ -264,6 +269,15 @@ func DebugAnalyzeBeforeStatistics() Option {
 	}
 }
 
+// WithQueryInterceptor adds a query logger for use in tests.
```
vroldanbet, line 272:
```suggestion
// WithQueryInterceptor adds an interceptor to all underlying postgres queries
```

vroldanbet (dismissed):
This is looking good to me, I'd suggest adding some testing to the `queryInterceptor` type to make sure it delegates to the underlying pool if the interceptor is nil

vroldanbet (approved):

PR #1124, opened 2023-01-26
Add caching of caveats in caching proxy for datastore

### LINKED ISSUE #892: Converge some of the datastore namespace loading methods
We have ReadNamespace, LookupNamespaces, and ListNamespaces in the datastore API, we can probably get rid of at least one of them.

josephschorr:
This PR adds support for caching caveats alongside namespaces in the datastore caching proxy and also adds support for having calls to the methods`Lookup{Caveats,Namespaces}WithNames` automatically be cached and use the cache when available.

As part of this PR, I also standardized the naming of the Lookup and List methods in the datastore for namespaces and caveats

Fixes #892

### DIFF
--- a/internal/datastore/crdb/caveat.go
+++ b/internal/datastore/crdb/caveat.go
@@ -62,7 +62,18 @@ func (cr *crdbReader) ReadCaveatByName(ctx context.Context, name string) (*core.
 	return loaded, revisionFromTimestamp(timestamp), nil
 }
 
-func (cr *crdbReader) ListCaveats(ctx context.Context, caveatNames ...string) ([]*core.CaveatDefinition, error) {
+func (cr *crdbReader) LookupCaveatsWithNames(ctx context.Context, caveatNames []string) ([]*core.CaveatDefinition, error) {
+	if len(caveatNames) == 0 {
+		return nil, nil
+	}
+	return cr.lookupCaveats(ctx, caveatNames)
+}
+
+func (cr *crdbReader) ListAllCaveats(ctx context.Context) ([]*core.CaveatDefinition, error) {
+	return cr.lookupCaveats(ctx, nil)
+}
+
+func (cr *crdbReader) lookupCaveats(ctx context.Context, caveatNames []string) ([]*core.CaveatDefinition, error) {
 	caveatsWithNames := listCaveat
 	if len(caveatNames) > 0 {
 		caveatsWithNames = caveatsWithNames.Where(sq.Eq{colCaveatName: caveatNames})
--- a/internal/datastore/crdb/caveat.go
+++ b/internal/datastore/crdb/caveat.go
@@ -22,7 +22,7 @@ var (
 	)
 	writeCaveat  = psql.Insert(tableCaveat).Columns(colCaveatName, colCaveatDefinition).Suffix(upsertCaveatSuffix)
 	readCaveat   = psql.Select(colCaveatDefinition, colTimestamp).From(tableCaveat)
-	listCaveat   = psql.Select(colCaveatName, colCaveatDefinition).From(tableCaveat).OrderBy(colCaveatName)
+	listCaveat   = psql.Select(colCaveatName, colCaveatDefinition, colTimestamp).From(tableCaveat).OrderBy(colCaveatName)
 	deleteCaveat = psql.Delete(tableCaveat)
 )
 
@@ -62,18 +62,23 @@ func (cr *crdbReader) ReadCaveatByName(ctx context.Context, name string) (*core.
 	return loaded, revisionFromTimestamp(timestamp), nil
 }
 
-func (cr *crdbReader) LookupCaveatsWithNames(ctx context.Context, caveatNames []string) ([]*core.CaveatDefinition, error) {
+func (cr *crdbReader) LookupCaveatsWithNames(ctx context.Context, caveatNames []string) ([]datastore.RevisionedCaveat, error) {
 	if len(caveatNames) == 0 {
 		return nil, nil
 	}
 	return cr.lookupCaveats(ctx, caveatNames)
 }
 
-func (cr *crdbReader) ListAllCaveats(ctx context.Context) ([]*core.CaveatDefinition, error) {
+func (cr *crdbReader) ListAllCaveats(ctx context.Context) ([]datastore.RevisionedCaveat, error) {
 	return cr.lookupCaveats(ctx, nil)
 }
 
-func (cr *crdbReader) lookupCaveats(ctx context.Context, caveatNames []string) ([]*core.CaveatDefinition, error) {
+type bytesAndTimestamp struct {
+	bytes     []byte
+	timestamp time.Time
+}
+
+func (cr *crdbReader) lookupCaveats(ctx context.Context, caveatNames []string) ([]datastore.RevisionedCaveat, error) {
 	caveatsWithNames := listCaveat
 	if len(caveatNames) > 0 {
 		caveatsWithNames = caveatsWithNames.Where(sq.Eq{colCaveatName: caveatNames})
@@ -83,7 +88,8 @@ func (cr *crdbReader) lookupCaveats(ctx context.Context, caveatNames []string) (
 	if err != nil {
 		return nil, fmt.Errorf(errListCaveats, err)
 	}
-	var allDefinitionBytes [][]byte
+	var allDefinitionBytes []bytesAndTimestamp
+
 	err = cr.executeWithTx(ctx, func(ctx context.Context, tx pgx.Tx) error {
 		rows, err := tx.Query(ctx, sql, args...)
 		if err != nil {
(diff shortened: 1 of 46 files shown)


internal/datastore/crdb/caveat.go
```diff
@@ -22,7 +22,7 @@ var (
 	)
 	writeCaveat  = psql.Insert(tableCaveat).Columns(colCaveatName, colCaveatDefinition).Suffix(upsertCaveatSuffix)
 	readCaveat   = psql.Select(colCaveatDefinition, colTimestamp).From(tableCaveat)
-	listCaveat   = psql.Select(colCaveatName, colCaveatDefinition).From(tableCaveat).OrderBy(colCaveatName)
+	listCaveat   = psql.Select(colCaveatName, colCaveatDefinition, colTimestamp).From(tableCaveat).OrderBy(colCaveatName)
```
vroldanbet, line 25: do we need a new index for this? 👀

internal/datastore/crdb/caveat.go
```diff
@@ -107,13 +114,16 @@ func (cr *crdbReader) lookupCaveats(ctx context.Context, caveatNames []string) (
 		return nil, fmt.Errorf(errListCaveats, err)
 	}
 
-	caveats := make([]*core.CaveatDefinition, 0, len(allDefinitionBytes))
-	for _, defBytes := range allDefinitionBytes {
+	caveats := make([]datastore.RevisionedCaveat, 0, len(allDefinitionBytes))
+	for _, bat := range allDefinitionBytes {
```
vroldanbet, line 118: bat?

pkg/datastore/datastore.go
```diff
@@ -196,6 +196,23 @@ func (sf SubjectsFilter) AsSelector() SubjectsSelector {
 	}
 }
 
+// SchemaDefinition represents a namespace or caveat definition under a schema.
+type SchemaDefinition interface {
```
vroldanbet, line 200: might be confusing due to the use of `definition` keyword in the schema language

internal/datastore/mysql/query_builder.go
```diff
@@ -65,7 +65,7 @@ func NewQueryBuilder(driver *migrations.MySQLDriver) *QueryBuilder {
 }
 
 func listCaveats(tableCaveat string) sq.SelectBuilder {
-	return sb.Select(colCaveatDefinition).From(tableCaveat).OrderBy(colName)
+	return sb.Select(colCaveatDefinition, colCreatedTxn).From(tableCaveat).OrderBy(colName)
```
vroldanbet, line 68: do we also need a new index here?

internal/datastore/postgres/caveat.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var (
 	writeCaveat  = psql.Insert(tableCaveat).Columns(colCaveatName, colCaveatDefinition)
-	listCaveat   = psql.Select(colCaveatDefinition).From(tableCaveat).OrderBy(colCaveatName)
+	listCaveat   = psql.Select(colCaveatDefinition, colCreatedXid).From(tableCaveat).OrderBy(colCaveatName)
```
vroldanbet, line 17: index?

internal/datastore/spanner/reader.go
```diff
@@ -149,12 +149,12 @@ func (sr spannerReader) ReadNamespaceByName(ctx context.Context, nsName string)
...
 }
 
-func (sr spannerReader) ListAllNamespaces(ctx context.Context) ([]*core.NamespaceDefinition, error) {
+func (sr spannerReader) ListAllNamespaces(ctx context.Context) ([]datastore.RevisionedNamespace, error) {
 	iter := sr.txSource().Read(
 		ctx,
 		tableNamespace,
 		spanner.AllKeys(),
-		[]string{colNamespaceConfig},
+		[]string{colNamespaceConfig, colNamespaceTS},
```
vroldanbet, line 157: same, do we need to consider an index here, or is this metadata available always for all fetched rows in Spanner?

internal/datastore/spanner/reader.go
```diff
@@ -179,7 +179,7 @@ func (sr spannerReader) LookupNamespacesWithNames(ctx context.Context, nsNames [
 		ctx,
 		tableNamespace,
 		spanner.KeySetFromKeys(keys...),
-		[]string{colNamespaceConfig},
+		[]string{colNamespaceConfig, colNamespaceTS},
```
vroldanbet, line 182: same, do we need to consider an index here?

internal/datastore/spanner/stats.go
```diff
@@ -35,7 +35,7 @@ func (sd spannerDatastore) Statistics(ctx context.Context) (datastore.Stats, err
 		ctx,
 		tableNamespace,
 		spanner.AllKeys(),
-		[]string{colNamespaceConfig},
+		[]string{colNamespaceConfig, colNamespaceTS},
```
vroldanbet, line 38: stats probably not that important, but for large instances this query could have impact capacity used-wise

internal/datastore/proxy/caching.go
```diff
@@ -36,72 +43,181 @@ func DatastoreProxyTestCache(t testing.TB) cache.Cache {
...
+
+func (r *definitionCachingReader) LookupCaveatsWithNames(
+	ctx context.Context,
+	caveatNames []string,
+) ([]datastore.RevisionedCaveat, error) {
+	return listAndCache[*core.CaveatDefinition](ctx, r, "caveat", caveatNames,
+		func(ctx context.Context, names []string) ([]datastore.RevisionedCaveat, error) {
+			return r.Reader.LookupCaveatsWithNames(ctx, names)
+		},
+		estimatedNamespaceDefinitionSize)
```
vroldanbet, line 136: it's using the wrong function: `estimatedCaveatDefinitionSize`

internal/datastore/proxy/caching.go
```diff
@@ -36,72 +43,181 @@ func DatastoreProxyTestCache(t testing.TB) cache.Cache {
...
 
-func (r *nsCachingReader) ReadNamespaceByName(
+func (r *definitionCachingReader) ReadNamespaceByName(
 	ctx context.Context,
-	nsName string,
+	name string,
 ) (*core.NamespaceDefinition, datastore.Revision, error) {
-	// Check the nsCache.
-	nsRevisionKey := nsName + "@" + r.rev.String()
+	return readAndCache[*core.NamespaceDefinition](ctx, r, "namespace", name,
```
vroldanbet, line 99: do we need the whole `namespace` prefix or can we go with `n`

internal/datastore/proxy/caching.go
```diff
@@ -36,72 +43,181 @@ func DatastoreProxyTestCache(t testing.TB) cache.Cache {
...
+			return r.Reader.LookupNamespacesWithNames(ctx, names)
+		},
+		estimatedNamespaceDefinitionSize)
+}
+
+func (r *definitionCachingReader) ReadCaveatByName(
+	ctx context.Context,
+	name string,
+) (*core.CaveatDefinition, datastore.Revision, error) {
+	return readAndCache[*core.CaveatDefinition](ctx, r, "caveat", name,
```
vroldanbet, line 121: do we need the whole `caveat` prefix or can we reduce memory usage with `c`

internal/datastore/crdb/caveat.go
josephschorr: I don't believe so; the name is the primary key and we're only selecting by name in the hot path (ListCaveats is returning everything)

internal/datastore/crdb/caveat.go
josephschorr: bytes and timestamp

pkg/datastore/datastore.go
josephschorr: yeah, but this is already the terminology I use in the schemadsl package; this is just repeated here.

internal/datastore/mysql/query_builder.go
josephschorr: I don't believe so, no

internal/datastore/postgres/caveat.go
josephschorr: See above

internal/datastore/spanner/reader.go
josephschorr: It should be available for all of them, but I'm not 100% certain about Spanner. @jakedt any thoughts?

internal/datastore/spanner/reader.go
josephschorr: see above

internal/datastore/spanner/stats.go
josephschorr: how would it be that different from the previous one?

internal/datastore/proxy/caching.go
josephschorr: good catch. fixed

internal/datastore/proxy/caching.go
josephschorr: Changed

internal/datastore/proxy/caching.go
josephschorr: changed

internal/datastore/crdb/caveat.go
vroldanbet: but then it won't be a covering index and will require access to disk - just want to make sure are being intentional about this

internal/datastore/mysql/query_builder.go
vroldanbet: potential concern is that this won't be a covering index anymore if it's not part of an existing index

internal/datastore/spanner/stats.go
vroldanbet: if there is an index, it shouldn't, but it may lead to an additional disk access if the request is not being served from the index anymore

vroldanbet (dismissed):

internal/datastore/mysql/query_builder.go
josephschorr: Possible, but I think that's already likely since the devision itself is not in the index

vroldanbet (approved):

Issue #1126, opened 2023-01-30
Unable to cleanly shut down SpiceDB process with REST gateway running


Meai:
On a nearly completely empty mysql+spicedb where I only added the sample schema with nothing else, I'm getting a ton of errors (like 50 per second) and it wont turn off properly anymore until I kill the process:

 ```
3:55PM WRN error attempting to perform garbage collection error="sql: Scan error on column index 0, name \"UTC_TIMESTAMP(6)\": unsupported Scan, storing driver.Value type []uint8 into type *time.Time" next-attempt-in=-0.000001
```

trying to shut down:

```
4:02PM INF shutting down datastore garbage collection worker
4:02PM ERR error waiting for garbage collector to shutdown error="context canceled"
4:02PM ERR error waiting for garbage collector to shutdown error="context canceled"
```

the shutdown problem even happens when i try to shut down immediately, before the scan errors happen.
I'm now trying adding ?parseTime=true to the connection string but I can already say that shutting down via ctrl+c still doesn't work, even after adding that. I saw that advice [here on stackoverflow](https://stackoverflow.com/questions/45040319/unsupported-scan-storing-driver-value-type-uint8-into-type-time-time)

**Comment by josephschorr** [comment]
@Meai What version of MySQL are you running against? Can you add `--log-level=trace` and retrieve the logs for the failure?

**Comment by Meai** [comment]
> @Meai What version of MySQL are you running against? Can you add `--log-level=trace` and retrieve the logs for the failure?

mysqld 8.0.31

for the shutdown via ctrl+c I dont see any new logs, it just always hangs forever. 
here are the logs with tracing enabled and no ?parseTime=true in the connectionstring:


```
7:47PM INF finished call grpc.code=OK grpc.component=server grpc.method=CheckPermission grpc.method_type=unary grpc.service=authzed.api.v1.PermissionsService grpc.start_time=2023-01-30T19:47:38+01:00 grpc.time_ms=0.214 peer.address=127.0.0.1:58040 protocol=grpc requestID=a60128bb01a5878eac8e3bbdbd3c6fa4
7:47PM DBG returning cached revision now=2023-01-30T19:47:39+01:00 requestID=a3182e1c23369dfbb79914a3d46f3680 valid=2023-01-30T19:47:40+01:00
7:47PM TRC adding sample duration to statistics duration=0.050666 requestID=a3182e1c23369dfbb79914a3d46f3680
[...]

PR #1129, opened 2023-01-30
Add an explicit error if parseTime is missing from MySQL URIs

josephschorr:
This is necessary for garbage collection and other time-based work, so make it required on construction

Related: #1126

### DIFF
--- a/internal/datastore/mysql/datastore.go
+++ b/internal/datastore/mysql/datastore.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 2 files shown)


josephschorr:
Depends on https://github.com/authzed/spicedb-short-urls/pull/2

vroldanbet (commented):
LGTM, would it be reasonable to add a test?

josephschorr:
I'll add one although we know it is testing successfully otherwise

vroldanbet (approved):

PR #1140, opened 2023-02-01
Fix flakiness of cert test

josephschorr:
Extend the expiration on certs in cert test to reduce flakiness


ecordell (approved):
LGTM

this seems worth a shot

PR #1141, opened 2023-02-02
Add release process for dev builds (nightlies)

ecordell:
This will build spicedb images on every push to main. 

Images will be pushed to `authzed/spicedb-git` repos with tags like `v<currentVersion + 0.0.1>-<shortsha>`

For example, running against main would publish:

```
ghcr.io/authzed/spicedb-git:v1.16.2-a3424cda
ghcr.io/authzed/spicedb-git:v1.16.2-a3424cda-debug
ghcr.io/authzed/spicedb-git:latest
ghcr.io/authzed/spicedb-git:latest-debug
```
(and equivalents in quay and docker hub)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -28,7 +28,6 @@ nfpms:
 furies:
   - account: "authzed"
     secret_name: "GEMFURY_PUSH_TOKEN"
-
 brews:
   - tap:
       owner: "authzed"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -110,35 +109,35 @@ docker_manifests:
   # Quay
   - name_template: "quay.io/authzed/spicedb:v{{ .Version }}"
     image_templates: [*amd_image_quay, *arm_image_quay]
-  - name_template: "quay.io/authzed/spicedb:latest"
+  - name_template: "quay.io/authzed/spicedb:{{ if .IsNightly }}devel{{ else }}latest{{ end }}"
```
vroldanbet, line 112: could we unintentionally push a development container as `latest` if this condition does not work?

ecordell: Yes; unfortunately I don't see a great way to test it (without publishing) since it will only build and tag the multi-arch manifest once it has pushed the other images. 

But it is verbatim from the goreleaser docs, see the bottom of this page: https://goreleaser.com/customization/nightlies/#what-is-skipped-when-using-nightly

vroldanbet (dismissed):

.goreleaser.yml
josephschorr: Yeah, I had the same question and concern, but if this is what they suggest, I guess its okay

jzelinskie (approved):

PR #1147, opened 2023-02-03
.github: use 8 cores for codeql

jzelinskie:
According to the documentation it should detect the number of threads and parallelize.


josephschorr (approved):
LGTM

PR #1149, opened 2023-02-07
pkg/cache: implement a central collector

jzelinskie:
This commit adds a Prometheus collector that records metrics for all allocated caches.

This fixes a bug where we weren't getting metrics on anything other than the the dispatch because it was manually being recorded in the caching dispatcher.

This changes the default to collect metrics which is the common case as the overhead is typically worth the added observability.

### DIFF
--- a/pkg/cache/cache_ristretto.go
+++ b/pkg/cache/cache_ristretto.go
@@ -13,25 +13,41 @@ import (
 	"github.com/authzed/spicedb/internal/dispatch/keys"
 )
 
-// NewCache creates a new ristretto cache from the given config.
-func NewCache(config *Config) (Cache, error) {
-	cache, err := ristretto.NewCache(&ristretto.Config{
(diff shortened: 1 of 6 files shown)


pkg/cache/cache_ristretto.go
```diff
@@ -13,25 +13,47 @@ import (
...
+	}
+
+	cache := wrapped{config, config.DefaultTTL, rcache}
+	mustRegisterCache(name, cache)
+	return cache, nil
+}
+
+// NewCache creates a new ristretto cache from the given config.
+func NewCache(config *Config) (Cache, error) {
+	cfg := ristrettoConfig(config)
```
vroldanbet, line 49: nit: with could refactor this method as the one above is very similar

vroldanbet (dismissed):

vroldanbet (approved):

PR #1152, opened 2023-02-08
make cache collector unregister on close

vroldanbet:
in cases where we need to test instantiating servers using `RunnableServer`, we have no control to unregister cache metrics, so global state is left behind, potentially causing failures in other tests

### DIFF
--- a/pkg/cache/cache_ristretto.go
+++ b/pkg/cache/cache_ristretto.go
@@ -39,7 +39,7 @@ func NewCacheWithMetrics(name string, config *Config) (Cache, error) {
 		return nil, err
 	}
 
-	cache := wrapped{config, config.DefaultTTL, rcache}
+	cache := wrapped{name, config, config.DefaultTTL, rcache}
 	mustRegisterCache(name, cache)
 	return cache, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/cache/cache_ristretto.go
```diff
@@ -67,3 +68,8 @@ var _ Cache = (*wrapped)(nil)
 
 func (w wrapped) GetMetrics() Metrics                   { return w.Cache.Metrics }
 func (w wrapped) MarshalZerologObject(e *zerolog.Event) { e.EmbedObject(w.config) }
+
+func (w wrapped) Close() {
```
jzelinskie, line 72: Don't we need to add the Close method to the Cache interface and call it in dispatcher?

vroldanbet: `wrapped` already implements ristretto `Cache` interface, which also has the `Close` method.
Ristretto Cache was already being closed through the `CachingDispatcher`

jzelinskie: oh i totally missed that there was already a close lmao

PR #1154, opened 2023-02-09
Mark caveats as a production-ready feature

josephschorr:
Also adds the bulk lookup for caveats in the expression evaluator

### DIFF
--- a/internal/caveats/run.go
+++ b/internal/caveats/run.go
@@ -13,6 +13,7 @@ import (
 	"github.com/authzed/spicedb/pkg/datastore"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/spiceerrors"
+	"github.com/authzed/spicedb/pkg/util"
 )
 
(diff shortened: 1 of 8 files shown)


internal/caveats/run.go
```diff
@@ -98,13 +99,67 @@ func runExpression(
...
+	if ok {
+		return caveat, deserialized, nil
+	}
+
+	deserialized, err := caveats.DeserializeCaveat(caveat.SerializedExpression)
+	if err != nil {
+		return caveat, nil, err
+	}
+
+	lc.deserializedCaveats[caveatDefName] = deserialized
```
vroldanbet, line 149: the assumption is that this won't get modified concurrently because the caveat tree is executed serially, right?

vroldanbet (commented):
shouldn't a test have failed because of the gating being removed? 🤔

internal/caveats/run.go
josephschorr: Correct

josephschorr:
> shouldn't a test have failed because of the gating being removed? 🤔

No, we didn't have a test for the gating

vroldanbet (approved):

Issue #1158, opened 2023-02-09
CockroachDB GC Window Change


chancesm:
The default datastore gc window in SpiceDB is 24 hours. This worked well as Cockroach's default gc window was 25 hours. When the datastore gc window is less than what spicedb is configured to use, an error occurs. CockroachDB is planning on dropping the default GC window down to 4 hours for their dedicated clusters (1.25 hours for serverless). This change will cause Spicedb to error out as the default values will end up conflicting. The change is documented here: https://github.com/cockroachdb/cockroach/issues/89233

Suggested solution: Change Spicedb's default datastore gc window to a value that matches one of CockroachDB's new default values (4 hours or 1.25 hours) so that new adopters do not run into issues when exploring Spicedb for the first time.

vroldanbet:
Unfortunately, we cannot change the default GC value as it would be a breaking change to SpiceDB deployments where client applications rely on `at_exact_snapshot` or `Watch` API. We added a new section in our documentation and referenced it in the logged error message.

Issue #606, opened 2022-05-18
Datastore-specific defaults


jakedt:
Right now in the `pkg/cmd/datastore` package we have a set of defaults that apply equally to all datastores. Sometimes this is not a single default that works for all datastores. For example, `spanner` is limited to 100 usersets in the query splittler based on the structural limitations of the datastore. All other implementations can split at ~1k usersets before they start running into query size constraints.

It would make sense to allow some options to let the datastore pick its own default. We should audit our datastore flags to figure out other cases where this may make sense, and then update the `pkg/cmd/datastore` package accordingly.

fearlessfe:
how to do this?

set the default SplitQueryCount to 0,  and the in  newPostgresDatastore or  newSpannerDatastore ... func, 
check whethe the SplitQueryCount is 0. if true, set the default value, else use the SplitQueryCount

miparnisari:
[PR #2867](https://github.com/authzed/spicedb/pull/2867) partially addresses this for the Watch API options. It adds a `DefaultsWatchOptions()` method on the `Datastore` interface so each implementation declares its own defaults for `CheckpointInterval`, `WatchBufferLength`, etc. and removes the per-datastore custom flags from  `kg/cmd/datastore/datastore.go`. The merge precedence lives in `pkg/datalayer/watchoptions.go`. 

The same pattern could be extended to other flags by adding similar `Defaults*()` methods to the `Datastore` interface and having the `DataLayer` object call that method.

(The downside of that PR is that anyone calling `datastore.Watch` directly instead of `datalayer.Watch` will have to be careful to pass the correct value for each setting. `datastore.Watch` now assumes they are valid.)

PR #1161, opened 2023-02-10
makes default GC window 1h

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/1158

see https://github.com/authzed/spicedb/issues/1158 see
https://github.com/cockroachdb/cockroach/issues/89233

Cockroach is changing their default to 4h in dedicated and 1.25h in Serverless.

This commit changes the lowest common denominator, sans some padding,
to avoid problems with folks running with CRDB and not being aware of those settings.

### DIFF
--- a/internal/datastore/crdb/options.go
+++ b/internal/datastore/crdb/options.go
@@ -192,7 +192,7 @@ func MaxRevisionStalenessPercent(stalenessPercent float64) Option {
 // GCWindow is the maximum age of a passed revision that will be considered
 // valid.
 //
-// This value defaults to 24 hours.
+// This value defaults to 1 hour.
 func GCWindow(window time.Duration) Option {
 	return func(po *crdbOptions) {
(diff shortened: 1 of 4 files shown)


ecordell:
Since there are different values for cockroach dedicated vs. serverless, maybe it would make sense to have a different default based on cockroach's setting?

vroldanbet:
>Since there are different GC windows for cockroach dedicated vs. serverless, maybe it would make sense to have a different default based on cockroach's setting?

It would, although given it's all handled via `cobra`, it requires the parsing to take place in order to understand which default to set. It wasn't immediately obvious, and thought that perhaps GC=1h may be even positive (busy databases 24h may be not enough and cause one large GC that times out).

josephschorr:
We should't change this for the other datastores. Rather, we should use this as an opportunity to do this: https://github.com/authzed/spicedb/issues/606

PR #1166, opened 2023-02-13
improve CRDB GC error message

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/1158

unfortunately we cannot change the default CRDB GC value as that would be a breaking change for deployments relying heavily on `at_exact_snapshot` request consistency and `Watch` API.

- For users of Cockroach dedicated, the value won't change automatically. Newly created clusters will have a GC of 4 hours, but users can change it 
- For users of Cockroach serverless the value will be automatically updated by the control plane to 1h15m - see [release notes](https://www.cockroachlabs.com/docs/releases/cloud.html#general-changes)


josephschorr (approved):
LGTM

PR #1169, opened 2023-02-15
does not return an error if GC windows aren't aligned

vroldanbet:
so that spicedbs dont crashloop if cockroach
changes Cockroach Serverless / Dedicated GC


josephschorr (dismissed):
LGTM

josephschorr (approved):

PR #1176, opened 2023-02-20
Skip loading of head revision on write calls

josephschorr:
These APIs do not use consistency blocks, so there is no need to load the head revision outside of their transactions

### DIFF
--- a/internal/services/v1/schema.go
+++ b/internal/services/v1/schema.go
@@ -10,7 +10,6 @@ import (
 
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/internal/middleware"
-	"github.com/authzed/spicedb/internal/middleware/consistency"
 	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
 	"github.com/authzed/spicedb/internal/middleware/usagemetrics"
 	"github.com/authzed/spicedb/internal/services/shared"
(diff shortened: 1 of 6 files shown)


pkg/middleware/consistency/consistency.go
```diff
@@ -16,13 +15,8 @@ func RevisionFromContext(ctx context.Context) datastore.Revision {
...
-// MustRevisionFromContext reads the selected revision out of a context.Context, computes a zedtoken
-// from it, and panics if it has not been set on the context.
-func MustRevisionFromContext(ctx context.Context) (datastore.Revision, *v1.ZedToken) {
-	rev := consistency.RevisionFromContext(ctx)
-	if rev == nil {
-		panic("consistency middleware did not inject revision")
-	}
-
-	return rev, zedtoken.MustNewFromRevision(rev)
+// RevisionFromContextOrError reads the selected revision out of a context.Context, computes a zedtoken
```
jzelinskie, line 18: same nitpick

jzelinskie (changes requested):

internal/services/v1/schema.go
```diff
@@ -46,15 +45,21 @@ type schemaServer struct {
 }
 
 func (ss *schemaServer) ReadSchema(ctx context.Context, in *v1.ReadSchemaRequest) (*v1.ReadSchemaResponse, error) {
-	readRevision, _ := consistency.MustRevisionFromContext(ctx)
-	ds := datastoremw.MustFromContext(ctx).SnapshotReader(readRevision)
+	// Schema is already read from the head revision.
```
vroldanbet, line 48: hrm, this comment assumes the reader has context about the previous implementation. I'd clarify that this method does not have consistency specified and thus we need to explicitly get `head` revision as it won't be available in the context.

internal/services/v1/schema.go
```diff
@@ -46,15 +45,21 @@ type schemaServer struct {
...
+	ds := datastoremw.MustFromContext(ctx)
+	headRevision, err := ds.HeadRevision(ctx)
+	if err != nil {
+		return nil, rewriteError(ctx, err)
+	}
+
+	reader := ds.SnapshotReader(headRevision)
```
vroldanbet, line 55: a potential optimization would be to have a method like `HeadReader` which executes the first query and returning the revision, and uses that revision for any potential subsequent queries. That way we avoid one roundtrip

vroldanbet (approved):
Other than jimmy's suggestion on the naming, this looks good to me

internal/services/v1/schema.go
josephschorr: Sorry; should have said "always". I've fixed

internal/services/v1/schema.go
josephschorr: Not sure how it avoids a roundtrip; `SnapshotReader(...)` should not be making a DB call until its actually used

josephschorr:
Rebased and updated

vroldanbet (approved):

jzelinskie (approved):

PR #1179, opened 2023-02-23
internal/datastore/crdb: split read/write connpools

jzelinskie:
Because CRDB and Postgres have a stateful protocol that consumes a connection while awaiting a response and writes can starve readers for connections from the pool when there is sufficient load on SpiceDB. 

This pull request splits out reads and writes for CockroachDB and PostgreSQL into two different connection pools. This should cause write latency to only impact writes rather than both reads and writes.


internal/datastore/crdb/crdb.go
```diff
@@ -74,23 +74,30 @@ func newCRDBDatastore(url string, options ...Option) (datastore.Datastore, error
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
 
-	poolConfig, err := pgxpool.ParseConfig(url)
+	readPoolConfig, err := pgxpool.ParseConfig(url)
 	if err != nil {
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
+	writePoolConfig := readPoolConfig.Copy()
```
ecordell, line 81: when I was testing locally, it seemed like this didn't create a copy with all of the values. I had to `pgxpool.ParseConfig(url)` again to get a working config.

pkg/cmd/datastore/datastore.go
```diff
@@ -111,11 +152,13 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
...
-	flagSet.IntVar(&opts.MinOpenConns, flagName("datastore-conn-min-open"), defaults.MinOpenConns, "number of minimum concurrent connections open in a remote datastore's connection pool")
-	flagSet.DurationVar(&opts.MaxLifetime, flagName("datastore-conn-max-lifetime"), defaults.MaxLifetime, "maximum amount of time a connection can live in a remote datastore's connection pool")
-	flagSet.DurationVar(&opts.MaxIdleTime, flagName("datastore-conn-max-idletime"), defaults.MaxIdleTime, "maximum amount of time a connection can idle in a remote datastore's connection pool")
-	flagSet.DurationVar(&opts.HealthCheckPeriod, flagName("datastore-conn-healthcheck-interval"), defaults.HealthCheckPeriod, "time between a remote datastore's connection pool health checks")
+
+	var legacyConnPool ConnPoolConfig
+	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-conn", DefaultConnPoolConfig(), &legacyConnPool)
+	deprecateUnifiedConnFlags(flagSet)
+	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-connpool-read", &legacyConnPool, &opts.ReadConnPool)
```
vroldanbet, line 159: why are we using `legacyConnPool` here as the default?

pkg/cmd/datastore/datastore.go
```diff
@@ -337,9 +387,9 @@ func newMySQLDatastore(opts Config) (datastore.Datastore, error) {
 		mysql.GCInterval(opts.GCInterval),
 		mysql.GCEnabled(!opts.ReadOnly),
 		mysql.GCMaxOperationTime(opts.GCMaxOperationTime),
-		mysql.ConnMaxIdleTime(opts.MaxIdleTime),
-		mysql.ConnMaxLifetime(opts.MaxLifetime),
-		mysql.MaxOpenConns(opts.MaxOpenConns),
+		mysql.MaxOpenConns(opts.ReadConnPool.MaxOpenConns),
```
vroldanbet, line 390: I think it could be argued some folks may consider setting the `WriteConnPool` parameter instead for those databases that do not support 2 conn pools. As a consequence, the user may think the flag is properly set, when it's in fact being ignored.

Ideally the application makes it difficult to make a mistake here.

pkg/cmd/datastore/datastore.go
```diff
@@ -279,11 +319,16 @@ func newCRDBDatastore(opts Config) (datastore.Datastore, error) {
...
-		crdb.ConnHealthCheckInterval(opts.HealthCheckPeriod),
-		crdb.MaxOpenConns(opts.MaxOpenConns),
-		crdb.MinOpenConns(opts.MinOpenConns),
+		crdb.MaxOpenReadConns(opts.ReadConnPool.MaxOpenConns),
+		crdb.MinOpenReadConns(opts.ReadConnPool.MinOpenConns),
+		crdb.ReadConnMaxIdleTime(opts.ReadConnPool.MaxIdleTime),
+		crdb.ReadConnMaxLifetime(opts.ReadConnPool.MaxLifetime),
+		crdb.ReadConnHealthCheckInterval(opts.ReadConnPool.HealthCheckInterval),
+		crdb.MaxOpenWriteConns(opts.WriteConnPool.MaxOpenConns),
+		crdb.MinOpenWriteConns(opts.WriteConnPool.MinOpenConns),
```
vroldanbet, line 328: perhaps moving the `Write` world to the beginning of the method name, so this method looks just like the other `Write*` methods? Feels weird a different naming convention was chosen for these.

internal/datastore/postgres/stats.go
```diff
@@ -45,7 +45,7 @@ func (pgd *pgDatastore) Statistics(ctx context.Context) (datastore.Stats, error)
 	var uniqueID string
 	var nsDefs []datastore.RevisionedNamespace
 	var relCount int64
-	if err := pgd.dbpool.BeginTxFunc(ctx, pgd.readTxOptions, func(tx pgx.Tx) error {
+	if err := pgd.readPool.BeginTxFunc(ctx, pgd.readTxOptions, func(tx pgx.Tx) error {
```
vroldanbet, line 48: I feel like the reason conn pool where split are less about "read and write" and more about "fast TX / slow TX". If our transactions in CRDB were faster, we wouldn't even need to do this split in the first place.

For the sake of the argument, this transaction here - if I recall correctly - is rather slow, and could put the same kind of contention over the read-connection pool that write TX did.

internal/datastore/postgres/options.go
```diff
@@ -107,52 +106,112 @@ func SplitAtUsersetCount(splitAtUsersetCount uint16) Option {
...
+// MaxOpenReadConns is the maximum size of the connection pool used for reads.
+//
+// This value defaults to having no maximum.
+func MaxOpenReadConns(conns int) Option {
+	return func(po *postgresOptions) { po.readPoolOpts.MaxOpenConns = &conns }
+}
+
+// MaxOpenWriteConns is the maximum size of the connection pool used for writes.
+//
+// This value defaults to having no maximum.
```
vroldanbet, line 212: Is this correct? I see defaults set in the CLI to:

```go
func DefaultConnPoolConfig() *ConnPoolConfig {
	return &ConnPoolConfig{
		MaxLifetime:         30 * time.Minute,
		MaxIdleTime:         30 * time.Minute,
		MaxOpenConns:        20,
		MinOpenConns:        10,
		HealthCheckInterval: 30 * time.Second,
	}
}
```

ecordell: ^ yes, it always defaults to 20, noticed this when I was testing it out

pkg/cmd/datastore/datastore.go
jzelinskie: Basically to support mixing and matching flags. The new flags prioritize over the old ones, so if you have `--datastore-conn-healthcheck-interval` and `--datastore-connpool-read-healthcheck-interval`, the latter takes precedent over the former.

pkg/cmd/datastore/datastore.go
jzelinskie: I agree -- especially since this PR doesn't split the pools for all of the datastores.

What's the precedent? We've got some other flags that are CRDB specific, right? What are we doing there?
Let's just be consistent in this PR and if we have to clean up that behavior, do everything at once in another PR.

pkg/cmd/datastore/datastore.go
jzelinskie: I tried to make it `(Read|Write)Conns` everywhere.

pkg/cmd/datastore/datastore.go
vroldanbet: let's assume a user wants to move to use MySQL with the new connection flag, because the old ones are deprecated. They may think they have to set the `write` flags, when in fact they have to set the `read` flags. Does that make sense?

pkg/cmd/datastore/datastore.go
vroldanbet: right except these 2 methods do not follow the convention

internal/datastore/postgres/options.go
jzelinskie: What do we want these defaults to be? I think write pools should always default to smaller than the read pools.
It could also make sense that these be different based on the datastore.

internal/datastore/postgres/stats.go
jzelinskie: > If our transactions in CRDB were faster, we wouldn't even need to do this split in the first place.

I don't think that this is true. Regardless of how fast writes could be, they'll still always be in a different order of magnitude than read queries. There will always be some scale where you'll end up seeing the writes contend reads for connections. This PR acknowledges that reality.

The ultimate fix for this issue is to use a transport that allows for multiplexing queries over a single connection (e.g. gRPC). The problem is that many SQL databases have stateful connection sessions, so they cannot reuse the connection until they've gotten a response.

internal/datastore/postgres/options.go
jzelinskie: Sorry -- I hijacked this with a different question

If you don't use this option to configure the datastore, it defaults to having no maximum.

jzelinskie: The write pool now defaults to half the read pool.

pkg/cmd/datastore/datastore.go
jzelinskie: Because MySQL doesn't use pgx, let's get this merged for now and we can work on getting parity to MySQL before the next release.

internal/datastore/crdb/crdb.go
```diff
@@ -380,7 +372,7 @@ func (cds *crdbDatastore) Features(ctx context.Context) (*datastore.Features, er
 	streamCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	time.AfterFunc(1*time.Second, cancel)
-	_, err = cds.pool.Exec(streamCtx, fmt.Sprintf(cds.beginChangefeedQuery, tableTuple, head))
+	_, err = cds.readPool.Exec(streamCtx, fmt.Sprintf(cds.beginChangefeedQuery, tableTuple, head))
```
ecordell, line 375: I think this is arguably more like a `write`

vroldanbet (approved):

internal/datastore/postgres/common/pgx.go
```diff
@@ -100,10 +100,48 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
...
+}
+
+// ConfigurePgx applies PoolOptions to a pgx connection pool confiugration.
+func (opts PoolOptions) ConfigurePgx(pgxConfig *pgxpool.Config) {
+	if opts.MaxOpenConns != nil {
+		pgxConfig.MaxConns = int32(*opts.MaxOpenConns)
+	}
+
+	if opts.MinOpenConns != nil {
+		pgxConfig.MinConns = int32(*opts.MinOpenConns)
```
ecordell, line 127: for crdb, maybe just default this to == max conns?

ecordell (approved):
LGTM

PR #1190, opened 2023-03-07
makes the request caveat context size configurable

vroldanbet:
makes it possible to configure the caveat context
size enforcement. This is useful when some
schemas may end up hitting this limit easily, but
we still need to have a limit in place to prevent abuse.

### DIFF
--- a/internal/services/v1/permissions_test.go
+++ b/internal/services/v1/permissions_test.go
@@ -1370,3 +1370,19 @@ func randString(length int) string {
 	}
 	return string(b)
 }
+
+func TestGetCaveatContext(t *testing.T) {
+	strct, err := structpb.NewStruct(map[string]any{"foo": "bar"})
+	require.NoError(t, err)
+
+	_, err = v1svc.GetCaveatContext(context.Background(), strct, 1)
+	require.ErrorContains(t, err, "request caveat context should have less than 1 bytes")
+
+	caveatMap, err := v1svc.GetCaveatContext(context.Background(), strct, 0)
+	require.NoError(t, err)
+	require.Contains(t, caveatMap, "foo")
+
+	caveatMap, err = v1svc.GetCaveatContext(context.Background(), strct, -1)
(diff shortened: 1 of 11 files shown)


pkg/cmd/serve.go
```diff
@@ -111,6 +111,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	cmd.Flags().BoolVar(&config.DisableVersionResponse, "disable-version-response", false, "disables version response support in the API")
 	cmd.Flags().Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	cmd.Flags().Uint16Var(&config.MaximumPreconditionCount, "update-relationships-max-preconditions-per-call", 1000, "maximum number of preconditions allowed for WriteRelationships and DeleteRelationships calls")
+	cmd.Flags().IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes")
```
josephschorr, line 114: document that `0` is no limit

internal/services/v1/permissions_test.go
```diff
@@ -1370,3 +1370,15 @@ func randString(length int) string {
...
 }
+
+func TestGetCaveatContext(t *testing.T) {
+	strct, err := structpb.NewStruct(map[string]any{"foo": "bar"})
+	require.NoError(t, err)
+
+	_, err = v1svc.GetCaveatContext(context.Background(), strct, 1)
+	require.ErrorContains(t, err, "request caveat context should have less than 1 bytes")
+
+	caveatMap, err := v1svc.GetCaveatContext(context.Background(), strct, 0)
```
josephschorr, line 1381: add a test with negative value

josephschorr (approved):
LGTM

PR #1195, opened 2023-03-08
CRDB retry backoff

ecordell:
Adds exponential backoff to tx retries for CRDB.

This is recommended by CRDB docs, but isn't implemented in `cockroach-go` (which we don't use). In theory this should help under heavy contention, but I would like to test it some more before merging.


internal/datastore/crdb/tx.go
```diff
@@ -61,10 +64,17 @@ func executeWithResets(ctx context.Context, fn innerFunc, maxRetries uint8) (err
...
+			log.Ctx(ctx).Warn().Err(err).Dur("after", after).Msg("retrying resetteable database error")
+			time.Sleep(after)
 			continue
 		}
-
+		if err != nil {
+			log.Ctx(ctx).Warn().Err(err).Msg("non-resettable database error")
+		}
+		if retries > 0 && err == nil {
+			log.Ctx(ctx).Warn().Uint8("retries", retries).Msg("resettable database error succeeded after retry")
```
vroldanbet, line 76: we may want to make this `info` level?

internal/datastore/crdb/tx.go
```diff
@@ -61,10 +64,17 @@ func executeWithResets(ctx context.Context, fn innerFunc, maxRetries uint8) (err
...
 		if resettable(ctx, err) {
-			log.Ctx(ctx).Warn().Err(err).Msg("retrying resetteable database error")
+			after := retry.BackoffExponentialWithJitter(100*time.Millisecond, 0.5)(uint(retries))
+			log.Ctx(ctx).Warn().Err(err).Dur("after", after).Msg("retrying resetteable database error")
+			time.Sleep(after)
 			continue
 		}
-
+		if err != nil {
+			log.Ctx(ctx).Warn().Err(err).Msg("non-resettable database error")
```
vroldanbet, line 73: do we need to log this? won't the error propagate up the stack and eventually get logged? maybe we can wrap the error with additional context if we want to add that information

internal/datastore/crdb/tx.go
```diff
@@ -61,10 +64,17 @@ func executeWithResets(ctx context.Context, fn innerFunc, maxRetries uint8) (err
 	for retries = 0; retries <= maxRetries; retries++ {
 		err = fn(ctx)
 		if resettable(ctx, err) {
-			log.Ctx(ctx).Warn().Err(err).Msg("retrying resetteable database error")
+			after := retry.BackoffExponentialWithJitter(100*time.Millisecond, 0.5)(uint(retries))
```
vroldanbet, line 67: Once this lands, I'd suggest opening an issue to make this parameterizable.

josephschorr: +1

vroldanbet (approved):

PR #1197, opened 2023-03-09
add request-scoped overlap key option for crdb

### LINKED ISSUE #1193: Support client-provided tenant ID for CRDB tx overlap key
The CRDB datastore [currently supports](https://authzed.com/docs/spicedb/selecting-a-datastore#configuration) 3 strategies for ensuring ZedTokens "work" by ensuring writes include overlapping keys ([blog](https://authzed.com/blog/prevent-newenemy-cockroachdb/)). However, the `secure` strategy effectively forces all write transactions to be processed serially within the datastore, which can be a significant throughput limitation for use cases like a data backfill into SpiceDB and limits horizontal scalability (more CRDB nodes won't help if all writes are bottlenecked on the node owning the "static" overlap row).

The prefix strategy is a great solution for schemas that can use differently-prefixed types for different tenants, but my use case is a multi-tenant application where it is not straightforward to use different SpiceDB types (prefixes) for each tenant. However, I do know the tenant ID at the time I'm making a `WriteRelationships` request, and (in my use case) strong consistency for writes in different tenants is usually not required.

I propose that SpiceDB provide some mechanism for applications to communicate a tenant id / consistency "domain" to SpiceDB (through an API parameter / request header / data relationship / ...) such that all writes within a tenant result in overlapping writes in CRDB (so that ZedTokens have the expected semantics), but writes for different tenants' data don't necessarily overlap.

ecordell:
This adds a new option to `--datastore-tx-overlap-strategy` called `request`.

When the `request` overlap strategy is used, transactions against CRDB are only forced to overlap if it is requested by a user. 

Requests can be forced to overlap by providing explicit overlap keys in the request metadata: 

```golang
ctx := metadata.NewOutgoingContext(context.Background(), metadata.New(map[string]string{
	"spicedb-overlap": "protected",
}))
_, err := client().V1().Permissions().WriteRelationships(ctx, &v1.WriteRelationshipsRequest{
	Updates: []*v1.RelationshipUpdate{...},
})
_, err := client().V1().Permissions().WriteRelationships(ctx, &v1.WriteRelationshipsRequest{
	Updates: []*v1.RelationshipUpdate{...},
})
```

Requests can also have multiple overlap keys:
```golang
metadata.New(map[string]string{
	"spicedb-overlap": "document-move,user-removed",
}
```
Any concurrent request with any matching key will be properly serialized by Cockroach.

This can significantly increase write throughput for Cockroach if the majority of requests do not require new-enemy protection. The `static` overlap strategy remains the default, so that users that have not spent the time to understand when and where a new enemy could occur for their specific applications remain protected.

The newenemy e2e test has also been updated to use this instead of `static`, to show that this mechanism can be used to protect specific workflows without paying the cost for every write.

See this [blog post](https://authzed.com/blog/prevent-newenemy-cockroachdb/) for more background on why this is necessary in the first place.

Fixes #1193


jzelinskie:
@ecordell do we think this still relevant and worth maintaining long-term?

internal/datastore/crdb/keys.go
```diff
@@ -49,3 +54,22 @@ func prefix(s string) string {
...
+
+// overlapKeysFromContext reads the request-provided initial overlap key set
+// from the grpc request metadata.
+func overlapKeysFromContext(ctx context.Context) keySet {
+	keys := newKeySet(ctx)
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return keys
+	}
+	for _, keyList := range md["spicedb-overlap"] {
```
jzelinskie, line 66: constant for the header name

internal/datastore/crdb/keys.go
```diff
@@ -49,3 +54,22 @@ func prefix(s string) string {
...
+		for _, key := range strings.Split(strings.TrimSpace(keyList), ",") {
```
jzelinskie, line 67: why trim twice? if you're going to trim each individually, it's kinda pointless here right?

internal/datastore/crdb/crdb.go
```diff
@@ -177,6 +182,7 @@ func newCRDBDatastore(url string, options ...Option) (datastore.Datastore, error
 		writePool,
 		config.watchBufferLength,
 		keyer,
+		keySetInit,
```
jzelinskie, line 185: instead of passing this in as a HOF, could we instead have a function that always gets called that takes the context and the keyer and then does a `switch` over the keyer?

jzelinskie (changes requested):

internal/datastore/crdb/crdb.go
ecordell: I wanted to have the `keyer` no-op immediately if you're providing the key in the request, since it gets called for every namespace touched on write.

internal/datastore/crdb/keys.go
```diff
@@ -1,10 +1,17 @@
...
 
-import "strings"
+import (
+	"context"
+	"strings"
+
+	"google.golang.org/grpc/metadata"
+)
+
+const overlapMDKey = "spicedb-overlap"
```
josephschorr, line 10: we should define this in authzed-go in here: https://github.com/authzed/authzed-go/blob/main/pkg/requestmeta/requestmeta.go

ecordell: updated

e2e/newenemy/newenemy_test.go
```diff
@@ -160,7 +162,8 @@ func TestNoNewEnemy(t *testing.T) {
 		spice.WithMetricsPort(9100),
 		spice.WithDashboardPort(8100),
 		spice.WithDBName(dbName))
-	require.NoError(t, protectedSpiceDB.Start(ctx, tlog, "protected"))
+	require.NoError(t, protectedSpiceDB.Start(ctx, tlog, "protected",
```
josephschorr, line 165: I think it might be worth having the tests for the non-per-request key remain as well

internal/datastore/crdb/keys.go
```diff
@@ -49,3 +55,23 @@ func prefix(s string) string {
 	}
 	return prefix
 }
+
+// overlapKeysFromContext reads the request-provided initial overlap key set
+// from the grpc request metadata.
+func overlapKeysFromContext(ctx context.Context) keySet {
```
josephschorr, line 61: what happens if they neglect to give a key? do we fail-safe or fail-open?

ecordell: fail open - the point is that you only put the key on requests you care about, so that you can get `insecure` performance without them

josephschorr: okay, can we document that more clearly in the flag then?

PR #1198, opened 2023-03-10
updates to OTel 1.14.0

vroldanbet:


vroldanbet:
We are already updated to 1.15.1

PR #1153, opened 2023-02-09
datastore/postgres: switch to DB snapshots as primary component of revisions

jakedt:
Fixes #1130

PR #1200, opened 2023-03-13
fixes behaviour of various datastores when HeadRevision is outside GC window

vroldanbet:
This started as an attempt to fix MemDB's behaviour upon a head revision that is outside GC window. After creating a test-case that run for all datastores, it surfaced MySQL and PostgreSQL had a similar problem, each one with its own flavour.

The problem manifested in the various databases in different ways:
- in MemDB, a request with an out-of-window head revision would return `invalid zedtoken: revision has expired`
- in PostgreSQL, a request with an out-of-window head revision would return that the provided resource type was unknown (because it wasn't able to determine a valid namespace)

The fixes depend on the datastore:
- MemDB: we use the machine timestamp instead of the last valid revision, and consider last revision valid even outside that window
- PostgreSQL: do not GC the most recent transaction (which changed in https://github.com/authzed/spicedb/pull/1153), and update query to consider it valid if it's the most recent TX
- MySQL: update query to consider it valid if it's the most recent TX

### DIFF
--- a/internal/datastore/memdb/revisions_test.go
+++ b/internal/datastore/memdb/revisions_test.go
@@ -1,0 +1,27 @@
+package memdb
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHeadRevision(t *testing.T) {
+	ds, err := NewMemdbDatastore(0, 0, 500*time.Millisecond)
+	require.NoError(t, err)
+
+	older, err := ds.HeadRevision(context.Background())
+	require.NoError(t, err)
+	err = ds.CheckRevision(context.Background(), older)
+	require.NoError(t, err)
+
+	time.Sleep(550 * time.Millisecond)
+
+	// GC window elapsed, last revision is returned even if outside GC window
+	newer, err := ds.HeadRevision(context.Background())
+	require.NoError(t, err)
+	err = ds.CheckRevision(context.Background(), newer)
+	require.NoError(t, err)
+}
--- a/internal/datastore/postgres/revisions.go
+++ b/internal/datastore/postgres/revisions.go
@@ -56,7 +56,14 @@ const (
 	//   %[5] Name of the snapshot column
 	queryValidTransaction = `
 	WITH minvalid AS (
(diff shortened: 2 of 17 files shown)


internal/datastore/memdb/revisions_test.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func TestHeadRevision(t *testing.T) {
+	ds, err := NewMemdbDatastore(0, 0, 500*time.Millisecond)
+	require.NoError(t, err)
+
+	older, err := ds.HeadRevision(context.Background())
+	require.NoError(t, err)
+	older.LessThan(revisionFromTimestamp(time.Now().UTC()))
+	time.Sleep(500 * time.Millisecond)
```
josephschorr, line 18: make this longer than 500ms to ensure the GC window has passed

internal/datastore/postgres/revisions.go
```diff
@@ -56,7 +56,14 @@ const (
 	//   %[5] Name of the snapshot column
 	queryValidTransaction = `
 	WITH minvalid AS (
-		SELECT %[1]s, %[5]s FROM %[2]s WHERE %[3]s >= NOW() - INTERVAL '%[4]f seconds' ORDER BY %[3]s ASC LIMIT 1
+		SELECT %[1]s, %[5]s
+        FROM %[2]s
+        WHERE 
+            %[3]s >= NOW() - INTERVAL '%[4]f seconds'
+          OR
+             %[3]s = (SELECT MAX(%[3]s) FROM %[2]s)
```
ecordell, line 64: not blocking, but if you end up touching this PR again the spacing here is slightly off

pkg/datastore/test/caveat.go
```diff
@@ -27,7 +27,7 @@ import (
 func CaveatNotFoundTest(t *testing.T, tester DatastoreTester) {
 	require := require.New(t)
 
-	ds, err := tester.New(0, veryLargeGCWindow, 1)
+	ds, err := tester.New(0, veryLargeGCInterval, veryLargeGCWindow, 1)
```
ecordell, line 30: not a big deal, but it doesn't really matter what the gcinterval is here, since GC won't remove anything inside the window

ecordell (approved):
LGTM

PR #1218, opened 2023-03-21
datastore/postgres: fix invalid slice appends in snapshots

jakedt:


internal/datastore/postgres/snapshot.go
```diff
@@ -151,6 +138,19 @@ func (s pgSnapshot) compare(rhs pgSnapshot) comparisonResult {
...
+func (s pgSnapshot) anyTXVisible(first uint64, others []uint64) bool {
+	if s.txVisible(first) {
+		return true
+	} else {
+		for _, txid := range others {
+			if s.txVisible(txid) {
+				return true
+			}
+		}
+	}
```
ecordell, line 150:
```suggestion
	}
	for _, txid := range others {
		if s.txVisible(txid) {
			return true
		}
	}
```

ecordell (changes requested):

PR #1222, opened 2023-03-22
reverts caveat covering index for postgres datastore

vroldanbet:
partially reverts https://github.com/authzed/spicedb/pull/1118

we observed no performance gains while load-testing the new index, but it did impact write throughput
significantly


josephschorr (approved):
LGTM

PR #1223, opened 2023-03-23
move tx overlap touch above the actual work

ecordell:


ecordell:
this seemed to make performance worse

PR #1226, opened 2023-03-24
Fix deduplication bug in reachable resources

josephschorr:
We were previously deduplicating dispatching in reachable resources based solely on the subject found, but without being transformed for the parent resource type. This resulted in branches that produced the *same* subject, but with a different relation, being skipped occasionally due to a race in the dispatching.

We now do the duplication check at the dispatch point, to ensure we're checking the exact subject being dispatched.

Also adds a consistency test which reproduces the issue. Thanks to `tr33` from Discord for the bug report and reproduction case!

### DIFF
--- a/internal/graph/reachableresources.go
+++ b/internal/graph/reachableresources.go
@@ -103,7 +103,7 @@ func (crr *ConcurrentReachableResources) ReachableResources(
 			}
 
 			rsm := subjectIDsToResourcesMap(rewrittenSubjectRelation, req.SubjectIds)
-			drsm := rsm.filterForDispatch(dispatched)
+			drsm := rsm.asReadOnly()
 
 			err := crr.redispatchOrReport(
 				ctx,
@@ -114,6 +114,7 @@ func (crr *ConcurrentReachableResources) ReachableResources(
 				entrypoint,
 				stream,
 				req,
+				dispatched,
 			)
 			if err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


internal/graph/reachableresources.go
```diff
@@ -103,7 +103,7 @@ func (crr *ConcurrentReachableResources) ReachableResources(
 			}
 
 			rsm := subjectIDsToResourcesMap(rewrittenSubjectRelation, req.SubjectIds)
-			drsm := rsm.filterForDispatch(dispatched)
+			drsm := rsm.lockForDispatch()
```
vroldanbet, line 106: The name made me think mutexes were involved. Perhaps `asReadOnly` would be better suited - the underlying structs already follow that naming.

internal/graph/resourcesubjectsmap.go
```diff
@@ -104,6 +93,25 @@ func (rsm dispatchableResourcesSubjectMap) resourceIDs() []string {
...
+	resourceIDs := rsm.resourceIDs()
+	filtered := make([]string, 0, len(resourceIDs))
+	for _, resourceID := range resourceIDs {
+		if dispatched.Add(&core.ObjectAndRelation{
+			Namespace: dispatchSubjectType.Namespace,
+			ObjectId:  resourceID,
+			Relation:  dispatchSubjectType.Relation,
+		}) {
+			filtered = append(filtered, resourceID)
+			continue
```
vroldanbet, line 108: maybe I'm missing something but is `continue` necessary here?

vroldanbet (commented):
LGTM, just a nit

internal/graph/resourcesubjectsmap.go
josephschorr: Removed. This was from a previous version

PR #1227, opened 2023-03-24
introduces configurable dispatch hashring replication factor

vroldanbet:
this new CLI allows parameterizing the replication factor
used in hashring used by the dispatch client-side gRPC balancer.

This was a bit more involved as the balancer registry is a global
singleton meant to be registered on an init block. This would have
required moving flag handling outside of cobra, which is not ideal.

As a solution I changed how the balancer is named, appending the replication
factor to the name, and introducing a lazy-load mechanism with a mutex:
- if a balancer with a specific replication factor does not exist, lock is
  acquired and the balancer is registered. This avoids races if multiple
  spicedb servers are running in the same process
- if a balancer exists already, it's not registered
- we do not unregister the balancer when the server is wind down, as
  it could affect another spicedb instance in the process using
  the same replication factor

### DIFF
--- a/pkg/balancer/hashring.go
+++ b/pkg/balancer/hashring.go
@@ -1,6 +1,7 @@
 package balancer
 
 import (
+	"fmt"
 	"math/rand"
 	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


pkg/cmd/server/server.go
```diff
@@ -108,6 +111,10 @@ type Config struct {
 	TelemetryInterval        time.Duration
 }
 
+const defaultBackendsPerKey = 1
```
ecordell, line 114: it would be nice to make this configurable too while we're here

pkg/balancer/hashring.go
```diff
@@ -35,12 +36,23 @@ var logger = grpclog.Component("consistenthashring")
 // `balancer.Register(consistent.NewConsistentHashringBuilder(hasher, factor, spread))`
 func NewConsistentHashringBuilder(hasher consistent.HasherFunc, replicationFactor uint16, spread uint8) balancer.Builder {
 	return base.NewBalancerBuilder(
-		BalancerName,
+		NameForReplicationFactor(replicationFactor),
```
ecordell, line 39: why change the name at all? we'll only ever register one configuration

pkg/cmd/server/server.go
ecordell: it's not very useful with smaller clusters, but with larger clusters it lets you spread the load for a hot key between multiple nodes

ecordell (approved):
LGTM

PR #1228, opened 2023-03-27
select for update overlap

ecordell:


ecordell:
the tests re-confirmed that this doesn't suffice for making txs overlap

PR #1231, opened 2023-03-29
datastore/benchmark: add a datastore driver benchmark

jakedt:


vroldanbet (commented):
We have a datastore test package that has per-datastore abstractions via `RunningEngineForTest`. Would it make sense to reuse it, and move the DS specific methods there?

josephschorr (approved):
LGTM

PR #1232, opened 2023-03-29
postgres, crdb: update pgx to v5

ecordell:


internal/datastore/postgres/common/pgx.go
```diff
@@ -102,7 +103,7 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
 		}
 	}
 	l := zerologadapter.NewLogger(log.Logger)
-	connConfig.Logger = levelMappingFn(l)
+	connConfig.Tracer = &tracelog.TraceLog{Logger: levelMappingFn(l), LogLevel: tracelog.LogLevelInfo}
```
vroldanbet, line 106: hmm, what does `LogLevel` mean here? should this be aligned with the log level of the application instead of hardcoded to info?

vroldanbet (commented):
LGTM, two comments

internal/datastore/postgres/common/interceptor.go
```diff
@@ -3,42 +3,48 @@ package common
...
-	"github.com/jackc/pgtype/pgxtype"
-	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 )
 
+// Querier holds common methods for connections and pools, equivalent to
+// Querier (which is deprecated for pgx v5)
+type Querier interface {
+	Exec(ctx context.Context, sql string, arguments ...interface{}) (pgconn.CommandTag, error)
```
jzelinskie, line 13: You should prefer `any` to `interface{}`

internal/datastore/postgres/common/pgx.go
ecordell: the `LogLevel` here just sets whether we should log all messages (info) or errors only (error). 

we already do the mapping from info -> debug here so I think it makes sense to leave it

internal/datastore/postgres/snapshot.go
```diff
@@ -1,31 +1,71 @@
...
-func (s *pgSnapshot) DecodeText(ci *pgtype.ConnInfo, src []byte) error {
-	s.status = pgtype.Undefined
-	asText := string(src)
-	components := strings.SplitN(asText, ":", 3)
+var (
+	_ pgtype.TextScanner = &pgSnapshot{}
+	_ pgtype.TextValuer  = &pgSnapshot{}
+)
+
+func (s *pgSnapshot) ScanText(v pgtype.Text) error {
```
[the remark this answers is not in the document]
jzelinskie: This comment should probably live in code and not just the PR unless we think it's immediately going to change.

jzelinskie (approved):

PR #1235, opened 2023-03-30
Add flags for maxlifetime jitter

ecordell:
Replaces #1229, which hardcodes the jitter value

### DIFF
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -41,6 +41,7 @@ var BuilderForEngine = map[string]engineBuilderFunc{
 type ConnPoolConfig struct {
 	MaxIdleTime         time.Duration
 	MaxLifetime         time.Duration
+	MaxLifetimeJitter   time.Duration
 	MaxOpenConns        int
 	MinOpenConns        int
 	HealthCheckInterval time.Duration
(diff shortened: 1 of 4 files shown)


pkg/cmd/datastore/datastore.go
```diff
@@ -44,6 +44,7 @@ type ConnPoolConfig struct {
 	MaxOpenConns        int
 	MinOpenConns        int
 	HealthCheckInterval time.Duration
+	MaxLifetimeJitter   time.Duration
```
jzelinskie, line 47: nitpick but i'd move this below the maxlifetime field in all the literals

pkg/cmd/datastore/datastore.go
```diff
@@ -53,6 +54,7 @@ func DefaultReadConnPool() *ConnPoolConfig {
 		MaxOpenConns:        20,
 		MinOpenConns:        20,
 		HealthCheckInterval: 30 * time.Second,
+		MaxLifetimeJitter:   10 * time.Minute,
```
jzelinskie, line 57: Do we want to explicitly not set this here and let it use the 20% default when unset?

jzelinskie (changes requested):

pkg/cmd/datastore/datastore.go
```diff
@@ -74,6 +75,7 @@ func RegisterConnPoolFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, defa
 	flagSet.IntVar(&opts.MaxOpenConns, flagName("max-open"), defaults.MaxOpenConns, "number of concurrent connections open in a remote datastore's connection pool")
 	flagSet.IntVar(&opts.MinOpenConns, flagName("min-open"), defaults.MinOpenConns, "number of minimum concurrent connections open in a remote datastore's connection pool")
 	flagSet.DurationVar(&opts.MaxLifetime, flagName("max-lifetime"), defaults.MaxLifetime, "maximum amount of time a connection can live in a remote datastore's connection pool")
+	flagSet.DurationVar(&opts.MaxLifetimeJitter, flagName("max-lifetime-jitter"), defaults.MaxLifetimeJitter, "waits rand(0, jitter) after a connection is open for max lifetime to actually close the connection")
```
jzelinskie, line 78: probably need to change the description here to include that "0" defaults to 20%

PR #1238, opened 2023-03-31
upsert overlap

ecordell:
This uses an `UPSERT` to create transaction overlap in CRDB. In benchmarks it appears to be slightly faster than the `INSERT ... ON CONFLICT`


ecordell:
Recent CRDB release notes make it sound as though SELECT FOR UPDATE may work for overlap now - it's worth re-running the tests with new versions.

ecordell:
I've filed https://github.com/authzed/spicedb/issues/2190 to continue looking at other options for CRDB overlap.

PR #1245, opened 2023-04-03
gomod: bump cobrautil

jzelinskie:
This changes the default for OTEL sample ratio to be 1% instead of 20%.

Our experience running SpiceDB in production has made it clear that 20% is far too large of a default sample ratio for most deployments.


josephschorr (approved):
LGTM

PR #1248, opened 2023-04-03
Extended object ids

jakedt:
Requires the API changes from https://github.com/authzed/api/tree/extended-object-ids


internal/datastore/mysql/migrations/zz_migration.0005_extend_object_id.go
```diff
@@ -0,0 +1,22 @@
+package migrations
+
+import "fmt"
+
+// Extend the object ID column to 1024 characters
+func adjustObjectIDLength(t *tables) string {
+	return fmt.Sprintf(
+		`ALTER TABLE %s
+		CONVERT TO CHARACTER SET latin1,
```
josephschorr, line 9: we should make absolutely sure that the IDs and the names of objects and relations are within latin1 in the authzed-go unit tests, once we start making changes there

PR #1252, opened 2023-04-06
Move to Go version 1.20.3 to fix reported vulns in Go 1.20.2

josephschorr:
See: https://groups.google.com/g/golang-announce/c/Xdv6JL9ENs8

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -32,7 +32,7 @@ on:  # yamllint disable-line rule:truthy
       - "internal/**"
       - "proto/**"
 env:
-  GO_VERSION: "~1.20.2"
+  GO_VERSION: "~1.20.3"
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


.github/workflows/build-test.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: "Build & Test"
-on:  # yamllint disable-line rule:truthy
+on: # yamllint disable-line rule:truthy
```
jzelinskie, line 3: This is double spaces on purpose. I'm pretty sure it's a yamllint rule

jzelinskie (changes requested):

.github/workflows/build-test.yaml
josephschorr: Fixed, but I think we should just ignore this going forward unless there is a good reason

jzelinskie (approved):

PR #1254, opened 2023-04-07
Add support for extended object ids

### LINKED ISSUE #1210: Proposal: Extend object IDs to support more values and encodings
# Situation

Right now the regex for object IDs is: `^(([a-zA-Z0-9_][a-zA-Z0-9/_|-]{0,127})|\\*)$`

Visualized by: https://regexper.com/#%5E%28%28%5Ba-zA-Z0-9_%5D%5Ba-zA-Z0-9%2F_%7C-%5D%7B0%2C127%7D%29%7C%5C*%29%24

Currently addressed use cases for this regex are:
- conforming hand-crafted object IDs
- Many Auth0 subject IDs which contain the provider name, a liter `|` and then the provider's unique identifier
- Base64 identifiers in url-safe encoding

Standard types that are known to not work natively:
- emails (contain `@` and `.`)
- JWT [sub claims](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.2) that contain URIs (contain `:`, `/`, `.`, `%` and others) or are otherwise non-conformant
- values which are longer than 128 characters in a conforming encoding.
- Standard Base64 values (contain `=`, `+`, and `/`)

The current guidance for using non-confirming types is to [base64 encode](https://en.wikipedia.org/wiki/Base64) them or one-way hash them.

One-way hashing is irreversible, which makes `LookupSubject` and `LookupResources` results less useful.

Base64 encoding increases the byte-length of the data by 1/3 due to storing 24 bits of binary data in 32 bits of ascii characters.
This is problematic because certain types are already quite long _before_ encoding. For example, an email can be up to 320 characters long before encoding, resulting in a string up to 427 characters after encoding.

It is non-trivial to arbitrarily extend the character set allowed. Right now object IDs are sometimes encoded directly into strings (for example, AuthZed's playground and zed validate file format), and some of the extended characters would make parsing these strings ambiguous.

It is also non-trivial to extend the length of the field, because the current object_id length assumptions may be baked into [spicedb](https://github.com/authzed/spicedb) datastore drivers or client storage. The latter is mitigated by the fact that the client would have had to give the value to the API in order to receive it back.

# Target

Ideally we would support more types, either natively or through encoding. In all cases, the length of the current object ID should be extended. For storage in the zed and playground file format, IDs could be encoded before serialization, allowing the use of extended characters in the API. For a partially extended character set, we might be able to keep the parser predictive (i.e. no backtracking) without encoding.

# Proposal (subject to discussion)

- Audit existing SpiceDB datastore implementations to verify compatibility with longer object IDs
- Poll the community (through this issue and other means) about whether extending the field length would cause them trouble
- Extend the `object_id` field to 1024 ascii characters.
- Extend the allowed encoding to accept `=`, `/`, and `+` to allow for standard Base64 values

josephschorr:
Fixes #1210


pkg/tuple/tuple.go
```diff
@@ -201,15 +207,25 @@ func Parse(tpl string) *core.RelationTuple {
...
+	resourceID := groups[stringz.SliceIndex(parserRegex.SubexpNames(), "resourceID")]
+	if err := ValidateResourceID(resourceID); err != nil {
+		return nil
+	}
+
+	subjectID := groups[stringz.SliceIndex(parserRegex.SubexpNames(), "subjectID")]
+	if err := ValidateSubjectID(subjectID); err != nil {
+		return nil
+	}
+
```
vroldanbet, line 219: I believe this is a redundant since `core.proto` already generates validation logic in `pkg/proto/core/v1/core.pb.validate.go`. We would be keeping 2 copies of regex that could drift-away, and do redundant and expensive regex checks.

pkg/tuple/tuple_test.go
```diff
@@ -297,6 +299,39 @@ var testCases = []struct {
...
+	},
+	{
+		input:          "testns:foo#testrel@user:" + superLongID,
+		expectedOutput: "testns:foo#testrel@user:" + superLongID,
+		tupleFormat: makeTuple(
+			ObjectAndRelation("testns", "foo", "testrel"),
+			ObjectAndRelation("user", superLongID, "..."),
+		),
+		relFormat: rel("testns", "foo", "testrel", "user", superLongID, ""),
+	},
```
vroldanbet, line 319: let's add a test with superlong subject and resource object ids

internal/services/v1/permissions_test.go
```diff
@@ -223,6 +222,13 @@ func TestCheckPermissions(t *testing.T) {
 			v1.CheckPermissionResponse_PERMISSIONSHIP_UNSPECIFIED,
 			codes.FailedPrecondition,
 		},
+		{
+			obj("document", "-base64YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK=="),
+			"view",
+			sub("user", "unknowngal", ""),
```
vroldanbet, line 228: add test with base64 encoded subject

pkg/tuple/tuple_test.go
```diff
@@ -297,6 +299,39 @@ var testCases = []struct {
...
+	{
+		input:          "testns:-base64YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==#testrel@user:-base65YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==",
+		expectedOutput: "testns:-base64YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==#testrel@user:-base65YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==",
+		tupleFormat: makeTuple(
+			ObjectAndRelation("testns", "-base64YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==", "testrel"),
+			ObjectAndRelation("user", "-base65YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==", "..."),
+		),
+		relFormat: rel("testns", "-base64YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==", "testrel", "user", "-base65YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==", ""),
+	},
```
vroldanbet, line 334: it seems like we are testing only the valid cases, but how can we be sure the validation is working if we are not testing an invalid tuple?

pkg/development/wasm/main_test.go
```diff
@@ -75,7 +75,7 @@ func TestInvalidRelationship(t *testing.T) {
 		},
 	})
 	require.NotNil(t, response.GetInternalError())
-	require.Equal(t, "invalid resource id; must be alphanumeric and between 1 and 127 characters", response.GetInternalError())
+	require.Equal(t, `invalid resource id; must match ([a-zA-Z0-9/_|\-=+]{1,})`, response.GetInternalError())
```
vroldanbet, line 78: there is a character size limit but it's no longer obvious through the regex, should we clarify that in the message?

pkg/datastore/test/tuples.go
```diff
@@ -317,6 +317,59 @@ func SimpleTest(t *testing.T, tester DatastoreTester) {
...
+	testCases := []string{
+		"simple",
+		"google|123123123123",
+		"--=base64YWZzZGZh-ZHNmZHPwn5iK8J+YivC/fmIrwn5iK==",
+		"veryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylong",
```
vroldanbet, line 325: should we also test invalid IDs?

pkg/tuple/tuple.go
josephschorr: This hasn't changed and I feel it is important to have this extra check in case the validation code fails for some reason, especially since the validation code could be removed externally without "SpiceDB" knowing

vroldanbet: it's true but we get the same level of confidence by testing all these things in SpiceDB. Removing this does not make this less secure IMO, and likewise I think the code is not more secure by having this here.

josephschorr: We actually don't: the constructed relationship/tuple is not having its `.Validate` called here. Even if we did, we'd need to keep `ValidateSubjectID` and the other methods around *anyway* because the `.Validate` method returns poor error messages, so we have to validate in a better fashion regardless

pkg/tuple/tuple_test.go
josephschorr: you mean together or...?

pkg/tuple/tuple_test.go
josephschorr: We test invalid cases above

pkg/development/wasm/main_test.go
josephschorr: Its clarified in the tooling

pkg/datastore/test/tuples.go
josephschorr: We can't; the datastore doesn't validate

pkg/tuple/tuple.go
vroldanbet: it sounds like the only problem here is the message returned by the generate `Validate` method?

pkg/tuple/tuple_test.go
vroldanbet: yeah, a test that has both a long subject id and long resource id

vroldanbet (approved):

PR #1263, opened 2023-04-13
Fix flakiness of bootstrap file in serve-testing by squashing all memdb revisions

josephschorr:
Also adds an integration test

Thanks to https://github.com/wscalf for the bug report and basis of the test case

Original bug report: https://discord.com/channels/844600078504951838/844600078948630559/1095721051020873779

wscalf — 12/04/2023, 7:44 AM
Hey there,

We're doing some integration tests with SpiceDB using the serve-testing command and using --load-configs and a YAML file containing schema and relationships, and we're noticing that sometimes (between 2% and 20% of the time), our tests fail- the schema appears to be there, but the relationships may not have been loaded yet- all test failures are due to missing relationships, and it looks like in the SpiceDB logs, the log message for serving comes before the log message for loading relationships:
{"level":"info","addr":":50051","network":"tcp","service":"grpc","workers":0,"insecure":true,"time":"2023-04-12T14:43:05Z","message":"grpc server started serving"}
{"level":"info","addr":":50052","network":"tcp","service":"readonly-grpc","workers":0,"insecure":true,"time":"2023-04-12T14:43:05Z","message":"grpc server started serving"}
{"level":"info","filePath":"/mnt/spicedb_bootstrap.yaml","schemaDefinitionCount":7,"time":"2023-04-12T14:43:06Z","message":"adding schema definitions"}

Should we be doing something differently? And as a side question, is --load-configs different from --datastore-bootstrap-files in some significant way?

### DIFF
--- a/cmd/spicedb/servetesting_race_test.go
+++ b/cmd/spicedb/servetesting_race_test.go
@@ -1,0 +1,64 @@
+//go:build docker && image
+// +build docker,image
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/ory/dockertest/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// Based on a test originally written by https://github.com/wscalf
+func TestCheckPermissionOnTesterNoFlakes(t *testing.T) {
+	_, b, _, _ := runtime.Caller(0)
+	basepath := filepath.Dir(b)
+	tester, err := newTester(t,
+		&dockertest.RunOptions{
+			Repository:   "authzed/spicedb",
+			Tag:          "ci",
+			Cmd:          []string{"serve-testing", "--load-configs", "/mnt/spicedb_bootstrap.yaml"},
+			Mounts:       []string{path.Join(basepath, "testdata/bootstrap.yaml") + ":/mnt/spicedb_bootstrap.yaml"},
+			ExposedPorts: []string{"50051/tcp", "50052/tcp", "8443/tcp", "8444/tcp"},
+		},
+		"",
+		true,
+	)
+	require.NoError(t, err)
+	defer tester.cleanup()
+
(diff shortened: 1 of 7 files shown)


internal/middleware/pertoken/pertoken.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	gcWindow             = 1 * time.Hour
-	revisionQuantization = 10 * time.Millisecond
+	revisionQuantization = 0 * time.Millisecond
```
vroldanbet, line 22: My concern with this change is that it would set a different default behaviour to `spicedb serve` and set wrong expectations. Why isn't asking folks to do `FullyConsistent` requests or use `AtLeastAsFresh` the recommended strategy in testing?

jakedt: I concur. This was intentional. Better to find any problems related to consistency in test than in production. If we were to try to solve it, I think the best way would be to have all timestamps requested before the "handover" point resolve to the handover snapshot, so at least the preloaded data is already consistent.

josephschorr: > Why isn't asking folks to do FullyConsistent requests or use AtLeastAsFresh the recommended strategy in testing?

Because there is an expectation that preloaded data is already "present" in the system; you can't get a ZedToken for it, and there shouldn't be an expectation to need to make a FullyConsistent request for data that was already there if it has been there for (simulated) days or years

> I think the best way would be to have all timestamps requested before the "handover" point resolve to the handover snapshot, so at least the preloaded data is already consistent.

how would we do so?

josephschorr: As discussed, squashed the memdb revisions

cmd/spicedb/servetesting_race_test.go
```diff
@@ -0,0 +1,82 @@
...
+				},
+			},
+		})
+
+		assert.NoError(t, err)
+		assert.Equal(t, v1.CheckPermissionResponse_PERMISSIONSHIP_HAS_PERMISSION, result.Permissionship, "Error on attempt #%d", i)
+	}
+}
+
+func newConnection(port string) (*authzed.Client, error) {
```
vroldanbet, line 69: is this dead code?

vroldanbet (commented):
LGMT just one question

vroldanbet (dismissed):

vroldanbet (approved):

PR #1265, opened 2023-04-16
Paginated datastore API

jakedt:


internal/services/v1/relationships.go
```diff
@@ -126,30 +130,49 @@ func (ps *permissionServer) ReadRelationships(req *v1.ReadRelationshipsRequest,
...
-	usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
-		DispatchCount: 1,
-	})
+	batchCount := datastoreQueryBatchSize
+	var cursor options.Cursor
+	var dispatchCount uint32
+	for batchCount == datastoreQueryBatchSize {
+		dispatchCount++
+		usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
+			DispatchCount: dispatchCount,
```
vroldanbet, line 139: where is dispatching happening here which makes this necessary?

internal/services/v1/relationships.go
```diff
@@ -126,30 +130,49 @@ func (ps *permissionServer) ReadRelationships(req *v1.ReadRelationshipsRequest,
 		return rewriteError(ctx, err)
 	}
 
-	usagemetrics.SetInContext(ctx, &dispatchv1.ResponseMeta{
-		DispatchCount: 1,
-	})
+	batchCount := datastoreQueryBatchSize
+	var cursor options.Cursor
+	var dispatchCount uint32
+	for batchCount == datastoreQueryBatchSize {
```
vroldanbet, line 136: This patterns will become useful in other spots as we move different parts of the codebase to pagination. I could see us creating a `RelationshipIterator` that abstracts the need to fetch the cursor and infinitely scrolls over the pages. WDYT?

vroldanbet, line 136: this can be tackled in a different PR since this is a more widespread problem, but for loops like this that incur expensive I/O operations, we want to check context cancelation. It would help the process free resources faster and spare wasteful I/O after cancelation is signaled.

vroldanbet (commented):
This looks good to me! 💯 I mostly pointed out missing tests, and the opportunity to create a new `RelationshipIterator` that abstracts pagination, when pagination is not exposed to API client (e.g. streaming APIs)

internal/services/v1/relationships.go
```diff
@@ -126,30 +130,49 @@ func (ps *permissionServer) ReadRelationships(req *v1.ReadRelationshipsRequest,
...
+		batchCount = 0
+		for tpl := tupleIterator.Next(); tpl != nil; tpl = tupleIterator.Next() {
+			batchCount++
+			err := resp.Send(&v1.ReadRelationshipsResponse{
+				ReadAt:       revisionReadAt,
+				Relationship: tuple.ToRelationship(tpl),
+			})
+			if err != nil {
+				tupleIterator.Close()
+				return err
```
vroldanbet, line 161: we need to call `rewriteError` here as well or else de gRPC error will be returned with a different error code

internal/datastore/common/sql.go
```diff
@@ -77,18 +95,120 @@ func NewSchemaQueryFilterer(schema SchemaInformation, initialQuery sq.SelectBuil
...
+		)
+	}
+
+	return sqf
+}
+
+func (sqf SchemaQueryFilterer) After(cursor *core.RelationTuple, order options.SortOrder) SchemaQueryFilterer {
+	switch order {
+	case options.ByResource:
+		sqf.queryBuilder = sqf.queryBuilder.Where(
```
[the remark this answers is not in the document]
jakedt: I started with that, taken directly from your code. It doesn't work for `spanner` though. I could split them and use both strategies, but this seems to work everywhere. Still need to scale test and see what indices we're missing.

internal/datastore/common/sql_test.go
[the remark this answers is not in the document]
jakedt: There are datastore tests for both modes of `TupleOrder` in the `pagination.go` test suite. Does that work, or are you thinking about something else?

internal/services/v1/relationships.go
jakedt: Before we were retuning `1` to represent the 1 datastore query being done. I thought that since we're now doing N datastore queries to fulfill the read, this would be a fair measure.

internal/datastore/common/tuple.go
```diff
@@ -1,40 +1,53 @@
...
+	"github.com/authzed/spicedb/pkg/datastore/options"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 )
 
-var errClosedIterator = errors.New("unable to iterate: iterator closed")
-
 // NewSliceRelationshipIterator creates a datastore.TupleIterator instance from a materialized slice of tuples.
-func NewSliceRelationshipIterator(tuples []*core.RelationTuple) RelationshipIterator {
-	return &sliceRelationshipIterator{tuples: tuples}
+func NewSliceRelationshipIterator(tuples []*core.RelationTuple, initialError error, order options.SortOrder) datastore.RelationshipIterator {
```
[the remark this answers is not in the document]
jakedt: It's actually not required anymore. Removed.

pkg/datastore/options/options.go
```diff
@@ -6,10 +6,22 @@ import (
 
 //go:generate go run github.com/ecordell/optgen -output zz_generated.query_options.go . QueryOptions ReverseQueryOptions
 
+type SortOrder int8
```
josephschorr, line 9: doc comments on exported members

internal/datastore/common/cursor.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+const (
+	errEncodingCursor = "error encoding cursor: %s"
+	errDecodingCursor = "error decoding cursor: %s"
+)
+
+func NoopCursorFunc(_ *core.RelationTuple) (options.Cursor, error) {
```
josephschorr, line 17: doc comments on exported members

internal/datastore/common/cursor.go
```diff
@@ -0,0 +1,49 @@
...
+	return "", datastore.ErrCursorsWithoutSorting
+}
+
+func EncodedProtoCursor(tpl *core.RelationTuple) (options.Cursor, error) {
```
josephschorr, line 21: Is the cursor for resuming *at* the tuple or *after*? I assume after, but if so, can you clarify the parameter name here?

internal/datastore/common/cursor.go
```diff
@@ -0,0 +1,49 @@
...
+	if tpl == nil {
+		return "", datastore.ErrCursorEmpty
+	}
+
+	marshalled, err := tpl.MarshalVT()
+	if err != nil {
+		return "", fmt.Errorf(errEncodingCursor, err)
+	}
+	return options.Cursor(base64.StdEncoding.EncodeToString(marshalled)), nil
```
josephschorr, line 30: Are these cursors ever sent outside of internal code? If so, and we ever have to change this format, we might want to consider using an encoded proto with a version, rather than just the tuple itself. If not, then no worries

pkg/datastore/datastore.go
```diff
@@ -367,6 +367,10 @@ type RelationshipIterator interface {
 	// Next returns the next tuple in the result set.
 	Next() *core.RelationTuple
 
+	// Cursor returns a cursor representing the last relationship returned if
```
josephschorr, line 370: maybe something like:

Cursor returns a cursor that can be used to resume reading of relationships from the last relationship returned. Only applies if a sort ordering was requested.

internal/services/v1/relationships.go
josephschorr: +1 to that idea

internal/datastore/common/cursor.go
jakedt: After, but this isn't the right place to make that determination. The input represents the last tuple returned to the user, so I renamed it `last`.

internal/datastore/common/cursor.go
jakedt: Changed cursors to just be a `RelationTuple` pointer.

internal/services/v1/relationships.go
jakedt: Done.

internal/datastore/memdb/readonly.go
```diff
@@ -345,16 +345,99 @@ func filterFuncForFilters(
 			return !found
 		}
 
+		tpl := &core.RelationTuple{
```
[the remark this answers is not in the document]
vroldanbet: friendly ping ☝🏻 can we make this non-allocating?

pkg/datastore/options/options.go
```diff
@@ -6,10 +6,29 @@ import (
...
 
+// SortOrder is an enum which represents the order in which the caller would like
+// the data returned.
+type SortOrder int8
+
+const (
+	// Unsorted lets the underlying datastore choose the order, or no order at all
+	Unsorted SortOrder = iota
+
+	// By resource sorts the relationships by the resource component first
```
vroldanbet, line 17:
```suggestion
	// ByResource sorts the relationships by the resource component first
```

pkg/datastore/pagination/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+	if pi.Err() != nil {
+		return nil
+	}
+
+	var next *core.RelationTuple
+	for next = pi.delegate.Next(); next == nil; next = pi.delegate.Next() {
+		if pi.delegate.Err() != nil {
+			pi.err = pi.delegate.Err()
+			return nil
+		}
```
vroldanbet, line 59: nit: 🤔 handing of error state seems inconsistent in this method. What do you think about adding an additional case to `Err()` and checking here `pi.Err()`, and also below after `startNewBatch`?

```suggestion
		if pi.Err() != nil {
			return nil
		}
```

pkg/datastore/pagination/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+
+		if pi.returnedFromBatch < pi.pageSize {
+			// No more tuples to get
+			return nil
+		}
+
+		cursor, err := pi.delegate.Cursor()
+		if errors.Is(err, datastore.ErrCursorEmpty) {
```
vroldanbet, line 67: it would seem we could ignore the error here if it's not `ErrCursorEmpty`

pkg/datastore/pagination/iterator.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/authzed/spicedb/internal/datastore/common"
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// NewPaginatedIterator creates an implementation of the datastore.Iterator
+// interface that internally paginates over datastore results.
+func NewPaginatedIterator(
```
vroldanbet, line 15: Does not have a test, but perhaps as I proposed in `pkg/datastore/test/pagination.go`, we could do integration test for the iterator there

pkg/datastore/pagination/iterator.go
jakedt: Why? I think as written it's more clear when and where a particular error is going to occur and get copied and become sticky in the parent iterator. For example, if we just checked delegate error in the err block, we could accidentally overwrite delegate and clear the error state.

pkg/datastore/pagination/iterator.go
jakedt: Good catch!

pkg/datastore/test/pagination.go
```diff
@@ -0,0 +1,315 @@
...
+
+				expected := sortedStandardData(tc.objectType, tc.ordering)
+
+				foreachTxType(ctx, ds, rev, func(reader datastore.Reader) {
+					// Test that if you ask for resume without an order we error
+					_, err := reader.QueryRelationships(ctx, datastore.RelationshipsFilter{
+						ResourceType: tc.objectType,
+					}, options.WithLimit(&testLimit), options.WithAfter(&core.RelationTuple{}))
+					require.ErrorIs(err, datastore.ErrCursorsWithoutSorting)
+
```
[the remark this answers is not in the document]
jakedt: I think of the paginated iterator as a helper written from the perspective of the datastore consumer. As such I don't think tests for it belong here. Ideally it would test using mocks. Let me see how long it will take to write that.

pkg/datastore/test/datastore.go
```diff
@@ -56,6 +56,11 @@ func AllExceptWatch(t *testing.T, tester DatastoreTester) {
 	t.Run("TestMultipleReadsInRWT", func(t *testing.T) { MultipleReadsInRWTTest(t, tester) })
 	t.Run("TestConcurrentWriteSerialization", func(t *testing.T) { ConcurrentWriteSerializationTest(t, tester) })
 
+	t.Run("TestOrdering", func(t *testing.T) { OrderingTest(t, tester) })
+	t.Run("TestLimit", func(t *testing.T) { LimitTest(t, tester) })
+	t.Run("TestOrderedLimit", func(t *testing.T) { OrderedLimitTest(t, tester) })
+	t.Run("TestResume", func(t *testing.T) { ResumeTest(t, tester) })
```
josephschorr, line 62: Might be better to be specific here: `TestOrderedQueryRels`, `TestResumeQueryRels`, etc

internal/datastore/common/sql.go
```diff
@@ -77,18 +95,120 @@ func NewSchemaQueryFilterer(schema SchemaInformation, initialQuery sq.SelectBuil
 	}
 }
 
+func (sqf SchemaQueryFilterer) TupleOrder(order options.SortOrder) SchemaQueryFilterer {
```
josephschorr, line 116: Can we add SQL tests for this (and below)

vroldanbet (approved):
LGTM!

PR #1267, opened 2023-04-17
Preloaded task runner for check dispatch

josephschorr:


internal/graph/preloadedtaskrunner.go
```diff
@@ -0,0 +1,104 @@
...
+func (tr *preloadedTaskRunner) selectTask() TaskFunc {
+	tr.lock.Lock()
+	defer tr.lock.Unlock()
+
+	if len(tr.tasks) == 0 {
+		return nil
+	}
+
+	task := tr.tasks[0]
+	tr.tasks = tr.tasks[1:]
```
jzelinskie, line 102: This can be one line according to the SliceTricks wiki: https://github.com/golang/go/wiki/SliceTricks#pop-frontshift

```go
task, tr.tasks = tr.tasks[0], tr.tasks[1:]
return task
```

internal/graph/preloadedtaskrunner.go
```diff
@@ -0,0 +1,104 @@
...
+		default:
+			// Select a task from the list, if any.
+			task := tr.selectTask()
+			if task == nil {
+				return
+			}
+
+			// Run the task. If an error occurs, store it and cancel any further tasks.
+			err := task(tr.ctx)
+			if err != nil {
```
jzelinskie, line 86: should we use this error somewhere?

internal/graph/preloadedtaskrunner.go
```diff
@@ -0,0 +1,104 @@
+package graph
+
+import (
+	"context"
+	"sync"
+)
+
+// preloadedTaskRunner is a task runner that invokes a series of preloaded tasks,
```
jzelinskie, line 8: Should we be pulling this out into a separate concurrency package?
It might only be useful in the graph now, but not forever!

jzelinskie (changes requested):

internal/graph/preloadedtaskrunner.go
josephschorr: Has to be of the form
```
	var task TaskFunc
	task, tr.tasks = tr.tasks[0], tr.tasks[1:]
	return task
```

internal/graph/preloadedtaskrunner.go
josephschorr: The expectation is that the runner is returning it to someone as part of its own response

internal/graph/preloadedtaskrunner.go
josephschorr: Eventually, yes, but I think for now its okay to keep this in the dispatch package

jzelinskie (approved):

PR #1277, opened 2023-04-20
Add v1 integration test for stream timeout

josephschorr:
Also exposes the parameter as a flag in case it needs to be overridden


internal/services/v1/relationships_test.go
```diff
@@ -1122,6 +1122,40 @@ func TestWriteRelationshipsUpdatesOverLimit(t *testing.T) {
...
+		RelationshipFilter: &v1.RelationshipFilter{
+			ResourceType: "document",
+		},
+	})
+	require.NoError(err)
+	time.Sleep(1 * time.Millisecond)
+
+	// Ensure the recv fails with a context cancelation.
+	_, err = stream.Recv()
+	require.ErrorContains(err, "context canceled")
```
vroldanbet, line 1156: `require.ErrorIs`?

josephschorr: Do we know the type of the context cancelation error?

pkg/cmd/serve.go
```diff
@@ -118,6 +118,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	cmd.Flags().Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	cmd.Flags().Uint16Var(&config.MaximumPreconditionCount, "update-relationships-max-preconditions-per-call", 1000, "maximum number of preconditions allowed for WriteRelationships and DeleteRelationships calls")
 	cmd.Flags().IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes. A value of zero or less means no limit")
+	cmd.Flags().DurationVar(&config.StreamingAPITimeout, "streaming-api-resp-timeout", 30*time.Second, "maximum duration a streaming API call waits for a response before it times out")
```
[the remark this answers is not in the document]
josephschorr: Updated

josephschorr:
Rebased and updated

internal/services/v1/relationships_test.go
vroldanbet: unless we are doing something unorthodox with context cancellation, we should. `ErrorIs` should be able to unroll wrapped errors and identify the context cancellation type (which is a variable in the `context` package)

vroldanbet (approved):

PR #1278, opened 2023-04-20
Fix issue where Watch in Postgres was looping endlessly

### LINKED ISSUE #1272: Endless loop in watch stream
We have the following problem, if we concurrently add two packs of relationships (in this case 2 batches of 5 relationships), then the watch service enters an endless loop, constantly giving back the latest records.
Originally was reproduced on version 1.19.1 but then clarified, reproduced on version 1.17.0, everything worked correctly on version 1.16.2

As I understand it, the problem is how the next snapshot is determined ([watch.js](https://github.com/authzed/spicedb/blob/5ae5f30ac01591a185d9f795a43773a52ea28852/internal/datastore/postgres/watch.go#L51)), but I did not master the current logic of the algorithm )

Some detail can be found here https://gist.github.com/aberkovsky/4ca8fd7227b83772ba6d3fc19cc18595

josephschorr:
This occurred if there were two exactly overlapping transactions, which caused the watch code to repeatedly "move backwards" when looking up the next revision

The fix is to ensure that all found transactions are accounted in the overall transaction revision passed to the next iteration of the watch loop

Fixes #1272

Thanks to @aberkovsky for the bug report and detailed repro information!

### DIFF
--- a/internal/datastore/postgres/postgres_test.go
+++ b/internal/datastore/postgres/postgres_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -141,6 +142,24 @@ func TestPostgresDatastore(t *testing.T) {
 					RevisionQuantization(0),
 					GCWindow(1*time.Millisecond),
 					WatchBufferLength(1),
+					MigrationPhase(config.migrationPhase),
+				))
+
+				t.Run("ConcurrentRevisionWatch", createDatastoreTest(
+					b,
+					ConcurrentRevisionWatchTest,
+					RevisionQuantization(0),
+					GCWindow(1*time.Millisecond),
+					WatchBufferLength(50),
+					MigrationPhase(config.migrationPhase),
+				))
+
+				t.Run("OverlappingRevisionWatch", createDatastoreTest(
+					b,
+					OverlappingRevisionWatchTest,
+					RevisionQuantization(0),
+					GCWindow(1*time.Millisecond),
+					WatchBufferLength(50),
 					MigrationPhase(config.migrationPhase),
 				))
 			}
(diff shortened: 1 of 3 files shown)


internal/datastore/postgres/postgres_test.go
```diff
@@ -738,6 +757,239 @@ func ConcurrentRevisionHeadTest(t *testing.T, ds datastore.Datastore) {
...
+			colObjectID,
+			colRelation,
+			colUsersetNamespace,
+			colUsersetObjectID,
+			colUsersetRelation,
+			colCaveatContextName,
+			colCaveatContext,
+			colCreatedXid,
+			nextx+1,
+		))
```
ecordell, line 964: Since the fix works by bumping the xid in the query forward, wouldn't it make sense to include a third write here that adds a _non_overlapping row at `nextx+2` to ensure it doesn't get skipped (It should be fine because the query uses `>=`, but just to be defensive)

internal/datastore/postgres/postgres_test.go
```diff
@@ -738,6 +757,239 @@ func ConcurrentRevisionHeadTest(t *testing.T, ds datastore.Datastore) {
...
+	require.True(r.IsReady)
+
+	rev, err := ds.HeadRevision(ctx)
+	require.NoError(err)
+
+	pds := ds.(*pgDatastore)
+	require.True(pds.watchEnabled)
+
+	prev := rev.(postgresRevision)
+	nextx := prev.snapshot.xmax + 1
```
ecordell, line 903: super nit, feel free to ignore: 
```suggestion
	nexttx := prev.snapshot.xmax + 1
```

internal/datastore/postgres/postgres_test.go
```diff
@@ -738,6 +757,239 @@ func ConcurrentRevisionHeadTest(t *testing.T, ds datastore.Datastore) {
...
+			colCaveatContext,
+			colCreatedXid,
+			nextx,
+		))
+		if err != nil {
+			return err
+		}
+
+		_, err = tx.Exec(ctx, fmt.Sprintf(
+			`INSERT INTO %s ("xid", "snapshot") VALUES ('%d', '%d:%d:')`,
```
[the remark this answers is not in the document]
josephschorr: As its a test, I think its fine

internal/datastore/postgres/postgres_test.go
josephschorr: I wanted the test to exactly reproduce what the user reported; I could add another test if you like?

ecordell (approved):
LGTM

PR #1283, opened 2023-04-25
CRDB: detect draining nodes on connection acquisition

ecordell:
At higher scales, creating new connections against CRDB dominates SpiceDB latency. 

We currently follow best-practices around connection draining by ensuring that our connection max lifetimes are <5m, which ensures all connections are drained properly during node drains in Cockroach Dedicated.

As we add more SpiceDB nodes and CRDB nodes, we increase the number of connections to take advantage of the extra capacity. CRDB connection latency increases as connection rate increases.

This PR checks the connection when we acquire it from the pool. If the node is draining, the connection is closed and we grab a new one instead. This allows us to extend lifetimes past 5m and still handle node drains.

This has a negative performance impact on CRDB at low scales - detecting when a node is draining requires another roundtrip, and this comes through in the benchmarks:

```go
goos: darwin
goarch: arm64
pkg: github.com/authzed/spicedb/internal/datastore/benchmark
                                                                       │   old.txt    │                new.txt                 │
                                                                       │    sec/op    │    sec/op      vs base                 │
DatastoreDriver/cockroachdb-overlap-static/TestTuple/SnapshotRead-10     866.3µ ± 12%   1720.0µ ±  9%   +98.56% (p=0.000 n=10)
DatastoreDriver/cockroachdb-overlap-static/TestTuple/Touch-10            11.93m ±  7%    11.56m ±  6%         ~ (p=0.165 n=10)
DatastoreDriver/cockroachdb-overlap-static/TestTuple/Create-10           13.10m ± 12%    12.89m ±  5%         ~ (p=0.353 n=10)
DatastoreDriver/cockroachdb-overlap-insecure/TestTuple/SnapshotRead-10   876.6µ ±  9%   1807.8µ ±  7%  +106.23% (p=0.000 n=10)
DatastoreDriver/cockroachdb-overlap-insecure/TestTuple/Touch-10          4.223m ± 10%    5.215m ±  9%   +23.49% (p=0.000 n=10)
DatastoreDriver/cockroachdb-overlap-insecure/TestTuple/Create-10         3.834m ±  9%    4.699m ± 15%   +22.56% (p=0.000 n=10)
geomean                                                                  3.526m          4.741m         +34.47%

                                                                       │   old.txt    │               new.txt                │
                                                                       │     B/op     │     B/op      vs base                │
[...]

PR #1284, opened 2023-04-25
Determine connection drain status async

ecordell:
Note: haven't seen this survive node drains as expected, not ready for review yet. See https://github.com/authzed/spicedb/pull/1283

This starts a worker that periodically (or on any connection error) asks Cockroach if any of its nodes are draining. If draining nodes are found, a shared `draining` flag is set, and only then do we check if connections are draining on acquire. 

The goal is to remove the roundtrip from the common case, while still being able to make use of it when nodes are draining.

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


ecordell:
Closing in favor of #1294

PR #1285, opened 2023-04-26
Crossfade revisions when a new optimized revision is picked

### LINKED ISSUE #640: Add revision fuzzing for picking optimized revisions
Right now when a new optimized revision becomes the de-facto choice, the entire existing cache is (practically) simultaneously invalidated. In order to decrease the effect of this cutover, we should probably phase between the outgoing optimized revision and the incoming optimized revision over some period of time. Other revision picking logic, such as `AtLeastAsFresh` consistency happens *after* the optimized revision picking and should therefore be unaffected.

jakedt:
Fixes #640

### DIFF
--- a/internal/datastore/common/revisions/optimized_test.go
+++ b/internal/datastore/common/revisions/optimized_test.go
@@ -123,12 +123,22 @@ func TestOptimizedRevisionCache(t *testing.T) {
 			defer cancel()
 
 			for _, expectedRevSet := range tc.expectedRevisions {
-				revision, err := or.OptimizedRevision(ctx)
-				require.NoError(err)
-				printableRevSet := lo.Map(expectedRevSet, func(val datastore.Revision, index int) string {
-					return val.String()
-				})
-				require.Contains(expectedRevSet, revision, "must return the proper revision, allowed set %#v, received %s", printableRevSet, revision)
+				awaitingRevisions := make(map[datastore.Revision]struct{}, len(expectedRevSet))
+				for _, rev := range expectedRevSet {
+					awaitingRevisions[rev] = struct{}{}
+				}
+
+				require.Eventually(func() bool {
+					revision, err := or.OptimizedRevision(ctx)
+					require.NoError(err)
+					printableRevSet := lo.Map(expectedRevSet, func(val datastore.Revision, index int) string {
+						return val.String()
+					})
+					require.Contains(expectedRevSet, revision, "must return the proper revision, allowed set %#v, received %s", printableRevSet, revision)
+
+					delete(awaitingRevisions, revision)
+					return len(awaitingRevisions) == 0
+				}, 1*time.Second, 1*time.Microsecond)
 
 				mockTime.Add(5 * time.Millisecond)
 			}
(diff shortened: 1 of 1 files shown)


internal/datastore/common/revisions/optimized.go
```diff
@@ -43,45 +41,71 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
...
+	// Prune the candidates that have definitely expired
+	var numToDrop uint
+	for _, candidate := range cor.candidates {
+		if candidate.validThrough.Add(cor.maxRevisionStaleness).Before(localNow) {
+			numToDrop++
+		} else {
+			break
+		}
+	}
+	cor.candidates = cor.candidates[numToDrop:]
```
ecordell, line 56: did you envision allowing a maxStaleness > 1? 

I imagine typically it will be between 0 and 1, in which case there will only ever be two candidates. We could probably have less locking by just storing two atomic revisions and picking between them.

jakedt: I did write it to be able to support more than two, I start with a `current` and `last`, but in the spirit of writing code that handles 0, 1, or N, I thought I would generalize it.

internal/datastore/common/revisions/optimized.go
```diff
@@ -43,45 +41,71 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
...
 	if err != nil {
 		return datastore.NoRevision, err
 	}
-	return lastQuantizedRevision.(datastore.Revision), err
+	return newQuantizedRevision.(datastore.Revision), err
 }
 
 // CachedOptimizedRevisions does caching and deduplication for requests for optimized revisions.
 type CachedOptimizedRevisions struct {
+	sync.Mutex
```
ecordell, line 100: I came across https://pkg.go.dev/github.com/puzpuzpuz/xsync/v2#RBMutex recently which based on the docs + benchmarks should have less contention for readers. We could also write our own benchmarks for optimized revision to see if it helps. 

Maybe this is premature, but every request goes through this path.

internal/datastore/common/revisions/optimized_test.go
```diff
@@ -118,10 +122,13 @@ func TestOptimizedRevisionCache(t *testing.T) {
...
 
-			for _, expectedRev := range tc.expectedRevisions {
+			for _, expectedRevSet := range tc.expectedRevisions {
 				revision, err := or.OptimizedRevision(ctx)
 				require.NoError(err)
-				require.True(expectedRev.Equal(revision), "must return the proper revision %s != %s", expectedRev, revision)
+				printableRevSet := lo.Map(expectedRevSet, func(val datastore.Revision, index int) string {
+					return val.String()
+				})
+				require.Contains(expectedRevSet, revision, "must return the proper revision, allowed set %#v, received %s", printableRevSet, revision)
```
ecordell, line 131: Could we tweak this test to show that all of the candidates actually do get selected some % of the time?

ecordell (changes requested):

internal/datastore/common/revisions/optimized.go
jakedt: In actual tests this seemed to make it slower.

jakedt:
Updated

ecordell (approved):
LGTM

PR #1300, opened 2023-05-05
internal/graph: use a dispatcher without goroutines at concurrency 1

vroldanbet:


josephschorr:
Handled by https://github.com/authzed/spicedb/commit/a3b0cac15a03e6ad1aef325f705589594e54f5c9

PR #1302, opened 2023-05-05
Fix serialization of custom types in caveat context

### LINKED ISSUE #1299: could not serialize context: proto: invalid type: types.IPAddress. please report this error
I am trying to develop an IP range check and I am using this as a base: https://authzed.com/docs/reference/caveats#ip-address-checking

Since I couldn't find a Playground example, I created one myself, based on your unit test https://github.com/authzed/spicedb/blob/main/pkg/caveats/types_test.go

Here is the Playground link: https://play.authzed.com/s/FGmHO5fqLhH8/assertions

When I try to run the assertion or generate expected relations, I get the following error:
`could not serialize context: proto: invalid type: types.IPAddress. please report this error`

I only need a working example in the Playground.

Cheers.

josephschorr:
We need to have another method on a shared interface to enable this to be generic and be enforced

Also adds some tests

Fixed #1299

### DIFF
--- a/pkg/caveats/types/ipaddress.go
+++ b/pkg/caveats/types/ipaddress.go
@@ -33,6 +33,10 @@ type IPAddress struct {
 	ip netip.Addr
 }
 
+func (ipa IPAddress) SerializedString() string {
+	return ipa.ip.String()
+}
+
 func (ipa IPAddress) ConvertToNative(typeDesc reflect.Type) (interface{}, error) {
 	switch typeDesc {
 	case reflect.TypeOf(""):
@@ -67,7 +71,7 @@ func (ipa IPAddress) Value() interface{} {
 	return ipa
 }
 
-var IPAddressType = registerCustomType(
+var IPAddressType = registerCustomType[IPAddress](
 	"ipaddress",
 	cel.ObjectType("IPAddress"),
 	func(value any) (any, error) {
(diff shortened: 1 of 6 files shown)


pkg/caveats/types/registration.go
```diff
@@ -80,7 +80,7 @@ func registerGenericType(
 }
 
 // registerCustomType registers a custom type that wraps a base CEL type.
-func registerCustomType(keyword string, baseCelType *cel.Type, converter typedValueConverter, opts ...cel.EnvOption) VariableType {
+func registerCustomType[T CustomType](keyword string, baseCelType *cel.Type, converter typedValueConverter, opts ...cel.EnvOption) VariableType {
```
vroldanbet, line 83: I believe this generic is not being used and can be dropped

pkg/caveats/types/ipaddress.go
```diff
@@ -67,7 +71,7 @@ func (ipa IPAddress) Value() interface{} {
 	return ipa
 }
 
-var IPAddressType = registerCustomType(
+var IPAddressType = registerCustomType[IPAddress](
```
vroldanbet, line 74: the generic type here could be dropped too

vroldanbet (commented):
LGTM, just left a suggestion

pkg/caveats/types/registration.go
josephschorr: It is explicitly added to type check against the interface, to ensure it meets the `CustomType` requirements

pkg/caveats/types/ipaddress.go
josephschorr: See above

vroldanbet (approved):

PR #1304, opened 2023-05-08
Skip initializing the telemetry collector if telemetry is disabled

josephschorr:


pkg/cmd/server/server.go
```diff
@@ -390,27 +392,32 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
...
+	} else if c.TelemetryEndpoint != "" {
+		log.Ctx(ctx).Debug().Msg("initializing telemetry collector")
+		registry, err := telemetry.RegisterTelemetryCollector(c.DatastoreConfig.Engine, ds)
 		if err != nil {
-			return nil, fmt.Errorf("unable to initialize metrics reporter: %w", err)
+			log.Ctx(ctx).Warn().Err(err).Msg("unable to initialize telemetry collector")
+		}
+		telemetryRegistry = registry
+
+		if telemetryRegistry != nil {
```
ecordell, line 410: Is this part needed? why not just return on the err on line 405 and then you wont reach the remote reporter init if the registry is nil.

josephschorr: I assumed there were cases where it might be nil even without an error

ecordell: Nope: https://github.com/josephschorr/spicedb/blob/4f89f60bd0ebe154d09ae23a19e055c52180a51d/internal/telemetry/metrics.go#L24-L94

josephschorr: Okay; updated

ecordell (approved):
LGTM

PR #1306, opened 2023-05-09
Add support for limited deletion of relationships

josephschorr:
Following this change, a caller can specify a limit on the number of relationships to be deleted. If the number of relationships to be deleted is in excess of the limit, then an error will be returned, unless partial deletion is requested

### DIFF
--- a/internal/services/v1/errors.go
+++ b/internal/services/v1/errors.go
@@ -177,6 +177,62 @@ func (err ErrDuplicateRelationshipError) GRPCStatus() *status.Status {
 	)
 }
 
+// ErrCouldNotTransactionallyDelete indicates that a deletion could not occur transactionally.
+type ErrCouldNotTransactionallyDelete struct {
+	error
+	limit  uint32
+	filter *v1.RelationshipFilter
+}
+
+// NewCouldNotTransactionallyDeleteErr constructs a new could not transactionally deleter error.
+func NewCouldNotTransactionallyDeleteErr(filter *v1.RelationshipFilter, limit uint32) ErrCouldNotTransactionallyDelete {
+	return ErrCouldNotTransactionallyDelete{
+		error: fmt.Errorf(
+			"found more than %d relationships to be deleted and partial deletion was not requested",
+			limit,
+		),
+		limit:  limit,
+		filter: filter,
+	}
+}
+
+// GRPCStatus implements retrieving the gRPC status for the error.
+func (err ErrCouldNotTransactionallyDelete) GRPCStatus() *status.Status {
+	metadata := map[string]string{
+		"limit":                strconv.Itoa(int(err.limit)),
+		"filter_resource_type": err.filter.ResourceType,
+	}
+
+	if err.filter.OptionalResourceId != "" {
+		metadata["filter_resource_id"] = err.filter.OptionalResourceId
+	}
+
+	if err.filter.OptionalRelation != "" {
+		metadata["filter_relation"] = err.filter.OptionalRelation
+	}
+
+	if err.filter.OptionalSubjectFilter != nil {
+		metadata["filter_subject_type"] = err.filter.OptionalSubjectFilter.SubjectType
+
+		if err.filter.OptionalSubjectFilter.OptionalSubjectId != "" {
+			metadata["filter_subject_id"] = err.filter.OptionalSubjectFilter.OptionalSubjectId
+		}
+
+		if err.filter.OptionalSubjectFilter.OptionalRelation != nil {
+			metadata["filter_subject_relation"] = err.filter.OptionalSubjectFilter.OptionalRelation.Relation
+		}
+	}
+
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.InvalidArgument,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_TOO_MANY_RELATIONSHIPS_FOR_TRANSACTIONAL_DELETE,
+			metadata,
+		),
+	)
+}
+
 func defaultIfZero[T comparable](value T, defaultValue T) T {
 	var zero T
 	if value == zero {
--- a/internal/services/v1/relationships.go
+++ b/internal/services/v1/relationships.go
@@ -275,10 +275,45 @@ func (ps *permissionServer) DeleteRelationships(ctx context.Context, req *v1.Del
 	}
 
 	ds := datastoremw.MustFromContext(ctx)
+	deletionProgress := v1.DeleteRelationshipsResponse_DELETION_PROGRESS_COMPLETE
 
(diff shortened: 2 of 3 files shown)


internal/services/v1/relationships.go
```diff
@@ -290,13 +325,18 @@ func (ps *permissionServer) DeleteRelationships(ctx context.Context, req *v1.Del
 			return err
 		}
 
+		if len(deleteMutations) > 0 {
+			return rwt.WriteRelationships(ctx, deleteMutations)
+		}
```
vroldanbet, line 330: This strategy is fine for now, but it requires to read all relationships and then issue write. Why wasn't `DELETE ... LIMIT` considered? If we used the relationship filter, it would provide a stable order to delete for. Was it because it wouldn't be possible to know if there are further rows to delete?

I also wondered if all the datastores we use are configured to run with serializable isolation level (I recall some proposals to relax PG). I think this would be relevant because it could be subject to phantom reads problem - you read 1000 rels, but some other transaction deleted 3 rels of your same set. Effectively when `WriteRelationships` is run, you've deleted 997 rels, thus not honored the limit

internal/services/v1/errors.go
```diff
@@ -177,6 +177,62 @@ func (err ErrDuplicateRelationshipError) GRPCStatus() *status.Status {
...
+			"found more than %d relationships to be deleted and partial deletion was not requested",
+			limit,
+		),
+		limit:  limit,
+		filter: filter,
+	}
+}
+
+// GRPCStatus implements retrieving the gRPC status for the error.
+func (err ErrCouldNotTransactionallyDelete) GRPCStatus() *status.Status {
```
vroldanbet, line 200: I almost feel like asking for a unit test. It's pretty simple logic, but regressions here means API breakage

internal/services/v1/relationships.go
```diff
@@ -290,13 +325,18 @@ func (ps *permissionServer) DeleteRelationships(ctx context.Context, req *v1.Del
 			return err
 		}
 
+		if len(deleteMutations) > 0 {
+			return rwt.WriteRelationships(ctx, deleteMutations)
```
vroldanbet, line 329: can you please add an OpenTelemetry log entry to the trace indicating the number of elements to be removed? Or is that visible already downstream via `WriteRelationships`?

internal/services/v1/relationships_test.go
```diff
@@ -982,6 +984,86 @@ func TestDeleteRelationships(t *testing.T) {
...
+
+			rev, err := zedtoken.DecodeRevision(resp.DeletedAt, ds)
+			require.NoError(err)
+			require.True(rev.GreaterThan(revision))
+			require.EqualValues(standardTuplesWithout(expected), readAll(require, client, resp.DeletedAt))
+			break
+		}
+	}
+
+	require.Less(iterations, 10)
```
vroldanbet, line 1064: Why such a loose check? We have 11 rels, and are deleting them in batches of 5. That's 3 calls, so let's check we iterated 3 times

internal/services/v1/relationships_test.go
```diff
@@ -982,6 +984,86 @@ func TestDeleteRelationships(t *testing.T) {
...
+			OptionalLimit:                 5,
+			OptionalAllowPartialDeletions: true,
+		})
+		require.NoError(err)
+
+		headRev, err = ds.HeadRevision(context.Background())
+		require.NoError(err)
+
+		afterDelete := readOfType(require, "document", client, zedtoken.MustNewFromRevision(headRev))
+		require.LessOrEqual(len(beforeDelete)-len(afterDelete), 5)
```
vroldanbet, line 1046: If this has 11 rels, it will be deleted in 3 iterations
- the first batch is tested with `i=0`
- the second batch is not tested, because it's partial, but `i != 0`
- the third batch is tested because deletion is complete

we could change the if calsuse below to check on this instead
```go
resp.DeletionProgress == v1.DeleteRelationshipsResponse_DELETION_PROGRESS_PARTIAL
```

then you can require for the partial case:
```go
require.Equal(len(beforeDelete)-len(afterDelete), 5)
```
and for the complete case
```go
require.Equal(len(beforeDelete)-len(afterDelete), 1)
```

internal/services/v1/relationships_test.go
```diff
@@ -982,6 +984,86 @@ func TestDeleteRelationships(t *testing.T) {
 	}
 }
 
+func TestDeleteRelationshipsBeyondLimit(t *testing.T) {
```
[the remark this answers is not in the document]
josephschorr: That's enforced in the Validate method on incoming calls

internal/services/v1/relationships.go
josephschorr: `DELETE ... LIMIT` is not supported on some datastores (such as MySQL)

This is all running under a transaction, so it should be safe, count wise

internal/services/v1/errors.go
josephschorr: We have a unit test for this: this error is checked in one of the service tests

internal/services/v1/relationships.go
josephschorr: It should be in the downstream call in the datastore, yes

internal/services/v1/relationships_test.go
josephschorr: Because I figured it was flexible if/when we change the test data (which happens from time to time). I can also make it strict, but this seemed simpler

internal/services/v1/relationships_test.go
josephschorr: Until we change the test data and all of this has to be modified, vs the above which will work for any additional data added just by changing the expected-to-be-deleted map

internal/services/v1/relationships.go
```diff
@@ -275,12 +275,47 @@ func (ps *permissionServer) DeleteRelationships(ctx context.Context, req *v1.Del
...
+	deletionProgress := v1.DeleteRelationshipsResponse_DELETION_PROGRESS_COMPLETE
 
 	revision, err := ds.ReadWriteTx(ctx, func(rwt datastore.ReadWriteTransaction) error {
 		if err := ps.checkFilterNamespaces(ctx, req.RelationshipFilter, rwt); err != nil {
 			return err
 		}
 
+		var deleteMutations []*core.RelationTupleUpdate
+
+		if req.OptionalLimit > 0 {
```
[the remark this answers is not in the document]
josephschorr: Added test

internal/services/v1/relationships_test.go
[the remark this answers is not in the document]
vroldanbet: but is there a test for it?

vroldanbet (approved):

PR #1308, opened 2023-05-10
Add better logging of config

josephschorr:
Logs the full (but sanitized) configuration on startup and provides access with /debug/config


pkg/cmd/util/util.go
```diff
@@ -295,37 +297,37 @@ func (d *disabledGrpcServer) NetDialContext(_ context.Context, _ string) (net.Co
...
 
 type HTTPServerConfig struct {
-	Address     string
-	TLSCertPath string
-	TLSKeyPath  string
-	Enabled     bool
+	HTTPAddress     string `debugmap:"visible"`
+	HTTPTLSCertPath string `debugmap:"visible"`
+	HTTPTLSKeyPath  string `debugmap:"visible"`
+	HTTPEnabled     bool   `debugmap:"visible"`
```
vroldanbet, line 303: why was the prefix needed?

josephschorr: Because over method name overlap since we're generating this and the grpc struct in the same module

pkg/cmd/server/server.go
```diff
@@ -58,75 +59,75 @@ var ConsistentHashringPicker = balancer.NewConsistentHashringPickerBuilder(
...
 type Config struct {
 	// API config
-	GRPCServer             util.GRPCServerConfig
-	GRPCAuthFunc           grpc_auth.AuthFunc
-	PresharedKey           []string
-	ShutdownGracePeriod    time.Duration
-	DisableVersionResponse bool
+	GRPCServer             util.GRPCServerConfig `debugmap:"visible"`
+	GRPCAuthFunc           grpc_auth.AuthFunc    `debugmap:"visible"`
+	PresharedSecureKey     []string              `debugmap:"sensitive"`
```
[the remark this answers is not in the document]
josephschorr: Make it clear it is a security-related item, which means optgen can require that the field is marked as `sensitive`

pkg/cmd/util/util.go
vroldanbet: It that's the case I think we should probably add `GRPC` prefix to `GRPCServerConfig` fields to keep everything homogeneous?

vroldanbet (approved):

vroldanbet (approved):

PR #1312, opened 2023-05-11
Use UPSERT for TOUCH in Postgres

### LINKED ISSUE #901: Proposal: Use upsert for TOUCH operations in supported datastores
Right now in datastores that do manual MVCC tracking (e.g. Postgres and MySQL), when a user does a `TOUCH` operation we always issue two queries: one to end the lifetime of any existing matching relationships, and one start a new lifetime for the relationship in its final state. This is an idempotent operation, and the end state for the datastore will be that the relationship now exists, whether it existed before or not.

```
{a:b#c@d:e[123:-]} -> {a:b#c@d:e[123:456], a:b#c@d:e[456:~]}
{} -> {a:b#c@d:e[456:~]}
```

This also gives us the ability to detect and report (through the Watch API) when a TOUCH operation was submitted by detecting the ending of one lifetime and the beginning of another.

The **downsides** of doing `TOUCH` this way are:
- we have to issue two queries instead of just one
- it creates more garbage that must be garbage collected
- the delete query is potentially causing more transactions to be identified as overlapping than necessary

I propose that we start using upsert semantics (e.g. `INSERT ... ON CONFLICT`) to do these writes to eliminate the downsides above. 

```
{a:b#c@d:e[123:-]} -> {a:b#c@d:e[123:~]}
{} -> {a:b#c@d:e[456:~]}
```

Without any additional work we would lose the ability to emit `TOUCH` events because we don't have the matching created and deleted transaction to key off of. Because it doesn't change the observable state of the datastore at any revision, it is hard to justify why emitting these touch events is strictly necessary. If this were deemed unacceptable we could potentially track successful upserts separately in a changelog table of some sort.

josephschorr:
This change makes use of UPSERT semantics for TOUCH operations in Postgres where applicable.

It does so by changing the WriteRelationships operation to the following:

1) Run CREATE operations as INSERT as we do today; any unique constraint failures are errors as they are today 2) Run TOUCH operations as INSERT with ON CONFLICT DO NOTHING and RETURNING all inserted rows. This will INSERT any relationships that are not currently present, while skipping those that would cause a conflict. The RETURNING allows the caller to track which relationships were successfully inserted 3) Run DELETE for each DELETE operation, *as well as any TOUCH operation which did not succeed and has changed caveats*. As we only need to "replace" a relationship on TOUCH if its caveat has changed, this ensures that only those relationships are deleted. This also makes use of a RETURNING to track which TOUCHed relationships were removed. 4) If any relationships for TOUCH were removed due to caveat name and/or context change, then INSERT those relationships with the updated caveat information to bring us to the desired state

Fixes #901 for postgres

Benchmarks, where `0`, `0.1`, etc represent the percentage of `CREATE` operations vs `TOUCH` operations (with `1` being 100% `CREATE`):
```
name                                                        old time/op    new time/op   
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0_-10        136ms           46ms     
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.1_-10      141ms           38ms     
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.25_-10     148ms           36ms     
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.5_-10      133ms           38ms     
DatastoreDriver/postgres/TestTuple/CreateAndTouch/1_-10       31.7ms         23.8ms     

name                                                        old alloc/op   new alloc/op   
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0_-10       9.11MB        12.35MB        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.1_-10     8.80MB        11.75MB        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.25_-10    7.92MB        10.40MB        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.5_-10     7.03MB         8.38MB        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/1_-10       5.18MB         5.26MB        

name                                                        old allocs/op  new allocs/op  
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0_-10         103k           148k        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.1_-10      97.9k         138.3k        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.25_-10     90.9k         124.5k        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/0.5_-10      78.9k         101.3k        
DatastoreDriver/postgres/TestTuple/CreateAndTouch/1_-10        55.0k          55.0k        
```


internal/datastore/postgres/readwrite.go
```diff
@@ -53,69 +57,220 @@ type pgReadWriteTXN struct {
...
 
+func appendForInsertion(builder sq.InsertBuilder, tpl *core.RelationTuple) sq.InsertBuilder {
+	var caveatName string
+	var caveatContext map[string]any
+	if tpl.Caveat != nil {
+		caveatName = tpl.Caveat.CaveatName
+		caveatContext = tpl.Caveat.Context.AsMap()
+	}
+
+	valuesToWrite := []interface{}{
```
jzelinskie, line 68:
```go
[]any{
    ...
}
```

jakedt:
Moved to draft.

#901 is a proposal, and we shouldn't commit to changing the semantics for one datastore until 1) the proposal is accepted, and 2) we're sure we can do it for all datastore impls.

josephschorr:
> we're sure we can do it for all datastore impls.

I don't believe we should block one datastore's improvements on another's; in this case, it seems the win is clear for Postgres and we can likely followup with CRDB.

jakedt:
There are other parts of the semantics discussed in the issue, namely that touches will no longer be emitted by watch. That's what we need agreement on, it's not just the perf.

josephschorr:
> There are other parts of the semantics discussed in the issue, namely that touches will no longer be emitted by watch. That's what we need agreement on, it's not just the perf.

In my opinion, having watch emit touches that made no changes does not provide value to any consumers of the API; in fact, I'd have argued they shouldn't have been there in the first place, but as they were a consequence of how TOUCH was implemented until now, it wasn't a huge deal.

As there has been no pushback against their removal in the proposal (which has been outstanding for ~7 months), I'd argue it is safe to remove them unless we get a forthcoming comment indicating otherwise

jakedt:
I'm fine with removing them, as you say, the proposal hasn't had any pushback. But we then need to make sure they don't come out in the watch stream regardless of which engine you're using, which is why I want to see the semantics of all engines updated to match.

josephschorr:
> I'm fine with removing them, as you say, the proposal hasn't had any pushback. But we then need to make sure they don't come out in the watch stream regardless of which engine you're using, which is why I want to see the semantics of all engines updated to match.

Given that the events are, for better or for worse, superfluous, I don't necessarily agree that they need to be fully filtered in all data stores, so long as we clearly document that some datastores might emit them. 

I didn't change any of the other datastores because I wanted to minimize the changes within this PR, but if you believe its better to do so, I'll make commits for each other datastore

josephschorr:
All datastores now support the new semantics

pkg/datastore/test/tuples.go
```diff
@@ -799,3 +842,59 @@ func onrToSubjectsFilter(onr *core.ObjectAndRelation) datastore.SubjectsFilter {
 		RelationFilter:     datastore.SubjectRelationFilter{}.WithNonEllipsisRelation(onr.Relation),
 	}
 }
+
+func ensureTuples(ctx context.Context, require *require.Assertions, ds datastore.Datastore, tpls ...*core.RelationTuple) {
```
jakedt, line 846: Combine these functions into one or extract out the common part into a helper.

pkg/datastore/test/watch.go
```diff
@@ -215,3 +215,115 @@ func WatchCancelTest(t *testing.T, tester DatastoreTester) {
...
+	)
+
+	verifyNoUpdates(require,
+		changes,
+		errchan,
+		false,
+	)
+
+	// TOUCH the relationship again with a caveat change and ensure it does appear in the watch.
+	changes, errchan = ds.Watch(ctx, afterNoChangesRevision)
```
jakedt, line 282: You could use `afterTouchRevision` here to cover a more broad swath of time to really ensure that the change doesn't come out.

internal/datastore/spanner/options.go
```diff
@@ -16,6 +16,22 @@ type spannerOptions struct {
...
+
+	migrationPhase string
+}
+
+type migrationPhase uint8
+
+const (
+	writeChangelogReadChangelog migrationPhase = iota
+	writeChangelogReadStream
+	complete
```
jakedt, line 28: I think when I did this for postgres, I made the zero value for the migrationPhase be complete, so the lack of a value would be the end migration state and it would make it easier for people who had fully migrated to stop passing the value.

internal/datastore/spanner/watch.go
```diff
@@ -2,142 +2,169 @@ package spanner
...
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
 const (
-	watchSleep = 100 * time.Millisecond
+	RelationTupleChangeStreamName = "relation_tuple_stream"
 )
 
-var queryChanged = sql.Select(allChangelogCols...).From(tableChangelog)
+// Copied from the spanner library
```
jakedt, line 24: Can you provide a git sha/reference? What is the license?

pkg/datastore/test/watch.go
```diff
@@ -215,3 +215,115 @@ func WatchCancelTest(t *testing.T, tester DatastoreTester) {
...
+		tuple.Parse("document:firstdoc#viewer@user:fred[thirdcaveat]"),
+	)
+
+	verifyNoUpdates(require,
+		changes,
+		errchan,
+		false,
+	)
+
+	// TOUCH the relationship again with a caveat change and ensure it does appear in the watch.
```
jakedt, line 281: Add a test for mutating caveat context as well.

internal/datastore/postgres/readwrite.go
```diff
@@ -53,69 +57,220 @@ type pgReadWriteTXN struct {
...
+		return fmt.Errorf(errUnableToWriteRelationships, err)
+	}
+
+	rows, err := rwt.tx.Query(ctx, sql, args...)
+	if err != nil {
+		return fmt.Errorf(errUnableToWriteRelationships, err)
+	}
+	defer rows.Close()
+
+	if len(touchMutationsByNonCaveat) > 0 {
```
jakedt, line 226: Would it be safe to just let this run even if it's empty? I think I would prefer that to having huge chunks of code disabled with an `if`.

internal/datastore/postgres/readwrite.go
```diff
@@ -275,4 +430,32 @@ func exactRelationshipClause(r *core.RelationTuple) sq.Eq {
 	}
 }
 
+func exactRelationshipDiffentCaveatClause(r *core.RelationTuple) sq.And {
```
jakedt, line 433: Different

jakedt (changes requested):

pkg/datastore/test/watch.go
josephschorr: We expect a change for this watch

internal/datastore/spanner/options.go
josephschorr: I actually went back and copied this directly from the 1.14 code and then modified it. I'm fine switching up the values, but since we always read the string value, I think its fine

jakedt: Yep, I see now it's the map that converts the zero value of the string to the `complete` phase.

internal/datastore/postgres/readwrite.go
josephschorr: Changed, although I'm sad to leave performance on the table

josephschorr:
Rebased and updated

internal/datastore/postgres/readwrite.go
jakedt: My linter complains about this, but obviously the one in CI doesn't.

jakedt (approved):

PR #1315, opened 2023-05-12
reduces flakiness of servetesting integration test

vroldanbet:
uses `retry.Pool` which has an exponential backoff, instead of
a fixed timeout that caused the previous
implementation to not complete in slow compute
e.g. GitHub Action Runners


josephschorr (approved):
LGTM

josephschorr (approved):

josephschorr (approved):

PR #1318, opened 2023-05-14
add magefiles for running project commands

ecordell:
This adds a [magefile](https://magefile.org/) to provide common project tasks, and switches CI to run them where possible.

We've been testing mage in the operator repo for a while and it's really convenient:

- Any operations that the steps need to run can lean on common go libraries, no (separate) dependency management needed.
- Almost all additional tooling that we use is available as a go package, so they can be run from the current go module and prevent version skew from local/CI or between dev environments.
- It has some basic utilities for doing `make`-like operations if we need to (running dependent tasks, checking for file changes, etc), but is otherwise a better fit than the file transformations of make. 
- We can (but haven't yet) build a shared library of common utilities that we can use across our authzed repos (similar to `authzed/actions`, but available for local use too).
- There is a zero-install option (which we use in CI to avoid the extra time to install mage; there's not much point to using it locally)
 
```sh
$ mage
Targets:
  build:testimage      Build the spicedb image for tests
  build:wasm           Build the wasm bundle
  deps:tidy            go mod tidy all go modules
  deps:update          go get -u all go dependencies
  gen:all              Run all generators in parallel
  gen:go               Run go codegen
  gen:proto            Run proto codegen
  lint:all             Run all linters
  lint:analyzers       Run all analyzers
  lint:extra           Lint everything that's not code
  lint:go              Run all go linters
  lint:gofumpt         Run gofumpt
  lint:golangcilint    Run golangci-lint
  lint:markdown        Lint markdown
  lint:vulncheck       Run vulncheck
  lint:yaml            Lint yaml
  test:all             Runs all test suites
  test:analyzers       Run the analyzer unit tests
  test:image           Run tests that run the built image
  test:integration     Run integration tests
  test:unit            Runs the unit tests
  test:wasm            Run wasm browser tests
  testcons:crdb        Run consistency tests for crdb
  testcons:mysql       Run consistency tests for mysql
  testcons:postgres    Run consistency tests for postgres
  testcons:spanner     Run consistency tests for spanner
  testds:crdb          Run datastore tests for crdb
  testds:mysql         Run datastore tests for mysql
  testds:postgres      Run datastore tests for postgres
  testds:spanner       Run datastore tests for spanner
```

Review note: this changes one of the names of the required jobs, so when we're ready to merge we'll need to switch the required job name to match.


.gitattributes
```diff
@@ -1 +0,0 @@
-*.pb*.go -diff linguist-generated=true
```
josephschorr, line 1: why is this removed?

.golangci.yaml
```diff
@@ -12,7 +12,7 @@ linters-settings:
       - 'github.com/jackc/pgx'
   gosec:
     excludes:
-      - 'G404' # Allow the usage of math/rand
+      - 'G404'  # Allow the usage of math/rand
```
josephschorr, line 15: is there a way we can turn this lint rule off? VSCode auto-formats it away and its a constant pain

buf.gen.yaml
```diff
@@ -11,15 +11,19 @@ managed:
       - 'buf.build/googleapis/googleapis'
 plugins:
   - name: 'go'
+    path: ["go", "run", "google.golang.org/protobuf/cmd/protoc-gen-go"]
```
josephschorr, line 14: does this make buf use the version found in `go.mod`?

go.mod
```diff
@@ -53,142 +57,311 @@ require (
...
+	4d63.com/gocheckcompilerdirectives v1.2.1 // indirect
+	4d63.com/gochecknoglobals v0.2.1 // indirect
 	cloud.google.com/go v0.110.0 // indirect
 	cloud.google.com/go/compute v1.18.0 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
 	cloud.google.com/go/iam v0.12.0 // indirect
 	cloud.google.com/go/longrunning v0.4.1 // indirect
-	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
-	github.com/Microsoft/go-winio v0.6.0 // indirect
+	github.com/Abirdcfly/dupword v0.0.11 // indirect
```
josephschorr, line 99: why so many more indirect deps all of a sudden?

magefiles/deps.go
```diff
@@ -0,0 +1,37 @@
+//go:build mage
+
+package main
+
+import "github.com/magefile/mage/mg"
+
+type Deps mg.Namespace
+
+var goModules = []string{
```
josephschorr, line 9: move this above the type def

magefiles/lint.go
```diff
@@ -0,0 +1,98 @@
...
+type Lint mg.Namespace
+
+// Run all linters
+func (l Lint) All() error {
+	mg.Deps(l.Go, l.Extra)
+	return nil
+}
+
+// Lint everything that's not code
+func (l Lint) Extra() error {
```
josephschorr, line 22: maybe `NonCode`?

.gitignore
```diff
@@ -1,2 +1,3 @@
 dist/
-*.wasm
\ No newline at end of file
+*.wasm
+tools/analyzers/analyzers
```
josephschorr, line 3: Why is this added?

.gitattributes
ecordell: We have a lot of proto plugins, I think it's useful to see the diffs when they change

buf.gen.yaml
ecordell: yep!

go.mod
ecordell: I added more to the tools.go

magefiles/lint.go
ecordell: I based it off of the GHA name, but don't have a strong preference either way

.gitignore
ecordell: I wasn't able to figure out how to run the analyzers with `go run` with the right directory context, so instead it builds them and then runs them. This ignores the built binary

.golangci.yaml
ecordell: I can look

ecordell: I don't see one, unless I'm mising it? https://yamllint.readthedocs.io/en/stable/rules.html

.gitignore
ecordell: I was able to get the analyzers to run directly without pre-compiling, so I removed this line.

The only things left that aren't directly `go run` from the module are the yaml and markdown linter, which are docker containers.

.golangci.yaml
josephschorr: D'oh

josephschorr (approved):
LGTM

PR #1330, opened 2023-05-18
Add retries around spawning images in servetesting

josephschorr:
It appears that runners sometimes get into a weird state, so this will spawn a completely new server image if that happens


ecordell (approved):
LGTM

PR #1331, opened 2023-05-18
Force custom plans in Postgres

josephschorr:
Automatically sets plan_cache_mode='force_custom_plan' if not overridden in the database URI. This setting bypasses the plan cache to workaround a bug in the Postgres optimizer that can result in significant performance degradation.

For background: https://spicedb.dev/d/force-custom-plan

### DIFF
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -123,7 +123,7 @@ func NewPostgresDatastore(
 }
 
 func newPostgresDatastore(
-	url string,
+	pgURL string,
 	options ...Option,
 ) (datastore.Datastore, error) {
 	config, err := generateConfig(options)
@@ -131,31 +131,39 @@ func newPostgresDatastore(
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
 
-	if config.migrationPhase != "" {
-		log.Info().
-			Str("phase", config.migrationPhase).
-			Msg("postgres configured to use intermediate migration phase")
-	}
-
-	// config must be initialized by ParseConfig
-	readPoolConfig, err := pgxpool.ParseConfig(url)
+	// Parse the DB URI into configuration.
+	parsedConfig, err := pgxpool.ParseConfig(pgURL)
 	if err != nil {
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
+
+	// Setup the default custom plan setting, if applicable.
+	pgConfig, err := defaultCustomPlan(parsedConfig)
+	if err != nil {
+		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	}
+
+	// Setup the config for each of the read and write pools.
+	readPoolConfig := pgConfig.Copy()
 	config.readPoolOpts.ConfigurePgx(readPoolConfig)
+
 	readPoolConfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
 		RegisterTypes(conn.TypeMap())
 		return nil
 	}
 
-	writePoolConfig, err := pgxpool.ParseConfig(url)
-	if err != nil {
-		return nil, fmt.Errorf(errUnableToInstantiate, err)
-	}
+	writePoolConfig := pgConfig.Copy()
 	config.writePoolOpts.ConfigurePgx(writePoolConfig)
+
 	writePoolConfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
 		RegisterTypes(conn.TypeMap())
 		return nil
+	}
+
+	if config.migrationPhase != "" {
+		log.Info().
+			Str("phase", config.migrationPhase).
+			Msg("postgres configured to use intermediate migration phase")
 	}
 
 	initializationContext, cancelInit := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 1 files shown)


internal/datastore/postgres/postgres.go
```diff
@@ -123,22 +124,45 @@ func NewPostgresDatastore(
...
+	if err != nil {
+		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	}
+
+	if existing, ok := parsed.ConnConfig.Config.RuntimeParams["plan_cache_mode"]; ok {
+		log.Info().Str("plan_cache_mode", existing).Msg("found plan_cache_mode in DB URL; leaving as-is")
+	} else {
+		parsedURL, err := url.Parse(pgURL)
+		if err == nil {
+			log.Warn().Msg("could not parse URL to automatically add `plan_cache_mode=force_custom_plans`; likely an old-style DSN")
```
jzelinskie, line 147: `log.Warn().Err(err).Msg(...)`

internal/datastore/postgres/postgres.go
```diff
@@ -123,22 +124,45 @@ func NewPostgresDatastore(
...
+	// See: https://spicedb.dev/d/force-custom-plan for more information.
+	parsed, err := pgxpool.ParseConfig(pgURL)
+	if err != nil {
+		return nil, fmt.Errorf(errUnableToInstantiate, err)
+	}
+
+	if existing, ok := parsed.ConnConfig.Config.RuntimeParams["plan_cache_mode"]; ok {
+		log.Info().Str("plan_cache_mode", existing).Msg("found plan_cache_mode in DB URL; leaving as-is")
+	} else {
+		parsedURL, err := url.Parse(pgURL)
```
jzelinskie, line 145: Why do we have to handle this case? Does pgx not handle both URI types?

internal/datastore/postgres/postgres.go
```diff
@@ -123,22 +124,45 @@ func NewPostgresDatastore(
...
+		if err == nil {
+			log.Warn().Msg("could not parse URL to automatically add `plan_cache_mode=force_custom_plans`; likely an old-style DSN")
+		}
+
+		queryParams := parsedURL.Query()
+		queryParams.Add("plan_cache_mode", "force_custom_plan")
+		parsedURL.RawQuery = queryParams.Encode()
+
+		pgURL = parsedURL.String()
+		log.Info().Msg("set plan_cache_mode='force_custom_plan' in Postgres DB URL")
```
jzelinskie, line 155: I think all of the logging related to this should include the spicedb.dev URL.
All spicedb.dev URLs should probably be stored under the same key (maybe "details_url"?).

jzelinskie (changes requested):

internal/datastore/postgres/postgres.go
josephschorr: PGX does, but we don't and PGX doesn't provide a way that i could find to modify the URL

josephschorr:
Updated

internal/datastore/postgres/postgres.go
jzelinskie: why are we even modifying the URL when we could just wrap this logic up into the config parser:

```go
func ParseConfig(connString string) (*ConnConfig, error) {
	config, err := pgxpool.ParseConfig(connString)
	if err != nil {
		return config, err
	}
	
	if existing, ok := config.ConnConfig.Config.RuntimeParams["plan_cache_mode"]; ok {
		log.Info().Debug("plan_cache_mode", existing).Msg("found plan_cache_mode in DB URI; leaving as-is")
	} else {
		config.ConnConfig.Config.RuntimeParams["plan_cache_mode"] = "force_custom_plan"
		log.Warn().
			Str("details-url", "https://spicedb.dev/d/force-custom-plan").
			Str("plan_cache_mode", "force_custom_plan").
			Msg("defaulting value in Postgres DB URI") // this wording in case we add more defaults in the future
	}

	return config, nil
}

readPoolConfig, err := ParseConfig(pgURL)
...
writePoolConfig, err := ParseConfig(pgURL)


```

jzelinskie (changes requested):

internal/datastore/postgres/postgres.go
josephschorr: I'm not certain that PGX passes `RuntimeParams` modifications to Postgres; I read through the code and didn't see it do so (unless I missed it). That's why I changed it to modify the URL

jzelinskie: it does, here: https://github.com/jackc/pgx/blob/70a200cff4d42098786c3649b2f3cd30b7a66f1b/pgconn/pgconn.go#L299-L302

so let's go with something along these lines:

```go
// defaultCustomPlan parses a Postgres URI and determines if a plan_cache_mode
// has been specified. If not, it defaults to "force_custom_plan".
// This worksaround a bug impacting performance documented here:
// https://spicedb.dev/d/force-custom-plan for more information.
func defaultCustomPlan(config *pgx.ConnConfig, err error) (*pgx.ConnConfig, error) {
	if err != nil {
		return config, err
	}

	if existing, ok := config.ConnConfig.Config.RuntimeParams["plan_cache_mode"]; ok {
		log.Info().
			Debug("plan_cache_mode", existing).
			Msg("found plan_cache_mode in DB URI; leaving as-is")
		return
	}

	config.ConnConfig.Config.RuntimeParams["plan_cache_mode"] = "force_custom_plan"
	log.Warn().
		Str("details-url", "https://spicedb.dev/d/force-custom-plan").
		Str("plan_cache_mode", "force_custom_plan").
		Msg("defaulting value in Postgres DB URI")

	return config, nil
}

...

pgConfig, err := defaultCustomPlan(pgxpool.ParseConfig(pgURL))
if err != nil {
}


readPoolConfig, err := pgConfig.Copy()
...
writePoolConfig, err := pgConfig.Copy()
```

josephschorr:
Updated as discussed

jzelinskie (approved):

jzelinskie (approved):

PR #1337, opened 2023-05-22
MySQL: nullable information_schema.tables columns and add test-suite for MySQL8

vroldanbet:
A user reported in Discord the following error

```
unable to query DB stats: sql: Scan error on column index 0, name "TABLE_ROWS": converting NULL to uint64 is unsupported
````

Some `information_schema.tables` columns are nullable, including `table_rows` used for SpiceDB statistics reported in telemetry. Depending on the configuration, specially in MySQL8, those values may not be populate and causes SpiceDB server to fail to start.

This changes the code to use `NullInt64`, and also runs the entire test-suite with V8.

### DIFF
--- a/internal/datastore/mysql/datastore_test.go
+++ b/internal/datastore/mysql/datastore_test.go
@@ -94,10 +94,30 @@ func TestMySQLDatastoreDSNWithoutParseTime(t *testing.T) {
 	require.ErrorContains(t, err, "https://spicedb.dev/d/parse-time-mysql")
 }
 
-func TestMySQLDatastore(t *testing.T) {
+func TestMySQL5Datastore(t *testing.T) {
 	b := testdatastore.RunMySQLForTesting(t, "")
 	dst := datastoreTester{b: b, t: t}
 	test.All(t, test.DatastoreTesterFunc(dst.createDatastore))
+	additionalMySQLTests(t, b)
+}
+
+func TestMySQL8Datastore(t *testing.T) {
+	b := testdatastore.RunMySQLForTestingWithOptions(t, testdatastore.MySQLTesterOptions{MigrateForNewDatastore: true, UseV8: true}, "")
+	dst := datastoreTester{b: b, t: t}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: internal/datastore/mysql/datastore_test.go [hunk 1]
```diff
@@ -90,14 +90,38 @@ func createDatastoreTest(b testdatastore.RunningEngineForTest, tf datastoreTestF
...
+
+func TestMySQL5DatastoreWithTablePrefix(t *testing.T) {
+	t.Parallel()
+	b := testdatastore.RunMySQLForTestingWithOptions(t, testdatastore.MySQLTesterOptions{MigrateForNewDatastore: true, Prefix: "spicedb_"}, "")
+	dst := datastoreTester{b: b, t: t, prefix: "spicedb_"}
+	test.All(t, test.DatastoreTesterFunc(dst.createDatastore))
+}
+
+func additionalMySQLTests(t *testing.T, b testdatastore.RunningEngineForTest) {
+	reg := prometheus.NewRegistry()
```
josephschorr, line 122: why do we need to register a new registry?

internal/datastore/mysql/datastore_test.go
```diff
@@ -90,14 +90,38 @@ func createDatastoreTest(b testdatastore.RunningEngineForTest, tf datastoreTestF
...
+func TestMySQL5Datastore(t *testing.T) {
+	t.Parallel()
 	b := testdatastore.RunMySQLForTesting(t, "")
 	dst := datastoreTester{b: b, t: t}
 	test.All(t, test.DatastoreTesterFunc(dst.createDatastore))
+	additionalMySQLTests(t, b)
+}
+
+func TestMySQL8Datastore(t *testing.T) {
+	t.Parallel()
```
josephschorr, line 107: I have concerns about parallel against running containers; are we sure this is stable?

vroldanbet: I'll remove it

josephschorr (approved):
LGTM

PR #1340, opened 2023-05-22
Add support for cursors on read relationships

### LINKED ISSUE #1339: Support cursors on read relationships
From #1163:

> The ReadRelationships API provides a streaming response, but no mechanism to "restart" reading from where a previous invocation left off. Providing a "start cursor" would enable a large ReadRelationships operation to be resumed from the application side.

josephschorr:
Fixes #1339

### DIFF
--- a/internal/services/v1/relationships.go
+++ b/internal/services/v1/relationships.go
@@ -21,6 +21,7 @@ import (
 	"github.com/authzed/spicedb/internal/namespace"
 	"github.com/authzed/spicedb/internal/relationships"
 	"github.com/authzed/spicedb/internal/services/shared"
+	"github.com/authzed/spicedb/pkg/cursor"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/datastore/options"
 	"github.com/authzed/spicedb/pkg/datastore/pagination"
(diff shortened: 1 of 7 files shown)


internal/services/v1/relationships.go
```diff
@@ -148,12 +149,47 @@ func (ps *permissionServer) ReadRelationships(req *v1.ReadRelationshipsRequest,
...
+
+	if req.OptionalCursor != nil {
+		decodedCursor, err := cursor.DecodeToDispatchCursor(req.OptionalCursor, rrRequestHash)
+		if err != nil {
+			return shared.RewriteError(ctx, err)
+		}
+
+		if len(decodedCursor.Sections) != 1 {
+			return shared.RewriteError(ctx, NewInvalidCursorErr("did not find expected resume relationship"))
+		}
```
vroldanbet, line 168: is this so that you cannot use a LR cursor in this API? Shouldn't this be centralized into a validation method?

internal/services/v1/relationships.go
```diff
@@ -148,12 +149,47 @@ func (ps *permissionServer) ReadRelationships(req *v1.ReadRelationshipsRequest,
...
+
+		parsed := tuple.Parse(decodedCursor.Sections[0])
+		if parsed == nil {
+			return shared.RewriteError(ctx, NewInvalidCursorErr("could not parse resume relationship"))
+		}
```
vroldanbet, line 173: maybe turn this into a helper like "ExactlyOneSection" which returns an error if it does not

vroldanbet (approved):

internal/services/v1/relationships.go
josephschorr: The hash check above makes sure you don't send a cursor from LR, but I wanted to be extra careful to not cause a bounds check error, so I added this

internal/services/v1/relationships.go
josephschorr: Same idea

PR #1342, opened 2023-05-22
increase timeout for health check tests

ecordell:
CRDB now waits until the pool has a minimum number of connections


josephschorr (approved):
LGTM

PR #1354, opened 2023-05-25
allow connpool flags to be written as conn-pool

ecordell:
This keeps the existing flags like `--datastore-connpool-read-max-open` but also allows them to be written as `--datastore-conn-pool-read-max-open`.

It's a minor difference, but it allows the field in the operator to be written more nicely as `datastoreConnPoolReadMaxOpen` instead of `datastoreConnpoolReadMaxOpen`. We could also make the change to support both only in the operator, but it seemed reasonable to do it here and help anyone out who might be misremembering the flag name.

### DIFF
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


pkg/cmd/datastore/datastore.go
```diff
@@ -167,8 +168,14 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	var legacyConnPool ConnPoolConfig
 	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-conn", DefaultReadConnPool(), &legacyConnPool)
 	deprecateUnifiedConnFlags(flagSet)
-	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-connpool-read", &legacyConnPool, &opts.ReadConnPool)
-	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-connpool-write", DefaultWriteConnPool(), &opts.WriteConnPool)
+	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-conn-pool-read", &legacyConnPool, &opts.ReadConnPool)
+	RegisterConnPoolFlagsWithPrefix(flagSet, "datastore-conn-pool-write", DefaultWriteConnPool(), &opts.WriteConnPool)
+	flagSet.SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
```
jzelinskie, line 173: Can you register multiple of these or should this be centralized in a place where we handle all legacy flags?

ecordell: it's once per flagset - I updated here to wrap any existing normalize func instead of overwriting it

PR #1356, opened 2023-05-26
introduces a flag to define the maximum relationship context size

vroldanbet:
Fixes https://github.com/authzed/spicedb/issues/1355

depends on https://github.com/authzed/api/pull/66
depends on https://github.com/authzed/authzed-go/pull/118

A new max-relationship-context-size flag is introduced with a default of 25KB. The gRPC RPC controller will fail to perform a WriteRelationship request if any of the relationships contained in the payload exceed the configured maximum size.

The API will return an InvalidArgument error with metadata:
- "relationship" that failed the condition
- "max_allowed_size" returns the configured limit by the server
- "relationship_size" returns the actual size of the relationship that failed the condition

### DIFF
--- a/pkg/cmd/serve.go
+++ b/pkg/cmd/serve.go
@@ -118,6 +118,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	cmd.Flags().Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	cmd.Flags().Uint16Var(&config.MaximumPreconditionCount, "update-relationships-max-preconditions-per-call", 1000, "maximum number of preconditions allowed for WriteRelationships and DeleteRelationships calls")
 	cmd.Flags().IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes. A value of zero or less means no limit")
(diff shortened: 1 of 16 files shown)


pkg/cmd/serve.go
```diff
@@ -118,6 +118,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	cmd.Flags().Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	cmd.Flags().Uint16Var(&config.MaximumPreconditionCount, "update-relationships-max-preconditions-per-call", 1000, "maximum number of preconditions allowed for WriteRelationships and DeleteRelationships calls")
 	cmd.Flags().IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes. A value of zero or less means no limit")
+	cmd.Flags().IntVar(&config.MaxRelationshipContextSize, "max-relationship-context-size", 25000, "maximum allowed size of the context to be stored in a relationship")
```
ecordell, line 121: isn't this defined in the database schema? what happens if you set it to a larger value than the db supports?

vroldanbet: @ecordell it's not defined in the database. Each database has a different maximum for the `JSON` datatype. I presume the database will fail the transaction if the size exceeds the maximum of the datatype.

ecordell (dismissed):

PR #1364, opened 2023-05-31
add support for termination log and custom exit codes

vroldanbet:
[termination log](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message) is a concept in Kubernetes that
allows applications providing information
upon termination that will be surfaced as part
of the status of a Pod. This is useful for troubleshooting
purposes but also can be used by the control plane.

The `termination.Error` also includes an `ExitCode` which will emit on process termination.

### DIFF
--- a/pkg/cmd/termination/termination.go
+++ b/pkg/cmd/termination/termination.go
@@ -7,9 +7,9 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"time"
 
 	log "github.com/authzed/spicedb/internal/logging"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 
 	"github.com/jzelinskie/cobrautil/v2"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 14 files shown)


pkg/cmd/termination/termination_test.go
```diff
@@ -0,0 +1,58 @@
...
+	var termErr Error
+	require.ErrorAs(t, err, &termErr)
+
+	jsonBytes, err := os.ReadFile(f.Name())
+	require.NoError(t, err)
+	var readErr Error
+	err = json.Unmarshal(jsonBytes, &readErr)
+	require.NoError(t, err)
+
+	require.Equal(t, publishedError.ErrorString, readErr.ErrorString)
```
ecordell, line 34: nit: I lean towards not using computed values in tests when possible. if there's a bug in `.Error()` that just returns nil for example, then these assertions will still all pass.

pkg/cmd/termination/termination.go
```diff
@@ -0,0 +1,120 @@
...
+
+// PublishError returns a new wrapping cobra run function that executes the provided argument runFunc, and
+// writes to disk an error returned by the latter if it is of type termination.Error.
+func PublishError(runFunc cobrautil.CobraRunFunc) cobrautil.CobraRunFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		runFuncErr := runFunc(cmd, args)
+		var termErr Error
+		if runFuncErr != nil && errors.As(runFuncErr, &termErr) {
+			terminationLogPath := cobrautil.MustGetString(cmd, terminationLogFlagName)
+			bytes, err := json.Marshal(termErr)
```
ecordell, line 94: there is a message length [limit of 4096](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#:~:text=Customizing%20the%20termination%20message&text=The%20termination%20message%20is%20intended,divided%20equally%20among%20each%20container.) here - maybe this should shed metadata if it's too long?

vroldanbet: @ecordell kube would truncate beyond 4096, which would be effectively the same. Maybe we should at least warn that the error is likely to be truncated?

pkg/cmd/termination/termination_test.go
vroldanbet: good point - what alternative do you suggest? I could add assertions to check those values are not nil

pkg/cmd/termination/termination.go
ecordell: a truncated json can be harder to parse (i.e. if we're going to display in the ui). I thought it would be nicer if we truncate that data before serializing so we still get valid json and the minimum data (like the error type / component / etc)?

vroldanbet: fair point, I can look into it 👍🏻

ecordell (approved):
LGTM

PR #1365, opened 2023-05-31
Parallelize the entrypoint computation in ReachableResources

josephschorr:
This brings some parallelism back that was lost during the switch to support cursors

We run each entrypoint in parallel, with the later branches being streamed to in-memory collections that are published in the correct order

### DIFF
--- a/internal/dispatch/stream.go
+++ b/internal/dispatch/stream.go
@@ -3,6 +3,7 @@ package dispatch
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 
 	grpc "google.golang.org/grpc"
 )
@@ -146,8 +147,41 @@ func (s *HandlingDispatchStream[T]) Context() context.Context {
 	return s.ctx
 }
 
+// CountingDispatchStream is a dispatch stream that counts the number of items published.
+// It uses an internal atomic int to ensure it is thread safe.
+type CountingDispatchStream[T any] struct {
+	Stream Stream[T]
+	count  *atomic.Uint64
+}
+
+func NewCountingDispatchStream[T any](wrapped Stream[T]) *CountingDispatchStream[T] {
+	return &CountingDispatchStream[T]{
+		Stream: wrapped,
+		count:  &atomic.Uint64{},
+	}
+}
+
+func (s *CountingDispatchStream[T]) PublishedCount() uint64 {
+	return s.count.Load()
+}
+
+func (s *CountingDispatchStream[T]) Publish(result T) error {
+	err := s.Stream.Publish(result)
+	if err != nil {
+		return err
+	}
+
+	s.count.Add(1)
+	return nil
+}
+
+func (s *CountingDispatchStream[T]) Context() context.Context {
+	return s.Stream.Context()
+}
+
 // Ensure the streams implement the interface.
 var (
 	_ Stream[any] = &CollectingDispatchStream[any]{}
 	_ Stream[any] = &WrappedDispatchStream[any]{}
+	_ Stream[any] = &CountingDispatchStream[any]{}
 )
--- a/internal/graph/cursors.go
+++ b/internal/graph/cursors.go
@@ -1,10 +1,14 @@
 package graph
 
 import (
+	"context"
+	"errors"
 	"strconv"
+	"sync"
 
 	"golang.org/x/exp/slices"
 
+	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/datastore/options"
 	v1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
@@ -54,6 +58,16 @@ func (ci cursorInformation) responsePartialCursor() *v1.Cursor {
 		AtRevision: ci.revision.String(),
 		Sections:   ci.outgoingCursorSections,
 	}
+}
+
+func (ci cursorInformation) withClonedLimits(ctx context.Context) (cursorInformation, context.Context) {
+	cloned, ctx := ci.limits.clone(ctx)
+	return cursorInformation{
+		currentCursor:          ci.currentCursor,
+		outgoingCursorSections: ci.outgoingCursorSections,
+		limits:                 cloned,
+		revision:               ci.revision,
+	}, ctx
 }
 
 // hasHeadSection returns true if the current cursor has the given name as the prefix of the cursor.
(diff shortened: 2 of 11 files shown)


internal/dispatch/stream.go
```diff
@@ -146,8 +146,47 @@ func (s *HandlingDispatchStream[T]) Context() context.Context {
 	return s.ctx
 }
 
+// CountingDispatchStream is a dispatch stream that counts the number of items published.
+// It uses an internal mutex to ensure it is thread safe.
+type CountingDispatchStream[T any] struct {
+	Stream Stream[T]
+	count  uint64
```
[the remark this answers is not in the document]
josephschorr: Good idea

internal/graph/preloadedtaskrunner.go
```diff
@@ -84,8 +99,9 @@ func (tr *preloadedTaskRunner) runner() {
 			// Run the task. If an error occurs, store it and cancel any further tasks.
 			err := task(tr.ctx)
 			if err != nil {
-				tr.cancel()
+				tr.storeErrorAndCancel(err)
 			}
+			tr.wg.Done()
```
ecordell, line 104: can't `startAndWait` deadlock if the task errors and we don't call done here?

internal/graph/preloadedtaskrunner.go
```diff
@@ -102,3 +118,31 @@ func (tr *preloadedTaskRunner) selectTask() TaskFunc {
...
+	}
+
+	for {
+		if len(tr.tasks) == 0 {
+			break
+		}
+
+		tr.tasks = tr.tasks[1:]
+		tr.wg.Done()
+	}
```
ecordell, line 147: I didn't think reslicing like this released the underlying array values until you re-assign over the same index?

maybe something like this?

```suggestion
    for i := 0; i < len(tr.tasks); i++ {
        tr.wg.Done()
    }
    tr.tasks = make([]TaskFunc, 0, len(tr.tasks))
```

internal/graph/cursors_test.go
```diff
@@ -212,3 +213,99 @@ func TestCombineCursorsWithNil(t *testing.T) {
...
+	items := []int{10, 20, 30, 40, 50}
+	parentStream := dispatch.NewCollectingDispatchStream[int](context.Background())
+	err = withParallelizedStreamingIterableInCursor[int, int](
+		context.Background(),
+		ci,
+		"iter",
+		items,
+		parentStream,
+		2,
+		func(ctx context.Context, cc cursorInformation, item int, stream dispatch.Stream[int]) error {
```
ecordell, line 236: nothing in the tests is currently asserting that there actually is concurrency compared to the old implementation

what about recording start/stop times for each runner and asserting overlap? or have the runners increment a shared counter on entry, decrement on exit, and assert the high watermark of 2?

internal/graph/cursors.go
```diff
@@ -310,3 +324,197 @@ func combineCursors(cursor *v1.Cursor, toAdd *v1.Cursor) (*v1.Cursor, error) {
...
+			if err != nil {
+				return err
+			}
+
+			return stream.completedTaskIndex(taskIndex)
+		})
+	}
+
+	// NOTE: since branches can be canceled if they have reached limits, the context Canceled error is ignored here.
+	if err := tr.startAndWait(); !errors.Is(err, context.Canceled) {
```
ecordell, line 399: but they'll also return canceled if the parent context is canceled right? and in that case we do want to return

internal/graph/cursors.go
```diff
@@ -310,3 +324,197 @@ func combineCursors(cursor *v1.Cursor, toAdd *v1.Cursor) (*v1.Cursor, error) {
...
+			// Remove the already emitted data from the overall limits.
+			done, err := ls.ci.limits.markAlreadyPublished(uint32(ls.countingStream.PublishedCount()))
+			if err != nil {
+				return err
+			}
+
+			defer done()
+		} else {
+			// Publish the results from the collected results of the child to the parent stream, in the correct order.
+			childStream := ls.childStreams[ls.toEmitIndex]
```
ecordell, line 501: I don't think this needs to be solved now, but it would be neat if we could do some sort of hybrid between the dispatch stream and the collecting stream.

If the first branch is totally done and the second branch is "halfway" done, we could just start streaming results out of the second branch.

i.e. treat the collecting stream like a buffer and not an aggregator

internal/graph/lookupresources.go
```diff
@@ -57,7 +57,7 @@ func (cl *CursoredLookupResources) LookupResources(
 
 		// Create a new handling stream that consumes the reachable resources results and publishes them
 		// as found resources if they are properly checked.
-		rrStream := dispatch.NewHandlingDispatchStream(ctx, func(result *v1.DispatchReachableResourcesResponse) error {
+		rrStream := dispatch.NewHandlingDispatchStream(withCancel, func(result *v1.DispatchReachableResourcesResponse) error {
```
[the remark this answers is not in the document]
josephschorr: a good idea; I'll add it to the linter list of ideas

internal/graph/limits.go
```diff
@@ -23,6 +25,16 @@ func newLimitTracker(ctx context.Context, optionalLimit uint32) (*limitTracker,
 	}, withCancel
 }
 
+// clone creates a copy of the limitTracker, inheriting the current limit.
+func (lt *limitTracker) clone(ctx context.Context) (*limitTracker, context.Context) {
+	withCancel, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
josephschorr: yep; it will cancel the parent `ctx` as well

internal/graph/preloadedtaskrunner.go
josephschorr: the conditional falls through on error

internal/graph/cursors_test.go
josephschorr: I don't think we need to assert parallelism; so long as it is correct, the parallelism is just a bonus for performance

internal/graph/cursors.go
josephschorr: that'll be handled at the parent call site though

internal/graph/cursors.go
josephschorr: Yeah, that was my original idea but it adds even more complexity to the system, so I figured we'd just do it this way now

internal/graph/cursors_test.go
ecordell: sure, but behavior-wise there's really nothing to distinguish the changes in this PR from the existing implementation. maybe a benchmark?

internal/graph/cursors.go
```diff
@@ -310,3 +324,197 @@ func combineCursors(cursor *v1.Cursor, toAdd *v1.Cursor) (*v1.Cursor, error) {
...
+}
+
+// completedTaskIndex indicates the the task at the specific index has completed successfully and that its collected
+// results should be published to the parent stream, so long as all previous tasks have been completed and published as well.
+func (ls *parallelLimitedIndexedStream[Q]) completedTaskIndex(index int) error {
+	ls.lock.Lock()
+	defer ls.lock.Unlock()
+
+	// Mark the task as completed.
+	ls.completedIndexes[index] = true
```
vroldanbet, line 477: Let's say stream 3 completes before stream 2, how's stream 3 going to emit results if it's waiting for stream 2, but stream 2 cannot be marked as completed because of the mutex being acquired by 3?

josephschorr: *stream* 3 is marked as completed; but it is not marked as published; `completed` is for the task, but publishing happens after it

internal/graph/preloadedtaskrunner.go
ecordell: but I don't think it will release the functions that are pointed to by the underlying array?

seems like a small change in exchange for free more garbage faster?

internal/graph/cursors_test.go
ecordell: if you've tested manually separate that seems fine. but it seemed easy to just add an atomic Uint and track a high water mark of concurrent processes (that would confirm concurrency and the the concurrency limit at the same time)

josephschorr: I added a test that ensures parallelism

internal/graph/cursors_test.go
```diff
@@ -212,3 +214,142 @@ func TestCombineCursorsWithNil(t *testing.T) {
...
+		5,
+		func(ctx context.Context, cc cursorInformation, item int, stream dispatch.Stream[int]) error {
+			lock.Lock()
+			encountered = append(encountered, item)
+			lock.Unlock()
+
+			return stream.Publish(item * 10)
+		})
+
+	require.Equal(t, len(expected), len(encountered))
```
ecordell, line 350: I think I'm missing how this tests concurrency?

ecordell (approved):
LGTM

internal/graph/cursors_test.go
josephschorr: `require.NotEqual(t, encountered, expected)` does so

PR #1367, opened 2023-06-01
Use correct error type for duplicate relationships

### LINKED ISSUE #1366: Duplicate updates return the wrong ErrorReason reason
When you submit a gRPC request in authzed-java with duplicate `RelationshipUpdate` entries, it returns an error as expected, but the `ErrorReason` listed in the response is `ERROR_REASON_INVALID_SUBJECT_TYPE`, which doesn't make much sense to me.

It seems like it ought to be `ERROR_REASON_UPDATES_ON_SAME_RELATIONSHIP`.

reify-tanner-stirrat:
Fixes #1366 
Duplicates #1368 

## Description
We noticed this when looking at error output for a duplicate relationship issue - it wasn't returning the correct error reason out of the enum.

## Changes
* Use the correct error type
## Testing
I suppose it would be good to write a test for this, but I'm not familiar with Go. If you'd like to see some tests I'll dig into it.

### DIFF
--- a/internal/services/v1/errors.go
+++ b/internal/services/v1/errors.go
@@ -168,7 +168,7 @@ func (err ErrDuplicateRelationshipError) GRPCStatus() *status.Status {
 		err,
 		codes.InvalidArgument,
 		spiceerrors.ForReason(
-			v1.ErrorReason_ERROR_REASON_INVALID_SUBJECT_TYPE,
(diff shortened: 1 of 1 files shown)


vroldanbet (approved):
LGTM! Thanks for reporting and fixing! 🎉

PR #1380, opened 2023-06-03
Re-parallelize the checking of resources in LookupResources

josephschorr:
This was removed during the cursoring+limits work due to the complexity involved, but is brought back via the use of a specialized stream implementation that queues all results to be processed and then published, maintaining the proper publishing order while running the checks in parallel where applicable

Basic benchmark when checks are necessary for the results:

```
v1.21.0:
BenchmarkServices/lookup_with_intersection/postgres/v1-10      	      10	  29911154 ns/op	10162012 B/op	  101606 allocs/op

v1.22.0-rc4:
BenchmarkServices/lookup_with_intersection/postgres/v1-10      	      10	6125200771 ns/op	241257368 B/op	 3540723 allocs/op

this PR:
BenchmarkServices/lookup_with_intersection/postgres/v1-10      	      10	  43249717 ns/op	20608956 B/op	  204650 allocs/op
```

### DIFF
--- a/internal/graph/checkingresourcestream.go
+++ b/internal/graph/checkingresourcestream.go
@@ -1,0 +1,532 @@
+package graph
+
+import (
+	"context"
+	"sort"
+	"sync"
+
+	"golang.org/x/exp/maps"
+
+	"github.com/authzed/spicedb/internal/dispatch"
+	"github.com/authzed/spicedb/internal/graph/computed"
+	"github.com/authzed/spicedb/pkg/datastore"
+	v1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+	"github.com/authzed/spicedb/pkg/util"
+)
+
+// possibleResource is a resource that was returned by reachable resources and, after processing,
+// may be returned by the lookup resources stream.
+type possibleResource struct {
+	// reachableResult is the result for this resource from the reachable resources stream.
+	reachableResult *v1.DispatchReachableResourcesResponse
+
+	// lookupResult is the result to be published by LookupResources, if the resource is actually
+	// permissioned. Will be nil before processing and nil after processing IF the resource needed
+	// to be checked and the check showed the resource was inaccessible.
+	lookupResult *v1.DispatchLookupResourcesResponse
+
+	// orderingIndex is the index of the resource result as returned by reachable resources. Used to
+	// maintain strict publishing order of results.
+	orderingIndex uint64
+}
+
+// resourceQueue is a queue for managing of possibleResources through the various states of the stream (queueing, processing and publishing).
+type resourceQueue struct {
+	ctx  context.Context
+	lock sync.Mutex
+
+	// toProcess are those resources (keyed by orderingIndex) that have not yet been processed (checked).
+	toProcess map[uint64]possibleResource
+
+	// toPublish are those resources (keyed by orderingIndex) that have been processed and
+	// are ready for publishing. Note that resources whose Check calls showed NO_PERMISSION
+	// will *also* be in this map, just with lookupResult set to nil. This is done to ensure
+	// strict ordering of publishing.
+	toPublish map[uint64]possibleResource
+
+	// beingProcessed are those resources (keyed by orderingIndex) that are currently being processed.
+	beingProcessed map[uint64]possibleResource
+}
+
+// addPossibleResource queues a resource for processing.
+func (rq *resourceQueue) addPossibleResource(pr possibleResource) {
+	rq.lock.Lock()
+	defer rq.lock.Unlock()
+
+	rq.toProcess[pr.orderingIndex] = pr
+}
+
+// updateToBePublished marks a resource as ready for publishing.
+func (rq *resourceQueue) updateToBePublished(pr possibleResource) {
+	rq.lock.Lock()
+	defer rq.lock.Unlock()
+
+	rq.toPublish[pr.orderingIndex] = pr
+	delete(rq.beingProcessed, pr.orderingIndex)
+}
+
+// markResourceCompleted marks that all work has been completed on the given resources.
+func (rq *resourceQueue) markResourceCompleted(pr possibleResource) {
+	rq.lock.Lock()
+	defer rq.lock.Unlock()
+
+	delete(rq.toPublish, pr.orderingIndex)
+}
+
+// selectResourcesToProcess marks all toProcess resources as being processing and returns those resources
+// for processing.
+//
+// If alwaysReturn is false, then resources will only be returned if they meet the chunk size, to ensure
+// checks occur in larger batches.
+func (rq *resourceQueue) selectResourcesToProcess(alwaysReturn bool) []possibleResource {
+	rq.lock.Lock()
+	defer rq.lock.Unlock()
+
+	toProcess := maps.Values(rq.toProcess)
+	if !alwaysReturn && len(toProcess) < int(datastore.FilterMaximumIDCount) {
+		return nil
+	}
+
+	for _, pr := range toProcess {
+		rq.beingProcessed[pr.orderingIndex] = pr
+	}
+
+	rq.toProcess = map[uint64]possibleResource{}
+	return toProcess
+}
+
+// resourcesToPossiblyPublish returns all resources in the toPublish state. This does *not* mark the resources
+// as published.
+func (rq *resourceQueue) resourcesToPossiblyPublish() []possibleResource {
+	rq.lock.Lock()
+	defer rq.lock.Unlock()
+
+	return maps.Values(rq.toPublish)
+}
+
+// checkingResourceStream is a Stream[*v1.DispatchLookupResourcesResponse] that consumes reachable resource
+// responses which are published to it, checks the resource (if necessary), and then publishes the resource
+// if reachable. This stream performs Checks for the possibly-inaccessible resources in a parallel fashion
+// but maintains the proper publishing order to the parent stream.
+//
+// Resources in the stream are processed as follows:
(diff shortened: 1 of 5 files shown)


internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+	defer rq.lock.Unlock()
+
+	for _, pr := range prs {
+		rq.toPublish[pr.orderingIndex] = pr
+		delete(rq.beingProcessed, pr.orderingIndex)
+	}
+}
+
+// markResourceCompleted marks that all work has been completed on the given resources.
+func (rq *resourceQueue) markResourceCompleted(prs ...possibleResource) {
```
vroldanbet, line 74: I don't see any call to this method with more than 1 resource. I'm not sure if the compiler is clever enough to avoid the allocation, but if there is no use we should probably remove it anyway, and turn it into a variadic when the need arises.

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+	rq.lock.Lock()
+	defer rq.lock.Unlock()
+
+	for _, pr := range prs {
+		rq.toProcess[pr.orderingIndex] = pr
+	}
+}
+
+// updateToBePublished marks resources as ready for publishing.
+func (rq *resourceQueue) updateToBePublished(prs ...possibleResource) {
```
vroldanbet, line 63: same as the comment below 👇🏻 remove variadic signature since there is no use with more than 1 element

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+	// will *also* be in this map, just with lookupResult set to nil. This is done to ensure
+	// strict ordering of publishing.
+	toPublish map[uint64]possibleResource
+
+	// beingProcessed are those resources (keyed by orderingIndex) that are currently being processed.
+	beingProcessed map[uint64]possibleResource
+}
+
+// addPossibleResources queues resources for processing.
+func (rq *resourceQueue) addPossibleResources(prs ...possibleResource) {
```
vroldanbet, line 53: same as the comment below 👇🏻 remove variadic signature since there is no use with more than 1 element

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+	// sem is a chan of length `concurrencyLimit` used to ensure the task runner does
+	// not exceed the concurrencyLimit with spawned goroutines.
+	sem chan token
+
+	// rq is the resourceQueue for managing the state of all resources returned by the reachable resources call.
+	rq *resourceQueue
+
+	// availableForProcessing is a channel which indicates to the processing worker(s) that work is available
+	// for processing.
+	availableForProcessing chan struct{}
```
vroldanbet, line 155: maybe `reachableResourceAvailable`? Or maybe turn the one below 👇🏻 into `processingCompleted`. Suggesting because by reading the code I was confused by which phase was completed through the `select` loops

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+
+	for _, rai := range toCheck.Values() {
+		checkResult := results[rai.reachableResult.Resource.ResourceId]
+
+		var permissionship v1.ResolvedResource_Permissionship
+		var missingFields []string
+
+		switch {
+		case checkResult == nil || checkResult.Membership == v1.ResourceCheckResult_NOT_MEMBER:
+			permissionship = v1.ResolvedResource_UNKNOWN
```
vroldanbet, line 440: I'd suggest adding a comment to clarify that `UNKNOWN` is never to be expected, but we are using it here to maintain the order right before publishing

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+				Metadata:            metadata,
+				AfterResponseCursor: rai.reachableResult.AfterResponseCursor,
+			}
+		}
+
+		rai.lookupResult = lookupResult
+		crs.rq.updateToBePublished(rai)
+	}
+
+	crs.availableForPublishing <- true
```
vroldanbet, line 472: wouldn't we want to signal element is available for publishing as soon as it is published, or you think that'd be much signaling overhead, and have `publishResourcesIfPossible` invoked on batches rather than on every element that becomes ready?

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+	}
+
+	for _, rai := range toCheck.Values() {
+		checkResult := results[rai.reachableResult.Resource.ResourceId]
+
+		var permissionship v1.ResolvedResource_Permissionship
+		var missingFields []string
+
+		switch {
+		case checkResult == nil || checkResult.Membership == v1.ResourceCheckResult_NOT_MEMBER:
```
vroldanbet, line 439: this case is not being tested by the newly added test in `internal/dispatch/graph/lookupresources_test.go`

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+			permissionship = v1.ResolvedResource_UNKNOWN
+
+		case checkResult != nil && checkResult.Membership == v1.ResourceCheckResult_MEMBER:
+			permissionship = v1.ResolvedResource_HAS_PERMISSION
+
+		case checkResult != nil && checkResult.Membership == v1.ResourceCheckResult_CAVEATED_MEMBER:
+			permissionship = v1.ResolvedResource_CONDITIONALLY_HAS_PERMISSION
+			missingFields = checkResult.MissingExprFields
```
vroldanbet, line 447: this case is not being tested by the newly added test in `internal/dispatch/graph/lookupresources_test.go`

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+				}
+			}
+
+			crs.rq.markResourceCompleted(current)
+		}
+	}
+}
+
+// setError sets an error that occurred.
+func (crs *checkingResourceStream) setError(err error) {
```
vroldanbet, line 338: we may want to test as well that all winds down properly on error

internal/graph/computed/computecheck.go
```diff
@@ -89,30 +90,42 @@ func computeCheck(ctx context.Context,
...
+			ResultsSetting:   setting,
+			Subject:          params.Subject,
+			Metadata: &v1.ResolverMeta{
+				AtRevision:     params.AtRevision.String(),
+				DepthRemaining: params.MaximumDepth,
+			},
+			Debug: debugging,
+		})
+		for _, resourceID := range resourceIDsToCheck {
+			resultMetadatas[resourceID] = checkResult.Metadata
```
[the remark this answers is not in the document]
josephschorr: Same reason as above: metadata must always be returned, even if there is an error

internal/graph/computed/computecheck.go
```diff
@@ -89,30 +90,42 @@ func computeCheck(ctx context.Context,
...
-		return nil, checkResult.Metadata, err
-	}
-
+	// Ensure that the number of resources IDs given to each dispatch call is not in excess of the maximum.
 	results := make(map[string]*v1.ResourceCheckResult, len(resourceIDs))
-	for _, resourceID := range resourceIDs {
-		computed, err := computeCaveatedCheckResult(ctx, params, resourceID, checkResult)
+	resultMetadatas := make(map[string]*v1.ResponseMeta, len(resourceIDs))
+
+	// TODO(jschorr): Should we make this run in parallel via the preloadedTaskRunner?
```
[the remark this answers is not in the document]
josephschorr: Limit is bounded by the LR itself, which is no more than 1000 items per iteration

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+
+// possibleResource is a resource that was returned by reachable resources and, after processing,
+// may be returned by the lookup resources stream.
+type possibleResource struct {
+	// reachableResult is the result for this resource from the reachable resources stream.
+	reachableResult *v1.DispatchReachableResourcesResponse
+
+	// lookupResult is the result to be published by LookupResources, if the resource is actually
+	// permissioned. Will be nil before processing and nil after processing IF the resource needed
+	// to be checked and the check failed.
```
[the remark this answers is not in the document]
josephschorr: Sure, I'll clarify

internal/graph/checkingresourcestream.go
```diff
@@ -0,0 +1,537 @@
...
+
+			toPublish = toPublish[1:]
+			crs.orderingIndexToBePublished++
+
+			// NOTE: lookupResult will be `nil` if the Check for the resource found that the resource is
+			// not actually accessible. The entry is kept in `toPublish` to ensure proper ordering is maintained
+			// on the parent stream.
+			if current.lookupResult != nil {
+				ok, done := crs.limits.prepareForPublishing()
+				defer done()
```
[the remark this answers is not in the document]
josephschorr: > but it would have incremented orderingIndexToBePublished and not published the resource.

That's fine: once the limit is reached, no additional publishing should occur.

internal/graph/checkingresourcestream.go
josephschorr: I figured the batch would make things slightly more efficient

internal/graph/checkingresourcestream.go
josephschorr: Added

internal/graph/checkingresourcestream.go
josephschorr: Added

josephschorr:
Updated

vroldanbet (approved):

PR #1383, opened 2023-06-07
correctly chunks large bootstrap files

vroldanbet:
the method was chunking on tuples but then
sending the full update list to the database,
which failed in CRDB due to hitting the maximum
number of statement placeholders with an error:

"lexical error: placeholder index must be between 1 and 65536"

### DIFF
--- a/pkg/validationfile/loader_test.go
+++ b/pkg/validationfile/loader_test.go
@@ -5,10 +5,13 @@ import (
 	"sort"
 	"testing"
 
-	"github.com/stretchr/testify/require"
-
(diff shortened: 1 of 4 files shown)


pkg/validationfile/loader_test.go
```diff
@@ -5,6 +5,10 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/authzed/spicedb/internal/datastore/proxy/proxy_test"
+	"github.com/authzed/spicedb/pkg/datastore"
```
josephschorr, line 9: fix ordering

josephschorr (approved):
LGTM

PR #1385, opened 2023-06-07
Fix a deadlock in preloaded task runner

josephschorr:
We need to make sure to completely cancel the runner if context was canceled


vroldanbet (approved):
LGTM

Issue #1357, opened 2023-05-28
ResourceExhausted Error Due to Non-Limited newRevisionsQuery and Uncontrolled Buffer Size


staskosatkin:
**Issue Description**

While using the Watch API, I've encountered a `ResourceExhausted` error. This problem can lead to the loss of events and disconnections. The issue seems to arise because the `newRevisionsQuery` lacks a limit, which may result in it returning a vast number of snapshots. Furthermore, the buffer size isn't controlled, and it could easily be smaller than the size of `changeToWrite`. This situation creates a risk to the consistency of changes received from the Watch API.

**Relevant Code Snippets**

https://github.com/authzed/spicedb/blob/6d2e468a14e237a3926853193fda04c65d17fe77/internal/datastore/postgres/watch.go#L31-L35
No limit clause present.

**Potential Overflow:**
https://github.com/authzed/spicedb/blob/6d2e468a14e237a3926853193fda04c65d17fe77/internal/datastore/postgres/watch.go#L93-L103
Here, size of `updates` (buffer) could be easily less than `changeToWrite` size.

**Proposed Solution**
I suggest to add a limit to newRevisionsQuery similar to the buffer size. Secondly, controlling the size of the buffer will allow us to continue watching from any point in time until the revisions weren't deleted by the garbage collector (by default it's 24h and it's really enough). This modification provides a great opportunity to maintain the consistency of the event stream, even during downtime of the Watch API's client.  Here's an example fix:

```
for _, changeToWrite := range changesToWrite {
	changeToWrite := changeToWrite

	for int(pgd.watchBufferLength) <= len(updates) {
		sleep := time.NewTimer(watchOverflowSleep)
		select {
		case <-sleep.C:
			continue
		case <-ctx.Done():
			errs <- datastore.NewWatchCanceledErr()
			return
		}
	}

	select {
	case updates <- &changeToWrite:
		// Nothing to do here, we've already written to the channel.
	default:
		errs <- datastore.NewWatchDisconnectedErr()
		return
	}
}
```
**Additional Context**

The proposed fix should improve the reliability of the watch API.

Thank you for your attention to this matter.

Best regards,
staskosatkin

josephschorr:
@staskosatkin seems reasonable. Would you be willing to open a PR?

staskosatkin:
@josephschorr I'll prepare PR, thanks

staskosatkin:
PR was opened https://github.com/authzed/spicedb/pull/1388

[...]

PR #1388, opened 2023-06-08
Extending Watch Api consistency. Postgres Buffer size fix. 

staskosatkin:
PR linked to issue https://github.com/authzed/spicedb/issues/1357

There is added buffer length checking before sending to the chan. So Watch API could continue streaming data. The one limit is GC period duration.

Please let me know if you have any concerns about this PR

### DIFF
--- a/internal/datastore/postgres/watch.go
+++ b/internal/datastore/postgres/watch.go
@@ -16,7 +16,8 @@ import (
 )
 
 const (
-	watchSleep = 100 * time.Millisecond
+	watchSleep       = 100 * time.Millisecond
+	watchSelectLimit = 100
 )
 
 type revisionWithXid struct {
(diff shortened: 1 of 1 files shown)


josephschorr:
@staskosatkin Can you please sign the CLI?

staskosatkin:
**_I have read the CLA Document and I hereby sign the CLA_**

staskosatkin:
Hi @josephschorr ! Could you please take a look?

josephschorr:
@staskosatkin Apologies, I was offline for a few days. Will take a look tomorrow

internal/datastore/crdb/watch.go
```diff
@@ -119,6 +120,17 @@ func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Rev
 				})
 
 				for _, change := range toEmit {
+					for int(cds.watchBufferLength) <= len(updates) {
```
josephschorr, line 123: So I'm not convinced this is necessary; if we remove the `default` clause from the select below, then the `updates <- change` should hang if the update channel is filled.

Can you try that approach?

josephschorr: @staskosatkin

josephschorr: Correct; the channel insert will hang, which is what we expect, correct?

staskosatkin: I will check this approach and update this PR after that. Thank you.

josephschorr: @staskosatkin Checking in on this

josephschorr: @staskosatkin Checking in again

josephschorr:
Closing due to lack of movement; Please feel to reopen if necessary

PR #1389, opened 2023-06-08
Properly handle context cancelation in LookupResources processing

josephschorr:


internal/services/shared/errors.go
```diff
@@ -129,6 +129,13 @@ func RewriteError(ctx context.Context, err error) error {
 	case errors.Is(err, context.DeadlineExceeded):
 		return status.Errorf(codes.DeadlineExceeded, "%s", err)
 	case errors.Is(err, context.Canceled):
+		err := context.Cause(ctx)
```
vroldanbet, line 132: TIL there is actually a cause to cancellation

internal/graph/reachableresources.go
```diff
@@ -270,6 +270,12 @@ func (crr *CursoredReachableResources) chunkedRedispatch(
 				return nil, nil
 			}
 
+			// If the number of results returned was less than the limit specified, then this is
+			// the final iteration and no cursor should be returned for the next iteration.
+			if rsm.len() < int(queryLimit) {
+				lastTpl = nil
```
vroldanbet, line 276: was the cursor being returned via API? can we add test for this?

internal/graph/lookupresources.go
```diff
@@ -55,14 +55,14 @@ func (cl *CursoredLookupResources) LookupResources(
 	reachableContext, cancelReachable := context.WithCancel(newContextForReachable)
 	defer cancelReachable()
 
-	limits, lCtx := newLimitTracker(lookupContext, req.OptionalLimit)
+	limits, _ := newLimitTracker(lookupContext, req.OptionalLimit)
```
vroldanbet, line 58: 👍🏻  reaching the limit should cancel RR but not LR. Im surprised we didn't have a test that would causes this to trigger.

internal/services/shared/errors.go
josephschorr: As of Go 1.20

internal/graph/reachableresources.go
josephschorr: No; it just meant an extra iteration was occurring

josephschorr:
Updated

vroldanbet (approved):

PR #1391, opened 2023-06-09
decouples middleware modifications

vroldanbet:
until now middleware modifications composed both unary and streaming middlewares. This worked under the assumption that we wanted the same order in each type of API method.

SpiceDB relies on `grpc-ecosystem/go-grpc-middleware` to build [chains of interceptors](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/bfde8086e68d6a129363b063ffd6518ab568467d/doc.go), which is not natively
supported by the go gRPC SDK. The order of execution is, assuming a slice of interceptors, from left to right.

Because of the design of gRPC Streaming APIs, composed of stream creation and actual streaming of messages, while the chain of interceptors will follow the expected middleware chain order, the actual interception of messages (`RecvMsg` and `SendMsg`) will happen in the opposite order.

Having middlewares depend on each other is a not good practice, but there are circumstances
where at the very least you need a middleware to be the first or the last in the chain, think
middlewares for auditing purposes should happen earlier, otherwise, it wouldn't catch requests that
fail earlier in the chain.

This can be addressed by reversing the order of the middleware to obtain the expected order, but this hasn't been possible because the `MiddlewareModification` API had both unary and streaming coupled.

This commit separates modifications to unary and streaming middlewares, generalized using Go generics.

ℹ️ The generated options rely on a modification to `ecordell/optgen` in order to support generics. See https://github.com/ecordell/optgen/pull/9


internal/services/integrationtesting/cert_test.go
```diff
@@ -137,24 +137,40 @@ func TestCertRotation(t *testing.T) {
...
+						Middleware: servicespecific.UnaryServerInterceptor,
+					},
+				},
+			},
+		}),
+		server.SetStreamingMiddlewareModification([]server.MiddlewareModification[grpc.StreamServerInterceptor]{
+			{
+				Operation: server.OperationReplaceAllUnsafe,
+				Middlewares: []server.ReferenceableMiddleware[grpc.StreamServerInterceptor]{
+					{
```
josephschorr, line 163: I wonder if we want to verify that the same interceptors are set on both unary and streaming, to ensure that we don't miss one?

pkg/cmd/server/middleware.go
```diff
@@ -11,21 +11,21 @@ import (
 // MiddlewareChain describes an ordered sequence of middlewares that can be modified
 // with one or more MiddlewareModification. This struct is used to facilitate the
 // creation and modification of gRPC middleware chains
-type MiddlewareChain struct {
-	chain []ReferenceableMiddleware
+type MiddlewareChain[T grpc.UnaryServerInterceptor | grpc.StreamServerInterceptor] struct {
```
josephschorr, line 14: I think you can define an interface that implements this constraint and just reference it

josephschorr (commented):
Do we have a test for the ordering?

pkg/cmd/server/middleware.go
vroldanbet: I like this is explicit. In which ways do you think adding a synthetic interface will improve the code?

josephschorr: Just remove repeating it everywhere

internal/services/integrationtesting/cert_test.go
ecordell: If we went back to having unary/streaming coupled into one, then you could do the validation at the aggregation instead:

```go
Middleware {
 MiddelwareID
 UnaryInterceptor
 StreamingInterceptor
}

MiddlewareChain {
   []Middleware
   UnaryOrder []MiddlewareID
   StreamingOrder []MiddlewareID
}
```

A validate method or a constructor could assert that every non-nil interceptor is somewhere in the order list?

Not sure this adds that much value, though.

vroldanbet: I'm ve gone with an approach that asserts in tests that the order executed is what we expect, see https://github.com/authzed/spicedb/pull/1395

vroldanbet: oh I see what you say 👍🏻

pkg/cmd/server/middleware.go
vroldanbet: done in https://github.com/authzed/spicedb/pull/1391/commits/67c0fcb33a5165ad8b51b2c0400d63d7708d5db1

vroldanbet:
>Do we have a test for the ordering?

I did a whole thing asserting order in https://github.com/authzed/spicedb/pull/1395

tools/analyzers/go.work.sum
josephschorr, line 1: Should this file be in the PR?

vroldanbet: The file predated this PR, it just got updated, which makes me think it should have been updated in a previous PR but wasn't. We have some actions in CI to fail if the repository is dirty, but somehow ignored this file.

The file was introduced in https://github.com/authzed/spicedb/commit/43678a7704335aee2e5b99aab2d75fae0d42b56b#diff-583a688fd5bc869dc63c64d19cb1dab06bfbde70115e13ad54f7e280ca57794eR1-R2. My understanding is its needed because we are adding a `go.work` file which acts as the good old `replace` directive to have this module use what's checked-in git.

@ecordell can you confirm this? ☝🏻

ecordell: Yes it was intentional. Normally we wouldn't commit go.work like this, but it's just to make the tools find all of the right dependencies.

josephschorr (approved):
LGTM

PR #1393, opened 2023-06-12
default termination-log to disabled to make sure SpiceDB does not fail to start

vroldanbet:
While `/dev/termination-log` is the default path to
termination logs in kube environments, for non-kube
environments the process may not have permission
to write to this path and would fail to start.

This makes the termination log default to disabled
if the flag has a value of empty string.


josephschorr (approved):
LGTM

PR #1395, opened 2023-06-13
grpc middleware order assertion

vroldanbet:
⚠️ built on top of https://github.com/authzed/spicedb/pull/1391

The behaviour of gRPC streaming API middlewares can be surprising w.r.t to order of execution. This is an attempt to force the developer introducing gRPC middleware in SpiceDB to be explicit about the order.

When run in tests, the fluent API builder will wrap each middlewares into another interceptor that will run assertions on the order of execution, and panic if that expected order is not met.

### DIFF
--- a/pkg/cmd/server/defaults.go
+++ b/pkg/cmd/server/defaults.go
@@ -142,35 +142,26 @@ func DefaultUnaryMiddleware(logger zerolog.Logger, authFunc grpcauth.AuthFunc, e
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareRequestID).
 			WithInterceptor(requestid.UnaryServerInterceptor(requestid.GenerateIfMissing(true))).
-			EnsureAlreadyExecuted(none).
-			EnsureNotExecuted(DefaultMiddlewareLog).
 			Done(),
 
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareLog).
 			WithInterceptor(logmw.UnaryServerInterceptor(logmw.ExtractMetadataField("x-request-id", "requestID"))).
-			EnsureAlreadyExecuted(DefaultMiddlewareRequestID).
-			EnsureNotExecuted(DefaultMiddlewareGRPCLog).
 			Done(),
 
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareGRPCLog).
 			WithInterceptor(grpclog.UnaryServerInterceptor(grpczerolog.InterceptorLogger(logger), defaultGRPCLogOptions...)).
-			EnsureAlreadyExecuted(DefaultMiddlewareLog).
-			EnsureNotExecuted(DefaultMiddlewareOTelGRPC).
 			Done(),
 
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareOTelGRPC).
 			WithInterceptor(otelgrpc.UnaryServerInterceptor()).
-			EnsureAlreadyExecuted(DefaultMiddlewareGRPCLog).
-			EnsureNotExecuted(DefaultMiddlewareGRPCProm).
 			Done(),
 
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareGRPCProm).
 			WithInterceptor(grpcprom.UnaryServerInterceptor).
-			EnsureAlreadyExecuted(DefaultMiddlewareOTelGRPC).
 			EnsureNotExecuted(DefaultMiddlewareGRPCAuth).
 			Done(),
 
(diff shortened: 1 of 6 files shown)


pkg/cmd/server/defaults.go
```diff
@@ -139,109 +139,171 @@ const (
...
-		},
-		{
-			Name:       DefaultInternalMiddlewareServerSpecific,
-			Internal:   true,
-			Middleware: servicespecific.UnaryServerInterceptor,
-		},
+		NewUnaryMiddleware().
+			WithName(DefaultMiddlewareRequestID).
+			WithInterceptor(requestid.UnaryServerInterceptor(requestid.GenerateIfMissing(true))).
+			EnsureAlreadyExecuted(none).
```
josephschorr, line 145: why not just not have this call?

pkg/cmd/server/defaults.go
```diff
@@ -139,109 +139,171 @@ const (
...
+			EnsureNotExecuted(DefaultMiddlewareLog).
+			Done(),
+
+		NewUnaryMiddleware().
+			WithName(DefaultMiddlewareLog).
+			WithInterceptor(logmw.UnaryServerInterceptor(logmw.ExtractMetadataField("x-request-id", "requestID"))).
+			EnsureAlreadyExecuted(DefaultMiddlewareRequestID).
```
josephschorr, line 152: I think we should have the Ensure calls when its necessary; if a middleware doesn't care about ordering relative to other middlewares, no need for the check

pkg/cmd/server/middleware.go
```diff
@@ -172,9 +176,9 @@ func (mc *MiddlewareChain[T]) validate(mod MiddlewareModification[T]) error {
 	}
 
 	// prevent appending/prepending a duplicate middleware
-	for _, middleware := range mod.Middlewares {
-		if existingNames.Has(middleware.Name) && mod.DependencyMiddlewareName == middleware.Name && mod.Operation != OperationReplace {
-			return fmt.Errorf("modification will cause a duplicate in chain: %s", middleware.Name)
+	for _, mw := range mod.Middlewares {
+		if existingNames.Has(mw.Name) && mod.DependencyMiddlewareName == mw.Name && mod.Operation != OperationReplace {
+			return fmt.Errorf("modification will cause a duplicate in chain: %s", mw.Name)
```
josephschorr, line 181: make a spiceerrors.MustBugf

pkg/cmd/server/middleware.go
```diff
@@ -210,3 +214,225 @@ func (mc *MiddlewareChain[T]) modify(modifications ...MiddlewareModification[T])
 	}
 	return nil
 }
+
+var none = uuid.NewString()
+
+func inTest() bool {
+	return flag.Lookup("test.v") != nil
```
josephschorr, line 221: can we use the function we already have defined in the testutil package?

pkg/cmd/server/middleware.go
```diff
@@ -210,3 +214,225 @@ func (mc *MiddlewareChain[T]) modify(modifications ...MiddlewareModification[T])
...
+}
+
+func mustHaveExecuted(ctx context.Context, expectedExecuted string) error {
+	if expectedExecuted == "" {
+		return nil
+	}
+
+	val := ctx.Value(interceptorsExecuted)
+	if val == nil {
+		panic("interception order validation bookkeeping not present in context")
```
josephschorr, line 409: use spiceerrors.MustBugf

pkg/cmd/server/defaults.go
vroldanbet: because then it defaults to not asserting that part, and `none` explicitly asserts that zero middleware have been executed

pkg/cmd/server/middleware.go
vroldanbet: I couldn't find any. Could you point me to it?

pkg/cmd/server/middleware.go
vroldanbet: That does not work - this is only executed only in tests, and if I have a panic here, because it happened in a random goroutine in the gRPC server, I couldn't recover that panic in the test I wrote to ensure these new helpers all work.

pkg/cmd/server/middleware.go
vroldanbet: I can add it but it does not provide value because none of this is executed in runtime

pkg/cmd/server/defaults.go
vroldanbet: Do you know if any of these middlewares care about ordering?

pkg/cmd/server/middleware.go
josephschorr: Ok

pkg/cmd/server/middleware.go
josephschorr: https://github.com/authzed/spicedb/blob/5d33a67c26f89275f839ee7841557c3552c45cd9/pkg/spiceerrors/bug.go#L10

pkg/cmd/server/middleware.go
josephschorr: ok

pkg/cmd/server/middleware.go
vroldanbet: it's unexported. If you are cool with that, I can export it

PR #1399, opened 2023-06-16
Fix generated membership for the development package when using indir…

josephschorr:
…ect caveats

Also adds a consistency test case for this scenario


ecordell (approved):
Do we not have any direct integration tests for the generated expected relations? Otherwise looks good.

josephschorr:
> Do we not have any direct integration tests for the generated expected relations? Otherwise looks good.

We do, but constructing it off of a schema is non-trivial and this seemed easier; I added a test to ensure Expand was returning the correct information as well.

PR #1400, opened 2023-06-20
Change the max depth exceeded error to be well-typed

josephschorr:


internal/dispatch/dispatch.go
```diff
@@ -86,23 +82,23 @@ type LookupSubjects interface {
 	) error
 }
 
-// HasMetadata is an interface for requests containing resolver metadata.
-type HasMetadata interface {
+// DispatchableRequest is an interface for requests.
+type DispatchableRequest interface {
```
vroldanbet, line 86: the interface does not say anything about dispatch tho, more like `MetadataProvider`

internal/services/shared/errors.go
```diff
@@ -66,6 +69,43 @@ func (err ErrSchemaWriteDataValidation) GRPCStatus() *status.Status {
...
+			},
+		),
+	)
+}
+
+// NewMaxDepthExceededError creates a new MaxDepthExceededError.
+func NewMaxDepthExceededError(allowedMaximumDepth uint32, isCheckRequest bool) error {
+	if isCheckRequest {
+		return MaxDepthExceededError{
+			fmt.Errorf("the check request has exceeded the allowable maximum depth of %d: this usually indicates a recursive or too deep data dependency. Try running zed with --explain to see the dependency. See: https://spicedb.dev/d/debug-max-depth-check", allowedMaximumDepth),
```
vroldanbet, line 98: I think that instead of putting the "suggestion" or "recommendation" in the error message, we could put it as a new field in the error metadata. That way error can be what the error is, and then there can be 1 or more recommendations to look into as extra metadata

internal/services/shared/errors.go
```diff
@@ -104,6 +149,14 @@ func RewriteError(ctx context.Context, err error) error {
 	case errors.As(err, &relationNotFoundError):
 		return spiceerrors.WithCodeAndReason(err, codes.FailedPrecondition, v1.ErrorReason_ERROR_REASON_UNKNOWN_RELATION_OR_PERMISSION)
 
+	case errors.As(err, &maxDepthError):
+		if config == nil {
+			return spiceerrors.MustBugf("missing config for API error")
+		}
+
+		_, isCheckRequest := maxDepthError.Request.(*dispatchv1.DispatchCheckRequest)
+		return NewMaxDepthExceededError(config.MaximumAPIDepth, isCheckRequest)
```
vroldanbet, line 158: why not getting that config parameter from the dispatch error instead?

vroldanbet (approved):

internal/dispatch/dispatch.go
josephschorr: Sure, but it is only used for dispatchable requests

internal/services/shared/errors.go
josephschorr: I disagree, only because the error message is what most people are going to read, and debugging the issue is going to be top of their mind

internal/services/shared/errors.go
josephschorr: Because the dispatch error doesn't have the maximum depth and we'd have to plumb it all the way through; dispatch simply takes in the *current* depth remaining, not the maximum depth overall

PR #1402, opened 2023-06-21
disambiguate middleware dependency assertion in streaming APIs

vroldanbet:
the newly added API to assert middleware execution order and dependencies in tests had an ambiguous situation with streaming APIs. gRPC streaming interceptors
can execute its logic in the interceptor
handler itself (run when the stream is created) or as part of the streaming process (while sending and receiving messages).

Unfortunately the API didn't distinguish between those two, and assumed interceptors always run its logic in the stream wrapper, and this is not always the case.

These changes allow asserting dependencies based on execution of the interceptor, or execution of interceptor wrapper.

This Commit also removes an assertion in the gRPC prom middleware, since the dependencies is from the grpcauth MW to the prom MW, and not the other way around.


ecordell (approved):
LGTM

PR #1403, opened 2023-06-21
Cursor memory usage improvements

josephschorr:
Reduce the size of cursors to improve memory usage and potentially reduce GC


proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -80,9 +80,6 @@ message DispatchExpandResponse {
 }
 
 message Cursor {
-  string at_revision = 1 [ (validate.rules).string = {
```
jzelinskie, line 83: this needs to be `reserve 1`

josephschorr: Added

internal/graph/cursors.go
```diff
@@ -65,84 +55,57 @@ func (ci cursorInformation) withClonedLimits() cursorInformation {
...
-func (ci cursorInformation) hasHeadSection(name string) (bool, error) {
-	if ci.currentCursor == nil || len(ci.currentCursor.Sections) == 0 {
-		return false, nil
-	}
-
-	if ci.currentCursor.Sections[0] != name {
-		return false, spiceerrors.MustBugf("expected cursor section %s in %v", name, ci.currentCursor.Sections)
+// headSectionValue returns the string value found at the head of the incoming cursor.
+// If the incoming cursor is empty, returns empty.
+func (ci cursorInformation) headSectionValue() (string, bool) {
```
jzelinskie, line 63: I feel like we could easily have a generic function for head of a slice and the caller easily just use that instead of having this specialized function

internal/graph/cursors.go
```diff
@@ -65,84 +55,57 @@ func (ci cursorInformation) withClonedLimits() cursorInformation {
...
-// If the incoming cursor is empty, returns empty. If the incoming cursor does not start with the name,
-// fails with an error.
-func (ci cursorInformation) sectionValue(name string) (string, error) {
-	if ci.currentCursor == nil || len(ci.currentCursor.Sections) < 2 {
-		return "", nil
+// integerSectionValue returns the *integer* found  at the head of the incoming cursor.
+// If the incoming cursor is empty, returns 0. If the incoming cursor does not start with an
+// int value, fails with an error.
+func (ci cursorInformation) integerSectionValue() (int, error) {
+	valueStr, hasValue := ci.headSectionValue()
```
jzelinskie, line 75:
```go
if ci.currentCursor != nil {
    return 0, nil
}

valueStr, ok := slicez.Head(ci.currentCursor.Sections)
if !ok {
    return 0, nil
}
```

jzelinskie:
This needs a rebase, but otherwise looks good.

jzelinskie (approved):

PR #1406, opened 2023-06-23
Add check traces to assertion errors for better debugging

josephschorr:


proto/internal/developer/v1/developer.proto
```diff
@@ -99,6 +99,14 @@ message DeveloperError {
   // context holds the context for the error. For schema issues, this will be the
   // name of the object type. For relationship issues, the full relationship string.
   string context = 7;
+
+  // debug_information is the debug information for the check, if this error was raised
```
vroldanbet, line 103:
```suggestion
  // debug_information is the debug information for the dispatched check, if this error was raised
```

pkg/development/wasm/operations_test.go
```diff
@@ -1097,6 +1124,15 @@ assertFalse:
 						errors = append(errors, response.GetOperationsResults().Results[1].GetValidationResult().ValidationErrors...)
 					}
 				}
+
+				if tc.expectCheckTraces {
+					require.NotNil(t, errors[0].CheckDebugInformation)
+					require.NotNil(t, errors[0].CheckResolvedDebugInformation)
+
+					errors[0].CheckDebugInformation = nil
```
vroldanbet, line 1132: add a comment clarifying this is unset so we don't have to craft the expected values in the test input.
Still, would it be much work to check the actual values?

vroldanbet (approved):

PR #1408, opened 2023-06-26
Reconnect logging and tracing into the disconnected LR/RR context

josephschorr:


internal/graph/context.go
```diff
@@ -3,14 +3,32 @@ package graph
...
 	"context"
 
+	"go.opentelemetry.io/otel/trace"
+
+	log "github.com/authzed/spicedb/internal/logging"
 	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
 )
 
 // branchContext returns a context disconnected from the parent context, but populated with the datastore.
 // Also returns a function for canceling the newly context, without canceling the parent context.
```
jakedt, line 13: newly created

internal/graph/context.go
```diff
@@ -3,14 +3,32 @@ package graph
...
 // This is used when cancelation of a child context should not propagate upwards.
```
jakedt, line 14: Context cancellation never propagates upwards.

jakedt (changes requested):

internal/graph/context.go
josephschorr: Removed

jakedt (approved):

PR #1413, opened 2023-06-28
more robust nil caveat context handling for Spanner

vroldanbet:
As I experimented with Cloud Spanner exports and imports, I discovered the Dataflow Job that runs generates AVRO, once imported, leaves the tuple rows with a different serialization that the original one. So when
ContextualizedCaveatFrom is run:
- for the caveat name: NullString.Valid is true, even though NullString.Value is empty string
- for the caveat context: NullJSON.Valid is true, even though NullJSON.Value is nil

That leads the ContextualizedCaveatFrom method to panic when asserting the type.

I'm not able to reproduce this, but based on the panic stacktrace and the dataset (no tuple had caveats), it is only be explained by the values described above (null types coming with Valid=true, but Value being actually nil).


ecordell (approved):
LGTM

PR #1416, opened 2023-06-29
implement a more flexible logging middleware

vroldanbet:
the community has signaled interest for more flexibility around configuring the logs emitted by SpiceDB, but it's limited to what the grpc logging middleware supports.

Implementing a new middleware isn't a lot of work, it gives the opportunity to find optimizations, and gives more flexibility on what kind of features SpiceDB should support.

This initial implementation tries to adhere as close as possible to the original middleware logs, but with a few changes:
- request payloads can be omitted based on an exclusion list
- response payloads can be omitted based on an exclusion list
- entire methods can be omitted based on an exclusion list (e.g. do not log healthcheck)
- time is reported as milliseconds instead of a duration. This makes it easy to craft queries in any log processor based on duration (e.g. which where the requests that took more than 25ms?)

### DIFF
--- a/pkg/cmd/server/defaults.go
+++ b/pkg/cmd/server/defaults.go
@@ -7,21 +7,6 @@ import (
 	"net/http"
 	"net/http/pprof"
 
-	"github.com/fatih/color"
-	"github.com/go-logr/zerologr"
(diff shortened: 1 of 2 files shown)


josephschorr:
Will reopen when rescheduled

PR #1418, opened 2023-06-30
rm -rf pkg/util

jzelinskie:
This reorganizes our various generic utilities into their own packages.

I also took a moment to do two things:
1) adjust our Set type to implement the zerolog marshaling interface so that we didn't need the intermediary SliceSet
2) replace uses of stringz with the generic versions provided by golang.org/x/exp/slices

### DIFF
--- a/internal/caveats/diff.go
+++ b/internal/caveats/diff.go
@@ -6,8 +6,8 @@ import (
 	"golang.org/x/exp/maps"
 
 	"github.com/authzed/spicedb/pkg/caveats/types"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
-	"github.com/authzed/spicedb/pkg/util"
 )
 
 // DeltaType defines the type of caveat deltas.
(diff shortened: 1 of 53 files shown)


internal/graph/check.go
```diff
@@ -153,7 +154,7 @@ func (cc *ConcurrentChecker) checkInternal(ctx context.Context, req ValidatedChe
 	}
 
 	// Deduplicate any incoming resource IDs.
-	resourceIds := util.UniqueSlice(req.ResourceIds)
+	resourceIds := slicez.Unique(req.ResourceIds)
```
josephschorr, line 157: `slicez`? Why not just `slices`?

internal/caveats/diff.go
```diff
@@ -95,8 +95,8 @@ func DiffCaveats(existing *core.CaveatDefinition, updated *core.CaveatDefinition
 	}
 
 	deltas := make([]Delta, 0, len(existing.ParameterTypes)+len(updated.ParameterTypes))
-	existingParameterNames := util.NewSet[string](maps.Keys(existing.ParameterTypes)...)
-	updatedParameterNames := util.NewSet[string](maps.Keys(updated.ParameterTypes)...)
+	existingParameterNames := mapz.NewSet(maps.Keys(existing.ParameterTypes)...)
```
josephschorr, line 98: shouldn't a set be under a sets package? Or, alternatively, have all of these be under a generic collections package under util

jzelinskie: I think this helps communicate the implementation/performance of the set. If we get a bunch more set logic, it makes sense to make it an independent package

internal/graph/check.go
jzelinskie: nah collision -- that's the idea with the z-suffix convention

josephschorr (approved):
LGTM

Issue #1082, opened 2023-01-02
reading credentials from file option


mstrYoda:
Currently, SpiceDB takes credentials as program args or env variables which is makes it quite hard to manage in Kubernetes.

If we set db credentials as env vars on deployments it exposes our credentials (all cluster users can see deployment manifest), same as deployment args.

WDYT about having a credential file option?

jzelinskie:
Did you know about secrefRefs for environment variables in Kubernetes?

```yaml
...
    spec: 
      containers: 
        - env: 
            - name: myenv
              valueFrom:
                secretKeyRef:
                  name: mysecret
                  key: mysecretKey
```

mstrYoda:
> Did you know about secrefRefs for environment variables in Kubernetes?
> 
> ```yaml
> ...
>     spec: 
>       containers: 
>         - env: 
>             - name: myenv
>               valueFrom:
>                 secretKeyRef:
>                   name: mysecret
>                   key: mysecretKey
> ```
> 
> You can see more of an [example here](https://github.com/authzed/examples/blob/main/kubernetes/example.yaml).

Thank you for your response 🙏  I actually aware of it but still it is a Kubernetes resource and can be accessible from a user that have secret access. For security reasons we are only allow credentials inside a pod, that can not be accessible outside of it. So, it would be nice to have option to read credentials from a file :/

jzelinskie:
Ah, that's an interesting use-case. If you don't mind me asking, how do the secrets get into the pod securely?

A good compromise for this might be adding support for reading `.env` files rather than having build out a config file format.

mstrYoda:
> Ah, that's an interesting use-case. If you don't mind me asking, how do the secrets get into the pod securely?
> 
> A good compromise for this might be adding support for reading `.env` files rather than having build out a config file format.
> 
> WDYT?

Well we are using a custom integration solution with Hashicorp Vault. We are securely getting secrets from vault to a pods file system by using sidecar integration.

`.env` files sounds feasible. I would like to contribute for that feature.

jzelinskie:
That sounds good! It'll probably be a contribution to cobrautil, our CLI utility library.

Probably a change somewhere around here that adds loading a dotenv before syncing with the environment: https://github.com/jzelinskie/cobrautil/blob/3f120e7f595f25cdd4c18501c09c9fbb10127801/cobrautil.go#L34

mstrYoda:
[...]

PR #1426, opened 2023-07-06
Load config from spicedb.env

suttod:
closes https://github.com/authzed/spicedb/issues/1082
Depends on https://github.com/jzelinskie/cobrautil/pull/48
Doc: https://github.com/authzed/docs/pull/127

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	github.com/jackc/pgx-zerolog v0.0.0-20230315001418-f978528409eb
 	github.com/jackc/pgx/v5 v5.4.2
 	github.com/johannesboyne/gofakes3 v0.0.0-20230506070712-04da935ef877
-	github.com/jzelinskie/cobrautil/v2 v2.0.0-20230714172849-80717639cec5
+	github.com/jzelinskie/cobrautil/v2 v2.0.0-20230814170201-d023222f49a5
 	github.com/jzelinskie/stringz v0.0.2
 	github.com/lib/pq v1.10.9
 	github.com/lthibault/jitterbug v2.0.0+incompatible
--- a/go.sum
+++ b/go.sum
@@ -903,8 +903,8 @@ github.com/julienschmidt/httprouter v1.2.0/go.mod h1:SYymIcj16QtmaHHD7aYtjjsJG7V
 github.com/julienschmidt/httprouter v1.3.0/go.mod h1:JR6WtHb+2LUe8TCKY3cZOxFyyO8IZAc4RVcycCCAKdM=
 github.com/julz/importas v0.1.0 h1:F78HnrsjY3cR7j0etXy5+TU1Zuy7Xt08X/1aJnH5xXY=
 github.com/julz/importas v0.1.0/go.mod h1:oSFU2R4XK/P7kNBrnL/FEQlDGN1/6WoxXEjSSXO0DV0=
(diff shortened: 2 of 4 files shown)


josephschorr:
@suttod Can you please rebase?

suttod:
@josephschorr rebased. Please note that it won't compile until https://github.com/jzelinskie/cobrautil/pull/48 is merged.

jzelinskie:
https://github.com/jzelinskie/cobrautil/pull/48 is merged.

josephschorr:
@suttod please rebase

suttod:
@jzelinskie added some error handling to cobrautil before merging the changes, so that existence of env file is required now. This breaks backward compatibility of spicedb cli (raises: `Error: open spicedb.env: no such file or directory` with default config). My expectation was to silently ignore loading env file, if it doesn't exist (or at most log an info-level message).

Now I'm somewhat confused, which would be the best from user/developer experience:

1. The cleanest option would be to add a command line flag / envvar option to load further variables from (eg. `--spicedb-envfile=./spicedb.env`). Though it would introduce the need of some preprocessing of env/flags, and might be more difficult to maintain.
2. Implement some fallback in pkg/cmd/server/default.go to pass error with loading envfile. Would require double execution of `cobrautil.CobraRunFunc` (when the first fails). Can introduce further regression as `fs.PathError` migh be raised from other source as well. 
3. Instead of using `SyncViperDotEnvPreRunE` of cobrautil, load envfile in sipcedb (`pkg/cmd/server/default.go:DefaultPreRunE`), and handle error if file doesn't exist (log/pass)
4. Keep cobrautil siletnly pass if file doesn't exist.

Which do you prefer?

jzelinskie:
Hey there! Sorry for the delayed response.

I added the additional error handling with the intent of enabling a caller to choose the behavior they want in the case when the file is missing. Some observations looking back on it now:

1. It's awkward because you have to wrap the function to catch the error
2. I probably need to wrap the error and define yet another error type so that it can be more easily detected
3. 1+2 is a bunch of boilerplate for what is ultimately the default behavior

I think ultimately most programs using dotenv expect the default to load or fallback. If you agree with 1-3, I'll change it back to your original behavior and add logging when it loads successfully.

suttod:
Hi, I ultimately agree with your observations, especially with the "load-or-fallback" behavior.

So lets go with my 4th suggestion:
> Keep cobrautil siletnly pass if file doesn't exist.

Will you change in cobrautil, or shall I raise a separate MR?

suttod:
> Hi, I ultimately agree with your observations, especially with the "load-or-fallback" behavior.
> 
> So lets go with my 4th suggestion:
> 
> > Keep cobrautil siletnly pass if file doesn't exist.
> 
> Will you change in cobrautil, or shall I raise a separate MR?

Raised a new [MR](https://github.com/jzelinskie/cobrautil/pull/49) in cobrautil that only prints a notification when spicedb.env not found. Also rebased to main branch.

jzelinskie:
I added a commit so that the info line produced in cobrautil can use the zerolog logger in SpiceDB. Can we pull that in as well?

suttod:
> I added a commit so that the info line produced in cobrautil can use the zerolog logger in SpiceDB. Can we pull that in as well?

Seems reasonable, but please note that when parsing the dotenv file, we are still before processing any configuration. Zerolog as well.

jzelinskie (approved):

PR #1428, opened 2023-07-06
Reenable support of deleting caveats relationships without specifying the caveat

### LINKED ISSUE #1427: Cannot delete caveated relationships without specifying caveat explicitly
The behaviour shown below seems broken to me. I have a schema with this caveated relation on the `runnable_command` object:

```
relation engineer: my_permission_model/user with is_in_engineering_okta_group
```

... and have inserted a caveated relationship. A `read` command confirms it's there. But it seems I now can't delete it:

```
$ zed --token=test --endpoint=localhost:50051 --insecure relationship read my_permission_model/runnable_command:bla
my_permission_model/runnable_command:bla engineer my_permission_model/user:2[is_in_engineering_okta_group]
$ zed --token=test --endpoint=localhost:50051 --insecure relationship delete my_permission_model/runnable_command:bla engineer my_permission_model/user:2
Error: rpc error: code = InvalidArgument desc = subjects of type `my_permission_model/user` are not allowed on relation `my_permission_model/runnable_command#engineer`
Usage:
  zed relationship delete <resource:id> <relation> <subject:id> [flags]

Flags:
  -h, --help   help for delete
      --json   output as JSON

Global Flags:
      --certificate-path string     path to certificate authoriy used to verify secure connections
      --endpoint string             spicedb gRPC API endpoint
      --insecure                    connect over a plaintext connection
      --log-format string           format of logs ("auto", "console", "json") (default "auto")
      --log-level string            verbosity of logging ("trace", "debug", "info", "warn", "error") (default "info")
      --no-verify-ca                do not attempt to verify the server's certificate chain and host name
      --permissions-system string   permissions system to query
      --skip-version-check          if true, no version check is performed against the server
      --token string                token used to authenticate to SpiceDB
```

The docs at https://authzed.com/docs/reference/caveats say:

> When deleting a relationship, a caveat does not need to be specified; the matching relationship will be deleted if present.

... but as you see, things don't seem to work as described. Instead, it just complains that the relationship I'm asking it to delete is one that isn't allowed to exist and refuses to delete it. (I think this happens because it wouldn't be allowed to exist without a caveat on it; I get the same error message if I try to create *such* a relationship without a `--caveat` argument.)

@josephschorr assures me in a support chat that the bug here is serverside and not in `zed`, so I'm filing here. :)

SpiceDB version: 1.21.0. Zed version: 0.10.2.

josephschorr:
Fixes #1427

### DIFF
--- a/internal/relationships/validation.go
+++ b/internal/relationships/validation.go
@@ -3,22 +3,86 @@ package relationships
 import (
 	"context"
 
+	"github.com/samber/lo"
+
 	"github.com/authzed/spicedb/internal/namespace"
 	"github.com/authzed/spicedb/pkg/caveats"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	ns "github.com/authzed/spicedb/pkg/namespace"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
-// ValidateRelationships performs validation on the given relationships, ensuring that
+// ValidateRelationshipUpdates performs validation on the given relationship updates, ensuring that
 // they can be applied against the datastore.
-func ValidateRelationships(
+func ValidateRelationshipUpdates(
+	ctx context.Context,
+	reader datastore.Reader,
+	updates []*core.RelationTupleUpdate,
+) error {
+	rels := lo.Map(updates, func(item *core.RelationTupleUpdate, _ int) *core.RelationTuple {
+		return item.Tuple
+	})
+
+	// Load namespaces and caveats.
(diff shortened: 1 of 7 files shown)


internal/relationships/validation.go
```diff
@@ -3,22 +3,86 @@ package relationships
...
+	}
+
+	return nil
+}
+
+// ValidateRelationshipsForCreateOrTouch performs validation on the given relationships to be written, ensuring that
+// they can be applied against the datastore.
+//
+// NOTE: This method *cannot* be used for relationships that will be deleted.
+func ValidateRelationshipsForCreateOrTouch(
```
jzelinskie, line 59: Either call this `...ForWrite` or rename the constant to be `...ForCreateOrTouch`. I lean towards the latter.

josephschorr: Done

jzelinskie (approved):

jzelinskie (approved):

PR #1430, opened 2023-07-11
Fix pgx leak in 5.4.x

vroldanbet:
this vendors commit 2bf5a614018d35b5db4259b5541497ab35175882
which includes fix to pgx memory leak.

See https://github.com/jackc/pgx/issues/1676

Also found an instance `Rows.Close()` wasn't being called


josephschorr (approved):
LGTM

Let's make sure to unpin once pgx contains the fix in a released version

PR #1436, opened 2023-07-13
Fixing a bug in the comment

Posya:
It seems to be an old comment in the code, in this line the datastore is being initialized


Posya:

PR #1438, opened 2023-07-14
Small fix

Posya:
It seems to be an old comment in the code, in this line the datastore is being initialized


Posya:

vroldanbet:
Hey @Posya, thanks for the typo fix. It seems that the commit log shows 2 different users, and 3 commits, would you mind squashing all together into a single commit with a single user?

PR #1440, opened 2023-07-14
Have remote dispatches be counted in the dispatch count

### LINKED ISSUE #226: Client-side and server-side dispatch cache should report different costs
The client-side cache should report `0` downstream dispatches, and the server side cache should report `1`. Right now it's hard-coded to always report zero:

https://github.com/authzed/spicedb/blob/ac17a3b5b03ccf34d4594e73587072e2c3a5ca1b/internal/dispatch/caching/caching.go#L133

This value should be configurable and the client-side (before remote dispatch) instantiation should use `0` and the server side (after remote dispatch) should use `1`.

josephschorr:
Fixes #226


ecordell (approved):
LGTM, we discussed some offline how this may overcount

ecordell (approved):
LGTM

Issue #1444, opened 2023-07-15
Broken garbage collection backoff (Postgres)


johanneswuerbach:
While we are seeing problems with garbage collection https://github.com/authzed/spicedb/issues/1443, there seems to be also a problem with the backoff implementation.

Eventually the `next-attempt-in` flips to `-0.000001`.

Example logs (all consecutive) 

```
{"level":"warn","error":"error retrieving watermark: no rows in result set","next-attempt-in":48186.303923,"time":"2023-07-15T08:04:01Z","message":"error attempting to perform garbage collection"}
{"level":"warn","error":"error retrieving watermark: no rows in result set","next-attempt-in":50827.821022,"time":"2023-07-15T08:04:49Z","message":"error attempting to perform garbage collection"}
{"level":"warn","error":"error retrieving watermark: no rows in result set","next-attempt-in":-0.000001,"time":"2023-07-15T08:05:40Z","message":"error attempting to perform garbage collection"}
{"level":"warn","error":"error retrieving watermark: no rows in result set","next-attempt-in":-0.000001,"time":"2023-07-15T08:05:41Z","message":"error attempting to perform garbage collection"}
{"level":"warn","error":"error retrieving watermark: no rows in result set","next-attempt-in":-0.000001,"time":"2023-07-15T08:05:41Z","message":"error attempting to perform garbage collection"}
```

Afterwards the garbage collection attempts seem to busy loop and create a lot outgoing connections.

* Version: `v1.23.0`
* Datastore: `postgres`

vroldanbet:
@johanneswuerbach thanks for reporting, I hit this problem recently while using the same library in a different project, the fix is trivial. I'll come up with a PR.

PR #1445, opened 2023-07-15
fixes incorrect usage of backoff in GC

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/1444

once `MaxElapsedTime` has elapsed, the backoff switches to the value of `Stop` which is -1. This means it
will effectively remove the backoff and a storm of requests will hit the DB.

I also changed it in every instance I found because the default behaviour is dangerous.

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -43,12 +43,13 @@ var (
 		Help:      "The number of stale namespaces deleted by the datastore garbage collection.",
 	})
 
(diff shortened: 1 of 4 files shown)


josephschorr:
@vroldanbet Can we add a test for this somehow? Perhaps initializing the GC with settings that get it to reproduce the old behavior?

vroldanbet:
@josephschorr done, please have a look

josephschorr (approved):
LGTM

Issue #1410, opened 2023-06-27
CEL Best Practices


TristonianJones:
As the CEL maintainer, I thought it would be useful to share the following tips and best practices to improve your experience with CEL:

### Use `DynType` over `AnyType`

The `AnyType` is intended to map to `google.protobuf.Any` and any object type rather than the set of all possible value types. A common gotcha is to use a `AnyType` result as the iteration range on a comprehension. On the CEL side, we intend to advise people against using the `AnyType` since cel-go developers mistake this as equivalent to a Go `any` value which is not the case.

###  Partial Eval

The result of a partial evaluation is not an error, but a `types.Unknown` value. In the following snippet, you can expect the behavior as follows:

```go
out, _, err := prg.Eval(<...>)
// Hard error, no additional data will change the result
if err != nil {
   // terminate evaluation
}
// Partial evaluation result, more data may shift the result to a known value or error
if types.IsUnknown(out) {
  // Expression ids from the AST nodes relevant to the completion of the evaluation
  // are included in the out.(types.Unknown) value.
}
// Success! You've got a known value
return val
```

### Attribute Errors

If your customers are encountering issues with runtime errors during field traversals (such as 'no such attribute'), consider enabling `cel.OptionalTypes()` which will make it simpler to write safe-traversal expressions: https://github.com/google/cel-spec/wiki/proposal-246

Hopefully this helps, and I would be happy to review a PR if you'd like some assistance on best practices with CEL.

Cheers!

-Tristan

ecordell:
@TristonianJones Thanks for reaching out!

@josephschorr is planning to make some updates to our CEL usage tomorrow, we'd love your feedback on that PR when it's available.

josephschorr:
@TristonianJones Can you provide some further guidance on the partial evaluation handling? Right now we handle this case by parsing the error message string (yes, its ugly): https://github.com/authzed/spicedb/blob/main/pkg/caveats/eval.go#L115

I tried changing the above to the latest version of CEL and the code to check `types.IsUnknown`, but it is returning false and I'm still receiving an error indicating that the attribute does not exist.

Do I perhaps need to turn off `celopts = append(celopts, cel.EvalOptions(cel.OptPartialEval))`?

Also: where is `OptionalTypes` defined? I don't see it in the `cel.` options list

TristonianJones:
@josephschorr thanks for the ping.

[...]

PR #1451, opened 2023-07-20
Implement CEL best practices

josephschorr:
Fixes https://github.com/authzed/spicedb/issues/1410


pkg/caveats/eval.go
```diff
@@ -106,45 +102,39 @@ func EvaluateCaveatWithConfig(caveat *CompiledCaveat, contextValues map[string]a
...
+
+	// If the value produced has Unknown type, then it means required context was missing.
+	if types.IsUnknown(val) {
+		unknownVal := val.(*types.Unknown)
+		missingVarNames := make([]string, 0, len(contextValues))
+		for _, id := range unknownVal.IDs() {
+			trails, ok := unknownVal.GetAttributeTrails(id)
+			if ok {
+				for _, attributeTrail := range trails {
+					missingVarNames = append(missingVarNames, attributeTrail.Variable())
```
TristonianJones, line 122: Consider using `AttributeTrail.String()` in the event that you have more specific missing attributes in the future.

pkg/caveats/types/ipaddress.go
```diff
@@ -26,7 +26,7 @@ func MustParseIPAddress(ip string) IPAddress {
 	return ipAddress
 }
 
-var ipaddressCelType = types.NewTypeValue("IPAddress", traits.ReceiverType)
+var ipaddressCelType = cel.ObjectType("IPAddress", traits.ReceiverType)
```
TristonianJones, line 29: Does the `IPAddress` type have fields that are accessible, or just methods? If you don't expose any fields to the caller directly, you can consider making this an `OpaqueType`.

TristonianJones (commented):
Looking good!

pkg/caveats/env.go
```diff
@@ -72,6 +72,10 @@ func (e *Environment) asCelEnvironment() (*cel.Env, error) {
 	// DefaultUTCTimeZone: ensure all timestamps are evaluated at UTC
 	opts = append(opts, cel.DefaultUTCTimeZone(true))
 
+	// OptionalTypes: enable optional typing syntax, e.g. `sometype?.foo`
+	// See: https://github.com/google/cel-spec/wiki/proposal-246
+	opts = append(opts, cel.OptionalTypes(cel.OptionalTypesVersion(0)))
```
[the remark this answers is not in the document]
josephschorr: Is there a downside to doing so?

pkg/caveats/eval.go
```diff
@@ -106,45 +102,39 @@ func EvaluateCaveatWithConfig(caveat *CompiledCaveat, contextValues map[string]a
...
-				return &CaveatResult{
-					val:             val,
-					details:         details,
-					parentCaveat:    caveat,
-					contextValues:   contextValues,
-					missingVarNames: strings.Split(found[2], " "),
-					isPartial:       true,
-				}, nil
+	val, details, err := prg.Eval(activation)
+
```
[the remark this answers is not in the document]
josephschorr: Good call. Fixed

pkg/caveats/eval.go
josephschorr: Changed

pkg/caveats/types/ipaddress.go
josephschorr: Just the `in_cidr` function right now. You'd recommend making it Opaque for that?

TristonianJones: Either way is fine. If you mark something as an object type, then there's some expectation that it will have fields that are directly accessible in the expression. If you always intend to expose functionality through methods, then consider an `OpaqueType` since it's slightly less overhead.

TristonianJones (approved):

pkg/caveats/types/ipaddress.go
josephschorr: Gotcha. Changed

pkg/caveats/env.go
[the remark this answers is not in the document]
vroldanbet: Do we want to add a test to make sure that optionals are enabled?
EDIT: saw you added some tests using optional types 👍🏻

pkg/caveats/eval.go
```diff
@@ -51,8 +47,8 @@ func (cr CaveatResult) PartialValue() (*CompiledCaveat, error) {
 		return nil, fmt.Errorf("result is fully evaluated")
 	}
 
-	expr := interpreter.PruneAst(cr.parentCaveat.ast.Expr(), cr.details.State())
-	return &CompiledCaveat{cr.parentCaveat.celEnv, cel.ParsedExprToAst(&exprpb.ParsedExpr{Expr: expr}), cr.parentCaveat.name}, nil
+	expr := interpreter.PruneAst(cr.parentCaveat.ast.Expr(), cr.parentCaveat.ast.SourceInfo().GetMacroCalls(), cr.details.State())
```
vroldanbet, line 50: I don't know what this does. Does this further remove parts of the AST that aren't necessary for evaluation and getting debug symbols? Macros can and have been used in caveats, would that make troubleshooting errors more complicated if source info is not present?

pkg/caveats/eval.go
```diff
@@ -106,46 +102,40 @@ func EvaluateCaveatWithConfig(caveat *CompiledCaveat, contextValues map[string]a
...
-				details:         details,
-				parentCaveat:    caveat,
-				contextValues:   contextValues,
-				missingVarNames: nil,
-				isPartial:       true,
-			}, nil
+	// If the value produced has Unknown type, then it means required context was missing.
+	if types.IsUnknown(val) {
+		unknownVal := val.(*types.Unknown)
+		missingVarNames := make([]string, 0, len(contextValues))
```
vroldanbet, line 121: why not `len(unknownVal.IDs())`?

pkg/caveats/eval_test.go
```diff
@@ -107,7 +107,7 @@ func TestEvaluateCaveat(t *testing.T) {
 			"",
 			true,
 			"",
-			[]string{"b"},
```
vroldanbet, line 110: oh wasn't this getting short-circuited before? 😕

pkg/caveats/types/ipaddress.go
```diff
@@ -26,7 +25,7 @@ func MustParseIPAddress(ip string) IPAddress {
 	return ipAddress
 }
 
-var ipaddressCelType = types.NewTypeValue("IPAddress", traits.ReceiverType)
+var ipaddressCelType = cel.OpaqueType("IPAddress")
```
vroldanbet, line 28: Is this a breaking change? is it possible this would render existing serialized CEL AST incompatible?

pkg/caveats/types/map.go
```diff
@@ -7,9 +7,9 @@ import (
 )
 
 func init() {
-	registerMethodOnDefinedType(cel.MapType(cel.StringType, cel.AnyType),
+	registerMethodOnDefinedType(cel.MapType(cel.StringType, cel.DynType),
```
vroldanbet, line 10: Is this a breaking change? is it possible this would render existing serialized CEL AST incompatible?

pkg/caveats/types/basic.go
```diff
@@ -73,7 +73,7 @@ func convertNumericType[T int64 | uint64 | float64](value any) (any, error) {
 }
 
 var (
-	AnyType     = registerBasicType("any", cel.AnyType, func(value any) (any, error) { return value, nil })
```
vroldanbet, line 76: Is this a breaking change? is it possible this would render existing serialized CEL AST incompatible?

internal/namespace/caveats.go
```diff
@@ -14,7 +14,16 @@ import (
 // definition, including usage of the parameters.
 func ValidateCaveatDefinition(caveat *core.CaveatDefinition) error {
 	// Ensure all parameters are used by the caveat expression itself.
-	deserialized, err := caveats.DeserializeCaveat(caveat.SerializedExpression)
+	parameterTypes, err := caveattypes.DecodeParameterTypes(caveat.ParameterTypes)
```
vroldanbet, line 17: do you foresee this could cause existing deployments to throw this error somehow now that we add the extra layer of type validation?

pkg/caveats/eval.go
josephschorr: It prune the AST of any executed branches, simplifying it down to only the remaining branches. This is tested in the unit tests

pkg/caveats/types/ipaddress.go
josephschorr: No, I don't believe so. It simply is the access semantics on the type

pkg/caveats/types/map.go
josephschorr: No

pkg/caveats/types/basic.go
josephschorr: No

internal/namespace/caveats.go
josephschorr: Yes, this is possible. I plan to put a warning in the release notes

vroldanbet: but how would existing deployments fix it?

josephschorr: they'd rewrite their caveat expression

vroldanbet: how would they tell ahead of time their caveat is "incorrect"

josephschorr: You can't; same way you can't tell today if it is incorrect if you take in a dynamic value and it type check fails at runtime. It would be the exact same thing: you passed a wrong value in context and it failed to eval, causing a type error

pkg/caveats/eval.go
josephschorr: Fixed

josephschorr:
Updated

vroldanbet (approved):

PR #1452, opened 2023-07-20
Remove the old "dashboard" from SpiceDB

josephschorr:
It didn't really provide a lot of value over what `zed` provides and the code had grown stale


ecordell (approved):
LGTM

PR #1453, opened 2023-07-21
introduce Bulk Check API

### LINKED ISSUE #621: Bulk check API
Hello,

I'm evaluating spicedb (very nice product btw !).

I'd like to handle the following use case:

> My front-end app calls my rest API to display the list of objects returned by the API.
> The last column of the list is an 'Actions' column.
> This column contains the list of actions that can be executed by the current user on the given object.
> 

I'd imagine calling SpiceDb API to evaluate a [list of permissions] for a given user and for a [list of objects].
This way, I'd know the buttons I need to display in each row of the actions column.

I could of course do that by calling the checkPermission API in a loop, but it won't scale

Is there something like a bulk checkPermission API ?

Thanks !

vroldanbet:
Introduces Bulk Check API, which allows issuing multiple CheckPermission requests at once

- API change: https://github.com/authzed/api/pull/75
- `authzed-go` changes: https://github.com/authzed/authzed-go/pull/136 

## Features Implemented
- clusters similar requests with different `resourceID` into a single dispatch
- clustered requests are dispatched with bounded concurrency 
- dispatch concurrency is added as option (but not exposed via CLI flags yet)
- chunks requests in case the number of resources to check exceed a default limit of 100
- does not fail the entire request if one element fails
- Debug is not supported for BulkChecks

## TODO

- [ ] How to test the number of dispatched checks?

Fixes #621


internal/services/v1/clustering.go
```diff
@@ -0,0 +1,77 @@
...
+	chunkedResourceIDs [][]string
+}
+
+type clusteringParameters struct {
+	atRevision           datastore.Revision
+	maximumAPIDepth      uint32
+	maxCaveatContextSize int
+}
+
+// clusterItems takes a slice of BulkCheckPermissionRequestItem and cluster them based
```
josephschorr, line 28: cluster -> clusters

internal/services/v1/clustering.go
```diff
@@ -0,0 +1,77 @@
...
+// on using the same resource type, subject type, subject id, and caveat.
```
josephschorr, line 29: same permission

internal/services/v1/clustering.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+		}
+	}
+
+	return maps.Values(clustered), nil
+}
+
+func bulkItemHash(item *v1.BulkCheckPermissionRequestItem) string {
+	contextHash := caveats.StableContextStringForHashing(item.Context)
+	key := item.Resource.ObjectType + item.Permission + item.Subject.Object.ObjectType + item.Subject.Object.ObjectId + normalizeSubjectRelation(item.Subject) + contextHash
```
josephschorr, line 75: so there is a risk (however small) that this fails due to not having separators

for example: resource_type=foo, permission=bar and resource_type=fo and permission=obar

Unlikely, but it *could* be a security issue if explicitly used, so we should make sure to have delimeters.

I'd use the same hashing code we're using for LR cursors, and extend it for check items here

internal/services/v1/clustering.go
```diff
@@ -0,0 +1,77 @@
...
+						Namespace: item.Subject.Object.ObjectType,
+						ObjectId:  item.Subject.Object.ObjectId,
+						Relation:  normalizeSubjectRelation(item.Subject),
+					},
+					CaveatContext: caveatContext,
+					AtRevision:    params.atRevision,
+					MaximumDepth:  params.maximumAPIDepth,
+					DebugOption:   computed.NoDebugging,
+				},
+				chunkedResourceIDs: [][]string{{item.Resource.ObjectId}},
```
josephschorr, line 65: might be worth allocating this as a slice of capacity equivalent to the max allowed

internal/services/v1/clustering_test.go
```diff
@@ -0,0 +1,288 @@
...
+func TestClusterItems(t *testing.T) {
+	testCases := []struct {
+		name      string
+		requests  []string
+		response  []expectedClusteredRequest
+		chunkSize uint16
+		err       string
+	}{
+		{
+			name: "different subjects cannot be clustered",
```
josephschorr, line 34: please add a test like I described above and ensure it doesn't accidentally cluster

internal/services/v1/clustering_test.go
```diff
@@ -0,0 +1,288 @@
...
+			MaxBulkCheckDispatchChunkSize = originalChunkSize
+			if tt.err != "" {
+				require.ErrorContains(t, err, tt.err)
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, len(tt.response), len(ccp))
+				sort.Slice(ccp, func(first, second int) bool {
+					firstParams := ccp[first].params
+					secondParams := ccp[second].params
+					firstKey := firstParams.ResourceType.Namespace + firstParams.ResourceType.Relation +
```
josephschorr, line 241: use the key function to do this

internal/services/v1/clustering_test.go
```diff
@@ -0,0 +1,288 @@
...
+				atRevision:           datastore.NoRevision,
+				maxCaveatContextSize: math.MaxInt,
+				maximumAPIDepth:      1,
+			}
+			originalChunkSize := MaxBulkCheckDispatchChunkSize
+			if tt.chunkSize > 0 {
+				MaxBulkCheckDispatchChunkSize = tt.chunkSize
+			}
+			ccp, err := clusterItems(context.Background(), cp, items)
+			MaxBulkCheckDispatchChunkSize = originalChunkSize
```
josephschorr, line 232: Noooooo :)

Pass the chunk size into the function

internal/services/v1/experimental.go
```diff
@@ -386,6 +405,138 @@ func (es *experimentalServer) BulkExportRelationships(
 	return nil
 }
 
+func (es *experimentalServer) BulkCheckPermission(ctx context.Context, req *v1.BulkCheckPermissionRequest) (*v1.BulkCheckPermissionResponse, error) {
+	atRevision, checkedAt, err := consistency.RevisionFromContext(ctx)
+	if err != nil {
+		return nil, es.rewriteError(ctx, err)
+	}
+
+	// identifies checks with same permission and caveats over different resources and cluster them
```
josephschorr, line 414: identifies -> Identify

josephschorr, line 414: indicate that the subject has to be the same as well

josephschorr:
Addressed my own review comments, handing over to @jzelinskie for review

irl-segfault:
huge

internal/services/integrationtesting/consistency_test.go
```diff
@@ -411,6 +418,31 @@ func validateLookupResources(t *testing.T, vctx validationContext) {
...
+										ObjectType: resourceRelation.Namespace,
+										ObjectId:   resolvedResource.ResourceObjectId,
+									},
+									Permission: resourceRelation.Relation,
+									Subject: &v1.SubjectReference{
+										Object: &v1.ObjectReference{
+											ObjectType: subject.Namespace,
+											ObjectId:   subject.ObjectId,
+										},
+										OptionalRelation: stringz.Default(subject.Relation, "", tuple.Ellipsis),
```
jzelinskie, line 433: there's a `DefaultEmpty` for this specific use case, haha

josephschorr: I copied this from elsewhere in our code

jzelinskie (changes requested):

internal/services/v1/clustering.go
```diff
@@ -0,0 +1,81 @@
...
+	"golang.org/x/exp/maps"
+
+	"github.com/authzed/spicedb/internal/graph/computed"
+	"github.com/authzed/spicedb/pkg/datastore"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+var MaxBulkCheckDispatchChunkSize = datastore.FilterMaximumIDCount
+
+type clusteredCheckParameters struct {
```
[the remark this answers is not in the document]
josephschorr: Done

internal/services/v1/hash.go
```diff
@@ -11,6 +11,17 @@ import (
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
+func computeBulkCheckPermissionItemHashWithoutResourceID(req *v1.BulkCheckPermissionRequestItem) (string, error) {
```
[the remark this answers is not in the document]
josephschorr: > Is this exclusively being used for grouping together the items in the bulk check? 

Yes. However, we need to be absolutely *certain* that the groupings are distinct, which is why I'm using the same method I use in LR and LS for call hash computation

> computeCallHash does a lot of work to ensure that the full protobuf types are hashed that is additional logic with allocations that are unnecessary.

Which protobuf types included in the hash are not necessary?

> If all you need is something unique and you don't want to store an struct as the key in the map, just do use the standard library's maphash(paramStrings)

So @vroldanbet  was doing this before, but it had a security issue in the way the string was computed; that is why I changed to an explicitly tested call hash like we use elsewhere

internal/services/v1/clustering.go
```diff
@@ -0,0 +1,81 @@
...
+
+type clusteringParameters struct {
+	atRevision           datastore.Revision
+	maximumAPIDepth      uint32
+	maxCaveatContextSize int
+}
+
+// clusterItems takes a slice of BulkCheckPermissionRequestItem and clusters them based
+// on using the same permission, subject type, subject id, and caveat.
+func clusterItems(ctx context.Context, params clusteringParameters, items []*v1.BulkCheckPermissionRequestItem, maxBatchSize uint16) ([]clusteredCheckParameters, error) {
```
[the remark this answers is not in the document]
josephschorr: > use our generic batching function in the places where we need it

Do you mean to extract it into chunks or...?

internal/services/v1/experimental.go
```diff
@@ -386,6 +405,139 @@ func (es *experimentalServer) BulkExportRelationships(
...
+						{
+							NamespaceName: item.params.Subject.Namespace,
+							RelationName:  stringz.DefaultEmpty(item.params.Subject.Relation, graph.Ellipsis),
+							AllowEllipsis: true,
+						},
+					}, ds)
+
+				var pairs []*v1.BulkCheckPermissionPair
+				var rcr map[string]*dispatchv1.ResourceCheckResult
+				if err == nil {
```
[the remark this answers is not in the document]
josephschorr: Fixed

internal/services/v1/experimental.go
```diff
@@ -386,6 +405,139 @@ func (es *experimentalServer) BulkExportRelationships(
...
+	permissionship, partialCaveat := checkResultToAPITypes(checkResult)
+	return &v1.BulkCheckPermissionPair_Item{
+		Item: &v1.BulkCheckPermissionResponseItem{
+			Permissionship:    permissionship,
+			PartialCaveatInfo: partialCaveat,
+		},
+	}
+}
+
+func requestItemFromResourceAndParameters(params clusteredCheckParameters, resourceID string) (*v1.BulkCheckPermissionRequestItem, error) {
```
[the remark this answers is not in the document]
josephschorr: The pair has different response block types based on whether an error occurred or not

josephschorr:
@jzelinskie Rebased and updated. Unfortunately, because the authzed-go changes are not yet merged, the tests will break on the slash tests that were recently added. We need to get the API merged and then rebase authzed-go

josephschorr:
Moved to latest version of authzed-go

internal/services/v1/experimental.go
```diff
@@ -386,6 +406,153 @@ func (es *experimentalServer) BulkExportRelationships(
...
+				Response: &v1.BulkCheckPermissionPair_Error{
+					Error: statusResp.Proto(),
+				},
+			})
+		}
+
+		return nil
+	}
+
+	appendResultsForCheck := func(params computed.CheckParameters, resourceIDs []string, metadata *dispatchv1.ResponseMeta, results map[string]*dispatchv1.ResourceCheckResult) error {
```
jzelinskie, line 459: metadata isn't actually being used here

internal/services/v1/experimental.go
```diff
@@ -386,6 +406,153 @@ func (es *experimentalServer) BulkExportRelationships(
...
+		return nil, es.rewriteError(ctx, err)
+	}
+
+	bulkResponseMutex := sync.Mutex{}
+	resp := &v1.BulkCheckPermissionResponse{CheckedAt: checkedAt}
+	tr := taskrunner.NewPreloadedTaskRunner(ctx, es.bulkCheckMaxConcurrency, len(groupedItems))
+
+	appendResultsForError := func(group groupedCheckParameters, err error) error {
+		bulkResponseMutex.Lock()
+		defer bulkResponseMutex.Unlock()
```
jzelinskie, line 433: We can move this to just above the for loop, because rewriting the error doesn't require holding the lock.

internal/services/v1/experimental.go
```diff
@@ -386,6 +406,153 @@ func (es *experimentalServer) BulkExportRelationships(
...
+		maxCaveatContextSize: es.maxCaveatContextSize,
+		maximumAPIDepth:      es.maximumAPIDepth,
+	}
+	groupedItems, err := groupItems(ctx, groupingParams, req.Items)
+	if err != nil {
+		return nil, es.rewriteError(ctx, err)
+	}
+
+	bulkResponseMutex := sync.Mutex{}
+	resp := &v1.BulkCheckPermissionResponse{CheckedAt: checkedAt}
```
jzelinskie, line 428: I can see pros and cons to extracting these into a struct and then having `FailGroup()`,`AppendResults()`,`Response()` methods.

I don't know if one is fundamentally more efficient (closure vs type).

jzelinskie (changes requested):

internal/services/v1/experimental.go
```diff
@@ -386,6 +406,153 @@ func (es *experimentalServer) BulkExportRelationships(
...
 
+func (es *experimentalServer) BulkCheckPermission(ctx context.Context, req *v1.BulkCheckPermissionRequest) (*v1.BulkCheckPermissionResponse, error) {
+	atRevision, checkedAt, err := consistency.RevisionFromContext(ctx)
+	if err != nil {
+		return nil, es.rewriteError(ctx, err)
+	}
+
+	// Identify checks with same permission+subject over different resources and group them. This is doable because
+	// the dispatching system already internally supports this kind of batching for performance.
+	groupingParams := groupingParameters{
```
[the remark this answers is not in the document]
josephschorr: Because it acts as a logical grouping and avoids more parameters to the call; I've inlined it

internal/services/v1/experimental.go
```diff
@@ -386,6 +406,153 @@ func (es *experimentalServer) BulkExportRelationships(
...
+						},
+						{
+							NamespaceName: group.params.Subject.Namespace,
+							RelationName:  stringz.DefaultEmpty(group.params.Subject.Relation, graph.Ellipsis),
+							AllowEllipsis: true,
+						},
+					}, ds)
+				if err != nil {
+					return appendResultsForError(group, err)
+				}
```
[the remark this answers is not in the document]
josephschorr: The error should be returned to the task runner; canceling the context above is (to me) a non-obvious way to handle errors

internal/services/v1/experimental.go
josephschorr: Good catch; fixed

internal/services/v1/experimental.go
josephschorr: *shrug* I think it would be about as complex

josephschorr:
Rebased and updated

internal/services/v1/experimental.go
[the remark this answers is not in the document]
jzelinskie: Wait isn't there a scenario where a second batch fails, and then it appends the whole group as failed after it's already appended the first batch in the group as a success?

internal/services/v1/experimental.go
jzelinskie: I don't see it fixed yet

josephschorr: https://github.com/authzed/spicedb/pull/1453/files#diff-46f0df22900bde355de03fe5c3e40b0dd39ad7332bd23b41a5dc8079e526d0afR482

internal/services/v1/experimental.go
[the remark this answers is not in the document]
josephschorr: Yes? Not sure I'm following the question here

internal/services/v1/experimental.go
jzelinskie: I took a peak at the assembly and I don't think the difference is huge. Let's go with this until a profiler tells us otherwise.

internal/services/v1/hash.go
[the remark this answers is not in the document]
jzelinskie: Looked into this in a bit. We could ultimately do something like

```go
pb := protobuf.CloneVT()
pb.ResourceID = ""
return hash(pb.MarshalVTStrict())
```

It's a hell of a lot less code to maintain and test, but I'm not sure it's more efficient. Once again, let's roll with what's here and only address what the profiler tells us to.

jzelinskie (approved):

PR #1455, opened 2023-07-23
move as many tools as possible into the mage module

ecordell:


vroldanbet (changes requested):
Changes look good to me but I believe you didn't mean to commit the compiled SpiceDB binary

ecordell:
Oops! Fixed

vroldanbet (approved):

PR #1456, opened 2023-07-25
moves analyzers to Go 1.20

vroldanbet:
this was causing problems with running
govulncheck


josephschorr (approved):
LGTM

vroldanbet:
This does not actually fix it, unfortunately

Issue #1457, opened 2023-07-26
make spanner changelog stream retention configurable


vroldanbet:
The SpiceDB Datastore implementation for Spanner now uses changelogs as of version 1.23.0 instead of its own bookkeeping changelog table. The migration used the default retention of 1 day. This is enough for most users, but some may want to change this. Spanner supports up to 7 days of data retention.

josephschorr:
So this raises an interesting question: The change stream is created via one of the migrations, but the retention period is configured in the same statement... so it would have to be a flag on the `migrate` call, and it would only be applicable once.

Any thoughts on how to make this devex better?

vroldanbet:
there is already a precedent of leaving certain DB configuration flags outside of `migrate`, like `cockroach_cluster_gc_window_nanos`. We can either have folks change that themselves or we could introduce support for migration revision-specific options. Not a fan of it though, would have to be pushed to the operator too, and require folks to edit CRs to add the specific options they want to be passed to the corresponding migration. Thoughts / other ideas @ecordell?

PR #1458, opened 2023-07-26
removes Spanner phased migrations

vroldanbet:
The challenge has been the GC related tests. Now with change streams GC is implicit via the configured retention. Given the Spanner emulator does still not have support for CS, I've disabled those tests that rely on GC.

I've also added the default retention of 24h for the remote clocks, used to check validity of a revision.

See https://cloud.google.com/spanner/docs/change-streams#data-retention
See https://github.com/authzed/spicedb/issues/1457


pkg/datastore/test/datastore.go
```diff
@@ -39,9 +39,36 @@ type TestableDatastore interface {
 	ExampleRetryableError() error
 }
 
-// AllExceptWatch runs all generic datastore tests on a DatastoreTester, except
-// those invoking the watch API.
-func AllExceptWatch(t *testing.T, tester DatastoreTester) {
+type Categories map[string]struct{}
```
ecordell, line 42: not worth changing everything just for this, but we might want to consider ginkgo for these tests. There are a lot of built-in features that would simplify this for us (i.e. there's an existing `Label` system for labelling / running tests with labels.

ecordell (approved):
LGTM

PR #1462, opened 2023-07-27
update prometheuspb module to new upstream domain

rxu-plaid:
https://buf.build/docs/migration-guides/migrate-remote-generation-alpha

the `prometheuspb` module used relies on a domain that no longer as a valid certificate. Following the steps outlined in the guide above, we can just switch the module to the new upstream domain

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/authzed/spicedb
 go 1.20
 
 require (
+	buf.build/gen/go/prometheus/prometheus/protocolbuffers/go v1.31.0-20230726221845-41588ce133c8.1
 	cloud.google.com/go/spanner v1.47.0
 	github.com/IBM/pgxpoolprometheus v1.1.1
 	github.com/Masterminds/squirrel v1.5.4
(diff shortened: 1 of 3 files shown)


rxu-plaid:

jzelinskie:
This change looks good to me, but the commit message/PR title is misleading.

We use local buf plugins explicitly for the because remote plugins are built with different compiler versions at different times. This breaks reproducibility in our build process, so we keep all versions pinned with local plugins. See here: https://github.com/authzed/spicedb/blob/main/buf.gen.yaml

This change is changing one place where we don't care about versions (because the Prometheus remote write API is standardized and very much so unchanged), so we just import the module from the BSR.

Maybe we could reword it something like

"update prometheuspb module to new upstream domain"

rxu-plaid:
makes sense, I made some changes and took the suggestion for the title

jzelinskie:
Can you edit the commit message too? That's what I'm most concerned about as that'll live in the git history.

rxu-plaid:
edited the commit message

jzelinskie (approved):

PR #1463, opened 2023-07-27
Change how dispatches are counted in LookupResources

josephschorr:
Before the cursoring change, LR and RR issued *batch* operations, with the batch dispatch counting as only a single dispatch. Following the cursoring change, which changed all results to be streamed back individually, the dispatch was counted in *every* result, which caused the aggregation to be MUCH larger than correct in the LookupResources API handler.

This change adds tests to ensure that the dispatch count is in line with expectations when using batching, and fixes the above issue by only counting the dispatch once on the first result streamed back. If the first result (or any subsequently) are filtered, then the dispatch count is "held" until another result is ready to be published and the count can be appended to that result. This does mean that we might get *lower* counts, but never higher.

### DIFF
--- a/internal/graph/checkingresourcestream.go
+++ b/internal/graph/checkingresourcestream.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"sync"
 
+	"go.uber.org/atomic"
 	"golang.org/x/exp/maps"
 
 	"github.com/authzed/spicedb/internal/dispatch"
@@ -181,6 +182,14 @@ type checkingResourceStream struct {
 	// lastResourceCursor is the cursor from the last received reachable resource result. Should *only* be accessed from queue()
 	// and waitForPublishing() (after reachable resources has completed).
 	lastResourceCursor *v1.Cursor
+
+	// dispatchesToBeReported is the number of dispatches that were skipped from being reported due
+	// to a resource being filtered, and whose count has to be attached to the next outgoing result.
+	dispatchesToBeReported atomic.Uint32
+
+	// cachedDispatchesToBeReported is the number of cached dispatches that were skipped from being reported due
+	// to a resource being filtered, and whose count has to be attached to the next outgoing result.
+	cachedDispatchesToBeReported atomic.Uint32
 
 	errSetter sync.Once
 	err       error
@@ -423,7 +432,7 @@ func (crs *checkingResourceStream) runProcess(alwaysProcess bool) (bool, error) 
 	}
 
 	// Issue the bulk check over all the resources.
-	results, resultsMetas, err := computed.ComputeBulkCheck(
+	results, checkResultMetadata, err := computed.ComputeBulkCheck(
 		crs.ctx,
 		crs.checker,
 		computed.CheckParameters{
@@ -439,6 +448,9 @@ func (crs *checkingResourceStream) runProcess(alwaysProcess bool) (bool, error) 
 	if err != nil {
 		return true, err
 	}
+
+	crs.dispatchesToBeReported.Add(checkResultMetadata.DispatchCount)
+	crs.cachedDispatchesToBeReported.Add(checkResultMetadata.CachedDispatchCount)
 
 	for _, rai := range toCheck.Values() {
 		checkResult := results[rai.reachableResult.Resource.ResourceId]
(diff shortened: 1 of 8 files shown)


internal/graph/checkingresourcestream.go
```diff
@@ -182,6 +183,10 @@ type checkingResourceStream struct {
 	// and waitForPublishing() (after reachable resources has completed).
 	lastResourceCursor *v1.Cursor
 
+	// dispatchesToBeReported is the number of dispatches that were skipped from being reported due
+	// to a resource being filtered, and whose count has to be attached to the next outgoing result.
+	dispatchesToBeReported atomic.Uint32
+
```
jakedt, line 189: please track cached dispatches as well until we figure out the long term solution for them in this kind of stochastic reporting case

internal/graph/checkingresourcestream.go
```diff
@@ -466,7 +471,15 @@ func (crs *checkingResourceStream) runProcess(alwaysProcess bool) (bool, error)
 		// Set the lookupResult iff the permissionship was a valid permission.
 		var lookupResult *v1.DispatchLookupResourcesResponse
 		if permissionship != v1.ResolvedResource_UNKNOWN {
-			metadata := combineResponseMetadata(rai.reachableResult.Metadata, resultsMetas[rai.reachableResult.Resource.ResourceId])
+			// NOTE: since the metadata returned for the check will encompass the dispatches shared for all results, only
```
jakedt, line 474: The assumption here is that the bulk check will return the dispatches on every result object? Could we instead return a metadata object that applies to the entire result set?

internal/graph/graph.go
```diff
@@ -97,3 +97,14 @@ func addCallToResponseMetadata(metadata *v1.ResponseMeta) *v1.ResponseMeta {
 		DebugInfo:           metadata.DebugInfo,
 	}
 }
+
+func addAmortizedCallToResponseMetadata(metadata *v1.ResponseMeta) *v1.ResponseMeta {
```
jakedt, line 101: This function doesn't actually add any calls to anything, amortized or not. I think you were trying to explain why we're _not_ adding it (because it's amortized) but the function should probably just explain what it does, and then you can explain why you're calling it at the callsite.

jakedt (changes requested):

internal/graph/checkingresourcestream.go
josephschorr: Not if debugging is enabled, no

josephschorr: I changed to returning a single metadata, but only if debugging is disabled. Otherwise, chunking is disabled.

internal/graph/checkingresourcestream.go
josephschorr: Added

internal/graph/graph.go
josephschorr: Changed the name

internal/graph/computed/computecheck.go
```diff
@@ -92,7 +99,7 @@ func computeCheck(ctx context.Context,
 
 	// Ensure that the number of resources IDs given to each dispatch call is not in excess of the maximum.
 	results := make(map[string]*v1.ResourceCheckResult, len(resourceIDs))
-	resultMetadatas := make(map[string]*v1.ResponseMeta, len(resourceIDs))
+	metadata := &v1.ResponseMeta{}
 
 	// TODO(jschorr): Should we make this run in parallel via the preloadedTaskRunner?
```
vroldanbet, line 104: unrelated to the fix in this PR, but it would appear the chunking is dispatching serially and blocking on the response, so for a wide relation this can increase the latency of the API request and potentially underutilize available resources.

The preloaded task runner is getting refactored in https://github.com/authzed/spicedb/pull/1453, so it would be a good opportunity to follow up and refactor the chunk function. If so I'd suggest opening an issue to keep track of this.

internal/graph/checkingresourcestream.go
```diff
@@ -466,7 +478,10 @@ func (crs *checkingResourceStream) runProcess(alwaysProcess bool) (bool, error)
 		// Set the lookupResult iff the permissionship was a valid permission.
 		var lookupResult *v1.DispatchLookupResourcesResponse
 		if permissionship != v1.ResolvedResource_UNKNOWN {
-			metadata := combineResponseMetadata(rai.reachableResult.Metadata, resultsMetas[rai.reachableResult.Resource.ResourceId])
+			metadata := rai.reachableResult.Metadata
+			metadata = crs.addSkippedDispatchCountToBePublished(metadata)
```
vroldanbet, line 482: do we have concerns about counts being held distorting the next member to return?

jakedt: That was the compromise in order to get accurate counts overall when streaming.

internal/graph/checkingresourcestream.go
```diff
@@ -476,6 +491,14 @@ func (crs *checkingResourceStream) runProcess(alwaysProcess bool) (bool, error)
 				Metadata:            metadata,
 				AfterResponseCursor: rai.reachableResult.AfterResponseCursor,
 			}
+		} else {
+			if rai.reachableResult.Metadata.DispatchCount > 0 {
```
jakedt, line 495: Can we not just always add without the if? My understanding is that this will not take a mutex...

josephschorr: Since it takes a LOCK instruction, I'd still prefer to avoid that if possible

jakedt (approved):

PR #1470, opened 2023-08-01
*: drop legacy proto library

jzelinskie:
We need to test this to be 100% sure, but I'm fairly certain this legacy dependency can be dropped.


josephschorr (approved):
LGTM

I suspect this is safe given unit and integration test passing; if it breaks during testing, it is easy to revert

Issue #1477, opened 2023-08-03
Graceful shutdown not working


Posya:
Shutting down the server with ctrl-c does not always shut down the server.
Below is an example of a log when the GC routine continues to run after the shutdown event. Helps only to kill the process by kill -9

```
12:42PM DBG datastore garbage collection completed successfully collected={"namespaces":0,"relationships":0,"transactions":0} duration=3.559866 highestTxID=CAE= nowTime=2023-08-03T09:42:31Z
12:42PM DBG datastore garbage collection scheduled for next run next-run-in=180000
^C12:44PM INF received interrupt
12:44PM INF shutting down
12:44PM INF http server stopped serving addr=:9090 service=metrics
12:44PM INF http server stopped serving addr=:8080 service=dashboard
12:44PM INF grpc server stopped serving addr=:50051 network=tcp service=grpc

12:45PM DBG Query module=pgx pgx={"args":[],"commandTag":"SELECT 1","pid":495,"sql":"SELECT NOW()","time":0.169277}
12:45PM DBG Query module=pgx pgx={"args":["2023-08-02T09:45:31.466434Z"],"commandTag":"SELECT 1","pid":495,"sql":"SELECT xid, snapshot FROM relation_tuple_transaction WHERE timestamp \u003c $1 ORDER BY xid DESC LIMIT 1","time":0.826674}
12:45PM DBG Query module=pgx pgx={"args":[{"Uint64":1,"Valid":true}],"commandTag":"DELETE 0","pid":499,"sql":"WITH rows AS (SELECT namespace, object_id, relation, userset_namespace, userset_object_id, userset_relation, created_xid, deleted_xid FROM relation_tuple WHERE deleted_xid \u003c $1 LIMIT 1000)\n\t\t  DELETE FROM relation_tuple\n\t\t  WHERE (namespace, object_id, relation, userset_namespace, userset_object_id, userset_relation, created_xid, deleted_xid) IN (SELECT namespace, object_id, relation, userset_namespace, userset_object_id, userset_relation, created_xid, deleted_xid FROM rows);\n\t","time":1.435842}
[...]

PR #1479, opened 2023-08-03
Add integration tests for graceful shutdown of SpiceDB

josephschorr:


josephschorr:
Reference: https://github.com/authzed/spicedb/issues/1477

ecordell (commented):
LGTM

I'm not sure we need the binary for this test; it should suffice to build the command object and run it, and then cancel the parent context.

josephschorr:
> I'm not sure we need the binary for this test; it should suffice to build the command object and run it, and then cancel the parent context.

Strictly, no, but I like the idea of actually testing the real signaling system too

Issue #1468, opened 2023-08-01
Hard limits in proto files


Posya:
I made a scheme based on the scheme from Google IAM example, but at the same time, for our use, the number of relations in the role definitions was increased to 1500.
At the same time, the scheme itself has grown in size to 900 kb.
Is it possible to change the limit in the proto file, which now does not allow schemas above 256 kb? 2...4 Mb.
Could you check in your environment if SpiceDB will work fine with such large schemas?
Are there any restrictions on the size of the schema (other than the specified limits in the proto files?)
By the way, is it possible to also increase the length of the relation? Now it can be maximum 64 bytes.

Posya:
I checked for myself: increasing the limit by 10 times did not lead to any problems.

Posya:
I'm using latest postgres container

PR #1480, opened 2023-08-03
Change MySQL definition columns to LONGBLOB in prep for supporting larger definitions

josephschorr:
Part #1 of #1468

While each definition will not likely be super large, there is a definite need for larger individual definitions

### DIFF
--- a/internal/datastore/mysql/migrations/zz_migration.0006_convert_definitions_to_longblob.go
+++ b/internal/datastore/mysql/migrations/zz_migration.0006_convert_definitions_to_longblob.go
@@ -1,0 +1,26 @@
+package migrations
+
+import "fmt"
+
+func convertNamespaceDefinitionToLongBlob(t *tables) string {
+	return fmt.Sprintf(`ALTER TABLE %s
+		MODIFY serialized_config LONGBLOB NOT NULL;`,
+		t.Namespace(),
+	)
+}
+
+func convertCaveatDefinitionToLongBlob(t *tables) string {
+	return fmt.Sprintf(`ALTER TABLE %s
+		MODIFY definition LONGBLOB NOT NULL;`,
+		t.Caveat(),
+	)
+}
+
+func init() {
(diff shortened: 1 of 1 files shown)


internal/datastore/mysql/migrations/zz_migration.0006_convert_definitions_to_longblob.go
```diff
@@ -0,0 +1,26 @@
+package migrations
+
+import "fmt"
+
+func convertNamespaceDefinitionToLongBlob(t *tables) string {
+	return fmt.Sprintf(`ALTER TABLE %s
+		MODIFY serialized_config LONGBLOB NOT NULL;`,
```
ecordell, line 7: We sure we don't want MEDIUMBLOB here? 

BLOB = 64KB
MEDIUMBLOB=16MB
LONGBLOB=4GB

64k to 4g seems like a big jump

josephschorr: It is merely a max size, still constrained by the overall schema size limits. Further, the other types have overhead: https://www.tutorialspoint.com/What-is-the-maximum-length-of-data-we-can-put-in-a-BLOB-column-in-MySQL

ecordell: that doc is definitely wrong, longblobs [use a 4-byte prefix](https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html#:~:text=L%20%2B%204%20bytes%2C%20where%20L%20%3C%20232) (the only distinction between the types is the number of bytes used for the length prefix)

but the 1 byte difference is probably not a big deal, lgtm

ecordell (approved):

internal/datastore/mysql/migrations/zz_migration.0006_convert_definitions_to_longblob.go
josephschorr: Darn internet... maybe we use MEDIUMBLOB then?

PR #1531, opened 2023-09-08
Update gRPC and Go versions

josephschorr:


jzelinskie (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/authzed/spicedb
 
-go 1.20
```
vroldanbet, line 3: Why bumping the version, is the codebase using any 1.21 specific API?

PR #1537, opened 2023-09-18
crdb: better max retry errors (especially with retries disabled)

ecordell:


internal/datastore/crdb/pool/slqerrors.go
```diff
@@ -27,8 +27,12 @@ type MaxRetryError struct {
 }
 
 func (e *MaxRetryError) Error() string {
-	return strconv.Itoa(int(e.MaxRetries)) + "max retries reached" + ": " + e.LastErr.Error()
+	if e.MaxRetries == 0 {
+		return "retries disabled: " + e.LastErr.Error()
+	}
+	return strconv.Itoa(int(e.MaxRetries)) + " max retries reached" + ": " + e.LastErr.Error()
```
josephschorr, line 33: use `fmt.Sprintf`?

internal/datastore/crdb/pool/pool.go
```diff
@@ -309,7 +309,7 @@ func (p *RetryPool) withRetries(ctx context.Context, fn func(conn *pgxpool.Conn)
 		log.Ctx(ctx).Warn().Err(err).Uint8("retries", retries).Msg("error is not resettable or retryable")
 		return err
 	}
-	return &MaxRetryError{MaxRetries: p.maxRetries, LastErr: err}
+	return &MaxRetryError{MaxRetries: maxRetries, LastErr: err}
```
josephschorr, line 312: so why this change?

ecordell: https://github.com/authzed/spicedb/pull/1537/files#diff-1b011b67a73f910b9841c71da8f83c3c99da6101f476e4d02bfb3d3335eda548R261

this will set the correct maxretries if it's been set to something other than the default (0 in the case of disabled)

josephschorr (approved):
LGTM

PR #1540, opened 2023-09-19
Fix a panic in Postgres revision parsing for incompatible ZedTokens

### LINKED ISSUE #1539: Panic on giving ZedTokens from CRDB to Postgres
When giving a ZedToken that was created by CRDB-backed SpiceDB to a Postgres-backed SpiceDB, a panic can be raised if the ZedToken contains a timestamp that is too large for Postgres to handle

This should instead return an error.

Thanks to `Duncan` in Discord for reporting (https://discord.com/channels/844600078504951838/1153638230051270698)

josephschorr:
If Postgres is passed a ZedToken from CRDB, it will (incorrectly) interpret the token and try to create a slice with too much capacity. This change caps the delta on the transaction IDs and returns an error in that scenario

Fixes #1539

### DIFF
--- a/internal/datastore/postgres/revisions.go
+++ b/internal/datastore/postgres/revisions.go
@@ -169,6 +169,14 @@ func parseRevisionProto(revisionStr string) (datastore.Revision, error) {
 	}, nil
 }
 
+// MaxLegacyXIPDelta is the maximum allowed delta between the xmin and
+// xmax revisions IDs on a *legacy* revision stored as a revision decimal.
+// This is set to prevent a delta that is too large from blowing out the
+// memory usage of the allocated slice, or even causing a panic in the case
+// of a VERY large delta (which can be produced by, for example, a CRDB revision
+// being given to a Postgres datastore accidentally).
+const MaxLegacyXIPDelta = 1000
+
 // parseRevisionDecimal parses a deprecated decimal.Decimal encoding of the revision
 // with an optional xmin component, in the format of revision.xmin, e.g. 100.99.
 // Because we're encoding to a snapshot, we want the revision to be considered visible,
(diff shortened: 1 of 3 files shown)


DBBrowne:
Are we sure it's a size issue, rather than an `int` vs `float` issue?

in our case, the similarly sized `GhUKEzE2OTM1NDA5NDQ5NTk3MjA1OTI=` does not cause the pre-change panic that `GiAKHjE2OTM1NDA5NDAzNzMwNDU3MjcuMDAwMDAwMDAwMQ==` does.

jzelinskie (dismissed):

josephschorr:
Yeah, its a capacity failure on trying to create a slice to track the overlap revisions. The panic occurs on a `uint64`, so it cannot be a float issue

DBBrowne:
> Yeah, its a capacity failure on trying to create a slice to track the revisions. The panic occurs on a `uint64`, so it cannot be a float issue

The failing token `GiAKHjE2OTM1NDA5NDAzNzMwNDU3MjcuMDAwMDAwMDAwMQ==` decodes to `\x1a \n\x1e1693540940373045727.0000000001`

So if it's a size issue, we'd expect `\x1a \n\x1e1693540940373045728` (`GiAKHjE2OTM1NDA5NDAzNzMwNDU3Mjg=`) to throw as well?

josephschorr:
> So if it's a size issue, we'd expect `\x1a \n\x1e1693540940373045728` (`GiAKHjE2OTM1NDA5NDAzNzMwNDU3Mjg=`) to throw as well?

We would, except the lack of a decimal point means that it is interpreted as size `0`. This is because legacy Postgres revisions were stored as a decimal representing a *range*, where `X.Y` was a *range* from Y->X.

In the case of the first revision given, `1693540940373045727.0000000001` became `1693540940373045727 -> 1` for a size of `1693540940373045726`. In your second example, `1693540940373045728` became `1693540940373045728->1693540940373045728`, with a size of `0`.

This is all because legacy Postgres revisions were stored in this particular format, and giving a CRDB revision that "just so happened" to hit the legacy parsing code perfectly

jakedt (approved):

DBBrowne:
Ah, interesting on the postgres history, ty.

so here the second part of the decimal timestamp is correctly being parsed as an xmin candidate, and correctly assigned to xmin on L208, as if it was a postgres revision.

it's correct then to return the error at that point, not dismiss the float timestamp sooner.

thanks again!

PR #1549, opened 2023-09-27
runs Postgres tests with all 4 latest versions

vroldanbet:
The reasoning is that we are considering Postgres 15 for its better query planner index selection, so this PR adds testing for multiple Postgres versions. At some point, we will drop support for older Postgres versions, but for now, let's from the minimum supported version to the latest.

### DIFF
--- a/internal/datastore/postgres/postgres_test.go
+++ b/internal/datastore/postgres/postgres_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/authzed/spicedb/internal/datastore/common"
 	pgcommon "github.com/authzed/spicedb/internal/datastore/postgres/common"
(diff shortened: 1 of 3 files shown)


internal/datastore/postgres/postgres_test.go
```diff
@@ -40,17 +41,36 @@ func (pgd *pgDatastore) ExampleRetryableError() error {
...
+		trace.WithSampler(trace.AlwaysSample()),
+	)
+	otel.SetTracerProvider(testTraceProvider)
+}
+
 func TestPostgresDatastore(t *testing.T) {
+	t.Parallel()
+
 	for _, config := range []struct {
 		targetMigration string
```
ecordell, line 61: this is left around from the phased migration in 1.14, I believe. Can probably remove it now?

ecordell (approved):

PR #1550, opened 2023-09-28
Add new index to postgres datastore

vroldanbet:
This adds 1 new indexes to optimize:
- GC operations

⚠️ important! This index will not be selected by the query planner in PG 13 and 14 due to the usage of the `xid8` type 

## [add tuned GC index](https://github.com/authzed/spicedb/commit/6fb514fe065a64e7085fe2d3c5da6bb776f02595) 

the original GC index was rarely selected
when GC'ing relations with millions of relationships.
This caused a sequential scan that could load potentially
Gigabytes worth of data, and overload the database.

⚠️ Postgres 13 and 14 do not correctly select this new index
because the query planner struggles with the xid8 datatype.
Postgres 15 will be needed for this index to
actually work. For scenarios where relations aren't very wide,
older PG versions may have an acceptable query latency for GC,
but we strongly recommend moving to PG15.

This migration also deletes the old suboptimal index after
the new one is created.

### DIFF
--- a/internal/datastore/postgres/migrations/zz_migration.0016_add_tuned_gc_index.go
+++ b/internal/datastore/postgres/migrations/zz_migration.0016_add_tuned_gc_index.go
@@ -1,0 +1,34 @@
+package migrations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const createTunedGCIndex = `CREATE INDEX CONCURRENTLY 
+	IF NOT EXISTS ix_gc_index
+	ON relation_tuple (deleted_xid DESC)
+    WHERE deleted_xid < '9223372036854775807'::xid8;`
+
+const deleteSuboptimalGCIndex = `DROP INDEX CONCURRENTLY IF EXISTS ix_relation_tuple_by_deleted_xid`
+
+func init() {
+	if err := DatabaseMigrations.Register("add-tuned-gc-index", "add-gc-covering-index",
+		func(ctx context.Context, conn *pgx.Conn) error {
+			if _, err := conn.Exec(ctx, createTunedGCIndex); err != nil {
+				return fmt.Errorf("failed to create new tuned GC Index: %w", err)
+			}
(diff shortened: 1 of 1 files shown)


josephschorr:
@vroldanbet I wonder if we should have a migration to run `ANALYZE TABLE` after the concurrent indexes are added to ensure the stats are correct for them?

vroldanbet:
@josephschorr from [this post](https://stackoverflow.com/questions/37261697/how-to-efficiently-vacuum-analyze-tables-in-postgres):

>A table of 40 Million rows, the auto vacuum will run when the table will receive 8 Million updates or deletes. Similarly the table needs to receive 4 Million updates or deletes in order for the auto analyze to start. Mostly tables of such size would become slow before this threshold is received, due to which manually VACUUM FULL ANALYZE is recommend on a regular basis.

This [other post](https://dba.stackexchange.com/questions/241257/is-it-necessary-to-analyze-a-table-after-an-index-has-been-created) suggests that analyze would be necessary if index has an expression to it, which is the case here:

>If the index is just on simple columns like in your case, it is not necessary to ANALYZE the table after you create the index.
That is because statistics on the value distribution of the table columns are always collected, no matter if there is an index on the column or not.

>However, if you are indexing an expression like upper(some_column) or (CAST(some_column AS date)), you should run ANALYZE after creating the index.
PostgreSQL will then also collect statistics on the value distribution of the indexed expression. This happens automatically whenever autoanalyze runs, but it is a good idea to do it manually right after creating the index so you have good statistics right away.

My concerns here are:
- is running `ANALYZE` effective if the index is being created `CONCURRENTLY`?
- Could running `ANALYZE` to a PG running already a `CREATE INDEX` statement add sufficient load to disrupt service?

josephschorr:
> * is running `ANALYZE` effective if the index is being created `CONCURRENTLY`?

Unlikely, but I believe the migration will wait for the index to complete (concurrently) before the next one is run, so the ANALYZE should apply afterwards

> Could running ANALYZE to a PG running already a CREATE INDEX statement add sufficient load to disrupt service?

Yes, but one would imagine that CREATE INDEX is not also running unless someone is running the migration command twice (at the same time)

PR #1552, opened 2023-09-28
Improve experience on pg serialization errors

vroldanbet:
ℹ️ Depends on https://github.com/authzed/authzed-go/pull/147 and https://github.com/authzed/api/pull/88 merging first

I tried to improve two aspects of serialization errors:
- retrying them on the server side to spare the client doing it themselves
- do exponential backoff
- better logging and more concise logging
- better gRPC response (`Aborted` instead of `Unknown`)

This greatly improved the odds of transactions to succeed.

## Details

serialization errors are not well handled in the Postgres datastore:
- they tend to be logged by the `pgx` driver with potentially very large SQL statements
- they are not retried, nor there is a backoff. There are also other types of errors that are not being retried, like `conn closed` (which pgx considers safe to retry)
- they return to the client as unknown instead of the recommended aborted gRPC code
- no information is logged on retries in `debug` level
- logs can get extremely verbose due to large SQL and args

Serialization errors can also manifest as a transaction rollback, and so code is added to retry and wrap the error accordingly, in which case gRPC `aborted` is returned.

Additionally, retries after serialization can lead to longer requests, and longer requests can hit the
deadline. Sometimes a cancellation leads to the pgx connection being asynchronously closed, losing information about the original cause that closed it. Unfortunately, there isn't an exported type for this, so it's detected via the string. The error is turned into a multi-error that has cancellation. In a local load test, evidence shows that in all cases the original reason was cancellation.

A call to `pgx.SafeToRetry` is added to handle various pgx internal errors (like `conn closed`) and determine if they can be retried. This has helped retry errors it didn't use to retry before.

Retries also now have a backoff, since the default maximum of 10 can knock down the database. The retry function is refactored, and the minimum wait is reduced since the total backoff grows pretty quickly, and will easily
make the client hit the typical deadlines. The log message for retries is moved to Debug since serialization errors can be pretty common so it shouldn't be considered a warning.

I also detected that the backoff function wasn't respecting the context cancellation, so even when the client had canceled, the goroutine would be still asleep. The function was changed to cancel immediately after the client signal.

### DIFF
--- a/internal/datastore/common/errors.go
+++ b/internal/datastore/common/errors.go
@@ -24,7 +24,7 @@ func (err SerializationError) GRPCStatus() *status.Status {
 		err,
 		codes.Aborted,
 		spiceerrors.ForReason(
-			v1.ErrorReason_ERROR_REASON_WRITE_OR_DELETE_PRECONDITION_FAILURE,
+			v1.ErrorReason_ERROR_REASON_SERIALIZATION_FAILURE,
 			map[string]string{},
 		),
 	)
--- a/internal/datastore/common/errors.go
+++ b/internal/datastore/common/errors.go
@@ -13,6 +13,32 @@ import (
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
+// SerializationError is returned when there's been a serialization
+// error while performing a datastore operation
+type SerializationError struct {
+	error
+}
+
+func (err SerializationError) GRPCStatus() *status.Status {
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.Aborted,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_WRITE_OR_DELETE_PRECONDITION_FAILURE,
+			map[string]string{},
+		),
+	)
+}
+
+func (err SerializationError) Unwrap() error {
+	return err.error
+}
+
+// NewSerializationError creates a new SerializationError
+func NewSerializationError(err error) error {
+	return SerializationError{err}
+}
+
 // CreateRelationshipExistsError is an error returned when attempting to CREATE an already-existing
 // relationship.
 type CreateRelationshipExistsError struct {
--- a/internal/datastore/postgres/common/pgx.go
+++ b/internal/datastore/postgres/common/pgx.go
@@ -107,6 +107,9 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
 	addTracer(connConfig, &tracelog.TraceLog{Logger: levelMappingFn(l), LogLevel: tracelog.LogLevelInfo})
 }
 
+// truncateLargeSQL takes arguments of a SQL statement provided via pgx's tracelog.LoggerFunc and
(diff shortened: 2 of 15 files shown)


internal/datastore/common/errors.go
```diff
@@ -13,6 +13,32 @@ import (
...
+type SerializationError struct {
+	error
+}
+
+func (err SerializationError) GRPCStatus() *status.Status {
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.Aborted,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_WRITE_OR_DELETE_PRECONDITION_FAILURE,
```
josephschorr, line 27: we should probably define a new API error code for this and change it here

internal/datastore/postgres/common/pgx.go
```diff
@@ -109,6 +107,52 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
 	addTracer(connConfig, &tracelog.TraceLog{Logger: levelMappingFn(l), LogLevel: tracelog.LogLevelInfo})
 }
 
+func truncateLargeSQL(data map[string]any) {
```
josephschorr, line 110: Add a comment that this does so in place

internal/datastore/postgres/common/pgx.go
```diff
@@ -109,6 +107,52 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
...
+	}
+	if argsData, ok := data["args"]; ok {
+		argsSlice, ok := argsData.([]any)
+		if ok && len(argsSlice) > maxSQLArgsLen {
+			data["args"] = argsSlice[:maxSQLArgsLen]
+		}
+	}
+}
+
+func IsCancellationError(err error) bool {
```
josephschorr, line 130: doc comments on exported members

internal/datastore/common/errors.go
vroldanbet: Sounds good 👍🏻

vroldanbet: Opened https://github.com/authzed/api/pull/88

internal/datastore/postgres/common/pgx.go
```diff
@@ -109,6 +107,56 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
 	addTracer(connConfig, &tracelog.TraceLog{Logger: levelMappingFn(l), LogLevel: tracelog.LogLevelInfo})
 }
 
+// truncateLargeSQL takes arguments of a SQL statement provided via pgx's tracelog.LoggerFunc and
+// truncates SQL statements and SQL arguments to placeholders that exceed a certain length. This helps
```
josephschorr, line 111: truncateLargeSQL takes arguments of a SQL statement provided via pgx's tracelog.LoggerFunc and
replaces SQL statements and SQL arguments with placeholders when the statements and/or arguments exceed a certain length

josephschorr (approved):
LGTM

PR #1553, opened 2023-09-30
Order onr comparisons by cardinality for short-circuiting

bradengroom:
This is a small optimization that carries forward the idea set in [onrEqual](https://github.com/authzed/spicedb/blob/846f79e3b9fb86c04a59037ab25a0280e0f7090a/internal/graph/check.go#L202). The comparisons are swapped so that the higher cardinality check happens first to allow for faster short-circuiting.


bradengroom:

bradengroom:
recheck

josephschorr (approved):
LGTM

PR #1562, opened 2023-10-03
Compare tuples without stringifying

bradengroom:
This updates the mysql datastore to compare tuples without stringifying.

### DIFF
--- a/pkg/tuple/onr.go
+++ b/pkg/tuple/onr.go
@@ -108,3 +108,12 @@ func StringsONRs(onrs []*core.ObjectAndRelation) []string {
 	sort.Strings(onrstrings)
 	return onrstrings
 }
+
+func OnrEqual(lhs, rhs *core.ObjectAndRelation) bool {
+	// Properties are sorted by highest to lowest cardinality to optimize for short-circuiting.
+	return lhs.ObjectId == rhs.ObjectId && lhs.Relation == rhs.Relation && lhs.Namespace == rhs.Namespace
+}
+
+func OnrEqualOrWildcard(tpl, target *core.ObjectAndRelation) bool {
+	return OnrEqual(tpl, target) || (tpl.ObjectId == PublicWildcard && tpl.Namespace == target.Namespace)
+}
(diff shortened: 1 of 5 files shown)


pkg/tuple/tuple.go
```diff
@@ -263,6 +271,11 @@ func Delete(tpl *core.RelationTuple) *core.RelationTupleUpdate {
 	}
 }
 
+func Equal(lhs, rhs *core.RelationTuple) bool {
+	// TODO(jschorr): Use a faster method then string comparison for caveats.
+	return OnrEqual(lhs.ResourceAndRelation, rhs.ResourceAndRelation) && OnrEqual(lhs.Subject, rhs.Subject) && MustStringCaveat(lhs.Caveat) == MustStringCaveat(rhs.Caveat)
```
bradengroom, line 276: We still stringify the caveat if everything else matches. There might be some optimizations that could be made here. For example, we could count the keys before stringifying. I'm not sure if this is worth doing.

vroldanbet: There is a function in the codebase that computes a hash of the caveat context. I'm not sure that'd be faster than stringifying tho.

pkg/tuple/onr.go
```diff
@@ -108,3 +108,12 @@ func StringsONRs(onrs []*core.ObjectAndRelation) []string {
 	sort.Strings(onrstrings)
 	return onrstrings
 }
+
+func OnrEqual(lhs, rhs *core.ObjectAndRelation) bool {
+	// Properties are sorted by highest to lowest cardinality to optimize for short-circuiting.
+	return lhs.ObjectId == rhs.ObjectId && lhs.Relation == rhs.Relation && lhs.Namespace == rhs.Namespace
+}
+
+func OnrEqualOrWildcard(tpl, target *core.ObjectAndRelation) bool {
```
vroldanbet, line 117: do we have tests for this?

vroldanbet (commented):
Thanks for your contribution! ✨ Changes look good to me, added some minor feedback

pkg/tuple/onr.go
bradengroom: Nope! They aren't new really. I just relocated them from `internal/graph/check.go`. I can write tests though

bradengroom:
Added tests for the equality functions.

The caveat hashing function I found looked like it accepts a different type (protobuf struct?) than I have. I wasn’t sure how strongly you felt about using the hashing function, but I’m happy to look into it more if you want.

vroldanbet (approved):
This looks good to me, thanks for contributing back to the project! 🎉

PR #1564, opened 2023-10-03
Change BulkCheckPermission response to match order of the request

### LINKED ISSUE #1563: `BulkCheckPermission` does not return results in the same order as requests

linux, macos, windows

amd64, arm64

### Steps to Reproduce

I have noticed that in some cases, `BulkCheckPermission` returns the results (permissions) in a different order than requests.

### Expected Result

I think the desired behavior would be to return results in the same order as requests

### Actual Result

-

josephschorr:
Fixes #1563

### DIFF
--- a/internal/services/v1/experimental.go
+++ b/internal/services/v1/experimental.go
@@ -34,11 +34,13 @@ import (
 	"github.com/authzed/spicedb/pkg/cursor"
 	"github.com/authzed/spicedb/pkg/datastore"
 	dsoptions "github.com/authzed/spicedb/pkg/datastore/options"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	"github.com/authzed/spicedb/pkg/genutil/slicez"
 	"github.com/authzed/spicedb/pkg/middleware/consistency"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	dispatchv1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
 	implv1 "github.com/authzed/spicedb/pkg/proto/impl/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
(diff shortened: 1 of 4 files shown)


internal/services/v1/experimental.go
```diff
@@ -520,11 +533,30 @@ func (es *experimentalServer) BulkCheckPermission(ctx context.Context, req *v1.B
...
+
+		pairHashes[pair] = pairItemHash
+	}
+
+	// Resort the results based on the incoming request order to make it easier to consume. This is not strictly
+	// necessary as each result contains the request that made it, but for the non-streaming version
+	// of this API, it makes sense.
+	slices.SortStableFunc(pairs, func(first *v1.BulkCheckPermissionPair, second *v1.BulkCheckPermissionPair) int {
+		return itemIndexByHash[pairHashes[first]] - itemIndexByHash[pairHashes[second]]
+	})
```
bradengroom, line 557: I assume this is going to be an O(n log n) sort, right? You could do the sorting in linear time by re-using your map from above.

```
sortedPairs := make([]*v1.BulkCheckPermissionPair, 0, len(req.Items))
for _, pair := range pairs {
    currHash := computeBulkCheckPermissionItemHash(pair.Request)
    pairIdx := itemIndexByHash[currHash]
    sortedPairs[pairIdx] = pair
}
```

bradengroom: Or if you don't want the added memory of another `sortedPairs` list, you could compute the hash and insert into `pairs` at the proper index in the first loop.

bradengroom (changes requested):
Thanks for doing this. I think you can actually support it without changing the runtime complexity of the endpoint 🙂

internal/services/v1/experimental.go
josephschorr: Updated. Thanks for the suggestion

bradengroom (approved):
I can't speak to the failing tests, but I'm approving because my feedback was addressed. Thanks!

internal/services/v1/experimental.go
```diff
@@ -520,11 +535,32 @@ func (es *experimentalServer) BulkCheckPermission(ctx context.Context, req *v1.B
...
+	// Run the checks in parallel.
 	if err := tr.StartAndWait(); err != nil {
 		return nil, es.rewriteError(ctx, err)
 	}
 
-	return resp, nil
+	// Resort the results based on the incoming request order to make it easier to consume. This is not strictly
+	// necessary as each result contains the request that made it, but for the non-streaming version
+	// of this API, it makes sense.
+	sortedPairs := make([]*v1.BulkCheckPermissionPair, len(req.Items))
```
vroldanbet, line 546: super nit but perhaps we could order the `pairs` slice in-place so we don't allocate another one?

vroldanbet (approved):

internal/services/v1/experimental.go
josephschorr: Done

vroldanbet (approved):

PR #1565, opened 2023-10-04
Use lo.Uniq function

bradengroom:
This removes the custom `slicez.Unique` function in favor of `lo.Uniq` function which does the exact same thing. Why own this code if you don't have to? 🙂

I wanted to make another change, but that got out of hand, so I left thoughts below.


internal/graph/check.go
```diff
@@ -155,7 +156,7 @@ func (cc *ConcurrentChecker) checkInternal(ctx context.Context, req ValidatedChe
 	}
 
 	// Deduplicate any incoming resource IDs.
-	resourceIds := slicez.Unique(req.ResourceIds)
+	resourceIds := lo.Uniq(req.ResourceIds)
```
bradengroom, line 159: I really wanted to swap this to use `mapz.NewSet[string]`. That would allow loop in this function to be removed
https://github.com/authzed/spicedb/blob/fe18f6700b1104f3f76807128cc0bb12508c2654/internal/graph/check.go#L534

```
	if resourceIds.Has(subject.ObjectId) {
		membershipSet := NewMembershipSet()
		membershipSet.AddDirectMember(subject.ObjectId, nil)
		resourceIds.Remove(subject.ObjectId)
		return membershipSet, resourceIds
	}
```

This blew up and I ended up updating a ton of stuff to use the set structure. This may or may not be a _bad_ idea, but it was getting out of hand enough that I thought I'd ask for thoughts from other before attempting that change. It was putting me down the path of updating [RelationshipsFilter](https://github.com/authzed/spicedb/blob/fe18f6700b1104f3f76807128cc0bb12508c2654/pkg/datastore/datastore.go#L60) to be a `mapz.Set`. This falls outside of the `internal/` package, so I'm not sure if that interface is something that can safely be changed.

bradengroom: Meh. Probably not worth it as I think about it.

josephschorr: If it reduces memory or CPU, then it might be worth it. You'd have to benchmark though

jzelinskie: I think it's more understandable API contract to return a set if you're guaranteeing uniqueness.

We should follow the execution path, if there are lots of appends, it's probably less efficient.

bradengroom: Alright thanks. I'll play with this idea in a separate PR.

jzelinskie:
Looks like this PR needs a `mage deps:tidy` and then it LGTM

bradengroom:
> Looks like this PR needs a `mage deps:tidy` and then it LGTM

Thanks! `mage` is new to me, so it would've taken me a while to find that.

josephschorr:
@bradengroom Please rebase your changes down to a single commit and then looks good

jzelinskie (approved):

Issue #1543, opened 2023-09-20
MySql some queries missing an index


nulian:
Not sure what call in spicedb does it but it generated a query like below which is fairly slow because it came up in our slow query log because created_transaction is missing an index in the MySQL version of it.

```
SELECT namespace, object_id, relation, userset_namespace, userset_object_id, userset_relation, caveat_name, caveat_context, created_transaction, deleted_transaction FROM relation_tuple WHERE ((created_transaction > 158582 AND created_transaction <= 158583) OR (deleted_transaction > 158582 AND deleted_transaction <= 158583));
```

### Expected Result

All queried columns having an index.

### Actual Result

Fast lookup query

josephschorr:
@nulian Be curious to see an `EXPLAIN` output from your database, just so we can compare to ours

nulian:
Not precisly the same database but is a based on a dump from it but this is the explain
Without index it takes somewhere between 500ms and 5s
With index it's 10ms - 150ms

```
|id|select_type|table|type|possible_keys|key|key_len|ref|rows|Extra|
|--|-----------|-----|----|-------------|---|-------|---|----|-----|
|1|SIMPLE|relation_tuple|ALL|ix_relation_tuple_by_deleted_transaction||||224739|Using where|

```

vroldanbet:
@nulian are you using the `Watch` API? Would you mind sharing the index you added?

This access pattern seems to be used by the MySQL datastore Watch API implementation:

https://github.com/authzed/spicedb/blob/8f6136b6f45b155caa665fd4b47bf492ae95acc7/internal/datastore/mysql/watch.go#L94-L103

this is the explain I got initially with 84K rows:

| EXPLAIN                                                                                                                                                                                                                                                                                                                                                                                          |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| -> Filter: (((relation_tuple.created_transaction > 158582) and (relation_tuple.created_transaction <= 158583)) or ((relation_tuple.deleted_transaction > 158582) and (relation_tuple.deleted_transaction <= 158583)))  (cost=7920 rows=16334) (actual time=71.3..71.3 rows=0 loops=1)
    -> Table scan on relation_tuple  (cost=7920 rows=77839) (actual time=0.0697..58.1 rows=84849 loops=1)
 |

Then I added the following index, the rows scanned where lower:

```sql
[...]

PR #1566, opened 2023-10-06
introduces an index to speed up Watch API calls

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/1543

See more details at https://github.com/authzed/spicedb/issues/1543#issuecomment-1744739489

This PR adds a new index to avoid sequential scans caused by the access patterns used in Watch API implementation

### DIFF
--- a/internal/datastore/mysql/migrations/zz_migration.0007_watch_api_rel_tuple_index.go
+++ b/internal/datastore/mysql/migrations/zz_migration.0007_watch_api_rel_tuple_index.go
@@ -1,0 +1,17 @@
+package migrations
+
+import "fmt"
+
+func addWatchAPIIndex(t *tables) string {
+	return fmt.Sprintf(`CREATE INDEX ix_relation_tuple_watch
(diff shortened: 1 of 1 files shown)


internal/datastore/mysql/migrations/zz_migration.0007_watch_api_index.go
```diff
@@ -0,0 +1,17 @@
+package migrations
+
+import "fmt"
+
+func addWatchAPIIndex(t *tables) string {
+	return fmt.Sprintf(`CREATE INDEX ix_watch
```
josephschorr, line 6: `ix_caveat_watch`?

vroldanbet: oh boy, good you found that - that's the wrong table! 🙈 
See https://github.com/authzed/spicedb/issues/1543#issuecomment-1744739489

josephschorr (approved):
LGTM

PR #1568, opened 2023-10-07
Update groupItems to use less memory

bradengroom:
The current implementation of [groupItems](https://github.com/authzed/spicedb/blob/ab6946b836229c2ce61bc0876844ed5849d1eca4/internal/services/v1/grouping.go#L29) calls `make` twice to allocate memory for building its results. This modifies the function to build its results in one pass and removes the need for the second `make` call. This should use less memory overall.

### DIFF
--- a/internal/services/v1/grouping.go
+++ b/internal/services/v1/grouping.go
@@ -7,14 +7,13 @@ import (
 
 	"github.com/authzed/spicedb/internal/graph/computed"
 	"github.com/authzed/spicedb/pkg/datastore"
-	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 )
 
 var MaxBulkCheckDispatchChunkSize = datastore.FilterMaximumIDCount
 
 type groupedCheckParameters struct {
-	params      computed.CheckParameters
+	params      *computed.CheckParameters
 	resourceIDs []string
 }
 
@@ -26,9 +25,8 @@ type groupingParameters struct {
 
 // groupItems takes a slice of BulkCheckPermissionRequestItem and groups them based
 // on using the same permission, subject type, subject id, and caveat.
-func groupItems(ctx context.Context, params groupingParameters, items []*v1.BulkCheckPermissionRequestItem) ([]groupedCheckParameters, error) {
-	groups := mapz.NewMultiMap[string, string]()
-	groupParams := make(map[string]computed.CheckParameters, len(items))
+func groupItems(ctx context.Context, params groupingParameters, items []*v1.BulkCheckPermissionRequestItem) (map[string]*groupedCheckParameters, error) {
+	res := make(map[string]*groupedCheckParameters)
 
 	for _, item := range items {
 		hash, err := computeBulkCheckPermissionItemHashWithoutResourceID(item)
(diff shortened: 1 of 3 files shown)


internal/services/v1/grouping.go
```diff
@@ -26,50 +25,51 @@ type groupingParameters struct {
...
-		grouped = append(grouped, groupedCheckParameters{
-			params:      groupParams[key],
-			resourceIDs: resourceIDs,
-		})
-	}
+	return res, nil
+}
 
-	return grouped, nil
+func checkParametersFromBulkCheckPermissionRequestItem(
```
bradengroom, line 55: I pulled this out just so that it was easier for me to focus on the algorithm piece. I'm happy to fold this back into `groupItems` now if that's preferred.

internal/services/v1/grouping.go
```diff
@@ -26,50 +25,51 @@ type groupingParameters struct {
 
 // groupItems takes a slice of BulkCheckPermissionRequestItem and groups them based
 // on using the same permission, subject type, subject id, and caveat.
-func groupItems(ctx context.Context, params groupingParameters, items []*v1.BulkCheckPermissionRequestItem) ([]groupedCheckParameters, error) {
-	groups := mapz.NewMultiMap[string, string]()
-	groupParams := make(map[string]computed.CheckParameters, len(items))
+func groupItems(ctx context.Context, params groupingParameters, items []*v1.BulkCheckPermissionRequestItem) (map[string]*groupedCheckParameters, error) {
```
bradengroom, line 28: I did change the signature here to return a map of `hash` -> `groupedCheckParameters`. Otherwise I would need to convert the dictionary back to a slice in the end with `maps.Values` which I think would cancel out the memory savings.

vroldanbet (commented):
Thanks for the contribution! Would you mind validating the improvements with a go `Benchmark` ?

bradengroom:
@vroldanbet 

Benchmark results on the current logic:
```
go test ./internal/services/v1/...  -bench . -benchmem
goos: darwin
goarch: arm64
pkg: github.com/authzed/spicedb/internal/services/v1
BenchmarkGrouping1Checks-10        	 1000000	      1046 ns/op	     592 B/op	      18 allocs/op
BenchmarkGrouping100Checks-10      	   12346	     97855 ns/op	   59391 B/op	    1375 allocs/op
BenchmarkGrouping10000Checks-10    	     127	   9190960 ns/op	 4793962 B/op	  131012 allocs/op
PASS
```

Benchmark results on the new logic:
```
go test ./internal/services/v1/...  -bench . -benchmem
goos: darwin
goarch: arm64
pkg: github.com/authzed/spicedb/internal/services/v1
BenchmarkGrouping1Checks-10        	 1000000	      1060 ns/op	     848 B/op	      20 allocs/op
BenchmarkGrouping100Checks-10      	   12512	     94771 ns/op	   38756 B/op	    1393 allocs/op
BenchmarkGrouping10000Checks-10    	     132	   8926897 ns/op	 3491259 B/op	  131209 allocs/op
PASS
```

So there's ~250 extra bytes for a single request, but it scales better with more requests (who will be calling the bulk endpoint with a single request?). About 20kb savings when there's 100 checks. About 1.3mb savings when there are 10k checks.

Here is the benchmark code (I realize it could be cleaned up a little, but I didn't worry about it since we discussed not merging it):
```go

func getRequests(n int) []*v1.BulkCheckPermissionRequestItem {
	requests := []string{}
	for subjectIdx := 0; subjectIdx < n; subjectIdx++ {
		for resourceIdx := 0; resourceIdx < n; resourceIdx++ {
			requests = append(requests, fmt.Sprintf("document:%d#view@user:%d", resourceIdx, subjectIdx))
		}
	}
	return lo.Map(requests, func(request string, _ int) *v1.BulkCheckPermissionRequestItem {
		rel := tuple.ParseRel(request)
		return &v1.BulkCheckPermissionRequestItem{
			Resource:   rel.Resource,
			Permission: rel.Relation,
			Subject:    rel.Subject,
		}
	})
}

func BenchmarkGrouping1Checks(b *testing.B) {
	cp := groupingParameters{
		atRevision:           datastore.NoRevision,
		maxCaveatContextSize: math.MaxInt,
		maximumAPIDepth:      1,
	}
	checks := getRequests(1)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		groupItems(context.Background(), cp, checks)
	}
}

func BenchmarkGrouping100Checks(b *testing.B) {
	cp := groupingParameters{
		atRevision:           datastore.NoRevision,
		maxCaveatContextSize: math.MaxInt,
		maximumAPIDepth:      1,
	}
	checks := getRequests(10)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		groupItems(context.Background(), cp, checks)
	}
}

func BenchmarkGrouping10000Checks(b *testing.B) {
	cp := groupingParameters{
		atRevision:           datastore.NoRevision,
		maxCaveatContextSize: math.MaxInt,
		maximumAPIDepth:      1,
	}
	checks := getRequests(100)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		groupItems(context.Background(), cp, checks)
	}
}

```

vroldanbet (approved):
Thanks for the benchmark, looks good to me!

PR #1571, opened 2023-10-08
Enable the experimental apis in the http gateway

thewunder:
It looks like the experimental apis are not usable in http, and it seems they were never included in the gateway handlers.


thewunder:
recheck

thewunder:

vroldanbet (approved):
Good catch, LGTM!

PR #1584, opened 2023-10-16
spanner: split read/write, add connpool config

jzelinskie:
This PR connects up the SpiceDB datastore flags for connection pooling with the underlying Spanner client's gRPC connection pooling.

This also enables gzip compression for the underlying Spanner gRPC connections.

Possible future work:
- Support the min read/write open connection flags -- this end up just being `grpcConns = max(config.min, config.max)` because we don't have access to set a minimum
- Use optgen to generate the spanner datastore config

### DIFF
--- a/internal/datastore/spanner/spanner.go
+++ b/internal/datastore/spanner/spanner.go
@@ -65,9 +65,10 @@ type spannerDatastore struct {
 	*revisions.RemoteClockRevisions
 	revision.DecimalDecoder
 
-	client   *spanner.Client
-	config   spannerOptions
-	database string
+	readClient  *spanner.Client
+	writeClient *spanner.Client
+	config      spannerOptions
+	database    string
 }
 
 // NewSpannerDatastore returns a datastore backed by cloud spanner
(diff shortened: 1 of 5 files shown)


vroldanbet:
@jzelinskie dep vuln failures are being addressed in https://github.com/authzed/spicedb/pull/1583

internal/datastore/spanner/spanner.go
```diff
@@ -203,7 +218,7 @@ func (sd spannerDatastore) Features(_ context.Context) (*datastore.Features, err
 }
 
 func (sd spannerDatastore) Close() error {
-	sd.client.Close()
+	sd.readClient.Close()
```
vroldanbet, line 221: shouldn't it also close the `writeClient`?

```suggestion
	sd.readClient.Close()
	sd.writeClient.Close()	
```

pkg/cmd/datastore/datastore.go
```diff
@@ -418,6 +418,8 @@ func newSpannerDatastore(opts Config) (datastore.Datastore, error) {
 		spanner.WatchBufferLength(opts.WatchBufferLength),
 		spanner.EmulatorHost(opts.SpannerEmulatorHost),
 		spanner.DisableStats(opts.DisableStats),
+		spanner.ReadConnsMaxOpen(opts.ReadConnPool.MaxOpenConns),
+		spanner.WriteConnsMaxOpen(opts.WriteConnPool.MaxOpenConns),
```
vroldanbet, line 422: The original default was `1`, see [here](https://github.com/googleapis/google-api-go-client/blob/25093c222db5ff0fec8d97782bd65d2661f39f9c/transport/grpc/dial.go#L92-L99). 

Now we are opening 20 connections by default. If multiplied by N SpiceDB replicas and 2 pools (before 1), this could cause problems for the Spanner instance. I'd suggest setting the maximum to the original behaviour of 1, and only move it up once we understand the implications.

vroldanbet (changes requested):
👍🏻 to the changes, I'm proposing to be more conservative with the new default.

ecordell (changes requested):
There's a couple of things in this PR that I'm not sure about.

First, each spanner client already maintains a pool of connections and a pool of sessions (4 connections, 100-400 session by default). The default session max is 100*the max connections, so this PR is making a huge change in the defaults (30 connections, 100-30000 sessions).

The spanner client also used to have a split between read and write sessions, but removed it:

https://github.com/googleapis/google-cloud-go/blob/main/spanner/session.go#L408-L415
https://github.com/googleapis/google-cloud-go/pull/7149

so I think it's worth digging into why that happened before we implement our own split on top

PR #1586, opened 2023-10-17
Configurable Spanner connection pool size, enable gzip

ecordell:
Builds on #1584, but removes the split pools for now so that we can test these other factors in isolation.


vroldanbet (approved):
LGTM

PR #1588, opened 2023-10-17
Add schema watch support in Spanner datastore driver

josephschorr:


internal/datastore/spanner/watch.go
```diff
@@ -19,8 +23,21 @@ import (
 
 const (
 	RelationTupleChangeStreamName = "relation_tuple_stream"
+	SchemaChangeStreamName        = "schema_change_stream"
 )
 
+var retryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
```
jzelinskie, line 29: Why not have a single metric if both crdb and cockroach are using this?

josephschorr: We could, but then it has to be defined in a central place, and that doesn't currently impact the other datastores

internal/datastore/spanner/watch.go
```diff
@@ -164,3 +181,175 @@ func contextualizedCaveatFromValues(values map[string]any) (*core.Contextualized
...
+	updates := make(chan *datastore.SchemaState, 10)
+	errs := make(chan error, 1)
+
+	lastRevision := afterRevision
+	retryCount := -1
+	go (func() {
+		running := true
+		for running {
+			retryCount++
+			retryHistogram.Observe(float64(retryCount))
```
vroldanbet, line 197: Why not a gauge or a count(1)? This number is basically monotonically increasing over time, what kind of graph were you planning to plot?

internal/datastore/spanner/watch.go
```diff
@@ -164,3 +181,175 @@ func contextualizedCaveatFromValues(values map[string]any) (*core.Contextualized
...
+	afterRevision := afterRevisionRaw.(revision.Decimal)
+	reader, err := changestreams.NewReaderWithConfig(
+		ctx,
+		project,
+		instance,
+		database,
+		SchemaChangeStreamName,
+		changestreams.Config{
+			StartTimestamp:    timestampFromRevision(afterRevision),
+			HeartbeatInterval: 100 * time.Millisecond, // 100ms is the minimum allowed
```
vroldanbet, line 244: N spicedb replicas running 10 heartbeats per second could be a bunch of heartbeats. Do we want to be more conservative, or perhaps expose this in a parameter now?

internal/datastore/spanner/watch.go
```diff
@@ -164,3 +181,175 @@ func contextualizedCaveatFromValues(values map[string]any) (*core.Contextualized
...
+
+func (sd spannerDatastore) watchSchemaWithoutRetry(ctx context.Context, afterRevisionRaw datastore.Revision, processUpdate func(update *datastore.SchemaState, err error)) {
+	project, instance, database, err := parseDatabaseName(sd.database)
+	if err != nil {
+		processUpdate(nil, err)
+		return
+	}
+
+	afterRevision := afterRevisionRaw.(revision.Decimal)
+	reader, err := changestreams.NewReaderWithConfig(
```
vroldanbet, line 236: is there a priority associated with change streams? should we put it by default to low priority so it does not disrupt the critical path?

internal/datastore/spanner/watch.go
```diff
@@ -164,3 +181,175 @@ func contextualizedCaveatFromValues(values map[string]any) (*core.Contextualized
 	}
 	return nil, nil
 }
+
+const maxSchemaWatchRetryCount = 15
+
+func (sd spannerDatastore) WatchSchema(ctx context.Context, afterRevision datastore.Revision) (<-chan *datastore.SchemaState, <-chan error) {
```
vroldanbet, line 187: can we add mocked tests at least if the emulator does not properly do changestreams?

internal/datastore/spanner/watch.go
```diff
@@ -164,3 +181,175 @@ func contextualizedCaveatFromValues(values map[string]any) (*core.Contextualized
...
+
+	err = reader.Read(ctx, func(result *changestreams.ReadResult) error {
+		// See: https://cloud.google.com/spanner/docs/change-streams/details
+		for _, record := range result.ChangeRecords {
+			for _, dcr := range record.DataChangeRecords {
+				changeRevision := revisionFromTimestamp(dcr.CommitTimestamp)
+				modType := dcr.ModType // options are INSERT, UPDATE, DELETE
+
+				for _, mod := range dcr.Mods {
+					primaryKeyColumnValues := mod.Keys.Value.(map[string]any)
```
vroldanbet, line 263: could this type assertion panic? maybe do it the safe way with `ok`?

vroldanbet (commented):
This looks good to me, I left a bunch of questions. I think we need to add some basic unit tests, even if it means mocking

internal/datastore/spanner/watch.go
```diff
@@ -164,3 +181,175 @@ func contextualizedCaveatFromValues(values map[string]any) (*core.Contextualized
...
+
+const maxSchemaWatchRetryCount = 15
+
+func (sd spannerDatastore) WatchSchema(ctx context.Context, afterRevision datastore.Revision) (<-chan *datastore.SchemaState, <-chan error) {
+	updates := make(chan *datastore.SchemaState, 10)
+	errs := make(chan error, 1)
+
+	lastRevision := afterRevision
+	retryCount := -1
+	go (func() {
```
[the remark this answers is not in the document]
josephschorr: I personally find it more readable but removed

internal/datastore/spanner/watch.go
josephschorr: Added a low priority

internal/datastore/spanner/watch.go
josephschorr: As discussed, will wait for the emulator's support to catch up

josephschorr:
Updated

pkg/cmd/datastore/datastore.go
```diff
@@ -210,6 +213,8 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	flagSet.StringVar(&opts.MigrationPhase, flagName("datastore-migration-phase"), "", "datastore-specific flag that should be used to signal to a datastore which phase of a multi-step migration it is in")
 	flagSet.Uint16Var(&opts.WatchBufferLength, flagName("datastore-watch-buffer-length"), 1024, "how many events the watch buffer should queue before forcefully disconnecting reader")
 
+	flagSet.DurationVar(&opts.SchemaWatchHeartbeat, flagName("datastore-schema-watch-heartbeat"), 0*time.Millisecond, "heartbeat time on the schema watch in the datastore (if supported). 0 means to default to the datastore's minium.")
```
vroldanbet, line 216:
```suggestion
	flagSet.DurationVar(&opts.SchemaWatchHeartbeat, flagName("datastore-schema-watch-heartbeat"), 0*time.Millisecond, "heartbeat time on the schema watch in the datastore (if supported). 0 means to default to the datastore's minimum.")
```

vroldanbet (approved):

pkg/cmd/datastore/datastore.go
josephschorr: Fixed

internal/datastore/spanner/watch.go
jzelinskie: Defining it in a central places seems pretty minor compared to users having to dashboard two different metrics.

PR #1590, opened 2023-10-18
Optimistic locking on Spanner read/write transactions

jzelinskie:
From my observations, SpiceDB has very little contention on rows in the Spanner datastore, so it's possible there are performance wins by defaulting to optimistic concurrency.


vroldanbet (approved):
LGTM

vroldanbet (approved):

PR #1591, opened 2023-10-18
.github: move to buildjet

jzelinskie:
This change migrates our GitHub Actions to all run on [BuildJet](buildjet.com), which is cheaper and faster than GitHub's provided service.

Previously we were relying on some DockerHub preventing rate limiting on GitHub's network, but they've ceased that behavior. We've migrated as many public images off DockerHub, so now we can take advantage of BuildJet.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -14,7 +14,7 @@ env:
   GO_VERSION: "~1.21.3"
 jobs:
   paths-filter:
-    runs-on: "ubuntu-latest"
+    runs-on: "buildjet-2vcpu-ubuntu-2204"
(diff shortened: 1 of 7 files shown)


.github/workflows/build-test.yaml
```diff
@@ -114,28 +102,12 @@ jobs:
           go-version: "${{ env.GO_VERSION }}"
       - name: "Integration tests"
         run: "go run mage.go testds:${{ matrix.datastore }}"
-
-  datastoreconsistency:
```
vroldanbet, line 118: why were these removed?

vroldanbet (approved):

PR #1592, opened 2023-10-18
spanner: add missing calls to RowIterator.Stop

vroldanbet:
Spanner Go client says we should be calling it


jzelinskie (approved):

vroldanbet:
not needed since `Do` calls `Stop`

jzelinskie:
I still think it might be worth merging this for sanity's sake. This isn't the first time folks have thought we're missing the Stop() calls to the RowIterator.

PR #1596, opened 2023-10-19
tracing ux: enriches traversal operations

vroldanbet:
⚠️ built on top of https://github.com/authzed/spicedb/pull/1598

this renders in a more obvious way how graph traversal is happening during check requests, showing
where is time being spent in the schema.

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/samber/lo"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	"github.com/authzed/spicedb/internal/dispatch"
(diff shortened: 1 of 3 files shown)


internal/graph/check.go
```diff
@@ -430,10 +446,20 @@ func mapFoundResources(result CheckResult, resourceType *core.RelationReference,
 func (cc *ConcurrentChecker) checkUsersetRewrite(ctx context.Context, crc currentRequestContext, rewrite *core.UsersetRewrite) CheckResult {
 	switch rw := rewrite.RewriteOperation.(type) {
 	case *core.UsersetRewrite_Union:
-		return union(ctx, crc, rw.Union.Child, cc.runSetOperation, cc.concurrencyLimit)
+		innerCtx := ctx
```
ecordell, line 449: not sure you need this, it seems fine to shadow like you do in the other branches

vroldanbet: you are right, I messed up while trying to only trace with there is more than one child

internal/graph/check.go
```diff
@@ -204,8 +208,8 @@ type directDispatch struct {
 }
 
 func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequestContext, relation *core.Relation) CheckResult {
-	log.Ctx(ctx).Trace().Object("direct", crc.parentReq).Send()
```
josephschorr, line 207: I'd keep this trace log for people not using tracing, but just `--log-level=trace`

josephschorr: My bad

josephschorr (approved):
LGTM

PR #1597, opened 2023-10-19
middleware/consistency: fix source of atLeast

jzelinskie:
Previously, we were recording all requests as being sourced from the request when the server could have picked the revision.


vroldanbet (approved):

vroldanbet (changes requested):
Unit test failures seem legit

PR #1598, opened 2023-10-20
propagates gRPC errors in graph package

vroldanbet:
When the datastore has a gRPC API, we may get
a grpc.Status as error, which does not respond to
`error.Is(ctx, context.Canceled)`.

Like in other parts of the code, we determine if
this is a `grpc.Status` and return it as is.


josephschorr (dismissed):
LGTM

josephschorr (approved):

PR #1599, opened 2023-10-20
Revert Optimistic Locking in Spanner

vroldanbet:
It appears it's not GA yet


ecordell (approved):
Couldn't we just put this behind a flag so that when it is GA in spanner we can enable it?

vroldanbet:
@ecordell yeah totally, let's do that in a follow up 👍🏻

PR #1602, opened 2023-10-20
add singleflight dispatcher for check requests

vroldanbet:
## What

This PR adds a new singleflight `dispatch.Dispatcher` capable of deduping `DispatchCheckRequest` (for now, more to come later).

is uses `janos/singleflight` instead of `x/sync/singleflight`, because the latter does not propagate cancelation, and
we want to abort any expensive operation (especially database operations) as soon as possible. janos/single will let the handler run until the last single-flighted caller is canceled. In exchange, it has to spin an extra goroutine compared to x/sync/singleflight.

## Previous iteration

- I previously tried to implement it as a gRPC Dispatcher API middleware, but that made things worse because we bypassed the caching dispatcher, which happens at a different level. Adding it at the dispatcher level proved to work as expected

## Other

- I also added a follow-up fix to https://github.com/authzed/spicedb/pull/1598, where the wrong error is being returned in the `rewriteGraphError` method


internal/dispatch/singleflight/singleflight.go
```diff
@@ -0,0 +1,86 @@
...
+	return &Dispatcher{delegate: delegate}
+}
+
+type Dispatcher struct {
+	delegate   dispatch.Dispatcher
+	checkGroup singleflight.Group[string, *v1.DispatchCheckResponse]
+}
+
+func (d *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckRequest) (*v1.DispatchCheckResponse, error) {
+	key, err := hashForDispatchCheck(req)
```
ecordell, line 36: only doing dispatchcheck as a test? or is there some reason to not singleflight the rest?

internal/dispatch/singleflight/singleflight.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return nil, status.Error(codes.Internal, "unexpected DispatchCheck error")
+	}
+
+	// TODO should we do a copy of the proto response?
+	v, isShared, err := d.checkGroup.Do(ctx, key, func(innerCtx context.Context) (*v1.DispatchCheckResponse, error) {
```
jzelinskie, line 42: In a previous experiment doing this I used `github.com/authzed/spicedb/internal/dispatch/keys.Handler` instead of recreating another hashing methodology.

internal/dispatch/singleflight/singleflight.go
ecordell: nvm, you mentioned this in the PR description

vroldanbet: There is no reason not to singleflight everything, I just wanted to get this out of the door ASAP and then we can follow up with singleflight for the rest of requests.

vroldanbet:
Closed in favour of https://github.com/authzed/spicedb/pull/1607

PR #1603, opened 2023-10-20
Add support for experimental secondary dispatching

josephschorr:
When enabled and configured, (for now, check) dispatches will also be made to the secondary dispatchers returned by the CEL expression. This will enable items such as cache prewarming by dispatching to a backup cluster while the primary cluster is still running

### DIFF
--- a/internal/dispatch/remote/cluster.go
+++ b/internal/dispatch/remote/cluster.go
@@ -5,11 +5,15 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/authzed/consistent"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/rs/zerolog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/internal/dispatch/keys"
@@ -18,7 +22,18 @@ import (
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
-type clusterClient interface {
+var dispatchCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "dispatch",
+	Name:      "remote_dispatch_handler_total",
+	Help:      "which dispatcher handled a request",
+}, []string{"request_kind", "handler_name"})
+
+func init() {
+	prometheus.MustRegister(dispatchCounter)
+}
+
+type ClusterClient interface {
 	DispatchCheck(ctx context.Context, req *v1.DispatchCheckRequest, opts ...grpc.CallOption) (*v1.DispatchCheckResponse, error)
 	DispatchExpand(ctx context.Context, req *v1.DispatchExpandRequest, opts ...grpc.CallOption) (*v1.DispatchExpandResponse, error)
 	DispatchReachableResources(ctx context.Context, in *v1.DispatchReachableResourcesRequest, opts ...grpc.CallOption) (v1.DispatchService_DispatchReachableResourcesClient, error)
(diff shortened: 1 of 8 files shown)


pkg/cmd/serve.go
```diff
@@ -115,6 +115,9 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	cmd.Flags().Uint16Var(&config.DispatchHashringReplicationFactor, "dispatch-hashring-replication-factor", 100, "set the replication factor of the consistent hasher used for the dispatcher")
 	cmd.Flags().Uint8Var(&config.DispatchHashringSpread, "dispatch-hashring-spread", 1, "set the spread of the consistent hasher used for the dispatcher")
 
+	cmd.Flags().StringToStringVar(&config.DispatchSecondaryUpstreamAddrs, "experimental-dispatch-secondary-upstream-addrs", nil, "secondary upstream addresses for dispatches, each with a name")
+	cmd.Flags().StringToStringVar(&config.DispatchSecondaryUpstreamExprs, "experimental-dispatch-secondary-upstream-exprs", nil, "map from request type to its associated CEL expression, which returns the secondary upstream(s) to be used for the request")
```
ecordell, line 119: can you document the available request types here?

internal/dispatch/remote/cluster.go
```diff
@@ -75,18 +89,99 @@ func (cr *clusterDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
...
-	err = adjustMetadataForDispatch(resp.Metadata)
 	return resp, err
 }
 
+type respTuple[S any] struct {
+	resp *S
+	err  error
+}
+
+func dispatchRequest[Q zerolog.LogObjectMarshaler, S any](ctx context.Context, cr *clusterDispatcher, reqKey string, req Q, handler func(context.Context, ClusterClient) (*S, error)) (*S, error) {
```
ecordell, line 113: can we tighten these types for Q and S?

internal/dispatch/remote/cluster.go
```diff
@@ -75,18 +89,99 @@ func (cr *clusterDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
...
+			secondaryResultChan <- resp
+		}()
+	}
+
+	var foundError error
+	select {
+	case <-withTimeout.Done():
+		return nil, fmt.Errorf("check dispatch has timed out")
+
+	case r := <-primaryResultChan:
```
ecordell, line 169: it would be nice to have a metric for secondary/primary dispatches. A counter with a label for which dispatcher handled it?

internal/dispatch/remote/cluster.go
```diff
@@ -75,18 +89,99 @@ func (cr *clusterDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
...
+
+		// Otherwise, if an error was found, log it and we'll return after *all* the secondaries have run.
+		// This allows an otherwise error-state to be handled by one of the secondaries.
+		foundError = r.err
+
+	case resp := <-secondaryResultChan:
+		return resp, nil
+	}
+
+	return nil, foundError
```
ecordell, line 182: returning without waiting for the secondary dispatch may not work for cache warming, since we may cancel the request before it can run and cache in the other cluster.

though perhaps this should be configured on the receiving cluster? could return immediately with no result while it runs the request, and let the primary do the main computation.

internal/dispatch/remote/cluster.go
```diff
@@ -75,18 +89,99 @@ func (cr *clusterDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
...
+	result, err := RunDispatchExpr(expr, req)
+	if err != nil {
+		log.Warn().Err(err).Msg("error when trying to evaluate the dispatch expression")
+	}
+
+	log.Trace().Str("secondary-dispatchers", strings.Join(result, ",")).Object("request", req).Msg("running secondary dispatchers")
+	for _, secondary := range cr.secondaryDispatch {
+		secondary := secondary
+
+		if !stringz.SliceContains(result, secondary.Name) {
```
ecordell, line 146: there's a stdlib `slices.Contains` now

internal/dispatch/remote/cluster.go
```diff
@@ -75,18 +89,99 @@ func (cr *clusterDispatcher) DispatchCheck(ctx context.Context, req *v1.Dispatch
...
+		primaryResultChan <- respTuple[S]{resp, err}
+	}()
+
+	result, err := RunDispatchExpr(expr, req)
+	if err != nil {
+		log.Warn().Err(err).Msg("error when trying to evaluate the dispatch expression")
+	}
+
+	log.Trace().Str("secondary-dispatchers", strings.Join(result, ",")).Object("request", req).Msg("running secondary dispatchers")
+	for _, secondary := range cr.secondaryDispatch {
```
ecordell, line 143: if `cr.secondaryDispatch` were a map keyed on name, you could just directly look up every dispatcher in the result

ecordell (changes requested):
Looking nice!

Offline we discussed the hashring: since this is using the same grpc config, secondary dispatches should use the same hash key but mapped into a different hashring (for the secondary cluster)

internal/dispatch/remote/cluster.go
josephschorr: Added

internal/dispatch/remote/cluster.go
josephschorr: Yeah, that was my intention: the receiving cluster would simply no-op and go off and fill the cache

josephschorr:
Updated

PR #1609, opened 2023-10-24
More spanner observability / Go Scheduler metrics

vroldanbet:
- adds opentelemetry integration into Spanner client
- exposes Spanner client metrics via prometheus
- enables All Go runtime metrics (including scheduler)


pkg/runtime/metrics.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
+)
+
+// prometheus client_golang by default registers a collector that collects all metrics, except scheduler metrics
+// this package unregisters the default collector and adds one that includes scheduler metrics
+//
+// in order to register this, the package must be imported anonymously
+func init() {
+	prometheus.DefaultRegisterer.Unregister(collectors.NewGoCollector())
```
ecordell, line 13: Ah nice, I spent a bunch of time trying to register the scheduler metrics but they kept including metrics that conflicted with existing ones as a side-effect.

ecordell (approved):

pkg/runtime/metrics.go
vroldanbet: Spent most of the time dealing with the same 😅  We could also create a new promhttp handler with a `Gatherer`, but would require a different HTTP endpoint and that's a mess. The reason we have to do this is because we rely heavily on the default prometheus Registry - we may want to revisit that.

PR #1615, opened 2023-10-26
log payloads, and add grpc duration as an integer

vroldanbet:
in order to better troubleshoot certain situations, it would be useful to load request and response payloads.

Also in certain log aggregators it's not possible to convert duration fields that are represented as a string, and this is useful in order to filter requests by a specific duration threshold. This commit changes the duration function to log it as an integer

Adds options to control request and response payloads via CLI

### DIFF
--- a/pkg/cmd/serve.go
+++ b/pkg/cmd/serve.go
@@ -47,6 +47,10 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	config.DispatchClusterMetricsEnabled = true
 	config.DispatchClientMetricsEnabled = true
 
+	// Flags for logging
+	cmd.Flags().BoolVar(&config.EnableRequestLogs, "grpc-log-requests-enabled", false, "logs API request payloads")
+	cmd.Flags().BoolVar(&config.EnableResponseLogs, "grpc-log-responses-enabled", false, "logs API response payloads")
+
 	// Flags for the gRPC API server
 	util.RegisterGRPCServerFlags(cmd.Flags(), &config.GRPCServer, "grpc", "gRPC", ":50051", true)
 	cmd.Flags().StringSliceVar(&config.PresharedSecureKey, PresharedKeyFlag, []string{}, "preshared key(s) to require for authenticated requests")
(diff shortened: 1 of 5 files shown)


pkg/cmd/server/defaults.go
```diff
@@ -116,7 +117,10 @@ var defaultGRPCLogOptions = []grpclog.Option{
 		return grpclog.DefaultServerCodeToLevel(code)
 	}),
 	// changes default logging behaviour to only log finish call message
-	grpclog.WithLogOnEvents(grpclog.FinishCall),
+	grpclog.WithLogOnEvents(grpclog.PayloadReceived, grpclog.PayloadSent, grpclog.FinishCall),
```
ecordell, line 120: Since payloads may be sensitive, we should probably put this behind a flag.

I'm okay approving this and then doing a flag in a followup before the next release if you prefer, just let me know.

vroldanbet: I can work on the flag!

pkg/cmd/serve.go
```diff
@@ -47,6 +47,10 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	config.DispatchClusterMetricsEnabled = true
 	config.DispatchClientMetricsEnabled = true
 
+	// Flags for logging
+	cmd.Flags().BoolVar(&config.EnableRequestLogs, "enable-request-logs", false, "logs API request payloads")
```
jzelinskie, line 51: Flags need to be named with a prefix for the subsystem they control, so that they are grouped together.
We should prefix this with `grpc` or `log`; probably log.

Any existing flags that start with `enable` were mistakes and should probably have their flags hidden and new ones introduced.

jzelinskie (changes requested):

pkg/cmd/serve.go
jzelinskie: `grpc-request-logs-enabled` `grpc-repsonse-logs-enabled`

jzelinskie (approved):

jzelinskie (approved):

PR #1621, opened 2023-10-27
adds flags to enable block and mutex profiles

vroldanbet:
this introduces two new flags that control the sampling rate for mutex and block profiles. The value is an integer representing the fraction of samples to gather, with 0 being none, 1 being 100%, and N being 1/N samples.

**Both are disabled disabled by default**

### Enable Mutex Profile for 100% of samples

```bash
spicedb serve --pprof-mutex-profile-rate 1
```

### Enable Mutex Profile for 0.1% of samples

```bash
spicedb serve --pprof-mutex-profile-rate 1000
```

### Enable Block Profile for 100% of samples

```bash
spicedb serve --pprof-block-profile-rate 1
```

### Enable Block Profile for 0.1% of samples

```bash
spicedb serve --pprof-block-profile-rate 1000
```

## How To Test This

To test this you can use good old pre-Grafana Pyroscope, which is significantly easier to run in local environments:

Assuming SpiceDB running locally, create `pyroscope.yaml`

```yaml
storage-path: /var/lib/pyroscope
scrape-configs:
- enabled-profiles:
  - cpu
  - mem
  - goroutines
  - mutex
  - block
  job-name: spicedb-job
  static-configs:
  - application: spicedb
    scrape-interval: 1s
    targets:
      - localhost:9090
```

```bash
open http://localhost:4040 && docker run --net host -v $(pwd)/pyroscope.yaml:/etc/pyroscope/server.yaml -p 4040:4040 pyroscope/pyroscope:latest server --config /etc/pyroscope/server.yaml
```


magefiles/lint.go
```diff
@@ -89,7 +89,7 @@ func (Lint) Analyzers() error {
 		"-closeafterusagecheck.must-be-closed-after-usage-types=github.com/authzed/spicedb/pkg/datastore.RelationshipIterator",
 		"-closeafterusagecheck.skip-pkg=github.com/authzed/spicedb/pkg/datastore,github.com/authzed/spicedb/internal/datastore,github.com/authzed/spicedb/internal/testfixtures",
 		"-paniccheck",
-		"-paniccheck.skip-files=_test,zz_",
+		"-paniccheck.skip-files=_test,zz_,profiling.go",
```
josephschorr, line 92: I'd remove this and just name the functions `mustWhatever`, to allow them to bypass

pkg/runtime/profiling.go
```diff
@@ -0,0 +1,43 @@
...
+	"runtime"
+	"strconv"
+)
+
+func init() {
+	initializeMutexProfile()
+	initializeBlockProfile()
+}
+
+// initializeMutexProfile Enables the mutex profile
```
josephschorr, line 14: initializeMutexProfile enables the mutex profile

pkg/runtime/profiling.go
```diff
@@ -0,0 +1,43 @@
...
+		val, err := strconv.ParseInt(provided, 10, 32)
+		if err != nil {
+			panic("invalid integer value provided in SPICEDB_MUTEX_PROFILE_RATE: " + provided)
+		}
+
+		runtime.SetMutexProfileFraction(int(val))
+	}
+}
+
+// initializeBlockProfile Enable the block profile
```
josephschorr, line 32: initializeBlockProfile enables the block profile

josephschorr (dismissed):
LGTM

jzelinskie:
Does this absolutely need to be ran in `init()`? It'd be nice if we could just make this a global CLI flag so that it works like every other env var

ecordell (approved):
LGTM

PR #1624, opened 2023-10-27
Ensure all datastores return an error if accessed after Close

### LINKED ISSUE #604: Audit the usage of Dispatcher and Datastore for use after being closed
Some of our fire-and-forgotten requests, that are no longer needed due to an earlier answer, can continue to use the datastore after it has been closed. We should probably track and audit each and every goroutine that we fire off for correctness and to prevent things from being used after they're closed.

josephschorr:
Fixes #604 by ensuring that attempts as use-after-close properly returns an error, which is safe for fire-and-forget operations

### DIFF
--- a/pkg/datastore/test/basic.go
+++ b/pkg/datastore/test/basic.go
@@ -1,0 +1,24 @@
+package test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func UseAfterCloseTest(t *testing.T, tester DatastoreTester) {
+	require := require.New(t)
+
+	// Create the datastore.
+	ds, err := tester.New(0, veryLargeGCInterval, veryLargeGCWindow, 1)
+	require.NoError(err)
+
+	// Immediately close it.
+	err = ds.Close()
+	require.NoError(err)
+
+	// Attempt to use and ensure an error is returned.
+	_, err = ds.HeadRevision(context.Background())
+	require.Error(err)
+}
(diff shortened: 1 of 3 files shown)


pkg/datastore/test/basic.go
```diff
@@ -0,0 +1,24 @@
...
+	// Create the datastore.
+	ds, err := tester.New(0, veryLargeGCInterval, veryLargeGCWindow, 1)
+	require.NoError(err)
+
+	// Immediately close it.
+	err = ds.Close()
+	require.NoError(err)
+
+	// Attempt to use and ensure an error is returned.
+	_, err = ds.HeadRevision(context.Background())
```
ecordell, line 22: doesn't this only show that `HeadRevision` can't be used after close?

ecordell (commented):
I'm not sure this addresses https://github.com/authzed/spicedb/issues/604, which I understand to be about continuing to use the datastore after a dispatch has been cancelled, and not when a datastore is closed (which really only happens at shutdown anyway).

pkg/datastore/test/basic.go
josephschorr: Yes; we could add an exhaustive list but since most operations start with revision checking, I expect any other reader operations will fail too

vroldanbet: @josephschorr wouldnt `OptimizedRevision` arguably be used more often than `HeadRevision`?

josephschorr: As I said, I can add more methods, but ultimately the only true way to be certain would be to test them all. I just wanted to verify that the standard path would result in an error

vroldanbet (approved):

PR #1625, opened 2023-10-27
Remove support for MySQL v5

### LINKED ISSUE #1338: Deprecation Notice: MySQL v5 support
MySQL v5 goes EOL in 5 months from now: https://endoflife.date/mysql

This is a notice that we will soon deprecate v5 support and move the MySQL datastore integration tests to use v8.

josephschorr:
Fixes #1338

To be merged no earlier than Oct 31, when MySQL v5 reaches EOL

### DIFF
--- a/internal/datastore/mysql/datastore_test.go
+++ b/internal/datastore/mysql/datastore_test.go
@@ -94,24 +94,11 @@ func TestMySQLDatastoreDSNWithoutParseTime(t *testing.T) {
 	require.ErrorContains(t, err, "https://spicedb.dev/d/parse-time-mysql")
 }
 
-func TestMySQL5Datastore(t *testing.T) {
-	b := testdatastore.RunMySQLForTesting(t, "")
-	dst := datastoreTester{b: b, t: t}
-	test.All(t, test.DatastoreTesterFunc(dst.createDatastore))
-	additionalMySQLTests(t, b)
-}
-
 func TestMySQL8Datastore(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/datastore/mysql/datastore_test.go
```diff
@@ -94,13 +94,6 @@ func TestMySQLDatastoreDSNWithoutParseTime(t *testing.T) {
...
 }
 
-func TestMySQL5Datastore(t *testing.T) {
-	b := testdatastore.RunMySQLForTesting(t, "")
-	dst := datastoreTester{b: b, t: t}
-	test.All(t, test.DatastoreTesterFunc(dst.createDatastore))
-	additionalMySQLTests(t, b)
-}
-
 func TestMySQL8Datastore(t *testing.T) {
```
ecordell, line 97: https://github.com/josephschorr/spicedb/blob/608f9eccb7ec17b63882f985f5ee90fb8f6d5649/internal/testserver/datastore/mysql.go#L37

I think you want to remove a bunch of the v5 stuff here as well

josephschorr: Looks like the file forgot to save; VSCode does that occasionally

vroldanbet:
>To be merged no earlier than Oct 31, when MySQL v5 reaches EOL

We can probably merge it already as the next release will almost certainly happen after Oct 31st

vroldanbet (approved):
LGTM

Issue #1217, opened 2023-03-21
support pgbouncer with SpiceDB Postgres datastore w/o session pooler mode


vroldanbet:
a few users have reported that having pgbouncer in front of Postgres does not work with the PG datastore. See this [discord discussion](https://discord.com/channels/844600078504951838/844600078948630559/1060960536117907517):

````
2 UNKNOWN: unable to find revision: ERROR: prepared statement \"lrupsc_89_0\" does not exist (SQLSTATE 26000)
````

[Another user reported](https://github.com/authzed/spicedb/issues/1992) an error like:
```
prepared statement name is already in use (SQLSTATE 08P01)
```

The issue seems to originate from the fact that SpiceDB uses a pgx execution mode that does 2 roundtrips, one to prepare the statement and another to execute it. PGBouncer would presumably switch the underlying connections, which can lead to errors like the above described. 

The current workaround, as described by some SpiceDB users, is to configure PgBouncer with connection pooler mode set to `session` and to ignore `plan_cache_mode` with `ignore_startup_parameters=plan_cache_mode`.

Ideally, a sustainable long-term solution is found that does not involve configuring specific settings in PgBouncer.

**Comment by vroldanbet** [comment]
Reopening as we are now testing it, but not documenting it, @bradengroom would you be so kind to add some docs? 🙏🏻 I think https://github.com/authzed/docs/blob/main/docs/spicedb/selecting-a-datastore.md would be the right place

**Comment by sashayakovtseva** [comment]
As I understand, session_pooling is the only possible way to use pgbouncer and it's not planned to support tx_pooling, correct?

**Comment by vroldanbet** [comment]
@sashayakovtseva from the work done by @bradengroom in https://github.com/authzed/spicedb/pull/1626, and also from a cursory look online, it appears that SpiceDB would need `session_pooling` for prepared statements. SpiceDB now has a nice test-suite contributed by @bradengroom that could help determining if it works.

It does not mean that _it won't be supported_, if there is a critical mass of folks requesting this, it could be supported, at the expense of query performance degradation (as you can see from https://github.com/authzed/spicedb/pull/1626).

I think some of the work on https://github.com/authzed/spicedb/pull/1626 could be reused to support having folks disabling prepared statements.

**Comment by pdeaudney** [comment]
In PostgreSQL you can also set GUC values such as `plan_cache_mode` on the database or user. https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-SQL

```
[...]

PR #1626, opened 2023-10-29
Disable prepared statements via pgx config

bradengroom:
Fixes https://github.com/authzed/spicedb/issues/1217

Instead of addressing this issue by documenting configuration required for pgbouncer to work, this PR removes the need for special pgbouncer configuration.

In the past, [a bug was found](https://spicedb.dev/d/force-custom-plan) with the use of prepared statements which impacted SpiceDB's performance. SpiceDB uses pgx to interact with Postgres which [uses prepared statements by default](https://github.com/jackc/pgx/wiki/Automatic-Prepared-Statement-Caching).

SpiceDB addressed the previously mentioned performance issue by injecting the `plan_cache_mode=force_custom_plan` setting into the postgres connection string. This fix mimicked the fix put in place by the author of [the blog post](https://spicedb.dev/d/force-custom-plan) outlining the original investigation into the bug.

While this fix was the only immediate option for the blog post author, SpiceDB can do better by changing pgx's execution mode. The use of `plan_cache_mode` does not truly disable prepared statements, but instead leaves pgx continuing its use of prepared statements while SpiceDB simultaneously asks postgres to ignore the query plans it computes for them upfront. Disabling prepared statements entirely via pgx has several benefits:
* If SpiceDB decides to selectively enable prepared statements for some queries in the future then it will be possible.
* Not disabling prepared statements via `plan_cache_mode` simplifies the story around PgBouncer deployments by removing the need  for special PgBouncer configuration. There are currently two PgBouncer configurations that are needed to use PgBouncer with SpiceDB:
  - `POOL_MODE=session` - This is needed to deal with [an issue reported by a user](https://github.com/authzed/spicedb/issues/1217) wherein PgBouncer was cleaning up prepared statements. (An odd issue to encounter given SpiceDB wishes to disable prepared statements.)
  - `IGNORE_STARTUP_PARAMETERS=plan_cache_mode` - PgBouncer does not understand the `plan_cache_mode` parameter, so it must be instructed to ignore it and pass it through to Postgres. This is another strange requirement. SpiceDB's method of disabling prepared statements is leaking out to PgBouncer users.

Discord conversation: https://discord.com/channels/844600078504951838/900405749405089812/1164379648650649700


internal/testserver/datastore/postgres.go
```diff
@@ -167,9 +167,6 @@ func (b *postgresTester) runPgbouncerForTesting(t testing.TB, pool *dockertest.P
 			"DB_NAME=*",     // Needed to make pgbouncer okay with the randomly named databases generated by the test suite
 			"AUTH_TYPE=md5", // use the same auth type as postgres
 			"MAX_CLIENT_CONN=" + POSTGRES_TEST_MAX_CONNECTIONS,
-			// params needed for spicedb
-			"POOL_MODE=session",                         // https://github.com/authzed/spicedb/issues/1217
-			"IGNORE_STARTUP_PARAMETERS=plan_cache_mode", // Tell pgbouncer to pass this param thru to postgres.
```
bradengroom, line 172: No longer needed 🙂

bradengroom:
I think the failing unit test is a flake?

internal/datastore/postgres/postgres.go
```diff
@@ -134,22 +134,25 @@ func newPostgresDatastore(
...
-		return nil, fmt.Errorf(errUnableToInstantiate, err)
-	}
+	// By default, pgx uses prepared statements for all queries. This disables
+	// the use of prepared statements in order to work around a bug impacting
+	// performance documented here:
+	// https://spicedb.dev/d/force-custom-plan.
+	// The docs for the different execution modes offered by pgx may be found
+	// here:
+	// https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode
+	parsedConfig.ConnConfig.DefaultQueryExecMode = pgx.QueryExecModeCacheDescribe
```
vroldanbet, line 144: The query exec mode can also be provided via the DB DSN. By hardcoding this here we are effectively overriding any value provided. While I don't think there is a reason for users to change this, it would be helpful in case this causes a regression to any deployment out there and folks need a way to revert back to the original behaviour.

vroldanbet (commented):
@bradengroom I'd suggest running the [datastore benchmarks](https://github.com/authzed/spicedb/tree/2bf82db9c7b462619588f261145c41255d9ac115/internal/datastore/benchmark) to make sure this is not causing any regression

jzelinskie:
cc @josephschorr @williamdclt

internal/datastore/postgres/postgres.go
bradengroom: This appears to be set with `default_query_exec_mode=cache_describe`. I'm not sure how you knew this, but I only see examples in `pgx` code/tests. Thanks for catching! Agreed that having an escape hatch is super valuable. I'll try to get that in place.

bradengroom:
Still WIP. I plan to write some tests for `defaultQueryExecMode`. For now, I think it's closer to what Victor suggested.

internal/datastore/postgres/postgres.go
```diff
@@ -477,25 +477,29 @@ func currentlyLivingObjects(original sq.SelectBuilder) sq.SelectBuilder {
...
 // This works around a bug impacting performance documented here:
 // https://spicedb.dev/d/force-custom-plan.
-func defaultCustomPlan(poolConfig *pgxpool.Config) (*pgxpool.Config, error) {
-	if existing, ok := poolConfig.ConnConfig.Config.RuntimeParams["plan_cache_mode"]; ok {
+//
+// The docs for the different execution modes offered by pgx may be found
+// here:
+// https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode
+func defaultQueryExecMode(poolConfig *pgxpool.Config) *pgxpool.Config {
+	if existing, ok := poolConfig.ConnConfig.Config.RuntimeParams["default_query_exec_mode"]; ok {
```
bradengroom, line 489: This won't work because pgx deletes this param from `RuntimeParams`
https://github.com/jackc/pgx/blob/cf6ef75f916648857cf7b46322d6d7af7d372917/conn.go#L173

I need to find another way to determine if the user omitted this vs wants to override

internal/datastore/postgres/postgres.go
```diff
@@ -477,25 +483,43 @@ func currentlyLivingObjects(original sq.SelectBuilder) sq.SelectBuilder {
...
-			Str("plan_cache_mode", existing).
-			Msg("found plan_cache_mode in DB URI; leaving as-is")
-		return poolConfig, nil
+//
+// The docs for the different execution modes offered by pgx may be found
+// here:
+// https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode
+func DefaultQueryExecMode(poolConfig *pgxpool.Config) *pgxpool.Config {
+
+	if !strings.Contains(poolConfig.ConnString(), "default_query_exec_mode") {
```
bradengroom, line 496: This is annoying, but we can't check RuntimeParams since pgx deletes this particular value.
https://github.com/jackc/pgx/blob/cf6ef75f916648857cf7b46322d6d7af7d372917/conn.go#L173

internal/datastore/postgres/postgres.go
```diff
@@ -97,6 +98,14 @@ var (
 	getNow = psql.Select("NOW()")
 
 	tracer = otel.Tracer("spicedb/internal/datastore/postgres")
+
+	supportedQueryExecModes = map[pgx.QueryExecMode]string{
+		pgx.QueryExecModeCacheStatement: "cache_statement",
+		pgx.QueryExecModeCacheDescribe:  "cache_describe",
+		pgx.QueryExecModeDescribeExec:   "describe_exec",
+		pgx.QueryExecModeExec:           "exec",
+		pgx.QueryExecModeSimpleProtocol: "simple_protocol",
```
bradengroom, line 107: TODO: I need to go through and verify which of these do not work. Some execution modes require type mappings to be registered, so they'll just fail. I'll circle back to this tomorrow.

bradengroom: Also write a test for this path.

internal/datastore/postgres/postgres.go
```diff
@@ -477,25 +483,43 @@ func currentlyLivingObjects(original sq.SelectBuilder) sq.SelectBuilder {
...
+// https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode
+func DefaultQueryExecMode(poolConfig *pgxpool.Config) *pgxpool.Config {
+
+	if !strings.Contains(poolConfig.ConnString(), "default_query_exec_mode") {
+		// the execution mode was not overridden by the user
+		poolConfig.ConnConfig.DefaultQueryExecMode = pgx.QueryExecModeCacheDescribe
+		log.Warn().
+			Str("details-url", "https://spicedb.dev/d/force-custom-plan").
+			Str("default_query_exec_mode", "cache_describe").
+			Msg("defaulting value in Postgres DB URI")
```
bradengroom, line 502: Please let me know if I'm not following the project's logging patterns.

vroldanbet:
@bradengroom is this ready for review?

bradengroom:
@vroldanbet I ran the benchmarks on both this branch and main to make sure I'm using up to date results.

this branch:
```
BenchmarkDatastoreDriver/postgres/TestTuple/SnapshotRead-10                                     	    2560	    521779 ns/op	   12890 B/op	     224 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadOnlyNamespace/BySubject-10        	      58	  20676485 ns/op	 2376906 B/op	   55159 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadOnlyNamespace/ByResource-10       	     100	  12058841 ns/op	 2376806 B/op	   55159 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadWithRelation/ByResource-10        	     338	   3385922 ns/op	 2378896 B/op	   55180 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadWithRelation/BySubject-10         	      78	  20775982 ns/op	 2378724 B/op	   55179 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadAllResourceFields/ByResource-10   	    2240	    521134 ns/op	   16757 B/op	     286 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadAllResourceFields/BySubject-10    	    2314	    538459 ns/op	   16758 B/op	     286 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SnapshotReverseRead-10                              	     100	  13424350 ns/op	 2377927 B/op	   55174 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/Touch-10                                            	     861	   1492052 ns/op	   12263 B/op	     185 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/Create-10                                           	     562	   1788423 ns/op	   10136 B/op	     147 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0_-10                                	      24	  71060014 ns/op	12762846 B/op	  145667 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0.1_-10                              	      15	  69570836 ns/op	12262022 B/op	  136301 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0.25_-10                             	      16	  70275073 ns/op	10857274 B/op	  122497 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0.5_-10                              	      20	  61499912 ns/op	 8762044 B/op	   99247 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/1_-10                                	      37	  40974766 ns/op	 5452780 B/op	   52962 allocs/op
```

main:
```
BenchmarkDatastoreDriver/postgres/TestTuple/SnapshotRead-10                                     	    2408	    462334 ns/op	   12890 B/op	     224 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadOnlyNamespace/BySubject-10        	      63	  20798763 ns/op	 2376729 B/op	   55159 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadOnlyNamespace/ByResource-10       	      84	  11989514 ns/op	 2376877 B/op	   55159 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadWithRelation/ByResource-10        	     356	   3298096 ns/op	 2378879 B/op	   55180 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadWithRelation/BySubject-10         	      79	  20965164 ns/op	 2378815 B/op	   55180 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadAllResourceFields/ByResource-10   	    2192	    500163 ns/op	   16758 B/op	     286 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SortedSnapshotReadAllResourceFields/BySubject-10    	    2118	    490966 ns/op	   16756 B/op	     286 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/SnapshotReverseRead-10                              	      96	  13861234 ns/op	 2377958 B/op	   55173 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/Touch-10                                            	     829	   1412580 ns/op	   12402 B/op	     185 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/Create-10                                           	     889	   1402953 ns/op	   10200 B/op	     144 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0_-10                                	      24	  48112885 ns/op	12167486 B/op	  145690 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0.1_-10                              	      28	  43083798 ns/op	11611001 B/op	  136315 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0.25_-10                             	      25	  43459102 ns/op	10314110 B/op	  122515 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/0.5_-10                              	      26	  46815375 ns/op	 8213800 B/op	   99264 allocs/op
BenchmarkDatastoreDriver/postgres/TestTuple/CreateAndTouch/1_-10                                	      37	  34119407 ns/op	 5110080 B/op	   52965 allocs/op
```

The largest difference is on the `CreateAndTouch` tests. `CreateAndTouch/0_-10` is registering as 22ms slower. I'm unsure if this is an acceptable regression or not (I assume not). It's worth noting that removing the use prepared statements will definitely be slower (by _some_ amount). Even though SpiceDB currently disables the use of cached plans for prepared statements, it is likely still benefiting from prepared statements in these ways:
* Fewer bytes over the wire (pgx only has to send `EXECUTE foo(args1, arg2, ...);` instead of the fully formed statement). This also incurs less of a parsing cost on the postgres side since there's just less to parse.
* Likely no statement -> logical plan conversion. Even though we've asked postgres not to reuse the same _physical_ plan, I would not be surprised if it is still reusing the same _logical_ plan.

While the benefits of turning off prepared statements provides a streamlined deployment story for pgbouncer users:
* No restriction to use the session pool_mode
* No need to ignore the plan_cache_mode setting

So I think that's the trade-off in this PR: perf vs ease of pgbouncer deployment. Given the nature of the project, I assume perf wins. If the Authzed folks agree then I'm happy to close this PR and go back to documenting the required PgBouncer configs. This was an interesting exercise regardless 🙂

I think it's not _as_ strange to document the pgbouncer config requirements now that they can be tied to perf improvements.

bradengroom:
It's also worth noting that the `CreateAndTouch` test creates 1000 tuples:
https://github.com/authzed/spicedb/blob/018aad9be9172e9c57999bdf3cd52c112e0bb00f/internal/datastore/benchmark/driver_bench_test.go#L199

So that 22 additional milliseconds is spread out over 1k operations. (22 _microseconds_ per operation). After realizing that, it doesn't actually feel that bad I think I would go for this. I'm still to pursue whatever the Authzed folks want though 🙂

bradengroom:
Sorry, one last comment. If this gets merged, it can be disabled by adding the following to your postgres connection string:
`?default_query_exec_mode=cache_statement&plan_cache_mode=force_custom_plan`

vroldanbet:
Hey @bradengroom, thank you so much for the thorough investigation, this is excellent work 💯 . I'm a bit on the fence here - I recall also looking into disabling prepared statements back in the day and I recall that it regressed somewhat the latencies. I'm going to run some tests internally to see what it looks like. Let's hold on merging this for now 🙏🏻

PR #1627, opened 2023-10-30
introduces flags to set min/max Spanner sessions

vroldanbet:
Aside from having a connection pool of gRPC sessions against the Spanner API, the client library also
keeps a number of sessions spread over those connections. This setting may need tweaking depending on the dispatch fan out of the schema, and may end blocking on
the session pool if not enough sessions are available.

From the Spanner docs:
>Set MinSessions=MaxSessions if the expected concurrency does
not change much during the application lifetime. This prevents the session pool from scaling up or down. Scaling the session pool up or down also consumes some resources.

### DIFF
--- a/internal/datastore/spanner/spanner.go
+++ b/internal/datastore/spanner/spanner.go
@@ -116,7 +116,11 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore, 
 		return nil, fmt.Errorf("failed to enable spanner GFE latency stats: %w", err)
 	}
 
-	client, err := spanner.NewClient(context.Background(), database,
+	cfg := spanner.DefaultSessionPoolConfig
+	cfg.MinOpened = config.minSessions
+	cfg.MaxOpened = config.maxSessions
+	client, err := spanner.NewClientWithConfig(context.Background(), database,
+		spanner.ClientConfig{SessionPoolConfig: cfg},
 		option.WithCredentialsFile(config.credentialsFilePath),
 		option.WithGRPCConnectionPool(max(config.readMaxOpen, config.writeMaxOpen)),
 		option.WithGRPCDialOption(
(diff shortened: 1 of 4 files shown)


internal/datastore/spanner/spanner.go
```diff
@@ -116,7 +116,16 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore,
...
 
-	client, err := spanner.NewClient(context.Background(), database,
+	client, err := spanner.NewClientWithConfig(context.Background(), database,
+		spanner.ClientConfig{
+			// obtained from DefaultSessionPoolConfig, only modified the min opened
+			SessionPoolConfig: spanner.SessionPoolConfig{
+				MinOpened:           config.minSessions,
+				MaxOpened:           config.maxSessions,
+				HealthCheckWorkers:  10,
+				HealthCheckInterval: 10 * time.Minute,
```
ecordell, line 126: The default in the spanner client is 50m - any reason for this change?

ecordell (commented):
one small question otherwise lgtm

internal/datastore/spanner/spanner.go
vroldanbet: Nope! An oversight from my side. Just pushed the change with 50 minutes. Thanks!

internal/datastore/spanner/spanner.go
```diff
@@ -116,7 +116,16 @@ func NewSpannerDatastore(database string, opts ...Option) (datastore.Datastore,
...
+	client, err := spanner.NewClientWithConfig(context.Background(), database,
+		spanner.ClientConfig{
+			// obtained from DefaultSessionPoolConfig, only modified the min opened
+			SessionPoolConfig: spanner.SessionPoolConfig{
+				MinOpened:           config.minSessions,
+				MaxOpened:           config.maxSessions,
+				HealthCheckWorkers:  10,
+				HealthCheckInterval: 50 * time.Minute,
+			},
+		},
```
jzelinskie, line 128: I think we should copy the spanner.DefaultSesssionPoolConfig and change only the values we need.

ecordell (approved):

jzelinskie (approved):

PR #1629, opened 2023-10-31
do not create custom canceled graph error

vroldanbet:
instead, use `context.Canceled`. This addresses the error being logged as an unexpected error in `graph.rewriteError()`, because it missed handling the custom cancelation error type


ecordell (approved):
LGTM

PR #1631, opened 2023-10-31
Fix flake in Postgres GC revision test by ensuring GC is run

### LINKED ISSUE #1619: Flake: TestPostgresDatastore/postgres-16-head-/TestRevisionGC
```
=== RUN   TestPostgresDatastore/postgres-16-head-/TestRevisionGC
    revisions.go:166: 
        	Error Trace:	/Users/vroldanbet/git/spicedb/pkg/datastore/test/revisions.go:166
        	            				/Users/vroldanbet/git/spicedb/pkg/datastore/test/datastore.go:104
        	Error:      	An error is expected but got nil.
        	Test:       	TestPostgresDatastore/postgres-16-head-/TestRevisionGC
        	Messages:   	expected revision head-1 to be outside GC Window
        --- FAIL: TestPostgresDatastore/postgres-16-head-/TestRevisionGC (1.13s)
```

josephschorr:
Fixes #1619

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -73,6 +73,10 @@ func RegisterGCMetrics() error {
 // GarbageCollector represents any datastore that supports external garbage
 // collection.
 type GarbageCollector interface {
+	HasGCRun() bool
+	MarkGCCompleted()
+	ResetGCCompleted()
+
 	ReadyState(context.Context) (datastore.ReadyState, error)
 	Now(context.Context) (time.Time, error)
 	TxIDBefore(context.Context, time.Time) (datastore.Revision, error)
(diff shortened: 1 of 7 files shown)


internal/datastore/common/gc.go
```diff
@@ -73,6 +73,10 @@ func RegisterGCMetrics() error {
 // GarbageCollector represents any datastore that supports external garbage
 // collection.
 type GarbageCollector interface {
+	HasGCRun() bool
+	MarkGCCompleted()
+	ResetGCCompleted()
```
vroldanbet, line 78: given this is only used in tests, can we perhaps extract it into its own interface. Then we can do type assertions in the test based on that interface. And the interface does not even need to be here, but closer to the consumer in the  test package using it

josephschorr: I prefer it being here to ensure that *all* GCable datastores are *forced* to implement it

vroldanbet: it can be forced if you make the test fail if the datastore does not implement the interface. Why making the API surface larger for something only used in a single test?

josephschorr: Because not all datastores implement the custom GC code

vroldanbet (approved):

PR #1638, opened 2023-11-03
Fix the caveat expr limit to be processed by our code

### LINKED ISSUE #1637: Caveat must appear within the first 100,000 bytes of the schema

linux

amd64

v1.26.0

### Steps to Reproduce

Run the nonexistant assertions on https://play.authzed.com/s/b1-RN001EQpW/schema

### Expected Result

✔️ Validated! after running the assertions

### Actual Result

```
invalid expression for caveat `unexpired_grant`: ERROR: unexpired_grant:-1:0: expression code point size exceeds limit: size: 106009, limit 100000
```
Screenshot 2023-11-03 at 11 08 47 AM

josephschorr:
Right now, as a hack to ensure caveat error messages have the correct line and col information, we replace the preceding schema with newlines. However, the CEL parser has a default limit on the size of the expression it can be given, which was causing the error. With this change, we turn off the CEL check and instead check the *trimmed* expression ourselves for a limit.

Long term, we'll want to issue a change in CEL to allow updating of the positioning information instead of this hack

Adds tests and fixes #1637


josephschorr:
@TristonianJones if you have any insights on how we could do this in a "supported" fashion, I'd be quite happy for the guidance

vroldanbet (approved):

TristonianJones:
> @TristonianJones if you have any insights on how we could do this in a "supported" fashion, I'd be quite happy for the guidance

The `common.Source` or maybe `ast.SourceInfo` can support a base line and column which can be used to ensure error messages include source offsets. Are you really parsing strings greater than 100KiB in length?

josephschorr:
> The `common.Source` or maybe `ast.SourceInfo` can support a base line and column which can be used to ensure error messages include source offsets
Can you point me to where I can set these? Last time I tried to do so, the error messages weren't be adjusted.

> Are you really parsing strings greater than 100KiB in length?

To ensure that the positioning is correct, we generate whitespace to mimic the position of the expression in the overall file, so yes, this can happen. The vast majority of it is just newlines.

PR #1639, opened 2023-11-03
speed up bulk import ingestion by batching writes

vroldanbet:
each individual write was causing a lot of overhead overhead on Hashicorp memdb. This cut down the import of a 130mb file from several minutes (I have up waiting) to 37 seconds


vroldanbet:
Something about bulk export is not working after the change, this needs more work

vroldanbet:
Closing, it appears the way this datastore uses `go-memdb` struggles to handle large transactions, as it has to keep all changes and eventually overflows and shifts to using a very slow algorithm. The reason this worked so fast initially was a bug in the code that mean tuples weren't being written.

PR #1640, opened 2023-11-06
Fix handling of recursive calls via singleflight dispatch

josephschorr:
Fixes the singleflight dispatcher so that if it sees the same key, for the same request ID, singleflight is bypassed to ensure that a recursion error is properly raised.

Note that this doesn't just directly raise an error in case the request ID is somehow overlapping.


internal/services/v1/permissions.go
```diff
@@ -327,6 +331,8 @@ func TranslateExpansionTree(node *core.RelationTupleTreeNode) *v1.PermissionRela
 
 func (ps *permissionServer) LookupResources(req *v1.LookupResourcesRequest, resp v1.PermissionsService_LookupResourcesServer) error {
 	ctx := resp.Context()
```
jzelinskie, line 333: You could inline this in the line below to avoid shadowing

jzelinskie (approved):
Only a minor nitpick

jzelinskie (approved):

PR #1642, opened 2023-11-09
Add a `datastore repair` command for revisions from Postgres backups

### LINKED ISSUE #1630: Fix Transaction ID counter for Postgres backups
### Problem Statement

When performing a backup of a Postgres database, Postgres does not backup/restore the transaction ID counter.
Because SpiceDB depends on this value being set, we need some means of setting the counter or incrementing it to the place where the previous database counter was set.

### Solution Brainstorm

After some investigation @jakedt was able to produce a benchmark for batching the creation of transactions to increment the Postgres transaction ID counter. With batch size of 10k its ~2.6 seconds per million transactions. This could probably use a bit more investigation to find an ideal batch size.

```go
func BenchmarkTxidPipeline(b *testing.B) {
	ctx := context.Background()

	conn, err := pgx.Connect(ctx, localhostDB)
	if err != nil {
		b.Fatalf("err: %s", err)
	}
	defer conn.Close(ctx)

	b.ResetTimer()

	// b.Logf("b.N: %d", b.N)
	for i := 0; i < b.N; {
		var batch pgx.Batch

		for j := 0; j < batchSize && i+j < b.N; j++ {
			// b.Logf("queuing query: %d", i+j)
			batch.Queue("begin;")
			batch.Queue("select pg_current_xact_id();")
			batch.Queue("rollback;")
			executed++
		}

		// b.Log("sending batch")

		i += batch.Len()

		br := conn.SendBatch(ctx, &batch)

		if err := br.Close(); err != nil {
			b.Fatalf("err: %s", err)
		}

		// b.Logf("value of i: %d", i)
	}

	b.Logf("total executed: %d", executed)
}
```
```
BenchmarkTxid-10            	     170	   6499092 ns/op
BenchmarkTxidPipeline-10    	  400071	      2705 ns/op
```

After this explored, we need to introduce this into SpiceDB.
A "repair" command could be added as maybe a part of a larger command for sanity-checking configurations (e.g. `spicedb status --repair`)

josephschorr:
Fixes #1630


pkg/cmd/datastore.go
```diff
@@ -81,3 +87,56 @@ func NewGCDatastoreCommand(programName string, cfg *datastore.Config) *cobra.Com
...
+				for _, op := range repairable.RepairOperations() {
+					fmt.Printf("\t%s: %s\n", op.Name, op.Description)
+				}
+				return nil
+			}
+
+			operationName := args[0]
+
+			log.Ctx(ctx).Info().Msg("Running repair...")
+			err = repairable.Repair(ctx, operationName, true)
```
vroldanbet, line 134: the only use in the codebase has progress bar enabled. I'd suggest remove this flag and always show the progress bar, and if folks complain, we can add a flag to the command, and then instead of passing it as an argument, we use it to instantiate a different implementation

pkg/cmd/datastore.go
```diff
@@ -81,3 +87,56 @@ func NewGCDatastoreCommand(programName string, cfg *datastore.Config) *cobra.Com
...
+				ds = wds.Unwrap()
+			}
+
+			repairable, ok := ds.(dspkg.RepairableDatastore)
+			if !ok {
+				return fmt.Errorf("datastore of type %T does not support the repair operation", ds)
+			}
+
+			if len(args) == 0 {
+				fmt.Println()
```
vroldanbet, line 123:
```suggestion
```

vroldanbet: Is this cosmetic to add a new linebreak after previous command output, or an oversight?

pkg/cmd/datastore.go
```diff
@@ -81,3 +87,56 @@ func NewGCDatastoreCommand(programName string, cfg *datastore.Config) *cobra.Com
...
+				return nil
+			}
+
+			operationName := args[0]
+
+			log.Ctx(ctx).Info().Msg("Running repair...")
+			err = repairable.Repair(ctx, operationName, true)
+			if err != nil {
+				return err
+			}
```
vroldanbet, line 137: nit
```suggestion
			}

```

internal/datastore/postgres/postgres.go
```diff
@@ -375,6 +378,98 @@ func (pgd *pgDatastore) ReadWriteTx(
...
+		return nil
+	}
+
+	var bar *progressbar.ProgressBar
+	if isatty.IsTerminal(os.Stderr.Fd()) && outputProgress {
+		bar = progressbar.Default(int64(counterDelta), "updating transactions counter")
+	}
+
+	for i := 0; i < counterDelta; i++ {
+		var batch pgx.Batch
```
vroldanbet, line 431: Have we determined how much impact does this loop have over a production postgres? I think we should throttle it.

internal/datastore/postgres/postgres.go
```diff
@@ -375,6 +378,98 @@ func (pgd *pgDatastore) ReadWriteTx(
...
+	switch operationName {
+	case repairTransactionIDsOperation:
+		return pgd.repairTransactionIDs(ctx, outputProgress)
+
+	default:
+		return fmt.Errorf("unknown operation")
+	}
+}
+
+const batchSize = 10000
```
vroldanbet, line 393: do we know the impact of such batch size on a production postgres? should this be at a more conservative value and expose a flag?

pkg/cmd/datastore.go
josephschorr: Its used in tests to not show things

pkg/cmd/datastore.go
josephschorr: Its cometic

internal/datastore/postgres/revisions.go
```diff
@@ -67,6 +67,9 @@ const (
 	SELECT minvalid.%[1]s, minvalid.%[5]s, pg_current_snapshot() FROM minvalid;`
 
 	queryCurrentSnapshot = `SELECT pg_current_snapshot();`
+
+	queryCurrentTransactionID = `SELECT pg_current_xact_id()::text::integer;`
+	queryLatestXID            = `SELECT max(xid) FROM relation_tuple_transaction;`
```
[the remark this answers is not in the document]
josephschorr: Preflight isn't really necessary, since it'll just show an error

pkg/cmd/datastore.go
```diff
@@ -81,3 +87,56 @@ func NewGCDatastoreCommand(programName string, cfg *datastore.Config) *cobra.Com
...
 	}
 }
+
+func NewRepairDatastoreCommand(programName string, cfg *datastore.Config) *cobra.Command {
+	return &cobra.Command{
+		Use:     "repair",
+		Short:   "executes datastore repair",
+		Long:    "Executes a repair operation for the datastore",
+		PreRunE: server.DefaultPreRunE(programName),
+		RunE: termination.PublishError(func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
josephschorr: For... printing and invoking a function?

internal/datastore/postgres/postgres.go
josephschorr: It shouldn't matter; it should only be run while SpiceDB is offline

pkg/cmd/datastore.go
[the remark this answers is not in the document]
vroldanbet: the `cmd` package does not have good test coverage, and I observed regressions specifically around configuration change. This function does not have configuration going on. I still think that commands deserve test coverage too, but I leave that up to you

internal/datastore/postgres/postgres.go
vroldanbet: Have that been made clear anywhere in the command description?

vroldanbet (dismissed):

vroldanbet (dismissed):

vroldanbet (approved):

vroldanbet (approved):

PR #1643, opened 2023-11-10
fix dispatch deduplication regression

vroldanbet:
regressed in https://github.com/authzed/spicedb/pull/1640

The previous approach caused an important regression in the deduplication rate because concurrent requests that are part of the same `requestID` will trigger the condition, but do not necessarily involve a loop. We observed production systems that deduplicated at a rate of 12% drop to 2%.

This approach uses a bloom filter to keep track of what parts of the schema have been traversed. If there is a false positive, the deduper will simply dispatch, so the loss in requests that could have been deduplicated is proportional to the false-positive probability configured.

The number of elements in the bloom filter is set to the `--dispatch-max-depth` configuration parameter. By default this value is set to `50`. This sets an upper boundary to the number of elements added to the bloom filter before the recursive-call limit error is returned. We know there will not be more than `--dispatch-max-depth` nested dispatch calls.

The size of a 50-element bloom filter with a 0.1% false-positive rate is 120 bytes ([90 according to the calculator](https://hur.st/bloomfilter/?n=50&p=0.001&m=&k=)). This is a reasonable tradeoff, especially given Snappy compression is enabled by default on the dispatch layer. We could increase it further and would still be reasonably sized.

The `RequestID` metadata in `ResolverMeta` was no longer needed because the dispatcher became stateless with this change, and the bloom filter is always request-specific. I left the field in the struct proto so as not to break Dispatch API as folks roll out the next SpiceDB version.

Tests were refactored to make them more readable, and a new test was added to simulate a loop.


internal/dispatch/singleflight/singleflight.go
```diff
@@ -53,19 +55,25 @@ func (d *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckReq
...
+	if errors.Is(err, ErrLoopDetect) {
+		singleFlightCount.WithLabelValues("DispatchCheck", "loop").Inc()
 		return d.delegate.DispatchCheck(ctx, req)
 	}
 
+	if err != nil {
+		return &v1.DispatchCheckResponse{Metadata: &v1.ResponseMeta{DispatchCount: 1}}, err
+	}
+
+	clonedReq := req.CloneVT()
```
josephschorr, line 72: This is going to be very inefficient - we should endeavor not to clone the response

internal/dispatch/singleflight/singleflight.go
```diff
@@ -84,28 +92,69 @@ func (d *Dispatcher) DispatchExpand(ctx context.Context, req *v1.DispatchExpandR
...
+	bf = bf.AddString(key)
+	modifiedBloomFilter, err := bf.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+
+	return modifiedBloomFilter, nil
+}
+
+func MustNewTraversalBloomFilter(depth uint) []byte {
```
josephschorr, line 147: doc comment

internal/graph/checkingresourcestream.go
```diff
@@ -455,12 +456,13 @@ func (crs *checkingResourceStream) runProcess(alwaysProcess bool) (bool, error)
...
-			AtRevision:    crs.req.Revision,
-			MaximumDepth:  crs.req.Metadata.DepthRemaining,
-			DebugOption:   computed.NoDebugging,
+			ResourceType:         crs.req.ObjectRelation,
+			Subject:              crs.req.Subject,
+			CaveatContext:        crs.req.Context.AsMap(),
+			AtRevision:           crs.req.Revision,
+			MaximumDepth:         crs.req.Metadata.DepthRemaining,
+			DebugOption:          computed.NoDebugging,
+			TraversalBloomFilter: singleflight.MustNewTraversalBloomFilter(uint(crs.req.Metadata.DepthRemaining)),
```
josephschorr, line 465: Don't use a `Must` call here; make it return a spicerrors.MustBugf and then return it here so it doesn't panic in prod

internal/services/v1/permissions.go
```diff
@@ -156,13 +157,11 @@ func (ps *permissionServer) ExpandPermissionTree(ctx context.Context, req *v1.Ex
...
 	}
 
-	requestID, ctx := requestid.GetOrGenerateRequestID(ctx)
-
 	resp, err := ps.dispatch.DispatchExpand(ctx, &dispatch.DispatchExpandRequest{
 		Metadata: &dispatch.ResolverMeta{
 			AtRevision:     atRevision.String(),
 			DepthRemaining: ps.config.MaximumAPIDepth,
-			RequestId:      requestID,
+			TraversalBloom: singleflight.MustNewTraversalBloomFilter(uint(ps.config.MaximumAPIDepth)),
```
josephschorr, line 164: ditto

internal/dispatch/singleflight/singleflight.go
vroldanbet: I can address this but I have concerns about it. The moment the same dispatch request is reused by two dispatch calls, it'd be leaking the state of the bloom filter. The worst that can happen is that singleflight will by bypassed, which is not terrible, mind you, but it would exhibit that behaviour and be difficult to understand why.

internal/dispatch/singleflight/singleflight_test.go
```diff
@@ -255,7 +255,7 @@ func assertCounterWithLabel(t *testing.T, gatherer prometheus.Gatherer, expected
 func bloomFilterForRequest(t *testing.T, keyHandler *keys.DirectKeyHandler, req *v1.DispatchCheckRequest) []byte {
 	t.Helper()
 
-	bloomFilter := bloom.NewWithEstimates(defaultBloomFilterSize, 0.01)
+	bloomFilter := bloom.NewWithEstimates(defaultBloomFilterSize, 0.001)
```
josephschorr, line 258: use `defaultFalsePositiveRate` here?

internal/dispatch/singleflight/singleflight.go
```diff
@@ -3,8 +3,11 @@ package singleflight
 import (
 	"context"
 	"encoding/hex"
+	"errors"
+	"fmt"
 	"strconv"
 
+	"github.com/bits-and-blooms/bloom/v3"
```
jzelinskie, line 10: Is this library better than https://github.com/tylertreat/BoomFilters which has been around for years and has many different types of probabilistic data structures?

internal/dispatch/singleflight/singleflight_test.go
vroldanbet: I don't think it's necessary, and it would import the `pkg/graph` package here, and it would need to be exported.

This package does not care about the size nor the probability of the bloom filter - it's specified externally by whoever initializes sets the `ResolverMeta`. The only contract is that it needs to use the same library.

So from the perspective of the tests, this could be any probability (we probably want to have it low so it does not flake).

internal/dispatch/singleflight/singleflight.go
vroldanbet: @jzelinskie I looked into that library, it's nice it has a lot of more data structures than just the classic bloom filter. From the API they are basically on par, but bits-and-blooms provides convenience AddString methods that spares you the type-casting.

From the performance point of view, bits-and-blooms seems significantly better, both time and space wise. I created 2 benchmarks that take 50 random strings, adds them, tests them and then serializes the filter, both on binary and in Gob serialization. This is what I got using `benchstat`:

```
benchstat bits-and-blooms.txt boomfilters.txt         
goos: darwin
goarch: arm64
pkg: github.com/authzed/spicedb/internal/dispatch/singleflight
                       │ bits-and-blooms.txt │           boomfilters.txt           │
                       │       sec/op        │   sec/op     vs base                │
AddTestSerialize-12              4.415µ ± 0%   5.544µ ± 0%  +25.59% (p=0.000 n=10)
AddTestSerializeGob-12           4.407µ ± 0%   5.562µ ± 0%  +26.20% (p=0.000 n=10)
geomean                          4.411µ        5.553µ       +25.89%

                       │ bits-and-blooms.txt │           boomfilters.txt            │
                       │        B/op         │     B/op      vs base                │
AddTestSerialize-12             1.406Ki ± 0%   1.586Ki ± 0%  +12.78% (p=0.000 n=10)
AddTestSerializeGob-12          1.406Ki ± 0%   1.586Ki ± 0%  +12.78% (p=0.000 n=10)
geomean                         1.406Ki        1.586Ki       +12.78%

                       │ bits-and-blooms.txt │            boomfilters.txt            │
                       │      allocs/op      │  allocs/op   vs base                  │
AddTestSerialize-12               10.00 ± 0%   123.00 ± 0%  +1130.00% (p=0.000 n=10)
AddTestSerializeGob-12            10.00 ± 0%   123.00 ± 0%  +1130.00% (p=0.000 n=10)
geomean                           10.00         123.0       +1130.00%
```

vroldanbet: For reference, the tests:

# bits-and-blooms

```go
package singleflight

import (
	"github.com/bits-and-blooms/bloom/v3"
	"testing"
)

var s = []string{
	"3#JQ9s7@",
	"f^&4L6$K",
	"2%hN5*G@",
	"p$#8q9!W",
	"1@L6t#E$",
	"m^&7P5!S",
	"4%kN3*H@",
	"j$#2r9@Q",
	"5@F6y#D$",
	"n^&1T7!U",
	"6%gM4*J@",
	"i$#3o9@R",
	"7@B5u#C$",
	"l^&2V6!X",
	"8%aK1*Z@",
	"h$#4m9@S",
	"9@G7x#Y$",
	"o^&5I3!N",
	"0%bL2*F@",
	"g$#6n9@T",
	"a@E8w#X$",
	"k^&4O1!M",
	"b%jP3*K@",
	"e$#7q9@U",
	"c@D5v#Z$",
	"d^&8R2!H",
	"f%tN4*L@",
	"i$#9s7@Q",
	"g@J6y#E$",
	"h^&5T3!U",
	"j%oM1*K@",
	"k$#8p9@R",
	"l@B7u#F$",
	"m^&2V5!X",
	"n%lK3*Z@",
	"o$#1m9@S",
	"p@G4x#C$",
	"q^&6I2!N",
	"r%bL4*F@",
	"s$#3n9@T",
	"t@E5w#Y$",
	"u^&7O1!M",
	"v%jP2*K@",
	"w$#4q9@U",
	"x@D6v#Z$",
	"y^&9R3!H",
	"z%tN5*L@",
}

func BenchmarkAddTestSerialize(b *testing.B) {
	for i := 0; i < b.N; i++ {
		bloomFilter := bloom.NewWithEstimates(defaultBloomFilterSize, 0.001)
		for _, randStr := range s {
			bloomFilter = bloomFilter.AddString(randStr)
		}
		for _, randStr := range s {
			bloomFilter.TestString(randStr)
		}
		_, _ = bloomFilter.MarshalBinary()
	}
}

func BenchmarkAddTestSerializeGob(b *testing.B) {
	for i := 0; i < b.N; i++ {
		bloomFilter := bloom.NewWithEstimates(defaultBloomFilterSize, 0.001)
		for _, randStr := range s {
			bloomFilter = bloomFilter.AddString(randStr)
		}
		for _, randStr := range s {
			bloomFilter.TestString(randStr)
		}
		_, _ = bloomFilter.GobEncode()
	}
}
```

# boomfilters

```go
package singleflight

import (
	"bytes"
	boom "github.com/tylertreat/BoomFilters"
	"testing"
)

var s = []string{
	"3#JQ9s7@",
	"f^&4L6$K",
	"2%hN5*G@",
	"p$#8q9!W",
	"1@L6t#E$",
	"m^&7P5!S",
	"4%kN3*H@",
	"j$#2r9@Q",
	"5@F6y#D$",
	"n^&1T7!U",
	"6%gM4*J@",
	"i$#3o9@R",
	"7@B5u#C$",
	"l^&2V6!X",
	"8%aK1*Z@",
	"h$#4m9@S",
	"9@G7x#Y$",
	"o^&5I3!N",
	"0%bL2*F@",
	"g$#6n9@T",
	"a@E8w#X$",
	"k^&4O1!M",
	"b%jP3*K@",
	"e$#7q9@U",
	"c@D5v#Z$",
	"d^&8R2!H",
	"f%tN4*L@",
	"i$#9s7@Q",
	"g@J6y#E$",
	"h^&5T3!U",
	"j%oM1*K@",
	"k$#8p9@R",
	"l@B7u#F$",
	"m^&2V5!X",
	"n%lK3*Z@",
	"o$#1m9@S",
	"p@G4x#C$",
	"q^&6I2!N",
	"r%bL4*F@",
	"s$#3n9@T",
	"t@E5w#Y$",
	"u^&7O1!M",
	"v%jP2*K@",
	"w$#4q9@U",
	"x@D6v#Z$",
	"y^&9R3!H",
	"z%tN5*L@",
}

func BenchmarkAddTestSerialize(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var buf bytes.Buffer
		bf := boom.NewUnstableBloomFilter(defaultBloomFilterSize, 0.001)
		for _, randStr := range s {
			bf.Add([]byte(randStr))
		}
		for _, randStr := range s {
			bf.Test([]byte(randStr))
		}
		_, _ = bf.WriteTo(&buf)
	}
}

func BenchmarkAddTestSerializeGob(b *testing.B) {
	for i := 0; i < b.N; i++ {
		bf := boom.NewUnstableBloomFilter(defaultBloomFilterSize, 0.001)
		for _, randStr := range s {
			bf.Add([]byte(randStr))
		}
		for _, randStr := range s {
			bf.Test([]byte(randStr))
		}
		_, _ = bf.GobEncode()
	}
}
```

pkg/graph/resolvermeta/resolvermeta.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/bits-and-blooms/bloom/v3"
+
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+const defaultFalsePositiveRate = 0.001
+
+// NewTraversalBloomFilter creates a new bloom filter sized to the provided number of elements and
+// with a predefined false-positive ratio of 1%.
```
jzelinskie, line 12: this is no longer 1% but 0.1%

internal/graph/reachableresources.go
```diff
@@ -549,7 +549,6 @@ func (crr *CursoredReachableResources) redispatchOrReport(
 				Metadata: &v1.ResolverMeta{
 					AtRevision:     parentRequest.Revision.String(),
 					DepthRemaining: parentRequest.Metadata.DepthRemaining - 1,
-					RequestId:      parentRequest.Metadata.RequestId,
```
jzelinskie, line 552: In the places where we're just removing request ID, why didn't we also add the bloom filter?
Isn't it possible that reachable resources could also fall into a loop?

vroldanbet: Singleflight dispatcher is only used on check and expand, so I think my thought was to only add this where needed.

internal/dispatch/singleflight/singleflight.go
```diff
@@ -84,28 +89,57 @@ func (d *Dispatcher) DispatchExpand(ctx context.Context, req *v1.DispatchExpandR
...
 	}
 	return v, err
 }
 
+var ErrLoopDetect = errors.New("traversal recursion loop detected")
+
+// validateTraversalRecursion determines from the ResolverMeta.TraversalBloom value if a traversal loop has happened
+// based on the argument key. As a bloom filter, this may lead to false positives,
+// so clients should take that into account.
+func validateTraversalRecursion(key string, meta *v1.ResolverMeta) ([]byte, error) {
```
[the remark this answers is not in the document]
vroldanbet: good suggestion! addressed in https://github.com/authzed/spicedb/pull/1643/commits/a17badf3fd9c8cff8edcab24676e13135a01973e

jzelinskie (approved):

PR #1650, opened 2023-11-15
singleflight dispatcher: do not double-singleflight on remote cluster dispatch

vroldanbet:
This PR includes some further improvements to the single-flight dispatcher.
The most important bit is the removal of it from the cluster dispatcher, which was causing keys to be added to the bloom filter on the egressing dispatcher and in the ingressing dispatcher. As a consequence, the false positive rate appeared abnormally high.

### DIFF
--- a/pkg/proto/dispatch/v1/02_resolvermeta.go
+++ b/pkg/proto/dispatch/v1/02_resolvermeta.go
@@ -12,7 +12,7 @@ import (
 
 func (x *ResolverMeta) RecordTraversal(key string) (possiblyLoop bool, err error) {
 	if key == "" {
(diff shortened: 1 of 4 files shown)


pkg/proto/dispatch/v1/02_resolvermeta.go
```diff
@@ -11,6 +11,10 @@ import (
 )
 
 func (x *ResolverMeta) RecordTraversal(key string) (possiblyLoop bool, err error) {
+	if key == "" {
+		return false, status.Error(codes.Internal, fmt.Errorf("missing key to be recorded in traversal").Error())
```
josephschorr, line 15: this should be a `spiceerrors.MustBugf`

josephschorr (approved):
LGTM

PR #1657, opened 2023-11-17
crdb: don't allow relationship counters to go negative

ecordell:
There is a problem when updating the counters that causes the total to possible go negative when reporting the counts, which panics when reading into a `uint`. This PR is a stop-gap to prevent the panics; a future change will fix how reporting works in CRDB to report correct counts.

### DIFF
--- a/internal/datastore/crdb/stats.go
+++ b/internal/datastore/crdb/stats.go
@@ -52,7 +52,7 @@ func (cds *crdbDatastore) Statistics(ctx context.Context) (datastore.Stats, erro
 	}
 
 	var nsDefs []datastore.RevisionedNamespace
-	var relCount uint64
+	var relCount int64
 
 	if err := cds.readPool.QueryRowFunc(ctx, func(ctx context.Context, row pgx.Row) error {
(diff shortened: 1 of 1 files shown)


internal/datastore/crdb/stats.go
```diff
@@ -26,7 +26,7 @@ const (
 
 var (
 	queryReadUniqueID         = psql.Select(colUniqueID).From(tableMetadata)
-	queryRelationshipEstimate = fmt.Sprintf("SELECT COALESCE(SUM(%s), 0) FROM %s AS OF SYSTEM TIME follower_read_timestamp()", colCount, tableCounters)
+	queryRelationshipEstimate = fmt.Sprintf("SELECT array_upper(ARRAY[COALESCE(SUM(%s), 0), 0], 1) FROM %s AS OF SYSTEM TIME follower_read_timestamp()", colCount, tableCounters)
```
josephschorr, line 29: since we know why these can go negative, maybe just read them into a signed int and `min` check them on the Go side?

josephschorr (approved):
LGTM

Issue #1659, opened 2023-11-27
Garbage collector interval not reset after transient error recovery


chriskdon:
### What platforms are affected?

linux, macos, windows

### What architectures are affected?

amd64, arm64

### What SpiceDB version are you using?

v1.27.0

### Steps to Reproduce

1. Set the `--datastore-gc-interval=5m`
   - This value doesn't strictly matter for this issue. It's used to make the explanation below more concrete.
2. Cause an error in the garbage collector (GC) (e.g. a deadlock wait timeout)
3. This causes the backoff in the GC to trigger, which sets `nextInterval = backoffInterval.NextBackOff()`
   - https://github.com/authzed/spicedb/blob/57bf05e1c63d8df16ccf571ce90200f9ce92fc1d/internal/datastore/common/gc.go#L138-L138

5. Let the GC recover and stop erroring
6. Notice that the `interval` time in the log will be stuck at the last backoff interval before recovery. The GC will either be running much faster or slower than expected.

The issue appears to be that `nextInterval` is not reset when the backoff is reset:
https://github.com/authzed/spicedb/blob/57bf05e1c63d8df16ccf571ce90200f9ce92fc1d/internal/datastore/common/gc.go#L144

### Expected Result

Once the error stops and the GC recovers, the interval should be reset to the configured 5 minutes. The garbage collector should continue to run at the configured interval.

`nextInterval` needs to be set to the configured interval:

```go
backoffInterval.Reset()
nextInterval = interval
```

I've made a PR in my fork to show the issue/fix better with a test: https://github.com/chriskdon/spicedb/pull/1. I'm happy to open a PR here if you'd like. 

### Actual Result

Once the transient GC error stops, the `interval` is not reset. This has the effect of causing the GC to run too fast or too slow, depending on whatever the last backoff interval was when it recovered.

PR #1663, opened 2023-11-28
Fix garbage collector interval and backoff resets

chriskdon:
Fixes https://github.com/authzed/spicedb/issues/1659

Resets `nextInterval` back to the configured interval when `backoffInterval.Reset()` is called on a successful garbage collection run.

It also fixes a different problem I discovered while writing tests. `backoffInterval.Reset()` must be called after initialization according to [these](https://pkg.go.dev/github.com/cenkalti/backoff/v4#ExponentialBackOff.Reset) docs:
> Reset the interval back to the initial retry interval and restarts the timer. Reset must be called before using b.

When this is not done, the `InitialInterval` value is ignored, and the default `500ms` is used until the first GC iteration completes successfully and `backoffInterval.Reset()` is called. I assume this is unintentional behaviour? 

-------

_I haven't written much Go, so please feel free to nitpick the PR. I'm open to any suggestions._ 😅

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -141,7 +141,10 @@ func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, gc GarbageColl
 					Msg("error attempting to perform garbage collection")
 				continue
 			}
+
 			backoffInterval.Reset()
+			nextInterval = interval
+
 			log.Ctx(ctx).Debug().
 				Dur("next-run-in", interval).
 				Msg("datastore garbage collection scheduled for next run")
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -111,6 +111,7 @@ func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, gc GarbageColl
 	backoffInterval.InitialInterval = interval
 	backoffInterval.MaxInterval = max(MaxGCInterval, interval)
 	backoffInterval.MaxElapsedTime = maxElapsedTime
+	backoffInterval.Reset()
 
 	nextInterval := interval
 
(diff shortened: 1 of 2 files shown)


internal/datastore/common/gc.go
```diff
@@ -111,6 +111,7 @@ func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, gc GarbageColl
 	backoffInterval.InitialInterval = interval
 	backoffInterval.MaxInterval = max(MaxGCInterval, interval)
 	backoffInterval.MaxElapsedTime = maxElapsedTime
+	backoffInterval.Reset()
```
chriskdon, line 114: This is the other issue I discovered during testing. Without this, the initial interval used by the backoff will be the default `500ms` until `backoffInterval.Reset()` is called. This is an edge case, but it can cause the wrong initial value to be used if the first iteration of the GC has an error.

internal/datastore/common/gc_test.go
```diff
@@ -3,42 +3,126 @@ package common
...
 	"github.com/prometheus/client_golang/prometheus"
 	promclient "github.com/prometheus/client_model/go"
+	"github.com/shopspring/decimal"
 	"github.com/stretchr/testify/require"
 )
 
-type testGC struct{}
+// Fake garbage collector that returns a new incremented revision each time
+// TxIDBefore is called.
+type fakeGC struct {
```
chriskdon, line 22: I updated the `testGC` to be more fake-like so it could be used to test different scenarios. In my case, I need it to intermittently fail and then recover. If you'd prefer a special, more straightforward GC implementation just for the new test, I'm happy to change it.

chriskdon:

internal/datastore/common/gc_test.go
vroldanbet: I think this is great 👍🏻

vroldanbet (approved):
👏🏻 good find, great contribution and excellent PR description and commit history. LGTM!

PR #1676, opened 2023-12-02
Respect dispatch concurrency limits for clusterdispatch

### LINKED ISSUE #1641: Respect concurrency limit settings for cluster dispatch

linux, macos, windows, others

amd64, arm64, others

v1.26.0

### Steps to Reproduce

1) launch spicedb with --dispatch-concurrency-limit=1 with debugger
2) send CheckPermission request

### Expected Result

After check is dispatched for the first time and cluster dispatcher is serving the request, I can see dispatchAllAsync (internal/graph/check.go:631) is started with limit 1 according to my settings.

### Actual Result

After check is dispatched for the first time and cluster dispatcher is serving the request, I can see dispatchAllAsync is started with default limit 50 (internal/graph/check.go:631) despite of the settings.

### Steps to fix

Function internal/dispatch/cluster/cluster.go:46 should be called during initial spicedb setup.

sashayakovtseva:
Fixes #1641

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -237,6 +237,9 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 
 	enableGRPCHistogram()
 
+	specificConcurrencyLimits := c.DispatchConcurrencyLimits
+	concurrencyLimits := specificConcurrencyLimits.WithOverallDefaultLimit(c.GlobalDispatchConcurrencyLimit)
+
 	dispatcher := c.Dispatcher
 	if dispatcher == nil {
 		cc, err := c.DispatchCacheConfig.WithRevisionParameters(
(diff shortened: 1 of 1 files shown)


sashayakovtseva:

jzelinskie:
I believe @vroldanbet might have discovered an issue with this approach. I'll let him chime in whenever he's available (it is the holiday season).

sashayakovtseva:
Hello,

any news here? what issue have you discovered?

vroldanbet (approved):
Hey @sashayakovtseva, my apologies for not getting back to this earlier.

There isn't so much of an issue but a concern around the impact of this change. While the unbounded dispatch configuration is not great for the health of the cluster, it does as well fundamentally change the baseline throughput of existing clusters, which may see a regression when this change is deployed in a non-trivial way, until they figure out that this change caps the dispatch concurrency.

I don't have an alternative, to be honest, other than rolling this out and observing the actual impact in practice, and tuning accordingly when required.

sashayakovtseva:
> While the unbounded dispatch configuration is not great for the health of the cluster, it does as well fundamentally change the baseline throughput of existing clusters

I should mention that currently dispatch is bounded with the default value of 50. So unless one doesn't explicitly set dispatch limit < 50 no regression should be observed

PR #1679, opened 2023-12-07
schemadsl/compiler: optionally skip validation

jzelinskie:
This allows the schema compiler to skip the validation step, which is useful for plumbing in zed for backup & restore where the server is already going to be doing validation.


internal/graph/computed/computecheck_test.go
```diff
@@ -938,7 +938,7 @@ func writeCaveatedTuples(ctx context.Context, _ *testing.T, ds datastore.Datasto
 	compiled, err := compiler.Compile(compiler.InputSchema{
 		Source:       "schema",
 		SchemaString: schema,
-	}, &empty)
+	}, &empty, false)
```
josephschorr, line 941: Make this into an options so we can add other stuff to it down the road

josephschorr (approved):
LGTM

PR #1683, opened 2023-12-12
fixes regression in compiler.Compile() contract

vroldanbet:
https://github.com/authzed/spicedb/pull/1679 introduced a flag to loosen validation of the schema definition.

This PR introduced a subtle but breaking change in the contract of the compile.Compile() method:
- before `Compile()` would fail if no object type prefix was defined _by default_. It supported compiling without prefix by setting the expected object prefix to "empty string", but this required the client to set that up explicitly.
- with the recent changes, `compile.Compile()` would by default allow schemas without prefix

The new default seems _unsafe_ compared to the original behavior.

This commit reverts that change and introduces a new Option "AllowUnprefixedObjectType", so that they are forced to opt in if they want to ignore prefixes. It's more verbose but safer.

- tests to explicitly state this formerly implicit contract were added
- tests for `SkipValidation` were also added


josephschorr (approved):
LGTM

PR #1684, opened 2023-12-13
fixes regression in Schema Watch led schema caching

vroldanbet:
Initialization of schema watch happens via the `UnwrappableDatastore` interface. Given the various proxies built around the `datastore.Datastore` interface, we came up with the notion of `UnwrappableDatastore` to be able to extract a particular type from the proxy chain.

The way this is correctly done is by unwrapping in a loop until the type is found. This was unfortunately not done in the Schema Watch initialization logic, which assumed the position of the actual underlying datastore at a specific level.

With the introduction of the datastore deduplication proxy in https://github.com/authzed/spicedb/pull/1610, the proxy chain changed, and the assumed order became incorrect, making the initialization logic unable to find an instance of the `SchemaWatchableDatastore`, even though it was there. As a consequence, the logic fell back to JIT caching strategy. 

This means schema watch-led caching is broken since the 1.27.0 release, but only for CRDB users: The Spanner datastore is not subject to this bug because the CRDB datastore has one extra proxy layer and the code as it was initially was able to check for the target interface in the immediate level and the next unwrapped level. 

This replaces every bespoke piece of code handling the unwrapping logic with a new function that does this correctly and more conveniently, sparing duplicate code and being more concise.

### DIFF
--- a/internal/datastore/postgres/postgres_shared_test.go
+++ b/internal/datastore/postgres/postgres_shared_test.go
@@ -1372,16 +1372,8 @@ func GCQueriesServedByExpectedIndexes(t *testing.T, _ testdatastore.RunningEngin
 	revision, err := ds.HeadRevision(ctx)
 	require.NoError(err)
 
-	for {
-		wds, ok := ds.(datastore.UnwrappableDatastore)
-		if !ok {
-			break
-		}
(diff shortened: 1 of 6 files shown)


josephschorr (approved):
LGTM

PR #1685, opened 2023-12-13
Add an integration test for schema watch

josephschorr:


vroldanbet (commented):
@josephschorr changes look good to me, although it does not seem like there is an assertion for actual schema watch logs

cmd/spicedb/schemawatch_integration_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	require.Equal(t, len(engines), len(datastore.Engines))
+
+	for driverName, shouldRun := range engines {
+		if !shouldRun {
+			continue
+		}
+
+		t.Run(driverName, func(t *testing.T) {
+			bridgeNetworkName := fmt.Sprintf("bridge-%s", uuid.New().String())
```
vroldanbet, line 36: seem like expensive tests, let's parallelize them

```suggestion
			t.Parallel()
			
			bridgeNetworkName := fmt.Sprintf("bridge-%s", uuid.New().String())
```

cmd/spicedb/schemawatch_integration_test.go
```diff
@@ -0,0 +1,127 @@
...
+	testdatastore "github.com/authzed/spicedb/internal/testserver/datastore"
+	"github.com/authzed/spicedb/pkg/datastore"
+)
+
+func TestSchemaWatch(t *testing.T) {
+	engines := map[string]bool{
+		"postgres":    false,
+		"mysql":       false,
+		"cockroachdb": true,
+		"spanner":     false,
```
vroldanbet, line 26: did you mean to disable postgres and spanner?

cmd/spicedb/schemawatch_integration_test.go
josephschorr: I don't like running datastore container tests in parallel; things to get wonky when we do that

cmd/spicedb/schemawatch_integration_test.go
josephschorr: Yes

vroldanbet (approved):

PR #1688, opened 2023-12-19
fixes WriteRelationships/BulkImport GRPC error codes on conflicts and retryable errors

vroldanbet:
Built to support https://github.com/authzed/zed/pull/316

Adjusts BulkImport so it returns gRPC codes to let clients know if it's ok to retry.
- duplicate relationship now reports gRPC code `AlreadyExists`
- retriable errors now return grpc code `Unavailable`

### DIFF
--- a/internal/datastore/crdb/pool/slqerrors.go
+++ b/internal/datastore/crdb/pool/slqerrors.go
@@ -5,6 +5,10 @@ import (
 	"fmt"
 
 	"github.com/jackc/pgx/v5/pgconn"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


internal/datastore/crdb/pool/slqerrors.go
```diff
@@ -43,6 +57,17 @@ type ResettableError struct {
 func (e *ResettableError) Error() string { return "resettable error" + ": " + e.Err.Error() }
 func (e *ResettableError) Unwrap() error { return e.Err }
 
+func (e *ResettableError) GRPCStatus() *status.Status {
+	return spiceerrors.WithCodeAndDetails(
+		e.Unwrap(),
+		codes.Unavailable, // return unavailable so clients are know it's ok to retry
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_UNSPECIFIED, // Since this wraps another error, we don't know the type
```
josephschorr, line 65: Maybe we don't include the reason?

josephschorr (approved):
LGTM

PR #1689, opened 2023-12-19
Ignore the source position when diffing permission expressions

josephschorr:


pkg/diff/namespace/diff.go
```diff
@@ -301,5 +302,12 @@ func isPermission(relation *core.Relation) bool {
 }
 
 func areDifferentExpressions(existing *core.UsersetRewrite, updated *core.UsersetRewrite) bool {
```
vroldanbet, line 304: if this changes namespace writes will not take place in `WriteSchema` API. Is this what users expect? They may expect the schema stored to be exactly the same thing as was previously written.

josephschorr: This is the correct implementation and this PR fixes a bug that was otherwise causing the namespace to be written when it should not have been. We don't store the exact schema either: it is rewritten on the way back out into a string

vroldanbet: so you say the schema is normalized before getting written, and this makes it so that it compares the _normalized_ schema, not just the input schema?

josephschorr: No, it compares the input schema but position information is not relevant for diffing

PR #1692, opened 2023-12-20
internal/datastore/spanner: force subject ID index

### LINKED ISSUE #1687: SpiceDB Spanner queries use different+worse index than e.g. same query in Spanner Studio

others

others

1.25.0 - 1.27.0

Not sure if this issue is present in earlier versions as well.

### Steps to Reproduce

**TLDR:**
- Use Spanner Datastore
- Run a LookupResources call
- Compare query performance and query plan from SpiceDB issued Spanner queries vs. ones ran in e.g. Spanner Studio
- Observe significant performance deficit

**Details**:
A LookupResource will issue multiple SQL queries targeting the Spanner datastore that are very similar, but will use different indices for the lookup, which depending on the amount of relations present will result in large performance degradations.

First query, using the `ix_relation_tuple_by_subject` index and returning in <20ms:
```sql
SELECT
  namespace,
  object_id,
  relation,
  userset_namespace,
  userset_object_id,
  userset_relation,
  caveat_name,
  caveat_context
FROM
  relation_tuple
WHERE
  ((userset_namespace = '<USERSET_NAMESPACE>'
      AND userset_object_id IN (('<OBJECT_ID>'))
      AND userset_relation = '...'))
  AND namespace = '<NAMESPACE>'
  AND relation = '<RELATION>'
ORDER BY
  userset_namespace,
  userset_object_id,
  userset_relation,
  namespace,
  object_id,
  relation
LIMIT
  9223372036854775807
```
Later query, using `ix_relation_tuple_by_subject_relation` index, returns in 2000-4000ms:
```sql
SELECT
  namespace,
  object_id,
  relation,
  userset_namespace,
  userset_object_id,
  userset_relation,
  caveat_name,
  caveat_context
FROM
  relation_tuple
WHERE
  ((userset_namespace = '<USERSET_NAMESPACE>'
      AND userset_object_id IN (('<OBJECT_ID>'))
      AND userset_relation = '...'))
  AND namespace = '<NAMESPACE>'
  AND relation = '<RELATION>'
  AND ((object_id > '<SOME OBJECT ID>'))
ORDER BY
  userset_namespace,
  userset_object_id,
  userset_relation,
  namespace,
  object_id,
  relation
LIMIT
  9223372036854775807
```

This only occurs with queries issued by SpiceDB. Running the EXACT SAME query inside e.g. Spanner Studio will always use the correct index and return in <20ms as well, yielding the same rows.

This issue was investigated by adding the `QueryWithStats()` Spanner call in `spanner.go` and verifying the planning and execution times as well as which indices where used.

The used `Query optimizer version` and `Statistics package` are the same between the queries in Spanner Studio and issued by SpiceDB.

### Expected Result

SpiceDB issued Spanner calls use the correct/fastest index for SQL queries

### Actual Result

`ix_relation_tuple_by_subject_relation` is used instead of `ix_relation_tuple_by_subject`, leading to a severe performance degradation.

We have stopped using `LookupResources` calls for now for cases with many relations, as the performance penalty was too severe.

jzelinskie:
Fixes #1687.

### DIFF
--- a/internal/datastore/spanner/reader.go
+++ b/internal/datastore/spanner/reader.go
@@ -28,7 +28,7 @@ func forceIndex(b sq.SelectBuilder, setColumns []string) (sq.SelectBuilder, erro
 	if slices.Contains(setColumns, colUsersetObjectID) && !slices.Contains(setColumns, colObjectID) {
 		// When using the Postgres dialect, Spanner uses magic comments to
 		// force indices.
-		b = b.From(tableRelationship + " /*@ FORCE_INDEX=idx_relation_tuple_by_subject */")
+		b = b.From(tableRelationship + " /*@ FORCE_INDEX = ix_relation_tuple_by_subject */")
 	}
 	return b, nil
 }
--- a/internal/datastore/spanner/reader.go
+++ b/internal/datastore/spanner/reader.go
@@ -6,8 +6,10 @@ import (
 	"time"
 
 	"cloud.google.com/go/spanner"
+	sq "github.com/Masterminds/squirrel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/exp/slices"
 	"google.golang.org/grpc/codes"
 
 	"github.com/authzed/spicedb/internal/datastore/common"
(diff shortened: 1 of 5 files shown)


internal/datastore/spanner/reader.go
```diff
@@ -16,6 +18,21 @@ import (
...
+//
+// You can find the documentation for forcing indices here:
+// https://cloud.google.com/spanner/docs/secondary-indexes#index-directive
+func forceIndex(b sq.SelectBuilder, setColumns []string) (sq.SelectBuilder, error) {
+	// The query is setting subject ID and not the resource ID; this is a
+	// reverse lookup.
+	if slices.Contains(setColumns, colUsersetObjectID) && !slices.Contains(setColumns, colObjectID) {
+		// When using the Postgres dialect, Spanner uses magic comments to
+		// force indices.
+		b = b.From(tableRelationship + " /*@ FORCE_INDEX=idx_relation_tuple_by_subject */")
```
kruegener, line 31:
```suggestion
		b = b.From(tableRelationship + " /*@ FORCE_INDEX = ix_relation_tuple_by_subject */")
```

jzelinskie:
Closing this as I think we need a different approach.

PR #1693, opened 2023-12-20
pkg/proto: upgrade vtprotobuf to support WKTs

jzelinskie:
This early-adopts better codegen for well-known types. Caveats should be the most improved.


josephschorr (approved):
LGTM

PR #1694, opened 2023-12-21
.github: use public-read-only docker account

jzelinskie:
Secrets are not passed to forks, so logging into a dummy account with a token limited to read-only for public repositories is the best thing we can do.


josephschorr (approved):
LGTM

PR #1695, opened 2023-12-22
Major cleanup and improvements around revision handling in the datastores

josephschorr:
1) Extract out the common revision types into a common library, add testing and change datastores to use the shared implementations
2) Change the non-decimal needed revisions to use a simple int64 or uint64
3) Change the HLC type used in CRDB to the CRDB-provided decimal type - this should result in reduced memory usage


internal/datastore/revisions/timestamprevision.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/authzed/spicedb/pkg/datastore"
+)
+
+// TimestampRevision is a revision that is a timestamp.
+type TimestampRevision int64
```
ecordell, line 12: why not uint64?

internal/datastore/revisions/hlcrevision.go
```diff
@@ -3,19 +3,20 @@ package revisions
...
+	"github.com/cockroachdb/apd"
 	"github.com/shopspring/decimal"
 
 	"github.com/authzed/spicedb/pkg/datastore"
 )
 
 // HLCRevision is a revision that is a hybrid logical clock, stored as a decimal.
 type HLCRevision struct {
-	decimal decimal.Decimal
+	decimal *apd.Decimal
```
ecordell, line 14: why a pointer here? if we directly embed, then an array of hlcs gc just has to check 1 object isntead of all of the entires.

internal/datastore/revisions/timestamprevision.go
josephschorr: Timestamps before supported negative numbers and I wasn't 100% certain if we needed to continue doing so, so I allowed it

internal/datastore/revisions/hlcrevision.go
josephschorr: All the New* methods return a pointer and all the accessors on it are on the pointer type

josephschorr: Changed to grab the value and use that

ecordell (dismissed):

ecordell (approved):

PR #1704, opened 2024-01-04
Add Union and Merge to Set

josephschorr:


alecmerdler (approved):

pkg/genutil/mapz/set.go
```diff
@@ -58,6 +58,23 @@ func (s *Set[T]) Extend(values []T) {
...
+func (s *Set[T]) Merge(other *Set[T]) {
+	for value := range other.values {
+		s.values[value] = struct{}{}
+	}
+}
+
+// Union adds all the values from the other set to this set,
+// returning a new set.
+func (s *Set[T]) Union(other *Set[T]) *Set[T] {
+	cpy := s.Copy()
```
jzelinskie, line 71: Why not just make this "return s.Copy().Merge(other)"

josephschorr: Parallels the other methods we have and, down the road, we may further optimize the implementation

PR #1705, opened 2024-01-08
expose revision parsing function for tests

vroldanbet:
since most of the Revision parsing APIs have been
recently refactored and the surface
is now exclusively the parsing functions
for each datastore


josephschorr (approved):
LGTM

PR #1706, opened 2024-01-09
Watch revision checkpoints using memdb datastore

alecmerdler:
Adds the ability to watch revision checkpoint events when using the memdb datastore implementation.


josephschorr (approved):
LGTM

PR #1707, opened 2024-01-09
Add timeout to watch buffers

josephschorr:
This ensures that even if the buffer fills, there is a window where it can be read from before the watch is terminated

### DIFF
--- a/internal/datastore/crdb/watch.go
+++ b/internal/datastore/crdb/watch.go
@@ -54,7 +54,12 @@ type changeDetails struct {
 }
 
 func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Revision, options datastore.WatchOptions) (<-chan *datastore.RevisionChanges, <-chan error) {
-	updates := make(chan *datastore.RevisionChanges, cds.watchBufferLength)
+	watchBufferLength := options.WatchBufferLength
+	if watchBufferLength <= 0 {
+		watchBufferLength = cds.watchBufferLength
+	}
+
+	updates := make(chan *datastore.RevisionChanges, watchBufferLength)
 	errs := make(chan error, 1)
 
 	features, err := cds.Features(ctx)
@@ -137,12 +142,28 @@ func (cds *crdbDatastore) watch(
 		errs <- err
 	}
 
+	watchBufferWriteTimeout := opts.WatchBufferWriteTimeout
+	if watchBufferWriteTimeout <= 0 {
+		watchBufferWriteTimeout = cds.watchBufferWriteTimeout
+	}
+
 	sendChange := func(change *datastore.RevisionChanges) bool {
 		select {
 		case updates <- change:
 			return true
 
 		default:
+			// If we cannot immediately write, setup the timer and try again.
+		}
+
+		timer := time.NewTimer(watchBufferWriteTimeout)
+		defer timer.Stop()
+
+		select {
+		case updates <- change:
+			return true
+
+		case <-timer.C:
 			errs <- datastore.NewWatchDisconnectedErr()
 			return false
 		}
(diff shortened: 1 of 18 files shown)


internal/datastore/mysql/watch.go
```diff
@@ -21,7 +21,12 @@ const (
 //
 // All events following afterRevision will be sent to the caller.
 func (mds *Datastore) Watch(ctx context.Context, afterRevisionRaw datastore.Revision, options datastore.WatchOptions) (<-chan *datastore.RevisionChanges, <-chan error) {
-	updates := make(chan *datastore.RevisionChanges, mds.watchBufferLength)
+	watchBufferLength := options.WatchBufferLength
+	if watchBufferLength <= 0 {
+		watchBufferLength = mds.watchBufferLength
```
jzelinskie, line 26: should probably warn when we do these fallbacks, but perhaps that's a tech debt pass for another time

josephschorr: I'll do it as a followup

jzelinskie (approved):

jzelinskie (approved):

alecmerdler (approved):

internal/datastore/crdb/watch.go
```diff
@@ -137,12 +142,28 @@ func (cds *crdbDatastore) watch(
 		errs <- err
 	}
 
+	watchBufferWriteTimeout := opts.WatchBufferWriteTimeout
+	if watchBufferWriteTimeout <= 0 {
+		watchBufferWriteTimeout = cds.watchBufferWriteTimeout
+	}
+
 	sendChange := func(change *datastore.RevisionChanges) bool {
```
vroldanbet, line 150: @josephschorr couldn't this code have been written like:

```
	sendChange := func(change *datastore.RevisionChanges) bool {
		select {
		case updates <- change:
			return true

		case <-time.After(watchBufferWriteTimeout):
			errs <- datastore.NewWatchDisconnectedErr()
			return false
		}
	}
```

pkg/cmd/datastore/datastore.go
```diff
@@ -212,7 +213,8 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	flagSet.Uint64Var(&opts.SpannerMaxSessions, flagName("datastore-spanner-max-sessions"), 400, "maximum number of sessions across all Spanner gRPC connections the client can have at a given time")
 	flagSet.StringVar(&opts.TablePrefix, flagName("datastore-mysql-table-prefix"), "", "prefix to add to the name of all SpiceDB database tables")
 	flagSet.StringVar(&opts.MigrationPhase, flagName("datastore-migration-phase"), "", "datastore-specific flag that should be used to signal to a datastore which phase of a multi-step migration it is in")
-	flagSet.Uint16Var(&opts.WatchBufferLength, flagName("datastore-watch-buffer-length"), 1024, "how many events the watch buffer should queue before forcefully disconnecting reader")
+	flagSet.Uint16Var(&opts.WatchBufferLength, flagName("datastore-watch-buffer-length"), 1024, "how large the watch buffer should be before blocking")
+	flagSet.DurationVar(&opts.WatchBufferWriteTimeout, flagName("datastore-watch-buffer-write-timeout"), 1*time.Second, "how long the watch buffer should queue before forcefully disconnecting the reader")
```
vroldanbet, line 217: @josephschorr `"how long the watch buffer should queue before forcefully disconnecting the reader once the buffer is full"`

pkg/datastore/datastore.go
```diff
@@ -342,6 +342,14 @@ type WatchOptions struct {
 	// If given the zero value, the datastore's default will be used. If smaller
 	// than the datastore's minimum, the minimum will be used.
 	CheckpointInterval time.Duration
+
+	// WatchBufferLength is the length of the buffer for the watch channel. If
+	// given the zero value, the datastore's default will be used.
+	WatchBufferLength uint16
```
vroldanbet, line 348: @josephschorr we probably want to specify some sort of upper limit in the various `Watch` implementations to make sure we don't OOMKill due to a massively large channel

josephschorr: That's why its a uint16 :)

PR #1717, opened 2024-01-18
Change telemetry failure to a warning and have Postgres check for its unique ID on startup

### LINKED ISSUE #1716: Change unable to init telemetry `error` to `warning` since it is not strict dependency to SpiceDB

linux

amd64

Latest docker image

### Steps to Reproduce

Having no telemetry, as far as I know.

```
spicedb_1  | {"level":"error","error":"unable to initialize telemetry collector: unable to query DB stats: unable to query unique ID: no rows in result set","time":"2024-01-18T12:05:32Z","message":"terminated with errors"}
```

### Expected Result

Is should be a `warning`. In some cases (for local development) there is no need for telemetry.

### Actual Result

It is an `error`.

josephschorr:
Fixes #1716

### DIFF
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -547,12 +547,19 @@ func (pgd *pgDatastore) ReadyState(ctx context.Context) (datastore.ReadyState, e
 	}
 
 	if version == headMigration {
+		// Ensure a datastore ID is present. This ensures the tables have not been truncated.
+		uniqueID, err := pgd.datastoreUniqueID(ctx)
+		if err != nil {
+			return datastore.ReadyState{}, fmt.Errorf("database validation failed: %w; if you have previously run `TRUNCATE`, this database is no longer valid and must be remigrated. See: https://spicedb.dev/d/truncate-unsupported", err)
+		}
+
+		log.Trace().Str("unique_id", uniqueID).Msg("postgres datastore unique ID")
 		return datastore.ReadyState{IsReady: true}, nil
 	}
 
 	return datastore.ReadyState{
 		Message: fmt.Sprintf(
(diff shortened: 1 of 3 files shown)


internal/datastore/postgres/postgres.go
```diff
@@ -547,12 +547,19 @@ func (pgd *pgDatastore) ReadyState(ctx context.Context) (datastore.ReadyState, e
 	}
 
 	if version == headMigration {
+		// Ensure a datastore ID is present. This ensures the tables have not been truncated.
+		uniqueID, err := pgd.datastoreUniqueID(ctx)
+		if err != nil {
+			return datastore.ReadyState{}, fmt.Errorf("database validation failed: %w; if you have previously run `TRUNCATE`, this database is no longer valid and must be remigrated. See: https://spicedb.dev/d/truncate-unsupported", err)
+		}
```
vroldanbet, line 554: truncating tables is really a problem for all datastores. And it's not only the unique ID, the "revision zero" is also needed. Each DB may have its own state written.

This improves the experience with Postgres but does not solve it holistically

josephschorr: It is, but we've encountered it primarily with Postgres and so I started there.

And yes, we need both, but if someone is truncating one table its likely the others are as well, so this is more a sanity check

vroldanbet: I do think this is the right place to solve it (in all of datastores `ReadyState` method). The risk is that folks can do all sorts of unexpected things to the underlying datastore, making assumptions of how it works.

I think this is probably fine for now, but we should open an issue and follow up on all datastores:
- add a test for all datastores
- check that the state of the datastore is in a good place (e.g. there is at least one revision, there exists at least one migration in the migration table, and the stats table is not empty)

vroldanbet (approved):

PR #1723, opened 2024-01-29
fix: add flags to configure how to handle zedtokens meant for a different datastore

### LINKED ISSUE #1541: Improvement: Prevent ZedToken's from being used cross-datastore
Right now, ZedTokens simply encode the datastore-provided revision information, with no additional information as to the *kind* of datastore used. This can result in weird behavior if a ZedToken from one SpiceDB cluster is used with a completely different cluster, or the datastore has been swapped out.

We should instead encode the datastore kind (and perhaps a unique ID) into the ZedToken, and either reject all such tokens, are automatically switch to full consistency (likely the former)

josephschorr:
The goal which is to prevent zedtokens not just across types but instances of the datastore as well. For this, we store a `uniqueID` per datastore instance and that gets included in the `zedToken`.

NOTE: ZedTokens are a bit longer now as a result of this change, but should still be well within the 1024 limit previously defined

Fixes #1541


josephschorr:
Rebased

josephschorr:
Rebased

josephschorr:
Updated

internal/datastore/crdb/stats.go
```diff
@@ -18,22 +18,33 @@ const (
...
+
+		var uniqueID string
 		if err := cds.readPool.QueryRowFunc(ctx, func(ctx context.Context, row pgx.Row) error {
 			return row.Scan(&uniqueID)
 		}, sql, args...); err != nil {
-			return datastore.Stats{}, fmt.Errorf("unable to query unique ID: %w", err)
+			return "", fmt.Errorf("unable to query unique ID: %w", err)
 		}
+
+		cds.uniqueID.Store(&uniqueID)
```
vroldanbet, line 37: I don't think this matters too much here because the value converges, but this can race. Ideally we do a CAS operation.

I'd prefer to see this implemented with `sync.Once` instead of an atomic pointer. The ID does not change for the lifetime of the server.

pkg/datastore/datastore.go
```diff
@@ -543,6 +543,11 @@ func (wo WatchOptions) WithCheckpointInterval(interval time.Duration) WatchOptio
 
 // ReadOnlyDatastore is an interface for reading relationships from the datastore.
 type ReadOnlyDatastore interface {
+	// UniqueID returns a unique identifier for the datastore. This identifier
+	// must be stable across restarts of the datastore if the datastore is
+	// persistent.
+	UniqueID(context.Context) (string, error)
```
vroldanbet, line 549: Did you consider alternatives using the `UniqueID` as a stable identifier for zedtokens?

I _assume_ (not present in the PR body) that the goal is making sure zedtokens from one datastore _type_ are not used in another datastore _type_. But what happens if you evolve the zedtoken implementation from one version to another for the same datastore, and want to force them to be dropped? the `uniqueID` wouldn't help, would it? 
Shouldn't we store a zedtoken versioning parameter like `datastoreType+zedtokenVersion`?

vroldanbet (commented):
Some early feedback, I'll continue tomorrow. Please describe in the PR body what problems are you trying to solve, to help folks reviewing the PR with the right context 🙏🏻

josephschorr:
> Some early feedback, I'll continue tomorrow. Please describe in the PR body what problems are you trying to solve and design choices you took to come up with this solution. It helps folks reviewing the PR with the right context 🙏🏻

I added the fixes; it was on the commit but not the PR

internal/datastore/crdb/stats.go
josephschorr: `sync.Once` won't work because if we get an error, we need to rerun this code the next time its called

pkg/datastore/datastore.go
josephschorr: Yes, but then it doesn't meet the goal which is to prevent zedtokens not just across *types* but *instances* as well

internal/datastore/crdb/stats.go
vroldanbet: Fair. Can we at least use the CAS operation to store it? if it fails, it probably means some other goroutine already set it.

Another point in favor of the atomic value is it does not rely on any mutex. `sync.Once` relies on it while it's not initialized, but once it is, the overhead of the mutex disapears.

internal/services/v1/relationships.go
```diff
@@ -353,8 +353,13 @@ func (ps *permissionServer) WriteRelationships(ctx context.Context, req *v1.Writ
 		writeUpdateCounter.WithLabelValues(v1.RelationshipUpdate_Operation_name[int32(kind)]).Observe(float64(count))
 	}
 
+	zedToken, err := zedtoken.NewFromRevision(ctx, revision, ds)
+	if err != nil {
+		return nil, ps.rewriteError(ctx, err)
+	}
+
 	return &v1.WriteRelationshipsResponse{
-		WrittenAt: zedtoken.MustNewFromRevision(revision),
```
vroldanbet, line 357: hmm, Im realizing we used `MustNewFromRevision` everywhere, when it's something that would panic in case of malformed proto encoding. I guess it worked because the middleware prevented any incorrect revision from making it through.

internal/services/v1/schema.go
```diff
@@ -97,9 +97,14 @@ func (ss *schemaServer) ReadSchema(ctx context.Context, _ *v1.ReadSchemaRequest)
 		DispatchCount: dispatchCount,
 	})
 
+	zedToken, err := zedtoken.NewFromRevision(ctx, headRevision, ds)
```
vroldanbet, line 100: at this point if feels like `NewFromRevision` should be a method in the `Datastore` interface, but I get the convenience of not relying on the datastore specially for tests

internal/services/v1/watch.go
```diff
@@ -52,11 +52,15 @@ func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchS
 
 	var afterRevision datastore.Revision
 	if req.OptionalStartCursor != nil && req.OptionalStartCursor.Token != "" {
-		decodedRevision, err := zedtoken.DecodeRevision(req.OptionalStartCursor, ds)
+		decodedRevision, tokenStatus, err := zedtoken.DecodeRevision(req.OptionalStartCursor, ds)
 		if err != nil {
 			return status.Errorf(codes.InvalidArgument, "failed to decode start revision: %s", err)
 		}
 
+		if tokenStatus == zedtoken.StatusMismatchedDatastoreID {
```
vroldanbet, line 60: I think context is missing on _why_ we are choosing to fail here. I would recommend adding a comment to clarify, since the implications are not immediately clear.

I'm trying to think what type of scenario are we trying to protect from by failing here, but also potential use-cases we could be hampering by doing so.

- Clearly zedtoken for a different datastore type should fail as those can fail in unexpected ways, even though we could potentially find ways some of them interoperate.
- Same datastore type, but different instances: we can't guarantee to observe the same transactions at the same timestamp in different instances.

In the case of Watch, reusing a zedtoken in different instances could lead to missing events.

The only scenario I can think that could be impacted by this are dual-writing to a new SpiceDB instance and doing an eventual cut-over (e.g. deep schema refactor). API calls with `at_least_as_fresh` can recover by momentarily switching to `minimize_latency` until the tokens are refreshed (although this could take _very long_ as it depends on resource changes! the system could be running in minimize for _a long time_). A cut-over to the Watch API will simply not work, regardless of the changes in this PR, as it can only work if they represent the same state of the world.

Long story short, I think this is the right call, and I wanted to think out lout and leave a trace of the why.

vroldanbet (changes requested):
We are transferring UUIDs in every API call for no reason other than to prevent an eventual migration within the same datastore type. That's **extremely** wasteful.

It's still within the zedtoken spec limit, but it is a very high price to pay for an incredible rare event (a legit one, don't get me wrong). We are storing UUIDs in the datastore as strings, which is the least compact way of storing them and transferring them over the wire. And we are forcing **everybody** to see the data transferred increase significantly (and latency!) to prevent a scenario that the system will not be subjected to in, I'll dare to say, practically 100% of its lifespan.

We need to rethink this. I get the zedtoken is a stateless token and that it needs to self-contain this information, but I think we can get 99% there by defining the datastore type as a small attribute in the token. I understand migrating from different instances of the same datastore can be problematic, but no one asked for this, and we could find alternative ways for it (e.g. a flag that forces SpiceDB to ignore requested consistency and fallback to `full_consistency` all the time, or `minimize_latency` if the customer can take the hit of ignoring the new enemy problem during the migration)

josephschorr:
> We are transferring UUIDs in every API call for no reason other than to prevent an eventual migration within the same datastore type. That's **extremely** wasteful.

Is it really? We could reduce the size of the ID to just a small prefix if we wish to save some bytes.

> 
> It's still within the zedtoken spec limit, but it is a very high price to pay for an incredible rare event (a legit one, don't get me wrong). We are storing UUIDs in the datastore as strings, which is the least compact way of storing them and transferring them over the wire. And we are forcing **everybody** to see the data transferred increase significantly (and latency!) to prevent a scenario that the system will not be subjected to in, I'll dare to say, practically 100% of its lifespan.

Except users have already been subject to it - this was spurred by two, different reports: one where a user moved between datastore types, but another where they moved between datastores of the *same* type (but accidentally)

> 
> We need to rethink this. I get the zedtoken is a stateless token and that it needs to self-contain this information, but I think we can get 99% there by defining the datastore type as a small attribute in the token. I understand migrating from different instances of the same datastore can be problematic, but no one asked for this, 

It has happened before, sadly.

> and we could find alternative ways for it (e.g. a flag that forces SpiceDB to ignore requested consistency and fallback to `full_consistency` all the time, or `minimize_latency` if the customer can take the hit of ignoring the new enemy problem during the migration). I'd argue that by adding such a flag we could completely avoid having to add the datastore type too.

This PR does add that flag, but we need a means of tracking to ensure that zedtokens are not used across instances of a datastore. The fact that we allow it now is, technically speaking, a small but real risk.

josephschorr:
Updated to only store 8 bytes of prefix of the datastore ID. Since datastores are *rarely* updated, this should be fine for detecting changes.

internal/datastore/crdb/stats.go
josephschorr: Why even do that though? The ID is guaranteed to be the same by the datastore, so no need to compare?

internal/middleware/consistency/consistency.go
```diff
@@ -55,27 +56,47 @@ func RevisionFromContext(ctx context.Context) (datastore.Revision, *v1.ZedToken,
 		handle := c.(*revisionHandle)
 		rev := handle.revision
 		if rev != nil {
-			return rev, zedtoken.MustNewFromRevision(rev), nil
+			ds := datastoremw.FromContext(ctx)
```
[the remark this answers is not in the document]
josephschorr: Are you asking me to add a comment stating that or add something somewhere? (sorry for forgetting)

internal/middleware/consistency/consistency.go
```diff
@@ -219,24 +246,48 @@ func (s *recvWrapper) RecvMsg(m interface{}) error {
...
+func pickBestRevision(ctx context.Context, requested *v1.ZedToken, ds datastore.Datastore, option MismatchingTokenOption) (datastore.Revision, bool, error) {
 	// Calculate a revision as we see fit
 	databaseRev, err := ds.OptimizedRevision(ctx)
 	if err != nil {
 		return datastore.NoRevision, false, err
 	}
 
 	if requested != nil {
-		requestedRev, err := zedtoken.DecodeRevision(requested, ds)
+		requestedRev, status, err := zedtoken.DecodeRevision(requested, ds)
```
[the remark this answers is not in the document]
josephschorr: Its explicitly not an error though; it only becomes an error if that is the setting on the CLI. I also do like that callers must handle it, as you said.

internal/middleware/consistency/consistency.go
```diff
@@ -147,11 +168,16 @@ func addRevisionToContextFromConsistency(ctx context.Context, req hasConsistency
...
 		ConsistentyCounter.WithLabelValues("snapshot", "request").Inc()
 
-		requestedRev, err := zedtoken.DecodeRevision(consistency.GetAtExactSnapshot(), ds)
+		requestedRev, status, err := zedtoken.DecodeRevision(consistency.GetAtExactSnapshot(), ds)
 		if err != nil {
 			return errInvalidZedToken
 		}
 
+		if status == zedtoken.StatusMismatchedDatastoreID {
+			log.Error().Str("zedtoken", consistency.GetAtExactSnapshot().Token).Msg("ZedToken specified references an older datastore but at-exact-snapshot was requested")
```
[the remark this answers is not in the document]
josephschorr: Changed to a warning

internal/services/v1/relationships.go
josephschorr: Yeah, we could move away from the `Must` approach, but it shouldn't fail in "normal" operation

internal/services/v1/schema.go
josephschorr: Yes and no; it uses the datastore's unique ID but its also not strictly tied to the datastore, so I prefer this approach

internal/services/v1/watch.go
josephschorr: Added a comment

josephschorr:
Rebased

internal/middleware/consistency/consistency.go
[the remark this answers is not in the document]
vroldanbet: No. The middleware framework has API to define dependencies between middleware instances, which makes it fail in dev if the order is not respected.

pkg/datastore/datastore.go
vroldanbet: >But what happens if you evolve the zedtoken implementation from one version to another for the same datastore, and want to force them to be dropped? the uniqueID wouldn't help, would it?
Shouldn't we store a zedtoken versioning parameter like datastoreType+zedtokenVersion?

Can you address this ☝️?

josephschorr:
Rebased

josephschorr:
Rebased

internal/datastore/memdb/memdb.go
```diff
@@ -108,6 +108,10 @@ func (mdb *memdbDatastore) MetricsID() (string, error) {
 	return "memdb", nil
 }
 
+func (mdb *memdbDatastore) UniqueID(_ context.Context) (string, error) {
+	return mdb.uniqueID, nil
```
miparnisari, line 111: this is an empty string right? why do we need to keep `uniqueID` around, just `return "", nil`?

pkg/cmd/serve.go
```diff
@@ -113,6 +113,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	apiFlags.Uint32Var(&config.MaxLookupResourcesLimit, "max-lookup-resources-limit", 1000, "maximum number of resources that can be looked up in a single request")
 	apiFlags.Uint32Var(&config.MaxBulkExportRelationshipsLimit, "max-bulk-export-relationships-limit", 10_000, "maximum number of relationships that can be exported in a single request")
 	apiFlags.BoolVar(&config.EnableRevisionHeartbeat, "enable-revision-heartbeat", true, "enables support for revision heartbeat, used to create a synthetic revision on an interval defined by the quantization window (postgres only)")
+	apiFlags.StringVar(&config.MismatchZedTokenBehavior, "mismatch-zed-token-behavior", "full-consistency", "behavior when a mismatched zedtoken is encountered. One of: full-consistency (treat as a full-consistency call), min-latency (treat as a min-latency call), error (return an error). defaults to full-consistency for safety.")
```
miparnisari:
```suggestion
	apiFlags.StringVar(&config.MismatchZedTokenBehavior, "mismatch-zed-token-behavior", "full-consistency", "behavior to enforce when an API call receives a zedtoken that was originally intended for a different kind of datastore. One of: full-consistency (treat as a full-consistency call, ignoring the zedtoken), min-latency (treat as a min-latency call, ignoring the zedtoken), error (return an error). defaults to full-consistency for safety.")
```

miparnisari: why "for safety"? isn't it "to prevent downtime"?

pkg/datastore/test/datastore.go
```diff
@@ -111,6 +111,7 @@ func AllWithExceptions(t *testing.T, tester DatastoreTester, except Categories,
 		runner = parallel
 	}
 
+	t.Run("TestUniqueID", func(t *testing.T) { UniqueIDTest(t, tester) })
```
miparnisari:
```suggestion
	t.Run("TestUniqueID", func(t *testing.T) { runner(tester, UniqueIDTest) })
```

pkg/datastore/test/basic.go
```diff
@@ -83,3 +83,21 @@ func DeleteAllDataTest(t *testing.T, tester DatastoreTester) {
...
+	// Create the datastore.
+	ds, err := tester.New(0, veryLargeGCInterval, veryLargeGCWindow, 1)
+	require.NoError(err)
+
+	// Ensure the unique ID is not empty.
+	uniqueID, err := ds.UniqueID(context.Background())
+	require.NoError(err)
+	require.NotEmpty(uniqueID)
+
+	// Ensure the unique ID is stable.
```
miparnisari, line 98: how is this code testing that it's stable across restarts? where is the restart happening? (it seems that every test is spinning up a new DS but not stopping it?)

pkg/middleware/consistency/consistency.go
```diff
@@ -232,24 +269,48 @@ func (s *recvWrapper) RecvMsg(m interface{}) error {
...
-		requestedRev, err := zedtoken.DecodeRevision(requested, ds)
+		requestedRev, status, err := zedtoken.DecodeRevision(requested, ds)
 		if err != nil {
 			return datastore.NoRevision, false, errInvalidZedToken
 		}
 
+		if status == zedtoken.StatusMismatchedDatastoreID {
+			switch option {
+			case TreatMismatchingTokensAsFullConsistency:
+				log.Warn().Str("zedtoken", requested.Token).Msg("ZedToken specified references a different datastore instance and SpiceDB is configured to treat this as a full consistency request")
```
miparnisari, line 294: non-blocking comment: assuming that spicedb logs all request parameters, including `consistency`, would it not be more useful to add a new field `overriden-consistency` or something like that? otherwise when you are going through logs looking for a particular request, you will not get the full picture just by looking at one log line; you will need to look at some lines before it

proto/internal/impl/v1/impl.proto
```diff
@@ -33,6 +33,10 @@ message DecodedZedToken {
   }
   message V1ZedToken {
     string revision = 1;
+
+    // datastore_unique_id_prefix is a prefix of the unique ID for the datastore that created
```
miparnisari, line 37: for future us, can we document why this is a prefix and not the entire thing?

maybe we could also document example values for the various datastores

internal/datastore/memdb/memdb.go
josephschorr: No; its generated on startup as a UUID

pkg/cmd/serve.go
josephschorr: It prevents errors from being returned

pkg/datastore/test/basic.go
josephschorr: It only verifies it is stable across calls; it is not verifying across restarts (nor does any other data storage test, mind you... that is assumed)

pkg/middleware/consistency/consistency.go
josephschorr: This is why I want to add some sort of proper warning framework; even `overriden-consistency` would be hard to see

proto/internal/impl/v1/impl.proto
josephschorr: It was deemed too big; see the discussion above

josephschorr:
Rebased and responded

proto/internal/impl/v1/impl.proto
miparnisari: Yes I saw that but i'm suggesting leaving a comment in the code with the summary

pkg/datastore/datastore.go
josephschorr: We'd try to parse the zedtoken, fail, and use the new fallback mechanism

PR #1739, opened 2024-02-07
Extend support for relationship filtering and add relationship filtering to other APIs

### LINKED ISSUE #315: Remove/Delete an object and all its relations (including used as subjects) in one call
Given this schema:
```
definition user {}

definition user_group {
	relation member: user
}

definition organization {
	relation member: user_group#member
}
```

And these relationships:
```
user_group:organization_members#member@user:user1
user_group:organization_members#member@user:user2
organization:myOrganization#member@organization_members#member
```

I have to run two calls to remove the existence of "organization_members" in its entirety:

```
_, err := a.client.DeleteRelationships(ctx, &v1.DeleteRelationshipsRequest{
    RelationshipFilter: &v1.RelationshipFilter{
        ResourceType:           "user_group",
        OptionalResourceId:  "organization_members",
    },
})

// This call removes these relationships
// user_group:organization_members#member@user:user1
// user_group:organization_members#member@user:user2

_, err = a.client.DeleteRelationships(ctx, &v1.DeleteRelationshipsRequest{
    RelationshipFilter: &v1.RelationshipFilter{
        ResourceType: "organization",
        OptionalSubjectFilter: &v1.SubjectFilter{
            SubjectType:       "user_group",
            OptionalSubjectId: "organization_members",
            OptionalRelation: &v1.SubjectFilter_RelationFilter{
                Relation: "member",
            },
        },
    },
})

// This call removes this relationship
// organization:myOrganization#member@organization_members#member
```

I'd like to have a call to remove the object and all its relations, that way I can be sure that there isn't any traces of it in the DB.

josephschorr:
> [!IMPORTANT]  
> If a relationships filter is used without a resource type, there can be a performance impact due to the existing indexes all starting with the resource type. We may examine adding additional indexes in the future, but for now, the performance impact should be noted

- Extends the relationship filter to allow for an optional `resource_type`, and a *prefix* for object IDs
- Adds support for the relationship filter on  BulkExport and Watch APIs

With this change, all objects can now be removed from the datastore with *two* calls (once with the object ID as the resource ID and once with it as the subject ID).

Fixes #315 (more or less; its two calls instead of one)


pkg/datastore/datastore.go
```diff
@@ -101,8 +107,44 @@ type RelationshipsFilter struct {
 	OptionalCaveatName string
 }
 
+// Test returns true iff the given relationship is matched by this filter.
```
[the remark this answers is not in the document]
josephschorr: `iff` is correct: if and only if

internal/services/v1/watch.go
```diff
@@ -97,20 +108,38 @@ func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchS
...
-		if _, ok := objectTypes[objectType]; ok {
-			filtered = append(filtered, update)
+		if !objectTypes.IsEmpty() && !objectTypes.Has(objectType) {
+			continue
+		}
+
+		if len(filters) > 0 {
+			// If there are filters, we need to check if the update matches any of them.
+			matched := false
+			for _, filter := range filters {
```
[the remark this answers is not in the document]
josephschorr: If you want to watch a subset of relationships you'd specify a filter for each type. Customers have requested it so they don't need to do the filtering on their end, which reduces overhead

pkg/datastore/datastore.go
```diff
@@ -101,8 +107,44 @@ type RelationshipsFilter struct {
...
+	if len(rf.OptionalSubjectsSelectors) > 0 {
+		for _, selector := range rf.OptionalSubjectsSelectors {
+			if selector.Test(relationship.Subject) {
+				return true
+			}
+		}
+		return false
+	}
+
+	if rf.OptionalCaveatName != "" {
```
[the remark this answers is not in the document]
josephschorr: Fixed

internal/services/v1/errors.go
```diff
@@ -337,6 +344,60 @@ func (err ErrInvalidCursor) GRPCStatus() *status.Status {
...
+}
+
+// GRPCStatus implements retrieving the gRPC status for the error.
+func (err ErrInvalidFilter) GRPCStatus() *status.Status {
+	// TODO(jschorr): Put a proper error reason in here.
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.InvalidArgument,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_UNSPECIFIED,
```
ecordell, line 359: any reason to wait on defining a real reason here?

internal/services/v1/experimental.go
```diff
@@ -647,29 +669,29 @@ func queryForEach(
...
 
-	cur := tuple.Parse(decoded.GetV1().GetSections()[0])
+	cur := tuple.Parse(decoded.GetV1().GetSections()[1])
 	if cur == nil {
-		return datastore.NoRevision, nil, errors.New("malformed cursor: invalid encoded relation tuple")
+		return datastore.NoRevision, "", nil, errors.New("malformed cursor: invalid encoded relation tuple")
 	}
 
-	return atRevision, cur, nil
+	return atRevision, decoded.GetV1().GetSections()[0], cur, nil
```
ecordell, line 696: maybe a note about what this new return value is?

internal/datastore/common/sql.go
```diff
@@ -275,6 +275,32 @@ func (sqf SchemaQueryFilterer) MustFilterToResourceIDs(resourceIds []string) Sch
...
+
+	// NOTE: we do *not* record the use of the resource ID column here, because it is not used
+	// statically and thus is necessary for sorting operations.
+	return sqf, nil
+}
+
+func (sqf SchemaQueryFilterer) MustFilterWithResourceIDPrefix(prefix string) SchemaQueryFilterer {
+	updated, err := sqf.FilterWithResourceIDPrefix(prefix)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
josephschorr: This function is only called from tests

internal/services/v1/errors.go
josephschorr: It requires an API repo change and I didn't want to block this change, so I left it as a TODO

internal/services/v1/watch.go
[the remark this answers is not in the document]
vroldanbet: That makes sense, it was just not immediately obvious from the API description. I've opened https://github.com/authzed/api/pull/94 to clarify this

internal/datastore/common/sql.go
```diff
@@ -325,12 +351,18 @@ func (sqf SchemaQueryFilterer) MustFilterWithRelationshipsFilter(filter datastor
 }
 
 func (sqf SchemaQueryFilterer) FilterWithRelationshipsFilter(filter datastore.RelationshipsFilter) (SchemaQueryFilterer, error) {
-	sqf = sqf.FilterToResourceType(filter.ResourceType)
+	if filter.OptionalResourceType != "" {
+		sqf = sqf.FilterToResourceType(filter.OptionalResourceType)
```
vroldanbet, line 355: `sqf` is the method received and is not a pointer. I assume this is not intended to have side effects and is just used for ergonomic convenience. To make it more obvious, create a local variable that does not shadow the receiver.

internal/datastore/common/sql.go
```diff
@@ -325,12 +351,18 @@ func (sqf SchemaQueryFilterer) MustFilterWithRelationshipsFilter(filter datastor
...
+	}
 
 	if filter.OptionalResourceRelation != "" {
 		sqf = sqf.FilterToRelation(filter.OptionalResourceRelation)
 	}
 
+	if len(filter.OptionalResourceIds) > 0 && filter.OptionalResourceIDPrefix != "" {
```
vroldanbet, line 362: This does not have a unit test, perhaps add one to assert that the condition will panic? (even though this won't happen as it is already checked in the service controllers)

internal/datastore/common/sql.go
```diff
@@ -275,6 +275,32 @@ func (sqf SchemaQueryFilterer) MustFilterToResourceIDs(resourceIds []string) Sch
...
+// starts with the specified prefix.
+func (sqf SchemaQueryFilterer) FilterWithResourceIDPrefix(prefix string) (SchemaQueryFilterer, error) {
+	if strings.Contains(prefix, "%") {
+		return sqf, spiceerrors.MustBugf("prefix cannot contain the percent sign")
+	}
+	if prefix == "" {
+		return sqf, spiceerrors.MustBugf("prefix cannot be empty")
+	}
+
+	sqf.queryBuilder = sqf.queryBuilder.Where(sq.Like{sqf.schema.colObjectID: prefix + "%"})
```
vroldanbet, line 288: Do we understand if this is properly covered by an index?

When cursors were introduced, having the various datastores match a reasonable index with `ReadRelationships` calls became rather unpredictable. I suspect adding another condition could further confuse query planners. It's worth testing various query scenarios for index coverage. We have some tests that support asserting on EXPLAIN statements (only postgres, unfortunately)

internal/datastore/memdb/readonly.go
```diff
@@ -261,13 +264,24 @@ func (r *memdbReader) mustLock() {
 }
 
 func iteratorForFilter(txn *memdb.Txn, filter datastore.RelationshipsFilter) (memdb.ResultIterator, error) {
-	index := indexNamespace
-	args := []any{filter.ResourceType}
-	if filter.OptionalResourceRelation != "" {
+	index := indexNamespace + "_prefix"
+	args := []any{}
```
vroldanbet, line 268:
```suggestion
	var args []any
```

internal/datastore/memdb/readonly.go
```diff
@@ -261,13 +264,24 @@ func (r *memdbReader) mustLock() {
 }
 
 func iteratorForFilter(txn *memdb.Txn, filter datastore.RelationshipsFilter) (memdb.ResultIterator, error) {
-	index := indexNamespace
-	args := []any{filter.ResourceType}
-	if filter.OptionalResourceRelation != "" {
+	index := indexNamespace + "_prefix"
```
vroldanbet, line 267: can you explain why `_prefix` was added to the default index? No index like that can be found in `spicedb/internal/datastore/memdb/schema.go`

internal/datastore/memdb/readwrite.go
```diff
@@ -252,10 +258,12 @@ func relationshipFilterFilterFunc(filter *v1.RelationshipFilter) func(interface{
 
 		// If it doesn't match one of the resource filters, filter it.
 		switch {
-		case filter.ResourceType != tuple.namespace:
+		case filter.ResourceType != "" && filter.ResourceType != tuple.namespace:
```
vroldanbet, line 261: not exercised via the datastore integration tests

internal/datastore/memdb/readwrite.go
```diff
@@ -252,10 +258,12 @@ func relationshipFilterFilterFunc(filter *v1.RelationshipFilter) func(interface{
...
 			return true
 		case filter.OptionalResourceId != "" && filter.OptionalResourceId != tuple.resourceID:
 			return true
+		case filter.OptionalResourceIdPrefix != "" && !strings.HasPrefix(tuple.resourceID, filter.OptionalResourceIdPrefix):
```
vroldanbet, line 265: not exercised via the datastore integration tests

internal/datastore/memdb/readwrite.go
```diff
@@ -252,10 +258,12 @@ func relationshipFilterFilterFunc(filter *v1.RelationshipFilter) func(interface{
...
+			return true
```
vroldanbet, line 266: also not exercised (but predate this PR) 👇🏻 

```
subjectFilter.SubjectType != tuple.subjectNamespace:
```

```
case subjectFilter.OptionalRelation != nil &&
				stringz.DefaultEmpty(subjectFilter.OptionalRelation.Relation, datastore.Ellipsis) != tuple.subjectRelation:
```

internal/datastore/common/sql.go
josephschorr: The wildcard will be because it is a prefix, but see the note at the top of the PR about indexes overall

internal/datastore/memdb/readonly.go
josephschorr: Its a built-in to the backing memdb for prefix filtering; I'll add a note

internal/services/v1/experimental.go
```diff
@@ -401,10 +427,6 @@ func (es *experimentalServer) BulkExportRelationships(
 				return es.rewriteError(ctx, err)
 			}
 		}
-
-		// Datastore namespace order might not be exactly the same as go namespace order
-		// so we shouldn't assume cursors are valid across namespaces
-		cur = nil
```
vroldanbet, line 407: can you clarify what happened here? I see a similar change was introduced earlier in the loop to make sure the cursor is reset for each iteration as the namespace changes

internal/datastore/common/sql.go
josephschorr: I don't generally like to assert panics and this is only a "break glass in case of emergency" panic (for the reason you specified), so I'm not 100% convinced it needs a unit test

internal/datastore/postgres/readwrite.go
```diff
@@ -429,14 +432,14 @@ func (rwt *pgReadWriteTXN) DeleteNamespaces(ctx context.Context, nsNames ...stri
 		switch {
 		case errors.As(err, &datastore.ErrNamespaceNotFound{}):
 			return err
+
 		case err == nil:
-			break
+			nsClauses = append(nsClauses, sq.Eq{colNamespace: nsName})
```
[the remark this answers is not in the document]
josephschorr: Has this changed appreciably?

internal/services/v1/experimental.go
```diff
@@ -647,29 +669,30 @@ func queryForEach(
 	return cur, nil
 }
 
-func decodeCursor(ds datastore.Datastore, encoded *v1.Cursor) (datastore.Revision, *core.RelationTuple, error) {
+func decodeCursor(ds datastore.Datastore, encoded *v1.Cursor) (datastore.Revision, string, *core.RelationTuple, error) {
```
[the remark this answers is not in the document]
josephschorr: yes, it has: we now have a namespace contained in it

internal/services/v1/experimental.go
josephschorr: Not sure I follow

internal/services/v1/relationships.go
```diff
@@ -114,16 +114,24 @@ type permissionServer struct {
 }
 
 func (ps *permissionServer) checkFilterComponent(ctx context.Context, objectType, optionalRelation string, ds datastore.Reader) error {
+	if objectType == "" {
```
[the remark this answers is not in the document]
josephschorr: the read test without a resource type does

internal/services/v1/relationships.go
```diff
@@ -114,16 +114,24 @@ type permissionServer struct {
...
+
 	relationToTest := stringz.DefaultEmpty(optionalRelation, datastore.Ellipsis)
 	allowEllipsis := optionalRelation == ""
 	return namespace.CheckNamespaceAndRelation(ctx, objectType, relationToTest, allowEllipsis, ds)
 }
 
-func (ps *permissionServer) checkFilterNamespaces(ctx context.Context, filter *v1.RelationshipFilter, ds datastore.Reader) error {
-	if err := ps.checkFilterComponent(ctx, filter.ResourceType, filter.OptionalRelation, ds); err != nil {
-		return err
+func (ps *permissionServer) validateRelationshipsFilter(ctx context.Context, filter *v1.RelationshipFilter, ds datastore.Reader) error {
```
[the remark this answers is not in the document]
josephschorr: Done

josephschorr:
Updated. Good catch on the deletion side: it was missing handlers for the prefix filtering

internal/services/v1/experimental.go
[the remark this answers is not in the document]
vroldanbet: in that case, don't we have a mechanism to version cursors already in place, and that we should be using it?

e.g. a cursor V1 is not compatible with cursor V2 or viceversa

internal/datastore/postgres/readwrite.go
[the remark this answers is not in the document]
vroldanbet: it hasn't, just surprised me to see logic next to error handling

vroldanbet (approved):

PR #1741, opened 2024-02-07
Prevent the staleness of an optimized revision from exceeding the GC window

josephschorr:
Currently, if the max staleness percentage is configured well in excess of the norm, it can result in OptimizedRevision occasionally returning revisions that are already expired


alecmerdler (approved):
Nice fix

jakedt (approved):

vroldanbet (commented):
Shouldn't we terminate with an error on startup instead of letting the server go healthy?

PR #1743, opened 2024-02-08
Reduce memory usage of WriteSchema

josephschorr:
We now only annotate the namespaces that have been changed or added, as annotation and canonicalization uses a significant amount of memory


vroldanbet (commented):
Can you perhaps write a benchmark that compares allocations before and after with a large schema?

josephschorr:
@vroldanbet I could but I verified it manually; not sure we really need this continuously tested

ecordell (approved):

PR #1744, opened 2024-02-13
goreleaser: init snap

jzelinskie:


jnsgruk:
Hey Jimmy, do you need any assistance moving this forward? There's some simple CI work we could do to start pushing this out to `edge` on merges to `main` if that's something you'd like to experiment with?

jzelinskie:
It looks like my last commit does the trick, but it doesn't follow the `snap.spicedb` naming convention.

What's next to get this plugged into the snap store channels?
We've got another goreleaser config and workflow for pushing builds of main that'd be trivial to update with the snaps.

jnsgruk:
Next step would be to [register your snap name](https://snapcraft.io/docs/registering-your-app-name). I'd recommend using a company account for signing up to snapcraft.io (like team@authzed.com or similar) so we can get it verified later.

Once you've signed up and registered, you'll need to [export a token](https://snapcraft.io/docs/snapcraft-authentication#heading--export) and set it in Github Actions secret as the variable `SNAPCRAFT_STORE_CREDENTIALS`. 

Finally - set up a job that publishes the `*.snap` files [like this] (https://github.com/parca-dev/parca/blob/c14c86b662c8c7cb12c633e6e0c668af20cc4dbb/.github/workflows/release.yml#L187-L206). The other option might just be to set `publish: true` in your `snapcrafts` section of `goreleaser`, but I've got less experience with how to auth that etc.

jnsgruk:
Looking at the goreleaser docs, you may need something like so in your `snapcrafts` section:

```yaml
extra_files:
      - source: completions/spicedb.bash
        destination: completions/spicedb.snap
        mode: 0755
```

Which will actually stage the completer, with the bonus that you can follow the snapcraft naming convention?

jzelinskie:
I'm not sure why, but extra_files appears to do nothing and I can only get `completer` to include the file when it ends with `.bash` and not `.snap`.

jzelinskie:
Sorry about this -- I was traveling for KubeCon and on vacation. Getting back to it soon.

jnsgruk:
Hey @jzelinskie! I did a bit of digging this morning, and I think the naming is fine.

I also came across this [guide](https://forum.snapcraft.io/t/debugging-tab-completion/4198) for debugging it if we run into problems. I note that in the example there they use:

```yaml
name: foo
version: v1
apps:
  bar:
    command: bar
    completer: bar-completer.sh # this is the one!
```

So I think you're good!

josephschorr (approved):
LGTM

PR #1747, opened 2024-02-19
fixes pgx min connection count always being set to max count

vroldanbet:
the logic ignored teh provided `opts.MinOpenConns` and always defaulted to the max value provided.

It affects both Postgres and CockroachDB datastores.


josephschorr (approved):
LGTM

Issue #1748, opened 2024-02-20
Use the same default port for the HTTP API across `serve` and `serve-testing` 


torbenw:
### What platforms are affected?

linux, macos, windows

### What architectures are affected?

amd64, arm64

### What SpiceDB version are you using?

v1.29.1

### Steps to Reproduce

1. Start SpiceDB with `serve --http-enabled` and look for `http server started serving` in the output.
2. Start SpiceDB with `serve-testing --http-enabled` and look for `http server started serving` in the output.

### Expected Result

The port for the HTTP API should be the same for `serve` and `serve-testing`.

### Actual Result

`serve` starts an HTTP server on port 8443:
`{"level":"info","addr":":8443","service":"http","insecure":true,"time":"2024-02-20T17:51:01Z","message":"http server started serving"}`

`serve-testing` starts an HTTP sever on port 8081:
`{"level":"info","addr":":8081","service":"http","insecure":true,"time":"2024-02-20T18:03:23Z","message":"http server started serving"}`

josephschorr:
@torbenw Should be a fairly small change on `serve-testing`; mind issuing a PR?

torbenw:
I've created pull request #1749 and changed 2 ports for `serve-testing`:
8081 -> 8443
8082 -> 8444

Now this should be equal to the ports of `serve` (but I did not test my changes, because I don't have anything set up for Go).

josephschorr:
Fixed by #1749. Thanks @torbenw!

PR #1749, opened 2024-02-20
Use the same default port for the HTTP API across serve and serve-testing

torbenw:
Issue #1748: Use the same default port for the HTTP API across serve and serve-testing


josephschorr:
@torbenw Please sign the CLA

torbenw:

vroldanbet (commented):
Changes look good to me, but this would be a breaking change for folks using `serve-testing`. Are we ok with that @josephschorr ?

josephschorr:
@vroldanbet I think so. It definitely should match and I think its okay to put a note about the changed `serve-testing` as it is, after all, only for testing

vroldanbet (approved):

PR #1751, opened 2024-02-21
reduces chunking allocations for wide relations

vroldanbet:
For most situations, when the number of elements to dispatch is below the default 100 elements, the slice with elements to dispatch was appropriately pre-allocated.

However, when dispatching large number of elements, the silence will grow from 1 up to the number of chunks, causing wasteful allocations in the critical path.

To simplify the logic, +1 is added to account for the situation the number of elements is less than the chunk size. For values above that it will cause one excess slice entry allocation, but seems like a reasonable tradeoff w.r.t extra annoying code to handle it.

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -383,7 +383,10 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 	it.Close()
 
 	// Convert the subjects into batched requests.
-	toDispatch := make([]directDispatch, 0, subjectsToDispatch.Len())
+	// To simplify the logic, +1 is added to account for the situation where
+	// the number of elements is less than the chunk size, and spare us some annoying code.
+	expectedNumberOfChunks := subjectsToDispatch.ValueLen()/int(crc.maxDispatchCount) + 1
+	toDispatch := make([]directDispatch, 0, expectedNumberOfChunks)
 	subjectsToDispatch.ForEachType(func(rr *core.RelationReference, resourceIds []string) {
 		chunkCount := 0.0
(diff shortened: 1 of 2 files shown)


internal/graph/check.go
```diff
@@ -383,7 +383,8 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 	it.Close()
 
 	// Convert the subjects into batched requests.
-	toDispatch := make([]directDispatch, 0, subjectsToDispatch.Len())
+	expectedNumberOfChunks := subjectsToDispatch.ValueLen()/int(crc.maxDispatchCount) + 1
```
jzelinskie, line 386: This definitely needs a comment along the lines of your PR description explaining the intention of the +1

vroldanbet: Done

internal/graph/check.go
```diff
@@ -383,7 +383,10 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 	it.Close()
 
 	// Convert the subjects into batched requests.
-	toDispatch := make([]directDispatch, 0, subjectsToDispatch.Len())
+	// To simplify the logic, +1 is added to account for the situation
```
josephschorr, line 386: situation where

internal/graph/check.go
```diff
@@ -601,7 +604,10 @@ func (cc *ConcurrentChecker) checkTupleToUserset(ctx context.Context, crc curren
...
```
josephschorr, line 607: ditto

josephschorr (approved):
LGTM

PR #1756, opened 2024-02-23
VSCode launch config

alecmerdler:
Adds the ability to run the `serve` command for SpiceDB using the VSCode debugger. Configuration will be read from an `.env.local` file.


josephschorr (approved):
LGTM

PR #1760, opened 2024-02-23
skip all steps for matrix jobs when the whole job should be skipped

ecordell:
when matrix jobs are required, skipping the whole job doesn't run them at all, which means required checks are missing from PRs


alecmerdler (approved):
Awesome, thanks!

PR #1764, opened 2024-02-23
Remove stale TODOs

josephschorr:
Most of the removed TODOs are changes that we won't end up making. In the case of the MySQL datastore, the Postgres datastore implementation from which it was copied has diverged significantly, so there isn't really a reason to try to refactor into a combined datastore anymore (since the PG one uses PG-specific code)

### DIFF
--- a/pkg/datastore/test/tuples.go
+++ b/pkg/datastore/test/tuples.go
@@ -384,8 +384,8 @@ func DeleteRelationshipsTest(t *testing.T, tester DatastoreTester) {
 
 			tRequire := testfixtures.TupleChecker{Require: require, DS: ds}
 
-			// TODO temporarily store tuples in multiple calls to ReadWriteTransaction since no Datastore
-			// handles correctly duplicate tuples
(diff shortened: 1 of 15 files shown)


pkg/datastore/test/tuples.go
```diff
@@ -384,8 +384,8 @@ func DeleteRelationshipsTest(t *testing.T, tester DatastoreTester) {
 
 			tRequire := testfixtures.TupleChecker{Require: require, DS: ds}
 
-			// TODO temporarily store tuples in multiple calls to ReadWriteTransaction since no Datastore
-			// handles correctly duplicate tuples
+			// NOTE: we tuples in multiple calls to ReadWriteTransaction because it is not allowed to change
```
ecordell, line 387: > we tuples

josephschorr: Ah, yes... I meant "I tuples". Good catch

(On a serious note: Fixed)

vroldanbet (approved):

vroldanbet (approved):

PR #1772, opened 2024-03-01
adds OpenTelemetry TraceID to logs

vroldanbet:
so we can correlate logs, traces and metric exemplars


ecordell (approved):
LGTM

I wonder if there's some way we could unify request IDs with traceIDs

vroldanbet:
>I wonder if there's some way we could unify request IDs with traceIDs

Yeah, I've had the a similar thought. We'd need to read a bit about Otel TraceIDs and determine if it has any restriction that would prevent us from feeding the `requestID` into it. Is that what you had in mind?

PR #1773, opened 2024-03-01
Move health check logs to debug level

vroldanbet:
The gRPC health check endpoint tends to cause noise in SpiceDB logs. This PR changes how the grpc log Middleware is installed so the grpc health route is logged at debug level.


jzelinskie:
Maybe we can set it so it only occurs when debug logging is set? When you first run something, it's super helpful to know if the healthcheck is reachable.

vroldanbet:
@jzelinskie please have another look. health check calls are now moved to debug level. It's not great because grpc middleware does not have native support for something like this, but the proposed approach gets the job done.

ecordell (approved):

ecordell (dismissed):
LGTM, though we'll want to test this and make sure everything logs as intended

PR #1792, opened 2024-03-13
CheckBulkPermissions

### LINKED ISSUE #1791: CheckBulkPermissions in SpiceDB
Implement the `PermissionsService/CheckBulkPermissions` API in SpiceDB.

alecmerdler:
Adds the `PermissionsService/CheckBulkPermissions` API method. This promotes the existing `ExperimentalService/BulkCheckPermission` API out of the experimental service, but does not remove the old method to ensure backwards compatibility.

To reduce code duplication, this refactors the bulk permission checking logic into a shared component that can be used by both the old and new methods. The old method performs protobuf message conversion on the input, passes the inputs to the shared method, and then converts the output.

Resolves #1791

### DIFF
--- a/internal/services/v1/bulkcheck.go
+++ b/internal/services/v1/bulkcheck.go
@@ -1,0 +1,251 @@
+package v1
+
+import (
+	"context"
+	"sync"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/jzelinskie/stringz"
+	"google.golang.org/grpc/status"
+
+	"github.com/authzed/spicedb/internal/dispatch"
+	"github.com/authzed/spicedb/internal/graph"
+	"github.com/authzed/spicedb/internal/graph/computed"
+	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
+	"github.com/authzed/spicedb/internal/middleware/usagemetrics"
+	"github.com/authzed/spicedb/internal/namespace"
+	"github.com/authzed/spicedb/internal/services/shared"
+	"github.com/authzed/spicedb/internal/taskrunner"
+	"github.com/authzed/spicedb/pkg/genutil"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
+	"github.com/authzed/spicedb/pkg/genutil/slicez"
+	"github.com/authzed/spicedb/pkg/middleware/consistency"
+	dispatchv1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+// bulkChecker contains the logic to allow ExperimentalService/BulkCheckPermission and
+// PermissionsService/CheckBulkPermissions to share the same implementation.
+type bulkChecker struct {
+	maxAPIDepth          uint32
+	maxCaveatContextSize int
+	maxConcurrency       uint16
+
+	dispatch dispatch.Dispatcher
+}
+
+func (bc *bulkChecker) checkBulkPermissions(ctx context.Context, req *v1.CheckBulkPermissionsRequest) (*v1.CheckBulkPermissionsResponse, error) {
+	atRevision, checkedAt, err := consistency.RevisionFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(req.Items) > maxBulkCheckCount {
+		return nil, NewExceedsMaximumChecksErr(uint64(len(req.Items)), maxBulkCheckCount)
+	}
+
+	// Compute a hash for each requested item and record its index(es) for the items, to be used for sorting of results.
+	itemCount, err := genutil.EnsureUInt32(len(req.Items))
+	if err != nil {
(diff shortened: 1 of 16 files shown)


internal/services/v1/bulkcheck.go
```diff
@@ -0,0 +1,249 @@
...
+	// Run the checks in parallel.
+	if err := tr.StartAndWait(); err != nil {
+		return nil, bc.rewriteError(ctx, err)
+	}
+
+	return &v1.CheckBulkPermissionsResponse{CheckedAt: checkedAt, Pairs: orderedPairs}, nil
+}
+
+func toCheckBulkPermissionsRequest(req *v1.BulkCheckPermissionRequest) *v1.CheckBulkPermissionsRequest {
+	items := make([]*v1.CheckBulkPermissionsRequestItem, len(req.Items))
```
josephschorr, line 204: you could also do `, 0, len(req.Items)` and `append`, but this works

internal/services/v1/bulkcheck.go
```diff
@@ -0,0 +1,249 @@
...
+	for i, pair := range resp.Pairs {
+		pairs[i] = &v1.BulkCheckPermissionPair{}
+		pairs[i].Request = &v1.BulkCheckPermissionRequestItem{
+			Resource:   pair.Request.Resource,
+			Permission: pair.Request.Permission,
+			Subject:    pair.Request.Subject,
+			Context:    pair.Request.Context,
+		}
+
+		switch pair.Response.(type) {
```
josephschorr, line 228: `switch t := pair.Response.(type) {` and then use `t.` below to avoid the casts

internal/services/v1/experimental.go
```diff
@@ -95,13 +82,18 @@ func NewExperimentalServer(dispatch dispatch.Dispatcher, permServerConfig Permis
...
-		defaultBatchSize:        uint64(config.DefaultExportBatchSize),
-		maxBatchSize:            uint64(config.MaxExportBatchSize),
-		dispatch:                dispatch,
-		maximumAPIDepth:         permServerConfig.MaximumAPIDepth,
-		maxCaveatContextSize:    permServerConfig.MaxCaveatContextSize,
-		bulkCheckMaxConcurrency: config.BulkCheckMaxConcurrency,
+		defaultBatchSize: uint64(config.DefaultExportBatchSize),
+		maxBatchSize:     uint64(config.MaxExportBatchSize),
 	}
+	es.bulkChecker = &bulkChecker{
```
josephschorr, line 88: assign this in the struct creation above

internal/services/v1/hash_test.go
```diff
@@ -419,11 +419,11 @@ func TestBulkCheckPermissionItemWithoutResourceIDHashStability(t *testing.T) {
 					},
 				},
 			},
-			"4fe0404757f60c7b",
+			"f518629690bd9dc0",
```
josephschorr, line 422: these hashes should not be changing?

internal/services/v1/relationships.go
```diff
@@ -103,6 +107,15 @@ func NewPermissionsServer(
 			),
 		},
 	}
+	ps.bulkChecker = &bulkChecker{
```
josephschorr, line 110: move into the struct above

internal/services/v1/bulkcheck.go
alecmerdler: This was a direct copy-paste from the previous implementation

internal/services/v1/experimental.go
alecmerdler: I would have done that if I didn't need to pass the `rewriteError()` method.

internal/services/v1/hash_test.go
josephschorr: May not be necessary to change

internal/services/v1/experimental.go
josephschorr: we should be rewriting errors at the call sites; so I'd remove that and just call rewriteError if/when an error occurs at the place where the shared method is invoked

alecmerdler: I tried that, but we have a [short-circuit in the bulk check implementation](https://github.com/authzed/spicedb/pull/1792/files#diff-4610fb4679843c5126b0e6b9a3b71f0649411314b9e122550f2506ec7cad1f9cR107-R112) which requires inspecting the result of `rewriteError()`; that's the only reason we are passing it in.

alecmerdler: Looking at the code, I guess we could just call `shared.RewriteError()` directly in the shared `checkBulkPermissions()` instead of passing in `rewriteError()`. Let me try that.

internal/services/v1/hash_test.go
alecmerdler: The goal in this pull request was to switch everything to using the new method and types and implement the old method in terms of the new method. Changing the call hashes to use the new method keeps the code consistent.

josephschorr: ok

josephschorr (approved):
LGTM

PR #1805, opened 2024-03-15
turns gRPC latency histogram into a toggleable option

vroldanbet:
so that custom builds of SpiceDB can decide to enable or disable it, as with the new version of the grpc-ecosystem prometheus middleware does not provide enough granularity.


josephschorr (approved):
LGTM

PR #1807, opened 2024-03-18
do not return backward incompatible `--explain` debug info in trailer

vroldanbet:
Previous change caused `zed` to fail to parse `--explain` trailer. This means new release causes clients to miss debug info trailer, but seems better than breaking them with an incompatible message.


josephschorr (dismissed):
Unfortunate, but necessary

josephschorr (dismissed):

bison (dismissed):

josephschorr (approved):

josephschorr (approved):

PR #1831, opened 2024-03-24
Use type information to optimize TOUCH operations in the PG datastore

josephschorr:
When a TOUCH is performed over a relationship for a relation+subject type that cannot be caveated, we can skip the UPDATE query used to check for the caveat changing, which is a significant performance boost


internal/datastore/postgres/readwrite.go
```diff
@@ -88,14 +90,88 @@ func appendForInsertion(builder sq.InsertBuilder, tpl *core.RelationTuple) sq.In
...
+	}
+
+	// Load the namespaces for any resources that are being TOUCHed and check if the relation being touched
+	// *can* have a caveat. If not, mark the relation as supported simplified TOUCH operations.
+	relationSupportSimplifiedTouch := mapz.NewSet[string]()
+	if touchedResourceNamespaces.IsEmpty() {
+		return relationSupportSimplifiedTouch, nil
+	}
+
+	namespaces, err := rwt.LookupNamespacesWithNames(ctx, touchedResourceNamespaces.AsSlice())
```
vroldanbet, line 109: So we are trading of an `UPDATE` for an extra `SELECT` operation.

internal/datastore/postgres/readwrite.go
```diff
@@ -88,14 +90,88 @@ func appendForInsertion(builder sq.InsertBuilder, tpl *core.RelationTuple) sq.In
...
+	if err != nil {
+		return nil, fmt.Errorf(errUnableToWriteRelationships, err)
+	}
+
+	if len(namespaces) == 0 {
```
vroldanbet, line 114: is this possible at all? We know `touchedResourceNamespaces` is not empty

internal/datastore/postgres/readwrite.go
```diff
@@ -88,14 +90,88 @@ func appendForInsertion(builder sq.InsertBuilder, tpl *core.RelationTuple) sq.In
...
+		return relationSupportSimplifiedTouch, nil
+	}
+
+	nsDefByName := make(map[string]*core.NamespaceDefinition, len(namespaces))
+	for _, ns := range namespaces {
+		nsDefByName[ns.Definition.Name] = ns.Definition
```
vroldanbet, line 120: do we need to keep the definition around instead of just the information we need (whether it has a caveat or not)?

internal/datastore/postgres/readwrite.go
```diff
@@ -88,14 +90,88 @@ func appendForInsertion(builder sq.InsertBuilder, tpl *core.RelationTuple) sq.In
...
+	// the existence of a relationship with a different caveat name and/or context which might need to be
+	// replaced.
+	relationSupportSimplifiedTouch, err := rwt.collectSimplifiedTouchTypes(ctx, mutations)
+	if err != nil {
+		return err
+	}
+
 	// Parse the updates, building inserts for CREATE/TOUCH and deletes for DELETE.
 	for _, mut := range mutations {
 		tpl := mut.Tuple
```
vroldanbet, line 177: All these operations happen within the span of a pgx transaction. `rwt.collectSimplifiedTouchTypes` is looping over the mutations 2 times, which could be up to 1000 elements max by default, keeping the transaction open. Could we leverage this loop to compute it? Is it worth writing a benchmark?

internal/datastore/postgres/readwrite.go
josephschorr: A much smaller `SELECT` operation, since it is only loading the namespaces needed

internal/datastore/postgres/readwrite.go
josephschorr: Most of the time, no, but if someone passes in invalid tuples, yes. Since the checks occur *before* WriteRels is called, we need to check again here

internal/datastore/postgres/readwrite.go
josephschorr: We determine the information by using the definition

internal/datastore/postgres/readwrite.go
josephschorr: I doubt 1000 iterations vs 2000 iterations of an in-mem slice will make an appreciable difference; the real extra delay will be loading the namespaces

internal/datastore/postgres/readwrite.go
vroldanbet: yeah, was jsut suggesting to keep what we needed in the value of the map, not the whole definition

vroldanbet (approved):

Issue #1842, opened 2024-03-28
Failure to delete relationships when not allowing partial deletions


jmadureira:
### What platforms are affected?

macos

### What architectures are affected?

arm64

### What SpiceDB version are you using?

v.1.30.0

### Steps to Reproduce

The following snippet is Java code but I'll try to explain each step.

Tests are run against postgreSQL and SpiceDB not running with `serve-testing`.

```java
public void test() {
        var testSchema = """
                definition identity/user {}

                definition test/organisation {
                  relation member: identity/user
                                
                  permission view = member
                }
                """;
        Function<Integer, Core.RelationshipUpdate> toRelationshipRequest = (i) -> {
            return Core.RelationshipUpdate.newBuilder()
                    .setRelationship(Core.Relationship.newBuilder()
                            .setResource(Core.ObjectReference.newBuilder()
                                    .setObjectType("test/organisation")
                                    .setObjectId(String.valueOf(i))
                                    .build())
                            .setRelation("member")
                            .setSubject(Core.SubjectReference.newBuilder()
                                    .setObject(Core.ObjectReference.newBuilder()
                                            .setObjectType("identity/user")
                                            .setObjectId("1")
                                            .build())
                                    .build())
                            .build())
                    .setOperation(Core.RelationshipUpdate.Operation.OPERATION_CREATE)
                    .build();
        };
        Function<Integer, PermissionService.CheckPermissionRequest> toCheckPermissionRequest = (i) -> {
            return PermissionService.CheckPermissionRequest.newBuilder()
                    .setResource(Core.ObjectReference.newBuilder()
                            .setObjectType("test/organisation")
                            .setObjectId(String.valueOf(i))
                            .build())
                    .setPermission("view")
                    .setSubject(Core.SubjectReference.newBuilder()
                            .setObject(Core.ObjectReference.newBuilder()
                                    .setObjectType("identity/user")
                                    .setObjectId("1")
                                    .build())
                            .build())
                    .setConsistency(PermissionService.Consistency.newBuilder().setFullyConsistent(true).build())
                    .build();
        };
        var schemaRequest = WriteSchemaRequest.newBuilder().setSchema(testSchema).build();
        // load the test schema
        assertThat(schemaService.writeSchema(schemaRequest)).succeedsWithin(Duration.ofSeconds(2));
        // create 100 test relationships
        var relationships = IntStream.range(0, 100).boxed().map(toRelationshipRequest).toList();
        writeRelationships(relationships);
[...]

PR #1843, opened 2024-03-29
Fix re-creating deleted relationships

alecmerdler:
Ensures it is possible to re-create relationships that were previously deleted using filters. This bug was caused by not including the `created_xid` column during deletion.

Addresses #1842


josephschorr (approved):
LGTM

PR #1845, opened 2024-04-01
Improve request-id propagation

vroldanbet:
SpiceDB supports a custom `request-id` header clients can use to trace requests. It's convenient, for example, when troubleshooting certain requests using `zed permission check --request-id <my-request-id>`.

Unfortunately `request-id` wasn't being properly propagated to:
- `pgx` driver logs for Postgres-based SpiceDB
- dispatched requests

These issues are fixed in this PR, making it possible to filter all the log entries of a request by a custom `request-id`, included those that are part emitted as part of SpiceDB's dispatch mechanism.

### DIFF
--- a/pkg/middleware/requestid/requestid.go
+++ b/pkg/middleware/requestid/requestid.go
@@ -41,7 +41,39 @@ type handleRequestID struct {
 	requestIDGenerator IDGenerator
 }
 
+func (r *handleRequestID) ClientReporter(ctx context.Context, meta interceptors.CallMeta) (interceptors.Reporter, context.Context) {
+	haveRequestID, requestID, ctx := r.requestIDFromContext(ctx)
+
+	if haveRequestID {
+		ctx = requestmeta.SetRequestHeaders(ctx, map[requestmeta.RequestMetadataHeaderKey]string{
+			requestmeta.RequestIDKey: requestID,
+		})
+	}
+
+	return interceptors.NoopReporter{}, ctx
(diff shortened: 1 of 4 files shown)


pkg/middleware/requestid/requestid.go
```diff
@@ -73,7 +73,25 @@ func (r *handleRequestID) ServerReporter(ctx context.Context, _ interceptors.Cal
 	return interceptors.NoopReporter{}, ctx
 }
 
-func (r *handleRequestID) requestIDFromContext(ctx context.Context) (bool, string, context.Context) {
+func (r *handleRequestID) fromContextOrGenerate(ctx context.Context) (bool, string, context.Context) {
+	haveRequestID, requestID, md := fromContext(ctx)
+
+	if !haveRequestID && r.generateIfMissing {
+		requestID, haveRequestID = r.requestIDGenerator(), true
```
josephschorr, line 80: why this form?

alecmerdler: Agreed, we don't use multiple assignments on the same line usually in this codebase.

vroldanbet: that style was already in place before the PR, but I agree.
Addressed in https://github.com/authzed/spicedb/pull/1845/commits/4e732a92f4ba831c60885c4d27ae5760741ac725

josephschorr (approved):
LGTM

PR #1849, opened 2024-04-01
Correct version requirement for datastore repair

alecmerdler:
The `spicedb datastore repair transaction-ids` command requires Postgres v15+ to succeed.


josephschorr (approved):
LGTM

PR #1850, opened 2024-04-01
LookupResources Postgres query optimization

alecmerdler:
Fixes potential slow reverse lookup queries when using Postgres by changing the order of the columns being compared in the cursor. By comparing the numerical IDs before the string relations, the query is able to short-circuit sooner and return faster.

### DIFF
--- a/internal/datastore/common/sql.go
+++ b/internal/datastore/common/sql.go
@@ -148,6 +148,7 @@ type nameAndValue struct {
 }
 
 func (sqf SchemaQueryFilterer) After(cursor *core.RelationTuple, order options.SortOrder) SchemaQueryFilterer {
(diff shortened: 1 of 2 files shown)


internal/graph/lookupresources.go
```diff
@@ -85,7 +85,7 @@ func (cl *CursoredLookupResources) LookupResources(
 		reachableResourcesCursor = newCursor
 
 		// If no additional reachable results were found or the request was unlimited, then we can stop.
-		if reachableCount == 0 || req.OptionalLimit == 0 {
+		if reachableCount == 0 {
```
josephschorr, line 88: don't remove this change; its useful independent of the above fix

josephschorr (approved):
LGTM

PR #1858, opened 2024-04-07
add support for AWS IAM authentication for postgres

j-white:
Here we add support for AWS IAM authentication for the postgres datastore

This fixes https://github.com/authzed/spicedb/issues/659

Verified this works in two different cases:
 1) Tested the `datastore migrate head` command via shell with SSO authentication
 2) Tested the `serve` command on ECS (Fargate) with a task role that has the `rds-db:connect` permission

In each case we're able to load the AWS credentials, generate an IAM token for the DB and authenticate successfully

PR #1867, opened 2024-04-16
add aws iam authentication for mysql

j-white:
Building off of https://github.com/authzed/spicedb/pull/1858, adds similar support for MySQL

Marking this as a draft since I haven't validated on AWS yet


internal/datastore/mysql/datastore.go
```diff
@@ -118,7 +118,27 @@ func newMySQLDatastore(ctx context.Context, uri string, options ...Option) (*Dat
...
+		err := parsedURI.Apply(mysql.BeforeConnect(func(ctx context.Context, config *mysql.Config) error {
+			config.User, config.Passwd, err = credentialsProvider.Get(ctx, config.Addr, config.User)
+			return err
+		}))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	connector, err := mysql.NewConnector(parsedURI)
```
j-white, line 141: We create the connector using `mysql.NewConnector(parsedURI)` instead of `mysql.MySQLDriver{}.OpenConnector(uri)`, so we can preserve the `BeforeConnect` hook we added

internal/datastore/mysql/migrations/driver.go
```diff
@@ -35,16 +37,29 @@ type MySQLDriver struct {
...
+		err := dbConfig.Apply(sqlDriver.BeforeConnect(func(ctx context.Context, config *sqlDriver.Config) error {
+			config.User, config.Passwd, err = credentialsProvider.Get(ctx, config.Addr, config.User)
+			return err
+		}))
+		if err != nil {
+			return nil, fmt.Errorf(errUnableToInstantiate, err)
+		}
+	}
+
+	connector, err := sqlDriver.NewConnector(dbConfig)
```
j-white, line 57: Similarly here, we create the db using `sqlDriver.NewConnector(dbConfig)` and `db := sql.OpenDB(connector)` instead of `sql.Open("mysql", dbConfig.FormatDSN())`, so we can preserve the `BeforeConnect` hook we added

internal/datastore/mysql/datastore.go
```diff
@@ -118,7 +118,29 @@ func newMySQLDatastore(ctx context.Context, uri string, options ...Option) (*Dat
...
+	if config.credentialsProviderName != "" {
+		credentialsProvider, err = datastore.NewCredentialsProvider(ctx, config.credentialsProviderName)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if credentialsProvider != nil {
+		// we must transmit the token over the connection, and not a hash
+		parsedURI.AllowCleartextPasswords = true
```
josephschorr, line 132: newline after this line.

Also, will this be the case for all credentials providers or just this one? We should log a warning here too

internal/datastore/mysql/migrations/driver.go
```diff
@@ -35,16 +37,32 @@ type MySQLDriver struct {
...
+func NewMySQLDriverFromDSN(url string, tablePrefix string, credentialsProvider datastore.CredentialsProvider) (*MySQLDriver, error) {
 	dbConfig, err := sqlDriver.ParseDSN(url)
 	if err != nil {
 		return nil, fmt.Errorf(errUnableToInstantiate, err)
 	}
 
-	db, err := sql.Open("mysql", dbConfig.FormatDSN())
+	if credentialsProvider != nil {
+		log.Debug().Str("name", credentialsProvider.Name()).Msg("using credentials provider")
+		// we must transmit the token over the connection, and not a hash
```
josephschorr, line 48: newline before each comment here

internal/datastore/mysql/migrations/driver.go
```diff
@@ -35,16 +37,32 @@ type MySQLDriver struct {
...
+		dbConfig.AllowCleartextPasswords = true
+		// add a before connect callback to trigger the token retrieval from the credentials provider
+		err := dbConfig.Apply(sqlDriver.BeforeConnect(func(ctx context.Context, config *sqlDriver.Config) error {
```
josephschorr, line 51: Can we move this shared code into a helper function?

internal/datastore/postgres/options.go
```diff
@@ -337,6 +337,10 @@ func MigrationPhase(phase string) Option {
 	return func(po *postgresOptions) { po.migrationPhase = phase }
 }
 
+// CredentialsProviderName the name of the CredentialsProvider implementation to use
+// for dynamically retrieving the datastore credentials at runtime
+//
+// Disabled by default.
```
josephschorr, line 343: Empty by default

josephschorr (commented):
Can we add a datastore test for MySQL to use it as well?

internal/datastore/mysql/datastore_test.go
```diff
@@ -711,6 +711,23 @@ func TestMySQLMigrationsWithPrefix(t *testing.T) {
 	req.NoError(rows.Err())
 }
 
+func TestMySQLWithAWSIAMCredentialsProvider(t *testing.T) {
```
j-white, line 714: here's a take at adding a test for this

alternatively, we could expect authentication to fail if we try connecting to the actual test datastore

internal/datastore/mysql/datastore.go
j-white: Good point. Say there is a new credentials provider that pulls the "normal" password from something like Vault, then this would not need to be set.

It looks like this can be set in the connection string as well using `allowCleartextPasswords=true`, so another option is to remove this and require the user to set it when they use AWS IAM auth. If they miss it it's not obvious though since the server just rejects it with access denied - had to pull out tcpdump to catch this one.

josephschorr: If it will always apply to the AWS creds provider, I say we return another param from the call that indicates whether the token is plaintext (or not) and switch based on that

j-white: Sound good. I opted to add a flag on the `CredentialsProvider` since we need to know this ahead of retrieving the actual token.

josephschorr:
Please rebase as well

josephschorr (approved):
LGTM

PR #1868, opened 2024-04-16
lsp: implement didChange and fix logging

jzelinskie:
- textDocdidChange events previous threw errors in certain clients
- nil pointer when logging crashed the entire process

### DIFF
--- a/internal/lsp/lsp.go
+++ b/internal/lsp/lsp.go
@@ -4,6 +4,7 @@ package lsp
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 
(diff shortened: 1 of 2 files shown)


internal/lsp/lsp.go
```diff
@@ -45,11 +46,18 @@ func (s *Server) Handle(ctx context.Context, conn *jsonrpc2.Conn, r *jsonrpc2.Re
 	jsonrpc2.HandlerWithError(s.handle).Handle(ctx, conn, r)
 }
 
+func logJSONPtr(msg *json.RawMessage) string {
```
jzelinskie, line 49: longer term: this should probably be a top-level zerolog abstraction built into our logging package

PR #1870, opened 2024-04-17
Include doc comments in resolver generated source

josephschorr:


pkg/development/resolver.go
```diff
@@ -143,9 +144,15 @@ func (r *Resolver) ReferenceAtPosition(source input.Source, position input.Posit
 			ColumnPosition: int(def.SourcePosition.ZeroIndexedColumnPosition),
 		}
 
-		targetSourceCode := fmt.Sprintf("definition %s {\n\t// ...\n}", def.Name)
+		docComment := ""
+		comments := namespace.GetComments(def.Metadata)
+		if len(comments) > 0 {
+			docComment = strings.Join(comments, "\n") + "\n"
```
jzelinskie, line 150: seems weird to add the final new line here rather than just in sprintf calls

josephschorr: Well, we don't want it present if there are no comments, so I add it here

jzelinskie (approved):

PR #1871, opened 2024-04-18
Add configurable limits for all APIs

josephschorr:


internal/services/v1/relationships.go
```diff
@@ -73,6 +73,22 @@ type PermissionsServerConfig struct {
...
+	// in a single DeleteRelationships call.
+	MaxDeleteRelationshipsLimit uint32
+
+	// MaxLookupResourcesLimit defines the maximum number of resources that can be looked up in a
+	// single LookupResources call.
+	MaxLookupResourcesLimit uint32
+
+	// MaxBulkExportRelationshipsLimit defines the maximum number of relationships that can be
+	// exported in a single BulkExportRelationships call.
+	MaxBatchExportRelationshipsLimit uint32
```
vroldanbet, line 91: should use the terminology of the API
```suggestion
	MaxBulkExportRelationshipsLimit uint32
```

internal/services/v1/relationships.go
```diff
@@ -81,13 +97,17 @@ func NewPermissionsServer(
...
+		MaxUpdatesPerWrite:               defaultIfZero(config.MaxUpdatesPerWrite, 1000),
+		MaximumAPIDepth:                  defaultIfZero(config.MaximumAPIDepth, 50),
+		StreamingAPITimeout:              defaultIfZero(config.StreamingAPITimeout, 30*time.Second),
+		MaxCaveatContextSize:             defaultIfZero(config.MaxCaveatContextSize, 4096),
+		MaxRelationshipContextSize:       defaultIfZero(config.MaxRelationshipContextSize, 25_000),
+		MaxDatastoreReadPageSize:         defaultIfZero(config.MaxDatastoreReadPageSize, 1_000),
+		MaxReadRelationshipsLimit:        defaultIfZero(config.MaxReadRelationshipsLimit, 1_000),
+		MaxDeleteRelationshipsLimit:      defaultIfZero(config.MaxDeleteRelationshipsLimit, 1_000),
+		MaxLookupResourcesLimit:          defaultIfZero(config.MaxLookupResourcesLimit, 1_000),
+		MaxBatchExportRelationshipsLimit: defaultIfZero(config.MaxBatchExportRelationshipsLimit, 100_000),
```
vroldanbet, line 110: didn't you use 10_000 as the new default in `internal/services/v1/experimental.go`?

internal/services/v1/relationships_test.go
vroldanbet, line 1: do we have an equivalent test for delete, where we assert optional_limit does not exceed the server-side configured limit?

vroldanbet (commented):
overal LGTM, observed some inconsistencies to address

internal/services/v1/relationships.go
josephschorr: whoops.

internal/services/v1/relationships_test.go
josephschorr: Yeah, we already had it

internal/services/v1/relationships.go
josephschorr: And I set the parameter to that value; this was the fallback default

vroldanbet (approved):

PR #1883, opened 2024-05-02
Improvements around usage of CEL

josephschorr:
Upgrades CEL to get structured errors out, allowing us to rewrite the location information on the errors instead of using the synthesized whitespace hack

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -16,8 +16,8 @@ require (
 )
 
 require (
-	github.com/antlr/antlr4/runtime/Go/antlr/v4 v4.0.0-20230512164433-5d1fd1a340c9 // indirect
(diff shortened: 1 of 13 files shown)


ecordell:
Haven't reviewed yet, but the CEL fork is needed for now - just update the fork if you want to update CEL.

Kube also imports CEL but at an older version, and the APIs changed backwards incompatibly. The fork is how we keep spicedb importable in spicedb-kubeapi-proxy. So not really a fork, it's a mirror.

josephschorr:
Ah... nertz. Okay, I'll update the mirror tomorrow and then update the commit to still move to the latest version. I'll also add a note in the code

vroldanbet (approved):

ecordell (approved):

PR #1886, opened 2024-05-03
refactor bulk export relationships logic

vroldanbet:
for a potential future use that bypasses the gRPC server

note: trivy build started failing because it flagged the spicedb image built by goreleaser as containing CVEs, because it's parsing the version in the ldflags, and infers its 0.0.1, which is the version the currently being generated in this pipeline. This was probably introduced by trivy 0.51.0, when it started parsing go binaries ldflags: https://github.com/aquasecurity/trivy/pull/6564. The right way to address this would be to generate the correct version, or something that does not follow semver.


josephschorr (approved):
LGTM

PR #1887, opened 2024-05-03
fetch git tags so that trivy sees the right binary version

vroldanbet:
this fixes an issue with trivy where it flags SpiceDB as vulnerable, possibly as of https://github.com/aquasecurity/trivy/pull/6564 include in version 0.51.0. It's flagged because it parses it as version 0.0.1-next.

Because the `checkout` action does not fetch the tags by default, goreleaser is unable the generate the right version.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -42,6 +42,14 @@ jobs:
     runs-on: "buildjet-2vcpu-ubuntu-2204"
     steps:
       - uses: "actions/checkout@v4"
+        with:
+          # Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0
+          # to fetch all history for all branches and tags. Refer here to learn which commit $GITHUB_SHA
+          # points to for different events.
+          #
+          # this is used so goreleaser generates the right version out of the tags, which we need so that
+          # trivy does not flag an old SpiceDB version
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


ecordell (dismissed):

.github/workflows/security.yaml
```diff
@@ -42,6 +42,8 @@ jobs:
     runs-on: "buildjet-2vcpu-ubuntu-2204"
     steps:
       - uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 0
```
ecordell, line 46: should this be fetch-tags or does this work?

jzelinskie: >Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer [here](https://docs.github.com/actions/using-workflows/events-that-trigger-workflows) to learn which commit $GITHUB_SHA points to for different events.

We should add a comment to this line

vroldanbet: >should this be fetch-tags or does this work?

@ecordell I tried that first. For some reason that didn't work. Could be an issue with the GH action, so I went the easy route

vroldanbet: https://github.com/actions/checkout/issues/1471

ecordell (approved):

PR #1888, opened 2024-05-03
expose BulkExportRelationships service controller logic

vroldanbet:
so that it can be used without the grpc server

### DIFF
--- a/pkg/services/v1/services.go
+++ b/pkg/services/v1/services.go
@@ -1,0 +1,17 @@
+package v1
+
+import (
+	"context"
+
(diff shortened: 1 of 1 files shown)


pkg/services/v1/services.go
```diff
@@ -0,0 +1,16 @@
+package v1
+
+import (
+	"context"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	servicesv1 "github.com/authzed/spicedb/internal/services/v1"
```
ecordell, line 7: lint error on the go import spacing

ecordell (approved):

PR #1892, opened 2024-05-08
Switch spanner datastore to use the built-in stats table for estimating rel count

josephschorr:
While this will be far less accurate of an estimate, it removes the need to write to a stats table on every write and delete, which should help with performance

### DIFF
--- a/internal/datastore/spanner/readwrite.go
+++ b/internal/datastore/spanner/readwrite.go
@@ -20,8 +20,7 @@ import (
 
 type spannerReadWriteTXN struct {
 	spannerReader
-	spannerRWT   *spanner.ReadWriteTransaction
-	disableStats bool
+	spannerRWT *spanner.ReadWriteTransaction
 }
 
 const inLimit = 10_000 // https://cloud.google.com/spanner/quotas#query-limits
@@ -36,12 +35,6 @@ func (rwt spannerReadWriteTXN) WriteRelationships(ctx context.Context, mutations
 		rowCountChange += countChange
 
 		if err := rwt.spannerRWT.BufferWrite([]*spanner.Mutation{txnMut}); err != nil {
-			return fmt.Errorf(errUnableToWriteRelationships, err)
-		}
-	}
-
-	if !rwt.disableStats {
-		if err := updateCounter(ctx, rwt.spannerRWT, rowCountChange); err != nil {
 			return fmt.Errorf(errUnableToWriteRelationships, err)
 		}
 	}
@@ -74,7 +67,7 @@ func spannerMutation(
 }
 
 func (rwt spannerReadWriteTXN) DeleteRelationships(ctx context.Context, filter *v1.RelationshipFilter, opts ...options.DeleteOptionsOption) (bool, error) {
-	limitReached, err := deleteWithFilter(ctx, rwt.spannerRWT, filter, rwt.disableStats, opts...)
+	limitReached, err := deleteWithFilter(ctx, rwt.spannerRWT, filter, opts...)
 	if err != nil {
 		return false, fmt.Errorf(errUnableToDeleteRelationships, err)
 	}
@@ -82,7 +75,7 @@ func (rwt spannerReadWriteTXN) DeleteRelationships(ctx context.Context, filter *
 	return limitReached, nil
 }
 
-func deleteWithFilter(ctx context.Context, rwt *spanner.ReadWriteTransaction, filter *v1.RelationshipFilter, disableStats bool, opts ...options.DeleteOptionsOption) (bool, error) {
+func deleteWithFilter(ctx context.Context, rwt *spanner.ReadWriteTransaction, filter *v1.RelationshipFilter, opts ...options.DeleteOptionsOption) (bool, error) {
 	delOpts := options.NewDeleteOptionsWithOptionsAndDefaults(opts...)
 	var delLimit uint64
 	if delOpts.DeleteLimit != nil && *delOpts.DeleteLimit > 0 {
@@ -94,24 +87,18 @@ func deleteWithFilter(ctx context.Context, rwt *spanner.ReadWriteTransaction, fi
 
 	var numDeleted int64
 	if delLimit > 0 {
-		nu, err := deleteWithFilterAndLimit(ctx, rwt, filter, disableStats, delLimit)
+		nu, err := deleteWithFilterAndLimit(ctx, rwt, filter, delLimit)
 		if err != nil {
 			return false, err
 		}
 		numDeleted = nu
 	} else {
-		nu, err := deleteWithFilterAndNoLimit(ctx, rwt, filter, disableStats)
+		nu, err := deleteWithFilterAndNoLimit(ctx, rwt, filter)
 		if err != nil {
 			return false, err
 		}
 
 		numDeleted = nu
-	}
-
-	if !disableStats {
-		if err := updateCounter(ctx, rwt, -1*numDeleted); err != nil {
-			return false, err
-		}
 	}
 
 	if delLimit > 0 && uint64(numDeleted) == delLimit {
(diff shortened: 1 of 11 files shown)


pkg/cmd/datastore/datastore.go
```diff
@@ -213,6 +214,7 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	flagSet.StringVar(&opts.SpannerEmulatorHost, flagName("datastore-spanner-emulator-host"), "", "URI of spanner emulator instance used for development and testing (e.g. localhost:9010)")
 	flagSet.Uint64Var(&opts.SpannerMinSessions, flagName("datastore-spanner-min-sessions"), 100, "minimum number of sessions across all Spanner gRPC connections the client can have at a given time")
 	flagSet.Uint64Var(&opts.SpannerMaxSessions, flagName("datastore-spanner-max-sessions"), 400, "maximum number of sessions across all Spanner gRPC connections the client can have at a given time")
+	flagSet.Uint64Var(&opts.SpannerEstimatedBytesPerRelationship, flagName("datastore-spanner-estimated-bytes-per-relationship"), spanner.DefaultEstimatedBytesPerRelationship, "estimated number of bytes per relationship tuple in the spanner instance")
```
alecmerdler, line 217: This does seem like a strange flag to have the user provide. While I understand the motivation of performance and simplifying the code, it feels clunky because it requires internal knowledge of how Spanner stores data on disk.

Even just querying Spanner for the size of a random stored relationship tuple when `Statistics()` is called would be probably be good enough to estimate the number of relationships, instead of providing this flag.

josephschorr: It is odd, true. I added it as a flag so users can set it IF they want more accurate estimates, not that the estimates are used for much as-is :)

alecmerdler: Given how many datastore-specific flags we already have, avoiding adding more when it isn't absolutely necessary will reduce potential tech debt if we decide to consolidate the flags in the future.

josephschorr: Think I should just hard code in then? Alternatively, we could make it load a random relationship and estimate based on that?

josephschorr: Changed to load some relationships and calculate the estimated size based on them, with a fallback if none found

alecmerdler (approved):
Nice work!

internal/datastore/spanner/readwrite.go
```diff
@@ -40,12 +39,6 @@ func (rwt spannerReadWriteTXN) WriteRelationships(ctx context.Context, mutations
 		}
 	}
 
-	if !rwt.disableStats {
```
vroldanbet, line 43: maybe we want to log a warning in the datastore initialization indicating the flag is basically ignored with spanner

internal/datastore/spanner/spanner.go
```diff
@@ -205,7 +216,7 @@ func (sd spannerDatastore) SnapshotReader(revisionRaw datastore.Revision) datast
 	return spannerReader{executor, txSource}
 }
 
-func (sd spannerDatastore) ReadWriteTx(ctx context.Context, fn datastore.TxUserFunc, opts ...options.RWTOptionsOption) (datastore.Revision, error) {
+func (sd *spannerDatastore) ReadWriteTx(ctx context.Context, fn datastore.TxUserFunc, opts ...options.RWTOptionsOption) (datastore.Revision, error) {
```
vroldanbet, line 219: make sure all method receivers in `spannerDatastore` are pointer receivers

internal/datastore/spanner/stats.go
```diff
@@ -46,57 +51,86 @@ func (sd spannerDatastore) Statistics(ctx context.Context) (datastore.Stats, err
...
-	// nolint:gosec
-	// G404 use of non cryptographically secure random number generator is not concern here,
-	// as this is only used to randomly distributed the counters across multiple rows and reduce write contention
-	_, err := rand.New(rng).Read(counterID)
-	if err != nil {
-		return fmt.Errorf("unable to select random counter: %w", err)
-	}
+	// If there is not yet a cached estimated bytes per relationship, read a few relationships and then
+	// compute the average bytes per relationship.
+	sd.cachedEstimatedBytesPerRelationshipLock.RLock()
```
vroldanbet, line 56: do we know how often is the `Statistics` method polled? I think it's by default 1h right? Are we sure it's not called from anywhere else?

internal/datastore/spanner/stats.go
```diff
@@ -46,57 +51,86 @@ func (sd spannerDatastore) Statistics(ctx context.Context) (datastore.Stats, err
...
-
-	counterID := make([]byte, 2)
-	// nolint:gosec
-	// G404 use of non cryptographically secure random number generator is not concern here,
-	// as this is only used to randomly distributed the counters across multiple rows and reduce write contention
-	_, err := rand.New(rng).Read(counterID)
-	if err != nil {
-		return fmt.Errorf("unable to select random counter: %w", err)
-	}
+	// If there is not yet a cached estimated bytes per relationship, read a few relationships and then
```
vroldanbet, line 54: why not `sync.Once`?

internal/datastore/spanner/readwrite.go
josephschorr: Meh, I don't think its necessary

internal/datastore/spanner/spanner.go
josephschorr: I did a global find-replace

josephschorr: And linter would warn

internal/datastore/spanner/stats.go
josephschorr: Its not but since there aren't any guarantees it won't be, I ensured it was locked

internal/datastore/spanner/stats.go
josephschorr: Because we only set it if we get a proper result which means it *may* need to be called a few times until it settles

internal/datastore/spanner/stats.go
```diff
@@ -46,57 +51,101 @@ func (sd spannerDatastore) Statistics(ctx context.Context) (datastore.Stats, err
...
-		return datastore.Stats{}, fmt.Errorf("unable to read row counts: %w", err)
-	}
-
-	return datastore.Stats{
-		UniqueID:                   uniqueID,
-		ObjectTypeStatistics:       datastore.ComputeObjectTypeStats(allNamespaces),
-		EstimatedRelationshipCount: uint64(estimate.Int64),
-	}, nil
-}
+	fmt.Println("starting stats")
```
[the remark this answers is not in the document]
josephschorr: Whoops, removed

internal/datastore/spanner/stats.go
```diff
@@ -46,57 +51,101 @@ func (sd spannerDatastore) Statistics(ctx context.Context) (datastore.Stats, err
...
+			totalByteCount += relationshipByteCount
+
+			return nil
+		}); err != nil {
+			return datastore.Stats{}, err
+		}
 
-func updateCounter(ctx context.Context, rwt *spanner.ReadWriteTransaction, change int64) error {
-	newValue := change
+		fmt.Println("got total relationships", totalRelationships, "total byte count", totalByteCount)
```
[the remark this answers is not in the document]
josephschorr: Whoops, removed

internal/datastore/spanner/spanner_test.go
```diff
@@ -38,5 +43,105 @@ func TestSpannerDatastore(t *testing.T) {
...
+CREATE TABLE fake_stats_table (
+  interval_end TIMESTAMP,
+  table_name STRING(MAX),
+  used_bytes INT64,
+) PRIMARY KEY (table_name, interval_end)
+`
+
+func FakeStatsTest(t *testing.T, ds datastore.Datastore) {
+	spannerDS := ds.(*spannerDatastore)
+	spannerDS.tableSizesStatsTable = "fake_stats_table"
```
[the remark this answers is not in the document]
josephschorr: Unfortunately not: 

> Many tables related to runtime introspection in the SPANNER_SYS schema (e.g., query stats tables) are not supported.

See:  https://github.com/GoogleCloudPlatform/cloud-spanner-emulator

josephschorr:
Updated

vroldanbet (approved):

PR #1894, opened 2024-05-09
Add ExperimentalComputablePermissions API

josephschorr:


internal/services/v1/experimental.go
```diff
@@ -497,6 +497,81 @@ func (es *experimentalServer) ExperimentalDiffSchema(ctx context.Context, req *v
 	return resp, nil
 }
 
+func (es *experimentalServer) ExperimentalComputablePermissions(ctx context.Context, req *v1.ExperimentalComputablePermissionsRequest) (*v1.ExperimentalComputablePermissionsResponse, error) {
```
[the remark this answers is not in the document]
vroldanbet: I'm also thinking there is an opportunity to support providing multiple namespace+relations pairs. If an update comes via Watch API and a transaction changes various relations, it could become very chatty over the wire to check each relation one by one.

internal/services/v1/experimental_test.go
```diff
@@ -1426,3 +1426,204 @@ func TestExperimentalDependentRelations(t *testing.T) {
...
+		{
+			name:           "invalid relation",
+			schema:         `definition user {}`,
+			definitionName: "user",
+			relationName:   "invalid",
+			expectedCode:   codes.FailedPrecondition,
+			expectedError:  "relation/permission `invalid` not found",
+		},
+		{
+			name: "basic",
```
vroldanbet, line 1457: While I do understand this expected response from the perspective of the reachability graph (`user:...`), it seems inconsistent for an API called `ExperimentalComputablePermissions`. I wouldn't expect this to return relations, and so it also wouldn't make sense the API has a field `IsPermission`.

It's not the same with `ExperimentalDependentRelations` because the name of the RPC explicitly states "Relations" so it fits both relations and permissions.

The options that come to mind are:
- create a type `ExPermissionReference` that drops that field, and filter for permissions in the controller
- rename the RPC to something like `ExperimentalComputableRelations` as it could return both.

pkg/typesystem/reachabilitygraph_test.go
```diff
@@ -17,6 +17,215 @@ import (
...
+
+				permission edit = editor
+				permission view = viewer + edit
+			}`,
+
+			"user",
+			"...",
+
+			[]string{"document#view", "document#viewer"},
+		},
```
vroldanbet, line 169: add test that uses subject relation with a computed userset

pkg/typesystem/reachabilitygraph_test.go
```diff
@@ -17,6 +17,215 @@ import (
...
+	}
```
vroldanbet, line 170: also add test where there is recursion in the schema. Also add to the other API tests

pkg/typesystem/reachabilitygraph.go
```diff
@@ -142,6 +144,78 @@ func (rg *ReachabilityGraph) RelationsEncounteredForResource(
...
+) ([]*core.RelationReference, error) {
+	if startingSubjectType.Namespace != rg.ts.nsDef.Name {
+		return nil, spiceerrors.MustBugf("gave mismatching namespace name for subject type to reachability graph")
+	}
+
+	allRelationNames := mapz.NewSet[string]()
+
+	subjectTypesToCheck := []*core.RelationReference{startingSubjectType}
+
+	// TODO(jschorr): optimize this to not require walking over all types recursively.
```
vroldanbet, line 161: We may want to add some sort of max depth to prevent infinite recursion that could happen via the schema? Perhaps adding a context deadline if this is taking too long as a precaution?

pkg/typesystem/reachabilitygraph.go
```diff
@@ -142,6 +144,78 @@ func (rg *ReachabilityGraph) RelationsEncounteredForResource(
 	return relationRefs, nil
 }
 
+// RelationsEncounteredForSubject returns all relations that are encountered when walking outward from a subject+relation.
+func (rg *ReachabilityGraph) RelationsEncounteredForSubject(
+	ctx context.Context,
+	allDefinitions []*core.NamespaceDefinition,
```
vroldanbet, line 150: nit: does this need to be "all definitions" or it could be any subset? The argument name would invite to think it has to be provided with all namespaces in the instance

pkg/typesystem/reachabilitygraph.go
```diff
@@ -142,6 +144,78 @@ func (rg *ReachabilityGraph) RelationsEncounteredForResource(
...
+		}
+
+		collected := &[]ReachabilityEntrypoint{}
+		for _, nsDef := range allDefinitions {
+			nts, err := rg.ts.TypeSystemForNamespace(ctx, nsDef.Name)
+			if err != nil {
+				return nil, err
+			}
+
+			nrg := ReachabilityGraphFor(&ValidatedNamespaceTypeSystem{nts})
```
vroldanbet, line 175: nit: shouldn't `TypeSystemForNamespace` return a `ValidatedNamespaceTypeSystem` then?

vroldanbet (commented):
LGTM overall, the only potential blocker I see is lack of tests and mitigation for schema recursion. I'd also like to see us supporting batched calls to both ComputablePermissions and DependentRelations, as I think that could be desirable for scenarios where you want to use these APIs for denormalization purposes.

internal/services/v1/experimental.go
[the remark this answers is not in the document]
josephschorr: I didn't think it was necessary, but we could add it as a followup. As for the multiple pairs, the expectation is that the caller will cache this information

internal/services/v1/experimental.go
```diff
@@ -497,6 +497,81 @@ func (es *experimentalServer) ExperimentalDiffSchema(ctx context.Context, req *v
...
+		return nil, shared.RewriteErrorWithoutConfig(ctx, err)
+	}
+
+	relations := make([]*v1.ExpRelationReference, 0, len(rr))
+	for _, r := range rr {
+		if r.Namespace == req.DefinitionName && r.Relation == req.RelationName {
+			continue
+		}
+
+		if req.OptionalDefinitionNameFilter != "" && !strings.HasPrefix(r.Namespace, req.OptionalDefinitionNameFilter) {
```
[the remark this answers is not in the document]
josephschorr: The filter applies to the results, so we can't do that

internal/services/v1/experimental_test.go
josephschorr: I left it named permissions because that's how people will primarily use it; we could rename it, but I felt it was more discoverable that way. Both functions return both.

pkg/typesystem/reachabilitygraph_test.go
josephschorr: We have one though?

pkg/typesystem/reachabilitygraph_test.go
josephschorr: We have one; a group refers to its own members

pkg/typesystem/reachabilitygraph.go
josephschorr: This is a structural walk, not a a data-driven one; there should be no case where it causes infinite recursion as the check below reduces the working set over time

pkg/typesystem/reachabilitygraph.go
josephschorr: Its everything

pkg/typesystem/reachabilitygraph.go
josephschorr: No; we know (internally here) that it is already validated but if the type system was loaded from memory, it may not have been validated yet

internal/services/v1/experimental.go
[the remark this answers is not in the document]
vroldanbet: 👍🏻 Let's make that explicit in the docs for now then

pkg/typesystem/reachabilitygraph_test.go
vroldanbet: oh I may missed it, just to be clear, something like `relation members: team#members` where members is a permission

vroldanbet (approved):

internal/services/v1/experimental_test.go
vroldanbet: It could confuse folks, but it's not a big deal anyway, just my 2 cents

Issue #1903, opened 2024-05-17
Optionally return operation statuses on WriteRelationships


benvernier-sc:
### Problem Statement

The SpiceDB docs cover a [Two writes & commits](https://authzed.com/docs/spicedb/concepts/relationships#two-writes--commit) scenario to do dual writes between an existing system and SpiceDB. Those docs suggest performing a `WriteRelationships` call as part of the primary write path for the data, then performing a `DeleteRelationships` request if a rollback if necessary.

The issue is that if I had a relationship that already existed, I performed a `WriteRelationships` request with a `TOUCH` operation on that existing relationship (which would have been a no-op internally) then want to rollback my operation, calling `DeleteRelationships` would actually leave my system in a different state than what it was in before I started the operation.

Similarly, if one of the operations of my `WriteRelationships` request was a `DELETE` operation but the relationship didn't actually exist in SpiceDB, I do not want to create the relationship on rollback as once again this would leave the system in a different state.

### Solution Brainstorm

When making a call to `WriteRelationships` (with either a single or multiple updates), there are cases where it would be useful to be able to get what the status of each operation was in the response.

In particular, knowing whether a `TOUCH` operation actually created a relationship or if it was a no-op, and knowing whether a `DELETE` operation actually deleted a relationship or if it was a no-op.

I acknowledge that systematically returning that information might come at a performance cost that existing use cases that don't need this information might not be comfortable with, so this could be an option set on the request to specify whether the status breakdown should be returned.

benvernier-sc:
@josephschorr I did a quick PoC for this:
- Suggested API change: https://github.com/authzed/api/pull/106
- Implementation of this API change for Postgres datastore: https://github.com/authzed/spicedb/pull/1906

Let me know your thoughts, if you're happy with the approach I can extend the implementation to cover the other datastores  and actually raise a PR for the change.

josephschorr:
@benvernier-sc Seems reasonable at first glance, although we'd need to clean up a bit the conditionals used around whether to add the `RETURNING` on not. 

I did a quick check through the other datastores:
- MySQL: uses `select for update`, so should be trivial to determine which rows were changed
- CRDB: supports `returning` so should be similar to the Postgres driver changes, although it will have the same performance impact
- Memdb: doesn't matter the performance impact, and trivial to implement
- Spanner: **⚠️ PROBLEM** - I'm unsure if we can do it in Spanner, which presents a problem around API compatibility. This will require more research

[...]

PR #1906, opened 2024-05-21
Optionally return status on WriteRelationships

benvernier-sc:
This PR is part of a proof of concept for https://github.com/authzed/spicedb/issues/1903.

It temporarily uses a `replace` statement in `go.mod` to use the files generated in https://github.com/benvernier-sc/authzed-go/tree/write_relationships_status from https://github.com/authzed/api/pull/106. 

It defines a new `core` protobuf message `RelationTupleUpdateStatus` that corresponds to the `RelationshipUpdateStatus` from the API protobuf. 

It then updates the signature of `ReadWriteTransaction.WriteRelationships` from
```go
type ReadWriteTransaction interface {
    // [...]
    WriteRelationships(ctx context.Context, mutations []*core.RelationTupleUpdate) error
    // [...]
}
```
to
```go
type ReadWriteTransaction interface {
    // [...]
    WriteRelationships(ctx context.Context, mutations []*core.RelationTupleUpdate, returnStatus bool) ([]*core.RelationTupleUpdateStatus, error)
    // [...]
}
```

This allows to specify whether the store should return the status of the operations or not, and adds a return field with said statuses that are only guaranteed to be populated if `returnStatus` was true.

All the usages of `WriteRelationships` store method were updated to ensure the code would still compile. The `WriteRelationships` gRPC handler was updated to actually pass the `WithStatus` option from the request and use the returned statuses.

So far only the Postgres implementation has been updated to honor the new options. All scenarios have been tested locally, but no automated tests have been written so far.


josephschorr:
@benvernier-sc Decided this approach didn't work?

PR #1913, opened 2024-05-23
goreleaser: use build.head? in install

jzelinskie:
This avoids any issues with ruby versions and File.exist.


josephschorr (approved):
LGTM

Issue #966, opened 2022-11-01
PROPOSAL: Add optional watch annotations to RelationshipUpdate


reify-tanner-stirrat:
## The Problem Statement
As the developer of an authorization service that uses SpiceDB as its backing store, for regulatory reasons, I need to have an audit log that shows what relationships were present in SpiceDB at what time in order to answer an auditor's question of "who had access to what at what time, and who authorized the change?"

## Proposed Solution
We can use the `WatchService` to answer "who had access to what at what time" by taking the stream of events and saving it to some durable persistence mechanism (Kafka in our case). However, additional information about the transaction (such as who authorized the change to SpiceDB) isn't currently stored on the relation and wouldn't show up in the `WatchService`.

The `WatchService` as it currently exists would be a desirable mechanism for audit because it's an in-order log of changes made to SpiceDB, which is our system of record and source of truth.

Attaching a notion of metadata to write requests seems like it could satisfy the above. As long as metadata that goes in with a `WriteRelationships` request comes out in the `WatchService`, it wouldn't even really need to be persisted to the backing store beyond the persistence mechanisms already present in the `WatchService`.

### Concretely
```protobuf
message RelationshipUpdate {
  enum Operation {
    OPERATION_UNSPECIFIED = 0;
    OPERATION_CREATE = 1;
    OPERATION_TOUCH = 2;
    OPERATION_DELETE = 3;
  }
  Operation operation = 1 [ (validate.rules).enum = {defined_only: true, not_in: [0]} ];
  Relationship relationship = 2 [ (validate.rules).message.required = true ];
  Struct watch_annotations = 3;
}
```
The understanding is that these annotations are *only* for consumption via the watch service, and are not intended for any other logical or semantic purpose.

The naming and location are a suggestion; if there's another location or name that provides the same functionality I'm all ears.

## Existing Alternatives
My team is currently discussing a few different ideas, all of which have drawbacks. We have changestreams coming from both web requests to the service and from a kafka consumer listening to changestreams from elsewhere in our system.

One is to write audit events before and after attempting a write to SpiceDB. This has the downside of leaving potentially indeterminate states in the audit log, especially if there are overlapping requests to our web interface that interact with the same object. It won't be clear which request landed first.

Another is to put a queue in the request path, which would help from the perspectives of durability and serializability, but comes at considerable complexity cost.

reify-tanner-stirrat:
One alternative that we've come up with that wouldn't be heinous is adding something like
```
definition audit_subject {}
[...]

PR #1914, opened 2024-05-26
Implement support for metadata associated with read-write transactions

josephschorr:
This will allow callers of APIs such as WriteRelationships and DeleteRelationships to assign metadata to the transaction that will be mirrored back out in the Watch API, to provide a means for correlating updates

First part of #966

Each datastore has a different means of correlating between the transaction and the associated change(s) coming out of the watch API:

- Memdb: The metadata is stored on the changes struct
- Postgres: A `metadata` column was added to the transactions table
- MySQL: A `metadata` column was added to the transactions table
- Spanner: A transaction tag is used to correlate the transaction and the metadata is stored in an auto-GC-ing table, indexed by the transaction tag
- CRDB: A row is inserted into a metadata table as part of the transaction, which includes the metadata. When seen at the same revision as the other changes, the metadata becomes correlated with that transaction. The rows are set to auto-GC after a period of time in CRDB >= v22

### DIFF
--- a/internal/datastore/common/changes.go
+++ b/internal/datastore/common/changes.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 
 	"golang.org/x/exp/maps"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/ccoveille/go-safecast"
 
@@ -37,6 +38,7 @@ type changeRecord[R datastore.Revision] struct {
 	definitionsChanged map[string]datastore.SchemaDefinition
 	namespacesDeleted  map[string]struct{}
 	caveatsDeleted     map[string]struct{}
+	metadata           map[string]any
 }
 
 // NewChanges creates a new Changes object for change tracking and de-duplication.
@@ -132,6 +134,25 @@ func (ch *Changes[R, K]) adjustByteSize(item sized, delta int) error {
 	return nil
 }
 
+// SetRevisionMetadata sets the metadata for the given revision.
+func (ch *Changes[R, K]) SetRevisionMetadata(ctx context.Context, rev R, metadata map[string]any) error {
+	if len(metadata) == 0 {
+		return nil
+	}
+
+	record, err := ch.recordForRevision(rev)
+	if err != nil {
+		return err
+	}
+
+	if len(record.metadata) > 0 {
+		return spiceerrors.MustBugf("metadata already set for revision")
+	}
+
+	maps.Copy(record.metadata, metadata)
+	return nil
+}
+
 func (ch *Changes[R, K]) recordForRevision(rev R) (changeRecord[R], error) {
 	k := ch.keyFunc(rev)
 	revisionChanges, ok := ch.records[k]
@@ -143,6 +164,7 @@ func (ch *Changes[R, K]) recordForRevision(rev R) (changeRecord[R], error) {
 			make(map[string]datastore.SchemaDefinition),
 			make(map[string]struct{}),
 			make(map[string]struct{}),
+			make(map[string]any),
 		}
 		ch.records[k] = revisionChanges
 	}
@@ -248,21 +270,25 @@ func (ch *Changes[R, K]) AddChangedDefinition(
 
 // AsRevisionChanges returns the list of changes processed so far as a datastore watch
 // compatible, ordered, changelist.
-func (ch *Changes[R, K]) AsRevisionChanges(lessThanFunc func(lhs, rhs K) bool) []datastore.RevisionChanges {
+func (ch *Changes[R, K]) AsRevisionChanges(lessThanFunc func(lhs, rhs K) bool) ([]datastore.RevisionChanges, error) {
 	return ch.revisionChanges(lessThanFunc, *new(R), false)
 }
 
 // FilterAndRemoveRevisionChanges filters a list of changes processed up to the bound revision from the changes list, removing them
 // and returning the filtered changes.
-func (ch *Changes[R, K]) FilterAndRemoveRevisionChanges(lessThanFunc func(lhs, rhs K) bool, boundRev R) []datastore.RevisionChanges {
-	changes := ch.revisionChanges(lessThanFunc, boundRev, true)
+func (ch *Changes[R, K]) FilterAndRemoveRevisionChanges(lessThanFunc func(lhs, rhs K) bool, boundRev R) ([]datastore.RevisionChanges, error) {
+	changes, err := ch.revisionChanges(lessThanFunc, boundRev, true)
+	if err != nil {
+		return nil, err
+	}
+
 	ch.removeAllChangesBefore(boundRev)
-	return changes
+	return changes, nil
 }
 
-func (ch *Changes[R, K]) revisionChanges(lessThanFunc func(lhs, rhs K) bool, boundRev R, withBound bool) []datastore.RevisionChanges {
+func (ch *Changes[R, K]) revisionChanges(lessThanFunc func(lhs, rhs K) bool, boundRev R, withBound bool) ([]datastore.RevisionChanges, error) {
 	if ch.IsEmpty() {
-		return nil
+		return nil, nil
 	}
 
 	revisionsWithChanges := make([]K, 0, len(ch.records))
@@ -273,7 +299,7 @@ func (ch *Changes[R, K]) revisionChanges(lessThanFunc func(lhs, rhs K) bool, bou
 	}
 
 	if len(revisionsWithChanges) == 0 {
-		return nil
+		return nil, nil
 	}
 
 	sort.Slice(revisionsWithChanges, func(i int, j int) bool {
@@ -299,9 +325,18 @@ func (ch *Changes[R, K]) revisionChanges(lessThanFunc func(lhs, rhs K) bool, bou
 		changes[i].ChangedDefinitions = maps.Values(revisionChangeRecord.definitionsChanged)
 		changes[i].DeletedNamespaces = maps.Keys(revisionChangeRecord.namespacesDeleted)
 		changes[i].DeletedCaveats = maps.Keys(revisionChangeRecord.caveatsDeleted)
+
+		if len(revisionChangeRecord.metadata) > 0 {
+			metadata, err := structpb.NewStruct(revisionChangeRecord.metadata)
+			if err != nil {
+				return nil, spiceerrors.MustBugf("failed to convert metadata to structpb: %v", err)
+			}
+
+			changes[i].Metadata = metadata
+		}
 	}
 
-	return changes
+	return changes, nil
 }
 
 func (ch *Changes[R, K]) removeAllChangesBefore(boundRev R) {
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -54,6 +54,7 @@ const (
 	tableTransactions        = "transactions"
 	tableCaveat              = "caveat"
 	tableRelationshipCounter = "relationship_counter"
+	tableTransactionMetadata = "transaction_metadata"
 
 	colNamespace      = "namespace"
 	colConfig         = "serialized_config"
@@ -79,6 +80,8 @@ const (
 	colCounterSerializedFilter = "serialized_filter"
 	colCounterCurrentCount     = "current_count"
 	colCounterUpdatedAt        = "updated_at_timestamp"
+	colExpiresAt               = "expires_at"
+	colMetadata                = "metadata"
 
 	errUnableToInstantiate = "unable to instantiate datastore"
 	errRevision            = "unable to find revision: %w"
@@ -207,6 +210,7 @@ func newCRDBDatastore(ctx context.Context, url string, options ...Option) (datas
 		analyzeBeforeStatistics: config.analyzeBeforeStatistics,
 		filterMaximumIDCount:    config.filterMaximumIDCount,
 		supportsIntegrity:       config.withIntegrity,
+		gcWindow:                config.gcWindow,
 	}
 	ds.RemoteClockRevisions.SetNowFunc(ds.headRevisionInternal)
 
@@ -289,6 +293,7 @@ type crdbDatastore struct {
 	writeOverlapKeyer       overlapKeyer
 	overlapKeyInit          func(ctx context.Context) keySet
 	analyzeBeforeStatistics bool
+	gcWindow                time.Duration
 
 	beginChangefeedQuery string
 	transactionNowQuery  string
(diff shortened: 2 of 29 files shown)


josephschorr:

josephschorr:
recheck

internal/datastore/postgres/revisions.go
```diff
@@ -234,11 +234,20 @@ func parseRevisionDecimal(revisionStr string) (datastore.Revision, error) {
...
 }
 
-func createNewTransaction(ctx context.Context, tx pgx.Tx) (newXID xid8, newSnapshot pgSnapshot, err error) {
+func createNewTransaction(ctx context.Context, tx pgx.Tx, metadata map[string]any) (newXID xid8, newSnapshot pgSnapshot, err error) {
 	ctx, span := tracer.Start(ctx, "createNewTransaction")
 	defer span.End()
 
-	cterr := tx.QueryRow(ctx, createTxn).Scan(&newXID, &newSnapshot)
+	if metadata == nil {
+		metadata = map[string]any{}
```
vroldanbet, line 242: let's create a global var for this so we don't allocate (I'm actually unsure if this allocates, but just in case)

internal/datastore/mysql/revisions.go
```diff
@@ -156,16 +154,26 @@ func (mds *Datastore) checkValidTransaction(ctx context.Context, revisionTx uint
 	return freshEnough.Bool, unknown.Bool, nil
 }
 
-func (mds *Datastore) createNewTransaction(ctx context.Context, tx *sql.Tx) (newTxnID uint64, err error) {
+func (mds *Datastore) createNewTransaction(ctx context.Context, tx *sql.Tx, metadata map[string]any) (newTxnID uint64, err error) {
 	ctx, span := tracer.Start(ctx, "createNewTransaction")
 	defer span.End()
 
-	createQuery := mds.createTxn
+	var wrappedMetadata caveatContextWrapper
```
vroldanbet, line 161: maybe rename `caveatContextWrapper` to `structpbWrapper`?

internal/datastore/spanner/spanner.go
```diff
@@ -12,6 +12,7 @@ import (
 	"cloud.google.com/go/spanner"
 	ocprom "contrib.go.opencensus.io/exporter/prometheus"
 	sq "github.com/Masterminds/squirrel"
+	"github.com/google/uuid"
```
vroldanbet, line 15: same suggestion as with CRDB: we can have Spanner generate the UUID  value and spare the network transfer, using `DEFAULT (GENERATE_UUID()),`. See https://cloud.google.com/spanner/docs/schema-design#uuid_primary_key

EDIT: I see this won't work here because we need to provide the transaction tag in the Spanner protocol 👍🏻

internal/datastore/spanner/migrations/zz_migration.0010_add_transaction_metadata_table.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"context"
+
+	"cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
+)
+
+const (
+	addTransactionMetadataTable = `CREATE TABLE transaction_metadata (
+			transaction_tag STRING(MAX) NOT NULL,
```
vroldanbet, line 11: if the tag is a UUID, use `STRING(36)` as the documentation suggests: https://cloud.google.com/spanner/docs/primary-key-default-value#universally_unique_identifier_uuid

internal/datastore/spanner/migrations/zz_migration.0010_add_transaction_metadata_table.go
```diff
@@ -0,0 +1,34 @@
...
+			created_at TIMESTAMP DEFAULT (CURRENT_TIMESTAMP()),
+			metadata JSON
+		) PRIMARY KEY (transaction_tag),
+		ROW DELETION POLICY (OLDER_THAN(created_at, INTERVAL 2 DAY))
```
vroldanbet, line 15: same feedback as with CRDB around TTL configuration.  Does `interval 2 day` means "GC every second day"?

internal/datastore/spanner/spanner.go
```diff
@@ -221,18 +222,50 @@ func (sd *spannerDatastore) SnapshotReader(revisionRaw datastore.Revision) datas
...
+	return metadata, nil
+}
+
 func (sd *spannerDatastore) ReadWriteTx(ctx context.Context, fn datastore.TxUserFunc, opts ...options.RWTOptionsOption) (datastore.Revision, error) {
 	config := options.NewRWTOptionsWithOptions(opts...)
 
 	ctx, span := tracer.Start(ctx, "ReadWriteTx")
 	defer span.End()
 
+	transactionTag := fmt.Sprintf("spicedb-%s", uuid.NewString())
```
vroldanbet, line 249: concatenate instead of `fmt`. Also why prepending `spicedb-`, seems unnecessary waste and UUID should guarantee they are unique, and we already correlate them with the table.

internal/datastore/spanner/spanner.go
```diff
@@ -252,15 +285,15 @@ func (sd *spannerDatastore) ReadWriteTx(ctx context.Context, fn datastore.TxUser
 		}
 
 		return nil
-	})
+	}, spanner.TransactionOptions{TransactionTag: transactionTag})
```
vroldanbet, line 288: any specific limitations around tagging a transaction? e.g. number of them? How long do they stay around in the DB? Do we have an understanding of how it affects write throughput?

Should we check if `config.Metadata` is present outside of the RTW not to tag transactions unless requested by the customer? I assume tagging is not free, and not every Spanner user may be using transaction metadata feature.

internal/datastore/common/changes.go
```diff
@@ -81,6 +83,22 @@ func (ch *Changes[R, K]) AddRelationshipChange(
...
 
+// SetRevisionMetadata sets the metadata for the given revision.
+func (ch *Changes[R, K]) SetRevisionMetadata(ctx context.Context, rev R, metadata map[string]any) error {
+	if len(metadata) == 0 {
+		return nil
+	}
+
+	record := ch.recordForRevision(rev)
+	if len(record.metadata) > 0 {
+		log.Ctx(ctx).Warn().Msg("metadata already set for revision")
```
vroldanbet, line 94: superfluous log entry as we are going to log the error anyway

vroldanbet (commented):
Left a few questions and suggestions, I think the approach makes sense to me. The main concerns are how does this affect write-throughput, but at first glance it shouldn't introduce any new hotspots.

A potential issue here is the overlap with the concept of tuple-level metadata, but we've certainly heard more about this use case than tuple-level metadata. I also do wonder if folks will eventually ask for a transactions API to read these back, but that's an easy thing to do if the time comes.


Finally: I understand these are just the Datastore API changes, but we'd have to make sure there are limits put in place so that folks don't abuse that new field.

internal/datastore/crdb/crdb.go
```diff
@@ -310,6 +312,23 @@ func (cds *crdbDatastore) ReadWriteTx(
 			Executor: pgxcommon.NewPGXExecutor(querier),
 		}
 
+		// If metadata is to be attached, write that row now.
+		if config.Metadata != nil {
+			randomKey := uuid.NewString()
+			expiresAt := time.Now().Add(25 * time.Hour)
```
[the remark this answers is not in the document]
josephschorr: Updated to `gcWindow` + 1 minute

internal/datastore/crdb/migrations/zz_migration.0007_add_transaction_metadata_table.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"strings"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	addTransactionMetadataTableQuery = `
+		CREATE TABLE transaction_metadata (
+			key VARCHAR PRIMARY KEY,
```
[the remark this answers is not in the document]
josephschorr: Changed type but can't set the default

internal/datastore/mysql/query_builder.go
```diff
@@ -99,8 +99,8 @@ func getLastRevision(tableTransaction string) sq.SelectBuilder {
 	return sb.Select("MAX(id)").From(tableTransaction).Limit(1)
 }
 
-func getRevisionRange(tableTransaction string) sq.SelectBuilder {
-	return sb.Select("MIN(id)", "MAX(id)").From(tableTransaction)
+func loadRevisionRange(tableTransaction string) sq.SelectBuilder {
```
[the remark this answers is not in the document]
josephschorr: It wasn't being used so I just repurposed the name

internal/datastore/spanner/spanner.go
josephschorr: Yeah, unfortunately

internal/datastore/spanner/migrations/zz_migration.0010_add_transaction_metadata_table.go
josephschorr: It means after 2 days

internal/datastore/spanner/spanner.go
josephschorr: Because we might end up using transaction tags of other forms, so I want a prefix

internal/datastore/spanner/spanner.go
josephschorr: It shouldn't have any impact, since I believe Spanner just sets it to empty by default

josephschorr:
Updated

josephschorr:
Rebased

tstirrat15 (approved):
LGTM, see comment

internal/datastore/postgres/revisions.go
```diff
@@ -235,11 +235,22 @@ func parseRevisionDecimal(revisionStr string) (datastore.Revision, error) {
 	}}, nil
 }
 
-func createNewTransaction(ctx context.Context, tx pgx.Tx) (newXID xid8, newSnapshot pgSnapshot, err error) {
+var emptyMetadata = map[string]any{}
```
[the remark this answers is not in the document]
josephschorr: Usually no, but we usually put NULL in if we can

internal/datastore/spanner/migrations/zz_migration.0010_add_transaction_metadata_table.go
```diff
@@ -0,0 +1,34 @@
...
+	"cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
+)
+
+const (
+	addTransactionMetadataTable = `CREATE TABLE transaction_metadata (
+			transaction_tag STRING(36) NOT NULL,
+			created_at TIMESTAMP DEFAULT (CURRENT_TIMESTAMP()),
+			metadata JSON
+		) PRIMARY KEY (transaction_tag),
+		ROW DELETION POLICY (OLDER_THAN(created_at, INTERVAL 2 DAY))
```
vroldanbet, line 15: Right now revisions will be around for an hour, which is the [default value](https://cloud.google.com/spanner/docs/timestamp-bounds#maximum_timestamp_staleness).

However, the way clients will retrieve the data is via the Watch API, and the table deltas are materialized into a change stream that has longer retention, [one day by default](https://github.com/authzed/spicedb/issues/1457). 

So I'm confused as to why we let the metadata sit around for 2 days, they will only be able to retrieve it for up to 1 day. Something like 25 hours should be sufficient.

internal/datastore/spanner/spanner.go
```diff
@@ -239,18 +240,50 @@ func (sd *spannerDatastore) SnapshotReader(revisionRaw datastore.Revision) datas
...
+	rs, err := sd.client.ReadWriteTransactionWithOptions(ctx, func(ctx context.Context, spannerRWT *spanner.ReadWriteTransaction) error {
 		txSource := func() readTX {
 			return &traceableRTX{delegate: spannerRWT}
 		}
 
+		if config.Metadata != nil {
+			// Insert the metadata into the transaction metadata table.
+			mutation := spanner.Insert(tableTransactionMetadata,
+				[]string{colTransactionTag, colMetadata},
+				[]any{transactionTag, config.Metadata.AsMap()},
```
vroldanbet, line 279: are you planning to add limit in the size of the metadata at the service level? (i.e. validate this in `WriteRelationships` and `DeleteRelationships`)?

internal/datastore/postgres/revisions.go
```diff
@@ -250,11 +250,22 @@ func parseRevisionDecimal(revisionStr string) (datastore.Revision, error) {
...
+func createNewTransaction(ctx context.Context, tx pgx.Tx, metadata map[string]any) (newXID xid8, newSnapshot pgSnapshot, err error) {
 	ctx, span := tracer.Start(ctx, "createNewTransaction")
 	defer span.End()
 
-	cterr := tx.QueryRow(ctx, createTxn).Scan(&newXID, &newSnapshot)
+	if metadata == nil {
+		metadata = emptyMetadata
+	}
+
+	sql, args, err := createTxn.Values(metadata).Suffix("RETURNING xid, snapshot").ToSql()
```
vroldanbet, line 263: Please turn `RETURNING xid, snapshot` into a constant that is created out of `colXID` and `colSnapshot` instead of hardcoding column names

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	if strings.Contains(version, "v22.1") {
+		if _, err := conn.Exec(ctx, addTransactionMetadataTableQueryWithBasicTTL); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	if strings.Contains(version, "v24.") {
```
vroldanbet, line 63: Once we move to CRDB v25, this won't work and it will fall back to the behaviour in CRDB 23 and v22.
We need semver comparison logic, ideally, we parse the version and then use the semver library (I assume such library must exist). It could be turn into a small helper over a pgx.Conn.

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	// ttl_expiration_expression support was added in CRDB v22.2, but the E2E tests
+	// use v21.2.
```
vroldanbet, line 12: Can you clarify what's the difference between the TTL support in 22.2 vs 21.2?

internal/datastore/spanner/migrations/zz_migration.0010_add_transaction_metadata_table.go
josephschorr: Only days is supported, so it had to be 1 day or 2:

> num_days is the number of days past the timestamp in the timestamp_column in which the row is marked for deletion. The value must be a non-negative integer and DAY is the only supported unit.

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
josephschorr: The `ttl_expiration_expression` property was added in 22.2

internal/datastore/spanner/migrations/zz_migration.0010_add_transaction_metadata_table.go
vroldanbet: Please add a comment to clarify this then.

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
vroldanbet: yeah but the pre 22.2 version has a different expression, what does it do?

internal/datastore/spanner/spanner.go
josephschorr: It would be done at the API level; that has not been added to the API yet (it will be done in a followup PR)

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
josephschorr: Switched to semver

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
josephschorr: It just expires after 1 day, vs using the column's value

vroldanbet (approved):

Issue #498, opened 2022-03-23
Set GOMAXPROCS based on Kubernetes CPUQuotas


jzelinskie:
There are tools like https://github.com/uber-go/automaxprocs that enable setting GOMAXPROCS dynamically based on the deployment environment.

From talking to Uber engineers, I get the impression that his library might not be maintained much in the future (cgroups v2) and the more promising solution is to wait for https://github.com/golang/go/issues/33803

ecordell:
The kube scheduler can change the visible number of CPUs visible to a running container (without restarting it). (I'm looking for explicit docs to back up this claim, but I'm pretty sure this is the case based on how other features work). 

Strawman scenario that  I believe can happen and would be bad with automaxprocs:

- SpiceDB container starts and detects 2 cpus, runs with GOMAXPROCS=2
- A "gauranteed" QoS class container gets scheduled on the same node, which steals 1 whole cpu away
- SpiceDB container is running with maxprocs=2, but actually only has 1 core available
- SpiceDB has tons of contention and performs badly

I think we could just take advantage of this ourselves:
https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/

By setting requests=limits=<a whole number of CPUs> on spicedb containers, we should get dedicated CPUs for spicedb to use. But then it's not super valuable to use automaxprocs (just set GOMAXPROCS env on the pod to equal the cpu request).

I'm not 100% sure how available cpu management features are in various cloud flavors of kube or if there are more caveats to be aware of.

PR #1921, opened 2024-06-03
adds automaxprocs and automemlimit

vroldanbet:
closes https://github.com/authzed/spicedb/issues/498

so that SpiceDB becomes aware of cgroup cpu and
memory limits, and automatically
adjusts the go runtime gomaxprocs and gomemlimit
parameters


vroldanbet:

cmd/spicedb/main.go
```diff
@@ -3,15 +3,18 @@ package main
...
+	"github.com/KimMachineGun/automemlimit/memlimit"
 	mcobra "github.com/muesli/mango-cobra"
 	"github.com/muesli/roff"
 	"github.com/rs/zerolog"
+	slogzerolog "github.com/samber/slog-zerolog/v2"
 	"github.com/sercand/kuberesolver/v5"
 	"github.com/spf13/cobra"
+	"go.uber.org/automaxprocs/maxprocs"
 	"google.golang.org/grpc/balancer"
-	_ "google.golang.org/grpc/xds"
```
ecordell, line 14: Return this please! This is what enables `xds` support for the hashring.

(We don't have tests for it which is why nothing failed - we should though)

ecordell (approved):
https://github.com/authzed/spicedb/issues/498#issuecomment-1247036753

^ this comment still stands, but I think for the most part it's not a huge deal because you can always set GOMAXPROCs explicitly, or you can roll the pod to pick up the new available values. It might be worth documenting this behavior though, since I don't think there are any kube docs for it.

cmd/spicedb/main.go
vroldanbet: shoot, sorry!

ecordell (approved):

PR #1928, opened 2024-06-07
Return a proper error code if a wildcard subject is specified

### LINKED ISSUE #1925: Error for use of wildcard in check is not well-typed
It should be an `INVALID_ARGUMENT`

josephschorr:
Fixes #1925


internal/services/shared/errors.go
```diff
@@ -176,8 +176,6 @@ func RewriteError(ctx context.Context, err error, config *ConfigForErrors) error
 	case errors.As(err, &datastore.ErrCounterNotRegistered{}):
 		return spiceerrors.WithCodeAndReason(err, codes.FailedPrecondition, v1.ErrorReason_ERROR_REASON_COUNTER_NOT_REGISTERED)
 
-	case errors.As(err, &graph.ErrInvalidArgument{}):
```
vroldanbet, line 179: Why removing the case, isn't it being potentially used elsewhere upstream?

josephschorr: I removed the error type

vroldanbet (approved):

PR #1930, opened 2024-06-08
Add the debug trace to the details of the recursion error

### LINKED ISSUE #1927: No debug info / tracing info when maximum depth exceeded (from SpiceDB version v1.27.0)

others

others

`v1.27.0` and above - No debug info / tracing info when maximum depth exceeded
`v1.26.0` - Debug info / tracing info exist when maximum depth exceeded

### Steps to Reproduce

1. Use the following schema:
    ```
            definition user {
                relation resource: resource;
                permission test = resource->test;
            }
            definition resource {
                relation user: user;
                permission test = user->test;
            }
    ```
2. Add relationships:
    ```
    zed relationship touch resource:1 user user:1
    zed relationship touch user:1 resource resource:1
    ```
3. Check for permissions with explain flag:
    ```
    zed permission check resource:1 test user:5 --explain
    ```

### Expected Result

```
11:06AM INF debugging requested on check
! resource:1 test (42.804084ms)
└── ! user:1 test (41.176917ms)
    └── ! resource:1 test (cycle) (38.891292ms)

11:06AM ERR terminated with errors error="rpc error: code = ResourceExhausted desc = the check request has exceeded the allowable maximum depth of 50: this usually indicates a recursive or too deep data dependency. Try running zed with --explain to see the dependency. See: https://spicedb.dev/d/debug-max-depth-check"
```

### Actual Result

```
11:08AM INF debugging requested on check
11:08AM WRN No debuging information returned for the check
11:08AM ERR terminated with errors error="rpc error: code = ResourceExhausted desc = the check request has exceeded the allowable maximum depth of 50: this usually indicates a recursive or too deep data dependency. Try running zed with --explain to see the dependency. See: https://spicedb.dev/d/debug-max-depth-check"
```

josephschorr:
Fixes #1927

### DIFF
--- a/internal/dispatch/singleflight/singleflight.go
+++ b/internal/dispatch/singleflight/singleflight.go
@@ -8,8 +8,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"resenje.org/singleflight"
@@ -82,21 +80,16 @@ func (d *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckReq
 		return d.delegate.DispatchCheck(ctx, req)
 	}
 
-	primary := false
-	v, isShared, err := d.checkGroup.Do(ctx, keyString, func(innerCtx context.Context) (*v1.DispatchCheckResponse, error) {
-		primary = true
+	sharedResp, isShared, err := d.checkGroup.Do(ctx, keyString, func(innerCtx context.Context) (*v1.DispatchCheckResponse, error) {
 		return d.delegate.DispatchCheck(innerCtx, req)
 	})
 
-	singleFlightCount.WithLabelValues("DispatchCheck", strconv.FormatBool(isShared)).Inc()
-	if err != nil {
-		return &v1.DispatchCheckResponse{Metadata: &v1.ResponseMeta{DispatchCount: 1}}, err
+	if sharedResp == nil {
+		sharedResp = &v1.DispatchCheckResponse{Metadata: &v1.ResponseMeta{DispatchCount: 1}}
 	}
 
-	span := trace.SpanFromContext(ctx)
-	singleflighted := isShared && !primary
-	span.SetAttributes(attribute.Bool("singleflight", singleflighted))
-	return v, err
+	singleFlightCount.WithLabelValues("DispatchCheck", strconv.FormatBool(isShared)).Inc()
+	return sharedResp, err
 }
 
 func (d *Dispatcher) DispatchExpand(ctx context.Context, req *v1.DispatchExpandRequest) (*v1.DispatchExpandResponse, error) {
--- a/internal/services/shared/errors.go
+++ b/internal/services/shared/errors.go
@@ -71,7 +71,7 @@ func (err ErrSchemaWriteDataValidation) GRPCStatus() *status.Status {
 
 // MaxDepthExceededError is an error returned when the maximum depth for dispatching has been exceeded.
 type MaxDepthExceededError struct {
-	error
+	*spiceerrors.ErrorWithAdditionalDetails
 
 	// AllowedMaximumDepth is the configured allowed maximum depth.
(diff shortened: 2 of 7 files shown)


internal/dispatch/singleflight/singleflight.go
```diff
@@ -82,21 +80,16 @@ func (d *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckReq
...
-	singleFlightCount.WithLabelValues("DispatchCheck", strconv.FormatBool(isShared)).Inc()
-	if err != nil {
-		return &v1.DispatchCheckResponse{Metadata: &v1.ResponseMeta{DispatchCount: 1}}, err
+	if sharedResp == nil {
+		sharedResp = &v1.DispatchCheckResponse{Metadata: &v1.ResponseMeta{DispatchCount: 1}}
 	}
 
-	span := trace.SpanFromContext(ctx)
-	singleflighted := isShared && !primary
-	span.SetAttributes(attribute.Bool("singleflight", singleflighted))
```
vroldanbet, line 98: why was this removed?

I understand that technically all requests are single-flighted, but that's not useful information. What we want to know with this is surfacing in the traces when certain traces are _simply not visible_ because they were singleflighted. This means these dispatches don't even appear, making the trace difficult to understand. This sets an attribute that at least lets the user know the reason spans are missing is because it was single-flighted.

internal/services/shared/errors.go
```diff
@@ -116,13 +116,26 @@ func AsValidationError(err error) *ErrSchemaWriteDataValidation {
...
+
 func RewriteErrorWithoutConfig(ctx context.Context, err error) error {
-	return RewriteError(ctx, err, nil)
+	return rewriteError(ctx, err, nil)
 }
 
 func RewriteError(ctx context.Context, err error, config *ConfigForErrors) error {
+	rerr := rewriteError(ctx, err, config)
+	if config != nil && config.DebugTrace != nil {
+		spiceerrors.WithAdditionalDetails(rerr, DebugTraceErrorDetailsKey, config.DebugTrace.String())
```
vroldanbet, line 133: does `DebugTrace.String()` do the prototext conversion?

pkg/spiceerrors/common.go
```diff
@@ -51,3 +54,35 @@ func AsErrorWithSource(err error) (*ErrorWithSource, bool) {
...
+}
+
+func (err *ErrorWithAdditionalDetails) WithAdditionalDetails(key string, value string) {
+	if err.AdditionalDetails == nil {
+		err.AdditionalDetails = make(map[string]string)
+	}
+	err.AdditionalDetails[key] = value
+}
+
+func (err *ErrorWithAdditionalDetails) Details(details map[string]string) map[string]string {
```
vroldanbet, line 82: I don't think the name of the function clarifies what it does. Maybe `Merge`?

vroldanbet (changes requested):
span information was removed, otherwise looks good to me

internal/dispatch/singleflight/singleflight.go
josephschorr: Turns out it was causing a race condition that the race detector only started to see once I added a test that exercised the code path. I realized we'd likely need a lock to fix it and, IMO, its not worth the overhead simply for a logging property

internal/services/shared/errors.go
josephschorr: Yes and this is tested

pkg/spiceerrors/common.go
josephschorr: Details is what it is called on the underlying gRPC error. I could make it `MergeDetails` or `AddToDetails`?

internal/dispatch/singleflight/singleflight.go
vroldanbet: why would it cause a race? unless the same span is being modified concurrently?

pkg/spiceerrors/common.go
vroldanbet: either works for me

internal/dispatch/singleflight/singleflight.go
josephschorr: It was the setting and accessing of the boolean under the handler

pkg/spiceerrors/common.go
josephschorr: Updated

vroldanbet (approved):

PR #1935, opened 2024-06-11
spanner: use stale reads for current_timestamp for optimized revision

ecordell:
The default is to use a strong read which does some coordination with the leader. For optimized revision computation, it should be fine to use the timestamp returned by the local spanner node and skip the extra coordination.

Testing with a single-region spanner saw these queries drop an order of magnitude.

It is probably also fine to use a stale read for the HeadRevision call, but holding off for further testing in case it could potentially cause a new enemy problem.

To test test (since testing this behavior requires a real spanner), I built a modified image with the following now function:
```go
func (sd *spannerDatastore) now(ctx context.Context) (time.Time, error) {
	go func() {
		start := time.Now()
		var timestamp time.Time
		defer func() {
			log.Ctx(ctx).Warn().Dur("duration", time.Now().Sub(start)).Time("now", timestamp).Str("type", "stale").Msg("now")
		}()
		if err := sd.client.Single().WithTimestampBound(spanner.ExactStaleness(sd.config.followerReadDelay)).Query(ctx, nowStmt).Do(func(r *spanner.Row) error {
			return r.Columns(&timestamp)
		}); err != nil {
			log.Ctx(ctx).Err(err).Msg("failed to stale read current timestamp")
		}
	}()
	start := time.Now()
	var timestamp time.Time
	defer func() {
		log.Ctx(ctx).Warn().Dur("duration", time.Now().Sub(start)).Time("now", timestamp).Str("type", "strong").Msg("now")
	}()
	if err := sd.client.Single().Query(ctx, nowStmt).Do(func(r *spanner.Row) error {
		return r.Columns(&timestamp)
	}); err != nil {
		return time.Time{}, err
	}

	return timestamp, nil
}
```
To compare the performance of stale vs. strong reads here.

Some logs from the test:
```json
{"level":"warn","requestID":"cpk60tjceiqd0bvvpmug","duration":5.713904,"now":"2024-06-11T14:34:30Z","type":"stale","time":"2024-06-11T14:34:30Z","message":"now"}
{"level":"warn","requestID":"cpk60tjceiqd0bvvpmug","duration":38.547463,"now":"2024-06-11T14:34:30Z","type":"strong","time":"2024-06-11T14:34:30Z","message":"now"}
```


ecordell:

ecordell:
recheck

internal/datastore/spanner/spanner.go
```diff
@@ -175,7 +175,11 @@ func NewSpannerDatastore(ctx context.Context, database string, opts ...Option) (
 		cachedEstimatedBytesPerRelationshipLock: sync.RWMutex{},
 		tableSizesStatsTable:                    tableSizesStatsTable,
 	}
-	ds.RemoteClockRevisions.SetNowFunc(ds.headRevisionInternal)
+	// Optimized revision and revision checking use a stale read for the
+	// current timestamp.
+	// Still investigating whether a stale read can be used for the
```
josephschorr, line 180: put a TODO here

PR #1937, opened 2024-06-13
Initial support for Intersection arrow

### LINKED ISSUE #597: Intersection Arrow Proposal
## Background
SpiceDB’s schema contains the arrow operation (known as `TupleToUserset` in Zanzibar), which performs a “walk” of relationships found on a relation (forming the `tupleset`), and for each object found, walks to the referenced relation or permission of a given name on the object (via the `computed_userset`). 

If **any** results are found by this walk, then the overall arrow operation resolves to `HAS_PERMISSION` for a Check:

```
definition user {}

definition folder {
  relation viewer: user
}

definition document {
  relation parent: folder
  permission view = parent->viewer
}
```

### Example Resolution Steps

1. Start a `parent`, and find all subjects for the current document object
2. For each folder found, walk to the `viewer` relation and perform a check for the user on that relation, with the **folder** as the new object
3. If **any** of the folders have the user as a viewer, then the user is considered to be able to `view` the document

## Problem

There are scenarios in which a permission might be defined not based on whether the user has permission on **any** of the parent object(s)'s relations, but rather **all** of the parent objects's relations.

As a concrete example, imagine a schema with permission roles, each of which is important when taken together:

```
definition user {}

definition role {
  relation viewer: user
}

definition document {
  relation requirement: role
  permission view = requirement->viewer // Works with just one requirement
}
```

If the writer of the schema wishes to define `view` such that users can only view a document if they are a `viewer` on **all** roles found in `requirement`, they currently cannot do so easily.

## Proposal

Introduce a new operator into the SpiceDB schema known as the “intersection arrow” or “and arrow”.

The arrow would act the same as the existing arrow operator, but for every subject found, would require that **all** checks succeed for the operator to return `HAS_PERMISSION`

Example:

```
definition user {}

definition role {
  relation viewer: user
}

definition document {
  relation requirement: role
  permission view = requirement-&->viewer // Needs all requirements
}
```

In the above example, if there are multiple roles within `requirement` on the document, then the user would be required to have `viewer` on **all found roles** for them to have `view` permission

### Changes 

- [x] Add either a new operation or a new field onto `TupleToUserset` in the core namespace API protos
    1. Likely a new field indicating the operation to be performed for each computed userset value, and a default to “union”
    2. If a new flag: add switches, on the flag, everywhere a TTU is currently used and ensure that a panic is raised in the default branch
- [x] Add the new syntax to the lexer
- [x] Add the new syntax to the parser, with a new node in the syntax tree
- [x] Have the schemadsl compiler and formatter support the new new node type
- [x] Have Check, Expand and Lookup be adapted to support the new construct/new flag on TTUs
    1. Lookup will need to consider it an intersection and use the slow path for now
- [ ] Update the playground to support the new syntax
- [x] Write many consistency tests
- [ ] Update schema documentation with the new operator and examples
- [ ] Write a blog post, with good examples

josephschorr:
Adds support for intersection arrow and new syntax for existing arrows.

Existing arrow, two syntaxes:

```
permission view = folder->view
permission view = folder.any(view)
```

New syntax for intersection arrow:

```
permission view = folder.all(view)
```

The `folder.all(view)` syntax will ensure that a subject is found for `view` in *all* folders for the parent resource

Fixes #597


internal/dispatch/graph/lookupresources_test.go
```diff
@@ -544,6 +544,39 @@ func TestLookupResourcesOverSchemaWithCursors(t *testing.T) {
...
+				tuple.MustParse("document:doc1#parent@folder:folder1-3"),
+				tuple.MustParse("folder:folder1-1#viewer@user:tom"),
+				tuple.MustParse("folder:folder1-2#viewer@user:tom"),
+				tuple.MustParse("folder:folder1-3#viewer@user:tom"),
+
+				tuple.MustParse("document:doc2#parent@folder:folder2-1"),
+				tuple.MustParse("document:doc2#parent@folder:folder2-2"),
+				tuple.MustParse("document:doc2#parent@folder:folder2-3"),
+				tuple.MustParse("folder:folder2-1#viewer@user:tom"),
+				tuple.MustParse("folder:folder2-2#viewer@user:tom"),
```
ecordell, line 574: add a `0` case: a document that is in no folders.

internal/dispatch/graph/lookupsubjects_test.go
```diff
@@ -707,6 +708,215 @@ func TestCaveatedLookupSubjects(t *testing.T) {
...
+					CaveatExpression: caveatAnd(
+						caveatAnd(
+							caveatAndCtx("caveat1", map[string]interface{}{"someparam1": 42}),
+							caveatAndCtx("caveat2", map[string]interface{}{"someparam2": 43}),
+						),
+						caveatAndCtx("anothercaveat", map[string]interface{}{"anotherparam": 43}),
+					),
+				},
+			},
+		},
```
ecordell, line 919: I don't see any test cases for (not just for LR, the other methods too):

- an `all` over a recursive relationship
- an `all` that is part of an expression

internal/graph/check.go
```diff
@@ -483,11 +483,23 @@ func (cc *ConcurrentChecker) runSetOperation(ctx context.Context, crc currentReq
...
+	case *core.SetOperation_Child_FunctionedTupleToUserset:
+		switch child.FunctionedTupleToUserset.Function {
+		case core.FunctionedTupleToUserset_FUNCTION_ANY:
+			return checkTupleToUserset(ctx, cc, crc, child.FunctionedTupleToUserset)
+
+		case core.FunctionedTupleToUserset_FUNCTION_ALL:
+			return checkIntersectionTupleToUserset(ctx, cc, crc, child.FunctionedTupleToUserset)
+
+		default:
+			return checkResultError(spiceerrors.MustBugf("unknown functioned tuple to userset function `%s`", child.FunctionedTupleToUserset.Function), emptyMetadata)
```
ecordell, line 496: nit/suggestion:
```suggestion
			return checkResultError(spiceerrors.MustBugf("unknown tuple to userset function `%s`", child.FunctionedTupleToUserset.Function), emptyMetadata)
```

internal/graph/expand.go
```diff
@@ -253,13 +263,21 @@ func (ce *ConcurrentExpander) expandComputedUserset(ctx context.Context, req Val
 	})
 }
 
-func (ce *ConcurrentExpander) expandTupleToUserset(_ context.Context, req ValidatedExpandRequest, ttu *core.TupleToUserset) ReduceableExpandFunc {
+type expandFunc func(ctx context.Context, start *core.ObjectAndRelation, requests []ReduceableExpandFunc) ExpandResult
+
+func expandTupleToUserset[T relation](
```
ecordell, line 268: I don't see any tests for expand?

internal/graph/lookupsubjects.go
```diff
@@ -188,16 +189,142 @@ func (cl *ConcurrentLookupSubjects) lookupViaComputed(
...
+					AtRevision:     parentRequest.Revision.String(),
+					DepthRemaining: parentRequest.Metadata.DepthRemaining - 1,
+				},
+			}, stream)
+		})
+	}
+	it.Close()
+
+	// Wait for all dispatched operations to complete.
+	if err := g.Wait(); err != nil {
```
ecordell, line 269: You shouldn't need to wait here - since this is an `AND` operation, you can just add in the sets as you get them (it's associative and commutative). That should keep memory lower too.

internal/namespace/canonicalization_test.go
```diff
@@ -375,6 +375,39 @@ func TestCanonicalization(t *testing.T) {
...
+					ns.TupleToUserset("owner", "something"),
+				)),
+				ns.MustRelation("difftuple", ns.Union(
+					ns.TupleToUserset("viewer", "something"),
+				)),
+				ns.MustRelation("third", ns.Union(
+					ns.MustFunctionedTupleToUserset("owner", "any", "something"),
+				)),
+				ns.MustRelation("thirdwithall", ns.Union(
+					ns.MustFunctionedTupleToUserset("owner", "all", "something"),
```
ecordell, line 397: Could you add some cases in here with expressions containing the fn arrows to make sure they canonicalize the same?

i.e. `owner.all(something) + foo` == `foo + owner.all(something)`

pkg/typesystem/reachabilitygraphbuilder.go
```diff
@@ -98,67 +98,103 @@ func computeRewriteOpReachability(ctx context.Context, children []*core.SetOpera
...
-						return err
-					}
-				}
+			switch child.FunctionedTupleToUserset.Function {
+			case core.FunctionedTupleToUserset_FUNCTION_ANY:
+				// Nothing to change.
+
+			case core.FunctionedTupleToUserset_FUNCTION_ALL:
+				// Mark as a conditional result.
+				operationResultState = core.ReachabilityEntrypoint_REACHABLE_CONDITIONAL_RESULT
```
ecordell, line 116: Why is this conditional? I would think it would be marked the same as a regular arrow.

internal/graph/check.go
```diff
@@ -575,16 +587,194 @@ func removeIndexFromSlice[T any](s []T, index int) []T {
...
+			}
+		},
+		cc.concurrencyLimit,
+	)
+	if err != nil {
+		return checkResultError(err, emptyMetadata)
+	}
+
+	// Create a membership set per-subject-type, representing the membership for each of the dispatched subjects.
+	resultsByDispatchedSubject := map[string]*MembershipSet{}
```
ecordell, line 688: I made a similar comment on one of the other graph methods, but you don't have to collect up all of the results before you combine them; you can combine as you get responses back.

ecordell (changes requested):
Overall this looks really good!

I'm mostly concerned about test coverage for more complex schemas, though I think there's some low-hanging performance fruit as well.

internal/dispatch/graph/lookupresources_test.go
josephschorr: Added

internal/dispatch/graph/lookupsubjects_test.go
josephschorr: Added

internal/graph/expand.go
josephschorr: Added

internal/namespace/canonicalization_test.go
josephschorr: Added

josephschorr:
Updated

ecordell (approved):

PR #1945, opened 2024-06-17
Fix/bulk loader nullstring

### LINKED ISSUE #1938: PostgreSQL Bulkload writes caveat names as null instead of empty string

linux, macos, windows, others

amd64, arm64, others

v.1.33.0

### Steps to Reproduce

1. Create Backup with CLI Tool (`zed backup create backup.bin`)
2. Restore Backup with CLI Tool (`zed backup restore backup.bin`)

But this will be the same over the normal bulkimport endpoint aswell.

### Expected Result

Writen Relationtuples in the database should be the same as in the source.

### Actual Result

The bulk imported tuples without a caveat have the `caveatname` as `null` value in the database. But this should be an empty string. This is problematic when we try to update an existing relation with a caveat later on because the sql check (`caveat_name <> 'some_caveat_name'`) fails for a null value.

It looks like the BulkLoad function uses a sql.NullString instead of a simple string, this causes the problem.

heissa83:
Fixes #1938 and also fixes problems like it in the future by using `IS DISTINCT FROM`instead of `<>` to compare a potential null value field. (This will make it so that already bulk loaded relations will work correctly).

### DIFF
--- a/internal/datastore/postgres/readwrite.go
+++ b/internal/datastore/postgres/readwrite.go
@@ -725,9 +725,7 @@ func exactRelationshipDifferentCaveatClause(r *core.RelationTuple) sq.And {
 			colUsersetRelation:  r.Subject.Relation,
 		},
 		sq.Or{
-			sq.NotEq{
-				colCaveatContextName: caveatName,
-			},
+			sq.Expr(fmt.Sprintf(`%s IS DISTINCT FROM ?`, colCaveatContextName), caveatName),
 			sq.NotEq{
 				colCaveatContext: caveatContext,
 			},
--- a/pkg/datastore/test/bulk.go
+++ b/pkg/datastore/test/bulk.go
@@ -119,6 +119,79 @@ func BulkUploadAlreadyExistsSameCallErrorTest(t *testing.T, tester DatastoreTest
 	grpcutil.RequireStatus(t, codes.AlreadyExists, err)
 }
 
+func BulkUploadEditCaveat(t *testing.T, tester DatastoreTester) {
+	tc := 10
+	require := require.New(t)
+	ctx := context.Background()
+
+	rawDS, err := tester.New(0, veryLargeGCInterval, veryLargeGCWindow, 1)
+	require.NoError(err)
+
+	ds, _ := testfixtures.StandardDatastoreWithSchema(rawDS, require)
+	bulkSource := testfixtures.NewBulkTupleGenerator(
+		testfixtures.DocumentNS.Name,
+		"caveated_viewer",
+		testfixtures.UserNS.Name,
+		tc,
+		t,
+	)
+
+	lastRevision, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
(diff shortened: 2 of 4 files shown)


heissa83:
recheck

heissa83:
I added a new test that checks that caveated relations can be updated after they were added by a BulkLoad. I also did the negative test locally and both changes i made fix the problem. (The test also showed that i was using the wrong column name in the first place, i also fixed that)

vroldanbet (changes requested):
one of the tests is failing

pkg/datastore/test/bulk.go
```diff
@@ -119,6 +119,85 @@ func BulkUploadAlreadyExistsSameCallErrorTest(t *testing.T, tester DatastoreTest
...
+	_, err = ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+		loaded, err := rwt.BulkLoad(ctx, bulkSource)
+		require.NoError(err)
+		require.Equal(uint64(tc), loaded)
+		return err
+	})
+	require.NoError(err)
+
+	head, err := ds.HeadRevision(ctx)
+	require.NoError(err)
```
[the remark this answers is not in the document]
heissa83: Changed it to use the revision from the last write beforehand.

heissa83:
Fixed the tests.
Sorry for some reason my docker daemon keeps crashing when i run the complete test suite locally.

internal/datastore/postgres/readwrite.go
```diff
@@ -725,7 +725,7 @@ func exactRelationshipDifferentCaveatClause(r *core.RelationTuple) sq.And {
 			colUsersetRelation:  r.Subject.Relation,
 		},
 		sq.Or{
-			sq.Expr(`? IS DISTINCT FROM ?`, colCaveatContextName, caveatName),
```
vroldanbet: Can you describe why this fixes the test issue? Both look equivalent to me

heissa83: > Can you describe why this fixes the test issue? Both look equivalent to me


In the first one (`? IS DISTINCT FROM ?`) the `caveat_name` parameter is interpreted as a string value, and so all the caveat names get checked against that, and they are always distinct from that (if you dont try to use a caveat named `caveat_name`). If it is directly written into the Expression (`caveat_name IS DISTINCT FROM ?`) it gets interpreted as the column name and the value of that is used to make the comparison.

So the difference is one will look like this in sql `"caveat_name" IS DISTINCT FROM "some_caveat_name"` and the other looks like `caveat_name IS DISTINCT FROM "some_caveat_name"` in the end.

josephschorr (dismissed):
LGTM

internal/datastore/postgres/readwrite.go
vroldanbet: makes sense, thanks for clarifying!

heissa83:
@vroldanbet do you still have a change request? If not can this PR be merged as it is right now?

vroldanbet (approved):
LGTM!

PR #1942, opened 2024-06-15
Support credential JSON for Spanner

lexcao:
Closes #1941

PR #1948, opened 2024-06-18
Fix empty value on optional credentialsJSON for Spanner

### LINKED ISSUE #1941: Allow credential JSON for Spanner options
### Problem Statement

I am using SpiceDB library mode with Spanner for testing.
And I want to pass `CredentialsJSON` for Spanner.

### Solution Brainstorm

Add `credentialsJSON` for `spannerOptions`

I will send a PR.

lexcao:
Closes #1941
This is the follow-up PR for #1942, which fix the optional value issue.

### DIFF
--- a/internal/datastore/spanner/spanner.go
+++ b/internal/datastore/spanner/spanner.go
@@ -141,14 +141,25 @@ func NewSpannerDatastore(ctx context.Context, database string, opts ...Option) (
 	cfg := spanner.DefaultSessionPoolConfig
 	cfg.MinOpened = config.minSessions
 	cfg.MaxOpened = config.maxSessions
-	client, err := spanner.NewClientWithConfig(context.Background(), database,
-		spanner.ClientConfig{SessionPoolConfig: cfg},
+
+	var spannerOpts []option.ClientOption
+	if config.credentialsJSON != nil {
+		spannerOpts = append(spannerOpts, option.WithCredentialsJSON(config.credentialsJSON))
+	}
+
+	spannerOpts = append(spannerOpts,
 		option.WithCredentialsFile(config.credentialsFilePath),
-		option.WithCredentialsJSON(config.credentialsJSON),
 		option.WithGRPCConnectionPool(max(config.readMaxOpen, config.writeMaxOpen)),
 		option.WithGRPCDialOption(
 			grpc.WithStatsHandler(otelgrpc.NewClientHandler()),
 		),
+	)
+
+	client, err := spanner.NewClientWithConfig(
+		context.Background(),
+		database,
+		spanner.ClientConfig{SessionPoolConfig: cfg},
+		spannerOpts...,
 	)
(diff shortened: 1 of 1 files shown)


vroldanbet (commented):
The change makes sense to me, but concerned it has taken 3 PRs to get this right. Can we add a test?

internal/datastore/spanner/spanner.go
```diff
@@ -141,14 +141,21 @@ func NewSpannerDatastore(ctx context.Context, database string, opts ...Option) (
...
+
 	client, err := spanner.NewClientWithConfig(context.Background(), database,
 		spanner.ClientConfig{SessionPoolConfig: cfg},
-		option.WithCredentialsFile(config.credentialsFilePath),
-		option.WithCredentialsJSON(config.credentialsJSON),
-		option.WithGRPCConnectionPool(max(config.readMaxOpen, config.writeMaxOpen)),
-		option.WithGRPCDialOption(
-			grpc.WithStatsHandler(otelgrpc.NewClientHandler()),
-		),
+		append(optionalOpts,
```
josephschorr, line 152: append this above the NewClient block

lexcao: Done, thanks.

lexcao:
Hi @vroldanbet 
I just thought it is options forwarding thing but I am not familiar with the code base at the beginning.
Is there any test case I can add?

vroldanbet:
@lexcao I've made client initialization simpler by making the option handle empty credentials JSON, which Spanner client does not handle well. I also added a test. Please have a look and make sure this works for you.

lexcao:
Hi @vroldanbet 
Thanks for your suggestions, but it would not resolve the issue.
The key point is `option.WithCredentialsJSON(config.credentialsJSON)` from Spanner, 
which should skip adding that option if the credentialsJSON is nil.

The internal operation in the `WithCredentialsJSON` method is making nil to `empty slice of bytes`, 
which would cause initialization error for Spanner with `empty slice`

https://github.com/googleapis/google-api-go-client/blob/e732ee3983555c5b0ca708033d9b8ec675f99019/option/option.go#L61-L70

vroldanbet:
@lexcao, you are right. I'd suggest removing my commit, and if anything adding a new commit only with the test. I'd also suggest opening a PR upstream spanner SDK to fix this annoying behaviour

lexcao:
Thanks, let me try to raise a PR for Spanner.

lexcao:
Hi @vroldanbet 
I just raised a PR in the upstream. 
https://github.com/googleapis/google-api-go-client/pull/2643
Waiting for review.

For now, I think we should add the fix for this, since it could break the use of `CredentialsFile` for Spanner.
I have updated my code, you can review now.

I will make it more ergonomic after the upstream accepts my PR then.

vroldanbet (approved):

PR #1949, opened 2024-06-18
Start on a steel thread test framework

josephschorr:
Will be used to test the new LS implementation and the new LR2 implementation


internal/services/steelthreadtesting/steelthread_test.go
```diff
@@ -0,0 +1,127 @@
...
+			ctx = datastoremw.ContextWithDatastore(ctx, ds)
+
+			// Load in the data.
+			_, _, err = validationfile.PopulateFromFiles(ctx, ds, []string{"testdata/" + tc.datafile})
+			require.NoError(t, err)
+
+			// Create a new server.
+			listener := bufconn.Listen(defaultConnBufferSize)
+
+			s := grpc.NewServer(
```
ecordell, line 53: there's already a helper for spinning up servers like this in `internal/testserver`

internal/services/steelthreadtesting/steelthread_test.go
```diff
@@ -0,0 +1,127 @@
...
+const defaultConnBufferSize = humanize.MiByte
+
+func TestSteelThreads(t *testing.T) {
+	for _, tc := range steelThreadTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancel()
+
+			// Create a new datastore.
+			ds, err := memdb.NewMemdbDatastore(0, 0*time.Second, memdb.DisableGC)
```
ecordell, line 41: I'm wondering about the choice of memdb here. We could definitely have bugs that would be datastore-specific and affect a steelthread test, right? Should we run against all backends?

magefiles/test.go
```diff
@@ -48,6 +48,20 @@ func (Test) Integration() error {
 	return goTest("./internal/services/integrationtesting/...", "-tags", "ci,docker", "-timeout", "15m")
 }
 
+// Steelthread Run steelthread tests
+func (Test) Steelthread() error {
```
ecordell, line 52: Add to `Test.All`?

ecordell: Related: I don't see any CI config for this. Where and how often will we run these?

internal/services/steelthreadtesting/steelresults/basic-lookup-subjects-uncursored-lookup-subjects-for-somedoc-results.yaml
```diff
@@ -0,0 +1,1601 @@
...
+- 'user-81'
+- 'user-82'
+- 'user-83'
+- 'user-84'
+- 'user-85'
+- 'user-86'
+- 'user-87'
+- 'user-88'
+- 'user-89'
+- 'user-9'
```
ecordell, line 1601: What if actually outputting the contents were a flag? 

Under typical circumstances, the stored results could just be like a chained digest of results or something along those lines that we can update as we go (would keep memory overhead low when running a really large steelthhread tests). Then you just need to store/compare the final hash.

If they're different of course you'd want to see the content differences (hence a flag), but we should be able to reconrstruct previous steelthread results just by running the suite at an earlier commit to compare.

internal/services/steelthreadtesting/steelthread_test.go
josephschorr: Yes, that was my plan as a followup

magefiles/test.go
josephschorr: The idea would be to run them manually, either via mage here or a manual action in CI

internal/services/steelthreadtesting/steelresults/basic-lookup-subjects-uncursored-lookup-subjects-for-somedoc-results.yaml
josephschorr: The idea was to allow those writing the tests to eyeball and validate the results manually; having a hash would make that harder

josephschorr: This is partially why I did nicely formatted YAML vs JSON

internal/services/steelthreadtesting/steelthread_test.go
ecordell: The testserver takes datastore init as an argument, we use it for datastore tests as well

internal/services/steelthreadtesting/steelresults/basic-lookup-subjects-uncursored-lookup-subjects-for-somedoc-results.yaml
ecordell: A flag allows eyeballing while also keeping the git storage and memory low for the common case, right?

This test case is quite small compared to what I understood you wanted to do with this suite. Maybe git LFS is another option?

internal/services/steelthreadtesting/steelthread_test.go
josephschorr: Okay, I'll switch to that

internal/services/steelthreadtesting/steelresults/basic-lookup-subjects-uncursored-lookup-subjects-for-somedoc-results.yaml
josephschorr: This case is small, the other one is less so, but yes, long term we'd likely want to store this in LFS

josephschorr: This also gets a lot smaller if we could use iterators in-code

ecordell (approved):

internal/services/steelthreadtesting/steelthread_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/gosimple/slug"
+	"github.com/stretchr/testify/require"
+)
+
+const defaultConnBufferSize = humanize.MiByte
+
+func TestSteelThreads(t *testing.T) {
+	for _, tc := range steelThreadTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			req := require.New(t)
```
vroldanbet, line 29: @josephschorr any reason steel threads are not running in parallel?

josephschorr: No, I just didn't do so in the initial PRs

PR #1951, opened 2024-06-20
enriches postgres revisions with txID and timestamp

vroldanbet:
consumers of the Datastore Watch API may want to access the postgres transaction of a revision, and
the recorded timestamp.

Since we don't want to load it in every call sites for performance reasons, a revision may be in a partial state, thus it should be retrieved via the accessor methods.


internal/datastore/postgres/revisions.go
```diff
@@ -284,6 +288,22 @@ func (pr postgresRevision) mustMarshalBinary() []byte {
 	return serialized
 }
 
+func (pr postgresRevision) OptionalTransactionID() (xid8, bool) {
```
josephschorr, line 291: Doc comment

internal/datastore/postgres/revisions.go
```diff
@@ -284,6 +288,22 @@ func (pr postgresRevision) mustMarshalBinary() []byte {
...
+	if !pr.optionalTxID.Valid {
+		return xid8{}, false
+	}
+
+	return pr.optionalTxID, true
+}
+
+func (pr postgresRevision) OptionalTimestamp() (uint64, bool) {
```
josephschorr, line 299: Doc comment

internal/datastore/postgres/revisions_test.go
```diff
@@ -83,6 +84,23 @@ func TestRevisionSerDe(t *testing.T) {
...
+	require.True(t, ok)
+	require.Equal(t, receivedTimestamp, testTimestamp)
+	txid, ok := pgr.OptionalTransactionID()
+	require.True(t, ok)
+	require.Equal(t, newXid8(1), txid)
+
+	pgr = postgresRevision{snapshot: pgSnapshot{}}
+	_, ok = pgr.OptionalTimestamp()
+	require.False(t, ok)
+	_, ok = pgr.OptionalTransactionID()
```
josephschorr, line 100: add tests for when it is available

pkg/datastore/test/counters.go
```diff
@@ -174,7 +174,7 @@ func UpdateRelationshipCounterTest(t *testing.T, tester DatastoreTester) {
 	require.Len(t, filters, 1)
 	require.Equal(t, "somedocfilter", filters[0].Name)
 	require.Equal(t, 1234, filters[0].Count)
-	require.Equal(t, updatedRev, filters[0].ComputedAtRevision)
+	require.True(t, updatedRev.Equal(filters[0].ComputedAtRevision))
```
josephschorr, line 177: Add a comment here as to why Equal is not used

vroldanbet: https://github.com/authzed/spicedb/pull/1951/commits/dc450332fa3d03ed97d176a50d1e33686e9fdf13

internal/datastore/postgres/revisions_test.go
```diff
@@ -83,6 +84,23 @@ func TestRevisionSerDe(t *testing.T) {
...
+	require.False(t, ok)
```
josephschorr, line 101: Add an assertion that they are still equal here, so long as the snapshot is the same

vroldanbet: done

PR #1968, opened 2024-07-03
Move to go 1.22.5 for a reporting go lang vuln

josephschorr:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/authzed/spicedb
 
-go 1.22.4
```
vroldanbet, line 3: I don't understand why go mod keeps updating those. And why even the minor version is relevant here. This used to mean "the go version you need to compile this project". Has anything changed in the go toolchain? It's annoying that this keeps bumping

josephschorr: Yes, it now means the go version to use and `toolchain` is the toolchain to use

josephschorr: Changed back since it broke WASM

vroldanbet (approved):

PR #1972, opened 2024-07-04
adjust pg revision timestamps

vroldanbet:
two follow up changes to https://github.com/authzed/spicedb/pull/1951
- the unit is nanos, to align it with the resolution of the inexact float64 provided by the CRDB HCL revision
- also provide timestamp and transaction ID values with checkpoint revisions, which were not being delivered via the Datastore Watch API

### DIFF
--- a/internal/datastore/postgres/postgres_shared_test.go
+++ b/internal/datastore/postgres/postgres_shared_test.go
@@ -5,6 +5,7 @@ package postgres
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"strings"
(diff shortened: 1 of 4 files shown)


internal/datastore/postgres/postgres_shared_test.go
```diff
@@ -1536,4 +1546,72 @@ func NullCaveatWatchTest(t *testing.T, ds datastore.Datastore) {
...
+	lowestRevision, err := ds.HeadRevision(ctx)
+	require.NoError(err)
+
+	// Run the watch API.
+	changes, errchan := ds.Watch(ctx, lowestRevision, datastore.WatchOptions{
+		Content: datastore.WatchRelationships | datastore.WatchSchema | datastore.WatchCheckpoints,
+	})
+	require.Zero(len(errchan))
+
+	// Manually insert a relationship with a NULL caveat. This is allowed, but can only happen due to
```
josephschorr, line 1564: Copy-paste error?

josephschorr (approved):
LGTM

PR #1973, opened 2024-07-05
Refactor: replace task management slice with buffered channel

lalalalatt:
- Change `tasks` from a slice to a buffered channel for better concurrency handling
- Modify `NewPreloadedTaskRunner` to initialize `tasks` as a buffered channel
- Update `Add` method to send tasks to the channel instead of appending to a slice
- Replace `Start` logic with a goroutine to run tasks from the channel
- Remove `spawnIfAvailable` and `runner` methods
- Add `run` method to handle task execution loop
- Implement `runTask` method to execute tasks and manage semaphore and wait group
- Simplify task cancellation in `emptyForCancel` by draining the task channel

### DIFF
--- a/internal/taskrunner/preloadedtaskrunner.go
+++ b/internal/taskrunner/preloadedtaskrunner.go
@@ -21,9 +21,10 @@ type PreloadedTaskRunner struct {
 	wg    sync.WaitGroup
 	err   error
 	lock  sync.Mutex
-	tasks []TaskFunc
+	tasks chan TaskFunc
 }
 
+// NewPreloadedTaskRunner creates a new PreloadedTaskRunner with the specified concurrency limit and initial capacity.
 func NewPreloadedTaskRunner(ctx context.Context, concurrencyLimit uint16, initialCapacity int) *PreloadedTaskRunner {
 	// Ensure a concurrency level of at least 1.
 	if concurrencyLimit <= 0 {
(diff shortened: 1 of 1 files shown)


vroldanbet (commented):
Hi,

thanks for your contribution, what specific problem are you trying to solve?

lalalalatt:
I am new to this project and I was reading the source code then found that the task queue in the PreloadTaskRunner is implemented with slice

I think it would be better to use a thread safe queue or channel to implement the task queue.

lalalalatt:
> Hi,
> 
> 
> 
> thanks for your contribution, what specific problem are you trying to solve?

I am new to this project and I was reading the source code then found that the task queue in the PreloadTaskRunner is implemented with slice

I think it would be better to use a thread safe queue or channel to implement the task queue.

lalalalatt:

vroldanbet:
@lalalalatt we appreciate the contribution but this is a complex and sensitive part of the system, and changes can lead to unexpected effects, this implementation has already been battle-tested in production, so I don't think we should change things for the sake of changing them.

Channels are a great tool for synchronization but they are a source of complex bugs. SpiceDB is an authorization service and therefore we strive to reduce risks as much as possible.

For reference, [this paper](https://songlh.github.io/paper/go-study.pdf) reveals that the usage of channels typically led to more bugs than shared memory:

>Shared memory vs. message passing. Our study found
that message passing does not necessarily make multithreaded programs less error-prone than shared memory.
In fact, message passing is the main cause of blocking bugs.
To make it worse, when combined with traditional synchronization primitives or with other new language features
and libraries, message passing can cause blocking bugs that
are very hard to detect. Message passing causes less nonblocking bugs than shared memory synchronization and surprisingly, was even used to fix bugs that are caused by wrong
shared memory synchronization. We believe that message
passing offers a clean form of inter-thread communication
and can be useful in passing data and signals. But they are
only useful if used correctly, which requires programmers
to not only understand message passing mechanisms well
but also other synchronization mechanisms of Go.

As I noted in discord, if you are looking to contribute, I think a good first step is to look into issues labeled with ["good first issue"](https://github.com/authzed/spicedb/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).

Issue #1953, opened 2024-06-20
Relationship Integrity


jzelinskie:
## Problem Statement

SpiceDB does not support a threat-model where the storage for the datastore is untrusted.
Today it is not possible to determine if relationships came from the SpiceDB API or were modified by a human with Datastore access. This is perfectly valid for those that fully trust the vendor providing services powering their SpiceDB Datastore. However, there are users that require additional integrity guarantees: proof that there has not been tampering with any data being consumed by SpiceDB.

For example, a customer support engineer for any one of the possible service providers for SpiceDB datastores could inadvertently modify data in the underlying storage causing changes to access being granted/revoked and downstream consumers of the Watch API from being inconsistency.

### Overview
Relationship Integrity is to be provided by including MAC checking into the SpiceDB datastore itself. When enabled, reading any relationship from the datastore will require checking the validity of a MAC of the relationship. If a relationship fails verification, an error is returned failing the entire request.
### Implementation
SpiceDB supports a pattern for implementing supersets of behavior for Datastore implementations by performing interface assertions. Today this is used to implement extensions to the Datastore interface for datastores that require [running background processes](https://github.com/authzed/spicedb/blob/v1.33.0/pkg/datastore/datastore.go#L516-L524) or [repairing any internal values](https://github.com/authzed/spicedb/blob/v1.33.0/pkg/datastore/datastore.go#L536-L546) after a major event such as failover or migration.

In order to progressively add Relationship Integrity support to the various datastores, it makes sense to introduce this feature as an extension. Datastores can optionally implement the interface to support the behavior; our first choice is CockroachDB as this vendor is not yet as ubiquitously trusted as cloud providers offering services for our other datastore implementations.

Logic for enforcing MAC checks will be implemented as a [Datastore proxy](https://github.com/authzed/spicedb/tree/v1.33.0/internal/datastore/proxy). Today, Datastore proxies are used in order to implement common behavior to all Datastores including collecting metrics, enforcing read-only access, hedging datastore access, and deduplicating datastore access.

MACs are to be stored alongside relationships; for datastores powered by relational databases, this implementation should be an additional column added to table storing relationships. Keys are provided similarly to any other secret in SpiceDB -- via the environment/flag system. AuthZed's managed products will manage keys using cloud providers' KMS services with the option to have this key be owned by customers.

[...]

PR #1980, opened 2024-07-08
Relationship integrity

josephschorr:
Fixes: #1953


internal/datastore/crdb/crdb.go
```diff
@@ -52,19 +52,23 @@ const (
...
-	colCaveatContext           = "caveat_context"
+	colNamespace         = "namespace"
+	colConfig            = "serialized_config"
+	colTimestamp         = "timestamp"
+	colTransactionKey    = "key"
+	colObjectID          = "object_id"
+	colRelation          = "relation"
+	colUsersetNamespace  = "userset_namespace"
+	colUsersetObjectID   = "userset_object_id"
+	colUsersetRelation   = "userset_relation"
```
jzelinskie, line 63: maybe put a new line here to keep all the caveat stuff together

internal/datastore/crdb/migrations/zz_migration.0007_add_integrity_columns.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	addIntegrityColumnsQuery = `
+		ALTER TABLE relation_tuple
+		ADD COLUMN IF NOT EXISTS integrity_hash BYTEA DEFAULT NULL,
+		ADD COLUMN IF NOT EXISTS integrity_key_id VARCHAR(255) DEFAULT NULL;
+	`
+)
+
+func init() {
+	err := CRDBMigrations.Register("add-integrity-columns", "add-relationship-counters-table", addIntegrityColumns, noAtomicMigration)
```
jzelinskie, line 18: is there a must register? i swear it's been written at some point

internal/datastore/crdb/reader.go
```diff
@@ -36,6 +36,9 @@ var (
 		colUsersetRelation,
 		colCaveatContextName,
 		colCaveatContext,
+		colIntegrityKeyID,
+		colIntegrityHash,
+		colTimestamp,
```
jzelinskie, line 41: do we need the timestamp if we have the key id?

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,397 @@
...
+	sb.WriteString(tpl.Subject.ObjectId)
+	sb.WriteString("#")
+	sb.WriteString(tpl.Subject.Relation)
+
+	if tpl.Caveat != nil && tpl.Caveat.CaveatName != "" {
+		sb.WriteString(" with ")
+		sb.WriteString(tpl.Caveat.CaveatName)
+		sb.WriteString(":")
+		sb.WriteString(tpl.Caveat.Context.String())
+	}
```
jzelinskie, line 123: is this something we can put in the tuple package? maybe something similar already exists

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,397 @@
...
+	return r.ds.ReadyState(ctx)
+}
+
+func (r *relationshipIntegrityProxy) RevisionFromString(serialized string) (datastore.Revision, error) {
+	return r.ds.RevisionFromString(serialized)
+}
+
+func (r *relationshipIntegrityProxy) Statistics(ctx context.Context) (datastore.Stats, error) {
+	return r.ds.Statistics(ctx)
+}
```
jzelinskie, line 177: i like to make functions like these 1 single line so they all stack together without newlines between. it makes it more clear that they're just boilerplate

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,397 @@
...
+)
+
+// NewRelationshipIntegrityProxy creates a new datastore proxy that ensures the integrity of
+// relationships by using HMACs to sign the data. The current key is used to sign new data,
+// and the expired keys are used to verify old data, if any.
+func NewRelationshipIntegrityProxy(ds datastore.Datastore, currentKey KeyConfig, expiredKeys []KeyConfig) (datastore.Datastore, error) {
+	currentKeyHMAC := hmacConfig{
+		keyID:     currentKey.ID,
+		expiredAt: currentKey.ExpiredAt,
+		hmacPool: &sync.Pool{
```
ecordell, line 47: I'm sure it depends, but for tiger I found that a buffered channel performed better than sync.Pool (`select` on retrieving from the pool to return a new instance if the pool is empty, similar to sync.Pool)

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,397 @@
...
+
+		if _, ok := keysByID[key.ID]; ok {
+			return nil, spiceerrors.MustBugf("found duplicate key ID: %s", key.ID)
+		}
+
+		keysByID[key.ID] = hmacConfig{
+			keyID:     key.ID,
+			expiredAt: key.ExpiredAt,
+			hmacPool: &sync.Pool{
+				New: func() any { return hmac.New(alg, key.Bytes) },
```
ecordell, line 72: it's probably worth benchmarking sha3 here, since you can use it as a MAC directly instead of through an HMAC

internal/datastore/crdb/reader.go
josephschorr: Yes, to ensure that if a key is expired, we can still accept relationships signed from before the expiration time

internal/datastore/proxy/relationshipintegrity.go
josephschorr: Yes, we could put it into there

internal/datastore/proxy/relationshipintegrity.go
vroldanbet: I also learned recently about https://github.com/google/highwayhash (and its Go native impl, https://github.com/minio/highwayhash), theoretically 1 order of magnitude faster than SHA256 thanks to SIMD, but says:

> HighwayHash is not a general-purpose cryptographic hash function (such as Blake2b, SHA-3 or SHA-2) and should not be used if strong collision resistance is required.

internal/datastore/crdb/migrations/zz_migration.0007_add_integrity_columns.go
josephschorr: Nope, we never added it. We could, but this seems fine

internal/datastore/proxy/relationshipintegrity.go
josephschorr: Tried but go is reformatting it back to this

josephschorr: Likely because the receiver is named `relationshipIntegrityProxy` and thus its making the lines too long. Shall I rename the struct?

internal/datastore/proxy/relationshipintegrity.go
josephschorr: Tried both and SHA3 was slower locally while highway is ever so slightly faster

josephschorr: Did some prelim tests and highwayhash is faster by about ~50% while SHA3 seemed to be slower.

However, I'm concerned about the lack of strong collision resistance. We'll need to investigate the implications of that.

josephschorr: Updated branch with the highwayhash impl for testing

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+func hasherForKey(key []byte) (hash.Hash, error) {
+	return highwayhash.New64(key)
+}
+
+type relationshipIntegrityProxy struct {
+	ds         datastore.Datastore
+	primaryKey hmacConfig
+	keysByID   map[string]hmacConfig
```
ecordell, line 94: it's usually a good idea to encode the hash algorithm/parameters in the key id for later retrieval (password hashing usually does this for example) so that the algorithm / parameters can be rotated alongside the keys themselves.

I.e. it's sha256 now but we decide we need to rotate to sha512 or sha3 etc.

josephschorr: That's why we have the algorithm version byte; we can map from that to the alg used

internal/datastore/proxy/relationshipintegrity.go
vroldanbet: The scenario this attempts to protect from is unauthorized modification of the state of the database. This requires sophisticated attacks and computing running for days to find a collision that could reveal the seed. If someone has access to the database and the machinery to run such an attack, the hash is not going to get in the way honestly, they probably have enough incentives to find other ways. I don't think remote attacks, with all the level of indirections, can give a clear signal to an attacker.

This is a decision the customer could themselves make and we could make it configurable via a flag. If the customer considers only collision-resistant hashes acceptable and are willing to accept the tradeoff in computational cost, then they can choose to do so. Highwayhash is by no means a "weak" hash, but one that makes the attack cost be sufficiently high to be unfeasible to most attackers.

>By contrast, 'strong' hashes such as SipHash or HighwayHash require infeasible attacker effort to find a hash collision (an expected 2^32 guesses of m per the birthday paradox) or recover the seed (2^63 requests). These security claims assume the seed is secret. It is reasonable to suppose s is initially unknown to attackers, e.g. generated on startup or even per-connection. 

Consider the authors describe how complicated an attack is:
>A timing attack by Wool/Bar-Yosef recovers 13-bit seeds by testing all 8K possibilities using millions of requests, which takes several days (even assuming unrealistic 150 us round-trip times). It appears infeasible to recover 64-bit seeds in this way.

ecordell: The context here was that (highwayhash isn't suitable for hmac)[https://github.com/minio/highwayhash/issues/13#issuecomment-528356795] (or at least we don't know that it is)

internal/datastore/proxy/relationshipintegrity.go
ecordell: Missed that, sorry

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,389 @@
...
+
+var (
+	versionByte = byte(0x01)
+	hashLength  = 16
+)
+
+// NewRelationshipIntegrityProxy creates a new datastore proxy that ensures the integrity of
+// relationships by using HMACs to sign the data. The current key is used to sign new data,
+// and the expired keys are used to verify old data, if any.
+func NewRelationshipIntegrityProxy(ds datastore.Datastore, currentKey KeyConfig, expiredKeys []KeyConfig) (datastore.Datastore, error) {
```
ecordell, line 41: I don't see this being used anywhere?

ecordell (commented):
Generally looking good, I left some comments throughout, but my main concerns are:

1. I don't see where the integrity proxy gets configured and installed...?
2. I don't see where/how spicedb loads signing keys. Is a commit missing?
3. I have some concerns about these fields being included in every query and not just when integrity is enabled. Wouldn't it be better to not query for those fields at all? I think this will mess with covering indexes, since I don't think we want them included in indexes.
4.  I think we need to have something in place for key rotation and, maybe more importantly, evolution of the canonical form. Let's say we add an extra field for native TTLs to tuples and we migrate caveat TTLs to a native form - that will require changes in the canonical form and re-signing with the keys.

internal/datastore/crdb/readwrite.go
```diff
@@ -51,7 +51,7 @@ type crdbReadWriteTXN struct {
 
 var (
 	upsertTupleSuffix = fmt.Sprintf(
-		"ON CONFLICT (%s,%s,%s,%s,%s,%s) DO UPDATE SET %s = now(), %s = excluded.%s, %s = excluded.%s WHERE (relation_tuple.%s <> excluded.%s OR relation_tuple.%s <> excluded.%s)",
+		"ON CONFLICT (%s,%s,%s,%s,%s,%s) DO UPDATE SET %s = now(), %s = excluded.%s, %s = excluded.%s, %s = excluded.%s, %s = excluded.%s WHERE (relation_tuple.%s <> excluded.%s OR relation_tuple.%s <> excluded.%s)",
```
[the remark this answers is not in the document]
josephschorr: Yeah, this is a common "issue" with adding these columns, but it does mean having multiple copies, which means more testing

internal/datastore/crdb/watch.go
```diff
@@ -248,6 +253,30 @@ func (cds *crdbDatastore) watch(
 				return
 			}
 
+			var integrity *core.RelationshipIntegrity
+
+			if details.After != nil && details.After.IntegrityKeyID != nil && details.After.IntegrityHashAsHex != nil && details.After.TimestampAsString != nil {
+				hexString := *details.After.IntegrityHashAsHex
+				hashBytes, err := hex.DecodeString(hexString[2:]) // drop the \x
```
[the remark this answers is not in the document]
josephschorr: Not according to testing

internal/datastore/mysql/migrations/zz_migration.0009_add_relationship_integrity_columns.go
```diff
@@ -0,0 +1,20 @@
+package migrations
+
+import "fmt"
+
+func addRelationshipIntegrityColumns(t *tables) string {
+	return fmt.Sprintf(`ALTER TABLE %s
+		ADD COLUMN integrity_key_id VARCHAR(255) DEFAULT NULL,
```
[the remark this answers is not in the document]
josephschorr: No, but do you have a good limit for it?

internal/datastore/mysql/migrations/zz_migration.0009_add_relationship_integrity_columns.go
```diff
@@ -0,0 +1,20 @@
+package migrations
+
+import "fmt"
+
+func addRelationshipIntegrityColumns(t *tables) string {
+	return fmt.Sprintf(`ALTER TABLE %s
+		ADD COLUMN integrity_key_id VARCHAR(255) DEFAULT NULL, 
+		ADD COLUMN integrity_hash TINYBLOB DEFAULT NULL,
```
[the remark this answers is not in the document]
josephschorr: Will switch; wasn't aware of it

internal/datastore/postgres/common/pgx.go
```diff
@@ -64,6 +73,15 @@ func queryTuples(ctx context.Context, sqlStatement string, args []any, span trac
 			if err != nil {
 				return fmt.Errorf(errUnableToQueryTuples, fmt.Errorf("unable to fetch caveat context: %w", err))
 			}
+
+			if integrityKeyID.Valid && integrityTimestamp != nil {
```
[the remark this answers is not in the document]
josephschorr: If we pushed validation into the datastores, we can no longer using a proxy to implement it as a common solution, as it would have no access to the integrity information

josephschorr:
Redesigned and rebased as discussed

internal/datastore/postgres/common/bulk.go
```diff
@@ -44,6 +45,12 @@ func (tg *tupleSourceAdapter) Values() ([]any, error) {
 	tg.valuesBuffer[6] = caveatName
 	tg.valuesBuffer[7] = caveatContext
 
+	if len(tg.colNames) > 8 && tg.current.Integrity != nil {
```
ecordell, line 48: I thought we weren't going to add this for anything except crdb. Am I mistaken?

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,389 @@
...
+
+func (r *relationshipIntegrityTx) BulkLoad(
+	ctx context.Context,
+	iter datastore.BulkWriteRelationshipSource,
+) (uint64, error) {
+	wrapped := &wrappedBulkLoadIterator{iter, r.parent}
+	return r.ReadWriteTransaction.BulkLoad(ctx, wrapped)
+}
+
+type wrappedBulkLoadIterator struct {
```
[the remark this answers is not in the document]
ecordell: ^

internal/datastore/proxy/relationshipintegrity_test.go
```diff
@@ -0,0 +1,419 @@
...
+		return tx.WriteRelationships(context.Background(), []*core.RelationTupleUpdate{
+			tuple.Create(tuple.MustParse("resource:foo#viewer@user:tom")),
+			tuple.Create(tuple.MustParse("resource:foo#viewer@user:fred")),
+			tuple.Touch(tuple.MustParse("resource:bar#viewer@user:sarah")),
+		})
+	})
+	require.NoError(t, err)
+
+	// Insert an invalid integrity hash for one of the relationships to be invalid by bypassing
+	// the
```
ecordell, line 131: bypassing the

internal/datastore/proxy/relationshipintegrity_test.go
```diff
@@ -0,0 +1,419 @@
...
+	_, err = pds.ReadWriteTx(context.Background(), func(ctx context.Context, tx datastore.ReadWriteTransaction) error {
+		return tx.WriteRelationships(context.Background(), []*core.RelationTupleUpdate{
+			tuple.Create(tuple.MustParse("resource:foo#viewer@user:tom")),
+			tuple.Create(tuple.MustParse("resource:foo#viewer@user:fred")),
+			tuple.Touch(tuple.MustParse("resource:bar#viewer@user:sarah")),
+		})
+	})
+	require.NoError(t, err)
+
+	// Insert an invalid integrity hash for one of the relationships to be invalid by bypassing
```
ecordell, line 197: bypassing the

pkg/datastore/test/tuples.go
```diff
@@ -1595,10 +1598,11 @@ func ConcurrentWriteSerializationTest(t *testing.T, tester DatastoreTester) {
 				close(waitToStart)
 			})
 			<-waitToFinish
-
 			return err
 		})
-		require.NoError(err)
+		if err != nil {
+			panic(err)
```
ecordell, line 1604: I know it's a test, but why panic when it's already in an errgroup?

pkg/tuple/tuple.go
```diff
@@ -572,3 +575,95 @@ func WithCaveat(tpl *core.RelationTuple, caveatName string, contexts ...map[stri
...
 }
+
+// CanonicalBytes converts a tuple to a canonical set of bytes. If the tuple is nil or empty, returns nil.
+// Can be used for hashing purposes.
+func CanonicalBytes(tpl *core.RelationTuple) ([]byte, error) {
+	if tpl == nil {
+		return nil, nil
+	}
+
+	var sb strings.Builder
```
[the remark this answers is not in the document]
ecordell: ^

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,482 @@
...
+}
+
+// computeRelationshipHash computes the HMAC hash of a relationship tuple.
+func computeRelationshipHash(tpl *corev1.RelationTuple, key *hmacConfig) ([]byte, error) {
+	bytes, err := tuple.CanonicalBytes(tpl)
+	if err != nil {
+		return nil, err
+	}
+
+	key.lock.Lock()
```
ecordell, line 162: this seems like it would be a bottleneck for large reads. what about a pool of hashers instead? either a `sync.Pool` or even just a `chan hasher`

internal/datastore/proxy/relationshipintegrity.go
```diff
@@ -0,0 +1,482 @@
...
+		keysByID:   keysByID,
+	}, nil
+}
+
+func hasherForKey(key []byte) (hash.Hash, error) {
+	return hmac.New(sha256.New, key), nil
+}
+
+type relationshipIntegrityProxy struct {
+	ds         datastore.Datastore
```
ecordell, line 141: if you just embed  `datastore.Datastore` instead of naming it, you can remove most of the methods in this file since they'll just pass through

internal/datastore/postgres/common/bulk.go
josephschorr: We aren't, but this is shared with CRDB, so I needed to handle it here

internal/datastore/proxy/relationshipintegrity.go
josephschorr: Exactly what I didn't want to be possible - if we add another read method down the road and we just embed, this won't fail to compile. I want it to do so to ensure that we don't accidentally forget to add validation to any new read method added

internal/datastore/crdb/crdb.go
```diff
@@ -447,6 +459,18 @@ func (cds *crdbDatastore) Features(ctx context.Context) (*datastore.Features, er
 	return features, err
 }
 
+func (cds *crdbDatastore) SupportsIntegrity() bool {
```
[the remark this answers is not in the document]
josephschorr: I wanted it to match an interface in the proxy around casting; if you like, I can move it into Features and just have the proxy check that call instead

internal/datastore/crdb/crdb_test.go
```diff
@@ -111,6 +123,69 @@ func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
...
+				GCWindow(gcWindow),
+				RevisionQuantization(revisionQuantization),
+				WatchBufferLength(watchBufferLength),
+				OverlapStrategy(overlapStrategyPrefix),
+				DebugAnalyzeBeforeStatistics(),
+				WithIntegrity(true),
+			)
+			require.NoError(t, err)
+
+			wrapped, err := proxy.NewRelationshipIntegrityProxy(ds, defaultKeyForTesting, nil)
```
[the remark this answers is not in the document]
josephschorr: Because it exists outside of CRDB's implementation; its a generic proxy that currently works with CRDB and Memdb (for testing)

internal/datastore/proxy/relationshipintegrity_test.go
josephschorr: Fixed

internal/datastore/proxy/relationshipintegrity_test.go
josephschorr: Fixed

pkg/datastore/test/tuples.go
josephschorr: Before I changed it, if an error was raised, the test runner would fail but not say why (because the require executes in a goroutine). The `panic` forces the error to be shown in full

internal/datastore/proxy/relationshipintegrity.go
josephschorr: I had a pool of them at first, but removed it because I didn't think we needed to lock it. I've added the pool back in

josephschorr:
Updated

internal/datastore/crdb/crdb.go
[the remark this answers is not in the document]
ecordell: I think a feature makes more sense - we do that for other stuff like Watch, etc

internal/datastore/crdb/crdb_test.go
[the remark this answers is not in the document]
ecordell: It just seems confusing that you have to do both - if you forget to add the proxy in, you will think you have integrity enabled but there will be no validation

internal/datastore/proxy/relationshipintegrity.go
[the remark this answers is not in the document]
ecordell: Screenshot 2024-08-27 at 12 23 36 PM

pkg/tuple/tuple.go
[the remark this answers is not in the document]
ecordell: Screenshot 2024-08-27 at 12 23 56 PM

internal/datastore/crdb/crdb_test.go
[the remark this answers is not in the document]
josephschorr: Well, no, you'd have integrity *data storage* enabled, but it would fail due to checks ensuring the data is actually present. Having it be in its own proxy file makes testing easier

josephschorr:
Updated

internal/datastore/crdb/readwrite.go
```diff
@@ -104,6 +113,50 @@ var (
...
+func (rwt *crdbReadWriteTXN) insertQuery() sq.InsertBuilder {
+	return psql.Insert(rwt.tupleTableName)
+}
+
+func (rwt *crdbReadWriteTXN) queryDeleteTuples() sq.DeleteBuilder {
+	return psql.Delete(rwt.tupleTableName)
+}
+
+func (rwt *crdbReadWriteTXN) queryWriteTuple() sq.InsertBuilder {
+	if rwt.withIntegrity {
```
ecordell, line 125: sorry for the late note - can we remove these ifs from every call? i.e. pick a different implementation of `crdbReadWriteTXN` on initialization instead?

josephschorr: That would be a lot of copy-paste of code and reducing reuse a lot

ecordell (approved):

ecordell (approved):

PR #1984, opened 2024-07-11
Add ability to toggle off specific warnings via magic comments

### LINKED ISSUE #1975: Add ability to toggle certain schema warnings off
### Problem Statement

This is a refiling of https://github.com/authzed/spicedb-vscode/issues/12 but in the proper repo.

First of all, thanks for making the new VS Code extension and LSP! The coloring, hovers, and lint are awesome to have during schema development. 🎉 

Onto the issue: My schema has something similar to this:
```
definition folder {
    relation editor: user
    permission create_child_folder = editor
    // ... etc
}
```

This produces this warning:
`Permission "create_child_folder" references parent type "folder" in its name; it is recommended to drop the suffix`

I understand what this warning is getting at, but it's not exactly relevant in this case – the "child_folder" just happens to be the same resource as the parent "folder", hence the suffix. Since other warnings are still valuable, it would be great to selectively turn off this warning. 

Generally, I'd like to toggle on/off whatever warnings are best for me. Perhaps a `// zed-disable-warning: parent-name-in-suffix` at the top of the file, or something in the `.vscode/settings` file?

### Solution Brainstorm

From @josephschorr in the linked issue:

> Should be easy to add something that uses the comment metadata on the nodes and disables it

josephschorr:
Example:

```
definition test {
  // spicedb-ignore-warning: permission-references-itself
  permission view = view	
}
```

Fixes #1975

### DIFF
--- a/internal/lsp/lsp_test.go
+++ b/internal/lsp/lsp_test.go
@@ -100,7 +100,7 @@ func TestDocumentWarningDiagnostics(t *testing.T) {
 	require.Equal(t, "full", resp.Kind)
 	require.Len(t, resp.Items, 1)
 	require.Equal(t, lsp.DiagnosticSeverity(lsp.Warning), resp.Items[0].Severity)
-	require.Equal(t, `Permission "view_resource" references parent type "resource" in its name; it is recommended to drop the suffix`, resp.Items[0].Message)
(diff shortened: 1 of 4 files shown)


alecmerdler (dismissed):
Nice!

alecmerdler (approved):

Issue #1923, opened 2024-06-04
CockroachDB index should to store caveats


jzelinskie:
### Problem Statement

A production system recommended the following index: `CREATE INDEX ON relation_tuple (namespace, relation, object_id, userset_relation) STORING (caveat_name, caveat_context);`

### Solution Brainstorm

This is the closest to our primary key constraint and should improve overall check performance.

jzelinskie:
This can also be applied to our reverse index, too

lalalalatt:
@jzelinskie I want to help with this PR, and I wonder if there is a way to generate the sql migration file, or just write it manually?

lalalalatt:
I just made a pull request (https://github.com/authzed/spicedb/pull/1988) and I’m seeking some clarification on how this change can enhance the check performance.

From my understanding, there is only one type of SQL query used in the check request, and its WHERE condition is the same as the new index. Would a standard index suffice to improve the query speed? Additionally, could you please explain why STORING (caveat_name, caveat_context); is necessary?

Thank you for your assistance!

vroldanbet:
Closing this as we've recently revamped index selection. Please reopen if it manifests again.

PR #1988, opened 2024-07-18
[Perf] Add Relationship Index Storing Caveat

lalalalatt:


lalalalatt:

lalalalatt:
recheck

internal/datastore/crdb/migrations/zz_migration.0007_add_relation_index_storing_caveat.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	addRelationshipIndexStoringCaveatQuery = `
+		CREATE INDEX idx_relationship_caveat_storing ON relation_tuple (namespace, relation, object_id, userset_relation) STORING (caveat_name, caveat_context);
```
ecordell, line 11: Is there a reason you've omitted `userset_namespace` and `userset_object_id` from the index? 

This may also be better as an `ALTER` to add the caveat storage to an existing index.

lalalalatt: Ok, thanks

lalalalatt:

peterxcli:

lalalalatt:
recheck

vroldanbet:
I believe this PR was opened to address https://github.com/authzed/spicedb/issues/1923

vroldanbet (changes requested):
I have concerns about this PR. The reason the issue was created in the first place is because the CRDB UI shows it as recommended. This is likely so that the query becomes served from the index, instead of having to go to disk to fetch the rest of the columns.

I have three concerns about the proposal:
- it drops the previous indexes. This means that until the new indexes are created, the system is left without an index. This has the potential to cause an incident on a production system, as it can push the database over saturation level.
- further increases the cost on the write path, as the system now has to write more information to disk, potentially making writes slower.
- increases the amount of storage needed. While it is reasonable to trade disk for query speed, customers running CRDB Dedicated can see their bill explode, as Cockroach Labs puts a hefty price tag on the disk capacity of their managed offerings.

I don't think we can move forward with this without:
- running a load test to assess how much queries improve and if it's worth the tradeoff
- determining the % increase in disk used
- finding a way to add those indexes without causing an incident (the easiest solution being to consider running 2 migrations, or exploring `ALTER` the indexes)

lalalalatt:
> I believe this PR was opened to address https://github.com/authzed/spicedb/issues/1923

Yes, you're right, thanks. I forgot to add the description.

lalalalatt:
> I have concerns about this PR. The reason the issue was created in the first place is because the CRDB UI shows it as recommended. This is likely so that the query becomes served from the index, instead of having to go to disk to fetch the rest of the columns.
> 
> 
> 
> I have three concerns about the proposal:
> 
> - it drops the previous indexes. This means that until the new indexes are created, the system is left without an index. This has the potential to cause an incident on a production system, as it can push the database over saturation level.
> 
> - further increases the cost on the write path, as the system now has to write more information to disk, potentially making writes slower.
> 
> - increases the amount of storage needed. While it is reasonable to trade disk for query speed, customers running CRDB Dedicated can see their bill explode, as Cockroach Labs puts a hefty price tag on the disk capacity of their managed offerings.
> 
> 
> 
> I don't think we can move forward with this without:
> 
> - running a load test to assess how much queries improve and if it's worth the tradeoff
> 
> - determining the % increase in disk used
> 
> - finding a way to add those indexes without causing an incident (the easiest solution being to consider running 2 migrations, or exploring `ALTER` the indexes) 

I would try to work on the stress test to test the two metrics you mentioned.


For the index drop and create issue, I have searched on that, (currently) it seems it is impossible to alter the standard index to the storing column index. 
Moreover, maybe require to contribute this feature to cockroachDB upstream.
But, For the current workaround, I think we can create the storing column index first, after finishing indexing, then it would be safe to drop the original index.

vroldanbet:
>But, For the current workaround, I think we can create the storing column index first, after finishing indexing, then it would be safe to drop the original index.

I'm not entirely sure this would work either. There are some caveats:
- The docs indicate that index creation and drop are executed concurrently: see https://www.cockroachlabs.com/docs/stable/create-index and https://www.cockroachlabs.com/docs/stable/drop-index. Thus there is a risk the old indexes are dropped and the new indexes are not ready yet.
- on top of that, there are some gotchas with multi-line DDL, see https://www.cockroachlabs.com/docs/stable/online-schema-changes#known-limitations, which could lead partially run migrations.

This is unfortunately a difficult migration to run without some careful planning. E.g. we have customers with Terabytes worth of SpiceDB relationships in their Cockroach clusters. This migration would:
- [increase the load on the cluster](https://www.cockroachlabs.com/docs/v24.2/schema-design-indexes#index-management) for a non-trivial amount of time (proportional to the amount of data stored)
- increase write transaction latency
- increase the disk capacity required to operate, as a new index covering all tuple columns is added.

I'd like to see these optimizations make it into `main`, but it's a risky change as-is. In an ideal world:
- The code can estimate if there is enough disk available to run the migration
- the old indexes are only dropped when the new indexes have been validated to be selected by the query planner and provide a visible performance gain
- the index creation runs with low priority and limits the impact on the cluster 

Perhaps a conservative approach would be:
- To introduce logic that determines the size of the indexes to be dropped, and checks if at least the same amount plus some headroom (to account for the newly stored caveat columns) is available on cluster
- the indexes are made [`NOT_VISIBLE`](https://www.cockroachlabs.com/docs/v24.2/indexes#performance) to begin with, and a feature flag in SpiceDB enables / disables the use of new index. It's enabled by default, but can be disabled if users detect performance regressions
- We can explore if using [`transaction priority`](https://www.cockroachlabs.com/docs/stable/transactions#transaction-priorities) buys us anything when it comes to running DDL statements like CREATE INDEX or DROP INDEX.
- On a different subsequent SpiceDB release the new indexes can be made visible, and the old indexes can be dropped safely.

Would like to hear your thoughts on this issue @josephschorr

lalalalatt:
https://github.com/authzed/spicedb/blob/d77601bfe8b927c78d348144684bfa972b14e199/pkg/datastore/test/watch.go#L49-L53
If I change tuple num to 512, then the test `256-true` would pass, while its testing name would be `512-true`

vroldanbet:
>If I change tuple num to 512, then the test 256-true would pass, while its testing name would be 512-true

I think that's fine, the goal of the test is to push so many changes the consumer cannot retrieve them for the channel fast enough, a timeout will be triggered if the channel is at capacity, and if enough time elapses the channel Watch API connection will be aborted. Feel free to increase it, id say this is likely a flake.

josephschorr:
This will be included in a series of index cleanup changes coming soon; closing for now

PR #1991, opened 2024-07-22
Remove apparently unneeded COALESCE call

### LINKED ISSUE #1986: Unnecessary COALESCE Used
https://github.com/authzed/spicedb/blob/fc0d6064efe7ede178d820718655e48b1ae622bd/internal/datastore/postgres/revisions.go#L35

In the first COALESCE, it is finding a single value else return null. COALESCE are pretty expensive and are used to make sure you don't return a NULL. Kill the COALESCE here would help speed this query up.

josephschorr:
Fixes #1986


vroldanbet (approved):
LGTM

PR #1997, opened 2024-07-25
additional dispatch chunk safeguards

vroldanbet:
the parameter is wired into various parts of the application and leaving it unset can lead to runtime errors


Verolop (approved):
👍

PR #2004, opened 2024-07-30
bump Docker to address security scanners surfacing CVE

vroldanbet:
this was a more elaborate change that I anticipated. bumping docker forced me to bump buf tooling.

Changes to generate proto seems expected, but
the  "buf format" tool changed dramatically:
- does no longer accept running on a directory outside of the context directory. This is not a trivial change because it forces us to move all dependencies to the main go.mod, including buf and all proto generators
- removes commas outside of proto definition, unclear if thats valid proto or a bug on the format tool


josephschorr (approved):
LGTM

PR #2005, opened 2024-07-30
Fix experimental LookupResources2 to shear the tree earlier on indirect permissions

josephschorr:


internal/graph/lr2streams.go
```diff
@@ -194,18 +160,77 @@ func (rdc *dispatchAndCheckRunner) runChecker(ctx context.Context, collected []*
...
-	// Kick off a dispatch at the current cursor.
-	rdc.taskrunner.Schedule(func(ctx context.Context) error {
-		return rdc.runDispatch(ctx, currentCursor)
+	wrappedStream := dispatch.NewHandlingDispatchStream(ctx, func(result *v1.DispatchLookupResources2Response) error {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+
+		default:
+		}
```
ecordell, line 213:
```suggestion
		if ctx.Err() != nil {
			return ctx.Err()
		}
```

internal/graph/lr2streams.go
```diff
@@ -53,87 +60,53 @@ func runDispatchAndChecker(
...
-	})
-
-	return nil
+	return rdc.taskrunner.Wait()
 }
 
-func (rdc *dispatchAndCheckRunner) runChecker(ctx context.Context, collected []*v1.DispatchLookupResources2Response) error {
+func (rdc *checkAndDispatchRunner) runChecker(ctx context.Context, startingIndex int) error {
 	rdc.lock.Lock()
 	if rdc.ci.limits.hasExhaustedLimit() {
```
ecordell, line 93: I think the limit should just be an atomic, and then you can get rid of the locking.

internal/graph/lr2streams.go
```diff
@@ -261,7 +287,7 @@ func publishResultToParentStream(
 
 	// The cursor for the response is that of the parent response + the cursor from the result itself.
 	afterResponseCursor, err := combineCursors(
-		ci.responsePartialCursor(),
+		responseCursor,
 		result.AfterResponseCursor,
 	)
 	if err != nil {
```
ecordell, line 293: I can't comment on it, but:

```go
        mappedResource, err := foundResources.mapPossibleResource(result.Resource)
        ...
	missingContextParameters := mapz.NewSet(mappedResource.MissingContextParams...)
	missingContextParameters.Extend(result.Resource.MissingContextParams)
	missingContextParameters.Extend(additionalMissingContext)
	mappedResource.MissingContextParams = missingContextParameters.AsSlice()
```
Why not just have `mappedResource.MissingContextParams` be a set so that you can just do:
```go
        mappedResource, err := foundResources.mapPossibleResource(result.Resource)
        mappedResource.MissingContextParams.Extend(result.Resource.MissingContextParams)
        mappedResource.MissingContextParams.Extend(additionalMissingContext)
```
and then convert to a slice when you need to?

ecordell, line 293: Can't comment on this either but:

```go
	metadata := result.Metadata
	if isFirstPublishCall {
		metadata = addCallToResponseMetadata(metadata)
		metadata = combineResponseMetadata(metadata, additionalMetadata)
	} else {
		metadata = addAdditionalDepthRequired(metadata)
	}
```

`addCallToResponseMetadata` increments Dispatch and DepthRequired
`addAdditionalDepthRequired` increments just DepthRequired

Shouldn't depth and dispatch go hand in hand? Why is dispatch only incremented on the first publish for a stream - or alternatively, why is depth required incremented for messages on an existing stream?

internal/dispatch/graph/lookupresources2_test.go
```diff
@@ -844,6 +999,278 @@ func TestLookupResources2EnsureCheckHints(t *testing.T) {
...
+			relationships: []*core.RelationTuple{
+				tuple.MustParse("document:masterplan#viewer@user:tom"),
+				tuple.MustParse("document:masterplan#viewer2@user:tom"),
+				tuple.MustParse("document:masterplan#editor@user:tom"),
+				tuple.MustParse("document:masterplan#admin@user:tom"),
+			},
+			resourceRelation: RR("document", "view_and_admin"),
+			subject:          ONR("user", "tom", "..."),
+			disallowedQueries: []*core.RelationReference{
+				RR("document", "viewer"),
```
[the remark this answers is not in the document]
josephschorr: Yeah and the check is the second half of the permission

internal/graph/lr2streams.go
josephschorr: Its not; we'd have to change the implementation and its use everywhere

internal/graph/lr2streams.go
josephschorr: Its a proto so it can't be a set

josephschorr: No; we only increase the dispatch count on the first result lest we double count it (this is what caused the dispatch count bug in LR many moons ago)

josephschorr:
Updated

internal/graph/lr2streams.go
ecordell: My suggestion was to change it to an atomic and update the callsites. That can be a followup though.

ecordell (approved):

PR #2006, opened 2024-07-30
Add server version middleware to serve-testing

### LINKED ISSUE #2003: SpiceDB version middleware not enabled in `serve-testing` mode

linux (possibly others)

amd64 (possibly others)

v1.34.0

### Steps to Reproduce

1. Start SpiceDB in `serve-testing` mode, e.g. `spicedb serve-testing --load-configs /test_schema.yaml`
2. Execute `zed version` command

### Expected Result

Zed outputs both Zed and SpiceDB version, e.g.:
```
client: zed v0.19.2
service: v1.34.0
```

### Actual Result

Zed outputs the following:
```
client: zed v0.19.2
service: (unknown)
```

Note: setting Zed's log level to debug (i.e. `--log-level debug`) results in an additional message indicating the server version header could not be read:

```
DBG error reading server version response header; it may be disabled on the server
```

josephschorr:
Fixes #2003


vroldanbet (changes requested):
Can you reuse the defaults here [instead](https://github.com/authzed/spicedb/blob/f30c5cb2f2eb043da2d0fa511ab74d758636c9e4/pkg/cmd/server/defaults.go#L214)? That way we won't have drifting in the future. It also uses the middleware framework, which is more robust when it comes to dependencies between middlewares.

josephschorr:
> Can you reuse the defaults here [instead](https://github.com/authzed/spicedb/blob/f30c5cb2f2eb043da2d0fa511ab74d758636c9e4/pkg/cmd/server/defaults.go#L214)? That way we won't have drifting in the future. It also uses the middleware framework, which is more robust when it comes to dependencies between middlewares.

`serve-testing` doesn't use most of those defaults and we've have to replace a few of them. I suppose we could inject/override some, but that has its own risks

josephschorr:
Updated to use the middleware system as we discussed

pkg/cmd/server/defaults.go
```diff
@@ -168,15 +168,75 @@ const (
...
 type MiddlewareOption struct {
-	logger                zerolog.Logger
-	authFunc              grpcauth.AuthFunc
-	enableVersionResponse bool
-	dispatcher            dispatch.Dispatcher
-	ds                    datastore.Datastore
-	enableRequestLog      bool
-	enableResponseLog     bool
-	disableGRPCHistogram  bool
+	Logger                  zerolog.Logger      `debugmap:"hidden"`
```
ecordell, line 173: Why are all of these exported now? I don't see that being used anywhere.

ecordell (approved):

PR #2014, opened 2024-08-07
Change the filter count check to a debug assertions

josephschorr:


pkg/spiceerrors/assert_on.go
```diff
@@ -0,0 +1,12 @@
+//go:build ci
+// +build ci
+
+package spiceerrors
+
+import "fmt"
+
+func DebugAssert(condition func() bool, format string, args ...any) {
+	if !condition() {
```
ecordell, line 9: why not have the condition fn return an error?

then you can do `DebugAssert(func() {return fmt.Errorf("whatever")}`

josephschorr: I personally think this reads better: it ensures that the conditional is a boolean and we could take other actions too (like log)

ecordell: But you won't be able to return messages that depend on computation in the assert fn

josephschorr: That's fine though; its a debug assertion

pkg/spiceerrors/assert_off.go
```diff
@@ -0,0 +1,8 @@
+//go:build !ci
+// +build !ci
+
+package spiceerrors
+
```
ecordell, line 5: Add a comment that this is always a no-op for prod builds

pkg/spiceerrors/assert_on.go
```diff
@@ -0,0 +1,12 @@
+//go:build ci
+// +build ci
+
+package spiceerrors
+
+import "fmt"
+
```
ecordell, line 7: Add a comment that this is only enabled in CI builds (and maybe point to the other impl for prod builds)

pkg/spiceerrors/assert_off.go
josephschorr: Added

pkg/spiceerrors/assert_on.go
josephschorr: Added

ecordell (approved):

PR #2017, opened 2024-08-09
Ensure debug information is returned for recursive checks that dispatch

josephschorr:


pkg/development/devcontext.go
```diff
@@ -340,6 +341,19 @@ func distinguishGraphError(ctx context.Context, dispatchError error, source devi
 		}, nil
 	}
 
+	details, ok := spiceerrors.GetDetails[*errdetails.ErrorInfo](dispatchError)
```
ecordell, line 344: why doesn't the previous case handle this?

josephschorr: Because the previous case is a non-gRPC wrapped error from the *local* dispatch vs a gRPC error from a remote dispatch

ecordell (approved):

PR #2018, opened 2024-08-12
Fix logging behavior around setting goproc limits

### LINKED ISSUE #2010: GOMAXPROCS, GOMEMLIMIT logs emitted on all CLI commands

linux, macos, windows

amd64, arm64

v.1.31.1

### Steps to Reproduce

1. Install SpiceDB v1.31.1
2. Run any command

### Expected Result

log-level is respected

### Actual Result

```
$ spicedb version
{"level":"debug","message":"maxprocs: Leaving GOMAXPROCS=10: CPU quota undefined"}
{"level":"info","time":"2024-08-01T11:25:19-04:00","GOMEMLIMIT":61847529062,"package":"github.com/KimMachineGun/automemlimit/memlimit","message":"GOMEMLIMIT is updated"}
spicedb v1.35.1
{"level":"debug","message":"maxprocs: No GOMAXPROCS change to reset"}
```

tstirrat15:
Fixes #2010 

## Description
In the implementation of #1921, we added memory and proc limits at the top level, along with logging from those libraries. Because it sits above the level of any of the commands, it didn't respect log levels. This fixes it by moving it into a `PreRunE` wrapper that sits below the level of log setup, which also means that it will be correctly formatted when running from the command line, and also only apply to commands that need it (we probably don't need it on `version`, for example).

## Notes
@ecordell made the point that this would probably be useful in Materialize as well; if it's desired, we can push this up into `cobrautil` and pull it into both projects that way.

## Questions
* Are there any downsides to pushing this limit setting down? It makes sense to me that it should happen as early as possible in process bootstrapping, and I don't know enough about the Go runtime to know whether these are things that can be modified willy-nilly on a running process.

## Changes
* Move setup logic from main into a dedicated `RunE` function
* Use that `RunE` function in the `DefaultPreRunE` function

## Testing
See that `spicedb version` no longer reports changes from the limit libraries. See that if you set `--log-level debug`, you see output related to these libs nicely-formatted early in the output of `spicedb serve` and friends.


pkg/cmd/limits/limits.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/spf13/cobra"
+	"go.uber.org/automaxprocs/maxprocs"
+
+	log "github.com/authzed/spicedb/internal/logging"
+)
+
+// SetLimitsRunE wraps the RunFunc with setup logic for memory limits and maxprocs
+// limits for the go process. Implementing it here means that we can use already-
+// established loggers and wrap only those commands that require it.
+func SetLimitsRunE() cobrautil.CobraRunFunc {
```
alecmerdler, line 19: This seems useful enough to put into `cobrautil` so it can be used across other projects.

alecmerdler (commented):
I would vote to move this into `cobrautil`.

alecmerdler (approved):
Nice

PR #2027, opened 2024-08-15
Ensure all resources are returned for relation check when caveats are specified

### LINKED ISSUE #2026: Permission check ignores further relationships when a relationship with caveat does not match

linux

amd64

v1.35.2

### docker-compose.yaml
```yaml
services:
  spicedb:
    image: "authzed/spicedb"
    command: "serve"
    restart: "always"
    ports:
      - "50051:50051"
    environment:
      - "SPICEDB_GRPC_PRESHARED_KEY=123456"
      - "SPICEDB_DATASTORE_ENGINE=postgres"
      - "SPICEDB_DATASTORE_CONN_URI=postgres://postgres:postgres@database:5432/spicedb?sslmode=disable"
    depends_on:
      - "migrate"

  migrate:
    image: "authzed/spicedb"
    command: "datastore migrate head"
    restart: "on-failure"
    environment:
      - "SPICEDB_DATASTORE_ENGINE=postgres"
      - "SPICEDB_DATASTORE_CONN_URI=postgres://postgres:postgres@database:5432/spicedb?sslmode=disable"
    depends_on:
      - "database"
      
  database:
    image: "postgres"
    command: -c track_commit_timestamp=on
    ports:
      - "5432:5432"
    environment:
      - "POSTGRES_DB=spicedb"
      - "POSTGRES_PASSWORD=postgres"
```
### schema.yaml
```yaml
schema: >-
  definition user {}

  /**
   * The count of current entries in the database must be lower than the limit.
   */
  caveat write_limit(limit uint, count uint) {
    count < limit
  }

  definition role {
    relation member: user
  }

  definition database {
      relation writer: role#member with write_limit

      permission write = writer
  }
relationships: |-
  database:listings#writer@role:default#member[write_limit:{"limit":2}]
  database:listings#writer@role:premium#member[write_limit:{"limit":4}]
  role:default#member@user:bob
  role:premium#member@user:bob
assertions:
  assertTrue:
    - 'database:listings#write@user:bob with {"count":3}'
validation: {}
```

## Run
1.  `docker compose -f ./docker-compose.yaml up`
2. `zed import ./schema.yaml`
3. `zed permission check database:listings write user:bob --caveat-context '{"count":3}' --consistency-full`

### Expected Result

I expected the permission check to succeed, which is the case when running Spicedb with the in-memory database or using `zed validate ./schema.yaml`.

### Actual Result

The permission check fails when using Postgres, MySql or the zed terminal on the Authzed playground. 
However, if I switch the values for the `limit` parameter and have relationships like
```
 database:listings#writer@role:default#member[write_limit:{"limit":4}]
 database:listings#writer@role:premium#member[write_limit:{"limit":2}]
```
instead of

```
database:listings#writer@role:default#member[write_limit:{"limit":2}]
database:listings#writer@role:premium#member[write_limit:{"limit":4}]
```
the permission check succeeds.
For me, it looks like Spicedb stops evaluation after encountering the first caveat relationship and ignores any further relationships.

josephschorr:
Fixes #2026

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -466,6 +466,7 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 	// Find the subjects over which to dispatch.
 	subjectsToDispatch := tuple.NewONRByTypeSet()
 	relationshipsBySubjectONR := mapz.NewMultiMap[string, *core.RelationTuple]()
+	hasCaveats := false
 
 	for tpl := it.Next(); tpl != nil; tpl = it.Next() {
 		if it.Err() != nil {
@@ -479,6 +480,9 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 
 		subjectsToDispatch.Add(tpl.Subject)
 		relationshipsBySubjectONR.Add(tuple.StringONR(tpl.Subject), tpl)
+		if tpl.Caveat != nil && tpl.Caveat.CaveatName != "" {
+			hasCaveats = true
+		}
 	}
 	it.Close()
 
@@ -499,6 +503,13 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 		dispatchChunkCountHistogram.Observe(chunkCount)
 	})
 
+	// If there are caveats on the incoming relationships, then we must require all results to be
+	// found, as we need to ensure that all caveats are used for building the final expression.
+	resultsSetting := crc.resultsSetting
+	if hasCaveats {
+		resultsSetting = v1.DispatchCheckRequest_REQUIRE_ALL_RESULTS
+	}
+
 	// Dispatch and map to the associated resource ID(s).
 	result := union(ctx, crc, toDispatch, func(ctx context.Context, crc currentRequestContext, dd directDispatch) CheckResult {
 		childResult := cc.dispatch(ctx, crc, ValidatedCheckRequest{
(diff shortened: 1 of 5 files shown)


internal/graph/check.go
```diff
@@ -499,14 +503,21 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 		dispatchChunkCountHistogram.Observe(chunkCount)
 	})
 
+	// If there are caveats on the incoming relationships, then we must require all results to be
+	// found, as we need to ensure that all caveats are used for building the final expression.
+	resultsSetting := crc.resultsSetting
+	if hasCaveats {
+		resultsSetting = v1.DispatchCheckRequest_REQUIRE_ALL_RESULTS
```
tstirrat15, line 510: Understanding check:

The issue is that we were short-circuiting when we found a single relation which matched the check but whose caveat was false and calling it false.

This fix says that when we have a caveat, we must find all relations that match the check so that we can evaluate the caveats on all of them and then `OR` them together so that we can decide whether the overall check was true.

The more efficient fix would be to continue evaluating relationships that match the check until we find one that also matches the caveat or we exhaust the set.

tstirrat15 (approved):
LGTM, see comment

internal/graph/check.go
tstirrat15: Ah I'm reading slack and realizing that my "more efficient fix" isn't how it would work for the caching reasons described.

josephschorr: Correct. We cannot just eval here because we build the *full* caveat expression and then evaluate outside of the dispatcher.

ecordell (dismissed):
LGTM

ecordell (approved):
it would be nice if there were a helper that could assert that two cel expressions are equivalent, instead of providing two options in the test

PR #2029, opened 2024-08-19
Cleanup handling of internal errors in Check dispatch

josephschorr:


internal/graph/check.go
```diff
@@ -404,22 +398,12 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 				return checkResultError(NewCheckFailureErr(it.Err()), emptyMetadata)
 			}
 
+			spiceerrors.DebugAssert(func() bool {
+				return tuple.OnrEqualOrWildcard(tpl.Subject, crc.parentReq.Subject)
```
tstirrat15, line 402: Curious - is this somehow capturing the local context for logging purposes? Or do we have a good enough idea of what's going on here that we don't need (for example) the tuple string?

tstirrat15 (approved):
LGTM

internal/graph/check.go
josephschorr: It would panic when tests are being run, so we'd have a full stack trace

PR #2030, opened 2024-08-19
Remove unnecessary branch from limit logic

tstirrat15:
## Description
It seems like it's a vestige of when there was additional logic in this file - otherwise the `== 0` case handles the decrementing properly.

## Changes
* Remove special handling of `== 1` case.
## Testing
Review. See that tests pass.


josephschorr (approved):
LGTM

PR #2037, opened 2024-08-20
Check data structure improvements

josephschorr:
Moves around and cleans up some data structures used in check


internal/developmentmembership/onrset_test.go
tstirrat15, line 1: Are there there any meaningful changes between this and the deleted file?

internal/developmentmembership/trackingsubjectset.go
```diff
@@ -100,9 +100,7 @@ func (tss *TrackingSubjectSet) getSetForKey(key string) datasets.BaseSubjectSet[
 			fs.excludedSubjects = excludedSubjects
 			fs.caveatExpression = caveatExpression
 			for _, source := range sources {
-				if source.relationships != nil {
```
tstirrat15, line 103: Why isn't this check necessary  anymore?

internal/graph/checkdispatchset.go
```diff
@@ -0,0 +1,166 @@
...
+		sort.Slice(entries, func(i, j int) bool {
+			iHasCaveat := entries[i].hasIncomingCaveats
+			jHasCaveat := entries[j].hasIncomingCaveats
+			if iHasCaveat == jHasCaveat {
+				return entries[i].objectID < entries[j].objectID
+			}
+			return iHasCaveat && !jHasCaveat
+		})
+
+		chunkCount := 0.0
```
tstirrat15, line 132: The float here surprises me - is that intentional?

tstirrat15 (approved):
See comments, otherwise LGTM

internal/developmentmembership/foundsubject.go
```diff
@@ -28,7 +28,7 @@ type FoundSubject struct {
 
 	// relations are the relations under which the subject lives that informed the locating
 	// of this subject for the root ONR.
-	relationships *tuple.ONRSet
+	relationships ONRSet
```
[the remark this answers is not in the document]
josephschorr: Since it wraps another reference now, it possibly removes another small item on the heap

internal/developmentmembership/onrset_test.go
josephschorr: No; it was moved

internal/developmentmembership/trackingsubjectset.go
josephschorr: I moved it into the set impl

internal/graph/checkdispatchset.go
josephschorr: Yes; we use it for the Prometheus metric below

internal/graph/check.go
```diff
@@ -447,56 +441,30 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
...
 	// Dispatch and map to the associated resource ID(s).
-	result := union(ctx, crc, toDispatch, func(ctx context.Context, crc currentRequestContext, dd directDispatch) CheckResult {
+	toDispatch := checksToDispatch.dispatchChunks(crc.dispatchChunkSize)
+	result := union(ctx, crc, toDispatch, func(ctx context.Context, crc currentRequestContext, dd checkDispatchChunk) CheckResult {
+		// If there are caveats on any of the incoming relationships for the subjects to dispatch, then we must require all
+		// results to be found, as we need to ensure that all caveats are used for building the final expression.
+		resultsSetting := crc.resultsSetting
+		if dd.hasIncomingCaveats {
+			resultsSetting = v1.DispatchCheckRequest_REQUIRE_ALL_RESULTS
+		}
```
vroldanbet, line 463: so the idea here is that because `dispatchChunks` now clusters non-ellipsis subjects relation by "caveat yes/no", we will reduce the number of dispatches that require _all_ because we've hopefully put all caveated the caveated ones in the minimum amount of chunks possible.

And this does not change with intersections and exclusions, since all dispatched elements have to be resolved.

I wonder how likely it is to find the "mixed" scenario (possible subject types with/without caveats). Because we are adding overhead by requiring sorting potentially very wide relations. Have you conducted benchmarks to understand the impact? wouldn't it perhaps be better to determine ahead of time if caveats are possible in the relation's type system? if caveats were present, the `CheckDispatchSet` will do the sorting, otherwise, it does not.

internal/graph/check.go
```diff
@@ -513,21 +481,24 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
...
 	return combineResultWithFoundResources(result, foundResources)
 }
 
-func mapFoundResources(result CheckResult, resourceType *core.RelationReference, relationshipsBySubjectONR *mapz.MultiMap[string, *core.RelationTuple]) CheckResult {
+func mapFoundResources(result CheckResult, resourceType relationRef, checksToDispatch *checkDispatchSet) CheckResult {
 	// Map any resources found to the parent resource IDs.
 	membershipSet := NewMembershipSet()
 	for foundResourceID, result := range result.Resp.ResultsByResourceId {
-		subjectKey := tuple.StringONRStrings(resourceType.Namespace, foundResourceID, resourceType.Relation)
+		resourceIDAndCaveats := checksToDispatch.mappingsForSubject(resourceType.namespace, foundResourceID, resourceType.relation)
```
vroldanbet, line 494: I think it's a good idea to avoid those string allocations, in previous profiling sessions I observed it amounting to a lot of garbage. The new tracker should help with that 👍🏻 

Creating these structs instead of synthesizing a new string as a key is a good strategy we can apply everywhere in the code base.

internal/graph/check.go
```diff
@@ -513,21 +481,24 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
...
+
+		spiceerrors.DebugAssert(func() bool {
+			return len(resourceIDAndCaveats) > 0
+		}, "found resource ID without associated caveats")
```
vroldanbet, line 498: Aren't we already asserting the same in `mappingsForSubject`?

internal/graph/check.go
```diff
@@ -835,11 +785,10 @@ func checkIntersectionTupleToUserset(
...
 
 			// Add any caveats on the subject from the starting relationship(s) as well.
-			subjectKey := tuple.StringONR(subject)
-			tuples, _ := relationshipsBySubjectONR.Get(subjectKey)
-			for _, relationTuple := range tuples {
-				if relationTuple.Caveat != nil {
-					caveats = append(caveats, wrapCaveat(relationTuple.Caveat))
+			resourceIDAndCaveats := checksToDispatch.mappingsForSubject(subject.Namespace, subject.ObjectId, subject.Relation)
+			for _, riac := range resourceIDAndCaveats {
+				if riac.caveat != nil {
```
vroldanbet, line 790: `mappingForSubject` asserts that a caveat exists. In runtime it wouldn't panic, and this portion would just ignore the resource, which does not seem right, shouldn't we return a spiceerror?

internal/graph/checkdispatchset_test.go
```diff
@@ -0,0 +1,387 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			set := newCheckDispatchSet()
+			for _, rel := range tc.relationships {
+				set.addForRelationship(rel)
+			}
+
+			chunks := set.dispatchChunks(tc.dispatchChunkSize)
+			for _, c := range chunks {
+				sort.Strings(c.resourceIds)
+			}
```
vroldanbet, line 374: why sorting, isn't this already supposed to be sorted by `dispatchChunks`?

internal/graph/checkdispatchset_test.go
```diff
@@ -0,0 +1,387 @@
...
+					resourceIds:        []string{"2", "3"},
+					hasIncomingCaveats: false,
+				},
+				{
+					resourceType:       relationRef{namespace: "group", relation: "member"},
+					resourceIds:        []string{"1", "4"},
+					hasIncomingCaveats: true,
+				},
+			},
+			map[string][]resourceIDAndCaveat{
```
vroldanbet, line 273: why isn't `group:4#member` showing up here?

internal/developmentmembership/onrset_test.go
```diff
@@ -0,0 +1,144 @@
+package developmentmembership
```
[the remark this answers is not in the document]
josephschorr: I didn't benchmark this because its only used for development tooling

internal/graph/check.go
josephschorr: something like `relation viewer: org#member | group#member with caveat` is likely fairly common and would be immediately improved by it. We can skip doing the sorting if no caveats are found at all (its easy to do) but then it makes the test's expected chunks inconsistent

internal/graph/check.go
josephschorr: Yep

internal/graph/check.go
josephschorr: Yes, but there are other callers so its best to be sure

internal/graph/check.go
josephschorr: `mappingForSubject` ensures that the *resource* ID exists; not that it is caveated. If there is no caveat, this entry will be `nil`

internal/graph/checkdispatchset_test.go
josephschorr: Was added before I had the call sorting it

internal/graph/checkdispatchset_test.go
josephschorr: Forgot to add it and the test only verifies what I put in that map

josephschorr:
Updated

internal/graph/check.go
vroldanbet: I guess what I'm mostly interested is that this does not regress the usual cases. If you could run a benchmark, that would help us understand the impact of the new changes.

josephschorr: It can't regress the cases vs what we have now in HEAD: the previous fix did result in worse performance; this resets it as best as we can in these cases

vroldanbet (approved):

PR #2038, opened 2024-08-23
Have diffexpr handle the case of adding to a single child expression

josephschorr:
Note that this still doesn't apply for exclusions, and they have a different expression tree format

### DIFF
--- a/pkg/diff/namespace/diffexpr.go
+++ b/pkg/diff/namespace/diffexpr.go
@@ -17,6 +17,11 @@ const (
 
 	// ExpressionChildrenChanged indicates that the children of the expression were changed.
 	ExpressionChildrenChanged ExpressionChangeType = "children-changed"
+
+	// ExpressionOperationExpanded indicates that the operation type of the expression was expanded
+	// from a union of a single child to multiple children under a union, intersection or another
+	// operation.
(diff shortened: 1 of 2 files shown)


pkg/diff/namespace/diffexpr.go
```diff
@@ -144,12 +149,20 @@ func DiffExpressions(existing *core.UsersetRewrite, updated *core.UsersetRewrite
...
 	if existingType != updatedType {
-		return &ExpressionDiff{
-			existing: existing,
-			updated:  updated,
-			change:   ExpressionOperationChanged,
-		}, nil
+		// If the expression has changed from a union with a single child, then
+		// treat this as a special case, since there wasn't really an operation
+		// before.
+		if existingType != "union" || len(existingOperation.Child) != 1 {
```
vroldanbet, line 157: why not returning expanded with intersections and exclusions too?

vroldanbet (approved):

pkg/diff/namespace/diffexpr.go
josephschorr: We don't create intersections/exclusions with a single child. Only union because we needed *some* operation to place there during compilation

PR #2041, opened 2024-08-26
Add a default connect timeout for watch in CRDB driver

josephschorr:
This ensures that if the new connection used for watch hangs, it will timeout

### DIFF
--- a/internal/datastore/crdb/watch.go
+++ b/internal/datastore/crdb/watch.go
@@ -88,12 +88,17 @@ func (cds *crdbDatastore) watch(
 	defer close(updates)
 	defer close(errs)
 
+	watchConnectTimeout := opts.WatchConnectTimeout
+	if watchConnectTimeout <= 0 {
+		watchConnectTimeout = cds.watchConnectTimeout
+	}
+
 	// get non-pooled connection for watch
 	// "applications should explicitly create dedicated connections to consume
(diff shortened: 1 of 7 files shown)


tstirrat15 (dismissed):
LGTM

internal/datastore/crdb/watch.go
```diff
@@ -24,6 +24,7 @@ import (
 const (
 	queryChangefeed       = "EXPERIMENTAL CHANGEFEED FOR %s WITH updated, cursor = '%s', resolved = '%s', min_checkpoint_frequency = '0';"
 	queryChangefeedPreV22 = "EXPERIMENTAL CHANGEFEED FOR %s WITH updated, cursor = '%s', resolved = '%s';"
+	watchConnectTimeout   = 1 * time.Second
```
vroldanbet, line 27: I'd suggest making this more conservative, and tunable. The code moves from no timeout at all to 1 second, and that could backfire in situations where a CRDB cluster is at capacity

josephschorr: That's actually kind of the point - it would fail fast and then they'd retry. It only applies on the initial connect

josephschorr: Made configurable

vroldanbet (approved):

PR #2043, opened 2024-08-27
Add analyzer to enforce usage of VT versions of marshalling and unmarshalling

tstirrat15:
## Description
Calling `proto.Marshal` requires calling reflection code, where the `VT` methods are called directly and are more efficient. This adds a linter that enforces that.

## Note
The exclusions ended up being fairly broad because of the size of the excluded packages. One thing that was discussed was implementing `// nolint` comments for this analyzer, but that would require either writing that manually into the analyzer or making use of a package that provides that functionality, and I wasn't sure whether we wanted to take that on right now.

## Changes
* Add new analyzer and tests
* Add analyzer to toolchain
* Add some exceptions for the analyzer
* Fix the problems identified by the analyzer
* Bump vtproto version to pick up changes that deal with nested one_ofs + messages like we have in the SetOperation definition
* Regenerate the protoc-generated files according to the above

## Testing
Review the code. See that lints go green.


tstirrat15:
Had a conversation with @josephschorr where it'd be nice to have a sort of snapshot test that serializes a schema out to protobuf bytes using both VT and non-VT, and then the test deserializes using the opposite and then asserts that they agree, with the idea being that it's a snapshot test that will alert us if our serialization or deserialization changes meaningfully.

tools/analyzers/protomarshalcheck/protomarshalcheck.go
```diff
@@ -0,0 +1,106 @@
...
+	"go/ast"
+	"regexp"
+	"strings"
+
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/ast/inspector"
+)
+
+func sliceMap(s []string, f func(value string) string) []string {
```
jzelinskie, line 15: I think there's a function for this in the lo library we import some places

tstirrat15: Is it named the same? I copied this code from an existing analyzer.

magefiles/lint.go
```diff
@@ -93,6 +93,12 @@ func (Lint) Analyzers() error {
 		"-paniccheck.skip-files=_test,zz_",
 		"-zerologmarshalcheck",
 		"-zerologmarshalcheck.skip-files=_test,zz_",
+		"-protomarshalcheck",
+		// Skip generated protobuf files for this check
+		"-protomarshalcheck.skip-files=.pb",
+		// Skip our dispatch codec logic that explicitly calls MarshalVT with proto.Marshal as a fallback
+		// Skip our internal telemetry reporter which uses a prometheus proto definition that we don't control
+		"-protomarshalcheck.skip-pkg=github.com/authzed/spicedb/pkg/proto/dispatch/v1,github.com/authzed/spicedb/internal/telemetry",
```
vroldanbet, line 101: aren't these ignored open-ended? Can we list by file at least? or better, can we add a `// nolint` annotation like what golangci-lint does?

tools/analyzers/protomarshalcheck/protomarshalcheck.go
vroldanbet: https://github.com/samber/lo?tab=readme-ov-file#map

vroldanbet (dismissed):
LGTM, just 2 nits

magefiles/lint.go
tstirrat15: Yeah, the issue is that `// nolint` needs to be implemented at the analyzer level - there isn't a way to do it at the parent framework level. There's this package that I was considering but I wasn't sure whether to treat it as in-scope for this work: https://pkg.go.dev/github.com/kyoh86/nolint#section-readme

internal/namespace/test/serialization_test.go
```diff
@@ -0,0 +1,159 @@
...
+					vtSerialized, err := caveatDef.MarshalVT()
+					require.NoError(err)
+					err = os.WriteFile(fmt.Sprintf("testdata/proto/%s-caveat-%s.vtproto", test.schema, caveatDef.Name), vtSerialized, 0o644)
+					require.NoError(err)
+				}
+			} else {
+				files, err := os.ReadDir("testdata/proto")
+				require.NoError(err)
+
+				definitionFiles := map[string]struct{}{}
```
josephschorr, line 74: you can use our custom `*mapz.NewSet[string]` here

internal/namespace/test/serialization_test.go
```diff
@@ -0,0 +1,159 @@
...
+				}
+
+				// For each namespace, we want to assert that all of the following are equivalent:
+				// standard serialization -> standard deserialization
+				// vt serialization -> standard deserialization
+				// standard serialization -> vt deserialization
+				// This is to validate that the vt serialization/deserialization isn't doing anything unexpected
+				// compared to the "official" serialization/deserialization
+
+				for definition := range definitions {
```
josephschorr, line 118: probably can extract this into a helper function and then reuse for caveats below; it can take the definition type in as a generic

magefiles/lint.go
vroldanbet: sounds reasonable 👍

vroldanbet (dismissed):
linter seems to be failing in CI, but changes look good to me!

vroldanbet (approved):

josephschorr (dismissed):
LGTM

josephschorr (approved):

PR #2044, opened 2024-08-28
Add configurable max buffer size for watch change tracker

josephschorr:


internal/datastore/common/changes.go
```diff
@@ -81,10 +87,14 @@ func (ch *Changes[R, K]) AddRelationshipChange(
...
 }
 
-func (ch *Changes[R, K]) recordForRevision(rev R) changeRecord[R] {
+func (ch *Changes[R, K]) recordForRevision(rev R) (changeRecord[R], error) {
 	k := ch.keyFunc(rev)
 	revisionChanges, ok := ch.records[k]
 	if !ok {
+		if ch.maxSize > 0 && uint64(len(ch.records)) >= ch.maxSize {
+			return changeRecord[R]{}, NewMaximumChangesSizeExceededError(ch.maxSize)
+		}
```
vroldanbet, line 96: The problem with this is that this is an entry count, not byte count. Relationship sizes vastly differ from a system to another, and in resource-constrained deployments, it may require a lot of experimentation to figure the right number. I think it's bad UX, and SpiceDB already has enough flags.

The first question is: can we come up with a default that works _most of the time_?

If instead this was bytes or a percent of available memory, it would be relatively easy to set expectations around this behaviour. The main use-case is to store relationships, so you could call `.SizeVT()` to compute this.

On top of that, we need to account for the [default cache sizes](https://github.com/authzed/spicedb/blob/06f52bf1016437be7af3f00815d72c37ef394395/pkg/cmd/serve.go#L24-L49). We currently use a [75% of free available system memory](https://github.com/authzed/spicedb/blob/a90724e32c3327b134f094fd3487813a89a2f43a/pkg/cmd/server/cacheconfig.go#L29) for caches. That leaves a 25% of free available memory for _everything else_. A default relative percentage would likely work best.

If we want to make it easy to use it should support absolute values (as a humanized size string like in kubernetes resource requests / limits) and relative values (a percentage of available memory limits), and the default should be relative (3% or 5%?)

vroldanbet (changes requested):
I think the idea should help, but I'd suggest making this easier to tune and ideally a setting that one rarely has to change.

internal/datastore/common/changes.go
josephschorr: The downside of using a byte count is having to either sum up the sizes every time *or* needing to maintain a roving counter that goes up and down. This also means having to do so for namespace and caveat entries as well. We can do so, but since the goal is to put *some* cap on the size, I started with something simpler. If you wish, I can change it to be based on a memory byte limit, but it will require extra computation

vroldanbet: As I mentioned, you can update the signature of the generics to have an interface that implements a method like `SizeVT()` which should be fairly efficient to compute the size of the proto.

I understand a cap is easy, but it becomes a maintenance burden once it's added. We should strive for flags that work _most of the time_ out of the box.

josephschorr: Changed

vroldanbet (approved):
Great stuff, thanks for addressing the feedback!

PR #2046, opened 2024-08-29
Fix data type for pg_class relcount

### LINKED ISSUE #2045: Upgrade from 1.28.0 to 1.35.3 brings error message about postgres stats

linux

amd64

Upgraded from 1.28.0 to 1.35.3

### Steps to Reproduce

User report, [here on discord](https://discord.com/channels/844600078504951838/844600078948630559/1278752964949053460)

Hello, we just upgraded to the latest version and started getting an error on start up (these errors are not affecting any functionality). The error seems to come from the fact that it reads `pg_class.reltuples` into an int, but in postgres, it is stored as a `real`/`float4`, which leads to parsing errors. 
```
unable to query DB stats: unable to read relationship count: can't scan into dest[0]: strconv.ParseInt: parsing "1.018981e+06": invalid syntax
```
Running the SQL command my self, I get:
```
authz=> SELECT reltuples FROM pg_class WHERE relname = 'relation_tuple';
  reltuples
--------------
 1.018981e+06
(1 row)
```
Location of the suspected bug: https://github.com/authzed/spicedb/blob/fd2978a1771fa9b09fdfb3ba05ba32bbd4006d69/internal/datastore/postgres/stats.go#L60

SpiceDB: v.1.35.3
Postgres: 16.1

### Expected Result

No error message

### Actual Result

Error message

josephschorr:
It is a float4. See: https://www.postgresql.org/docs/current/catalog-pg-class.html

Fixes #2045

### DIFF
--- a/internal/datastore/postgres/stats.go
+++ b/internal/datastore/postgres/stats.go
@@ -57,7 +57,7 @@ func (pgd *pgDatastore) Statistics(ctx context.Context) (datastore.Stats, error)
 
 	var uniqueID string
 	var nsDefs []datastore.RevisionedNamespace
-	var relCount int64
+	var relCount float64
 	if err := pgx.BeginTxFunc(ctx, pgd.readPool, pgd.readTxOptions, func(tx pgx.Tx) error {
 		if pgd.analyzeBeforeStatistics {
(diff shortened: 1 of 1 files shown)


tstirrat15 (approved):
LGTM

PR #2061, opened 2024-09-09
Fix security errors in lint steps

tstirrat15:
## Description
A new go version was published that fixes a security vulnerability. This takes that patch.

## Changes
* Bump go version in root and magefile
## Testing
Review. See that things pass.

### DIFF
--- a/internal/datastore/common/changes.go
+++ b/internal/datastore/common/changes.go
@@ -120,9 +120,10 @@ func (ch *Changes[R, K]) adjustByteSize(item sized, delta int) error {
 		return spiceerrors.MustBugf("byte size underflow")
 	}
 
-	// We checked for underflow above, so the current byte size
-	// should fit in a uint64
-	currentByteSize, _ := safecast.ToUint64(ch.currentByteSize)
+	currentByteSize, err := safecast.ToUint64(ch.currentByteSize)
+	if err != nil {
+		return spiceerrors.MustBugf("could not cast currentByteSize to uint64: %v", err)
+	}
 
 	if currentByteSize > ch.maxByteSize {
 		return datastore.NewMaximumChangesSizeExceededError(ch.maxByteSize)
--- a/internal/datastore/common/changes.go
+++ b/internal/datastore/common/changes.go
@@ -6,6 +6,8 @@ import (
 
 	"golang.org/x/exp/maps"
 
+	"github.com/ccoveille/go-safecast"
+
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/pkg/datastore"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
@@ -118,7 +120,11 @@ func (ch *Changes[R, K]) adjustByteSize(item sized, delta int) error {
 		return spiceerrors.MustBugf("byte size underflow")
 	}
 
-	if ch.currentByteSize > int64(ch.maxByteSize) {
+	// We checked for underflow above, so the current byte size
+	// should fit in a uint64
+	currentByteSize, _ := safecast.ToUint64(ch.currentByteSize)
+
+	if currentByteSize > ch.maxByteSize {
 		return datastore.NewMaximumChangesSizeExceededError(ch.maxByteSize)
 	}
 
--- a/internal/datastore/common/gc_test.go
+++ b/internal/datastore/common/gc_test.go
@@ -68,7 +68,7 @@ func (gc *fakeGC) DeleteBeforeTx(_ context.Context, rev datastore.Revision) (Del
 
 	revInt := rev.(revisions.TransactionIDRevision).TransactionID()
 
-	return gc.deleter.DeleteBeforeTx(int64(revInt))
+	return gc.deleter.DeleteBeforeTx(revInt)
 }
 
 func (gc *fakeGC) HasGCRun() bool {
@@ -101,22 +101,22 @@ func (gc *fakeGC) GetMetrics() gcMetrics {
 
 // Allows specifying different deletion behaviors for tests
 type gcDeleter interface {
-	DeleteBeforeTx(revision int64) (DeletionCounts, error)
+	DeleteBeforeTx(revision uint64) (DeletionCounts, error)
 }
 
 // Always error trying to perform a delete
 type alwaysErrorDeleter struct{}
 
-func (alwaysErrorDeleter) DeleteBeforeTx(_ int64) (DeletionCounts, error) {
+func (alwaysErrorDeleter) DeleteBeforeTx(_ uint64) (DeletionCounts, error) {
 	return DeletionCounts{}, fmt.Errorf("delete error")
 }
 
 // Only error on specific revisions
 type revisionErrorDeleter struct {
-	errorOnRevisions []int64
+	errorOnRevisions []uint64
 }
 
-func (d revisionErrorDeleter) DeleteBeforeTx(revision int64) (DeletionCounts, error) {
+func (d revisionErrorDeleter) DeleteBeforeTx(revision uint64) (DeletionCounts, error) {
 	if slices.Contains(d.errorOnRevisions, revision) {
 		return DeletionCounts{}, fmt.Errorf("delete error")
 	}
@@ -178,7 +178,7 @@ func TestGCFailureBackoffReset(t *testing.T) {
 		// Error on revisions 1 - 5, giving the exponential
 		// backoff enough time to fail the test if the interval
 		// is not reset properly.
-		errorOnRevisions: []int64{1, 2, 3, 4, 5},
+		errorOnRevisions: []uint64{1, 2, 3, 4, 5},
 	})
 
 	ctx, cancel := context.WithCancel(context.Background())
--- a/internal/datastore/crdb/pool/balancer.go
+++ b/internal/datastore/crdb/pool/balancer.go
@@ -3,11 +3,13 @@ package pool
 import (
 	"context"
 	"hash/maphash"
+	"math"
 	"math/rand"
 	"slices"
 	"strconv"
 	"time"
 
+	"github.com/ccoveille/go-safecast"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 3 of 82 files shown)


ecordell (dismissed):

ecordell (commented):
there's another go.mod in e2e as well

internal/datastore/postgres/xid8.go
```diff
@@ -139,7 +139,7 @@ func (c Uint64Codec) DecodeDatabaseSQLValue(m *pgtype.Map, oid uint32, format in
 	if err != nil {
 		return nil, err
 	}
-	return int64(n), nil
+	return n, nil
```
tstirrat15, line 142: This is the only change that makes me really uneasy. I'm not sure why this cast was here in the first place. Lint isn't complaining; we'll see whether there are test issues.

internal/datastore/revisions/commonrevision.go
```diff
@@ -77,10 +77,3 @@ type WithTimestampRevision interface {
 	TimestampNanoSec() int64
 	ConstructForTimestamp(timestampNanoSec int64) WithTimestampRevision
 }
-
-// WithIntegerRepresentation is an interface that can be implemented by a revision to
-// provide an integer representation of the revision.
-type WithIntegerRepresentation interface {
-	datastore.Revision
-	IntegerRepresentation() (int64, uint32)
-}
```
tstirrat15, line 86: I grepped for usages of this interface and its `IntegerRepresentation` method and didn't find anything. I'm assuming it's dead code.

internal/datastore/revisions/hlcrevision.go
```diff
@@ -55,12 +56,14 @@ func parseHLCRevisionString(revisionStr string) (datastore.Revision, error) {
 	}
 
 	paddedLogicalClockStr := pieces[1] + strings.Repeat("0", logicalClockLength-len(pieces[1]))
-	logicalclock, err := strconv.ParseInt(paddedLogicalClockStr, 10, 32)
+	logicalclock, err := strconv.ParseUint(paddedLogicalClockStr, 10, 32)
```
tstirrat15, line 59: This gets us more directly to a uint32.

internal/dispatch/graph/lookupresources2_test.go
```diff
@@ -701,6 +702,10 @@ func TestLookupResources2OverSchemaWithCursors(t *testing.T) {
 					foundChunks := [][]*v1.DispatchLookupResources2Response{}
 					for {
 						stream := dispatch.NewCollectingDispatchStream[*v1.DispatchLookupResources2Response](ctx)
+
+						uintPageSize, err := safecast.ToUint32(pageSize)
+						require.NoError(err)
```
tstirrat15, line 707: I made a bunch of these changes to tests. They're overkill, especially since the values are statically defined in most cases, so I can walk these back out if desired.

pkg/cache/cache_otter.go
```diff
@@ -66,7 +68,13 @@ func (wtc *otterCache[K, V]) Get(key K) (V, bool) {
 
 func (wtc *otterCache[K, V]) Set(key K, value V, cost int64) bool {
 	keyString := key.KeyString()
-	return wtc.cache.Set(keyString, valueAndCost[V]{value, uint32(cost)})
+	uintCost, err := safecast.ToUint32(cost)
+	if err != nil {
+		// We make an assumption that if the cast fails, it's because the value
+		// was too big, so we set to maxint in that case.
+		uintCost = math.MaxUint32
+	}
```
tstirrat15, line 76: Please check this

pkg/cmd/datastore/datastore.go
```diff
@@ -469,6 +470,11 @@ func newCRDBDatastore(ctx context.Context, opts Config) (datastore.Datastore, er
 		return nil, errors.New("read replicas are not supported for the CockroachDB datastore engine")
 	}
 
+	maxRetries, err := safecast.ToUint8(opts.MaxRetries)
+	if err != nil {
+		return nil, errors.New("max-retries could not be cast to uint8")
+	}
```
tstirrat15, line 476: Another branch that we're unlikely to hit, but it could result in very weird behavior if we do.

pkg/development/assertions.go
```diff
@@ -42,14 +43,18 @@ func runAssertions(devContext *DevContext, assertions []blocks.Assertion, expect
 	for _, assertion := range assertions {
 		tpl := tuple.MustFromRelationship[*v1t.ObjectReference, *v1t.SubjectReference, *v1t.ContextualizedCaveat](assertion.Relationship)
 
+		// NOTE: zeroes are fine here to mean "unknown"
```
tstirrat15, line 46: If safecast can't successfully cast, it returns zero. For pointing at line numbers, I think this should be fine (especially since you'd have to do something weird to end up out of bounds here).

pkg/genutil/ensure.go
```diff
@@ -13,16 +17,18 @@ func MustEnsureUInt32(value int) uint32 {
 
 // EnsureUInt32 ensures that the specified value can be represented as a uint32.
 func EnsureUInt32(value int) (uint32, error) {
```
tstirrat15, line 19: These functions are basically doing `safecast` things anyway, so I rewrote them in terms of safecast to silence the errors. We can upstream it later.

tstirrat15 (commented):
See additional comments

internal/datastore/common/changes.go
```diff
@@ -118,7 +120,11 @@ func (ch *Changes[R, K]) adjustByteSize(item sized, delta int) error {
 		return spiceerrors.MustBugf("byte size underflow")
 	}
 
-	if ch.currentByteSize > int64(ch.maxByteSize) {
+	// We checked for underflow above, so the current byte size
+	// should fit in a uint64
+	currentByteSize, _ := safecast.ToUint64(ch.currentByteSize)
```
josephschorr, line 125: return the error even if we think it cannot be hit; its safer than ignoring it

internal/datastore/common/changes.go
```diff
@@ -118,7 +120,11 @@ func (ch *Changes[R, K]) adjustByteSize(item sized, delta int) error {
...
+
+	if currentByteSize > ch.maxByteSize {
```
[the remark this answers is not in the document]
josephschorr: I think we should return the error as a MustBugf even if we know it can be safe; it is "extra" code but it means if anything changes we're guaranteed to catch the error vs having a silent bug appear

internal/datastore/common/gc_test.go
```diff
@@ -68,7 +68,7 @@ func (gc *fakeGC) DeleteBeforeTx(_ context.Context, rev datastore.Revision) (Del
 
 	revInt := rev.(revisions.TransactionIDRevision).TransactionID()
 
-	return gc.deleter.DeleteBeforeTx(int64(revInt))
```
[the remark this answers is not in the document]
josephschorr: Yeah, seems so

internal/datastore/crdb/pool/balancer.go
```diff
@@ -84,7 +86,15 @@ type nodeConnectionBalancer[P balancePoolConn[C], C balanceConn] struct {
...
-	seed := int64(new(maphash.Hash).Sum64())
+	seed := int64(0)
+	for seed == 0 {
+		// Sum64 returns a uint64, and safecast will return 0 if it's not castable,
+		// which will happen about half the time (?). We just keep running it until
+		// we get a seed that fits in the box.
+		// Subtracting math.MaxInt64 should mean that we retain the entire range of
+		// possible values.
+		seed, _ = safecast.ToInt64(new(maphash.Hash).Sum64() - math.MaxInt64)
+	}
```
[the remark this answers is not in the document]
josephschorr: I think we should just make the seed a uint64 then? @ecordell?

internal/datastore/crdb/pool/pool.go
```diff
@@ -141,12 +142,16 @@ func (p *RetryPool) ID() string {
 
 // MaxConns returns the MaxConns configured on the underlying pool
 func (p *RetryPool) MaxConns() uint32 {
-	return uint32(p.pool.Config().MaxConns)
+	// This should be non-negative
+	maxConns, _ := safecast.ToUint32(p.pool.Config().MaxConns)
```
josephschorr, line 146: If we have enough of these examples, we should add our own MustToUint32 method and just have it panic if it fails

internal/datastore/crdb/readwrite.go
```diff
@@ -430,7 +431,11 @@ func (rwt *crdbReadWriteTXN) DeleteRelationships(ctx context.Context, filter *v1
 	}
 
 	rwt.relCountChange -= modified.RowsAffected()
-	if delLimit > 0 && uint64(modified.RowsAffected()) == delLimit {
+	rowsAffected, err := safecast.ToUint64(modified.RowsAffected())
+	if err != nil {
+		return false, spiceerrors.MustBugf("could not cast RowsAffected to uint64")
```
josephschorr, line 436: always wrap the error when using MustBugf

josephschorr: In fact, we should probably add a custom linter to ensure this at some point. Might be worth filing an issue

pkg/cache/cache_otter.go
josephschorr: Log it as well, but this seems fine

pkg/development/assertions.go
josephschorr: We should either log the error or have it panic

internal/datastore/crdb/pool/balancer.go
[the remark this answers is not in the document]
ecordell: `NewSource()` and `Seed()` both take an int64

the requirements for this are super minimal, we really just want "don't do the same exact thing on every replica 100% of the time". it's okay for it to not be unique, it's even okay for it to be 0 sometimes.

can we just divide it by 2 and then cast?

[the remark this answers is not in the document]
tstirrat15: Yeah, we can `// ignore gosec` for this particular case, or we can use the approach I came up with. I'm fine with either.

josephschorr (approved):
LGTM

PR #2064, opened 2024-09-16
add continuous checkpointing to Datastore Features

vroldanbet:
this helps consumer understand the behaviour of checkpoints from the datastore without having to know which actual datastore is it talking to.


josephschorr (approved):
LGTM

PR #2069, opened 2024-09-18
Add support for secondary dispatching on LR2

josephschorr:


internal/dispatch/remote/cluster.go
```diff
@@ -216,6 +216,147 @@ func dispatchRequest[Q requestMessage, S responseMessage](ctx context.Context, c
...
+}
+
+type receiver[S responseMessage] interface {
+	Recv() (S, error)
+	grpc.ClientStream
+}
+
+const secondaryCursorPrefix = "$$secondary:"
+
+func publishClient[Q requestMessageWithCursor, R responseMessageWithCursor](ctx context.Context, client receiver[R], stream dispatch.Stream[R], secondaryDispatchName string) error {
```
ecordell, line 236: is `Q` used anywhere?

internal/dispatch/remote/cluster.go
```diff
@@ -216,6 +216,147 @@ func dispatchRequest[Q requestMessage, S responseMessage](ctx context.Context, c
...
+	if cursor != nil && len(cursor.Sections) > 0 {
+		if strings.HasPrefix(cursor.Sections[0], secondaryCursorPrefix) {
+			cursorLockedSecondaryName = strings.TrimPrefix(cursor.Sections[0], secondaryCursorPrefix)
+			cursor.Sections = cursor.Sections[1:]
+		}
+	}
+
+	// If no secondary dispatches are defined, just invoke directly.
+	if len(cr.secondaryDispatchExprs) == 0 || len(cr.secondaryDispatch) == 0 {
+		return publishClient[Q](withTimeout, client, stream, "")
```
ecordell, line 303: define a constant for the primary dispatcher name?

internal/dispatch/remote/cluster_test.go
```diff
@@ -242,6 +259,228 @@ func TestSecondaryDispatch(t *testing.T) {
...
+					Relation:  "somerelation",
+				},
+				SubjectIds: []string{"foo"},
+				TerminalSubject: &corev1.ObjectAndRelation{
+					Namespace: "foo",
+					ObjectId:  "bar",
+					Relation:  "...",
+				},
+				OptionalCursor: &v1.Cursor{
+					Sections:        []string{secondaryCursorPrefix + "tertiary"},
```
ecordell, line 362: add one extra test case where the expression is `"['secondary']"` and the cursor is also for `secondary`?

internal/dispatch/remote/cluster.go
josephschorr: Its used for the type of the incoming message, off of which I grab the cursor. I could have the req just be that interface directly, but I always prefer well-typed wherever possible

ecordell (approved):

PR #2070, opened 2024-09-18
pkg/proto: adopt CodecV2 and gRPC buffer pooling

jzelinskie:
This replaces our existing gRPC codec (which enabled us to use vtprotobuf) with the new v2 codec interface that can support buffer pooling.

Huge thanks to engineers from Google and LinkedIn for fleshing out this work and upstreaming the necessary changes. There was a great talk from gRPConf I can share once it's uploaded to YouTube.

I did regenerate our go.mod which combined all of the indirect dependencies into one block. If there were purposely split out, I can fix it.

See the sister PR to this introducing the Codec for Vitess: https://github.com/vitessio/vitess/pull/16790


pkg/proto/dispatch/v1/01_codec.go
```diff
@@ -20,38 +18,51 @@ import (
...
+			}
+			return mem.BufferSlice{mem.SliceBuffer(buf)}, nil
+		}
+		pool := mem.DefaultBufferPool()
+		buf := pool.Get(size)
+		if _, err := m.MarshalToSizedBufferVT((*buf)[:size]); err != nil {
+			pool.Put(buf)
+			return nil, err
+		}
+		return mem.BufferSlice{mem.NewBuffer(buf, pool)}, nil
```
tstirrat15, line 48: Curious - what does `NewBuffer` do here? Haven't we already reserved the relevant memory in the `pool.Get` call above?

Alternatively, is there some reading about how this approach works generally?

tstirrat15 (commented):
See comment; I'll let someone more familiar with the internals give an approval.

pkg/proto/dispatch/v1/01_codec.go
jzelinskie: By and large, this code is inspired by the upstream default codec: https://github.com/grpc/grpc-go/blob/v1.66.2/encoding/proto/proto.go

`mem.NewBuffer` is less about allocating bytes and more about tracking the lifetime and ownership of the buffer: https://github.com/grpc/grpc-go/blob/v1.66.2/mem/buffers.go#L83-L102

vmg:
@jzelinskie We have quite a bit of work left before the Vitess PR is green... I would really like to see how these changes affect spicedb benchmarks. Do you have a plan to benchmark this branch? Cheers!

pkg/proto/dispatch/v1/01_codec.go
```diff
@@ -20,38 +18,52 @@ import (
...
 	if m, ok := v.(vtprotoMessage); ok {
-		return m.MarshalVT()
-	}
-
-	if m, ok := v.(proto.Message); ok {
-		return proto.Marshal(m)
+		size := m.SizeVT()
+		if mem.IsBelowBufferPoolingThreshold(size) {
+			buf := make([]byte, size)
+			n, err := m.MarshalToSizedBufferVT(buf[:size])
```
josephschorr, line 37: if we're allocating the buffer as size of `size`, why do we need to slice here and below?

jzelinskie: I cleaned this up a bit to try and make it more clear (and after checking guarantees from the pool implementations).
I'm being defensive for the scenario where the number of bytes written are not actually the exact size of the buffer, which looks possible based on the generated code.

jzelinskie:
> @jzelinskie We have quite a bit of work left before the Vitess PR is green... I would really like to see how these changes affect spicedb benchmarks. Do you have a plan to benchmark this branch? Cheers!

We're getting out 2 stable releases of SpiceDB and then looking to begin testing this in some different staging environments to see how it goes.

josephschorr (approved):
LGTM

PR #2079, opened 2024-09-25
Enable LRv2 by default and update the steelthread tests

josephschorr:
The steelthread tests have to be updated because the "indirect" tests now return results in a slightly different order in LRv2. Manually verified that the results themselves are the same (except duplicates, which are allowed)


tstirrat15 (approved):
LGTM

PR #2084, opened 2024-10-01
Add API support for transaction metadata on WriteRels and DeleteRels

josephschorr:


internal/services/v1/relationships.go
```diff
@@ -42,6 +43,8 @@ var writeUpdateCounter = promauto.NewHistogramVec(prometheus.HistogramOpts{
 	Buckets:   []float64{0, 1, 2, 5, 10, 15, 25, 50, 100, 250, 500, 1000},
 }, []string{"kind"})
 
+const MaximumTransactionMetadataSize = 65000 // bytes. Limited by the BLOB size used in MySQL driver
```
tstirrat15, line 46: Is it worth making this datastore-specific? Or will we burn that bridge when we get to it?

internal/services/v1/relationships.go
```diff
@@ -367,7 +374,28 @@ func (ps *permissionServer) WriteRelationships(ctx context.Context, req *v1.Writ
...
+		return nil
+	}
+
+	b, err := metadata.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	if len(b) > MaximumTransactionMetadataSize {
+		return NewTransactionMetadataTooLargeErr(len(b), MaximumTransactionMetadataSize)
```
tstirrat15, line 388: Are we guaranteed that the client encoding will match the JSON encoding? i.e. will the size that that the client sees on their machine match what we're validating against here?

tstirrat15 (dismissed):
See comments, otherwise LGTM

internal/services/v1/relationships.go
josephschorr: No; this is a rough estimate but since the drivers store JSON too, it should be sufficient

tstirrat15 (dismissed):

tstirrat15 (approved):
Reapproving after rebas

PR #2085, opened 2024-10-02
reduce build times in CI

vroldanbet:
optimized tests with various changes. 40% improvement in CI in some cases.

- datastore tests now run parallel. CRDB and Postgres had issues so it's disabled parallelization there
- Parallelized integrity / no integrity tests in CRDB
- removed code coverage by default in unit tests - was not being used
- added `t.Parallel()` in the unit tests taking the longest

# before
image

# after
image

### DIFF
--- a/internal/dispatch/graph/check_test.go
+++ b/internal/dispatch/graph/check_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"go.uber.org/goleak"
 
 	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/internal/datastore/memdb"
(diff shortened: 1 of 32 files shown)


Verolop (approved):
🚀

internal/dispatch/graph/check_test.go
```diff
@@ -30,7 +29,7 @@ import (
 var ONR = tuple.ObjectAndRelation
 
 func TestSimpleCheck(t *testing.T) {
-	defer goleak.VerifyNone(t, append(testutil.GoLeakIgnores(), goleak.IgnoreCurrent())...)
```
tstirrat15, line 33: Was this no longer necessary?

tstirrat15 (commented):
Curious about goleak change

internal/dispatch/graph/check_test.go
vroldanbet: no - it's moved into a `TestMain` function which makes it execute for all tests in the package

PR #2093, opened 2024-10-16
metrics: spicedb_environment_info from telemetry

### LINKED ISSUE #2086: Introduce spicedb_info metric
### Problem Statement

Ops folks need a better way to aggregate basic information about the nodes in their SpiceDB deployments.

### Solution Brainstorm

This metric actually exists already: [`spicedb_telemetry_info`](https://github.com/authzed/spicedb/blob/main/TELEMETRY.md#spicedb_telemetry_info-gauge).

SpiceDB should be updated to export this metric under the normal `/metrics` endpoint, too.

jzelinskie:
This change reuses the logic for `spicedb_telemetry_info` for a metric registered to the default Prometheus registry.

Fixes #2086

### DIFF
--- a/internal/telemetry/metrics.go
+++ b/internal/telemetry/metrics.go
@@ -17,16 +17,10 @@ import (
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/internal/middleware/usagemetrics"
 	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/promutil"
 )
 
-// RegisterTelemetryCollector registers a collector for the various pieces of
-// data required by SpiceDB telemetry.
-func RegisterTelemetryCollector(datastoreEngine string, ds datastore.Datastore) (*prometheus.Registry, error) {
-	registry := prometheus.NewRegistry()
-
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
-
+func SpiceDBClusterInfoCollector(ctx context.Context, subsystem, dsEngine string, ds datastore.Datastore) (promutil.CollectorFunc, error) {
 	nodeID, err := os.Hostname()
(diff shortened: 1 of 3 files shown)


pkg/cmd/server/server.go
```diff
@@ -459,6 +459,15 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 	closeables.AddCloser(gatewayCloser)
 	closeables.AddWithoutError(gatewayServer.Close)
 
+	infoCollector, err := telemetry.SpiceDBClusterInfoCollector(context.TODO(), "environment", c.DatastoreConfig.Engine, ds)
```
vroldanbet, line 462: please use the context
```suggestion
	infoCollector, err := telemetry.SpiceDBClusterInfoCollector(ctx, "environment", c.DatastoreConfig.Engine, ds)
```

internal/telemetry/metrics.go
```diff
@@ -57,9 +50,42 @@ func RegisterTelemetryCollector(datastoreEngine string, ds datastore.Datastore)
 				"arch":       runtime.GOARCH,
 				"go":         buildInfo.GoVersion,
```
vroldanbet, line 51: we are already providing go version through the standard Go extended runtime metrics. It's available through `go_info`

jzelinskie: Yeah, this is an artifact of reusing the same metric across telemetry and non-telemetry usage. Should I make it optional or just keep it?

pkg/cmd/server/server.go
jzelinskie: thanks for this, i did some refactoring

internal/telemetry/metrics.go
vroldanbet: it's ok, I don't think it hurts, just wanted to call it out

vroldanbet (approved):

Issue #2094, opened 2024-10-19
Aggressive postgres instantiation ctx timeout causes intermittent failures in Datastore initialisation


anoopsundial:
### What platforms are affected?

linux

### What architectures are affected?

others

### What SpiceDB version are you using?

v1.31.0

### Steps to Reproduce

1. Start Spice DB locally with a remote cloud postgres which can take more than 5 second to respond.
2. Before the initialisation steps can complete the ctx times out very frequently eventhough intermittently[](url)

### Expected Result

The initialisation should be able to work with slow or remote postgres instances.
It should either be made configurable. Or increased to higher number like 20/30 . As this is the initialisation flow.

The exact line were this is set is below:
https://github.com/authzed/spicedb/blob/4262bfd23266888fdfff5e01a24f76e6d30c98ef/internal/datastore/postgres/postgres.go#L227

### Actual Result

The context times out very frequently . An example log below

`2024-10-18 15:53:02 {"level":"trace","preshared-keys-count":1,"time":"2024-10-18T10:23:02Z","message":"using gRPC auth with preshared key(s)"}
2024-10-18 15:53:02 {"level":"trace","preshared-key-1-length":13,"time":"2024-10-18T10:23:02Z","message":"preshared key configured"}
2024-10-18 15:53:02 {"level":"info","time":"2024-10-18T10:23:02Z","message":"using postgres datastore engine"}
2024-10-18 15:53:02 {"level":"warn","details-url":"https://spicedb.dev/d/force-custom-plan","plan_cache_mode":"force_custom_plan","time":"2024-10-18T10:23:02Z","message":"defaulting value in Postgres DB URI"}
2024-10-18 15:53:07 {"level":"debug","pgx":{"args":[],"err":"timeout: context deadline exceeded","pid":23643,"sql":"SHOW track_commit_timestamp;","time":645.80925},"time":"2024-10-18T10:23:07Z","message":"Query"}
2024-10-18 15:53:07 {"level":"error","error":"failed to create datastore: timeout: context deadline exceeded","time":"2024-10-18T10:23:07Z","message":"terminated with errors"}`

anoopsundial:
Have added a PR to address the issue. Please do consider. If it needs to be fixed differently please do let me know.
https://github.com/authzed/spicedb/pull/2095

anoopsundial:
Thanks for reviewing and merging the fix. 

[...]

PR #2095, opened 2024-10-19
Increase the aggressive postgres instantiation timeout of 5 seconds to 30 seconds #2094

anoopsundial:
The instantiation ctx timeout for postgres datastore is aggressive a 5 seconds.

This causes intermittent initialisation failures in environments where the postgres is slow / remote etc. Since this is an instantiation flow, its ok to increase the timeout to a higher number like 30 seconds. This can avoid such failures.

### DIFF
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -224,7 +224,7 @@ func newPostgresDatastore(
 			Msg("postgres configured to use intermediate migration phase")
 	}
 
-	initializationContext, cancelInit := context.WithTimeout(context.Background(), 5*time.Second)
+	initializationContext, cancelInit := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancelInit()
 
 	readPool, err := pgxpool.NewWithConfig(initializationContext, readPoolConfig)


anoopsundial:

anoopsundial:
recheck

anoopsundial:
Would be great if someone can take a quick look at this PR as its just a 1 line change.

anoopsundial:
> @anoopsundial I'm ok with the change, but if it's taking that long to connect to PG, I suspect the SpiceDB->PG latency is going to be so large that it will lead to terrible performance overall.
> 
> Why do you think is it taking that long for your SpiceDB to reach PG?

Thanks for looking into this @vroldanbet 

This is specifically for Local Dev Testing. 
We create a local spicedb instance and connect it to our staging environment in cloud to create a backup. 
This is the primary reason it's taking longer for response.
And we primarily use this mode to create a backup and restore it to a local spicedb instance for local testing.
In the proper production use-case we do not face this issue.

vroldanbet:
Understood. I'm not sure if you are familiar with https://github.com/authzed/zed, but it comes with a backup, redaction, and restore functionality so that you can hit your staging SpiceDB directly. I can also think of doing a `pg_dump` and restoring to a local PG instance. To make a `pg_dump` work, run `spicedb datastore repair` after restoring the backup into your local PG.

vroldanbet (approved):

PR #2097, opened 2024-10-23
Adjust CRDB watch to set the checkpoint frequency equal to the resolved

josephschorr:
Should reduce watch pressure from the caller to CRDB


josephschorr:
Closing; will reopen if need be in the future

PR #2099, opened 2024-10-23
Remove LookupResources v1 implementation

tstirrat15:
## Description
Now that v2 is on by default, this cleans up the old v1 implementation. It should also reduce our test times by getting rid of a branch of the consistency tests.

## Changes
Will annotate.

## Testing
See that everything goes green.


internal/services/integrationtesting/consistency_test.go
```diff
@@ -60,15 +60,10 @@ func TestConsistency(t *testing.T) {
...
-						useLRV2 := useLRV2
-						t.Run(fmt.Sprintf("lrv2-%t", useLRV2), func(t *testing.T) {
-							for _, chunkSize := range []uint16{5, 10} {
-								t.Run(fmt.Sprintf("lrv2-%t", useLRV2), func(t *testing.T) {
-									t.Parallel()
-									runConsistencyTestSuiteForFile(t, filePath, dispatcherKind == "caching", chunkSize, useLRV2)
-								})
-							}
+					for _, chunkSize := range []uint16{5, 10} {
+						t.Run(fmt.Sprintf("chunk-size-%t", chunkSize), func(t *testing.T) {
```
tstirrat15, line 64: I think this is what should be reported here but I'm not 100% sure.

internal/services/v1/permissions.go
```diff
@@ -393,158 +393,7 @@ func TranslateExpansionTree(node *core.RelationTupleTreeNode) *v1.PermissionRela
...
-	if req.OptionalCursor != nil {
-		_, ok, err := cursor.GetCursorFlag(req.OptionalCursor, lrv2CursorFlag)
-		if err != nil {
-			return ps.rewriteError(resp.Context(), err)
-		}
-
-		if ok {
-			return ps.lookupResources2(req, resp)
-		}
-	}
```
tstirrat15, line 410: I figured this was no longer necessary since the flag itself isn't really necessary anymore.

internal/services/v1/permissions_test.go
```diff
@@ -641,7 +641,6 @@ func TestLookupResources(t *testing.T) {
 									MaxPreconditionsCount:           1000,
 									StreamingAPITimeout:             30 * time.Second,
 									MaxRelationshipContextSize:      25000,
-									UseExperimentalLookupResources2: useV2,
```
tstirrat15, line 644: There's a bunch of changes like this that are just propagating the option removal.

pkg/proto/core/v1/util.go
tstirrat15, line 1: I ran `deadcode` against the project and this file was called out as being entirely unused. I'm assuming that means it can be deleted.

tstirrat15 (commented):
See comments

internal/services/v1/permissions.go
```diff
@@ -393,158 +393,7 @@ func TranslateExpansionTree(node *core.RelationTupleTreeNode) *v1.PermissionRela
 	}
 }
 
-const lrv2CursorFlag = "lrv2"
```
josephschorr, line 396: keep the cursor flag; we might need it again in the future and we don't want to break existing cursors. we should just return an error if the cursor does not have it

pkg/cmd/serve.go
```diff
@@ -158,7 +158,6 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 
 	experimentalFlags := nfs.FlagSet(BoldBlue("Experimental"))
 	// Flags for experimental features
-	experimentalFlags.BoolVar(&config.EnableExperimentalLookupResources, "enable-experimental-lookup-resources", true, "enables the experimental version of the lookup resources API")
```
josephschorr, line 161: don't remove, as it will break callers. Instead, deprecate it and if someone gives a value of `false`, issue a warning that the flag will be removed in the future

internal/services/v1/permissions.go
josephschorr: I don't see this change reverted

josephschorr (approved):
LGTM

PR #2100, opened 2024-10-24
fixes CRDB watch implementation accumulating past updates

vroldanbet:
when the timestamp to start emitting updates from
is way in the past, CRDB will not emit checkpoints.

As a consequence, every update in the w.r.t to the moment the changefeed was created will accumulate
in memory and OOM the process with a large enough
backlog.

The proposed solution is to compute one checkpoint from the real-time stream of updates, and then use that as the high-watermark for the backlog of
changes from the past. We need to emit all the updates before we can emit a checkpoint, so downstream callers would have to handle it accordingly.


vroldanbet:
superseded by https://github.com/authzed/spicedb/pull/2120

PR #2101, opened 2024-10-24
Copy old schema package into new package

tstirrat15:
## Description
The first part of implementing composable schemas. We decided that we want to create a new package and build on top of it while development is in flight. Doing the copying as a separate PR means that the first PR where meaningful changes happen will have a meaningful diff.

## Changes
* Copy old schema package into new package
* Update import references to point at new package

## Testing
See that things are green


ecordell (approved):
I assume this happened after a discussion with @josephschorr that this is the route we want to take?

Personally I'm not convinced that this is the best plan; we know we have some schema feature work coming up and we'll have to make them in two places now. 

I won't block it if you have consensus on this approach though.

josephschorr:
@ecordell Yes, this is the preferred approach. While we'll have to copy the changes into both packages, they will diverge fairly soon and until we have an IR for schema, we can't break back-compat with the existing schema parser. Adding conditionals into the existing parser will be quite ugly too, so I figured this was a better solution, even if it does require a bit more work upfront

PR #2102, opened 2024-10-25
Enable native histograms in server latency metrics

bison:
Enables [Native Histograms](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) for gRPC server metrics.

As far as I know, there's not really a reason to not enable this. Clients that don't support it will still get the same format they always have. We should do some more thorough research and confirm that though.


alecmerdler (dismissed):
Nice!

vroldanbet (approved):
Please note that this would enable native histograms only for the gRPC prometheus middleware, which I think is great but ideally we enable it for each histogram in the application

Issue #1437, opened 2023-07-13
Support for multi-test validation of schemas


josephschorr:
## Background

Schema in SpiceDB is written as a single unit via the `WriteSchema` call. This is an important requirement for type checking and validation, but comes with a cost of making it harder for different teams using the same unified permission system to manage changes to their own portions of a schema.

Furthermore, the current file format passed to tooling such as Playground, `zed validate` and the [https://github.com/authzed/action-spicedb-validate](<https://github.com/authzed/action-spicedb-validate>) GitHub Action only supports a single set of test relationships and associated assertions/expected relationships, making the testing of different scenarios much harder.

## Proposed Solution

Define a new modified file format for `.zed` files that allows for the definition of both schema data (`caveat` and `definition`), as well as a set of validation tests, each with their own relationships, assertions and expected relationships. This new file format will also support `import`-ing of other `.zed` files via a new `import` keyword.

To create the "final" schema for `WriteSchema`, a new `zed build` command will be supplied that performs import resolution, type checks the combined schema, runs the validation tests (if any) and, finally, returns or saves the combined schema to be sent to SpiceDB

> **Note**<br>The semantics of the `WriteSchema` call *will not change*; imports will *not* be supported via that call, under this proposal.

### Proposed File Format Example

```zed
// Import statement allows for importing of object type definitions and caveats from other .zed files locally or via git
from .somelocalzedfile import sometype
from "../../someotherfile" import someothertype
from github.com/mycomp/myauthzteam@v2 import user

// Schema can reference an imported type
definition resource { 
  relation viewer: user
}

/**
 * Tests if the document can be viewed directly, or via being a
 * member of the group.
 */
test SomeSortOfThing {
  relationships {
    document:somedoc viewer user:someuser
    team:someteam    member user:anotheruser with `caveat expression goes here`
  }
  assertions {
    document:somedoc     view user:someuser     // Positive assertion
    document:anotherdoc !view user:someuser     // Negative assertion
    document:anotherdoc  view user:someuser when {
      now: “2022-12-21 12:34:00+0000”
    }
  }
  expected {
    document:firstdoc view (
      user:tom, user:fred are viewer of document:firstdoc
      user:sarah                is  member of team:someteam
    ) when {
        now: “2022-12-21 12:34:00+0000”
    }
  }
}
```

#### zed build

[...]

PR #2103, opened 2024-10-25
Implement parsing of local imports

tstirrat15:
Part of #1437 

## Description
A first step in implementing the composable schema proposal. This implements the relative import syntax in the parser.

## Changes
* Add a new test with the desired local import syntax
* Add `from` and `import` as keywords
* Add new import node type and predicates for it
* Add parser logic to consume import syntax
* Add generated test output
* Fix trivy usage so that we don't hit rate limits
## Testing
Review. See that tests are green.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -72,7 +72,7 @@ jobs:
         with:
           distribution: "goreleaser-pro"
           version: "2.3.2"
-          args: "release --clean --split --snapshot --single-target --skip=chocolatey"
+          args: "release --clean --split --snapshot --single-target"
         env:
           GORELEASER_KEY: "${{ secrets.GORELEASER_KEY }}"
       - name: "Obtain container image to scan"
--- a/pkg/composableschemadsl/parser/parser.go
+++ b/pkg/composableschemadsl/parser/parser.go
@@ -62,6 +62,9 @@ Loop:
 		case p.isKeyword("caveat"):
 			rootNode.Connect(dslshape.NodePredicateChild, p.consumeCaveat())
 
+		case p.isKeyword("from"):
+			rootNode.Connect(dslshape.NodePredicateChild, p.consumeImport())
+
 		default:
 			p.emitErrorf("Unexpected token at root level: %v", p.currentToken.Kind)
 			break Loop
@@ -574,6 +577,20 @@ func (p *sourceParser) tryConsumeIdentifierLiteral() (AstNode, bool) {
 	return identNode, true
 }
 
+// consumeIdentifierLiteral is similar to the above, but attempts and errors
+// rather than checking the token type beforehand
+func (p *sourceParser) consumeIdentifierLiteral() (AstNode, bool) {
+	identNode := p.startNode(dslshape.NodeTypeIdentifier)
+	defer p.mustFinishNode()
+
+	identifier, ok := p.consumeIdentifier()
+	if !ok {
+		return identNode, false
+	}
+	identNode.MustDecorate(dslshape.NodeIdentiferPredicateValue, identifier)
+	return identNode, true
+}
+
 func (p *sourceParser) tryConsumeNilExpression() (AstNode, bool) {
 	if !p.isKeyword("nil") {
 		return nil, false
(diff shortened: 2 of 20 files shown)


.github/workflows/security.yaml
```diff
@@ -72,10 +72,19 @@ jobs:
         with:
           distribution: "goreleaser-pro"
           version: "2.3.2"
-          args: "release --clean --split --snapshot --single-target --skip=chocolatey"
```
tstirrat15, line 75: There's no chocolatey so there's no reason for it

.github/workflows/security.yaml
```diff
@@ -72,10 +72,19 @@ jobs:
...
         env:
           GORELEASER_KEY: "${{ secrets.GORELEASER_KEY }}"
       - name: "Obtain container image to scan"
         run: 'echo "IMAGE_VERSION=$(jq .version dist/linux_amd64_v1/metadata.json --raw-output)" >> $GITHUB_ENV'
-      - name: "run trivy on release image"
-        run: "docker run -v /var/run/docker.sock:/var/run/docker.sock aquasec/trivy image --format table --exit-code 1 --ignore-unfixed --vuln-type os,library --no-progress --severity CRITICAL,HIGH,MEDIUM authzed/spicedb:v${{ env.IMAGE_VERSION }}-amd64"
+      - uses: "aquasecurity/trivy-action@master"
+        name: "run trivy on release image"
+        with:
+          scan-type: "image"
```
tstirrat15, line 83: Use the action instead of the hand-rolled command

.github/workflows/security.yaml
josephschorr: Where did it go?

pkg/composableschemadsl/parser/parser.go
```diff
@@ -485,6 +488,7 @@ func (p *sourceParser) tryConsumeArrowExpression() (AstNode, bool) {
 				return nil, false
 			}
 
+			// TODO: is this the time to do this?
```
josephschorr, line 491: yep!

pkg/composableschemadsl/parser/parser.go
```diff
@@ -584,3 +588,50 @@ func (p *sourceParser) tryConsumeNilExpression() (AstNode, bool) {
...
+	// from ...
+	p.consumeKeyword("from")
+
+	// Consume alternating periods and identifiers
+	for {
+		if _, ok := p.consume(lexer.TokenTypePeriod); !ok {
+			return importNode
+		}
+
+		segmentNode, ok := p.tryConsumeIdentifierLiteral()
```
josephschorr, line 605: This isn't a try consume; it should be a consume, because a missing identifier here should be an error. Please add a test case for this

pkg/composableschemadsl/parser/parser.go
```diff
@@ -584,3 +588,50 @@ func (p *sourceParser) tryConsumeNilExpression() (AstNode, bool) {
...
+		importNode.Connect(dslshape.NodeImportPredicatePathSegment, segmentNode)
+
+		if !p.isToken(lexer.TokenTypePeriod) {
+			// If we don't have a period as our next token, we move
+			// to the next step of parsing.
+			break
+		}
+	}
+
+	p.consumeKeyword("import")
```
josephschorr, line 618: Return here if the keyword was not found

pkg/composableschemadsl/parser/parser.go
```diff
@@ -584,3 +588,50 @@ func (p *sourceParser) tryConsumeNilExpression() (AstNode, bool) {
...
+
+	// Consume alternating identifiers and commas until we reach the end of the import statement
+	for {
+		definitionNode, ok := p.tryConsumeIdentifierLiteral()
```
josephschorr, line 622: Similarly, not a try consume because it is required

pkg/composableschemadsl/parser/parser_test.go
```diff
@@ -122,6 +122,7 @@ func TestParser(t *testing.T) {
 		{"arrow illegal operations test", "arrowillegalops"},
 		{"arrow illegal function test", "arrowillegalfunc"},
 		{"caveat with keyword parameter test", "caveatwithkeywordparam"},
+		{"local imports test", "localimport"},
```
josephschorr, line 125: add a bunch of error cases as well

.github/workflows/security.yaml
tstirrat15: It's in `.goreleaser.windows.yaml`, which isn't used by this command.

.github/workflows/security.yaml
tstirrat15: I'll walk this back out with the changes to the DBs that @josephschorr put through.

pkg/composableschemadsl/parser/parser.go
tstirrat15: I was going off of the implementation of this function, which does seem to use `consume`: https://github.com/authzed/spicedb/blob/c8e739ce370133d6e99ea20651f98f9f39160eb7/pkg/composableschemadsl/parser/parser.go#L576

I'll add a test case, though.

pkg/composableschemadsl/parser/tests/localimport_keyword_in_identifiers.zed.expected
```diff
@@ -0,0 +1,121 @@
...
+      end-rune = 30
+      input-source = local imports with keyword in identifiers test
+      start-rune = 0
+      child-node =>
+        NodeTypeError
+          end-rune = 30
+          error-message = Expected one of: [TokenTypeComma], found: TokenTypeKeyword
+          error-source = caveat
+          input-source = local imports with keyword in identifiers test
+          start-rune = 32
```
tstirrat15, line 16: Why doesn't this line up with the error below?

pkg/composableschemadsl/parser/tests/localimport_malformed_identifier_set.zed.expected
```diff
@@ -0,0 +1,109 @@
...
+      end-rune = 30
+      input-source = local imports with malformed identifiers set test
+      start-rune = 0
+      child-node =>
+        NodeTypeError
+          end-rune = 30
+          error-message = Expected one of: [TokenTypeComma], found: TokenTypeKeyword
+          error-source = definition
+          input-source = local imports with malformed identifiers set test
+          start-rune = 33
```
tstirrat15, line 16: Same question here. Is this expected?

tstirrat15 (commented):
See questions

pkg/composableschemadsl/parser/parser.go
tstirrat15: I'm gonna do that in a follow-on PR

tstirrat15, line 491:
```suggestion
```

pkg/composableschemadsl/parser/tests/localimport_keyword_in_identifiers.zed.expected
josephschorr: This typically is a result of missing a finishNode call somewhere, but I don't see it. You'll need to trace to see why that is

josephschorr (approved):
LGTM

PR #2105, opened 2024-10-28
Make all and any into keywords

tstirrat15:
## Description
One of the things that we wanted to do with the new schema version. `any` and `all` should be keywords befitting their use in intersection arrow syntax, but we didn't want to break existing usages; we're making the breaking change now.

## Changes
* Add `any` and `all` to keyword list
* Add `consumeKeywords` impl function to capture getting one of a set of keywords
* Update caveat parsing code to allow for `any` as a keyword in type logic
* Update test case to reflect new error messaging
* Update lexer tests and make keyword tests exhaustive
## Testing


josephschorr (approved):
LGTM

PR #2107, opened 2024-10-30
Updates to tuple helpers

tstirrat15:
## Description
Came up the context of authzed/zed#431. The lack of context in the string helper isn't clear from the name, and having a `MustParseV1` function makes writing tests in zed and elsewhere easier.

## Changes
* Rename fn
* Update usage
* Add new helper for tests

## Testing
Review.


josephschorr (approved):
LGTM

PR #2116, opened 2024-11-04
Update compiler to consume local imports

tstirrat15:
## Description
Part of getting composable schemas implemented. This actually implements the traversal.

## Notes
This is intentionally incremental, and is not intended to be the release candidate. Future work includes:

* Import cycle tracking
* Only including requested imports (rather than all imports)

## Changes
* Plumb an `existingNames` value through the compiler so that it can track the definitions it's already seen
* Add file importer logic
* Implement import logic the translator
## Testing
Review. See that tests go green.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,6 +1,6 @@
 ---
 name: "Lint"
-on: # yamllint disable-line rule:truthy
+on:  # yamllint disable-line rule:truthy
   push:
     branches:
       - "!dependabot/*"
@@ -20,7 +20,7 @@ jobs:
       - name: "Check Licenses"
         uses: "authzed/actions/go-license-check@main"
         with:
-          ignore: "buf.build" # Has no license information
+          ignore: "buf.build"  # Has no license information
 
   go-lint:
     name: "Lint Go"
--- a/pkg/composableschemadsl/compiler/compiler.go
+++ b/pkg/composableschemadsl/compiler/compiler.go
@@ -9,6 +9,7 @@ import (
 	"github.com/authzed/spicedb/pkg/composableschemadsl/dslshape"
 	"github.com/authzed/spicedb/pkg/composableschemadsl/input"
 	"github.com/authzed/spicedb/pkg/composableschemadsl/parser"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 )
 
@@ -52,20 +53,33 @@ func (cs CompiledSchema) SourcePositionToRunePosition(source input.Source, posit
 type config struct {
 	skipValidation   bool
 	objectTypePrefix *string
+	// In an import context, this is the folder containing
+	// the importing schema (as opposed to imported schemas)
+	sourceFolder string
 }
 
 func SkipValidation() Option { return func(cfg *config) { cfg.skipValidation = true } }
 
+// Config for the prefix attached to all definitions, such as in Serverless
+// where it's `someorganization/` in front of each definition.
 func ObjectTypePrefix(prefix string) ObjectPrefixOption {
 	return func(cfg *config) { cfg.objectTypePrefix = &prefix }
 }
 
+// Config that does not supply the prefix but requires the prefix on all objects.
 func RequirePrefixedObjectType() ObjectPrefixOption {
 	return func(cfg *config) { cfg.objectTypePrefix = nil }
 }
 
+// Config that allows for no prefix. This is the "normal" default.
 func AllowUnprefixedObjectType() ObjectPrefixOption {
 	return func(cfg *config) { cfg.objectTypePrefix = new(string) }
+}
+
+// Config that supplies the root source folder for compilation. Required
+// for relative import syntax to work properly.
+func SourceFolder(sourceFolder string) Option {
+	return func(cfg *config) { cfg.sourceFolder = sourceFolder }
 }
 
 type Option func(*config)
@@ -74,6 +88,11 @@ type ObjectPrefixOption func(*config)
 
 // Compile compilers the input schema into a set of namespace definition protos.
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
+	names := mapz.NewSet[string]()
+	return compileImpl(schema, names, prefix, opts...)
+}
+
+func compileImpl(schema InputSchema, existingNames *mapz.Set[string], prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
 	cfg := &config{}
 	prefix(cfg) // required option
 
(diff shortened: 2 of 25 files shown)


pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -52,6 +52,10 @@ func (cs CompiledSchema) SourcePositionToRunePosition(source input.Source, posit
 type config struct {
 	skipValidation   bool
 	objectTypePrefix *string
+	existingNames    []string
+	// In an import context, this is the folder containing
+	// the importing schema (as opposed to imported schemas)
+	sourceFolder string
```
tstirrat15, line 58: I don't really like how I need to plumb this through three levels in order to get it to the next compilation context. I'd welcome recommendations on a different way to structure this.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -22,6 +23,8 @@ type translationContext struct {
 	mapper           input.PositionMapper
 	schemaString     string
 	skipValidate     bool
+	existingNames    []string
```
tstirrat15, line 26: Should this be a `*mapz.Set` up and down the stack?

tstirrat15 (commented):
A couple of thoughts

pkg/composableschemadsl/compiler/translator.go
josephschorr: Likely

pkg/composableschemadsl/compiler/compiler.go
josephschorr: Why can't you put it on the translationContext?

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,51 @@
...
+
+	var schemaBytes []byte
+	schemaBytes, err := os.ReadFile(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read schema file: %w", err)
+	}
+	// TODO: do we want this sort of logging here?
+	log.Trace().Str("schema", string(schemaBytes)).Str("file", filepath).Msg("read schema from file")
+
+	compiled, err := Compile(InputSchema{
```
josephschorr, line 34: If you call an internal compile, then this could be passed the translationContext directly and then you wouldn't need a different context here or the need to pass it into Compile

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,54 @@
...
+	pathSegments []string
+	sourceFolder string
+	names        *mapz.Set[string]
+}
+
+const SchemaFileSuffix = ".zed"
+
+func ImportFile(importContext *ImportContext) (*CompiledSchema, error) {
+	relativeFilepath := constructFilePath(importContext.pathSegments)
+	filePath := path.Join(importContext.sourceFolder, relativeFilepath)
```
vroldanbet, line 25: what happens if the imports are outside of the root context folder? is that supported, and if not, what kind of opinions/guidelines does it impose on the structure of a composable schema?

josephschorr: No; the eventual goal will be to support this using git-based imports

.github/workflows/lint.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: "Lint"
-on: # yamllint disable-line rule:truthy
+on:  # yamllint disable-line rule:truthy
```
josephschorr, line 3: Why this change and below?

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,54 @@
...
+
+type ImportContext struct {
+	pathSegments []string
+	sourceFolder string
+	names        *mapz.Set[string]
+}
+
+const SchemaFileSuffix = ".zed"
+
+func ImportFile(importContext *ImportContext) (*CompiledSchema, error) {
```
josephschorr, line 23: Unless you see a need to export this, should we just keep this internal (i.e. lowercase)?

tstirrat15: That works for me

.github/workflows/lint.yaml
tstirrat15: `yamllint` wants two spaces before an inline comment and appeared to be failing the lint check as a result.

pkg/composableschemadsl/compiler/importer.go
tstirrat15: The biggest effect that I see is that you won't be able to do a relative import *within* the package that walks up a directory. this means that you couldn't have:

```
app1/app1.zed
app2/app2.zed
util.zed
```

and have both app files reference the util file. I have a feeling this will be a tripping point; it'll either have to be well-documented or something we change.

.github/workflows/lint.yaml
josephschorr: Blech

pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -68,12 +72,21 @@ func AllowUnprefixedObjectType() ObjectPrefixOption {
 	return func(cfg *config) { cfg.objectTypePrefix = new(string) }
 }
 
+func SourceFolder(sourceFolder string) Option {
```
josephschorr, line 75: Add a doc comment explaining what this does. Probably good to add doc comments to the other exported functions too

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,54 @@
...
+	"path"
+	"path/filepath"
+
+	"github.com/rs/zerolog/log"
+
+	"github.com/authzed/spicedb/pkg/composableschemadsl/input"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
+)
+
+type ImportContext struct {
```
josephschorr, line 15: make this context unexported as well

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,54 @@
...
+	pathSegments []string
+	sourceFolder string
+	names        *mapz.Set[string]
+}
+
+const SchemaFileSuffix = ".zed"
+
+func importFile(importContext *ImportContext) (*CompiledSchema, error) {
```
josephschorr, line 23: why does this take in a `*` vs just the importContext struct itself?

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,54 @@
...
+
+	newSourceFolder := filepath.Dir(filePath)
+
+	var schemaBytes []byte
+	schemaBytes, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read schema file: %w", err)
+	}
+	// TODO: do we want this sort of logging here?
+	log.Trace().Str("schema", string(schemaBytes)).Str("file", filePath).Msg("read schema from file")
```
josephschorr, line 35: yes; trace is good

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -0,0 +1,54 @@
...
+
+	compiled, err := compileImpl(InputSchema{
+		// TODO: should this point to the schema file? What is this for?
```
josephschorr, line 38: yes; it gives the file path for error messaging

pkg/composableschemadsl/compiler/importer.go
tstirrat15: Mostly me not having built an intuition yet for when you would use one vs the other.

josephschorr (approved):
LGTM

PR #2117, opened 2024-11-04
Fix MySQL test breakage caused by daylight savings change

josephschorr:
Seriously, it broke due to the DST change because it just *happen* to match the -5 hours window used for the test.

The fix is to move the test that changes the instance's timezone to its own, isolated, MySQL so it doesn't interfere with the other tests


tstirrat15 (approved):
LGTM

alecmerdler (approved):
amazing

PR #2120, opened 2024-11-05
introduce emission strategy into CockroachDB Watch API

vroldanbet:
depends on https://github.com/authzed/spicedb/pull/2122
introduces `WatchOptions.EmissionStrategy` to handle limitations of CockroachDB changefeeds.

In CRDB, checkpointing for a span only occurs once its catch-up scan completes. As a result, resolved timestamps for the changefeed won't be emitted until all spans have been checkpointed and moved past the initial high-water mark.
Unfortunately, no workarounds exist to make the changefeed emit resolved timestamps during a catch-up scan. This is a known issue on CRL's radar.

This is important because, without checkpoints, it means changes are accumulated in memory, and depending on the use case can lead to increased memory pressure and out-of-memory errors

This commit introduces a new change emission strategy option into WatchOptions:
- `EmitWhenCheckpointedStrategy` is the original strategy, which accumulates until a checkpoint is emitted.
- `EmitImmediatelyStrategy` is the new strategy that streams changes right away. It's important to note that makes the client responsible of:
  - deduplicating changes
  - ordering revisions
  - accumulate changes until a checkpoint is emitted

Essentially the client is responsible for doing all the work that used to be done with the `Change` struct helper.

For every other datastore, calling Watch API with `EmitImmediatelyStrategy` will return an error. We don't consider there is an evident need at the moment for other data stores. Once we've proven this strategy is viable in production workloads, we will study implementing a Watch API strategy that buffers to disk to address out-of-memory issues, and potentially deprecate the emit immediately strategy in favor of one that does all the heavy lifting for the client.

### DIFF
--- a/internal/datastore/crdb/watch.go
+++ b/internal/datastore/crdb/watch.go
@@ -10,7 +10,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/authzed/spicedb/internal/datastore/common"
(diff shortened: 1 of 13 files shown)


internal/datastore/crdb/watch.go
```diff
@@ -193,7 +202,130 @@ func (cds *crdbDatastore) watch(
...
+
+type (
+	sendChangeFunc func(change *datastore.RevisionChanges) error
+	sendErrorFunc  func(err error)
+)
+
+func (cds *crdbDatastore) processChanges(ctx context.Context, changes pgx.Rows, sendError sendErrorFunc, sendChange sendChangeFunc, opts datastore.WatchOptions, streaming bool) {
+	var tracked ChangeTracker[revisions.HLCRevision, revisions.HLCRevision]
+	if streaming {
+		tracked = &StreamingChangeProvider{
```
josephschorr, line 321: why is this exported?

vroldanbet: it does not have to, this is the internal package. I'm ok making it unexported, and same for the interface

josephschorr: Please do

vroldanbet: done

josephschorr (approved):
LGTM

adinhodovic (approved):

PR #2122, opened 2024-11-06
fix HCLRevision mismatch on zero logical clock

vroldanbet:
While working on https://github.com/authzed/spicedb/pull/2120, I detected something I wasn't expecting: when comparing the revision out of the datastore write, and what came out of Watch API, the revisions were different.
We use `SHOW COMMIT TIMESTAMP` to retrieve a CRDB [transaction timestamp](https://github.com/cockroachdb/cockroach/pull/80848). The value coming out of the `update` field in change streams had the same value, but the difference was the notation: one included the logical clock, the other didn't, but logical clocks were the same (zero):

- revision generated out of the transaction in `readTransactionCommitRev` uses `NewForHLC(hlcNow)`, which takes a `Decimal`. This in turn calls `decimal.String()`, which returns a number stripped out of the decimal part if it's zero.
- revision obtained out of the changefeeds uses `revisions.HLCRevisionFromString(details.Updated)`, and `details.Updated` always comes with the decimal part, even when it's zero

Both timestamps were the same, but had a different string representation, and led to a different `HCLRevision` logical lock, hence `.Equal()` method failing.

The problem originates in the use of the offset to represent the logical part of the clock. The representation of "zero" wasn't being used consistently, and in some parts of the code the revision was initialized with logical clock of zero, when it needed the offset padding.

⚠️ One of the consequences of this change is that moving forward string representation of the revision will be _longer_ as it will always include the logical clock (when using `HLCRevision`, that is, which only applies to CRDB). This means caches that append the revision to keys will be longer.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -67,6 +67,9 @@ jobs:
           format: "table"
           exit-code: "1"
           severity: "CRITICAL,HIGH,MEDIUM"
+        env:
+          TRIVY_DB_REPOSITORY: "public.ecr.aws/aquasecurity/trivy-db"
+          TRIVY_JAVA_DB_REPOSITORY: "public.ecr.aws/aquasecurity/trivy-java-db"
       - uses: "goreleaser/goreleaser-action@v6"
         id: "goreleaser"
         with:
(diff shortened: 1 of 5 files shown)


josephschorr (approved):
LGTM

PR #2123, opened 2024-11-06
Handle circular and duplicate imports

tstirrat15:
## Description
Circular imports are a normal problem in an import system. This handles them, along with imports via multiple paths, by keeping a global registry of visited files and warning and short-circuiting when a file is visited multiple times.

## Notes
This is a fairly naive implementation for the time being. One thing that we have going for us is that a schema is entirely static - there's no logic to run at import time, so import order doesn't particularly matter.

## Changes
* Add test cases for circular imports and imports via multiple paths
* Refactor `compileImpl` to take a compilation context and add visited files to that context
* Track visited files in the importer

## Testing
Review. See that things go green. Suggest any test cases I haven't thought of.

### DIFF
--- a/pkg/composableschemadsl/compiler/importer.go
+++ b/pkg/composableschemadsl/compiler/importer.go
@@ -13,12 +13,19 @@ import (
 )
 
 type importContext struct {
-	pathSegments []string
-	sourceFolder string
-	names        *mapz.Set[string]
+	pathSegments         []string
+	sourceFolder         string
+	names                *mapz.Set[string]
+	locallyVisitedFiles  *mapz.Set[string]
+	globallyVisitedFiles *mapz.Set[string]
 }
 
 const SchemaFileSuffix = ".zed"
+
+type ErrCircularImport struct {
+	error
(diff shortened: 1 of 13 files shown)


pkg/composableschemadsl/compiler/importer.go
```diff
@@ -26,25 +28,50 @@ func importFile(importContext importContext) (*CompiledSchema, error) {
 
 	newSourceFolder := filepath.Dir(filePath)
 
-	var schemaBytes []byte
+	currentLocallyVisitedFiles := importContext.locallyVisitedFiles.Copy()
+
+	if ok := currentLocallyVisitedFiles.Add(filePath); !ok {
+		return nil, fmt.Errorf("circular import detected: %s has been visited on this branch", filePath)
```
josephschorr, line 34: this should be a well-typed developer error, attached to the node as an error, rather than returning as an internal error

josephschorr (dismissed):
LGTM

josephschorr (approved):

PR #2131, opened 2024-11-12
Add subject filters in schema relation delete to force use of the index

josephschorr:
This should allow for much faster queries under the transaction

### DIFF
--- a/internal/services/shared/schema.go
+++ b/internal/services/shared/schema.go
@@ -318,9 +318,45 @@ func sanityCheckNamespaceChanges(
 	for _, delta := range diff.Deltas() {
 		switch delta.Type {
 		case nsdiff.RemovedRelation:
+			// NOTE: We add the subject filters here to ensure the reverse relationship index is used
+			// by the datastores. As there is no index that has {namespace, relation} directly, but there
+			// *is* an index that has {subject_namespace, subject_relation, namespace, relation}, we can
+			// force the datastore to use the reverse index by adding the subject filters.
+			var previousRelation *core.Relation
+			for _, relation := range existing.Relation {
+				if relation.Name == delta.RelationName {
+					previousRelation = relation
+					break
+				}
+			}
+
+			if previousRelation == nil {
+				return nil, spiceerrors.MustBugf("relation `%s` not found in existing namespace definition", delta.RelationName)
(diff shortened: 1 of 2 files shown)


internal/services/shared/schema_test.go
```diff
@@ -11,64 +11,305 @@ import (
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/schemadsl/input"
+	"github.com/authzed/spicedb/pkg/tuple"
 )
 
 func TestApplySchemaChanges(t *testing.T) {
```
vroldanbet, line 17: we are missing a test with wildcards

internal/services/shared/schema.go
```diff
@@ -318,9 +318,45 @@ func sanityCheckNamespaceChanges(
...
+				}
+			}
+
+			if previousRelation == nil {
+				return nil, spiceerrors.MustBugf("relation `%s` not found in existing namespace definition", delta.RelationName)
+			}
+
+			subjectSelectors := make([]datastore.SubjectsSelector, 0, len(previousRelation.TypeInformation.AllowedDirectRelations))
+			for _, allowedType := range previousRelation.TypeInformation.AllowedDirectRelations {
+				if allowedType.GetRelation() != "" {
```
vroldanbet, line 339: I think this worked by _happenstance_, but the ellipsis was not being handled correctly, as it's not modeled as empty string at this level of the stack, but `...`.

There is also no explicit handling of wildcards, but my understanding is that's modeled as an `userset_object_id = *`, but `userset_relation = ""`, so it ended up falling in ellipsis branch, which would generate the wrong `SubjectRelationFilter`.

I've force-pushed the changes, let me know what you think

vroldanbet (changes requested):

vroldanbet (commented):
I force pushed the changes, if you think they look fine, we can approve this.

We have some examples on how to test for Datastore API calls selecting specific indexes, I'd suggest adding a test to make sure this does not regress, but also confirm that what we wanted to achieve actually is happening.

vroldanbet (approved):

PR #2137, opened 2024-11-20
add a flag to allow spicedb to run against non-head migrations

### LINKED ISSUE #2135: Add a flag to override allowed migrations in health check
### Problem Statement

Right now, the health check for SpiceDB is very restrictive; in most cases (except where we hardcode specific exceptions) the datastore health check requires that the database be at exactly one migration level, the latest for that specific version of SpiceDB.

This health check only has to succeed once on startup, and then the health check ignores the migration state. Under normal circumstances, this doesn't cause problems during upgrades because new SpiceDB versions (with newer migrations) will start up while the old instances are still running.

This can cause (and has caused) problems if the "old" instances of SpiceDB need to restart for some reason after a newer migration has run - they will fail to come up healthy. This is in spite of the majority of the migrations for SpiceDB being fully backwards-compatible (the exceptions being phased migrations, in which case there is always some step to run that is backwards-compatible).

We can't simply allow the "next" migration, because we don't know what the "next" one is before we publish a specific version of spicedb. We also may want to skip ahead several migrations (i.e. the spicedb operator allows this as long as it is safe).

### Solution Brainstorm

We could potentially add something smarter to the health check, so that SpiceDB only ensures that it has what it needs to work and not a specific migration name (i.e. check that field X exists on table Y).

But a simpler option would just be to add a flag, i.e. `--alowed-migrations` that allow a user to specify a set of additional migrations that are considered safe on startup. The operator can set this flag to the migration of the newer version and roll the "old" version with that flag before deploying the migration itself. Then the old instances will continue to work even if they need to restart / reschedule after the migration has run.

ecordell:
fixes #2135

### DIFF
--- a/internal/datastore/common/migrations_test.go
+++ b/internal/datastore/common/migrations_test.go
@@ -1,0 +1,69 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/pkg/datastore"
+)
+
+func TestMigrationReadyState(t *testing.T) {
+	tests := []struct {
+		name                        string
+		additionalAllowedMigrations []string
+		headMigration               string
+		version                     string
+		want                        datastore.ReadyState
+	}{
+		{
+			name:          "matches head migration",
+			headMigration: "initial",
+			version:       "initial",
+			want:          datastore.ReadyState{IsReady: true},
+		},
+		{
+			name:                        "matches additional migration",
+			headMigration:               "initial",
+			additionalAllowedMigrations: []string{"additional"},
+			version:                     "additional",
+			want:                        datastore.ReadyState{IsReady: true},
+		},
+		{
+			name:                        "matches one of several additional migrations",
+			headMigration:               "initial",
+			additionalAllowedMigrations: []string{"additional", "additional2", "additional3"},
+			version:                     "additional2",
+			want:                        datastore.ReadyState{IsReady: true},
+		},
+		{
+			name:                        "matches head migration when additional migrations are allowed",
+			headMigration:               "initial",
+			additionalAllowedMigrations: []string{"additional"},
+			version:                     "initial",
+			want:                        datastore.ReadyState{IsReady: true},
+		},
+		{
+			name:          "doesn't match head migration",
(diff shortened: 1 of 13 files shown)


internal/datastore/common/migrations_test.go
```diff
@@ -0,0 +1,61 @@
...
+			additionalAllowedMigrations: []string{"additional"},
+			version:                     "additional",
+			want:                        datastore.ReadyState{IsReady: true},
+		},
+		{
+			name:          "matches head migration",
+			headMigration: "initial",
+			version:       "initial",
+			want:          datastore.ReadyState{IsReady: true},
+		},
```
tstirrat15, line 37: Dupe of top test?

internal/datastore/mysql/datastore.go
```diff
@@ -214,6 +214,11 @@ func newMySQLDatastore(ctx context.Context, uri string, replicaIndex int, option
 	// ID value makes this idempotent (i.e. safe to execute concurrently).
 	createBaseTxn := fmt.Sprintf("INSERT IGNORE INTO %s (id, timestamp) VALUES (1, FROM_UNIXTIME(1))", driver.RelationTupleTransaction())
 
+	headMigration, err := migrations.Manager.HeadRevision()
+	if err != nil {
+		return nil, fmt.Errorf("invalid head migration found for mysql: %w", err)
+	}
+
```
tstirrat15, line 221: Was this a reorganization or was it required by the refactor?

tstirrat15 (approved):
See questions, otherwise LGTM

internal/datastore/common/migrations_test.go
ecordell: nice catch! this was supposed to be a case matching the head revision but with additional ones allowed.

internal/datastore/mysql/datastore.go
ecordell: Yeah, previously this was called indirectly by

```
compatible, err := migrations.Manager.IsHeadCompatible(currentMigrationRevision)
```

tstirrat15 (approved):
LGTM

PR #2139, opened 2024-11-21
issue a checkpoint when head revision moved outside an application transaction

vroldanbet:
it's possible for `pg_current_snapshot()` (used for HeadRevision) in PG to return a new xmin:xmax outside of application transactions. For example, running `ANALYZE;` could bump it.

In this situation, we need to emit a Checkpoint, in case clients triggered a call to Watch API based on a different HeadRevision from what they had cached locally (e.g. relevant to compute change deltas between revisions).

The proposal is to call `HeadRevision` if there are no changes. This turns the Watch API heartbeat into 2 queries instead of 1, which can add an extra load to the database.


internal/datastore/postgres/revisions.go
```diff
@@ -91,15 +92,28 @@ func (pgd *pgDatastore) HeadRevision(ctx context.Context) (datastore.Revision, e
 	ctx, span := tracer.Start(ctx, "HeadRevision")
 	defer span.End()
 
+	result, err := pgd.getHeadRevision(ctx, pgd.readPool)
+	if err != nil {
+		return nil, err
+	}
+	if result == nil {
```
ecordell, line 99: why not have `getHeadRevision` return datastore.NoRevision here?

internal/testserver/datastore/postgres.go
```diff
@@ -25,7 +25,7 @@ const (
 	POSTGRES_TEST_USER            = "postgres"
 	POSTGRES_TEST_PASSWORD        = "secret"
 	POSTGRES_TEST_PORT            = "5432"
-	POSTGRES_TEST_MAX_CONNECTIONS = "2000"
+	POSTGRES_TEST_MAX_CONNECTIONS = "3000"
```
ecordell, line 28: why this change?

pkg/datastore/test/watch.go
```diff
@@ -727,7 +727,7 @@ func WatchCheckpointsTest(t *testing.T, tester DatastoreTester) {
 	require.NoError(err)
 
 	changes, errchan := ds.Watch(ctx, lowestRevision, datastore.WatchOptions{
-		Content:            datastore.WatchCheckpoints | datastore.WatchRelationships | datastore.WatchSchema,
+		Content:            datastore.WatchCheckpoints | datastore.WatchRelationships | datastore.WatchSchema, //
```
ecordell, line 730:
```suggestion
		Content:            datastore.WatchCheckpoints | datastore.WatchRelationships | datastore.WatchSchema,
```

pkg/datastore/test/watch.go
```diff
@@ -770,7 +770,7 @@ func WatchEmissionStrategyTest(t *testing.T, tester DatastoreTester) {
 	require.NoError(err)
 
 	changes, errchan := ds.Watch(ctx, lowestRevision, datastore.WatchOptions{
-		Content:            datastore.WatchCheckpoints | datastore.WatchRelationships,
+		Content:            datastore.WatchCheckpoints | datastore.WatchRelationships, //
```
ecordell, line 773:
```suggestion
		Content:            datastore.WatchCheckpoints | datastore.WatchRelationships,
```

internal/datastore/postgres/watch.go
```diff
@@ -197,9 +219,18 @@ func (pgd *pgDatastore) Watch(
 	return updates, errs
 }
 
-func (pgd *pgDatastore) getNewRevisions(ctx context.Context, afterTX postgresRevision) ([]postgresRevision, error) {
+func (pgd *pgDatastore) getNewRevisions(ctx context.Context, afterTX postgresRevision, returnHeadRevision bool) ([]postgresRevision, *postgresRevision, error) {
```
ecordell, line 222: personally I think I'd find it easier to come back to this code if it were split into `getNewRevisions` and `getNewRevisionsWithHead`

Then you wouldn't have to do any nil checking for the `optionalHeadRevision` (because you'd just return an error from `getNewRevisionsWithHead`)

internal/datastore/postgres/watch.go
```diff
@@ -181,8 +181,30 @@ func (pgd *pgDatastore) Watch(
...
+							IsCheckpoint: true,
+						}) {
+							return
+						}
+					}
+				}
 
+				sleep := time.NewTimer(watchSleep)
 				select {
 				case <-sleep.C:
```
ecordell, line 209: nit: I think `case <-time.NewTimer(watchSleep).C` would work here?

ecordell (commented):
nice work finding this! very subtle bug.

left a couple of small comments, looks good overall

internal/datastore/postgres/revisions.go
vroldanbet: because the return type is `pgRevision`. I did that so I didn't have to cast from `datastore.Revision` to `pgRevision`.

pkg/datastore/test/watch.go
vroldanbet: sorry - I marked all tests locally that could be affected by this to review them. fixed.

internal/datastore/postgres/watch.go
vroldanbet: so this method is only used in 1 place, that's the postgres watch loop. And we pass  `returnHeadRevision` arg to it as true if the client requested `datastore.WatchCheckpoints`, which is optional (e.g. SpiceDB Watch API does not currently return checkpoints), so it was added as an optimization in case someone does not need checkpoints.

so that refactor would be useful if we used the method in 2 different places, and one of them needed either always head, or never head, but that's not the case - it's entirely driven by the `Datastore.Watch` call inputs. Does that make sense?

ecordell (approved):

PR #2142, opened 2024-11-25
Schema compiler support for relationship expiration trait

josephschorr:
Part 2 of support for expiration trait: adds support for the new trait in the schema IR, the compiler and the generator


proto/internal/core/v1/core.proto
```diff
@@ -412,8 +412,18 @@ message AllowedRelation {
...
+  /**
+   * required_expiration defines the required expiration on this relation.
+   */
+  ExpirationTrait required_expiration = 7;
 }
 
+/**
+ * ExpirationTrait is an expiration trait of a relation.
+ */
+message ExpirationTrait {}
```
tstirrat15, line 425: Is this functionally just a presence/absence marker?

tstirrat15 (approved):
See comments, otherwise LGTM

PR #2143, opened 2024-11-26
adds index to support efficient querying of Watch API

vroldanbet:
we identified Watch API was causing a full table scan on large postgres-backed SpiceDB, overloading the database.

This commit adds an index that reduces the execution time of relationship changes to be emitted via Watch API by more than three orders of magnitude on tables with several gigabytes worth of data.

### DIFF
--- a/internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
+++ b/internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
@@ -1,0 +1,23 @@
+package migrations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+)
+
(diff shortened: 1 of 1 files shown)


internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
```diff
@@ -0,0 +1,23 @@
+package migrations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const addWatchAPIIndexToRelationTupleTable = `CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_watch_index ON relation_tuple (created_xid);`
```
Verolop, line 10: just a thought for later: since the query also filters on `deleted_xid`, we might want to try a composite index on (created_xid, deleted_xid) down the line.

`CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_watch_index ON relation_tuple (created_xid, deleted_xid);`

It could make queries with both columns even faster by improving how rows are filtered. Definitely not a blocker tho.

PR #2144, opened 2024-11-26
Add explicit option to enable expiration in schema

josephschorr:
This will allow us to test without allowing the keyword otherwise


tstirrat15 (approved):
LGTM

PR #2149, opened 2024-11-29
fix: schema watch causes incorrect delete event for schema changes

vroldanbet:
the computation of schema change events was incorrect due to
the use of the wrong snapshot XID for changed namespaces / caveats.
It manifested itself as a change event, followed by a delete event.


josephschorr (approved):
LGTM

PR #2152, opened 2024-12-03
End to end support for experimental first-class relationship expiration feature

### LINKED ISSUE #1262: GC of relationships expired via caveats
With the caveats implementation now available, we have the option to support TTL on relationships by doing something like follows:

```
  caveat is_not_expired(current_time timestamp, expiration timestamp) {
    current_time < expiration
  }

  definition document {
    relation viewer: user with is_not_expired
    permission view = viewer
  }
```
(example took from here https://github.com/authzed/spicedb/issues/282#issuecomment-1478314417)

Some use-cases might require using SpiceDB to store solely "session related" permissions (ephemeral). This would cause having a datastore that grows rapidly in size. It would be nice to support some kind of GC to prevent having an ever growing datastore containing relationships that are technically "expired".

josephschorr:
Fixes #1262 by superseding it

When the flag is enabled, relationship expiration can be used in schema via the `use expiration` directive and the `expiration` keyword on `relation`'s:

```
use expiration

definition user {}

definition resource {
  relation viewer: user with expiration
}
```

When a relationship reaches its expiration, it is immediately filtered from all relationship queries, making it appear as-if it was immediately deleted.

Subsequently (typically 24 hours later), the relationship will be actually removed from the underlying datastore to free space.

> [!WARNING]  
> Currently, relationships are filtered using the **wall clock** found *within the datastore*, which has the following consequences:
> 1. For multi-master datastores such as CRDB and Spanner, there may be some slight skew between nodes on clocks, which could lead to different results if queries are called very close to the expiration time for relationship(s) to be returned
> 2. Caches are *not* updated if a relationship expires within the caching window, e.g. if a permission answer is computed at time N, a relationship used as part of that computation expires at time N+1 and the cache entry is valid until N+10, the caller will see the cached answer even if the relationship has expired - this is in line with caches reflecting the state of the world when they were computed, but is a change from current behavior when using caveats for expiration
>  3. `at_exact_snapshot` calls, as they are made to the datastore, *will* reflect the expiration as of "now", which means they can return *different* results depending on when they are invoked. This is a change from current behavior when using caveats for expiration. Note that this may be changed in the future to instead use the timestamp of the revision itself; this change is currently under discussion

> [!NOTE]  
> Because expiring relationships are not removed immediately upon expiration, a `CREATE` operation will fail if the relationship is attempted to be re-created. Like caveats, a `TOUCH` can be used to change the expiration and move the relationship forward

> [!NOTE]  
> This PR does not yet implement returning of the `expires_at` in CheckPermissionResponse and CheckPermission debug information. That will be done in a followup PR.


pkg/cmd/testserver/testserver.go
```diff
@@ -84,6 +84,7 @@ func (c *Config) Complete() (RunnableTestServer, error) {
 				MaxLookupResourcesLimit:         c.MaxLookupResourcesLimit,
 				MaxBulkExportRelationshipsLimit: c.MaxBulkExportRelationshipsLimit,
 				DispatchChunkSize:               defaultMaxChunkSize,
+				ExpiringRelationshipsEnabled:    true,
```
tstirrat15, line 87: Same here - should this reference the config?

pkg/development/devcontext.go
```diff
@@ -166,12 +166,13 @@ func (dc *DevContext) RunV1InMemoryService() (*grpc.ClientConn, func(), error) {
...
 	ps := v1svc.NewPermissionsServer(dc.Dispatcher, v1svc.PermissionsServerConfig{
-		MaxUpdatesPerWrite:    50,
-		MaxPreconditionsCount: 50,
-		MaximumAPIDepth:       50,
-		MaxCaveatContextSize:  0,
+		MaxUpdatesPerWrite:           50,
+		MaxPreconditionsCount:        50,
+		MaximumAPIDepth:              50,
+		MaxCaveatContextSize:         0,
+		ExpiringRelationshipsEnabled: true,
```
tstirrat15, line 173: Here as well

tstirrat15 (dismissed):
See comments, otherwise LGTM

internal/datastore/common/gc.go
```diff
@@ -185,23 +193,31 @@ func RunGarbageCollection(gc GarbageCollector, window, timeout time.Duration) er
...
 	collected, err := gc.DeleteBeforeTx(ctx, watermark)
 
+	expiredRelationshipsCount, eerr := gc.DeleteExpiredRels(ctx)
+
 	// even if an error happened, garbage would have been collected. This makes sure these are reflected even if the
 	// worker eventually fails or times out.
 	gcRelationshipsCounter.Add(float64(collected.Relationships))
 	gcTransactionsCounter.Add(float64(collected.Transactions))
 	gcNamespacesCounter.Add(float64(collected.Namespaces))
+	gcExpiredRelationshipsCounter.Add(float64(expiredRelationshipsCount))
```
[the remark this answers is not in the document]
josephschorr: Yes

internal/datastore/crdb/crdb.go
```diff
@@ -344,21 +347,25 @@ func (cds *crdbDatastore) ReadWriteTx(
...
 		}
 
-		// If metadata is to be attached, write that row now.
-		if config.Metadata != nil && len(config.Metadata.GetFields()) > 0 {
-			expiresAt := time.Now().Add(cds.gcWindow).Add(1 * time.Minute)
-			insertTransactionMetadata := psql.Insert(tableTransactionMetadata).
-				Columns(colExpiresAt, colMetadata).
-				Values(expiresAt, config.Metadata.AsMap())
+		// Write metadata onto the transaction.
+		metadata := config.Metadata.AsMap()
```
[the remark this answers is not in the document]
josephschorr: Line 362 executes the insert statement for it; I fixed the reference

internal/datastore/crdb/watch.go
```diff
@@ -549,7 +590,9 @@ func (cds *crdbDatastore) processChanges(ctx context.Context, changes pgx.Rows,
 					return
 				}
 
-				if err := tracked.SetRevisionMetadata(ctx, rev, details.After.Metadata); err != nil {
+				adjustedMetadata := details.After.Metadata
```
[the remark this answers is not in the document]
josephschorr: In this case, its safe because we're not reusing it anywhere else

internal/services/v1/relationships.go
```diff
@@ -115,6 +118,7 @@ func NewPermissionsServer(
 		MaxLookupResourcesLimit:         defaultIfZero(config.MaxLookupResourcesLimit, 1_000),
 		MaxBulkExportRelationshipsLimit: defaultIfZero(config.MaxBulkExportRelationshipsLimit, 100_000),
 		DispatchChunkSize:               defaultIfZero(config.DispatchChunkSize, 100),
+		ExpiringRelationshipsEnabled:    true,
```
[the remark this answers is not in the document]
josephschorr: Yes, for testing. We then explicitly disable it when starting SpiceDB, as *its* default is `false`

pkg/cmd/testserver/testserver.go
josephschorr: No; I think its fine to enable by default in the test server

tstirrat15 (dismissed):
LGTM

pkg/tuple/hashing.go
```diff
@@ -42,7 +43,8 @@ func CanonicalBytes(rel Relationship) ([]byte, error) {
 
 	if rel.OptionalExpiration != nil {
 		sb.WriteString(" with $expiration:")
-		sb.WriteString(rel.OptionalExpiration.Format(expirationFormat))
+		truncated := rel.OptionalExpiration.UTC().Truncate(time.Second)
```
vroldanbet, line 46: why truncating to second resolution?

pkg/tuple/v1_test.go
```diff
@@ -44,3 +46,26 @@ func TestJoinSubjectRef(t *testing.T) {
...
+		if tc.stableCanonicalization == "" {
+			continue
+		}
+
+		t.Run("relationship/"+tc.input, func(t *testing.T) {
+			v1rel := ToV1Relationship(tc.relFormat)
+			testutil.RequireProtoEqual(t, tc.v1Format, v1rel, "mismatch in v1 proto")
+			adjusted := tc.relFormat
+
+			// Converting to V1 removes the timezones.
```
vroldanbet, line 62: Does this mean that when the user writes with timezone, and reads back via `ReadRelationships`, they get UTC timestamps? I think that's fine, but could make the life of the user harder if they are not properly handling timezones. Please document in PR body.

pkg/datastore/test/watch.go
```diff
@@ -74,7 +74,7 @@ func WatchTest(t *testing.T, tester DatastoreTester) {
 
 			opts := datastore.WatchOptions{
 				Content:                 datastore.WatchRelationships,
-				WatchBufferLength:       128,
+				WatchBufferLength:       50,
```
vroldanbet, line 77: why the change?

internal/datastore/memdb/memdb.go
```diff
@@ -105,11 +105,11 @@ func (mdb *memdbDatastore) SnapshotReader(dr datastore.Revision) datastore.Reade
 	defer mdb.RUnlock()
 
 	if len(mdb.revisions) == 0 {
-		return &memdbReader{nil, nil, fmt.Errorf("memdb datastore is not ready")}
+		return &memdbReader{nil, nil, fmt.Errorf("memdb datastore is not ready"), time.Now()}
```
vroldanbet, line 108: we could accidentally instantiate `memdbReader` without transaction timestamp, and impact expiration logic. Please create constructor function that forces us to add it.

internal/datastore/memdb/schema.go
```diff
@@ -101,30 +100,18 @@ func (r relationship) String() string {
 		caveat = "[" + r.caveat.caveatName + "]"
 	}
 
-	return r.namespace + ":" + r.resourceID + "#" + r.relation + "@" + r.subjectNamespace + ":" + r.subjectObjectID + "#" + r.subjectRelation + caveat
+	expiration := ""
+	if r.expiration != nil {
+		expiration = "[expiration:" + r.expiration.Format(time.RFC3339Nano) + "]"
```
vroldanbet, line 105: didn't we define a constant for the format we adopted somewhere? Please reuse to avoid drift

internal/relationships/errors.go
```diff
@@ -48,7 +48,8 @@ func NewInvalidSubjectTypeError(
 			if allowedType.RequiredCaveat != nil &&
 				allowedType.RequiredCaveat.CaveatName != "" &&
 				allowedType.Namespace == relationship.Subject.ObjectType &&
-				allowedType.GetRelation() == relationship.Subject.Relation {
+				allowedType.GetRelation() == relationship.Subject.Relation &&
+				(allowedType.RequiredExpiration != nil) == (relationship.OptionalExpiration != nil) {
```
vroldanbet, line 52: please add test

internal/services/v1/experimental.go
```diff
@@ -196,6 +196,13 @@ func (a *bulkLoadAdapter) Next(_ context.Context) (*tuple.Relationship, error) {
 	a.current.Subject.ObjectID = a.currentBatch[a.numSent].Subject.Object.ObjectId
 	a.current.Subject.Relation = stringz.DefaultEmpty(a.currentBatch[a.numSent].Subject.OptionalRelation, tuple.Ellipsis)
 
+	if a.currentBatch[a.numSent].OptionalExpiresAt != nil {
+		t := a.currentBatch[a.numSent].OptionalExpiresAt.AsTime()
```
vroldanbet, line 200: super nit

```suggestion
	if expiresAt := a.currentBatch[a.numSent].OptionalExpiresAt; expiresAt != nil {
		t := expiresAt.AsTime()
```

internal/services/v1/permissions.go
```diff
@@ -785,6 +785,13 @@ func (a *loadBulkAdapter) Next(_ context.Context) (*tuple.Relationship, error) {
 	a.current.Subject.ObjectID = a.currentBatch[a.numSent].Subject.Object.ObjectId
 	a.current.Subject.Relation = stringz.DefaultEmpty(a.currentBatch[a.numSent].Subject.OptionalRelation, tuple.Ellipsis)
 
+	if a.currentBatch[a.numSent].OptionalExpiresAt != nil {
```
vroldanbet, line 788: please add V1 API test that asserts expiration is being bulk loaded correctly.

internal/services/v1/schema.go
```diff
@@ -35,15 +35,17 @@ func NewSchemaServer(additiveOnly bool) v1.SchemaServiceServer {
 				usagemetrics.StreamServerInterceptor(),
 			),
 		},
-		additiveOnly: additiveOnly,
+		additiveOnly:        additiveOnly,
+		expiringRelsEnabled: expiringRelsEnabled,
```
vroldanbet, line 39: super nit: maybe invert this flag, since enabled is the default?

internal/services/v1/schema_test.go
```diff
@@ -585,3 +585,120 @@ func TestSchemaInvalid(t *testing.T) {
...
+	_, err = v1client.WriteRelationships(context.Background(), &v1.WriteRelationshipsRequest{
+		Updates: []*v1.RelationshipUpdate{tuple.MustUpdateToV1RelationshipUpdate(tuple.Create(
+			toWrite,
+		))},
+	})
+	require.Nil(t, err)
+
+	newSchema := "definition document {\n\trelation somerelation: user\n}\n\ndefinition user {}"
+
+	// Attempt to change the schema to remove the expiration, which should work.
```
vroldanbet, line 694: can you also do the inverse? add rel that has expiration, change schema to remove non-expirable type

pkg/typesystem/typesystem.go
```diff
@@ -691,6 +691,46 @@ func (nts *TypeSystem) TypeSystemForNamespace(ctx context.Context, namespaceName
 	return NewNamespaceTypeSystem(nsDef, nts.resolver)
 }
 
+// RelationDoesNotAllowCaveatsOrTraitsForSubject returns true if and only if it can be conclusively determined that
+// the given subject type does not accept any caveats or traits on the given relation. If the relation does not have type information,
+// returns an error.
+func (nts *TypeSystem) RelationDoesNotAllowCaveatsOrTraitsForSubject(relationName string, subjectTypeName string) (bool, error) {
```
vroldanbet, line 697: please add method test

pkg/cmd/server/server.go
```diff
@@ -108,18 +108,19 @@ type Config struct {
...
+	MaximumPreconditionCount                 uint16        `debugmap:"visible"`
+	MaxDatastoreReadPageSize                 uint64        `debugmap:"visible"`
+	StreamingAPITimeout                      time.Duration `debugmap:"visible"`
+	WatchHeartbeat                           time.Duration `debugmap:"visible"`
+	MaxReadRelationshipsLimit                uint32        `debugmap:"visible"`
+	MaxDeleteRelationshipsLimit              uint32        `debugmap:"visible"`
+	MaxLookupResourcesLimit                  uint32        `debugmap:"visible"`
+	MaxBulkExportRelationshipsLimit          uint32        `debugmap:"visible"`
+	EnableExperimentalLookupResources        bool          `debugmap:"visible"`
+	EnableExperimentalRelationshipExpiration bool          `debugmap:"visible"`
```
vroldanbet, line 123: please add default annotation

internal/datastore/spanner/watch.go
```diff
@@ -189,6 +189,18 @@ func (sd *spannerDatastore) watch(
 				changeRevision := revisions.NewForTime(dcr.CommitTimestamp)
 				modType := dcr.ModType // options are INSERT, UPDATE, DELETE
 
+				// See: https://cloud.google.com/spanner/docs/ttl
+				// > TTL supports auditing its deletions through change streams. Change
+				// > streams data records that track TTL changes to a database have the
+				// > transaction_tag field set to RowDeletionPolicy and the
+				// > is_system_transaction field set to true.
+				if modType == "DELETE" && dcr.TransactionTag == "RowDeletionPolicy" && dcr.IsSystemTransaction {
```
vroldanbet, line 197: isn't there any constant in the spanner SDK? Maybe open an issue in the repo since this seems API that could break any time? Same for `DELETE`

internal/datastore/spanner/watch.go
```diff
@@ -189,6 +189,18 @@ func (sd *spannerDatastore) watch(
...
+					// Skip deletions that are performed by TTL policy.
+					// TODO(jschorr): once we decide to emit events for GCed expired rels, change to emit those
```
vroldanbet, line 199: Please document in PR body, and open follow up issue documenting the gap

internal/datastore/spanner/migrations/zz_migration.0011_add_expiration_support.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
+)
+
+const (
+	addExpirationColumn = `
+		ALTER TABLE relation_tuple ADD COLUMN expires_at TIMESTAMP
+	`
+
+	addExpirationSupport = `
```
vroldanbet, line 14: please document in PR body that GC is hardcoded and cannot be changed, and open issue describing the gap

pkg/tuple/hashing.go
josephschorr: CRDB doesn't store the nanos and it was breaking tests, so I truncated it for integrity purposes

pkg/tuple/v1_test.go
josephschorr: Yes, especially because V1 does so

pkg/datastore/test/watch.go
josephschorr: See the commit message: it was required to fix the test

internal/datastore/memdb/memdb.go
josephschorr: I checked: its forced by virtue of us using the structural form everywhere and this is internal to the package

internal/datastore/memdb/schema.go
josephschorr: We do, but I don't want to expose it in the tuple package. This is for debugging anyway and doesn't need to match the string format entirely (note it doesn't include caveat context)

internal/relationships/errors.go
josephschorr: I did, in the errors tests

internal/services/v1/permissions.go
josephschorr: I did

internal/services/v1/schema.go
josephschorr: I prefer to leave it as "is this enabled"

internal/services/v1/schema_test.go
josephschorr: There should be a test above that inverts the write. We also have other tests that remove non-expirable types

pkg/typesystem/typesystem.go
```diff
@@ -691,6 +691,46 @@ func (nts *TypeSystem) TypeSystemForNamespace(ctx context.Context, namespaceName
...
+	foundSubjectType := false
+	for _, allowedRelation := range typeInfo.GetAllowedDirectRelations() {
+		if allowedRelation.GetNamespace() == subjectTypeName {
+			foundSubjectType = true
+			if allowedRelation.GetRequiredCaveat() != nil && allowedRelation.GetRequiredCaveat().CaveatName != "" {
+				return false, nil
+			}
+			if allowedRelation.GetRequiredExpiration() != nil {
+				return false, nil
+			}
```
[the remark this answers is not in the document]
josephschorr: New traits will break many other type system tests; I'd much prefer to be explicit here

pkg/cmd/server/server.go
josephschorr: Its not defaulted here - we default it on the flag

internal/datastore/spanner/watch.go
josephschorr: Not that I saw, unfortunately

internal/datastore/spanner/watch.go
josephschorr: Its not a gap, so much as an additional feature we *may* support down the line

internal/datastore/postgres/gc.go
```diff
@@ -68,6 +68,15 @@ func (pgd *pgDatastore) TxIDBefore(ctx context.Context, before time.Time) (datas
 	return postgresRevision{snapshot: snapshot, optionalTxID: value}, nil
 }
 
+func (pgd *pgDatastore) DeleteExpiredRels(ctx context.Context) (int64, error) {
+	return pgd.batchDelete(
+		ctx,
+		tableTuple,
+		gcPKCols,
+		sq.Lt{colExpiration: time.Now().Add(-1 * pgd.gcWindow)},
```
vroldanbet, line 76: Why are we using `time.Now()` here, but for relationships we call `gc.Now()` which calls the datastore? Please use the same function, this is not in the critical path.

internal/datastore/crdb/migrations/zz_migration.0009_add_expiration_support.go
```diff
@@ -0,0 +1,67 @@
...
+		ALTER TABLE relation_tuple_with_integrity
+		ADD COLUMN IF NOT EXISTS expires_at TIMESTAMPTZ DEFAULT NULL;
+	`
+
+	// ttl_expiration_expression support was added in CRDB v22.2.
+	addExpirationPolicy = `
+		ALTER TABLE relation_tuple SET (ttl_expiration_expression = 'expires_at', ttl_job_cron = '@daily');
+
+		ALTER TABLE relation_tuple_with_integrity SET (ttl_expiration_expression = 'expires_at', ttl_job_cron = '@daily');
+	`
```
vroldanbet, line 26: any specific reason we don't want to garbage collect those more frequently? is daily the recommended value?

internal/datastore/crdb/crdb.go
```diff
@@ -90,6 +91,8 @@ const (
 	queryTransactionNowPreV23 = querySelectNow
 	queryTransactionNow       = "SHOW COMMIT TIMESTAMP"
 	queryShowZoneConfig       = "SHOW ZONE CONFIGURATION FOR RANGE default;"
+
+	spicedbTransactionKey = "$spicedb_transaction_key"
```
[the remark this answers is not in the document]
josephschorr: I want to prevent any overlap

internal/datastore/crdb/watch.go
```diff
@@ -361,6 +363,47 @@ func (cds *crdbDatastore) processChanges(ctx context.Context, changes pgx.Rows,
...
+				//
+				// 1) Rows deleted by the TTL cleanup in CRDB will not have any metadata associated with the
+				//    transaction.
+				// 2) No other operations besides DELETE will be performed by the TTL cleanup.
+				// 3) Only filter rows that had an expires_at that are before the current time.
+				if _, ok := revChange.Metadata.AsMap()[spicedbTransactionKey]; !ok {
+					if len(revChange.ChangedDefinitions) == 0 && len(revChange.DeletedNamespaces) == 0 &&
+						len(revChange.DeletedCaveats) == 0 {
+						hasNonTTLEvent := false
+						for _, relChange := range revChange.RelationshipChanges {
```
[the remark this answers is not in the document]
josephschorr: > it would not filter expired relationships if it finds a delete first

You mean if it finds a *non-delete*?

> it would not filter expired relationships if it finds a tuple without expiration first

Which is correct: there cannot, by definition, be a relationship without an expiration in the TTL-based deletion transaction

> it would filter relationships that are not expired, or have no expiration because it found first an expired relationship.

No? It checks all the updates in the change

internal/datastore/crdb/watch.go
```diff
@@ -403,8 +446,12 @@ func (cds *crdbDatastore) processChanges(ctx context.Context, changes pgx.Rows,
 				return
 			}
 
-			var integrity *core.RelationshipIntegrity
+			var expiration *time.Time
+			if details.After != nil {
+				expiration = details.After.RelationshipExpiration
+			}
 
+			var integrity *core.RelationshipIntegrity
```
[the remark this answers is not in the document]
josephschorr: We have them - its done automatically when the test suite is run with integrity enabled

internal/datastore/crdb/watch.go
```diff
@@ -361,6 +363,47 @@ func (cds *crdbDatastore) processChanges(ctx context.Context, changes pgx.Rows,
...
+								hasNonTTLEvent = true
+								break
+							}
+
+							if relChange.Relationship.OptionalExpiration == nil {
+								hasNonTTLEvent = true
+								break
+							}
+
+							if relChange.Relationship.OptionalExpiration.After(time.Now()) {
```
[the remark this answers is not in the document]
josephschorr: The long term goal is to switch from using the local clocks entirely; this is an interim solution

internal/datastore/mysql/migrations/zz_migration.0010_add_expiration.go
```diff
@@ -0,0 +1,27 @@
+package migrations
+
+import "fmt"
+
+func addExpirationToRelationTupleTable(t *tables) string {
+	return fmt.Sprintf(`ALTER TABLE %s
+		ADD COLUMN expiration DATETIME NULL DEFAULT NULL;`,
```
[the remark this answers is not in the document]
josephschorr: I don't consider anything beyond a second to be necessary for this feature - if your expiration relies on fractions of a second, we're already in trouble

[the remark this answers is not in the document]
josephschorr: This is also why we truncate to seconds in the integrity tests

internal/datastore/mysql/query_builder.go
```diff
@@ -174,6 +175,7 @@ func queryRelationships(tableTuple string) sq.SelectBuilder {
 		colUsersetRelation,
 		colCaveatName,
 		colCaveatContext,
+		colExpiration,
 	).From(tableTuple)
 }
```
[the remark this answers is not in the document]
josephschorr: They do; all use the same query builder which adds the filter

internal/datastore/postgres/gc.go
josephschorr: Changed

internal/datastore/crdb/migrations/zz_migration.0009_add_expiration_support.go
josephschorr: Because it has to be outside of the GC window and we recommend GC windows no greater than 24 hours. We can't really support a "dynamic" option here

pkg/tuple/v1_test.go
vroldanbet: what is V1?

pkg/datastore/test/watch.go
vroldanbet: why would a smaller buffer be needed to fix a test?

internal/datastore/crdb/crdb.go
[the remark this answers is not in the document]
vroldanbet: I was thinking disk usage impact, but I realized this is bounded by the GC window.

internal/datastore/crdb/watch.go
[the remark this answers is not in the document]
vroldanbet: and that's fine, but we want to future proof it

internal/datastore/crdb/watch.go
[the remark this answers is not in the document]
vroldanbet: we discussed offline, this is fine:
- this code will only handle transactions originating from GC'ed information, which is only generated by CRDB, never from user-generated transactions
- in CRDB stuff will be GC'ed after 24h, so that `time.Now()` call there drifting is irrelevant

internal/datastore/mysql/migrations/zz_migration.0010_add_expiration.go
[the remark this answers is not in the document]
vroldanbet: then this needs to be documented as well, please add to the PR body

internal/datastore/mysql/query_builder.go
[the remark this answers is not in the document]
vroldanbet: yep, forgot that I actually traced that myself.

vroldanbet (approved):

tstirrat15 (approved):
LGTM

PR #2154, opened 2024-12-04
Changes to address flaky DB tests

josephschorr:
1) Add additional logging when the flake occurs
2) Remove parallel from the PG tests, as they seem to be causing the PG containers to die
3) Make sure to cleanup connections


tstirrat15 (dismissed):
LGTM

PR #2087, opened 2024-10-04
Add currently-enabled workflow to release windows

tstirrat15:
Fixes #2053 

## Description
This should get windows releases going.

## Changes
* Add a release workflow github action
* Add a missing `authors` field in the config
* Change checksum filename so it doesn't overlap with primary release

## Testing
See that it's available here: https://community.chocolatey.org/packages/spicedb

PR #2158, opened 2024-12-09
Remove now-unused windows workflow

tstirrat15:
## Description
This is superseded by what's in https://github.com/authzed/spicedb/blob/main/.github/workflows/release-windows.yaml, which uses a different runner and a different mechanism. I forgot to clean this up in the context of #2087.

## Changes
* Remove now-unused windows release flow
## Testing
Review.


alecmerdler (approved):
Thanks!

PR #2159, opened 2024-12-09
GC improvements: GC only on a single node and add a missing index in PG

josephschorr:
1) Have GC lock so that it only runs on a single node at a time
2) Add a missing index in the Postgres datastore for GC

This should reduce datastore CPU pressure

### DIFF
--- a/pkg/datastore/test/revisions.go
+++ b/pkg/datastore/test/revisions.go
@@ -131,7 +131,7 @@ func RevisionGCTest(t *testing.T, tester DatastoreTester) {
 	gcable, ok := ds.(common.GarbageCollector)
 	if ok {
 		gcable.ResetGCCompleted()
-		require.Eventually(func() bool { return gcable.HasGCRun() }, 5*time.Second, 50*time.Millisecond, "GC was never run as expected")
+		require.Eventually(func() bool { return gcable.HasGCRun() }, 10*time.Second, 150*time.Millisecond, "GC was never run as expected")
 	}
 
 	// FIXME currently the various datastores behave differently when a revision was requested and GC Window elapses.


josephschorr:
Redesigned to use the native locks as discussed

internal/datastore/postgres/migrations/zz_migration.0023_add_missing_gc_index.go
```diff
@@ -0,0 +1,25 @@
+package migrations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const addGCIndexForRelationTupleTransaction = `CREATE INDEX CONCURRENTLY
```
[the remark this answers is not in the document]
josephschorr: Added a comment into the code

ecordell (approved):
LGTM, though need to resolve the test failures

internal/datastore/mysql/locks.go
```diff
@@ -0,0 +1,33 @@
...
+	var acquired int
+	if err := row.Scan(&acquired); err != nil {
+		return false, err
+	}
+
+	return acquired == 1, nil
+}
+
+func (mds *Datastore) releaseLock(ctx context.Context, lockName lockName) error {
+	_, err := mds.db.ExecContext(ctx, `
```
vroldanbet, line 27: Please add comment from MySQL docs:

```suggestion
        // A lock obtained with GET_LOCK() is released explicitly by executing RELEASE_LOCK()
        // or implicitly when your session terminates (either normally or abnormally).
	_, err := mds.db.ExecContext(ctx, `
```

internal/datastore/postgres/migrations/zz_migration.0022_add_missing_gc_index.go
```diff
@@ -0,0 +1,39 @@
...
+//
+// Planning Time: 0.098 ms
+// Execution Time: 5706.192 ms
+// (6 rows)
+const addGCIndexForRelationTupleTransaction = `CREATE INDEX CONCURRENTLY 
+	IF NOT EXISTS ix_relation_tuple_transaction_xid_desc_timestamp
+	ON relation_tuple_transaction (xid DESC, timestamp);`
+
+func init() {
+	if err := DatabaseMigrations.Register("add-missing-gc-index", "add-expiration-support",
```
vroldanbet, line 29: please rename to `add-index-for-transaction-gc`

internal/datastore/mysql/locks.go
```diff
@@ -0,0 +1,33 @@
...
+
+type lockName string
+
+const (
+	// gcRunLock is the lock name for the garbage collection run.
+	gcRunLock lockName = "gc_run"
+)
+
+func (mds *Datastore) tryAcquireLock(ctx context.Context, lockName lockName) (bool, error) {
+	// Acquire the lock, with max 1s timeout.
```
vroldanbet, line 13: // please add comment indicating `GET_LOCK` is reentrant. From the docs:

```
It is even possible for a given session to acquire multiple locks for the same name. Other sessions cannot acquire a lock with that name until the acquiring session releases all its locks for the name.
```

internal/datastore/postgres/locks.go
```diff
@@ -0,0 +1,32 @@
...
+
+type lockID uint32
+
+const (
+	// gcRunLock is the lock ID for the garbage collection run.
+	gcRunLock lockID = 1
+)
+
+func (pgd *pgDatastore) tryAcquireLock(ctx context.Context, lockID lockID) (bool, error) {
+	// Acquire the lock.
```
vroldanbet, line 15: Please link to: https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS

vroldanbet: Please also mention that  pg locks are reentrant, but with a caveat

```
A lock can be acquired multiple times by its owning process; for each completed lock request there must be a corresponding unlock request before the lock is actually released
...
If a session already holds a given advisory lock, additional requests by it will always succeed, even if other sessions are awaiting the lock; this statement is true regardless of whether the existing lock hold and new request are at session level or transaction level.
```

internal/datastore/postgres/locks.go
```diff
@@ -0,0 +1,32 @@
...
+	if err := row.Scan(&lockAcquired); err != nil {
+		return false, err
+	}
+	return lockAcquired, nil
+}
+
+func (pgd *pgDatastore) releaseLock(ctx context.Context, lockID lockID) error {
+	_, err := pgd.writePool.Exec(ctx, `
+		SELECT pg_advisory_unlock($1)
+	`, lockID)
```
vroldanbet, line 30: Please check the result. If it returns false, we know something is wrong, so we should at least log it.

vroldanbet (commented):
Changes look good to me, BUT:

Please note this **only solves a part of the problem**: it prevents multiple nodes from GC'ing simultaneously, which will avoid contention and spikes, but it does not elect a leader node who is the only one that should be running GC.

For example, if 100 nodes are doing GC and have their timers sufficiently skewed, it's possible to have 100 nodes GC'ing one after the other (this is an extreme, worst-case scenario). It won't happen at the same time, but it is still an inefficient use of datastore compute and can cause a sustained load on the datastore, proportional to the number of nodes.

josephschorr:
> For example, if 100 nodes are doing GC and have their timers sufficiently skewed, it's possible to have 100 nodes GC'ing one after the other (this is an extreme, worst-case scenario). It won't happen at the same time, but it is still an inefficient use of datastore compute and can cause a sustained load on the datastore, proportional to the number of nodes.

That's true, but following the first GC, the next N GCs should more or less no-op. I considered adding a "last time GCed", but that would require a new table

josephschorr:
Updated

vroldanbet (dismissed):

ecordell (approved):

PR #2165, opened 2024-12-12
Change GC test to always call GC directly

josephschorr:
This should hopefully remove the flaky nature of the test

### DIFF
--- a/pkg/datastore/test/revisions.go
+++ b/pkg/datastore/test/revisions.go
@@ -92,6 +92,50 @@ func RevisionSerializationTest(t *testing.T, tester DatastoreTester) {
 	require.NoError(meta.Validate())
 }
 
+// GCProcessRunTest tests whether the custom GC process runs for the datastore.
+// For datastores that do not have custom GC processes, will no-op.
+func GCProcessRunTest(t *testing.T, tester DatastoreTester) {
+	require := require.New(t)
+	gcWindow := 300 * time.Millisecond
+	gcInterval := 500 * time.Millisecond
+
+	ds, err := tester.New(0, gcInterval, gcWindow, 1)
+	require.NoError(err)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	testCaveat := createCoreCaveat(t)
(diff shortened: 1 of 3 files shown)


pkg/datastore/test/revisions.go
```diff
@@ -125,13 +127,16 @@ func RevisionGCTest(t *testing.T, tester DatastoreTester) {
 	require.NoError(err)
 	require.NoError(ds.CheckRevision(ctx, head), "expected head revision to be valid in GC Window")
 
-	// Make sure GC kicks in after the window.
-	time.Sleep(300 * time.Millisecond)
```
ecordell, line 129: would this work with 600 ms instead? it seems like the flakiness could be related to jitter?

josephschorr: The flakiness was around the GC process itself running, rather than the window having passed. So long as we have *at least* 300ms, we should be fine (since it is the window size)

ecordell: I'm still concerned about this - if the GC window is 300ms and we're waiting for _10s_ and it's not running again, something seems off somewhere?

ecordell: One thing that might help illuminate - what if we change the `hasGCRun` bool to an atomic uint that we increment on each GC run?

The test can print that value in the eventually check and we can see if GC just isn't running or if there's some issue with marking it as having run.

josephschorr: It is incredibly hard to reproduce locally, but I suspect the problem is that the GC is encountering errors on the locking because the PG container is overloaded. This, in turn, causes the GC runner to continuously "push out" running the GC process (because of the exponential backoff). However, waiting for the *background* process to run is not the point of the test, so I changed it to always run it directly.

We could add a distinct test to ensure the GC process itself runs once, without having it run at an interval of `10ms` (which is what the test had here previously), but we have distinct GC tests for those in each datastore, so I think it is fine to change the generic one to just run it manually

ecordell: What other test asserts that the background GC process actually runs? having trouble finding it.

josephschorr: MySQL appears to have it but Postgres does not. I can add a test to the PG driver if you like or add one that just does a wait-and-see to see if GC runs at all in the generic tests

ecordell (approved):

PR #2166, opened 2024-12-13
Add basic steelthread tests for bulk import and export of relationships

josephschorr:


alecmerdler (approved):

Verolop (approved):
looks good to me.
I realize these are supposed to cover minimal actions, but would you consider adding a negative path test to check whether the system will fail gracefully or just crash in the presence of garbage data. Not a blocker though!

PR #2170, opened 2024-12-16
pkg/cmd: add metric for logical checks

jzelinskie:
This tracks the "logical checks" which includes not only check requests but each item in a bulk check request.

### DIFF
--- a/pkg/cmd/server/defaults.go
+++ b/pkg/cmd/server/defaults.go
@@ -22,6 +22,7 @@ import (
 	"github.com/jzelinskie/cobrautil/v2/cobraproclimits"
 	"github.com/jzelinskie/cobrautil/v2/cobrazerolog"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
@@ -30,6 +31,8 @@ import (
 	"google.golang.org/grpc/codes"
 
 	"github.com/authzed/authzed-go/pkg/requestmeta"
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/authzed/grpcutil"
 
 	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/internal/logging"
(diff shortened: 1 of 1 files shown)


pkg/cmd/server/defaults.go
```diff
@@ -406,6 +415,11 @@ func DefaultStreamingMiddleware(opts MiddlewareOption) (*MiddlewareChain[grpc.St
 			WithInterceptor(serverversion.StreamServerInterceptor(opts.EnableVersionResponse)).
 			Done(),
 
+		NewStreamMiddleware().
+			WithName(DefaultMiddlewareLogicalChecks).
+			WithInterceptor(noopStreamInterceptor).
```
vroldanbet, line 420: Should it be marked `internal`? I lean towards no,  just thinking out loud. Internal middlewares should not be replaced or removed because they are crucial to SpiceDB's correct functioning. Nothing would break in SpiceDB, but if the metric disappears, it will affect folks running SpiceDB who rely on it.

pkg/cmd/server/defaults.go
```diff
@@ -465,6 +483,29 @@ func DefaultDispatchMiddleware(logger zerolog.Logger, authFunc grpcauth.AuthFunc
...
+	response, err := handler(ctx, request)
+	if err == nil {
+		switch m := request.(type) {
+		case *v1.CheckPermissionRequest:
+			svc, method := grpcutil.SplitMethodName(info.FullMethod)
+			LogicalChecksCounter.WithLabelValues(method, svc, "unary").Add(1)
+		case *v1.CheckBulkPermissionsRequest:
+			svc, method := grpcutil.SplitMethodName(info.FullMethod)
+			LogicalChecksCounter.WithLabelValues(method, svc, "unary").Add(float64(len(m.GetItems())))
+		default:
```
vroldanbet, line 503: please add unit test testing testing both request types, and with/without error

pkg/cmd/server/defaults.go
```diff
@@ -465,6 +483,29 @@ func DefaultDispatchMiddleware(logger zerolog.Logger, authFunc grpcauth.AuthFunc
 		}
 }
 
+var LogicalChecksCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "logical",
+	Name:      "permission_checks_total",
```
vroldanbet, line 489: Is there a better namespace for this metric? I'm not sure "logical" is a subsystem. Should it be aligned with the namespace of the grpc prom middleware? or perhaps the subsystem is a permissions service?

josephschorr:
Closing as we added a metric in https://github.com/authzed/spicedb/pull/2325

PR #2173, opened 2024-12-18
Add some additional datastore tests to improve coverage

josephschorr:


pkg/datastore/test/counters.go
```diff
@@ -65,6 +67,41 @@ func RelationshipCounterOverExpiredTest(t *testing.T, tester DatastoreTester) {
...
+
+	// Run multiple registrations of the counter in parallel and ensure only
+	// one succeeds.
+	var numSucceeded atomic.Int32
+	var numFailed atomic.Int32
+	var wg sync.WaitGroup
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			_, err = ds.ReadWriteTx(context.Background(), func(ctx context.Context, tx datastore.ReadWriteTransaction) error {
```
Verolop, line 84: there seems to be a data race due to all goroutines writing to the shared err variable.
Quick workaround: introducing a new variable inside the goroutine:
``` 
localErr := ds.ReadWriteTx(context.Background(), func(ctx context.Context, tx datastore.ReadWriteTransaction) error {

File: pkg/datastore/test/counters.go [hunk 2]
```diff
@@ -65,6 +67,41 @@ func RelationshipCounterOverExpiredTest(t *testing.T, tester DatastoreTester) {
...
+				return tx.RegisterCounter(ctx, "document", &core.RelationshipFilter{
+					ResourceType: testfixtures.DocumentNS.Name,
+				})
+			})
+			if err != nil {
```
Verolop, line 89: updating the reference to the newly introduced variable:
```
if localErr != nil {
    require.Contains(t, localErr.Error(), "counter with name `document` already registered")

Verolop (changes requested):

tstirrat15 (approved):
LGTM

pkg/datastore/test/counters.go
josephschorr: Good catch. Fixed

josephschorr:
Updated

Verolop (approved):

PR #2177, opened 2024-12-19
Add option to enable query parameters to appear in traces

josephschorr:
This should for easier debugging of query performance when necessary


tstirrat15 (approved):
LGTM

PR #2179, opened 2024-12-20
Move caveat loading into a shared runner to reduce overhead in dispatch

josephschorr:
Rather than re-processing the caveat on each iteration of the relationships loop in LR2, we use a caching loader to only process once per caveat

### DIFF
--- a/internal/caveats/run.go
+++ b/internal/caveats/run.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"maps"
 
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+
 	"github.com/authzed/spicedb/pkg/caveats"
 	caveattypes "github.com/authzed/spicedb/pkg/caveats/types"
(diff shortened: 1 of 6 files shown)


internal/caveats/run.go
```diff
@@ -26,125 +31,101 @@ const (
...
-	return !result.Value() && !result.IsPartial()
-}
-
-func isTrueResult(result ExpressionResult) bool {
-	return result.Value() && !result.IsPartial()
 }
 
-func runExpression(
+// RunCaveatExpression runs a caveat expression over the given context and returns the result.
+func (cr *CaveatRunner) RunCaveatExpression(
```
alecmerdler, line 62: It looks like we might be missing a unit test for this new struct and method.

josephschorr: `RunSingleCaveatExpression` does invoke this method, but I'll add a unit test just the same

josephschorr: Added

alecmerdler (approved):
Nice improvement!

PR #2189, opened 2025-01-04
Add asterisk import syntax for import all

tstirrat15:
## Description
This is intended to provide an escape hatch for users who don't want to explicitly list all of the imports they're bringing in from a source file. It should make explicit import behavior less painful in some cases as well, and is intended as a compromise between implicit and explicit import syntax.

## Changes
* Add `*` syntax for imports to say "import everything"
* Add tests for that syntax
* Add predicate as marker on the node

## Testing
Review. See that tests pass.


tstirrat15:
Pausing this pending a design discussion. It'll probably be closed.

tstirrat15:
We're going a different direction.

PR #2193, opened 2025-01-07
Add support for debug traces in Check Bulk Permission

josephschorr:


internal/services/v1/debug.go
```diff
@@ -75,6 +96,10 @@ func convertCheckTrace(ctx context.Context, caveatContext map[string]any, ct *di
 		subRelation = ""
 	}
 
+	spiceerrors.DebugAssert(func() bool {
+		return len(ct.Results) == 1
+	}, "multiple results in a single check trace")
```
Verolop, line 101: there seems to be an inconsistency in the check trace conversion logic. It's causing the wildcard exclusion test to fail because those operations produce multiple results by definition. Isn't that what we want?

josephschorr: Looking

josephschorr: Fixed

Verolop (approved):

pkg/middleware/nodeid/nodeid.go
tstirrat15, line 1: Curious - why did adding tracing to bulk check require adding the concept of a node ID? Is this because a bulk check can potentially break up its dispatch to different nodes and we want to identify them? Why wasn't it necessary in the unary check case?

tstirrat15 (approved):
LGTM, had a question

pkg/middleware/nodeid/nodeid.go
josephschorr: It didn't; the node ID is an additional benefit for debugging

PR #2197, opened 2025-01-08
make index creation idempotent

josephschorr:
1) Make each non-transactional migration step contain only a single migration operation. This ensures that if the operation fails, we don't end up "halfway" into a migration step. In the future, we'll enforce this via testing.
2) Change how concurrent indexes are created. See: https://www.shayon.dev/post/2024/225/stop-relying-on-if-not-exists-for-concurrent-index-creation-in-postgresql/ for why this is important

### DIFF
--- a/internal/datastore/postgres/migrations/index.go
+++ b/internal/datastore/postgres/migrations/index.go
@@ -1,0 +1,57 @@
+package migrations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/authzed/spicedb/internal/datastore/postgres/common"
+)
+
+const createIndexTemplate = `
+CREATE INDEX CONCURRENTLY 
+	%s
+	ON
+	%s`
+
+const dropIndexTemplate = `
+	DROP INDEX CONCURRENTLY IF EXISTS 
+	%s;
+`
+
+const timeoutMessage = "This typically indicates that your database global statement_timeout needs to be increased and/or spicedb migrate command needs --migration-timeout increased (1h by default)"
+
+// createIndexConcurrently creates an index concurrently, dropping the existing index if it exists to ensure
+// that indexes are not left in a partially constructed state.
+// See: https://www.shayon.dev/post/2024/225/stop-relying-on-if-not-exists-for-concurrent-index-creation-in-postgresql/
+func createIndexConcurrently(ctx context.Context, conn *pgx.Conn, indexName, creationClause string) error {
+	dropIndexSQL := fmt.Sprintf(dropIndexTemplate, indexName)
+	if _, err := conn.Exec(ctx, dropIndexSQL); err != nil {
+		if common.IsQueryCanceledError(err) {
+			return fmt.Errorf(
+				"timed out while trying to drop index %s before recreating it: %w. %s",
+				indexName,
+				err,
+				timeoutMessage,
+			)
+		}
+
+		return fmt.Errorf("failed to drop index %s before creating it: %w", indexName, err)
+	}
+
+	createIndexSQL := fmt.Sprintf(createIndexTemplate, indexName, creationClause)
+	if _, err := conn.Exec(ctx, createIndexSQL); err != nil {
+		if common.IsQueryCanceledError(err) {
+			return fmt.Errorf(
+				"timed out while trying to create index %s: %w. %s",
+				indexName,
+				err,
+				timeoutMessage,
+			)
+		}
+
+		return fmt.Errorf("failed to create index %s: %w", indexName, err)
+	}
+	return nil
+}
--- a/internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
+++ b/internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
@@ -2,20 +2,17 @@ package migrations
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/jackc/pgx/v5"
 )
 
-const addWatchAPIIndexToRelationTupleTable = `CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_watch_index ON relation_tuple (created_xid);`
+const watchAPIIndexToRelationTupleTable = `
+	relation_tuple (created_xid);`
 
 func init() {
 	if err := DatabaseMigrations.Register("add-watch-api-index-to-relation-tuple-table", "add-metadata-to-transaction-table",
 		func(ctx context.Context, conn *pgx.Conn) error {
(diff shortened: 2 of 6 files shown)


vroldanbet (commented):
This approach seems sensible, although we cannot enforce migration designs like this moving forward.

I don't see why this has to be done in different migrations if the column creation is also designed to be idempotent (e.g., checking the catalog to see if the column was already created). Ideally, we have self-contained migrations, but I don't strongly oppose doing this—it would also work.

Example of checking if column exists:
```sql
DO $$
BEGIN
    -- Add column if it doesn't exist
    IF NOT EXISTS (
        SELECT 1 
        FROM information_schema.columns 
        WHERE table_name = 'your_table' 
        AND column_name = 'new_column'
    ) THEN
        EXECUTE 'ALTER TABLE your_table ADD COLUMN new_column type';
    END IF;
END $$;
```

I also wondered why we weren't running those in the same transaction (supported via the migration framework), but it's impossible to run `CREATE INDEX CONCURRENTLY` transactionally with another statement.

There is a [bunch of caveats](https://www.postgresql.org/docs/17/sql-createindex.html#INDEX-RELOPTION-AUTOSUMMARIZE:~:text=is%20off.-,Building%20Indexes%20Concurrently,-Creating%20an%20index) with `CREATE INDEX CONCURRENTLY`.

One issue I see is that the idempotent drop/create index can enter an unrecoverable (and potentially disruptive by sustaining high-load for an extended period?) state if timeouts (migration timeout, statement timeout) are hit.

1. migration runs
2. drop index
3. create index
4. index creation fails due to SQL statement timeout or migration timeout, marked as invalid
5. migration terminates with an error
6. GOTO 1 (operator retries migration)

Coding defensively would be complex, so I suggest we log the error and provide some advice in the log (e.g.,` Please consider increasing your statement timeout and migration timeout using flags...`) and let a human intervene.

I also did some research, and there is no scenario in which a session terminates and index creation continues, so we are guaranteed that it will always end up in an `invalid` state if it has not succeeded (it will never run detached).

josephschorr:
Updated as discussed

internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
```diff
@@ -2,20 +2,17 @@ package migrations
...
 import (
 	"context"
-	"fmt"
 
 	"github.com/jackc/pgx/v5"
 )
 
-const addWatchAPIIndexToRelationTupleTable = `CREATE INDEX CONCURRENTLY IF NOT EXISTS ix_watch_index ON relation_tuple (created_xid);`
+const watchAPIIndexToRelationTupleTable = `
+	ON relation_tuple (created_xid);`
```
vroldanbet, line 10: `ON` seems [required as per the docs](https://www.postgresql.org/docs/current/sql-createindex.html), seems off to force callsite of `createIndexconcurrently` to provide it.

internal/datastore/postgres/migrations/index.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/authzed/spicedb/internal/datastore/postgres/common"
+)
+
+const createIndexTemplate = `
+CREATE INDEX CONCURRENTLY 
+	%s
+	%s`
+
+const dropIndexTemplate = `
+	DROP INDEX CONCURRENTLY IF EXISTS
```
vroldanbet, line 18: So, if the index is partial, it's deleted as expected. However, if, for some reason, the index is completed, but the migration version is not committed to the DB (partial failure), the whole thing will be recreated from scratch. I guess it's a lot of unnecessary work, but safe work to do nevertheless.

I'm trying to imagine a scenario in which this runs again despite the migration being executed—it will unnecessarily drop the index. I understand that `IF EXIST` returns true on invalid and valid indexes.

All that work is spared only if the migration is skipped when completed, which should be the case.

internal/datastore/postgres/migrations/index.go
```diff
@@ -0,0 +1,52 @@
...
+
+// createIndexConcurrently creates an index concurrently, dropping the existing index if it exists to ensure
+// that indexes are not left in a partially constructed state.
+// See: https://www.shayon.dev/post/2024/225/stop-relying-on-if-not-exists-for-concurrent-index-creation-in-postgresql/
+func createIndexConcurrently(ctx context.Context, conn *pgx.Conn, indexName, creationClause string) error {
+	dropIndexSQL := fmt.Sprintf(dropIndexTemplate, indexName)
+	if _, err := conn.Exec(ctx, dropIndexSQL); err != nil {
+		if common.IsQueryCanceledError(err) {
+			return fmt.Errorf(
+				"timed out while trying to drop index %s before recreating it: %w. This typically indicates that your statement_timeout needs to be increased",
```
vroldanbet, line 30:
```suggestion
				"timed out while trying to drop index %s before recreating it: %w. This typically indicates that your database global statement_timeout needs to be increased and/or spicedb migrate command needs --migration-timeout increased (1h by default)",
```

internal/datastore/postgres/migrations/index.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		return fmt.Errorf("failed to drop index %s before creating it: %w", indexName, err)
+	}
+
+	createIndexSQL := fmt.Sprintf(createIndexTemplate, indexName, creationClause)
+	if _, err := conn.Exec(ctx, createIndexSQL); err != nil {
+		if common.IsQueryCanceledError(err) {
+			return fmt.Errorf(
+				"timed out while trying to create index %s: %w. This typically indicates that your statement_timeout needs to be increased",
```
vroldanbet, line 43:
```suggestion
				"timed out while trying to create index %s: %w. This typically indicates that your database global statement_timeout needs to be increased and/or spicedb migrate command needs --migration-timeout increased (1h by default)",
```

vroldanbet (approved):

internal/datastore/postgres/migrations/zz_migration.0020_add_watch_api_index.go
josephschorr: Changed

internal/datastore/postgres/migrations/index.go
josephschorr: > I'm trying to imagine a scenario in which this runs again despite the migration being executed—it will unnecessarily drop the index. I understand that IF EXIST returns true on invalid and valid indexes.

The only scenario where that could happen is if the index was created and then the connection dies right before the  revision is updated in the table. In that scenario recreating the index is I fine in my opinion.

josephschorr:
Updated

vroldanbet (approved):

PR #2198, opened 2025-01-09
Move unlock call to a background context in GC

josephschorr:
This ensures that if the GC operation times out, the unlock still can run

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -80,17 +80,38 @@ func RegisterGCMetrics() error {
 // GarbageCollector represents any datastore that supports external garbage
 // collection.
 type GarbageCollector interface {
+	// HasGCRun returns true if a garbage collection run has been completed.
 	HasGCRun() bool
+
+	// MarkGCCompleted marks that a garbage collection run has been completed.
 	MarkGCCompleted()
+
+	// ResetGCCompleted resets the state of the garbage collection run.
 	ResetGCCompleted()
 
+	// LockForGCRun attempts to acquire a lock for garbage collection. This lock
+	// is typically done at the datastore level, to ensure that no other nodes are
+	// running garbage collection at the same time.
 	LockForGCRun(ctx context.Context) (bool, error)
-	UnlockAfterGCRun(ctx context.Context) error
 
+	// UnlockAfterGCRun releases the lock after a garbage collection run.
+	// NOTE: this method does not take a context, as the context used for the
+	// reset of the GC run can be canceled/timed out and the unlock will still need to happen.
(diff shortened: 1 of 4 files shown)


internal/datastore/common/gc.go
```diff
@@ -196,7 +196,7 @@ func RunGarbageCollection(gc GarbageCollector, window, timeout time.Duration) er
 	}
 
 	defer func() {
-		err := gc.UnlockAfterGCRun(ctx)
+		err := gc.UnlockAfterGCRun(context.Background())
```
tstirrat15, line 199:
```suggestion
		// Using a separate context to run the DB unlock logic ensures that a timeout
		// won't prevent a lock release.
		err := gc.UnlockAfterGCRun(context.Background())
```

tstirrat15 (approved):
Added a comment for my own edification, LGTM

tstirrat15 (approved):
LGTM

internal/datastore/common/gc_test.go
```diff
@@ -39,11 +41,23 @@ func newFakeGC(deleter gcDeleter) fakeGC {
 	}
 }
 
-func (*fakeGC) LockForGCRun(ctx context.Context) (bool, error) {
+func (gc *fakeGC) LockForGCRun(ctx context.Context) (bool, error) {
+	gc.lock.Lock()
+	defer gc.lock.Unlock()
```
ecordell, line 46:
```suggestion
	defer func() {
	  gc.lock.Unlock()
	  wasUnlocked = true
	 }
```

internal/datastore/common/gc.go
```diff
@@ -196,7 +196,9 @@ func RunGarbageCollection(gc GarbageCollector, window, timeout time.Duration) er
 	}
 
 	defer func() {
-		err := gc.UnlockAfterGCRun(ctx)
+		// NOTE: this uses a separate context to run the DB unlock logic, ensuring that
+		// a timeout in the parent context won't prevent a lock release.
+		err := gc.UnlockAfterGCRun(context.Background())
```
ecordell, line 201: wouldn't this be clearer?
```
err := gc.UnlockAfterGCRun()
```
and then the implementations for the datastores can handle it as appropriate (i.e. it may make sense to have an internal timeout and kill the session)

josephschorr: Not sure if it is clearer (as a caller I'd expect a context), but I can document it

josephschorr:
Updated

ecordell (approved):

tstirrat15 (approved):
LGTM

PR #2199, opened 2025-01-09
Switch postgres to use a set and return an error if a duplicate caveat name is given

josephschorr:
This condition should never happen in real code, but the constraint failure error was causing the PG driver to retry over and over, making the test incredibly slow. Instead, we now check on the client side and return a non-retryable error.


tstirrat15 (approved):
LGTM. Only question is whether this would be a hot path, and if the use of a set instead of a slice is a concern. If not, :shipit:

josephschorr:
@tstirrat15 we need to collect the names anyway, so converting from a set -> slice should be fairly low impact

PR #2201, opened 2025-01-10
Remove sleep in stats test unless needed

josephschorr:
The sleep was originally added for CRDB, which requires time for follower reads to catch up. However, the other datastores don't need this, which meant an extra 5s delay in most datastore tests


tstirrat15 (approved):
Yep, makes sense

PR #2202, opened 2025-01-10
Small changes around node IDs and trace IDs

josephschorr:
1) Moves the default node ID to use a hash in case the hostname has anything sensitive in it
2) Moves the trace ID generation to a single method, in case we choose to change its implementation in the future to something less intensive to run

### DIFF
--- a/pkg/middleware/nodeid/nodeid.go
+++ b/pkg/middleware/nodeid/nodeid.go
@@ -2,8 +2,10 @@ package nodeid
 
 import (
 	"context"
+	"fmt"
 	"os"
 
+	"github.com/cespare/xxhash/v2"
 	middleware "github.com/grpc-ecosystem/go-grpc-middleware/v2"
 	"google.golang.org/grpc"
 )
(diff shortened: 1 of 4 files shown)


pkg/middleware/nodeid/nodeid.go
```diff
@@ -40,7 +42,14 @@ func FromContext(ctx context.Context) (string, error) {
 		if err != nil {
 			return "", err
 		}
-		defaultNodeID = spiceDBPrefix + hostname
+
+		// Hash the hostname to get the final default node ID.
+		hasher := xxhash.New()
```
tstirrat15, line 47: Nit: Move this out of the function?

tstirrat15 (approved):
LGTM

pkg/middleware/nodeid/nodeid.go
josephschorr: Since we cache the name globally after computation, I think this is fine

PR #2203, opened 2025-01-10
Delete LookupResources v1, ReachableResources and all helper code

josephschorr:
SpiceDB no longer dispatches to these methods, so we can remove them


internal/dispatch/graph/check_test.go
```diff
@@ -27,7 +27,10 @@ import (
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
-var ONR = tuple.ONR
+var (
+	ONR = tuple.ONR
+	RR  = tuple.RR
```
tstirrat15, line 32: Why was this necessary?

tstirrat15: Ah, I take it this was just the place it made sense to move it within the package?

tstirrat15 (approved):
LGTM

internal/dispatch/graph/check_test.go
josephschorr: It was in one of the deleted files but we still need it here, so I moved it

PR #2205, opened 2025-01-11
Change feature detection for CRDB watch to not require waiting

josephschorr:
Instead, we purposefully give an invalid parameter, which returns an expected error, but only if watch is supported

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -577,24 +577,26 @@ func (cds *crdbDatastore) features(ctx context.Context) (*datastore.Features, er
 		return nil, err
 	}
 
-	// streams don't return at all if they succeed, so the only way to know
-	// it was created successfully is to wait a bit and then cancel
-	streamCtx, cancel := context.WithCancel(ctx)
-	defer cancel()
-	time.AfterFunc(1*time.Second, cancel)
-
-	_ = cds.writePool.ExecFunc(streamCtx, func(ctx context.Context, tag pgconn.CommandTag, err error) error {
-		if err != nil && errors.Is(err, context.Canceled) {
-			features.Watch.Status = datastore.FeatureSupported
-			features.Watch.Reason = ""
-		} else if err != nil {
-			features.Watch.Status = datastore.FeatureUnsupported
-			features.Watch.Reason = fmt.Sprintf("Range feeds must be enabled in CockroachDB and the user must have permission to create them in order to enable the Watch API: %s", err.Error())
+	// Start a changefeed with an invalid value. If we get back an invalid value error (SQLSTATE 22023)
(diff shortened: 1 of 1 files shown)


internal/datastore/crdb/crdb.go
```diff
@@ -577,24 +577,26 @@ func (cds *crdbDatastore) features(ctx context.Context) (*datastore.Features, er
...
-		if err != nil && errors.Is(err, context.Canceled) {
-			features.Watch.Status = datastore.FeatureSupported
-			features.Watch.Reason = ""
-		} else if err != nil {
-			features.Watch.Status = datastore.FeatureUnsupported
-			features.Watch.Reason = fmt.Sprintf("Range feeds must be enabled in CockroachDB and the user must have permission to create them in order to enable the Watch API: %s", err.Error())
+	// Start a changefeed with an invalid value. If we get back an invalid value error (SQLSTATE 22023)
+	// then we know that the datastore supports watch. If we get back any other error, then we know that
+	// the datastore does not support watch emits or there is a permissions issue.
+	_ = cds.writePool.ExecFunc(ctx, func(ctx context.Context, tag pgconn.CommandTag, err error) error {
```
vroldanbet, line 583: The changestream does not run through any connection pools based on a recommendation from CRL; why is it being done that way here?

not sure if this runs in a loop or a certain cadence, but that would cause connections to be terminated (because of the error)

internal/datastore/crdb/crdb.go
```diff
@@ -577,24 +577,26 @@ func (cds *crdbDatastore) features(ctx context.Context) (*datastore.Features, er
...
+			if pgerr.Code == "22023" {
+				features.Watch.Status = datastore.FeatureSupported
+				return nil
+			}
+		}
+
+		features.Watch.Status = datastore.FeatureUnsupported
+		features.Watch.Reason = fmt.Sprintf("Range feeds must be enabled in CockroachDB and the user must have permission to create them in order to enable the Watch API: %s", err.Error())
+		return nil
+	}, fmt.Sprintf(cds.beginChangefeedQuery, cds.schema.RelationshipTableName, head, "-1s"))
```
vroldanbet, line 599: Is `-1` the invalid argument here?

internal/datastore/crdb/crdb.go
josephschorr: Because its going to fail either way; the reason to not use a pool is because it can be a long-running operation

internal/datastore/crdb/crdb.go
josephschorr: Yes

vroldanbet (approved):

internal/datastore/crdb/crdb.go
josephschorr: Its only called on startup, so it should be fine

PR #2206, opened 2025-01-13
add schemaFile to ValidationFile

kartikaysaxena:
Related authzed/zed#453

Added `schemaFile` to ValidationFile which could be used to specify path of the schema.

### DIFF
--- a/pkg/validationfile/fileformat.go
+++ b/pkg/validationfile/fileformat.go
@@ -39,6 +39,9 @@ type ValidationFile struct {
 	// ValidationTuples are the validation tuples, in tuple string syntax.
 	// Deprecated: only for internal use. Use `relationships`.
 	ValidationTuples []string `yaml:"validation_tuples"`
+
+	// Schema file represents the path specified for the schema file
+	SchemaFile string `yaml:"schemaFile"`
 }
 
 // ParseAssertionsBlock parses the given contents as an assertions block.
--- a/pkg/validationfile/fileformat_test.go
+++ b/pkg/validationfile/fileformat_test.go
@@ -101,6 +101,18 @@ validation:
 	}
 }
 
+func TestDecodeValidationFileWithoutSchema(t *testing.T) {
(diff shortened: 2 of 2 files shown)


tstirrat15:
@kartikaysaxena is this all it took for SpiceDB changes? I was worried that a missing `schema:` key would cause the validation file parsing for that key to fail if it wasn't present.

tstirrat15:
The author has validated that this change works as desired over in authzed/zed#455

kartikaysaxena:
> @kartikaysaxena is this all it took for SpiceDB changes? I was worried that a missing `schema:` key would cause the validation file parsing for that key to fail if it wasn't present.

Added a check if schema and schemaFile aren't specified, also, afaik it panics when schema is not present and we try to compile it in zed, updated the checks there too.

tstirrat15:
It looks like there are some lint and validation failures - can you have a look at those?

tstirrat15:
There's still a test failure. I think it's related to the new parsing you implemented - I was thinking of putting that validation up at the level of zed rather than down here in the parser. I think what you had before with just adding the `SchemaFile` key to the struct was probably fine.

kartikaysaxena:
> There's still a test failure. I think it's related to the new parsing you implemented - I was thinking of putting that validation up at the level of zed rather than down here in the parser. I think what you had before with just adding the `SchemaFile` key to the struct was probably fine.

Yes now the parser enforces that there must be either schema or schemafile in the file,  and so the test expects the same, sure would remove that.

tstirrat15:
Yeah, I'm okay with that. We can either drop that validation for now or do it in a different way.

tstirrat15 (approved):
LGTM

tstirrat15:
sorry, one more thing - can you squash the commits? I usually do it with `git rebase -i main` and then reword the first commit and fixup the others. I'll merge it after that.

tstirrat15 (approved):
LGTM

PR #2207, opened 2025-01-15
Implement simpler import syntax

tstirrat15:
## Description
After discussion, we decided to go with an `import "some/path/to/file.zed"` as the syntax and drop the explicit list of imports. This is based on an assumption that users mostly want to decompose a singles schema file into several, and code reuse is not high on the list of priorities.

This changes the parser, compiler, and importer behavior to match.

## Changes
* Update parser behavior
* Update parser tests and remove now-irrelevant tests
* Add parser impl for string consume
* Update compiler, translator, and importer behavior
* Add validation of local import

## Testing


pkg/composableschemadsl/parser/tests/localimport_with_unterminated_string.zed
```diff
@@ -1,4 +1,4 @@
-from .path.to. import user, persona
+import "path/to/user.zed
```
josephschorr, line 1: Add an error case test where the string is not closed

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -84,6 +84,15 @@ func importFile(importContext importContext) (*CompiledSchema, error) {
...
-func constructFilePath(segments []string) string {
-	return path.Join(segments...) + SchemaFileSuffix
+// Take a filepath and ensure that it's local to the current context.
+func validateFilepath(path string) error {
+	// NOTE: This is slightly overly restrictive; it should theoretically be possible
+	// to take a given filepath and figure out whether it's local to the context where
+	// the compiler is being invoked, rather than whether it's local to the source
+	// folder of the current context. The assumption is that that won't matter
+	// right now, and we can fix it if we need to.
+	if !filepath.IsLocal(path) {
```
josephschorr, line 94: we should also add a check for `..` anywhere in the path; if so, we can also quick-error

pkg/composableschemadsl/parser/parser_impl.go
```diff
@@ -198,6 +198,37 @@ func (p *sourceParser) tryConsumeKeyword(keyword string) bool {
 	return true
 }
 
+// consumeString consumes a string token and returns the unwrapped string or adds an error node.
+func (p *sourceParser) consumeString() (string, bool) {
```
josephschorr, line 202: let's name this `consumeStringLiteral`

pkg/composableschemadsl/parser/tests/localimport_with_singlequotes.zed
```diff
@@ -1,4 +1,4 @@
-from .path.to. import user, persona
+import 'path/to/user.zed'
```
josephschorr, line 1: Add one more test which uses ' and " within the *other* kind of quotes

pkg/composableschemadsl/compiler/importer.go
tstirrat15: Like at the parser level?

pkg/composableschemadsl/parser/parser_impl.go
```diff
@@ -198,6 +198,37 @@ func (p *sourceParser) tryConsumeKeyword(keyword string) bool {
...
+	return consumedString, true
+}
+
+// tryConsumeString attempts to consume an expected string token and return the unwrapped string.
+func (p *sourceParser) tryConsumeStringLiteral() (string, bool) {
+	wrappedStringToken, ok := p.tryConsume(lexer.TokenTypeString)
+	if !ok {
+		return "", false
+	}
+	wrappedString := wrappedStringToken.Value
```
josephschorr, line 219: nit: newline before large comments here and above

josephschorr (approved):
LGTM

PR #2212, opened 2025-01-28
Implement partials compilation

tstirrat15:
## Description
This implements partials support in the composable schema language.

## Changes
Gonna annotate.

## Testing
Review. See that tests pass. Suggest any missing tests.

### DIFF
--- a/pkg/composableschemadsl/compiler/compiler.go
+++ b/pkg/composableschemadsl/compiler/compiler.go
@@ -86,31 +86,8 @@ type Option func(*config)
 
 type ObjectPrefixOption func(*config)
 
-type compilationContext struct {
-	// The set of definition names that we've seen as we compile.
-	// If these collide we throw an error.
-	existingNames *mapz.Set[string]
-	// The global set of files we've visited in the import process.
-	// If these collide we short circuit, preventing duplicate imports.
-	globallyVisitedFiles *mapz.Set[string]
-	// The set of files that we've visited on a particular leg of the recursion.
-	// This allows for detection of circular imports.
-	// NOTE: This depends on an assumption that a depth-first search will always
-	// find a cycle, even if we're otherwise marking globally visited nodes.
-	locallyVisitedFiles *mapz.Set[string]
-}
-
 // Compile compilers the input schema into a set of namespace definition protos.
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
-	cctx := compilationContext{
-		existingNames:        mapz.NewSet[string](),
-		globallyVisitedFiles: mapz.NewSet[string](),
-		locallyVisitedFiles:  mapz.NewSet[string](),
-	}
-	return compileImpl(schema, cctx, prefix, opts...)
-}
-
-func compileImpl(schema InputSchema, cctx compilationContext, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
 	cfg := &config{}
 	prefix(cfg) // required option
 
@@ -118,23 +95,30 @@ func compileImpl(schema InputSchema, cctx compilationContext, prefix ObjectPrefi
 		fn(cfg)
 	}
 
-	mapper := newPositionMapper(schema)
-	root := parser.Parse(createAstNode, schema.Source, schema.SchemaString).(*dslNode)
-	errs := root.FindAll(dslshape.NodeTypeError)
-	if len(errs) > 0 {
-		err := errorNodeToError(errs[0], mapper)
+	root, mapper, err := parseSchema(schema)
+	if err != nil {
+		return nil, err
+	}
+
+	// NOTE: import translation is done separately so that partial references
+	// and definitions defined in separate files can correctly resolve.
+	err = translateImports(importResolutionContext{
+		globallyVisitedFiles: mapz.NewSet[string](),
+		locallyVisitedFiles:  mapz.NewSet[string](),
+		sourceFolder:         cfg.sourceFolder,
+	}, root)
+	if err != nil {
 		return nil, err
 	}
 
 	compiled, err := translate(translationContext{
-		objectTypePrefix:     cfg.objectTypePrefix,
-		mapper:               mapper,
-		schemaString:         schema.SchemaString,
-		skipValidate:         cfg.skipValidation,
-		sourceFolder:         cfg.sourceFolder,
-		existingNames:        cctx.existingNames,
-		locallyVisitedFiles:  cctx.locallyVisitedFiles,
-		globallyVisitedFiles: cctx.globallyVisitedFiles,
+		objectTypePrefix:   cfg.objectTypePrefix,
+		mapper:             mapper,
+		schemaString:       schema.SchemaString,
+		skipValidate:       cfg.skipValidation,
+		existingNames:      mapz.NewSet[string](),
+		compiledPartials:   make(map[string][]*core.Relation),
+		unresolvedPartials: mapz.NewMultiMap[string, *dslNode](),
 	}, root)
 	if err != nil {
 		var withNodeError withNodeError
@@ -146,6 +130,17 @@ func compileImpl(schema InputSchema, cctx compilationContext, prefix ObjectPrefi
 	}
 
 	return compiled, nil
+}
+
+func parseSchema(schema InputSchema) (*dslNode, input.PositionMapper, error) {
+	mapper := newPositionMapper(schema)
+	root := parser.Parse(createAstNode, schema.Source, schema.SchemaString).(*dslNode)
+	errs := root.FindAll(dslshape.NodeTypeError)
+	if len(errs) > 0 {
+		err := errorNodeToError(errs[0], mapper)
+		return nil, nil, err
+	}
+	return root, mapper, nil
 }
 
 func errorNodeToError(node *dslNode, mapper input.PositionMapper) error {
--- a/pkg/composableschemadsl/compiler/compiler_test.go
+++ b/pkg/composableschemadsl/compiler/compiler_test.go
@@ -89,6 +89,238 @@ func TestCompile(t *testing.T) {
 			},
 		},
 		{
+			"simple partial",
+			withTenantPrefix,
+			`partial view_partial {
+				relation user: user;
+			}
+
+			definition simple {
+				...view_partial
+			}`,
+			"",
+			[]SchemaDefinition{
+				namespace.Namespace("sometenant/simple",
+					namespace.MustRelation("user", nil,
+						namespace.AllowedRelation("sometenant/user", "..."),
+					),
+				),
+			},
+		},
+		{
+			"more complex partial",
+			withTenantPrefix,
+			`
+			definition user {}
+			definition organization {}
+
+			partial view_partial {
+				relation user: user;
+				permission view = user
+			}
+
+			definition resource {
+				relation organization: organization
+				permission manage = organization
+
+				...view_partial
+			}
+			`,
+			"",
+			[]SchemaDefinition{
+				namespace.Namespace("sometenant/user"),
+				namespace.Namespace("sometenant/organization"),
+				namespace.Namespace("sometenant/resource",
+					namespace.MustRelation("organization", nil,
+						namespace.AllowedRelation("sometenant/organization", "..."),
+					),
+					namespace.MustRelation("manage",
+						namespace.Union(
+							namespace.ComputedUserset("organization"),
+						),
+					),
+					namespace.MustRelation("user", nil,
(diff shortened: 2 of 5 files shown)


pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -86,55 +86,39 @@ type Option func(*config)
...
-	existingNames *mapz.Set[string]
-	// The global set of files we've visited in the import process.
-	// If these collide we short circuit, preventing duplicate imports.
-	globallyVisitedFiles *mapz.Set[string]
-	// The set of files that we've visited on a particular leg of the recursion.
-	// This allows for detection of circular imports.
-	// NOTE: This depends on an assumption that a depth-first search will always
-	// find a cycle, even if we're otherwise marking globally visited nodes.
-	locallyVisitedFiles *mapz.Set[string]
-}
```
tstirrat15, line 101: These were previously values that we passed through the tree, because the import logic invoked compilation. That isn't possible with partials since a partial can be defined in one file and used in another, so among other changes, this PR makes the import logic operate on the DSL tree as an intermediate step.

pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -86,55 +86,39 @@ type Option func(*config)
...
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
-	cctx := compilationContext{
-		existingNames:        mapz.NewSet[string](),
-		globallyVisitedFiles: mapz.NewSet[string](),
-		locallyVisitedFiles:  mapz.NewSet[string](),
-	}
-	return compileImpl(schema, cctx, prefix, opts...)
-}
-
-func compileImpl(schema InputSchema, cctx compilationContext, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
```
tstirrat15, line 113: Flattening these since there's no longer an invocation in the import logic

pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -86,55 +86,39 @@ type Option func(*config)
...
 	}
 
-	compiled, err := translate(translationContext{
-		objectTypePrefix:     cfg.objectTypePrefix,
-		mapper:               mapper,
-		schemaString:         schema.SchemaString,
-		skipValidate:         cfg.skipValidation,
+	// NOTE: import translation is done separately so that partial references
+	// and definitions defined in separate files can correctly resolve.
+	err = translateImports(importTranslationContext{
```
tstirrat15, line 105: Note that the import process doesn't produce a new tree - it operates on the tree started in the root file and recursively and mutatively flattens things on. I'm not entirely sure I like the approach; I'd welcome other suggestions.

pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -148,6 +132,17 @@ func compileImpl(schema InputSchema, cctx compilationContext, prefix ObjectPrefi
...
+func parseSchema(schema InputSchema) (*dslNode, input.PositionMapper, error) {
+	mapper := newPositionMapper(schema)
+	root := parser.Parse(createAstNode, schema.Source, schema.SchemaString).(*dslNode)
+	errs := root.FindAll(dslshape.NodeTypeError)
+	if len(errs) > 0 {
+		err := errorNodeToError(errs[0], mapper)
+		return nil, nil, err
+	}
+	return root, mapper, nil
+}
```
tstirrat15, line 144: I pulled this out into a separate function because the import functionality parses as a part of its work.

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -26,63 +25,18 @@ type CircularImportError struct {
 	filePath string
 }
 
-func importFile(importContext importContext) (*CompiledSchema, error) {
```
tstirrat15, line 29: Lots of logic moved out of here and into `translateImport` because we're no longer recursing here.

pkg/composableschemadsl/compiler/importer.go
```diff
@@ -26,63 +25,18 @@ type CircularImportError struct {
...
-
+func importFile(filePath string) (*dslNode, error) {
 	schemaBytes, err := os.ReadFile(filePath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read schema file: %w", err)
 	}
 	log.Trace().Str("schema", string(schemaBytes)).Str("file", filePath).Msg("read schema from file")
 
-	return compileImpl(InputSchema{
+	parsedSchema, _, err := parseSchema(InputSchema{
```
tstirrat15, line 35: Note that this function now returns a parsed schema instead of a compiled schema, too.

pkg/composableschemadsl/compiler/node.go
```diff
@@ -35,6 +35,15 @@ func (tn *dslNode) Connect(predicate string, other parser.AstNode) {
 	tn.children[predicate].PushBack(other)
 }
 
+// Used to preserve import order when doing import operations on AST
+func (tn *dslNode) ConnectAndHoistMany(predicate string, other *list.List) {
```
tstirrat15, line 39: I welcome bikeshedding on this.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -20,14 +21,13 @@ import (
...
 
 type translationContext struct {
-	objectTypePrefix     *string
-	mapper               input.PositionMapper
-	schemaString         string
-	skipValidate         bool
-	existingNames        *mapz.Set[string]
-	locallyVisitedFiles  *mapz.Set[string]
-	globallyVisitedFiles *mapz.Set[string]
-	sourceFolder         string
```
tstirrat15, line 30: These moved into the import translation context.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -20,14 +21,13 @@ import (
...
+	objectTypePrefix    *string
+	mapper              input.PositionMapper
+	schemaString        string
+	skipValidate        bool
+	existingNames       *mapz.Set[string]
+	compiledPartials    map[string][]*core.Relation
+	partialWaitingRooms *mapz.MultiMap[string, *dslNode]
```
tstirrat15, line 30: These are for the partial compilation process. `partialWaitingRooms` could probably move into the `translatePartials` function if desired.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -92,17 +100,6 @@ func translate(tctx translationContext, root *dslNode) (*CompiledSchema, error)
 			}
 
 			orderedDefinitions = append(orderedDefinitions, def)
-
-		case dslshape.NodeTypeImport:
-			compiled, err := translateImport(tctx, topLevelNode, names)
-			if err != nil {
-				return nil, err
```
tstirrat15, line 99: This logic moved into `translateImports`

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -273,6 +261,64 @@ func translateObjectDefinition(tctx translationContext, defNode *dslNode) (*core
 	return ns, nil
 }
 
+// NOTE: this function and the translatePartial variation differ in how they treat a missing reference.
+// This function treats that as an error state, since all partials should be resolved by that point;
+// the partial variant treats that as a skip.
+func translateDefinitionRelationsAndPermissions(tctx translationContext, astNode *dslNode) ([]*core.Relation, error) {
```
tstirrat15, line 267: I imagine there's some deduplication to be done between this and the function below, but the point at which they differ is stuck in the middle. I'd be fine leaving this as a target for future refactoring.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -696,27 +742,167 @@ func addWithCaveats(tctx translationContext, typeRefNode *dslNode, ref *core.All
...
+	// This allows for detection of circular imports.
+	// NOTE: This depends on an assumption that a depth-first search will always
+	// find a cycle, even if we're otherwise marking globally visited nodes.
+	locallyVisitedFiles *mapz.Set[string]
+	sourceFolder        string
+}
+
+// Takes a parsed schema and recursively translates import syntax and replaces
+// import nodes with parsed nodes from the target files
+func translateImports(itctx importTranslationContext, root *dslNode) error {
```
tstirrat15, line 759: This is now where the import logic is happening.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -696,27 +742,167 @@ func addWithCaveats(tctx translationContext, typeRefNode *dslNode, ref *core.All
...
+		if topLevelNode.GetType() == dslshape.NodeTypePartial {
+			err := translatePartial(tctx, topLevelNode)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	if tctx.partialWaitingRooms.Len() != 0 {
+		return fmt.Errorf("could not resolve partials: [%s]. this may indicate a circular reference", strings.Join(tctx.partialWaitingRooms.Keys(), ", "))
 	}
```
tstirrat15, line 844: This engenders an assumption that the algorithm using this approach is correct. So far it looks correct, but I don't know that I could prove it right now.

tstirrat15 (commented):
See annotations

pkg/composableschemadsl/compiler/compiler_test.go
```diff
@@ -88,6 +88,182 @@ func TestCompile(t *testing.T) {
...
+			`,
+			"could not resolve partials",
+			[]SchemaDefinition{},
+		},
+		{
+			"definition reference to nonexistent partial",
+			withTenantPrefix,
+			`
+			definition simple {
+				...some_partial
```
josephschorr, line 261: add some tests that have concrete impls next to the partials

pkg/composableschemadsl/compiler/node.go
josephschorr, line 39: does the order really matter?

pkg/composableschemadsl/compiler/translator.go
josephschorr: why is it called a waiting room?

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -58,6 +58,14 @@ func translate(tctx translationContext, root *dslNode) (*CompiledSchema, error)
 	// as we do our walk
 	names := tctx.existingNames.Copy()
 
+	// Do an initial pass to translate partials and add them to the
+	// translation context. This ensures that they're available for
+	// subsequent reference in definition compilation.
+	err := translatePartials(tctx, root)
```
josephschorr, line 64: call this `collectPartials`, since it isn't actually translating them into the root of the tree

pkg/composableschemadsl/compiler/translator.go
josephschorr: pass a flag and change behavior on that bool

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -696,27 +742,169 @@ func addWithCaveats(tctx translationContext, typeRefNode *dslNode, ref *core.All
 	return nil
 }
 
-func translateImport(tctx translationContext, importNode *dslNode, names *mapz.Set[string]) (*CompiledSchema, error) {
-	path, err := importNode.GetString(dslshape.NodeImportPredicatePath)
-	if err != nil {
-		return nil, err
+type importTranslationContext struct {
```
josephschorr, line 745: maybe `importResolutionContext`

pkg/composableschemadsl/compiler/translator.go
tstirrat15: It was a name i reached for. The idea is that the values of a given key are the partials whose compilation depends on the compilation of the partial named in the key, and my brain went to "waiting room" for that concept. I can change that to "unresolvedPartials" or something if that's preferable.

pkg/composableschemadsl/compiler/node.go
tstirrat15: It doesn't, but it means that tests pass without regeneration (which makes me feel better about not having missed something) and that the generated output matches my mental model of imports (in terms of imports being replaced by the contents of the referenced files).

josephschorr: Okay. Maybe name "ConnectAtFront"?

tstirrat15: I'd be more keen to unwind it in a later refactor if we want to drop it as an optimization

josephschorr: ok

josephschorr (approved):
LGTM

PR #2214, opened 2025-01-28
Fixes for replicated query iterator

josephschorr:


internal/datastore/proxy/strictreplicated.go
```diff
@@ -305,32 +108,96 @@ func (rr *strictReadReplicatedReader) LookupCaveatsWithNames(ctx context.Context
...
+				if !yield(tuple.Relationship{}, err) {
+					return
+				}
+			}
+
+			isFirstResult = false
+			if !yield(result, nil) {
+				return
+			}
+		}
```
tstirrat15, line 163: Why was this additional layer necessary?

tstirrat15 (dismissed):
LGTM, had question

internal/datastore/proxy/strictreplicated.go
josephschorr: Yes; we need to catch it here as well

internal/datastore/proxy/strictreplicated.go
```diff
@@ -305,32 +108,96 @@ func (rr *strictReadReplicatedReader) LookupCaveatsWithNames(ctx context.Context
...
+
+				if !yield(tuple.Relationship{}, err) {
+					return
+				}
+			}
+
+			isFirstResult = false
+			if !yield(result, nil) {
+				return
+			}
```
vroldanbet, line 162: no code coverage here - can you please also add an assertion to the test that checks the happy path? (revision IS in replica)

internal/datastore/proxy/replicated.go
```diff
@@ -305,32 +306,83 @@ func (rr *strictReadReplicatedReader) LookupCaveatsWithNames(ctx context.Context
...
+	it, err := handler(sr)(ctx, filter, options...)
+
+	// Check for a RevisionUnavailableError, which indicates the replica does not contain the requested
+	// revision. In this case, use the primary instead. This may not be returned on this call from
+	// wrapped datastores that defer the actual query until the iterator is used.
+	if err != nil {
+		if errors.As(err, &common.RevisionUnavailableError{}) {
+			log.Trace().Str("revision", rr.rev.String()).Msg("replica does not contain the requested revision, using primary")
+			return handler(rr.primary.SnapshotReader(rr.rev))(ctx, filter, options...)
+		}
```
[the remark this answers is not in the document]
josephschorr: It could be if the underlying implementation changes; since that is not guaranteed, we still need to check for it here to future proof this

internal/datastore/proxy/strictreplicated.go
josephschorr: Added

vroldanbet (dismissed):
Thanks for addressing the feedback, LGTM

vroldanbet (approved):

PR #2216, opened 2025-01-29
Have LookupSubjects return Unimplemented if a cursor is used

josephschorr:
This will give user feedback until support is added


tstirrat15 (approved):
LGTM

PR #2219, opened 2025-01-30
Change strict reading PG to only return rows when valid

josephschorr:
This is necessary for the error and retry logic to work in the strict read proxy

### DIFF
--- a/internal/datastore/postgres/postgres_shared_test.go
+++ b/internal/datastore/postgres/postgres_shared_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/authzed/spicedb/internal/datastore/common"
 	pgcommon "github.com/authzed/spicedb/internal/datastore/postgres/common"
 	pgversion "github.com/authzed/spicedb/internal/datastore/postgres/version"
+	"github.com/authzed/spicedb/internal/datastore/proxy"
 	"github.com/authzed/spicedb/internal/testfixtures"
 	testdatastore "github.com/authzed/spicedb/internal/testserver/datastore"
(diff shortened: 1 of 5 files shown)


internal/datastore/proxy/checkingreplicated_test.go
```diff
@@ -79,14 +79,14 @@ func TestReplicatedReaderReturnsExpectedError(t *testing.T) {
 }
 
 type fakeDatastore struct {
-	isPrimary bool
-	revision  datastore.Revision
+	state    string
```
vroldanbet, line 82: why the change?

internal/datastore/postgres/postgres_shared_test.go
```diff
@@ -1568,6 +1579,60 @@ func LockingTest(t *testing.T, ds datastore.Datastore, ds2 datastore.Datastore)
...
+		OptionalResourceType: "resource",
+	})
+	require.NoError(err)
+
+	found, err := datastore.IteratorToSlice(it)
+	require.NoError(err)
+	require.NotEmpty(found)
+
+	// Perform a read at a manually constructed revision beyond head, which should fallback to the primary.
+	badRev := postgresRevision{
```
vroldanbet, line 1615: can you test the replica and make sure it does return error with this revision?

vroldanbet (approved):

internal/datastore/proxy/checkingreplicated_test.go
josephschorr: So I can have more test cases

Issue #1995, opened 2024-07-24
Schema Change Events via Watch API


saltheman:
### Problem Statement

Today we get events about Relationships via the SpiceDB Watch API. It will be nice to get a stream of events
when the schema changes for example new relation added/removed, new permission added/remove, new object definition added removed, a relation subject changed etc.

josephschorr:
This is already supported on the datastore side in most datastores, so we'd need to:

1) come up with an updated API definition
2) implement in the remaining datastores (I believe just MySQL and memdb)

vroldanbet:
@josephschorr `MemDB` is [already implemented](https://github.com/authzed/spicedb/pull/1720)

miparnisari:
Based on https://github.com/authzed/spicedb/blob/main/pkg/datastore/datastore.go#L47-L60 it looks like we can only emit events for:
- type added
- type changed
- type deleted
- caveat added
- caveat changed
- caveat deleted

So, there wouldn't be events for changes to relations within types. Is this OK @josephschorr?

josephschorr:
@miparnisari The changes for relations are found in `ChangedDefinitions`; you'll need to run the SchemaDiff code to determine the changes within a changed definition.

I recommend we start with the API-side for the new event in Watch, so we can determine what the various diff event types are

miparnisari:
@saltheman is there a scenario where you _only_ want to get updates to the schema, and no updates to the relationships?

miparnisari:
Closing as this was introduced in [v1.42.0](https://github.com/authzed/spicedb/releases/tag/v1.42.0).

PR #2221, opened 2025-01-31
feat: implement schema watch for MySQL datastore driver

kartikaysaxena:
Part of #1995 similar to how it was done for PG and other datastores


kartikaysaxena:
closing in favour of #2223 for better indexing support first

josephschorr:
@kartikaysaxena Why would this be impacted by that PR?

PR #2222, opened 2025-01-31
Remove parallel test to reduce flakiness

josephschorr:
Also adds logs for when the test fails


tstirrat15 (approved):
LGTM! I like the log change.

PR #2224, opened 2025-01-31
feat: schema watch support for mysql driver

kartikaysaxena:
implements schema watch for mysql datastore

### DIFF
--- a/internal/datastore/mysql/watch.go
+++ b/internal/datastore/mysql/watch.go
@@ -3,11 +3,13 @@ package mysql
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/internal/datastore/revisions"
 	"github.com/authzed/spicedb/pkg/datastore"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/tuple"
 
 	sq "github.com/Masterminds/squirrel"
@@ -28,12 +30,6 @@ func (mds *Datastore) Watch(ctx context.Context, afterRevisionRaw datastore.Revi
 
 	updates := make(chan *datastore.RevisionChanges, watchBufferLength)
 	errs := make(chan error, 1)
-
-	if options.Content&datastore.WatchSchema == datastore.WatchSchema {
-		close(updates)
-		errs <- errors.New("schema watch unsupported in MySQL")
-		return updates, errs
-	}
 
 	if options.EmissionStrategy == datastore.EmitImmediatelyStrategy {
 		close(updates)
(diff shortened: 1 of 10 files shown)


kartikaysaxena:
@josephschorr I thought this PR would need refactoring after the changes to base of datastore driver, after looking closely it seems that this could go ahead as well, my bad.

internal/datastore/mysql/watch.go
```diff
@@ -177,7 +173,30 @@ func (mds *Datastore) loadChanges(
...
+	// Load namespace changes for the revision range.
+	if options.Content&datastore.WatchSchema == datastore.WatchSchema {
+		if err := mds.loadNamespaceChanges(ctx, afterRevision, newRevision, stagedChanges); err != nil {
+			return nil, 0, err
+		}
+	}
+
+	// Load caveat changes for the revision range.
+	if options.Content&datastore.WatchSchema == datastore.WatchSchema {
+		if err := mds.loadCaveatChanges(ctx, afterRevision, newRevision, stagedChanges); err != nil {
```
josephschorr, line 189: Put this under the if statement above

internal/datastore/mysql/watch.go
```diff
@@ -265,7 +284,130 @@ func (mds *Datastore) loadChanges(
...
+		)
+		if err != nil {
+			return
+		}
+		loaded := &core.NamespaceDefinition{}
+		if err := loaded.UnmarshalVT(config); err != nil {
+			return fmt.Errorf("unable to parse changed namespace: %w", err)
+		}
+
+		if createdTxn > afterRevision && createdTxn <= newRevision {
```
josephschorr, line 332: why the filter here if we're doing so in the SQL query?

internal/datastore/mysql/watch.go
```diff
@@ -265,7 +284,130 @@ func (mds *Datastore) loadChanges(
...
+	if err = rows.Err(); err != nil {
+		return fmt.Errorf("unable to load changes: %w", err)
+	}
+
+	return
+}
+
+func (mds *Datastore) loadCaveatChanges(ctx context.Context, afterRevision uint64, newRevision uint64, stagedChanges *common.Changes[revisions.TransactionIDRevision, uint64]) (err error) {
+	sql, args, err := mds.QueryChangedCaveatsQuery.Where(sq.Or{
+		sq.And{
```
josephschorr, line 353: We should be able to combine these methods

internal/datastore/mysql/watch.go
kartikaysaxena: The query can return a wider net of rows in the watch window, hence a double check, was implemented earlier for relationships too, also tested without the if statement, datastore tests tend to break in that case.

internal/datastore/mysql/watch.go
kartikaysaxena: done, should I do same for loadRelationships and combine that too? (right now `loadSchemaChanges` is only called when schema watch is enabled)

internal/datastore/mysql/watch.go
```diff
@@ -265,7 +277,95 @@ func (mds *Datastore) loadChanges(
...
+				err = rows.Scan(&config, &createdTxn, &deletedTxn)
+				if err != nil {
+					return fmt.Errorf("unable to parse changed namespace: %w", err)
+				}
+				def := &core.NamespaceDefinition{}
+				if err := def.UnmarshalVT(config); err != nil {
+					return fmt.Errorf("unable to parse changed namespace: %w", err)
+				}
+				loaded = def
+				objName = def.Name
```
josephschorr, line 334: don't need to set objName, since its defined on the SchemaDefinition

internal/datastore/mysql/watch.go
josephschorr: Ah

internal/datastore/mysql/watch.go
kartikaysaxena: done

miparnisari:
Based on this comment https://github.com/authzed/spicedb/pull/2924#issuecomment-4006820689 I'm closing this PR, since this feature will be deprecated soon anyways. Thanks @kartikaysaxena for the contribution and sorry it didn't get a timely merge.

PR #2226, opened 2025-02-02
Change default version for CRDB testing to latest

josephschorr:
Also adds ability to test at different versions


vroldanbet (commented):
changes LGTM, do we want to also add a new GHA matrix to test with multiple CRDB versions in CI?

PR #2229, opened 2025-02-03
Add metrics for usage of the replication proxies in the datastore

josephschorr:
Should help provide some insights into when the read replicas are being used vs the primary

### DIFF
--- a/internal/datastore/proxy/checkingreplicated.go
+++ b/internal/datastore/proxy/checkingreplicated.go
@@ -6,10 +6,29 @@ import (
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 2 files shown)


bison (commented):
Do we have some kind of name or stable ID for a replica? If there are multiple, can we label the counters with it?

josephschorr:
> Do we have some kind of name or stable ID for a replica? If there are multiple, can we label the counters with it?

Currently, no. We'd have to extract it from the URI or the "position"; we could label them `replica_0`, `replica_1`, or we could try to pull some info from the connection URI

vroldanbet (approved):

PR #2231, opened 2025-02-03
Add a metric tracking the selected replica

josephschorr:
Allows for understanding which replica(s) are being used

### DIFF
--- a/internal/datastore/common/url.go
+++ b/internal/datastore/common/url.go
@@ -1,0 +1,19 @@
+package common
+
+import (
+	"errors"
+	"net/url"
+)
+
+// MetricsIDFromURL extracts the metrics ID from a given datastore URL.
+func MetricsIDFromURL(dsURL string) (string, error) {
+	if dsURL == "" {
+		return "", errors.New("datastore URL is empty")
+	}
+
+	u, err := url.Parse(dsURL)
+	if err != nil {
+		return "", errors.New("could not parse datastore URL")
+	}
+	return u.Host + u.Path, nil
+}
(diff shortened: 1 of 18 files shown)


internal/datastore/common/url.go
```diff
@@ -0,0 +1,19 @@
...
+func MetricsIDFromURL(dsURL string) (string, error) {
+	if dsURL == "" {
+		return "", errors.New("datastore URL is empty")
+	}
+
+	u, err := url.Parse(dsURL)
+	if err != nil {
+		return "", errors.New("could not parse datastore URL")
+	}
+	return u.Host + u.Path, nil
```
vroldanbet, line 18: note that returning host+path will lead to explosion of cardinality when having many spiced clusters in the same physical database. I do see the value in disambiguating telemetry across spicedb clusters, just something to note.

internal/datastore/memdb/memdb.go
```diff
@@ -100,6 +100,10 @@ type snapshot struct {
 	db       *memdb.MemDB
 }
 
+func (mdb *memdbDatastore) MetricsID() (string, error) {
+	return "memdb-" + mdb.uniqueID, nil
```
vroldanbet, line 104: this almost certainly cause explosion in cadinality for these metrics. E.g. when MemDB SpiceDB is embedded in another process, each restart will bring different uniqueID. I'd personally just have it as `memdb` - the scenario of multiple SpiceDB processes in memory is less likely, and the explosion in cardinality can have severe cost implications

josephschorr: Changed

josephschorr:
Updated

vroldanbet (approved):

PR #2232, opened 2025-02-03
Remove the reference for changes being emitted by Watch

josephschorr:
A change should always be found in the channel, so there is little benefit to it being a reference


tstirrat15 (approved):
For my own edification: the idea is that this prevents nil pointer derefs and also means that if the struct is small enough and the go runtime deems it appropriate, it may allocate it on the stack instead of the heap.

PR #2235, opened 2025-02-04
magefile: added combined test coverage support for `mage test:all`

### LINKED ISSUE #1348: `mage test:all` should produce a combined code coverage result
If we add the `-coverprofile` flag to ever `go test` func in `magefiles/test.go`, we can aggregate them all into a single coverage file in `All()`.

This'll be useful to track (perfect coverage is a non-goal of the project), but is also [required for submitting to awesome-go](https://github.com/avelino/awesome-go/pull/3771)

kartikaysaxena:
Fixes #1348 `mage test:all` now generates combined coverprofile aggregated into `coverage.txt`


magefiles/util.go
```diff
@@ -21,12 +21,24 @@ func goTest(path string, args ...string) error {
 
 // run go test in a directory
 func goDirTest(dir string, path string, args ...string) error {
-	testArgs := append([]string{"test", "-failfast", "-count=1"}, args...)
+	testArgs := append([]string{
+		"test",
+		"-covermode=atomic",
+		"-coverprofile=coverage.txt",
```
josephschorr, line 27: Won't this overwrite the coverage.txt file when each individual test suite runs? Can you try running `mage test:all` and ensure it combines the coverage?

kartikaysaxena: Changed the approach, now each test writes to a different file and removes those files in the end, `go tool cover -html=coverage.txt` now shows 
Screenshot 2025-02-06 at 12 25 43 AM

magefiles/util.go
```diff
@@ -210,3 +223,45 @@ func run(dir string, env map[string]string, stdout, stderr io.Writer, cmd string
 	err = c.Run()
 	return sh.CmdRan(err), sh.ExitStatus(err), err
 }
+
+func sanitizePath(path string) string {
```
josephschorr, line 227: I'd recommend hashing the path here

kartikaysaxena: Done

vroldanbet (commented):
Seems like the PR enables coverage by default on `all`, whereas before it was opt in. Do we have concerns that would add overhead on the inner loop? Or is the overhead negligible?

kartikaysaxena:
> Seems like the PR enables coverage by default on `all`, whereas before it was opt in. Do we have concerns that would add overhead on the inner loop? Or is the overhead negligible?

I used `time mage test:all` to check and the overhead was costing over a minute, followed the pattern for other targets for generating coverage reports which was earlier in `UnitCover`, now we have a dedicated `test:allcover` for this, is this correct? Would love the feedback and any improvements further if necessary

magefiles/util.go
```diff
@@ -210,3 +221,40 @@ func run(dir string, env map[string]string, stdout, stderr io.Writer, cmd string
...
+
+	args := []string{"run", "github.com/wadey/gocovmerge@latest"}
+	args = append(args, files...)
+
+	err = RunSh(goCmdForTests(), WithV(), WithStdout(f))(args...)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
```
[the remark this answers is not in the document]
kartikaysaxena: was for deleting individual coverage files after merging them. reverting

magefiles/util.go
```diff
@@ -210,3 +221,40 @@ func run(dir string, env map[string]string, stdout, stderr io.Writer, cmd string
...
+		return fmt.Errorf("no coverage files found")
+	}
+
+	f, err := os.Create("coverage.txt")
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	args := []string{"run", "github.com/wadey/gocovmerge@latest"}
```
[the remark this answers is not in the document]
kartikaysaxena: it seems we need to build a binary for this for specifying the `-cover` flag before the tests which would generate the files(tests should be carried on the same binary?)

kartikaysaxena:
Utilised uuid instead, also passed the bool through a context to `goDirTest` and `goDirTestWithEnv`, now works like `mage test:all -cover=true`

magefiles/util.go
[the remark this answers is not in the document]
vroldanbet: And is there a reason we need to stick to the text cover files?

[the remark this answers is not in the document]
kartikaysaxena: `go test -coverprofile` generates text-based files (current approach), raw coverage files are generated by testing the binary built using `go build -cover`, does the approach need a overhaul?

magefiles/util.go
```diff
@@ -3,30 +3,59 @@
...
+	}, args...)
+	if cover, _ := ctx.Value("cover").(bool); cover {
+		if err := os.MkdirAll("coverage", 0o755); err != nil {
+			return fmt.Errorf("failed to create coverage directory: %w", err)
+		}
+		testArgs = append(testArgs, []string{
+			"-covermode=atomic",
+			fmt.Sprintf("-coverprofile=coverage-%s.txt", uuid.New().String()),
+		}...)
+	}
```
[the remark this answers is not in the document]
kartikaysaxena: done with `testWithArgs`

magefiles/test.go
```diff
@@ -19,51 +20,62 @@ var emptyEnv map[string]string
...
+	}
+	ctx = context.WithValue(ctx, "cover", cover)
+	mg.CtxDeps(ctx, t.Unit, t.Integration, t.Steelthread, t.Image, t.Analyzers,
 		ds.Crdb, ds.Postgres, ds.Spanner, ds.Mysql,
 		c.Crdb, c.Spanner, c.Postgres, c.Mysql)
+
 	return nil
 }
 
+func (t Test) Combine() error {
```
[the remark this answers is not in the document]
kartikaysaxena: done

magefiles/test.go
```diff
@@ -19,51 +22,56 @@ var emptyEnv map[string]string
 func (t Test) All() error {
 	ds := Testds{}
 	c := Testcons{}
-	mg.Deps(t.Unit, t.Integration, t.Steelthread, t.Image, t.Analyzers,
+	cover := parseCommandLineFlags()
+	ctxMyKey := ctxkey.New[bool]()
+	ctx := ctxMyKey.WithValue(context.Background(), cover)
```
tstirrat15, line 27: Does it work to declare `ctxMyKey` twice? I'd expect that that would give you two different identities of keys, which wouldn't allow you to pull values out of context. The idea is that you'd give the key a name that means something and declare it once at the top of the file. There's an example in the [ctxkey examples dir](https://github.com/authzed/ctxkey/blob/main/examples/http/main.go#L23-L24).

Would you please declare this key once at the top of the module with a meaningful name and then reuse it?

tstirrat15 (commented):
See comment; once that's through it looks good to me.

magefiles/test.go
```diff
@@ -19,51 +22,56 @@ var emptyEnv map[string]string
 func (t Test) All() error {
 	ds := Testds{}
 	c := Testcons{}
-	mg.Deps(t.Unit, t.Integration, t.Steelthread, t.Image, t.Analyzers,
+	cover := parseCommandLineFlags()
+	ctxMyKey := ctxkey.New[bool]()
```
vroldanbet, line 26: This should be defined as a global var, a key to a specific value.

magefiles/test.go
kartikaysaxena: was doing that to avoid a global var, seems like it should be the way to go for this.

magefiles/util.go
```diff
@@ -3,33 +3,60 @@
...
 	return RunSh(goCmdForTests(), WithV(), WithDir(dir), WithEnv(env), WithArgs(testArgs...))(path)
 }
 
+func testWithArgs(ctx context.Context, args ...string) ([]string, error) {
+	testArgs := append([]string{
+		"test",
+		"-failfast",
+		"-count=1",
+	}, args...)
+	if cover := ctxMyKey.MustValue(ctx); cover {
```
vroldanbet, line 48: `Must` will panic - it's the naming pattern for methods used in tests so you don't have to check errors. Please use the non panicking variant.

vroldanbet (changes requested):

magefiles/util.go
kartikaysaxena: used `ctxMyKey.Value`

magefiles/test.go
```diff
@@ -9,22 +9,23 @@ import (
...
 
 // All Runs all test suites
-func (t Test) All() error {
+func (t Test) All(ctx context.Context) error {
 	ds := Testds{}
 	c := Testcons{}
 	cover := parseCommandLineFlags()
-	ctxMyKey := ctxkey.New[bool]()
-	ctx := ctxMyKey.WithValue(context.Background(), cover)
+	ctx = ctxMyKey.WithValue(context.Background(), cover)
```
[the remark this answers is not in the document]
kartikaysaxena: yeah overlooked that, thanks for pointing out!

vroldanbet (approved):

PR #2242, opened 2025-02-14
import: improve err msg

kartikaysaxena:
Fixes authzed/zed#463, error now explicitly reports that there is something wrong with the import

### DIFF
--- a/pkg/composableschemadsl/compiler/importer.go
+++ b/pkg/composableschemadsl/compiler/importer.go
@@ -19,7 +19,7 @@ type CircularImportError struct {
 func importFile(filePath string) (*dslNode, error) {
 	schemaBytes, err := os.ReadFile(filePath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read schema file: %w", err)
+		return nil, fmt.Errorf("failed to read import in schema file: %w", err)
 	}
 	log.Trace().Str("schema", string(schemaBytes)).Str("file", filePath).Msg("read schema from file")
 


pkg/composableschemadsl/compiler/importer.go
```diff
@@ -19,7 +19,7 @@ type CircularImportError struct {
 func importFile(filePath string) (*dslNode, error) {
 	schemaBytes, err := os.ReadFile(filePath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read schema file: %w", err)
+		return nil, fmt.Errorf("failed to read import in schema file: %w", err)
```
josephschorr, line 22: This is good, but we should also decorate it with the source location information from the source file(s) in which the import was used. It would require tracking them in the import context, but it should be doable.

Something like: `failed to read import in schema file: %w; referenced at [...]`

tstirrat15:
I think the idea of this was that it'd include the information about where in the source it happened:

> Ideally this would say something about the import statement on line 1 of root.zed, etc.

I think it'll require turning this error: https://github.com/authzed/spicedb/blob/main/pkg/composableschemadsl/compiler/importer.go#L21-L23

into a WithSourceError like this one: https://github.com/authzed/spicedb/blob/main/pkg/composableschemadsl/compiler/translator.go#L94-L96

and then pulling that change into zed.

pkg/composableschemadsl/compiler/importer.go
kartikaysaxena: done, tested with zed, it returns `terminated with errors error="parse error in `simple.zed`, line 2, column 1: failed to read import in schema file"` on a faulty import

kartikaysaxena:
Now `zed preview schema compile` gives the following error on a faulty import
```
1:33AM ERR terminated with errors error="parse error in `simple.zed`, line 2, column 1: failed to read import in schema file"
```
does this look good?

tstirrat15 (approved):
This looks good to me now.

PR #2244, opened 2025-02-17
Improve handling of watch errors

josephschorr:
Adds more cases where retryable error is returned and adds a too-stale error for CRDB

A followup will add too-stale for the other datastores


vroldanbet (approved):
LGTM

PR #2245, opened 2025-02-18
Support follower read delay flag with Postgres and MySQL datastores

ecordell:
In CRDB and Spanner, follower read delay is used to artificially pick an older timestamp that should be synchronized to the follower nodes.

We have a similar situation in Postgres and MySQL with read replicas, where we'd like to pick a quantized revision that we think is likely to have been replicated, but prior to this PR, that flag was not used in those datastores.

To use this with postgres or mysql, set `--datastore-follower-read-delay-duration` to any non-default value (`4.8s`) (The postgres and mysql default is actually 0s even though the flag default is 4.8s)

### DIFF
--- a/internal/datastore/postgres/options.go
+++ b/internal/datastore/postgres/options.go
@@ -19,6 +19,7 @@ type postgresOptions struct {
 	watchBufferLength       uint16
 	watchBufferWriteTimeout time.Duration
 	revisionQuantization    time.Duration
+	followerReadDelay       time.Duration
 	gcWindow                time.Duration
 	gcInterval              time.Duration
 	gcMaxOperationTime      time.Duration
@@ -74,6 +75,8 @@ const (
 	defaultColumnOptimizationOption          = common.ColumnOptimizationOptionNone
 	defaultIncludeQueryParametersInTraces    = false
 	defaultExpirationDisabled                = false
+	// no follower delay by default, it should only be set if using read replicas
+	defaultFollowerReadDelay = 0
 )
 
 // Option provides the facility to configure how clients within the
@@ -99,6 +102,7 @@ func generateConfig(options []Option) (postgresOptions, error) {
 		columnOptimizationOption:       defaultColumnOptimizationOption,
 		includeQueryParametersInTraces: defaultIncludeQueryParametersInTraces,
 		expirationDisabled:             defaultExpirationDisabled,
+		followerReadDelay:              defaultFollowerReadDelay,
 	}
 
 	for _, option := range options {
@@ -279,6 +283,14 @@ func WatchBufferWriteTimeout(watchBufferWriteTimeout time.Duration) Option {
 // This value defaults to 5 seconds.
 func RevisionQuantization(quantization time.Duration) Option {
 	return func(po *postgresOptions) { po.revisionQuantization = quantization }
+}
+
+// FollowerReadDelay is the amount of time to round down the current time when
+// reading from a read replica is expected.
+//
+// This value defaults to 0 seconds.
+func FollowerReadDelay(delay time.Duration) Option {
+	return func(po *postgresOptions) { po.followerReadDelay = delay }
 }
 
 // MaxRevisionStalenessPercent is the amount of time, expressed as a percentage of
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -25,7 +25,10 @@ import (
 
 type engineBuilderFunc func(ctx context.Context, options Config) (datastore.Datastore, error)
 
-const MaxReplicaCount = 16
+const (
+	MaxReplicaCount          = 16
+	DefaultFollowerReadDelay = 4_800 * time.Millisecond
+)
 
 const (
 	MemoryEngine    = "memory"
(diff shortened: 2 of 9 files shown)


vroldanbet (dismissed):

internal/datastore/postgres/options.go
```diff
@@ -74,6 +75,7 @@ const (
 	defaultColumnOptimizationOption          = common.ColumnOptimizationOptionNone
 	defaultIncludeQueryParametersInTraces    = false
 	defaultExpirationDisabled                = false
+	defaultFollowerReadDelay                 = 0
```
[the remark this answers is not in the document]
ecordell: it's not just that - if you don't have any read replicas you definitely don't want a follower read delay. will add a note.

ecordell:
Update here: I tested this with an aurora global DB replicating from us-east-1 to ap-east-1. With a quantization of 5s and a follower read delay of 2s, the ap-east-1 spicedbcluster was always able to find the revision replicated.

pkg/cmd/datastore/datastore.go
```diff
@@ -25,7 +25,10 @@ import (
 
 type engineBuilderFunc func(ctx context.Context, options Config) (datastore.Datastore, error)
 
-const MaxReplicaCount = 16
+const (
+	MaxReplicaCount          = 16
```
josephschorr, line 29: Where is `MaxReplicaCount` used?

ecordell: I didn't add it (you did) but it's used when validating the replica config for a pg instance.

josephschorr: Oh, sorry. I thought it was new. My bad.

pkg/cmd/datastore/datastore.go
```diff
@@ -336,6 +339,12 @@ func NewDatastore(ctx context.Context, options ...ConfigOption) (datastore.Datas
 		o(opts)
 	}
 
+	if (opts.Engine == PostgresEngine || opts.Engine == MySQLEngine) && opts.FollowerReadDelay == DefaultFollowerReadDelay {
+		// Set the default follower read delay for postgres and mysql to 0 -
+		// this should only be set if read replicas are used.
```
josephschorr, line 344: Can we add a check so it is only set if a read replica is enabled for PG/MySQL?

ecordell: We need to set the follower read delay to 0 if read replicas _aren't_ enabled, which is what this is doing.

I suppose we could also check that if a delay is set and you're not using a replica, we error, but I don't know that there won't be reasons to use a read delay even without an explicit replica (i.e. maybe you're talking to a primary or a replica through a proxy that hides them?)

josephschorr: > but I don't know that there won't be reasons to use a read delay even without an explicit replica

I'd prefer if we raised the error; we can always remove it later, but if you're talking to a replica through a proxy, SpiceDB at minimum should be in read only mode

ecordell: PgBouncer for example can route writes to a primary and reads to a replica. I'm pretty sure the "cluster" endpoints in aurora also work similarly, and can distribute the load to multiple backends (only one of which may be a primary).

josephschorr: > PgBouncer for example can route writes to a primary and reads to a replica

And if the system is not configured with SpiceDB knowing those reads are going to a replica, it will fail to read the latest revision and be incorrect.

I'll defer on this issue but it opens risks

ecordell: There's no case where this will return anything "incorrect" - the only option is returning a quantized revision, and that may or may not be replicated. I'm not following if you have a correctness concern.

This flag _can_ be used for a transparent proxy that balances reads between a primary and a replica.

Let's say there's replication lag of `2s`, and we quantize on `5s`, and there's a proxy that routes read requests randomly between the primary and the replica (and we have 1 rev / s for simplicity):

Without the flag: 

- Primary is at rev 11, it will tell you the current quantized revision is 10
- Secondary is at rev 9, it will tell you the current quantized revision is 5

If initial request hits the secondary, it will pick 5 and the request will work. If the initial request hits the primary, it will pick 10 and then any request that hits the secondary will fail with a replication error.

With the flag set to delay: 3s

- Primary is at rev 11, it will tell you "now" is 8 and the current quantized revision is 5
- Secondary is at rev 9, it will tell you  "now" is 6  and the current quantized revision is 5

Now both instances will pick 5, which is already replicated because the lag is 2s.

Here's the full table of revision selection for two quantizations:

| Primary Revision  | Secondary Revision (primary - lag) |  Primary Revision (adjusted) | Secondary Revision (adjusted) |  Primary Quantized Rev | Secondary Quantized Rev | 
| ------------- | ------------- | -------- | --------- | ----- | ----- |
| 5 | 3  | 2 | 0 |  0 | 0 |
| 6 | 4  | 3 | 1 |  0 | 0 |
| 7 | 5  | 4 | 2 |  0 | 0 |
| 8 | 6  | 5 | 3 |  5 | 0 |
| 9 | 7  | 6 | 4 |  5 | 0 |
| 10 | 8  | 7 | 5 |  5 | 5 |
| 11 | 9 | 8 | 6 |  5 | 5 |
| 12 | 10  | 9 | 7 |  5 | 5 |
| 13 | 11  | 10 | 8 |  10 | 5 |
| 14 | 12  | 11 | 9 |  10 | 5 |
| 15 | 13  | 12 | 10 |  10 | 10 |

Note that in every case both nodes always pick a revision that is available in both replicas. The only time it would fail is when lag exceeds the follower read delay, but that's what you'd expect (and what happens in the "non-transparent" case).

josephschorr: I think you misunderstood what I was saying: "reads to a replica" is where the problem comes in, not this flag specifically. If you make a fully consistent request but the read of the latest revision hits the replica without SpiceDB knowing, it will potentially return an older revision and therefore what should be a fully consistent read is now... not.

> This flag can be used for a transparent proxy that balances reads between a primary and a replica.

That's my point - such transparent proxies are *broken*, which is why we have very strict rules on when it is safe to use a replica and we have the strict-read-checking injection of SQL assertions as well

josephschorr (approved):
LGTM

pkg/cmd/datastore/datastore.go
josephschorr: As discussed offline, we'll keep the flag open

vroldanbet, line 344: @ecordell shouldn't this be checking that read replicas are disabled?

PR #2252, opened 2025-02-26
make pg datastore continuously checkpoint using a revision heartbeat

vroldanbet:
alternative implementation to https://github.com/authzed/spicedb/pull/2247

## Context

Postgres datastore does not offer a continuous checkpointing Watch API because Postgres revisions are not time-based.
In the other databases, any timestamp represents a fully fledged revision, while in the Postgres datastore, a timestamp does not necessarily represent a transaction.

This is only relevant for consumers of the Watch API, and more particularly, consumers of the datastore Watch API, which exposes the concept of Watch Checkpoints. Consumers of the Postgres Watch API don't observe the passing of time, and this is important for any system that needs to report lag as a health metric.

## The solution

Introduces a leader-elected worker that will insert an artificial transaction into the transactions table so long there isn't any other revision already in the same quantization window.

This fixes three problems

## Problem 1: Cache Thrashing with Out-Of-Band changes

When out-of-band writes happen in the same cluster, it will make the `OptimizedRevision` continuously move, and this will impact revision quantization and thus caching, essentially turning into `fully_consistent` behavior. This can happen in SpiceDB clusters that share Postgres instances with another SpiceDB cluster or unrelated workloads.

## Problem 2: Watch API Continuous Checkpointing

If no changes are happening in the cluster, Watch API does not emit any changes or checkpoints, which makes it difficult for clients to know what the reported lag is, but also makes it difficult to reestablish a dropped Watch API connection because the client would be holding onto a very old revision potentially outside of the GC window.

## Problem 3: Bug on Watch API with out-of-band changes

This change fixes a latent bug identified while reasoning about all the different pieces in place to support Postgres datastore watch API. 

In commit https://github.com/authzed/spicedb/commit/74dc7b26a8d1ec70266f7ac99edee10d11cb269f a change was introduced to address the lack of checkpoints when the Postgres datastore moved its `pg_current_snapshot()` outside of application changes (e.g. `VACUUM` operations, or other local databases in the same Postgres instance). The inconsistency identified is that `pg_current_snapshot()` will be emitted as a checkpoint, but a new transaction can come in and be evaluated at that transaction because `pg_current_snapshot()` returns _the snapshot at which the query is evaluated_, and that does not consume it if not transaction ID was assigned (e.g. there was no commit involved). Consequently, it's possible to observe one checkpoint at revision T and then a change that comes after that at the same revision T, which violates the semantics of the checkpoint acting as a high watermark (the client must
not observe pop-in after the checkpoint).

The issue is now solved because it uses the data fed from the table using a synthetic revision.

### DIFF
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -311,6 +311,15 @@ func newPostgresDatastore(
 		followerReadDelayNanos,
 	)
 
+	revisionHeartbeatQuery := fmt.Sprintf(
+		insertHeartBeatRevision,
+		colXID,
+		tableTransaction,
+		colTimestamp,
+		quantizationPeriodNanos,
+		colSnapshot,
+	)
+
 	validTransactionQuery := fmt.Sprintf(
 		queryValidTransaction,
 		colXID,
@@ -353,12 +362,13 @@ func newPostgresDatastore(
 		watchBufferWriteTimeout: config.watchBufferWriteTimeout,
 		optimizedRevisionQuery:  revisionQuery,
 		validTransactionQuery:   validTransactionQuery,
+		revisionHeartbeatQuery:  revisionHeartbeatQuery,
 		gcWindow:                config.gcWindow,
 		gcInterval:              config.gcInterval,
 		gcTimeout:               config.gcMaxOperationTime,
 		analyzeBeforeStatistics: config.analyzeBeforeStatistics,
 		watchEnabled:            watchEnabled,
-		gcCtx:                   gcCtx,
+		workerCtx:               gcCtx,
 		cancelGc:                cancelGc,
 		readTxOptions:           pgx.TxOptions{IsoLevel: pgx.RepeatableRead, AccessMode: pgx.ReadOnly},
 		maxRetries:              config.maxRetries,
@@ -367,6 +377,7 @@ func newPostgresDatastore(
 		inStrictReadMode:        config.readStrictMode,
 		filterMaximumIDCount:    config.filterMaximumIDCount,
 		schema:                  *schema,
+		quantizationPeriodNanos: quantizationPeriodNanos,
 	}
 
 	if isPrimary && config.readStrictMode {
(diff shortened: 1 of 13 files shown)


pkg/cmd/serve.go
```diff
@@ -169,6 +169,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 			Msg("The old implementation of LookupResources is no longer available, and a `false` value is no longer valid. Please remove this flag.")
 	}
 
+	experimentalFlags.BoolVar(&config.EnableExperimentalRevisionHeartbeat, "enable-experimental-revision-heartbeat", false, "enables experimental support for postgres revision heartbeat, used to create a synthetic revision on a given interval (postgres only)")
```
ecordell, line 172: I would lean towards this being `true` by default right away, since I think the behavior without it could be characterized as a bug if you have _any_ other workload talking to the same PG instance. As an optimization you could disable it if you're talking to a DB that is dedicated to a single spicedb, and know you don't need the watch heartbeats.

i.e. I'm not sure it's worth soaking this as default-false for a release, but I don't feel strongly if you disagree.

ecordell (approved):
LGTM, I left a couple of comments that you can address in this PR or a followup, whichever you prefer.

internal/datastore/postgres/postgres.go
```diff
@@ -721,7 +741,7 @@ func (pgd *pgDatastore) OfflineFeatures() (*datastore.Features, error) {
 				Status: datastore.FeatureUnsupported,
 			},
 			ContinuousCheckpointing: datastore.Feature{
-				Status: datastore.FeatureUnsupported,
+				Status: datastore.FeatureSupported,
```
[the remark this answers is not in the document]
vroldanbet: hmm, that's fair

pkg/cmd/serve.go
vroldanbet: I don't disagree with any of that - I was just being conservative. I think the part that I wouldn't be so ok with is the fact we are turning an experimental flag on by default, which seems to contradict the definition of experimental flag 😅 . Should I make it non-experimental?

ecordell: yeah I think I'd lean towards not experimental, and then it's something you can disable if you really want a tiny bit of extra perf from pg and know it's safe.

vroldanbet:
follow ups on https://github.com/authzed/spicedb/pull/2254

PR #2262, opened 2025-03-06
Convert packages to use the new schema package

barakmich:


tstirrat15 (approved):
LGTM! I appreciate that you changed the names of vars and not just the types - it made it easier to follow what had changed.

PR #2263, opened 2025-03-07
OS-222: Require definitions and partials to have distinct names

tstirrat15:
## Description
Currently, it's possible to give a definition and a partial the same name, which is fine but potentially misleading/confusing. This makes it so that if a definition or a caveat has the same name as an existing partial, there's an error.

Note that it also brings the notion of prefixes into partials. I'm not entirely sure if that makes sense, because a prefix will never be a part of a realized schema, but I don't think it should be a problem.

## Changes
* Make the `compiledPartials` map a reference instead of a value so that it can be read post-modification
* Add new tests for collision
* Add prefixes to the names of partials so that they match the names of caveats/definitions
* Add checks for collision

## Testing
Review. See that tests go green

### DIFF
--- a/pkg/composableschemadsl/compiler/translator.go
+++ b/pkg/composableschemadsl/compiler/translator.go
@@ -30,7 +30,7 @@ type translationContext struct {
 	enabledFlags     []string
 	existingNames    *mapz.Set[string]
 	// The mapping of partial name -> relations represented by the partial
-	compiledPartials *map[string][]*core.Relation
+	compiledPartials map[string][]*core.Relation
 	// A mapping of partial name -> partial DSL nodes whose resolution depends on
 	// the resolution of the named partial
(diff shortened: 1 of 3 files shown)


pkg/composableschemadsl/compiler/translator.go
```diff
@@ -30,7 +30,7 @@ type translationContext struct {
 	enabledFlags     []string
 	existingNames    *mapz.Set[string]
 	// The mapping of partial name -> relations represented by the partial
-	compiledPartials map[string][]*core.Relation
+	compiledPartials *map[string][]*core.Relation
```
barakmich, line 33: Is there a reason to make this a pointer-to-map? It looks like the other lines would be fine with just a regular map

barakmich (dismissed):

PR #2268, opened 2025-03-10
Decouple validation file schema compilation from yaml parsing

tstirrat15:
## Description
Part of making it possible to run `zed validate` on files that use composable schema syntax. In order to have different compilation behaviors in different cases, we need to not run the compilation as a part of yaml parsing. This implements that.

## Changes
Will annotate.

## Testing

### DIFF
--- a/internal/datastore/proxy/schemacaching/types.go
+++ b/internal/datastore/proxy/schemacaching/types.go
@@ -1,6 +1,6 @@
 package schemacaching
 
-import "github.com/authzed/spicedb/pkg/schemadsl/compiler"
+import "github.com/authzed/spicedb/pkg/commonschemadsl"
 
 // *DefinitionSizeVTMultiplier are the mulitipliers to be used for
 // estimating the in-memory cost of a SchemaDefinition based on its
@@ -16,6 +16,6 @@ const (
 )
 
 type schemaDefinition interface {
-	compiler.SchemaDefinition
+	commonschemadsl.SchemaDefinition
 	SizeVT() int
 }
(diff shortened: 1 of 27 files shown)


.golangci.yaml
```diff
@@ -37,10 +37,10 @@ linters:
     - 'rowserrcheck'
     - 'staticcheck'
     - 'stylecheck'
-    - 'tenv'
```
tstirrat15, line 40: This was marked as deprecated in favor of `usetesting` when I ran `golangci-lint run` locally.

internal/services/integrationtesting/cert_test.go
```diff
@@ -46,8 +46,7 @@ func TestCertRotation(t *testing.T) {
 		waitFactor = 2
 	)
 
-	certDir, err := os.MkdirTemp("", "test-certs-")
-	require.NoError(t, err)
+	certDir := t.TempDir()
```
tstirrat15, line 49: one of the lint recommendations.

pkg/validationfile/blocks/schema.go
```diff
@@ -1,70 +1,28 @@
...
+// position of the schema within the validation file.
+type SchemaWithPosition struct {
 	// Schema is the schema found.
 	Schema string
 
 	// SourcePosition is the position of the schema in the file.
 	SourcePosition spiceerrors.SourcePosition
-
-	// CompiledSchema is the compiled schema.
-	CompiledSchema *compiler.CompiledSchema
```
tstirrat15, line 25: This got pushed out into `CompileSchema`.

tstirrat15:
We took a different approach in authzed/zed#478 so this is no longer necessary.

PR #2272, opened 2025-03-13
Secondary dispatch improvements: Add LookupSubjects support and dynamic hedging

josephschorr:
Adds LookupSubjects support for secondary dispatching and adds dynamic hedging support to secondary+primary dispatching

### DIFF
--- a/internal/dispatch/remote/cluster.go
+++ b/internal/dispatch/remote/cluster.go
@@ -236,7 +236,7 @@ const (
 	primaryDispatcher     = "$primary"
 )
 
-func publishClient[R responseMessageWithCursor](ctx context.Context, client receiver[R], stream dispatch.Stream[R], secondaryDispatchName string) error {
+func publishClient[R responseMessage](ctx context.Context, client receiver[R], stream dispatch.Stream[R], secondaryDispatchName string) error {
 	for {
 		select {
 		case <-ctx.Done():
@@ -256,11 +256,13 @@ func publishClient[R responseMessageWithCursor](ctx context.Context, client rece
 			}
 
 			if secondaryDispatchName != primaryDispatcher {
-				afterResponseCursor := result.GetAfterResponseCursor()
-				if afterResponseCursor == nil {
-					return spiceerrors.MustBugf("received a nil after response cursor for secondary dispatch")
+				if supportsCursors, ok := any(result).(responseMessageWithCursor); ok {
+					afterResponseCursor := supportsCursors.GetAfterResponseCursor()
+					if afterResponseCursor == nil {
+						return spiceerrors.MustBugf("received a nil after response cursor for secondary dispatch")
+					}
+					afterResponseCursor.Sections = append([]string{secondaryCursorPrefix + secondaryDispatchName}, afterResponseCursor.Sections...)
 				}
-				afterResponseCursor.Sections = append([]string{secondaryCursorPrefix + secondaryDispatchName}, afterResponseCursor.Sections...)
 			}
 
 			serr := stream.Publish(result)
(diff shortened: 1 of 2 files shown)


internal/dispatch/remote/cluster.go
```diff
@@ -86,8 +112,37 @@ type clusterDispatcher struct {
...
+
+	secondaryDispatch               map[string]SecondaryDispatch
+	secondaryDispatchExprs          map[string]*DispatchExpr
+	secondaryInitialResponseDigests map[string]*digestAndLock
+}
+
+// digestAndLock is a struct that holds a TDigest and a lock to protect it.
+type digestAndLock struct {
+	digest *tdigest.TDigest
+	lock   sync.RWMutex
```
vroldanbet, line 124: I'm worried about the contention this will cause when dispatching at large scale, specially on spikey requests.

We could use `TryLock` and `RTryLock` on write and read paths.
- If we add a new add result time, it would be fine to _drop it_ if we can't acquire the lock with `TryLock`. At a certain scale, all those dropped stats would likely not be meaningful.
If we are reading the current time, we could do `TryRLock`. If we don't get it, we could use a cached value (although that will need another lock 😢 ) or just a default value.

https://github.com/puzpuzpuz/xsync offers a reader-biased mutex based on a paper. Unfortunately, it seems like writes and reads in this code are going to be 50/50. However, we could make it more reader-biased if we at least implement `TryLock` on the read path, and stick to the same behavior in `getWaitTime`

What do you think?

All of this needs of a load-test to understand behavior at scale.

vroldanbet: I went ahead and wrote a benchmark:

```go
func BenchmarkDAL(b *testing.B) {
	dal := &digestAndLock{
		digest: tdigest.NewWithCompression(1000),
		lock:   sync.RWMutex{},
	}

	heavyWritesFn := func() {
		for i := 0; i < b.N*10; i++ {
			dal.addResultTime(1)
		}
	}

	go heavyWritesFn()
	go heavyWritesFn()
	go heavyWritesFn()

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		dal.getWaitTime()
	}
	b.StopTimer()
}

```

Here is the change
```
func (dal *digestAndLock) addResultTime(duration time.Duration) {
	if dal.lock.TryLock() {
		dal.digest.Add(float64(duration.Milliseconds()), 1)
		dal.lock.Unlock()
	}
}
```

I observed 2 orders of magnitude better throughput

```
BenchmarkDAL-12        31273	     40102 ns/op	     327 B/op	       0 allocs/op
BenchmarkNewDAL-12     1000000	      1281 ns/op	       0 B/op	       0 allocs/op
```

Also tested xsync, and as expected (this is 50/50 workload) it did not improve things, and in fact was slightly worse in this benchmark

```
BenchmarkXsyncDAL-12    	 1000000	      1363 ns/op	       2 B/op	       0 allocs/op
```

vroldanbet (approved):
Changes look good to me, I added a suggestion to optimize contention, but does not have to block landing this

internal/dispatch/remote/cluster.go
```diff
@@ -363,6 +424,15 @@ func dispatchStreamingRequest[Q requestMessage, R responseMessage](
 
 		clientCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), cr.dispatchOverallTimeout)
```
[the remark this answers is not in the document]
josephschorr: No; its still attached at the dispatch layer

internal/dispatch/remote/cluster.go
vroldanbet: Talked offline, we will go with this change 👍🏻

PR #2276, opened 2025-03-13
bug-fix: fix typo in flag prefix

kartikaysaxena:
A user recently reported that they can't see the effect of setting the env var `SPICEDB_DATASTORE_READ_REPLICA_CONN_POOL_READ_MAX_OPEN`, this is because the flags here are registered with a prefix of `datastore-read-replica-conn-pool` without the `read` in the end (docs suggest the flags here should be like `--datastore-read-replica-conn-pool-read-*`), this fixes that.


tstirrat15 (approved):
This looks good to me; will merge.

PR #2279, opened 2025-03-14
[schema] fix automatic validation if the source is pre-validated

barakmich:
Cache only definitions that are validated. If it's coming from a datastore, it's pre-validated; if it's local it may not be. This allows us to skip validating if we've already done the work.

### DIFF
--- a/pkg/schema/typesystem_validation.go
+++ b/pkg/schema/typesystem_validation.go
@@ -14,11 +14,19 @@ import (
 
 // GetValidatedDefinition runs validation on the type system for the definition to ensure it is consistent.
 func (ts *TypeSystem) GetValidatedDefinition(ctx context.Context, definition string) (*ValidatedDefinition, error) {
-	def, err := ts.GetDefinition(ctx, definition)
+	def, validated, err := ts.getDefinition(ctx, definition)
 	if err != nil {
 		return nil, err
 	}
-	return def.Validate(ctx)
+	if validated {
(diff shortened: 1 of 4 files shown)


pkg/schema/typesystem_validation.go
```diff
@@ -27,7 +35,7 @@ func (def *Definition) Validate(ctx context.Context) (*ValidatedDefinition, erro
 
 		// Validate the usersets's.
 		usersetRewrite := relation.GetUsersetRewrite()
-		rerr, err := graph.WalkRewrite(usersetRewrite, func(childOneof *core.SetOperation_Child) (interface{}, error) {
+		rerr, err := graph.WalkRewrite(usersetRewrite, func(childOneof *core.SetOperation_Child) (any, error) {
```
tstirrat15, line 38: curious, what's the difference between an empty interface and `any`?

tstirrat15 (approved):
Changes make sense; see question

pkg/schema/typesystem_validation.go
josephschorr: Nothing; `any` is an alias for `interface{}`, its just the new way of doing so: 

> For convenience, the predeclared type any is an alias for the empty interface.

barakmich: And it's starting to be a lint (at least in my editor) so I just fixed it while I was there

tstirrat15 (approved):
LGTM

PR #2281, opened 2025-03-17
goreleaser-windows: fix deprecated option and update copyright year

kartikaysaxena:
Was looking for the config for chocolatey when found this, `format` is deprecated as per [deprecation notices](https://goreleaser.com/deprecations/#archivesformat_overridesformat), also updates copyright year, just a small nit here.


tstirrat15 (approved):
LGTM

PR #2284, opened 2025-03-19
Schema Change Events via Watch API

miparnisari:
https://github.com/authzed/spicedb/issues/1995


internal/services/v1/watch.go
```diff
@@ -107,6 +103,24 @@ func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchS
...
+						return status.Errorf(codes.Canceled, "watch canceled by user: %s", err)
+					}
+				}
+				if update.IsCheckpoint {
+					if err := stream.Send(&v1.WatchResponse{
+						IsCheckpoint:                update.IsCheckpoint,
+						ChangesThrough:              zedtoken.MustNewFromRevision(update.Revision),
+						OptionalTransactionMetadata: update.Metadata,
+					}); err != nil {
+						return status.Errorf(codes.Canceled, "watch canceled by user: %s", err)
```
ecordell, line 121: could construct the `WatchResponse` above and then only call `send` once, so that we don't have to copy this error message around?

internal/services/v1/watch.go
```diff
@@ -157,3 +188,18 @@ func filterUpdates(objectTypes *mapz.Set[string], filters []datastore.Relationsh
 
 	return filtered
 }
+
+func convertWatchKindToContent(kinds []v1.WatchKind) datastore.WatchContent {
```
ecordell, line 192: this is correct but a small unit test would be good, just in case we change these types in the future

ecordell (commented):
looks good overall! couple of small suggestions and a caveat that I haven't tried running this myself yet locally

internal/services/v1/watch.go
miparnisari: What you are suggesting is:

```
WatchResponse{schema_change: ..., is_checkpoint: true}
```

My understanding from @josephschorr is that we want separate messages:

```
WatchResponse{schema_change: ... }
WatchResponse{is_checkpoint: true}
```

internal/services/v1/watch.go
miparnisari: Added! Though it feels hackish to make this function public just to be able to test it...

josephschorr: If you add another test file within `package v1`, then you don't need to make it public. I'd do that instead

internal/services/v1/watch.go
josephschorr: Yeah, you need a distinct send

ecordell (approved):
LGTM

PR #2286, opened 2025-03-21
upload coverage to codecov

miparnisari:


ecordell (approved):
LGTM

Whatever we decide to do to lock down actions, that seems orthogonal to adding coverage

magefiles/test.go
```diff
@@ -53,7 +53,7 @@ func (t Test) Unit(ctx context.Context) error {
 
 func (Test) unit(ctx context.Context, coverage bool) error {
 	fmt.Println("running unit tests")
-	args := []string{"-tags", "ci,skipintegrationtests", "-race", "-timeout", "10m", "-count=1"}
+	args := []string{"-tags", "ci,skipintegrationtests", "-race", "-timeout", "10m", "-count=1", "-covermode=atomic", "-coverprofile=coverage.txt"}
```
vroldanbet, line 56: @miparnisari it seems like there is a `coverage` argument here, shouldn't that be respected? i.e. if coverage=true, then add the flags.

Wouldn't this effectively run coverage _always_, which means we get the overhead while testing locally?

miparnisari: You're right! Sent https://github.com/authzed/spicedb/pull/2291

PR #2294, opened 2025-03-27
Change PG and CRDB datastore drivers to use PGX cancelation

josephschorr:
This should (in most cases) allow the connection to be reused when the parent context is canceled, instead of our current behavior of severing the context and allowing the query to finish processing. Severing was initially done because the default behavior in PGX is to *close* the connection, which added overhead on other requests

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -77,9 +77,19 @@ func RegisterGCMetrics() error {
 	return nil
 }
 
-// GarbageCollector represents any datastore that supports external garbage
-// collection.
-type GarbageCollector interface {
+// GarbageCollectableDatastore represents a datastore supporting external
+// and explicit garbage collection.
+type GarbageCollectableDatastore interface {
+	// BuildGarbageCollector builds a garbage collector for the datastore.
+	//
+	// Each instance is considered isolated and will be used for the duration
+	// of a *single* garbage collection call.
+	BuildGarbageCollector(ctx context.Context) (GarbageCollector, error)
+
+	// ReadyState returns the current state of the datastore. Note that this does not
+	// operate over the dedicated connection.
+	ReadyState(context.Context) (datastore.ReadyState, error)
+
 	// HasGCRun returns true if a garbage collection run has been completed.
 	HasGCRun() bool
 
@@ -88,7 +98,10 @@ type GarbageCollector interface {
 
 	// ResetGCCompleted resets the state of the garbage collection run.
 	ResetGCCompleted()
+}
 
+// GarbageCollector represents a runnable garbage collector for a datastore.
+type GarbageCollector interface {
 	// LockForGCRun attempts to acquire a lock for garbage collection. This lock
 	// is typically done at the datastore level, to ensure that no other nodes are
 	// running garbage collection at the same time.
@@ -98,9 +111,6 @@ type GarbageCollector interface {
 	// NOTE: this method does not take a context, as the context used for the
 	// reset of the GC run can be canceled/timed out and the unlock will still need to happen.
 	UnlockAfterGCRun() error
-
-	// ReadyState returns the current state of the datastore.
-	ReadyState(context.Context) (datastore.ReadyState, error)
 
 	// Now returns the current time from the datastore.
 	Now(context.Context) (time.Time, error)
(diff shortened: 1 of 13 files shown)


internal/datastore/postgres/postgres.go
```diff
@@ -185,6 +185,25 @@ func newPostgresDatastore(
 		}
 	}
 
+	var gcPoolConfig *pgxpool.Config
+	if isPrimary {
+		// Configure a pool of size 1 for the GC connection.
+		gcPoolConfig = pgConfig.Copy()
```
ecordell, line 191: I don't think you need a new pool here, you can just `Acquire()` one from the write pool?

josephschorr: But if that connection dies, will it reconnect?

internal/datastore/postgres/locks.go
```diff
@@ -26,7 +25,7 @@ func (pgd *pgDatastore) tryAcquireLock(ctx context.Context, lockID lockID) (bool
 	// > even if other sessions are awaiting the lock; this statement is true regardless of whether the
 	// > existing lock hold and new request are at session level or transaction level.
 	// See: https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS
-	row := pgd.writePool.QueryRow(ctx, `
+	row := pgd.gcConn.QueryRow(ctx, `
```
tstirrat15, line 28: Do we only use this lock for GC? Is it a problem if it's used for something else? Are we using the GC connection pool because we assume that it will usually be free?

tstirrat15 (dismissed):
See comments, otherwise LGTM

internal/datastore/postgres/postgres.go
ecordell: no, but I was imagining you'd just acquire one at the beginning of the gc process, which is periodic

ecordell: also - if the connection dies, you will have lost the gc lock, so you'll _need_ to reconnect explicitly and re-lock

ecordell (changes requested):
Cancellation changes look great, very excited about those

I'm not convinced of the gc pool - since the locks are per-session, losing a connection and reconnecting transparently in the background is bad. I think we just need to acquire from the write pool and then manually handle re-acquiring in the GC loop if needed

internal/datastore/common/gc.go
```diff
@@ -88,31 +91,46 @@ type GarbageCollector interface {
...
+}
+
+// GarbageCollector represents any datastore that supports external garbage
+// collection.
+type GarbageCollector[C any] interface {
+	RunnableGarbageCollection
+
+	// AcquireConnection acquires a connection to the datastore for use by the GC
+	// process. If no dedicated connection is necessary, return nil.
+	AcquireConnection(ctx context.Context) (C, error)
```
ecordell, line 103: I think you could take C out of the interface:
```go
	AcquireConnection(ctx context.Context) error
	ReleaseConnection()
```
and instead track it as an internal member of the type that implements `GarbageCollector`

josephschorr: There is no requirement that the datastore itself act as a singleton garbage collector, so strictly speaking, we can't. We could have the datastore return an instance of a GarbageCollection interface, and then we could remove the connection, but that's a different design

ecordell: Wouldn't this just not be implemented by datastores that don't need a garbage collector?

internal/datastore/postgres/postgres.go
```diff
@@ -162,7 +162,7 @@ func newPostgresDatastore(
 		}
 	}
 
-	// Setup the config for each of the read and write pools.
+	// Setup the config for each of the read, write and GC pools.
```
ecordell, line 165: nit: there is no GC pool anymore

internal/datastore/postgres/gc.go
```diff
@@ -42,15 +53,32 @@ func (pgd *pgDatastore) ResetGCCompleted() {
...
 }
 
-func (pgd *pgDatastore) Now(ctx context.Context) (time.Time, error) {
+func (pgg *pgGarbageCollector) Close() {
+	pgg.isClosed = true
+	pgg.conn.Release()
+}
+
+func (pgg *pgGarbageCollector) LockForGCRun(ctx context.Context) (bool, error) {
+	return pgg.pgd.tryAcquireLock(ctx, pgg.conn, gcRunLock)
```
ecordell, line 62: I think the locking could probably move in to the garbage collector too?

ecordell (approved):
LGTM - new interface looks good. left a couple of small nits but nothing that needs to block.

internal/datastore/postgres/gc.go
josephschorr: We use it for non-GC too

PR #2295, opened 2025-03-31
fix leader election retry interval for revision heartbeat

vroldanbet:
each SpiceDB nodes attempts to acquire a postgres lock to become the leader in charge of publishing the
heartbeat revision.

Unfortunately I overlooked adding both the heartbeat and the jitter computed out of it, leading to
SpiceDB nodes trying to acquire locks
in very rapid succession.

### DIFF
--- a/internal/graph/lookupresources2.go
+++ b/internal/graph/lookupresources2.go
@@ -115,7 +115,8 @@ func (crr *CursoredLookupResources2) afterSameType(
 	req ValidatedLookupResources2Request,
 	parentStream dispatch.LookupResources2Stream,
 ) error {
-	ctx, span := tracer.Start(ctx, "lookupViaReachability")
+	reachabilityForString := req.ResourceRelation.Namespace + "#" + req.ResourceRelation.Relation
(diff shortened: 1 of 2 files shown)


internal/graph/lookupresources2.go
```diff
@@ -145,9 +146,7 @@ func (crr *CursoredLookupResources2) afterSameType(
 			spiceerrors.DebugAssert(func() bool {
 				return err == nil
 			}, "Error in entrypoint.DebugString()")
-			ctx, span := tracer.Start(ctx, "entrypoint", trace.WithAttributes(
-				attribute.String("entrypoint", ds),
-			))
+			ctx, span := tracer.Start(ctx, "entrypoint: "+ds, trace.WithAttributes())
```
tstirrat15, line 149: Is this to help with filtering?

tstirrat15 (approved):
LGTM

miparnisari (approved):

internal/graph/lookupresources2.go
vroldanbet: Basically to make the traces easier to navigate.

PR #2296, opened 2025-03-31
remove limit in AtRevision in dispatch ResolverMeta

ecordell:


vroldanbet (approved):

tstirrat15:
Are there potential performance issues with allowing this field to be unbounded? Or is it no worse than the current state of affairs?

PR #2300, opened 2025-04-01
support both old and new read replica flag prefixes

ecordell:
This addresses a regression in supported flags in 1.42.0, so that read replica connection pool flags with both the old prefix (`datastore-read-replica-conn-pool`) and the new one (`datastore-read-replica-conn-pool-read`) are supported.

There is one edge case in this implementation: if a user explicitly sets a new flag, but sets it to the same value as the default, and also sets the old flag, then the old flag will take precedence (even though the new one is specified). I thought this was unlikely to be hit in practice and users should still see a deprecation warning if they do. We could avoid this issue, but it would require some more refactoring of flag handling so that after the flags are parsed we can lookup what has been actually set on the cli (vs the values in the config struct).

```sh
$ spicedb serve --datastore-read-replica-conn-pool-max-open=30 --grpc-preshared-key=abc
Flag --datastore-read-replica-conn-pool-max-open has been deprecated, please use the flags with the prefix "datastore-read-replica-conn-pool-read" instead of "datastore-read-replica-conn-pool"
<....>
DatastoreConfig.ReadReplicaConnPool.MaxOpenConns=30

$ spicedb serve --datastore-read-replica-conn-pool-read-max-open=30 --grpc-preshared-key=abc
<....>
DatastoreConfig.ReadReplicaConnPool.MaxOpenConns=30
```

### DIFF
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -192,8 +192,6 @@ func (c *closeableStack) CloseIfError(err error) error {
 // if there is no error, a completedServerConfig (with limited options for
 // mutation) is returned.
 func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
-	log.Ctx(ctx).Info().Fields(helpers.Flatten(c.DebugMap())).Msg("configuration")
-
 	closeables := closeableStack{}
 	var err error
 	defer func() {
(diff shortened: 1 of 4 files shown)


pkg/cmd/server/server.go
```diff
@@ -528,6 +529,47 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 	}, nil
 }
 
+func (c *Config) supportOldAndNewReadReplicaConnectionPoolFlags() {
```
tstirrat15, line 532: Ooof. But also I guess this isn't terrible.

tstirrat15 (approved):
Thank you!

PR #2301, opened 2025-04-01
Use golangci-lint v2 for linting

tstirrat15:
## Description
v2 of golangci-lint was recently released. This is a part of dealing with the breaking changes and getting us up to v2.

## Changes
* Migrate the config
* Change magefile to point at binary rather than command
* Add golangci-lint action to lint steps
* Fix lint errors
## Testing
Review. See that things go green.


internal/datastore/common/sql.go
```diff
@@ -467,10 +467,11 @@ func (sqf SchemaQueryFilterer) FilterWithRelationshipsFilter(filter datastore.Re
 		csqf = csqf.FilterWithCaveatName(filter.OptionalCaveatName)
 	}
 
-	if filter.OptionalExpirationOption == datastore.ExpirationFilterOptionHasExpiration {
+	switch filter.OptionalExpirationOption {
```
tstirrat15, line 470: This was one of a set of new lint issues that were automatically fixed by --fix.

internal/datastore/mysql/migrations/manager.go
```diff
@@ -12,7 +12,7 @@ const migrationNamePattern = `^[_a-zA-Z]*$`
 
 var (
 	noNonatomicMigration migrate.MigrationFunc[Wrapper]
-	noTxMigration        migrate.TxMigrationFunc[TxWrapper] // nolint: deadcode, unused, varcheck
```
tstirrat15, line 15: These linters don't exist anymore and golangci was warning about them.

magefiles/lint.go
```diff
@@ -73,7 +73,10 @@ func (Lint) Gofumpt() error {
 // Golangcilint Run golangci-lint
 func (Lint) Golangcilint() error {
 	fmt.Println("running golangci-lint")
-	return RunSh("go", WithV())("run", "github.com/golangci/golangci-lint/cmd/golangci-lint", "run", "--fix")
+	// NOTE: we run the binary here because the maintainers of golangci-lint strongly recommend
+	// that you don't run from source: https://golangci-lint.run/welcome/install/#install-from-sources
+	// We use the action to install in CI and you should install locally to run locally.
+	return sh.RunV("golangci-lint", "run", "--fix")
```
tstirrat15, line 79: This seems like potentially a meaningful change.

pkg/datastore/test/watch.go
```diff
@@ -829,10 +829,7 @@ func WatchEmissionStrategyTest(t *testing.T, tester DatastoreTester) {
 	features, err := ds.Features(ctx)
 	require.NoError(err)
 
-	expectsWatchError := false
-	if !(features.WatchEmitsImmediately.Status == datastore.FeatureSupported) {
-		expectsWatchError = true
-	}
+	expectsWatchError := (features.WatchEmitsImmediately.Status == datastore.FeatureSupported)
```
tstirrat15, line 832: Check this too

tstirrat15 (commented):
See comments

pkg/datastore/test/watch.go
miparnisari: This should be `expectsWatchError := (features.WatchEmitsImmediately.Status != datastore.FeatureSupported)`

vroldanbet: good catch!

magefiles/lint.go
tstirrat15: Fair

magefiles/lint.go
```diff
@@ -73,7 +73,7 @@ func (Lint) Gofumpt() error {
 // Golangcilint Run golangci-lint
 func (Lint) Golangcilint() error {
 	fmt.Println("running golangci-lint")
-	return RunSh("go", WithV())("run", "github.com/golangci/golangci-lint/cmd/golangci-lint", "run", "--fix")
+	return RunSh("go", WithV())("run", "github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.0.2", "run", "--fix")
```
vroldanbet, line 76: We are pinning the tool versions via `tools.go` and `go.mod`, so setting the version here is not appropriate, and we are also in turn not defining the v2 as a dependency in `tools.go`

vroldanbet (changes requested):
sorry, one last issue

vroldanbet (dismissed):

vroldanbet (approved):

PR #2308, opened 2025-04-05
bug-fix(schema): prevent schema write failures when migrating to caveated relation

### LINKED ISSUE #2302: Schema write failing when migrating to caveated relation
### Description
  
When attempting to migrate a schema with a relation that includes subject types with caveats, the migration fails with an error indicating that the allowed type cannot be removed, despite the existing relation being properly caveated.

  ### Steps to Reproduce

  1. Create a schema with a relation that includes subject types with caveats:
     
```
     definition user {}

     caveat only_on_tuesday(day_of_week string) {
       day_of_week == 'tuesday'
     }

     definition document {
       relation writer: user
       relation reader: user | user with only_on_tuesday

       permission edit = writer
       permission view = reader + edit
     }
  ```

2. Create a relationship with the caveat: `document:firstdoc#reader@user:tom[only_on_tuesday]`

3. Attempt to update the schema after removing `user | ` and leaving only `user with only_on_tuesday`

```
     definition user {}

     caveat only_on_tuesday(day_of_week string) {
       day_of_week == 'tuesday'
     }

     definition document {
       relation writer: user
       relation reader: user with only_on_tuesday

       permission edit = writer
       permission view = reader + edit
     }
```

### Actual Result

The `zed schema write` fails: 

```
$ zed schema write schema.zed       
7:03AM FTL failed to write schema error="rpc error: code = InvalidArgument desc = cannot remove allowed type `user` from relation `reader` in object definition `document`, as a relationship exists with it"
```

The relationship exists but conforms to the `user with only_on_tuesday` format: 

```
zed relationship read document:firstdoc reader user:tom
document:firstdoc reader user:tom[only_on_tuesday:{"day_of_week":"tuesday"}]
```

### Expected Result

The `zed schema write` should succeed since the existing relation has the caveat for `user with only_on_tuesday`. SpiceDB should recognize that the existing relationships conform to the schema with caveated subjects.

  ### Environment

  - SpiceDB version: v1.41.0
  - Datastore: Postgres 
  - Client: zed CLI

  ### Impact

This issue blocks schema migrations even when existing relationships conform to the updated schema.

### PR DESCRIPTION
**Pull request by kartikaysaxena**
Fixes #2302

Added a check and a test for the case while migrating a schema with a relation that includes subject types with caveats when a caveated relationship exists

### CODE REVIEW

File: internal/services/shared/schema.go [hunk 1]
```diff
@@ -419,14 +419,19 @@ func sanityCheckNamespaceChanges(
...
-				return diff, err
+			if qyrErr != nil {
+				return diff, qyrErr
+			}
+
+			for relationTuple, iterErr := range qyr {
+				if iterErr != nil {
+					return diff, iterErr
+				}
+				if !(relationTuple.OptionalCaveat != nil && relationTuple.OptionalCaveat.CaveatName != optionalCaveatName) {
```
josephschorr, line 430: this could have to iterate between potentially millions of relationships in a worse-case scenario. We need to add a new filter option to RelationshipFilter to filter based on the caveat (or not) (like we do for expiration) and then specify that filter

josephschorr: @kartikaysaxena I'm going to prep a change to add the support in the datastore for the correct filter; once that is in, you can rebase this PR on top and change to use that datastore filter

kartikaysaxena: makes sense, sure!

josephschorr: PR merged; please rebase and update

kartikaysaxena: It seems like the fix was done in that PR itself (migrating from `OptionalCaveatName` to `OptionalCaveatNameFilter`) so this adds just a test for the same.

josephschorr: Yeah, I just realized I accidentally fixed it :)

josephschorr (approved):
LGTM

PR #2315, opened 2025-04-09
Add a concept of a caveat TypeSet to allow overriding the types available to caveat processing

josephschorr:


barakmich (commented):
I like the fact it's all piped through! But then it immediately reads to me as though there's a config `struct` waiting to happen for the params to some of the `NewX` functions, specifically in `internal/dispatch/graph/graph.go` and `internal/graph/lookupresources2.go` (and maybe `internal/services/v1/schema.go`)

barakmich:
> > I like the fact it's all piped through! But then it immediately reads to me as though there's a config `struct` waiting to happen for the params to some of the `NewX` functions, specifically in `internal/dispatch/graph/graph.go` and `internal/graph/lookupresources2.go` (and maybe `internal/services/v1/schema.go`)
> 
> Almost certainly. Do you think we should do it now or as a followup?

Might as well; easy enough to find the locations since it follows this diff

pkg/cmd/server/pgserver.zip
tstirrat15, line 1: Unintentional?

pkg/caveats/types/typeset.go
```diff
@@ -0,0 +1,71 @@
...
+	customMethodsOnTypes []cel.EnvOption
+
+	// isFrozen indicates whether the TypeSet is frozen. A frozen TypeSet cannot be modified.
+	isFrozen bool
+}
+
+// Freeze marks the TypeSet as frozen. A frozen TypeSet cannot be modified.
+func (ts *TypeSet) Freeze() {
+	ts.isFrozen = true
+}
```
tstirrat15, line 32: Is this a convention? Where does this actually prevent the TypeSet from being modified?

tstirrat15 (approved):
LGTM, pending removal of the zip and a question

pkg/cmd/server/pgserver.zip
josephschorr: Yep

pkg/caveats/types/typeset.go
josephschorr: Check below: it can't be used if not frozen

tstirrat15: Right, it can't be used, but nothing prevents me from touching a frozen typeset (unless I'm missing something)

tstirrat15 (dismissed):
LGTM

barakmich (approved):
Boy howdy that's a lot of files. Fortunately, mostly signature-driven.

LGTM

PR #2316, opened 2025-04-10
document packages

miparnisari:
I glazed over the code to see what is implemented and to get a feeling of the overall organization. (And I fixed some typos along the way)

What stood out to me is that there are lot of **folders with the same name within `internal` and within `pkg`**. The line of what goes where seems blurry. E.g.:
- `graph` (only 2 files in `pkg`?)
- `caveats`. And, surprisingly, different `Evaluation Results` structs exist in both folders
- `namespace` / `schema` 

I believe that this could be simplified by merging folders, but this could break existing consumers of `pkg` if we move to `internal`. Who are the code consumers of `pkg` today? Do we provide guarantees on its stability?

**Other things I noticed:**
- I think it'd be good to document in https://authzed.com/docs/spicedb/ what request and response headers are available.
- the code used by the Playground appeared to be spread across a couple of folders

I'd be happy to submit follow-up PRs for all these things 😄

### DIFF
--- a/e2e/doc.go
+++ b/e2e/doc.go
@@ -1,0 +1,2 @@
+// Package e2e contains a series of tests, including tests that verify SpiceDB addresses the New Enemy problem.
+package e2e
--- a/internal/datasets/subjectset.go
+++ b/internal/datasets/subjectset.go
@@ -44,7 +44,7 @@ func (ss SubjectSet) UnionWithSet(other SubjectSet) error {
 	return ss.BaseSubjectSet.UnionWithSet(other.BaseSubjectSet)
 }
 
-// WithParentCaveatExpression returns a copy of the subject set with the parent caveat expression applied
+// WithParentCaveatExpression returns a copy of the subject set with the parent caveat expression applied
 // to all members of this set.
 func (ss SubjectSet) WithParentCaveatExpression(parentCaveatExpr *core.CaveatExpression) SubjectSet {
 	return SubjectSet{ss.BaseSubjectSet.WithParentCaveatExpression(parentCaveatExpr)}
--- a/internal/middleware/servicespecific/doc.go
+++ b/internal/middleware/servicespecific/doc.go
@@ -1,0 +1,2 @@
+// Package servicespecific defines middleware that injects other middlewares.
+package servicespecific
--- a/internal/services/doc.go
+++ b/internal/services/doc.go
@@ -1,0 +1,2 @@
+// Package services contains all the gRPC controllers.
+package services
--- a/pkg/cmd/doc.go
+++ b/pkg/cmd/doc.go
@@ -1,0 +1,2 @@
+// Package cmd defines various public and internal commands.
+package cmd
--- a/pkg/development/doc.go
+++ b/pkg/development/doc.go
@@ -1,0 +1,2 @@
+// Package development contains code that runs in the Playground.
+package development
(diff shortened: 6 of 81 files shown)


internal/developmentmembership/foundsubject.go
```diff
@@ -1,3 +1,4 @@
+// Package developmentmembership defines operations with sets. To be used in tests only.
```
miparnisari, line 1: I hope i got this right... `developmentmembership` wasn't a clear name 😅

e2e/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package e2e contains a series of tests, including against the New Enemy problem.
```
vroldanbet, line 1:
```suggestion
// Package e2e contains a series of tests, including tests that verify SpiceDB addresses the New Enemy problem.
```

pkg/cmd/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package cmd defines various public commands and internal commands (e.g. to serve the playground).
```
vroldanbet, line 1: SpiceDB does not serve the playground

internal/services/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package services contains code to register and manage gRPC services, including Schema and Watch services.
```
vroldanbet, line 1: I'd say this is the package that includes all gRPC controllers

pkg/zedtoken/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package zedtoken converts decimal.Decimal to zedtoken and vice versa
```
vroldanbet, line 1: this seems very implementation specific, perhaps "APIs to handle zedtokens"?

vroldanbet (approved):

pkg/cmd/doc.go
miparnisari: Right. But for example, the LSP command, isn't that used by the playground?

vroldanbet: I'm not aware the Playground has been updated to use the LSP. I know it was developed for the Visual Studio Code extension, and any IDE that supports LSP

miparnisari: Ok, removed references to the playground.

internal/datasets/subjectset.go
```diff
@@ -44,7 +44,7 @@ func (ss SubjectSet) UnionWithSet(other SubjectSet) error {
 	return ss.BaseSubjectSet.UnionWithSet(other.BaseSubjectSet)
 }
 
-// WithParentCaveatExpression returns a copy of the subject set with the parent caveat expression applied
+// WithParentCaveatExpression returns a copy of the subject set with the parent caveat expression applied
```
tstirrat15, line 47: nbsp?

internal/middleware/servicespecific/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package servicespecific defines middleware that injects other middlewares.
```
tstirrat15, line 1:
```suggestion
// Package servicespecific defines middleware that injects other middlewares for service-specific use cases.
```

pkg/development/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package development contains code that runs in the Playground.
+package development
```
tstirrat15, line 2: I think this is also used by `zed`. Maybe "used by other libraries to facilitate schema development?"

tstirrat15 (approved):
LGTM, see comments

PR #2322, opened 2025-04-11
internal/datastore/pg: server-side loop for repair

jzelinskie:
This change replaces our batched-send strategy for incrementing the xact_id with a PL/pgSQL for-loop. Locally, this leads to a x3 throughput improvement.

### DIFF
--- a/internal/datastore/postgres/postgres.go
+++ b/internal/datastore/postgres/postgres.go
@@ -69,15 +69,6 @@ const (
 	repairBatchSize = 1_000_000
 )
 
-// queryLoopXactID performs pg_current_xact_id() in a server-side loop in the
-// database in order to increment the xact_id.
-var queryLoopXactID = fmt.Sprintf(`DO $$
-BEGIN
-  FOR i IN 1..%d LOOP
-    PERFORM pg_current_xact_id(); ROLLBACK;
-  END LOOP;
-END $$;`, repairBatchSize)
-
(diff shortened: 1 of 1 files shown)


internal/datastore/postgres/postgres.go
```diff
@@ -558,17 +572,9 @@ func (pgd *pgDatastore) repairTransactionIDs(ctx context.Context, outputProgress
...
 	for i := 0; i < counterDelta; i++ {
-		var batch pgx.Batch
-
-		batchCount := min(batchSize, counterDelta-i)
-		for j := 0; j < batchCount; j++ {
-			batch.Queue("begin;")
-			batch.Queue("select pg_current_xact_id();")
-			batch.Queue("rollback;")
-		}
+		batchCount := min(repairBatchSize, counterDelta-i)
```
vroldanbet, line 575: This is no longer used to drive the number of iterations—it's hardcoded to 1M. This means the last iteration may only need a slight increase, but the query will still execute 1M, causing an excess increase in the snapshot xact_id.

I don't think this is a big deal, but wanted to check if it's deliberate.

jzelinskie: I pulled this into a function that wraps sprintf. Sprintf overhead is minimal compared to the time spent server-side and we already do validation that we're only working with integers, so there isn't an opportunity for injection.

vroldanbet (approved):

PR #2325, opened 2025-04-14
Add logical checks telemetry

josephschorr:


internal/telemetry/metrics.go
```diff
@@ -144,8 +155,11 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
 		log.Warn().Err(err).Msg("unable to collect datastore statistics")
 	}
 
+	logicalChecksCount := loadLogicalChecksCount()
```
tstirrat15, line 158: Given that loading zeroes things out, are we concerned at all about the value failing to be sent in this block? Or is a dropped metric send here and there not gonna matter too much for what this metric is supposed to do?

tstirrat15 (dismissed):
LGTM, had question

internal/telemetry/metrics.go
josephschorr: A dropped metric is probably fine; this is a rough estimate anyway

vroldanbet (changes requested):

internal/telemetry/logicalchecks.go
```diff
@@ -0,0 +1,16 @@
+package telemetry
+
+import "sync/atomic"
+
+var logicalChecksCountTotal atomic.Uint64
```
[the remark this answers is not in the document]
josephschorr: Prometheus library uses the same `atomic.AddUint64`: https://github.com/prometheus/client_golang/blob/main/prometheus/counter.go#L133

All docs I've found indicate that using a mutex is slower than an atomic increment.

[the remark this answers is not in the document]
vroldanbet: The codebase already has mutexes in various spots, so this adds to the pile of synchronization. I understand doing something that is better will be relatively complex and prone to bugs.

internal/services/v1/permissions.go
```diff
@@ -461,6 +466,8 @@ func (ps *permissionServer) LookupResources(req *v1.LookupResourcesRequest, resp
 	}
 
 	alreadyPublishedPermissionedResourceIds := map[string]struct{}{}
+	var totalCountPublished uint64
+	defer telemetry.RecordLogicalChecks(totalCountPublished)
```
vroldanbet, line 470: This does not work. You are capturing the value of `totalCountPublished` as zero.

Demonstration:
```go
package telemetry

import (
	"testing"
)

func TestDoesNotWork(t *testing.T) {
	var count uint64
	defer func() {
		require.Equal(t, uint64(1), loadLogicalChecksCount())
	}()
	defer RecordLogicalChecks(count)
	count++
}

func TestDoesWork(t *testing.T) {
	var count uint64
	defer func() {
		require.Equal(t, uint64(1), loadLogicalChecksCount())
	}()
	defer func() {
		RecordLogicalChecks(count)
	}()
	count++
}
```

vroldanbet (changes requested):

internal/services/v1/permissions.go
josephschorr: Fixed

vroldanbet (approved):

PR #2327, opened 2025-04-14
Force indexes on Postgres and change subject sort order to match index

josephschorr:
Requires `pg_hint_plan` to be installed for the index forcing to work

### DIFF
--- a/internal/datastore/postgres/gc.go
+++ b/internal/datastore/postgres/gc.go
@@ -67,6 +67,9 @@ func (pgd *pgDatastore) TxIDBefore(ctx context.Context, before time.Time) (datas
 		return datastore.NoRevision, err
 	}
 
+	// Force the proper index for the GC operation.
+	sql = "/*+ IndexOnlyScan(" + schema.TableTransaction + " " + schema.IndexSortedRelationTupleTransaction.Name + ") */" + sql
+
 	var value xid8
 	var snapshot pgSnapshot
 	err = pgd.readPool.QueryRow(ctx, sql, args...).Scan(&value, &snapshot)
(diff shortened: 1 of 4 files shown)


internal/datastore/postgres/gc.go
```diff
@@ -145,11 +152,16 @@ func (pgd *pgDatastore) batchDelete(
...
+	index *common.IndexDefinition,
 ) (int64, error) {
 	sql, args, err := psql.Select(pkCols...).From(tableName).Where(filter).Limit(gcBatchDeleteSize).ToSql()
 	if err != nil {
 		return -1, err
 	}
+	if index != nil {
+		// Force the proper index for the GC operation.
+		sql = "/*+ IndexOnlyScan(" + tableName + " " + index.Name + ") */" + sql
+	}
```
vroldanbet, line 164: please turn this into a helper function (a helper that reuses `forcedIndex`)?

internal/datastore/postgres/gc.go
```diff
@@ -67,6 +67,9 @@ func (pgd *pgDatastore) TxIDBefore(ctx context.Context, before time.Time) (datas
 		return datastore.NoRevision, err
 	}
 
+	// Force the proper index for the GC operation.
+	sql = "/*+ IndexOnlyScan(" + schema.TableTransaction + " " + schema.IndexSortedRelationTupleTransaction.Name + ") */" + sql
```
vroldanbet, line 71: Please clarify in the PR body that `pg_hint_plan` receives index hints as SQL comments

vroldanbet: please turn this into a helper function

PR #2328, opened 2025-04-15
postgres: optimize CheckRevision query

vroldanbet:
the condition in postgres' CheckRevision datastore could lead to slow queries for very large
transactions table. This was caused by the
`timestamp = (SELECT MAX(timestamp) FROM relation_tuple_transaction)` portion of the query, which led to filtering rows which could have been discarded in the first place by the other expression.

This query was contributed by one of our customers.


tstirrat15 (approved):
Breaking down the query makes sense to me; the results are impressive.

PR #2330, opened 2025-04-15
Further optimize the cases when transaction metadata is written in CRDB

josephschorr:
Since the metadata is only really needed if user specified OR there are only DELETE operations in the transaction, we only write the data in those cases

### DIFF
--- a/internal/datastore/crdb/readwrite.go
+++ b/internal/datastore/crdb/readwrite.go
@@ -49,8 +49,9 @@ var (
 
 type crdbReadWriteTXN struct {
 	*crdbReader
-	tx             pgx.Tx
-	relCountChange int64
+	tx                          pgx.Tx
+	relCountChange              int64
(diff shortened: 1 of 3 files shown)


tstirrat15 (approved):
LGTM

internal/datastore/crdb/readwrite.go
```diff
@@ -49,8 +49,9 @@ var (
 
 type crdbReadWriteTXN struct {
 	*crdbReader
-	tx             pgx.Tx
-	relCountChange int64
+	tx                          pgx.Tx
+	relCountChange              int64
+	hasNonExpiredDeletionChange bool
```
miparnisari, line 54: @josephschorr sorry for the late review. Why is `expired` in the field's name? it seems to me that a better name would be `hasNonDeletionChange`. Since expiration doesn't affect when this flag is set.

PR #2334, opened 2025-04-16
add mutex analyzer

miparnisari:


tools/analyzers/mutexcheck/mutexcheck.go
```diff
@@ -0,0 +1,163 @@
...
+		Flags:    *flagSet,
+	}
+}
+
+func isMutexType(field *ast.Field) ([]string, bool) {
+	isMutexType := false
+	var fieldNames []string
+	if se, ok := field.Type.(*ast.SelectorExpr); ok {
+		if ident, ok := se.X.(*ast.Ident); ok {
+			if ident.Name == "sync" && se.Sel.Name == "Mutex" {
```
josephschorr, line 150: We should check for `RWMutex` as well

internal/taskrunner/preloadedtaskrunner.go
```diff
@@ -19,9 +19,9 @@ type PreloadedTaskRunner struct {
 	sem chan struct{}
 
 	wg    sync.WaitGroup
-	err   error
+	err   error // GUARDED_BY(lock)
 	lock  sync.Mutex
```
josephschorr, line 23: mind reordering it so the lock is above the fields here?

internal/taskrunner/taskrunner.go
```diff
@@ -19,11 +19,11 @@ type TaskRunner struct {
 
 	// err holds the error returned by any task, if any. If the context is canceled,
 	// this err will hold the cancelation error.
-	err error
+	err error // GUARDED_BY(lock)
```
josephschorr, line 22: ditto

josephschorr (approved):
LGTM

PR #2335, opened 2025-04-16
Simplify the registration in the caveats typeset

josephschorr:
Also allows for registration of other custom CEL env


tstirrat15 (approved):
LGTM

PR #2338, opened 2025-04-17
Fix logger for certmanager

tstirrat15:
## Description
Prior to this change, you'd see error messages in log output about log registration.

This is `certmanager` complaining that it isn't hooked up to a logger. This fixes that problem.

## Changes
* Add registration with certmanager logger
## Testing
Review. See that it doesn't complain on startup when you're running with SSL certs.


ecordell (dismissed):
LGTM but looks like a lint issue

ecordell (approved):

PR #2341, opened 2025-04-18
Add additional logs to the flaky integration test runner

josephschorr:
Should help us further debug why it cannot connect when the container appears to start successfully and serve


tstirrat15 (dismissed):
LGTM

PR #2342, opened 2025-04-18
Minimal decoupling of validation file parsing and schema compilation

tstirrat15:
## Description
A subset of the work in #2268, this pulls out schema compilation from schema parsing so that we can pass in the caveat typesystem more easily.

## Changes
* Move schema compilation logic that was previously in parsing into its own function
* Remove CompiledSchema from yaml blocks used in parsing
* Amend callsites accordingly
## Testing
Review

### DIFF
--- a/pkg/validationfile/loader.go
+++ b/pkg/validationfile/loader.go
@@ -2,8 +2,11 @@ package validationfile
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
+
+	"github.com/ccoveille/go-safecast"
 
 	log "github.com/authzed/spicedb/internal/logging"
 	dsctx "github.com/authzed/spicedb/internal/middleware/datastore"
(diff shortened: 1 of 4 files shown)


pkg/validationfile/loader.go
```diff
@@ -170,3 +183,42 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
 
 	return &PopulatedValidationFile{schemaStr, objectDefs, caveatDefs, rels, files}, revision, err
 }
+
+// CompileSchema takes a SchemaWithPosition and returns the compiled schema, or else an error.
+// TODO: this is probably the wrong place for this, in part because it's coupling to the compiler
+// implementation.
+func CompileSchema(schemaWithPosition blocks.SchemaWithPosition) (*compiler.CompiledSchema, error) {
```
josephschorr, line 190: Have this take in the caveat type set as well, and pass it via option to the compile call below

josephschorr (approved):
LGTM

PR #2356, opened 2025-04-24
pin github actions

miparnisari:
See https://github.com/ossf/scorecard/blob/cd152cb6742c5b8f2f3d2b5193b41d9c50905198/docs/checks.md#pinned-dependencies


vroldanbet (approved):
LGTM, please note we also run dependabot updates for actions, we should keep an eye in case we stop seeing the updates coming through

PR #2357, opened 2025-04-24
pin dockerfile images

miparnisari:
See https://github.com/ossf/scorecard/blob/1e72b3ac6e297e29c63156145f945a464f087675/docs/checks.md#pinned-dependencies

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,7 +1,7 @@
 # vim: syntax=dockerfile
-ARG BASE=cgr.dev/chainguard/static:latest
+ARG BASE=cgr.dev/chainguard/static@sha256:1ff7590cbc50eaaa917c34b092de0720d307f67d6d795e4f749a0b80a2e95a2c
 
-FROM golang:1.24.2-alpine3.20 AS health-probe-builder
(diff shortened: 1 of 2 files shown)


Dockerfile.release
```diff
@@ -1,7 +1,7 @@
 # vim: syntax=dockerfile
-ARG BASE=cgr.dev/chainguard/static:latest
+ARG BASE=cgr.dev/chainguard/static@sha256:1ff7590cbc50eaaa917c34b092de0720d307f67d6d795e4f749a0b80a2e95a2c
```
vroldanbet, line 2: I was concerned we would not get updates for this image, which happens regularly. The chainguard folks describe this in [the docs](https://images.chainguard.dev/directory/image/static/overview):

>The static image is not versioned, since there's hardly any software in it that has a useful version. Most updates to the image are due to routine rebuilds of the packages included in the image, and don't constitute any functional change.

>In general it should be safe to rely on the :latest or :latest-glibc tag, but you may also want to pin the image reference by digest and use a tool like [digestabot](https://github.com/chainguard-dev/digestabot) to automate image digest updates.

Theoretically, if I understand https://github.com/dependabot/dependabot-core/issues/7387 correctly, dependabot should be able to update images pinned by sha too

vroldanbet (approved):
LGTM, left a note on potential issues updating chainguard image

PR #2366, opened 2025-04-28
make consistency middleware return invalid revision error as gRPC `InvalidArgument`

vroldanbet:
When a malformed zedtoken was provided, the API would return an `Unknown` gRPC error code. This changes the consistency middleware to return `InvalidArgument` error code.

It also adds a missing test verifying `OutOfBounds` error code is returned when the revision is past the GC Window, and turns every `fmt.Errorf` call into a `status.Errorf` call.


tstirrat15 (approved):
LGTM

PR #2369, opened 2025-04-28
Set pgx CancelRequestContextWatcherHandler.CancelRequestDelay to fixcancelation delay

vroldanbet:
This fixes a 100ms `QueryRelationship` delay caused pgx when handling context cancelation


josephschorr (approved):
LGTM

PR #2370, opened 2025-04-28
improvements to context severing

vroldanbet:
Several improvements to the use of context-severing strategies in the codebase

### DIFF
--- a/internal/datastore/context.go
+++ b/internal/datastore/context.go
@@ -3,30 +3,11 @@ package datastore
 import (
 	"context"
 
-	"go.opentelemetry.io/otel/trace"
-
-	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/datastore/options"
 	"github.com/authzed/spicedb/pkg/datastore/test"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 )
-
-// SeparateContextWithTracing is a utility method which allows for severing the
-// context between grpc and the datastore to prevent context cancellation from
-// killing database connections that should otherwise go back to the connection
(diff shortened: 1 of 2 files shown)


internal/graph/cursors.go
```diff
@@ -457,7 +457,7 @@ func (ls *parallelLimitedIndexedStream[Q]) forTaskIndex(ctx context.Context, ind
 	// Create a new cursor with cloned limits, because each child task which executes (in parallel) will need its own
 	// limit tracking. The overall limit on the original cursor is managed in completedTaskIndex.
 	childCI := currentCursor.withClonedLimits()
-	childContext, cancelDispatch := branchContext(ctx)
+	childContext, cancelDispatch := context.WithCancelCause(ctx)
```
ecordell, line 460: The previous version severed the context completely - did you mean to write WithoutCancel here?

internal/datastore/context.go
```diff
@@ -66,15 +47,15 @@ func (p *ctxProxy) IsStrictReadModeEnabled() bool {
 }
 
 func (p *ctxProxy) OptimizedRevision(ctx context.Context) (datastore.Revision, error) {
```
ecordell, line 49: Is this proxy even used anywhere anymore?

internal/graph/cursors.go
vroldanbet: I deliberately intended to _not_ sever the context. I couldn't find a reason to justify severed context. This is what prompted this PR: I observed that cancelation wasn't getting propagated to `ReverseQueryRelationship` calls in the data store, which means we could be _not _ freeing_ resources.

internal/datastore/context.go
vroldanbet: Yes, MySQL datastore still uses it.

Issue #2371, opened 2025-04-28
Typed Permissions


jzelinskie:
### Problem Statement

Today, relation definitions are required to be typed (e.g. `relation members: user`).
Permissions definitions implicitly adopt a type that is the union of any types referenced in its definition expression (e.g. `permission view = members + members->apikeys` `view` has type of `user | apikey`). But what if I wanted to enforce that only users can view permissions?

### Solution Brainstorm

Add syntax to permission definitions that'll throw errors when the type the expression doesn't evaluate to the same thing. For example, `permission view: user = members + members->apikeys` which would error because the expression is of type `user | apikey` and not just `user`.

artie-shevchenko:
that's coming soon, right?

vroldanbet:
@artie-shevchenko I think it may be about to be done, or done. @barakmich mind clarifying?

josephschorr:
This is done and released in https://github.com/authzed/spicedb/releases/tag/v1.45.1

PR #2374, opened 2025-04-29
Implement a typechecking function in schema

barakmich:
Addressing the implementation part of #2371  

Also opened a new issue stemming from the new `schema.Graph` object not covering all cases.


pkg/schema/type_check.go
```diff
@@ -0,0 +1,113 @@
...
+	"context"
+	"fmt"
+
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
+	corev1 "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+const ellipsesRelation = "..."
+
+func (ts *TypeSystem) GetTypesForRelation(ctx context.Context, defName string, relationName string) (*mapz.Set[string], error) {
```
josephschorr, line 13: Let's name this `GetRecursiveSubjectTypesForRelation` and add a doc comment explaining what it returns. I'd also change the return to be a slice of names, rather than the set

pkg/schema/type_check_test.go
```diff
@@ -0,0 +1,194 @@
...
+			definition user {}
+
+			definition organization {
+				relation member: user
+			}
+
+			definition resource {
+				relation viewer: organization#member
+				relation banned: user
+				permission view = viewer - banned
```
josephschorr, line 152: So this raises an interesting question: the target subject types of `view` here are *technically* `user` *and* `organization#member`, since you can issue a check for organization#member here. Question is: do we include that in this set?

josephschorr: I think we do include it and make it required to be specified

barakmich: `organization#member` is also a `user` though (or, really, the set of things that are on the other end of that relation -- and here we have computed subtypes again). I'm okay with it being specified, though

josephschorr: Yeah, but its valid, so I think we should add it

pkg/schema/type_check.go
```diff
@@ -0,0 +1,119 @@
...
+	id := fmt.Sprint(defName, "#", relationName)
+	if seen.Has(id) {
+		return nil, nil
+	}
+	seen.Add(id)
+	def, err := ts.GetDefinition(ctx, defName)
+	if err != nil {
+		return nil, err
+	}
+	rel, ok := def.GetRelation(relationName)
```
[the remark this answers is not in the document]
miparnisari: The `wsl` linter does this for you 😄  https://github.com/bombsimon/wsl

pkg/schema/type_check_test.go
```diff
@@ -0,0 +1,193 @@
...
+				relation member: user
+			}
+
+			definition resource {
+				relation viewer: organization#member
+			}
+		`,
+			expected: map[string][]string{
+				"organization#member": {"user"},
+				"resource#viewer":     {"user"},
```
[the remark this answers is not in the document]
miparnisari: I believe that, using the "terminal vs non-terminal" parlance, you need to return both terminal and non-terminal types. The reason being, imagine you have

```
definition user {}
definition organization {
	relation member: user
}

definition resource {
	relation aaa: organization#member
	relation bbb: user
	permission view = aaa & bbb
}
```

`user` is a terminal type, but `organization#member` is not.

If you issued `CheckPermission(doc:budget, view, organization:spicedb#member)`, you'd get zero results because there is no way of writing relation `b` for `organization:spicedb#member`.

Similarly, `CheckPermission(doc:budget, view, user:maria)`, would give you zero results because there is no way of writing relation `a for `user:maria`.

Based on the description in https://github.com/authzed/spicedb/issues/2371, I believe the point of this work is to warn people that relation `view` will effectively never return true because `len_of_terminal-and_non_terminal_types(resource#view) > 1`. (Or at least this is what I hope it's used for!)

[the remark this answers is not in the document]
miparnisari: Sorry, I deleted my previous comment because it was wrong 😄 

If the use case for this work is warn people that their permission doesn't make sense because it will never ever return `true`, I believe that you should not include `organization#member` because it's not a terminal type.

E.g. if your permission is

```
relation aaa: user
relation bbb: organization#member
permission view = aaa & bbb
```

`view` can return true for some users (if the `user` is member of some organization)

However, if it was

```
relation aaa: user
relation bbb: apiKey
permission view = aaa & bbb
```


`view` will never return true no matter what relationships you write.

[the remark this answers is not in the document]
josephschorr: It necessary because it is a valid subject type for the check, and for LS. The type annotation is indicating the types of allowed subjects, so it should be present.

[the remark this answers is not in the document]
miparnisari: But we're not gonna use this code to validate the subject types for Check and LS, right?

pkg/schema/type_check.go
```diff
@@ -0,0 +1,144 @@
...
+	"context"
+	"fmt"
+
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
+	corev1 "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+const ellipsesRelation = "..."
+
+// GetRecursiveTypesForRelation returns, for a given definition and relation, are the potential
```
josephschorr: are -> all

pkg/schema/type_check.go
```diff
@@ -0,0 +1,144 @@
...
+// subject definition names of that relation.
```
josephschorr: subject definition names -> terminal subject types?

pkg/schema/type_check.go
```diff
@@ -0,0 +1,144 @@
...
+func (ts *TypeSystem) GetRecursiveTypesForRelation(ctx context.Context, defName string, relationName string) ([]string, error) {
```
josephschorr: if we're going to have this only return the terminals, maybe rename to `GetRecursiveTerminalTypesForRelation`

pkg/schema/type_check.go
```diff
@@ -0,0 +1,144 @@
...
+	set, err := ts.getTypesForRelationInternal(ctx, defName, relationName, seen, false)
+	if err != nil {
+		return nil, err
+	}
+	return set.AsSlice(), nil
+}
+
+// GetRecursiveSubtypesForRelation returns, for a given definition and relation, are the potential
+// subject definition names of that relation, as well as any relation subtypes (eg, `group#member`) that may occur.
+func (ts *TypeSystem) GetRecursiveSubtypesForRelation(ctx context.Context, defName string, relationName string) ([]string, error) {
```
josephschorr: maybe `GetFullRecursiveSubjectTypesForRelation`

pkg/schema/type_check.go
```diff
@@ -0,0 +1,144 @@
...
+	return nil, asTypeError(NewMissingAllowedRelationsErr(defName, relationName))
+}
+
+func (ts *TypeSystem) getTypesForInfo(ctx context.Context, defName string, rel *corev1.TypeInformation, seen *mapz.Set[string], addRelations bool) (*mapz.Set[string], error) {
+	out := mapz.NewSet[string]()
+	for _, dr := range rel.GetAllowedDirectRelations() {
+		if dr.GetRelation() == ellipsesRelation {
+			out.Add(dr.GetNamespace())
+		} else if dr.GetRelation() != "" {
+			if addRelations {
```
josephschorr: addRelations -> addNonTerminalTypes

pkg/schema/type_check_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+			res := ResolverForCompiledSchema(*schema)
+			ts := NewTypeSystem(res)
+			_, err = ts.GetRecursiveTypesForRelation(context.Background(), "resource", "view")
+			require.Error(t, err)
+			_, err = ts.GetRecursiveSubtypesForRelation(context.Background(), "resource", "view")
+			require.Error(t, err)
+		})
+	}
+}
```
josephschorr, line 469: add a TODO for me to wire this call into the consistency tests as well, so I can validate over them all too

josephschorr (approved):
LGTM

PR #2376, opened 2025-04-30
Improve CRDB index forcing logic to be more fine-grain

josephschorr:
Also elides forcing the index in cases where we aren't certain the existing indexes should be used


barakmich (approved):
This is fine, but in the long run we shouldn't need to write these heuristics

PR #2380, opened 2025-05-02
Cache Features call in CRDB and skip outputting the check error

josephschorr:
Makes the logs less noisy and avoids re-doing the features work on every `Watch` call

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/IBM/pgxpoolprometheus"
@@ -284,7 +285,9 @@ type crdbDatastore struct {
 	beginChangefeedQuery string
 	transactionNowQuery  string
 
-	featureGroup singleflight.Group[string, *datastore.Features]
+	featuresGroup  singleflight.Group[string, *datastore.Features]
+	cachedFeatures *datastore.Features // GUARDED_BY(featuresLock)
+	featuresLock   sync.Mutex
 
 	pruneGroup           *errgroup.Group
 	ctx                  context.Context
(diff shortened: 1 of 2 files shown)


internal/datastore/crdb/crdb.go
```diff
@@ -516,9 +519,25 @@ func (cds *crdbDatastore) OfflineFeatures() (*datastore.Features, error) {
 }
 
 func (cds *crdbDatastore) Features(ctx context.Context) (*datastore.Features, error) {
-	features, _, err := cds.featureGroup.Do(ctx, "", func(ictx context.Context) (*datastore.Features, error) {
+	cds.featuresLock.Lock()
+	cached := cds.cachedFeatures
+	cds.featuresLock.Unlock()
```
ecordell, line 524: technically this permission can change over time, the account spicedb is using is granted changefeed permission after starting up.

internal/datastore/postgres/common/pgx.go
```diff
@@ -74,12 +75,19 @@ func ConfigurePGXLogger(connConfig *pgx.ConnConfig) {
...
 			if errArg, ok := data["err"]; ok {
 				err, ok := errArg.(error)
 				if ok && (common.IsCancellationError(err) || IsQueryCanceledError(err) || IsSerializationError(err) || IsReplicationLagError(err)) {
 					logger.Log(ctx, tracelog.LogLevelDebug, msg, data)
 					return
 				}
+
+				// NOTE: this error is raised *on purpose* by the CRDB datastore when checking if watch
+				// is enabled. It is not a real error, and therefore should not be logged.
+				if strings.Contains(err.Error(), "negative durations are not accepted") {
```
tstirrat15, line 88: Only nit would be that it'd be nice if there was something that we could `As` here.

tstirrat15 (approved):
LGTM, assuming that we're string checking because that was the only avenue available to us.

internal/datastore/crdb/crdb.go
josephschorr: Yes, it could, but at that point it would error anyway

ecordell: It could start without access and then gain access after startup

josephschorr: Yeah, but that's also fine. We don't need to support that case; restart the pod

PR #2381, opened 2025-05-06
pin github action and dockerfile.release base image

miparnisari:
Thanks to https://securityscorecards.dev/viewer/?uri=github.com/authzed/spicedb I noticed that I forgot to pin a couple of places.

The `Dockerfile.release` pinning suggestion is also recommended by https://hadolint.github.io/hadolint/.

(Also opened https://github.com/ossf/scorecard/issues/4617)


tstirrat15:
We should add to our projects generally :thinking: not a today project but it'd be nice to have.

tstirrat15 (approved):
LGTM

miparnisari:
> We should add to our projects generally 🤔 not a today project but it'd be nice to have.

add _what_ to our projects?

PR #2382, opened 2025-05-06
remove direct dependency to aws sdk v1

miparnisari:
Reference: https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/migrate-gosdk.html

Also: https://github.com/aws/aws-sdk-go?tab=readme-ov-file#warning-this-sdk-is-in-maintenance-mode


go.mod
```diff
@@ -49,6 +51,7 @@ require (
 	github.com/go-sql-driver/mysql v1.9.2
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/snappy v1.0.0
+	github.com/golangci/golangci-lint/v2 v2.1.5
```
miparnisari, line 54: This shouldn't be here 🤔  This should be a tooling dependency...

tstirrat15: It is a tooling dependency, but it doesn't work properly unless you either entirely externalize it or run it from the package root with the root's go.mod. This was a part of the back-and-forth with the golangci-lint v2 PRs.

internal/services/v0/sharestore.go
```diff
@@ -132,7 +125,7 @@ func (s3s *s3ShareStore) createBucketForTesting() error {
 		Bucket: aws.String(s3s.bucket),
 	}
 
-	_, err := s3s.s3Client.CreateBucket(cparams)
+	_, err := s3s.s3Client.CreateBucket(context.TODO(), cparams)
```
tstirrat15, line 128: Should this be a `Background` or something?

internal/services/v0/sharestore.go
```diff
@@ -158,16 +151,16 @@ func (s3s *s3ShareStore) LookupSharedByReference(reference string) (SharedDataV2
...
-	result, err := s3s.s3Client.GetObjectWithContext(ctx, &s3.GetObjectInput{
+	result, err := s3s.s3Client.GetObject(ctx, &s3.GetObjectInput{
 		Bucket: aws.String(s3s.bucket),
 		Key:    aws.String(key),
 	})
 	if err != nil {
-		var aerr awserr.Error
-		if errors.As(err, &aerr) && aerr.Code() == s3.ErrCodeNoSuchKey {
+		var nsk *types.NoSuchKey
+		if errors.As(err, &nsk) {
```
tstirrat15, line 160: if the type is a pointer, should this be a direct reference (sans &)?

tstirrat15 (commented):
See comments, otherwise LGTM (will review again)

internal/services/v0/sharestore.go
miparnisari: No, based on the example here: https://pkg.go.dev/errors#example-As

it shouldn't

tstirrat15 (approved):
LGTM

PR #2385, opened 2025-05-08
Add additional forced CRDB index in special case of no object IDs

josephschorr:
Before this change, the non-ID index would never be selected by read or delete relationships in CRDB. Now, it is used for the single, special case

### DIFF
--- a/internal/datastore/crdb/schema/indexes.go
+++ b/internal/datastore/crdb/schema/indexes.go
@@ -88,6 +88,19 @@ func IndexingHintForQueryShape(schema common.SchemaInformation, qs queryshape.Sh
 
 // IndexForFilter returns the index to use for a given relationships filter or nil if no index is forced.
 func IndexForFilter(schema common.SchemaInformation, filter datastore.RelationshipsFilter) *common.IndexDefinition {
+	// Special case: if the filter specifies the resource type and relation and the subject type and relation, then
+	// the schema diff index can be used.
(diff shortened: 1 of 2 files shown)


internal/datastore/crdb/schema/indexes.go
```diff
@@ -88,6 +88,19 @@ func IndexingHintForQueryShape(schema common.SchemaInformation, qs queryshape.Sh
 
 // IndexForFilter returns the index to use for a given relationships filter or nil if no index is forced.
 func IndexForFilter(schema common.SchemaInformation, filter datastore.RelationshipsFilter) *common.IndexDefinition {
+	// Special case: if the filter specifies the resource type and relation and the subject type and relation, then
+	// the schema diff index can be used.
+	if filter.OptionalResourceType != "" &&
+		filter.OptionalResourceRelation != "" &&
+		len(filter.OptionalSubjectsSelectors) == 1 &&
+		filter.OptionalSubjectsSelectors[0].OptionalSubjectType != "" &&
+		filter.OptionalSubjectsSelectors[0].RelationFilter.NonEllipsisRelation != "" &&
```
vroldanbet, line 97: Not a blocker, just an idea: I wish we could turn this into some fluent API helper

vroldanbet (approved):
LGTM, please add a description to the body on the index that was being selected, and the ones that should be selected instead

PR #2387, opened 2025-05-08
add more default linters

miparnisari:
Follow-up of https://github.com/authzed/spicedb/pull/2382#discussion_r2078341249, we went deep and found that we're missing a bunch of default linters.

image

I also added a few other ones.


tstirrat15 (approved):
LGTM

tstirrat15 (approved):
LGTM

PR #2389, opened 2025-05-09
Secondary dispatch improvements

josephschorr:


internal/dispatch/remote/cluster.go
```diff
@@ -55,10 +55,13 @@ var primaryDispatch = prometheus.NewCounterVec(prometheus.CounterOpts{
 
 // defaultStartingPrimaryHedgingDelay is the delay used by default for primary calls (when secondaries are available),
 // before statistics are available to determine the actual delay.
-const defaultStartingPrimaryHedgingDelay = 5 * time.Millisecond
+const defaultStartingPrimaryHedgingDelay = 1 * time.Millisecond
+
+// maximumHedgingDelay is the maximum delay used for hedging requests to the primary dispatcher.
+const maximumHedgingDelay = 5 * time.Millisecond
```
vroldanbet, line 61: I'd suggest that you make this configurable. Under certain load conditions this may not be enough

vroldanbet (approved):

internal/dispatch/remote/cluster.go
josephschorr: The concern is that if we're going much beyond this, then the overhead for normal calls will be too large vs the "extra work" we're doing

vroldanbet: I think it's entirely context dependent. We could have a data store dispatches in average in 10ms, and a secondary that does so in 7ms. You still get net gains out of dispatching to the secondary, but the cap won't allow it.

josephschorr: Remember - this is the delay from the *secondary* dispatch, which means it should never be more than a few ms max; if it is more than that, then we don't want to delay the primary anyway, since the secondary is being overwhelmingly slow

internal/dispatch/remote/cluster.go
```diff
@@ -881,6 +964,46 @@ func (srst *supportedResourceSubjectTracker) updateForError(err error) {
...
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
+		log.Trace().Str("request-key", s.reqKey).Msg("primary dispatch was canceled before waiting")
+		return
+
+	case <-timer.C:
+		log.Trace().Str("request-key", s.reqKey).Msg("primary dispatch finished waiting")
+		hedgeActualWaitHistogram.WithLabelValues(s.reqKey).Observe(s.waitTime.Seconds())
```
vroldanbet, line 1005: This should be in a defer block, because if the context is cancelled, we are not capturing the actual wait in that case.

internal/dispatch/remote/cluster.go
```diff
@@ -881,6 +964,46 @@ func (srst *supportedResourceSubjectTracker) updateForError(err error) {
...
+}
+
+func (s *primarySleeper) sleep(ctx context.Context) {
+	if s.waitTime <= 0 {
+		return
+	}
+
+	s.lock.Lock()
+	ctx, s.cancel = context.WithCancel(ctx)
+	s.lock.Unlock()
```
vroldanbet: Instead of using a lock here, you could create the context on `newPrimarySleeper`, that way you'll know there is no possibility of races.

internal/dispatch/remote/cluster.go
```diff
@@ -295,50 +339,47 @@ func dispatchSyncRequest[Q requestMessage, S responseMessage](
...
-						return
-					}
+			default:
+				log.Trace().Str("secondary", secondary.Name).Msg("running secondary dispatch")
+				startTime := time.Now()
+				resp, err := handler(withTimeout, secondary.Client)
+				endTime := time.Now()
+
+				// Cancel the primary sleeper as soon as any result comes back.
+				primarySleeper.cancelSleep()
```
vroldanbet: Why do we cancel the sleeper unconditionally? If the secondary has returned a legit result, the primary could be executed before **the result is sent over the channel**, which is what cancels the context. As a consequence, it will trigger an unnecessary primary request.

Please note that `primarySleeper.cancelSleep` does not cancel `withTimeout` context: the `sleep` method is creating a new context, and thus that cancellation is not propagated.

You could solve this by passing the top-level context and the cancellation function to the `primarySleeper` helper instead of creating a new context.

vroldanbet (changes requested):

internal/dispatch/remote/cluster.go
josephschorr: Exactly: if the context is canceled, it is not the actual wait time, so we do not want to log it. The actual wait time is only logged once the timer has completed

internal/dispatch/remote/cluster.go
josephschorr: If I do so, it is not resuable

internal/dispatch/remote/cluster.go
josephschorr: As soon as the secondary is complete, the primary should execute. If the secondary has returned an error, we want the primary to execute immediately and if it is not going to return an error... we want it to execute immediately anyway, since it won't start (since it just received an result). In either scenario, we want the primary to stop waiting

josephschorr: Moved to only canceling the sleep on secondary error, to prevent the primary from *ever* even starting

vroldanbet (approved):

PR #2394, opened 2025-05-16
fix(CI): CLA GitHub Action permissions

squat:
The cla-assistant GitHub Action is currently failing with:
```
Error: Could not update the JSON file: Error occured when creating a pull request comment: Resource not accessible by integration
```
I suspect this was caused by a recent change to the workflow's
permissiosn in https://github.com/authzed/spicedb/pull/2353.

This should fix the issue, according to the workflow's
upstream's upstream
https://github.com/contributor-assistant/github-action?tab=readme-ov-file#configure-contributor-license-agreement-within-two-minutes.

Signed-off-by: squat <lserven@gmail.com>

### DIFF
--- a/.github/workflows/cla.yaml
+++ b/.github/workflows/cla.yaml
@@ -13,10 +13,10 @@ on:  # yamllint disable-line rule:truthy
     types:
       - "checks_requested"
 permissions:
-  actions: write
-  contents: read
-  pull-requests: write
-  statuses: write
+  actions: "write"
+  contents: "write"
+  pull-requests: "write"
+  statuses: "write"
 jobs:
   cla:
     name: "Check Signature"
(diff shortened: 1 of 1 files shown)


squat:
Hmm this seems not to be the fix. There is probably something more going on in the permissions for the GitHub PAT.

.github/workflows/cla.yaml
```diff
@@ -13,7 +13,10 @@ on:  # yamllint disable-line rule:truthy
     types:
       - "checks_requested"
 permissions:
-  contents: "read"
+  actions: "write"
+  contents: "write"
```
miparnisari, line 17:
```suggestion
  contents: "read" # CLA signatures are stored in https://github.com/authzed/cla
```

miparnisari (approved):

PR #2400, opened 2025-05-22
migrate to AWS sdk v2 while keeping support for Google storage

miparnisari:
## Description

- deprecate AWS SDK v1 in favor of v2
- include a fix for when we use AWS SDK v2 with Google Storage: https://github.com/aws/aws-sdk-go-v2/issues/1816

This PR deviates from our original idea of using the Google SDK when `--share-store=google`, because it would mean a breaking change: the Google SDK doesn't support an access key / secret key pair for authentication. I see this PR as a reasonable alternative.

## Testing

TO DO. I don't know how to get the Playground to talk to my local devtools instance.


miparnisari:
Cancelling this PR as the playground now embeds the AWS SDK 👏 https://github.com/authzed/playground/blob/f99d6390b3aef74c9f4e8e7fdcd7dab060392d4f/api/share.ts#L1

PR #2402, opened 2025-05-23
Handle both kinds of cancelation errors in test, part 2

miparnisari:
Companion of https://github.com/authzed/spicedb/pull/2390, we just missed one other place.


josephschorr (approved):
LGTM

PR #2403, opened 2025-05-23
include integration tests in code coverage

miparnisari:


magefiles/test.go
```diff
@@ -8,33 +8,21 @@ import (
...
+func (t Test) All(ctx context.Context) {
 	ds := Testds{}
 	c := Testcons{}
-	cover := parseCommandLineFlags()
-	ctx = ctxMyKey.WithValue(ctx, cover)
 	mg.CtxDeps(ctx, t.Unit, t.Integration, t.Steelthread, t.Image, t.Analyzers,
 		ds.Crdb, ds.Postgres, ds.Spanner, ds.Mysql,
 		c.Crdb, c.Spanner, c.Postgres, c.Mysql)
-
-	if !cover {
```
miparnisari, line 33: Presumably, nobody cares about `mage test:all` running coverage or not. I assume that we can leave coverage collection to CI/CD

.github/workflows/build-test.yaml
```diff
@@ -106,8 +105,14 @@ jobs:
         with:
           username: "${{ env.DOCKERHUB_PUBLIC_USER }}"
           password: "${{ env.DOCKERHUB_PUBLIC_ACCESS_TOKEN }}"
-      - name: "Integration tests"
-        run: "go run mage.go test:integration"
+      - name: "Integration tests with coverage"
+        run: "go run mage.go test:integrationCover"
+      - name: "Coverage"
```
miparnisari, line 110: I don't love that this step is everywhere, but it works for now. In the future we can probably cache the various `coverage.txt` files with different names and then push them all in one go.

josephschorr (approved):
LGTM

PR #2408, opened 2025-05-23
update codecov.yaml

miparnisari:
Validated via `curl -X POST --data-binary @codecov.yaml https://codecov.io/validate`


josephschorr (dismissed):
LGTM

PR #2424, opened 2025-06-02
remove serve-devtools command

miparnisari:
## Description

After https://github.com/authzed/playground/commit/f99d6390b3aef74c9f4e8e7fdcd7dab060392d4f, Playground no longer depends on code defined here. 

So i'm deprecating the `serve-devtools` command and removing code that it used.


josephschorr (approved):
LGTM

PR #2430, opened 2025-06-05
Fix flaky reporter test

josephschorr:


internal/telemetry/reporter.go
```diff
@@ -134,7 +135,7 @@ func RemoteReporter(
 	if _, err := url.Parse(endpoint); err != nil {
 		return nil, fmt.Errorf("invalid telemetry endpoint: %w", err)
 	}
-	if interval < MinimumAllowedInterval {
+	if !strings.Contains(endpoint, "127.0.0.1") && interval < MinimumAllowedInterval {
```
barakmich, line 138: Technically, should check the IP for "localhost-y" -- eg, `::1`

barakmich (approved):

internal/telemetry/reporter.go
josephschorr: It should but its only there to allow the test to bypass it; if it breaks, we can change it

PR #2434, opened 2025-06-09
revert pgx context cancellation

vroldanbet:
reverts https://github.com/authzed/spicedb/pull/2294

We've observed unreliable behavior from pgx cancellation:
- connections were returned to the pool before the cancelation error was returned. This led to subsequent API requests fail with `SQLSTATE 57014`
- Context cancellation can cause the system enter an unrecoverable feedback loop were the connection pool is being starved because of deadlines.

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -21,6 +21,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"resenje.org/singleflight"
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -0,0 +1,5 @@
+package spicedb
```
josephschorr: ?

vroldanbet: whoops, sorry, some leftover garbage I had locally

josephschorr (approved):
LGTM

PR #2437, opened 2025-06-09
Add further filtering of forced indexes on CRDB to discount other shapes

josephschorr:


internal/datastore/crdb/schema/indexes.go
```diff
@@ -103,12 +104,16 @@ func IndexForFilter(schema common.SchemaInformation, filter datastore.Relationsh
...
+					return nil // Cannot use an index with a prefix and a relation.
+				}
+
 				resourceFieldDepth = 3
 			}
 		}
 	}
 
-	subjectFieldDepth := 0
+	subjectFieldDepths := mapz.NewSet[int]()
```
tstirrat15, line 116: I'm not entirely clear on the intent of using a map here - I can see that the semantic is that there should only be  a single field depth, but I'm not clear on when there could be multiple, or why that would indicate that an index shouldn't be used.

tstirrat15 (approved):
LGTM, appreciate the test coverage

internal/datastore/crdb/schema/indexes.go
josephschorr: Its a set: we need to make sure all filters have the same depth

internal/datastore/crdb/schema/indexes.go
```diff
@@ -120,7 +125,16 @@ func IndexForFilter(schema common.SchemaInformation, filter datastore.Relationsh
...
+		subjectFieldDepths.Add(sfd)
+	}
+
+	if subjectFieldDepths.Len() > 1 {
+		return nil
+	}
+
+	subjectFieldDepth := 0
+	if !subjectFieldDepths.IsEmpty() {
+		subjectFieldDepth = subjectFieldDepths.AsSlice()[0]
```
miparnisari, line 137: ~I don't understand this line - you have an unordered set of depths, you convert it to a slice (again, unordered) and then arbitrarily pick the first one? why?~

EDIT: sorry, i get it now. The set doesn't allow duplicate elements.

josephschorr:
```
	if subjectFieldDepths.Len() > 1 {
		return nil
	}
```
right before it makes it so that there can only be a single element

miparnisari (approved):

PR #2440, opened 2025-06-10
Change the schema for CRDB to have the subjects sort match the index field order

josephschorr:


vroldanbet (commented):
Shouldn't this made the index selection middleware for tests fail if it was selecting the wrong index?

josephschorr:
> Shouldn't this made the index selection middleware for tests fail if it was selecting the wrong index?

It wasn't selecting the wrong index, but rather using the cursor fields in the wrong order, which still works, but it likely slower

vroldanbet (approved):

PR #2446, opened 2025-06-11
do not count closed conns due to cancellation toward health check

vroldanbet:
We noticed that when clients called APIs that propagate cancellation errors like `WriteRelationships`, a tight deadline could cause the write connection pool to get starved. This is because the `RetryPool` considers an error to count toward the health check when it has caused a connection to be closed.

This commit modifies the code to ensure that this is ignored when the connection is closed as part of context cancellation. This should prevent marking as unhealthy a node that is observing connections being closed, which is caused by the client's cancellation or deadlines. The idea is to prevent the client from inducing this condition.

### DIFF
--- a/.github/workflows/cla.yaml
+++ b/.github/workflows/cla.yaml
@@ -22,7 +22,7 @@ jobs:
     name: "Check Signature"
     runs-on: "depot-ubuntu-24.04-small"
     steps:
-      - uses: "authzed/actions/cla-check@9013d08e1002d122cc87f21d9ed43063555642d0" # main
+      - uses: "authzed/actions/cla-check@391defc4658e3e4ac6e53ba66da5b90a3b3f80e2" # main
         with:
           github_token: "${{ secrets.GITHUB_TOKEN }}"
           cla_assistant_token: "${{ secrets.CLA_ASSISTANT_ACCESS_TOKEN }}"
--- a/internal/datastore/crdb/pool/pool.go
+++ b/internal/datastore/crdb/pool/pool.go
@@ -33,8 +33,28 @@ type ctxDisableRetries struct{}
 
 var CtxDisableRetries ctxDisableRetries
 
+var isConnClosed = isConnClosedFunc
+
+func isConnClosedFunc(conn *pgxpool.Conn) bool {
+	return conn.Conn().IsClosed()
+}
+
+var getConn = getConnFunc
+
+func getConnFunc(conn *pgxpool.Conn) *pgx.Conn {
+	return conn.Conn()
+}
+
+type Pool interface {
+	Config() *pgxpool.Config
+	AcquireAllIdle(ctx context.Context) []*pgxpool.Conn
+	Close()
+	Stat() *pgxpool.Stat
+	Acquire(ctx context.Context) (c *pgxpool.Conn, err error)
+}
+
 type RetryPool struct {
-	pool          *pgxpool.Pool
+	pool          Pool
 	id            string
 	healthTracker *NodeHealthTracker
 
@@ -158,7 +178,7 @@ func (p *RetryPool) MinConns() uint32 {
 // connection on error, or retrying on a retryable error.
 func (p *RetryPool) ExecFunc(ctx context.Context, tagFunc func(ctx context.Context, tag pgconn.CommandTag, err error) error, sql string, arguments ...any) error {
 	return p.withRetries(ctx, func(conn *pgxpool.Conn) error {
-		tag, err := conn.Conn().Exec(ctx, sql, arguments...)
+		tag, err := getConn(conn).Exec(ctx, sql, arguments...)
 		return tagFunc(ctx, tag, err)
 	})
 }
@@ -167,7 +187,7 @@ func (p *RetryPool) ExecFunc(ctx context.Context, tagFunc func(ctx context.Conte
 // connection on error, or retrying on a retryable error.
 func (p *RetryPool) QueryFunc(ctx context.Context, rowsFunc func(ctx context.Context, rows pgx.Rows) error, sql string, optionsAndArgs ...any) error {
 	return p.withRetries(ctx, func(conn *pgxpool.Conn) error {
-		rows, err := conn.Conn().Query(ctx, sql, optionsAndArgs...)
+		rows, err := getConn(conn).Query(ctx, sql, optionsAndArgs...)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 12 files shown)


internal/datastore/crdb/pool/pool_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/time/rate"
+)
+
+type mockPgxPool struct {
```
josephschorr: this is a fake, not a mock :)

internal/datastore/crdb/pool/pool.go
```diff
@@ -33,8 +33,28 @@ type ctxDisableRetries struct{}
...
+	return conn.Conn().IsClosed()
+}
+
+var getConn = getConnFunc
+
+func getConnFunc(conn *pgxpool.Conn) *pgx.Conn {
+	return conn.Conn()
+}
+
+type Pool interface {
```
josephschorr: doc comments

internal/datastore/crdb/pool/pool.go
```diff
@@ -33,8 +33,28 @@ type ctxDisableRetries struct{}
 
 var CtxDisableRetries ctxDisableRetries
 
+var isConnClosed = isConnClosedFunc
```
josephschorr: why are these function refs? Please document

internal/datastore/crdb/pool/pool.go
```diff
@@ -298,10 +321,15 @@ func (p *RetryPool) withRetries(ctx context.Context, fn func(conn *pgxpool.Conn)
 
 			common.SleepOnErr(ctx, err, retries)
 
-			conn, err = p.acquireFromDifferentNode(ctx, nodeID)
-			if err != nil {
-				return fmt.Errorf("error acquiring connection from pool after retry %d: %w", retries, err)
+			// if this is the last attempt to retry, do not unnecessarily attempt to acquire a new connection
+			if retries != maxRetries {
```
josephschorr: perhaps `retries < maxRetries`?

internal/datastore/crdb/pool/pool.go
```diff
@@ -424,3 +447,18 @@ func wrapRetryableError(ctx context.Context, err error) error {
 	}
 	return err
 }
+
+// returns error if either the context or the provided error is of type context.DeadlineExceeded or context.Canceled
+func contextError(ctx context.Context, err error) error {
```
josephschorr: maybe have it return the error and a bool?

internal/datastore/crdb/pool/pool_test.go
```diff
@@ -0,0 +1,103 @@
...
+	return &pgxpool.Conn{}, nil
+}
+
+func TestWithRetries(t *testing.T) {
+	originalClosedFunc := isConnClosed
+	t.Cleanup(func() {
+		isConnClosed = originalClosedFunc
+	})
+
+	isConnClosed = func(conn *pgxpool.Conn) bool {
```
josephschorr: can we put these on the RetryPool itself?

internal/datastore/crdb/pool/pool.go
vroldanbet: They are there just to be able to test the whole thing. I can document that's the reason, does it help?

internal/datastore/crdb/pool/pool.go
vroldanbet: Why? `nil` value is enough signal. It's an anti-pattern to have functions that have error and a boolean

internal/datastore/crdb/pool/pool.go
josephschorr: If there isn't another way to replace them, yes

internal/datastore/crdb/pool/pool.go
josephschorr: Its an unexpected pattern. If you're going to do so, maybe call it `asContextError`?

internal/datastore/crdb/pool/pool_test.go
vroldanbet: what do you mean? As fields in the struct?

josephschorr: Yeah, with good defaults

.github/workflows/cla.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     name: "Check Signature"
     runs-on: "depot-ubuntu-24.04-small"
     steps:
-      - uses: "authzed/actions/cla-check@9013d08e1002d122cc87f21d9ed43063555642d0" # main
+      - uses: "authzed/actions/cla-check@391defc4658e3e4ac6e53ba66da5b90a3b3f80e2" # main
```
miparnisari, line 25: 🤔 why update this one?

vroldanbet: It's not strictly necessary.

internal/datastore/crdb/pool/pool_test.go
vroldanbet: fixed in https://github.com/authzed/spicedb/pull/2446/commits/353d29c35074b2372fc180cb10d3068cdee1941c

internal/datastore/crdb/pool/pool.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/2446/commits/353d29c35074b2372fc180cb10d3068cdee1941c

internal/datastore/crdb/pool/pool.go
vroldanbet: renamed in https://github.com/authzed/spicedb/pull/2446/commits/85c9dd9ab65a6ae4fd031adb26acabebf3d7868e

internal/datastore/crdb/pool/pool.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/2446/commits/85c9dd9ab65a6ae4fd031adb26acabebf3d7868e

internal/datastore/crdb/pool/pool.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/2446/commits/85c9dd9ab65a6ae4fd031adb26acabebf3d7868e

vroldanbet:
Found the behaviour during an upgrade changed. Need to go back to the drawing board.

PR #2451, opened 2025-06-12
LookupResources v3, based on a new cursored iterator library

josephschorr:
Implement a new cursorediterator library for defining cursor-based iterator trees that also handle context, errors and cancelation transparently.

Then use it to implement LookupResources3

### DIFF
--- a/internal/graph/lookupresources3.go
+++ b/internal/graph/lookupresources3.go
@@ -18,6 +18,7 @@ import (
 	"github.com/authzed/spicedb/internal/graph/computed"
 	"github.com/authzed/spicedb/internal/graph/hints"
 	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
+	"github.com/authzed/spicedb/internal/telemetry/otelconv"
 	caveattypes "github.com/authzed/spicedb/pkg/caveats/types"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/datastore/options"
@@ -198,7 +199,7 @@ type ValidatedLookupResources3Request struct {
 // This allows the operation to resume at any point in the complex iteration hierarchy without
 // losing progress or re-processing completed work.
 func (crr *CursoredLookupResources3) LookupResources3(req ValidatedLookupResources3Request, stream dispatch.LookupResources3Stream) error {
-	ctx, span := tracer.Start(stream.Context(), "LookupResources3")
+	ctx, span := tracer.Start(stream.Context(), otelconv.EventDispatchLookupResources3)
 	defer span.End()
 
 	if req.TerminalSubject == nil {
@@ -299,7 +300,7 @@ func (crr *CursoredLookupResources3) unlimitedLookupResourcesIter(
 	ctx context.Context,
 	lctx lr3ctx,
 ) iter.Seq2[result, error] {
-	ctx, span := tracer.Start(ctx, "unlimitedLookupResourcesIter")
+	ctx, span := tracer.Start(ctx, otelconv.EventDispatchLR3UnlimitedResults)
 	defer span.End()
 
 	var cursor cter.Cursor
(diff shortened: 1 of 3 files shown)


internal/dispatch/graph/lookupresources3_test.go
```diff
@@ -0,0 +1,1362 @@
...
+				genRels("document", "editor", "user", "tom", 1510),
+			),
+			RR("document", "view"),
+			ONR("user", "tom", "..."),
+			nil,
+			genResourceIds("document", 1510),
+			nil,
+		},
+		{
+			"basic exclusion",
```
barakmich, line 373: this doesn't actually test exclusion, but could with two other tuples (another viewer, and a banned)

internal/dispatch/graph/lookupresources3_test.go
```diff
@@ -0,0 +1,1362 @@
...
+  			 }`,
+			genRelsWithCaveat("document", "viewer", "user", "tom", "somecaveat", map[string]any{"somecondition": 42}, 0, 2450),
+			RR("document", "view"),
+			ONR("user", "tom", "..."),
+			nil,
+			genResourceIds("document", 2450),
+			nil,
+		},
+		{
+			"excluded items",
```
barakmich, line 448: combine with previous exclusion test?

internal/dispatch/graph/lookupresources3_test.go
```diff
@@ -0,0 +1,1362 @@
...
+				genRels("document", "editor", "user", "tom", 510),
+			),
+			RR("document", "view"),
+			ONR("user", "tom", "..."),
+			nil,
+			genResourceIds("document", 510),
+			nil,
+		},
+		{
+			"union and excluded union",
```
barakmich, line 408: also doesn't do what it says

proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -123,6 +124,26 @@ message DispatchLookupResources2Response {
...
 }
 
+message DispatchLookupResources3Request {
+  ResolverMeta metadata = 1 [(validate.rules).message.required = true];
+
+  core.v1.RelationReference resource_relation = 2 [(validate.rules).message.required = true];
+  core.v1.RelationReference subject_relation = 3 [(validate.rules).message.required = true];
+  repeated string subject_ids = 4;
+
+  core.v1.ObjectAndRelation terminal_subject = 5 [(validate.rules).message.required = true];
```
barakmich, line 134: This name is a little confusing to me at first glance -- specifically how it's different than `subject_ids` (I suspect it's on redispatch but a comment might make things clearer)

internal/graph/lookupresources3.go
```diff
@@ -0,0 +1,1230 @@
...
+) iter.Seq2[result, error] {
+	ctx, span := tracer.Start(ctx, "dispatchIter")
+	defer span.End()
+
+	if !rm.isPopulated() {
+		// If there are no relationships to dispatch, return an empty iterator.
+		return cter.UncursoredEmpty[result]()
+	}
+
+	subjectIDs := rm.subjectIDsToDispatch()
```
barakmich, line 802: I know we talked about it, but if this gets very wide, this could start to get unwieldy -- which I get, it's chunked so there's a fixed size, but then the cursor holds the chunk... are we sure the pipes are configured right to support it? (And/or appropriate logging to know if the request ever made it, or was rejected or dropped by, eg, a service mesh configured with too low a request body size)

josephschorr:
> `internal/dispatch/graph/lookupresources3_test.go` doesn't seem to always test what it says on the tin...

Its a duplicate of lookupresources2_test.go, changed to call LR3.

internal/dispatch/graph/lookupresources3_test.go
josephschorr: It does; it tests that the exclusion operation works as expected, even with a nil set on the right side

internal/dispatch/graph/lookupresources3_test.go
josephschorr: This one does have items on the right hand side

internal/dispatch/graph/lookupresources3_test.go
josephschorr: Same deal: it executes an exclusion, it just has an empty set on the right side as a sanity check

proto/internal/dispatch/v1/dispatch.proto
josephschorr: `terminal_subject` is the subject for which the overall LR call is being made: we need it to perform the checks for intersection/exclusion, and it cannot be the current "intermediate" set of subject(s).

internal/graph/lookupresources3.go
josephschorr: They should be, but we won't know for sure without widespread testing

barakmich (dismissed):

Issue #2386, opened 2025-05-08
Enable `go vet`'s linters


tstirrat15:
The `golangci-lint` configuration doesn't currently include `go vet`'s lint rules. There's a set of problems that can be caught by static analysis that won't be caught by the compiler, such as `errors.As` invocations where `errors.As` takes an `any` as its second arg.

We ran `go vet ./...` on the repo as a test, and it seems like most of the work here will be either reordering fields or adding names to structs whose order of instantiated fields don't match their order of declared fields, mostly in tests.

rkgarcia:
Someone is working on this?

James9074:
@rkgarcia / @miparnisari I'm picking this one up today, enabling all TODO'd modules and working through them [here](https://github.com/authzed/spicedb/pull/2455). Should be finished by EOD.

Edit: https://github.com/authzed/spicedb/pull/2455 is out for review.

PR #2455, opened 2025-06-17
Enable all linters and resolve all outstanding lint issues (Resolves issue 2386)

James9074:
As per https://github.com/authzed/spicedb/issues/2386, this PR uncomments `gocritic`, `govet`, and `testifylint`, and resolves all outstanding lint findings. Note that `govet` did not actually return anything, but `gocritic` and `testifylint` did. Those were resolved, along with a number of stubborn yaml comment spacing complaints within the github workflows yamls.

I generally resolved each finding in its own commit, for clarity. Additionally, I ran the full suite of unit tests after each fix and ensured no bloat in test execution time. The test suite took ~2.5 minutes before and after all changes on my machine.

**Before:**

See [this](https://gist.github.com/James9074/351276fc413779f390b6d23fddeadbb0) for the raw output of all findings as of main/HEAD

**After:**
```bash
mage lint:all                                                                                                                                                    ✔  at 04:25:57 PM  
generating buf
generating go
Generating options for common.SchemaInformation...
Generated 1 options
Generating options for graph.ConcurrencyLimits...
Generated 1 options
Generating options for options.ExperimentalServerOptions...
Generated 1 options
Generating options for datastore.ConnPoolConfig...
Generated 1 options
Generating options for datastore.Config...
Generated 1 options
Generating options for datastore.RelIntegrityKey...
Generated 1 options
Generating options for server.CacheConfig...
Generated 1 options
Generating options for server.MiddlewareOption...
Generated 1 options
Generating options for server.Config...
Generated 1 options
Generating options for testserver.Config...
Generated 1 options
Generating options for util.GRPCServerConfig, HTTPServerConfig...
Generated 1 options
Generating options for options.QueryOptions, ReverseQueryOptions, RWTOptions...
Generated 1 options
Generating options for options.DeleteOptions...
Generated 1 options
running vulncheck
formatting go
running golangci-lint
running analyzers
0 issues.
```

### DIFF
--- a/internal/datastore/crdb/options.go
+++ b/internal/datastore/crdb/options.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/utils/ptr"
+
 	"github.com/authzed/spicedb/internal/datastore/common"
 	pgxcommon "github.com/authzed/spicedb/internal/datastore/postgres/common"
 	log "github.com/authzed/spicedb/internal/logging"
@@ -111,6 +113,16 @@ func generateConfig(options []Option) (crdbOptions, error) {
 		log.Warn().Msg("filterMaximumIDCount not set, defaulting to 100")
 	}
 
+	// Default to 30m jitter for CockroachDB database pools when not explicitly set
+	// or when explicitly set to zero (which happens when using pkg/cmd/datastore defaults)
+	if computed.readPoolOpts.ConnMaxLifetimeJitter == nil || *computed.readPoolOpts.ConnMaxLifetimeJitter == 0 {
+		computed.readPoolOpts.ConnMaxLifetimeJitter = ptr.To(30 * time.Minute)
+	}
+
+	if computed.writePoolOpts.ConnMaxLifetimeJitter == nil || *computed.writePoolOpts.ConnMaxLifetimeJitter == 0 {
+		computed.writePoolOpts.ConnMaxLifetimeJitter = ptr.To(30 * time.Minute)
+	}
+
 	return computed, nil
 }
 
@@ -183,7 +195,7 @@ func ReadConnMaxLifetime(lifetime time.Duration) Option {
 // ReadConnMaxLifetimeJitter is an interval to wait up to after the max lifetime
 // to close the connection.
 //
-// This value defaults to 20% of the max lifetime.
+// For CockroachDB, this value defaults to 30 minutes when not explicitly set.
 func ReadConnMaxLifetimeJitter(jitter time.Duration) Option {
 	return func(po *crdbOptions) { po.readPoolOpts.ConnMaxLifetimeJitter = &jitter }
 }
@@ -199,7 +211,7 @@ func WriteConnMaxLifetime(lifetime time.Duration) Option {
 // WriteConnMaxLifetimeJitter is an interval to wait up to after the max lifetime
 // to close the connection.
 //
-// This value defaults to 20% of the max lifetime.
+// For CockroachDB, this value defaults to 30 minutes when not explicitly set.
 func WriteConnMaxLifetimeJitter(jitter time.Duration) Option {
 	return func(po *crdbOptions) { po.writePoolOpts.ConnMaxLifetimeJitter = &jitter }
 }
--- a/internal/datastore/crdb/options_test.go
+++ b/internal/datastore/crdb/options_test.go
@@ -1,0 +1,67 @@
+package crdb
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfiguration(t *testing.T) {
+	tests := []struct {
+		name     string
+		options  []Option
+		validate func(t *testing.T, config crdbOptions)
+	}{
+		{
+			name:    "default jitter configuration",
+			options: []Option{},
+			validate: func(t *testing.T, config crdbOptions) {
+				require.NotNil(t, config.readPoolOpts.ConnMaxLifetimeJitter)
+				require.Equal(t, 30*time.Minute, *config.readPoolOpts.ConnMaxLifetimeJitter)
+
+				require.NotNil(t, config.writePoolOpts.ConnMaxLifetimeJitter)
+				require.Equal(t, 30*time.Minute, *config.writePoolOpts.ConnMaxLifetimeJitter)
+			},
+		},
+		{
+			name: "explicit jitter values preserved",
+			options: []Option{
+				ReadConnMaxLifetimeJitter(10 * time.Minute),
+				WriteConnMaxLifetimeJitter(15 * time.Minute),
+			},
+			validate: func(t *testing.T, config crdbOptions) {
+				// Should preserve explicitly set values
+				require.NotNil(t, config.readPoolOpts.ConnMaxLifetimeJitter)
+				require.Equal(t, 10*time.Minute, *config.readPoolOpts.ConnMaxLifetimeJitter)
+
+				require.NotNil(t, config.writePoolOpts.ConnMaxLifetimeJitter)
+				require.Equal(t, 15*time.Minute, *config.writePoolOpts.ConnMaxLifetimeJitter)
+			},
+		},
+		{
+			name: "zeros values applies defaults",
+			options: []Option{
+				// This simulates what happens when pkg/cmd/datastore passes zero values
+				// from ConnPoolConfig.MaxLifetimeJitter (which defaults to 0)
+				ReadConnMaxLifetimeJitter(time.Duration(0)),
+				WriteConnMaxLifetimeJitter(time.Duration(0)),
+			},
+			validate: func(t *testing.T, config crdbOptions) {
+				require.NotNil(t, config.readPoolOpts.ConnMaxLifetimeJitter)
+				require.Equal(t, 30*time.Minute, *config.readPoolOpts.ConnMaxLifetimeJitter)
+
+				require.NotNil(t, config.writePoolOpts.ConnMaxLifetimeJitter)
+				require.Equal(t, 30*time.Minute, *config.writePoolOpts.ConnMaxLifetimeJitter)
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			config, err := generateConfig(tt.options)
+			require.NoError(t, err)
+			tt.validate(t, config)
+		})
+	}
+}
--- a/pkg/cmd/datastore/datastore.go
+++ b/pkg/cmd/datastore/datastore.go
@@ -85,7 +85,7 @@ func RegisterConnPoolFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, defa
 	flagSet.IntVar(&opts.MaxOpenConns, flagName("max-open"), defaults.MaxOpenConns, "number of concurrent connections open in a remote datastore's connection pool")
 	flagSet.IntVar(&opts.MinOpenConns, flagName("min-open"), defaults.MinOpenConns, "number of minimum concurrent connections open in a remote datastore's connection pool")
 	flagSet.DurationVar(&opts.MaxLifetime, flagName("max-lifetime"), defaults.MaxLifetime, "maximum amount of time a connection can live in a remote datastore's connection pool")
-	flagSet.DurationVar(&opts.MaxLifetimeJitter, flagName("max-lifetime-jitter"), defaults.MaxLifetimeJitter, "waits rand(0, jitter) after a connection is open for max lifetime to actually close the connection (default: 20% of max lifetime)")
+	flagSet.DurationVar(&opts.MaxLifetimeJitter, flagName("max-lifetime-jitter"), defaults.MaxLifetimeJitter, "waits rand(0, jitter) after a connection is open for max lifetime to actually close the connection (default: 20% of max lifetime, 30m for CockroachDB)")
 	flagSet.DurationVar(&opts.MaxIdleTime, flagName("max-idletime"), defaults.MaxIdleTime, "maximum amount of time a connection can idle in a remote datastore's connection pool")
 	flagSet.DurationVar(&opts.HealthCheckInterval, flagName("healthcheck-interval"), defaults.HealthCheckInterval, "amount of time between connection health checks in a remote datastore's connection pool")
 }


James9074:

James9074:
@vroldanbet It's a wall of changes, but ready for you to look! I took great care in validating each one, ensuring it logically made sense and wouldn't change the spirit of each test or function. Feel free to suggest any changes - most are cut and dry but a few (mostly the `Requires within a goroutine` issues) could have been written a number of different ways.

internal/datastore/postgres/strictreader_test.go
```diff
@@ -73,21 +73,21 @@ func TestStrictReaderDetectsLagErrors(t *testing.T) {
 			reader.wrapped = mc
 			err := reader.ExecFunc(t.Context(), nil, "SELECT 1")
 			if expected != nil {
-				require.ErrorAs(t, err, &expected)
+				require.ErrorAs(t, err, &expected) //nolint:testifylint
```
tstirrat15: What was the issue here?

internal/datastore/common/gc_test.go
```diff
@@ -267,13 +273,17 @@ func TestGCUnlockOnTimeout(t *testing.T) {
...
 		interval := 10 * time.Millisecond
 		window := 10 * time.Second
-		timeout := 1 * time.Millisecond
+		timeout := 1 * time.Minute
 
-		require.Error(t, StartGarbageCollector(ctx, gc, interval, window, timeout))
+		errCh <- StartGarbageCollector(ctx, gc, interval, window, timeout)
 	}()
+	time.Sleep(20 * time.Millisecond) // Give goroutine time to acquire lock
+	cancel()
```
tstirrat15, line 285: This seems like a slightly different semantic - this is testing cancellation behavior rather than the timeout behavior. Was there something about the timeout behavior that wasn't amenable to refactoring into channels?

internal/datastore/postgres/xid8.go
```diff
@@ -70,7 +70,7 @@ func (Uint64Codec) PlanEncode(_ *pgtype.Map, _ uint32, format int16, value any)
 			return encodePlanUint64CodecBinaryUint64Valuer{}
 		}
 	case pgtype.TextFormatCode:
-		switch value.(type) {
+		switch value.(type) { //nolint:gocritic
```
tstirrat15: Curious here too.

internal/datastore/proxy/schemacaching/intervaltracker_test.go
```diff
@@ -179,13 +179,13 @@ func TestIntervalTrackerGC(t *testing.T) {
...
 	// GC anything older than 5s, which shouldn't change anything.
 	result := tracker.removeStaleIntervals(5 * time.Second)
-	require.False(t, false, result)
-	require.Equal(t, 2, len(tracker.sortedEntries))
+	require.False(t, result)
+	require.Len(t, tracker.sortedEntries, 2)
 
 	// GC anything older than 5ms. There should still be a single entry because it is unbounded.
 	result = tracker.removeStaleIntervals(5 * time.Millisecond)
-	require.False(t, false, result)
```
tstirrat15, line 187: Huh... was that just wrong?

internal/datastore/proxy/schemacaching/watchingcache_test.go
```diff
@@ -18,6 +19,10 @@ import (
 	"github.com/authzed/spicedb/pkg/testutil"
 )
 
+func errorAs(err error, target any) bool {
+	return errors.As(err, target)
+}
+
```
tstirrat15: Why was this necessary?

pkg/cmd/util/util.go
```diff
@@ -14,7 +14,6 @@ import (
 
 	"github.com/jzelinskie/cobrautil/v2/cobraotel"
 	"github.com/jzelinskie/stringz"
-	// Register Snappy S2 compression
 	_ "github.com/mostynb/go-grpc-compression/experimental/s2"
```
tstirrat15, line 17: Why this change? I personally like it when empty imports are annotated with why they're there.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -513,9 +513,7 @@ func translateExpressionDirect(tctx *translationContext, expressionNode *dslNode
 		if err != nil {
 			return nil, err
 		}
-		leftOps := collapseOps(leftOperation, lookup)
-		rightOps := collapseOps(rightOperation, lookup)
-		ops := append(leftOps, rightOps...)
+		ops := append(collapseOps(leftOperation, lookup), collapseOps(rightOperation, lookup)...)
```
tstirrat15, line 516: What was it complaining about here?

pkg/development/assertions.go
```diff
@@ -31,9 +31,9 @@ func RunAllAssertions(devContext *DevContext, assertions *blocks.Assertions) ([]
 		return nil, err
 	}
 
-	failures := append(trueFailures, caveatedFailures...)
-	failures = append(failures, falseFailures...)
-	return failures, nil
+	trueFailures = append(trueFailures, caveatedFailures...)
+	trueFailures = append(trueFailures, falseFailures...)
+	return trueFailures, nil
```
tstirrat15: Same here - is this about reducing allocations? If so I'd prefer that `trueFailures` is renamed to `failures` and then that change is propagated.

internal/datastore/postgres/strictreader_test.go
James9074: `error-is-as: second argument to require.ErrorAs should not be *error (testifylint)`

Shoot, this and the other lint ignore were done early on, after those I opted to not use //nolint for any of these, and forgot to go back and address these. 

I've updated this in a new commit to:

```
require.Error(t, err)
require.IsType(t, expected, err)
```

(You could also define a custom error type but that seems overkill here)

internal/datastore/postgres/xid8.go
James9074: For this one, I was preserving the switch pattern above it for readability:

```go
func (Uint64Codec) PlanEncode(_ *pgtype.Map, _ uint32, format int16, value any) pgtype.EncodePlan {
	switch format {
	case pgtype.BinaryFormatCode:
		switch value.(type) { // <---- Switch
		case uint64:
			return encodePlanUint64CodecBinaryUint64{}
		case Uint64Valuer:
			return encodePlanUint64CodecBinaryUint64Valuer{}
		}
	case pgtype.TextFormatCode:
		switch value.(type) { // <---- Switch
		case uint64:
			return encodePlanUint64CodecTextUint64{}
		}
	}

	return nil
}
```

After changing to an if statement it looked inconsistent - was going to see what y'all thought about that. For all other lint findings related to switch vs if I changed it to what the linter wanted. I've gone ahead and updated this to an if/else to remove that lintignore, if you want it reverted lmk.

internal/datastore/proxy/schemacaching/intervaltracker_test.go
James9074: Yeah I did a double take on this one too. I bet it was originally written as `require.Equal(t, false, result)` and changed it to `require.False` last minute without updating the args.

internal/datastore/proxy/schemacaching/watchingcache_test.go
James9074: It's not! I couldn't get `errors.As` to work inline in the goroutine and this was an artifact of frustrated debugging, forgot to clean it up after I realized I was calling `errors.as` instead of `errors.As` below. Fixed in new commit.

pkg/cmd/util/util.go
James9074: So this one is annoying:
`gci` and `goimports` are competing. gci wants the blank line above that comment removed, and `goimports` wants a space above it.

Running `gci write pkg/cmd/util/util.go` reorders the whole import order and _still_ produces a `goimports` error. I must have figeted with these for 20 minutes before giving in - Eventually I found this to be the only working combination that satisfied both without an ugly `//nolint`, by just removing the comments hoping anyone curious enough would look at the commit history.

Thoughts?

pkg/composableschemadsl/compiler/translator.go
James9074: `appendAssign: append result not assigned to the same slice (gocritic)` - When we do `append(leftOps, rightOps...)`, if leftOps isn't large enough, there would be a new underlying array created. Since we're assigning it back to `ops` and not `leftOps`, The original `leftOps` slice header would still point to the old memory. The linter is upset at that potential reallocation. At least I _think_ that's what's going on here

We can either assign it back to the same slice 
```
leftOps = append(leftOps, rightOps...)
ops := leftOps
```

Or do this, which is more verbose but kinda bloated imo
```
ops := make([]YourType, 0, len(leftOps)+len(rightOps))
ops = append(ops, leftOps...)
ops = append(ops, rightOps...)
```

or as I did, just consolidate it all down to one line since it's kinda obvious what we're doing anyways.

Either way, whichever you'd prefer is fine by me. I'll update it accordingly.

pkg/development/assertions.go
James9074: Yup, same deal as above. Sounds good - I changed it to `failures` throughout in a new commit.

internal/datastore/common/gc_test.go
James9074: This one is complicated - technically it was refactored into channels, but the context has to be cancelled to force the GC to return an error. I'd like to revisit this one in a new PR, but FWIW the previous state doesn't reliably test GC timeout unlocking either since that `requires` within the goroutine doesn't seem to actually fail the test like it should. The test needs to be restructured a bit, and fakegc might need to be expanded with more options. Probably outside the scope of this particular PR - but at least the current testing logic has not degraded with this change (still capturing lock/unlock)

pkg/composableschemadsl/compiler/translator.go
tstirrat15: Nah, this seems fine to me; I was just curious. Thank you for the detail!

internal/datastore/common/gc_test.go
tstirrat15: Okay, that works for me.

pkg/cmd/util/util.go
tstirrat15: Can we append the comments to the end of the line rather than having be on the previous line?

James9074: I did try that too, gci doesn't like it either. I'll keep poking at this, if I find a good solution I'll submit another PR

James9074:
@tstirrat15 Anything else I need to get this merged before it gets stale? Unable to merge it myself - not sure if @vroldanbet has to review or if he was auto-included.

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
```diff
@@ -55,11 +55,7 @@ func addTransactionMetadataTable(ctx context.Context, conn *pgx.Conn) error {
 		return err
 	}
 
-	re, err := regexp.Compile(semver.SemVerRegex)
-	if err != nil {
-		return fmt.Errorf("failed to compile regex: %w", err)
-	}
-
+	re := regexp.MustCompile(semver.SemVerRegex)
```
vroldanbet: Raising a panic does not provide a good user experience with a CLI. I'd suggest we continue to return an error.

internal/datastore/crdb/migrations/zz_migration.0009_add_expiration_support.go
```diff
@@ -47,11 +47,7 @@ func addExpirationSupport(ctx context.Context, conn *pgx.Conn) error {
 		return err
 	}
 
-	re, err := regexp.Compile(semver.SemVerRegex)
-	if err != nil {
-		return fmt.Errorf("failed to compile regex: %w", err)
-	}
-
+	re := regexp.MustCompile(semver.SemVerRegex)
```
vroldanbet: Raising a panic does not provide a good user experience with a CLI. I'd suggest we continue to return an error.

internal/datastore/postgres/xid8.go
```diff
@@ -25,7 +25,7 @@ type xid8 struct {
 	Valid  bool
 }
 
-func newXid8(u uint64) xid8 {
+func NewXid8(u uint64) xid8 {
```
vroldanbet: Why was this exported?

internal/datastore/revisions/hlcrevision_test.go
```diff
@@ -108,7 +108,7 @@ func TestInexactFloat64(t *testing.T) {
 			rev, err := HLCRevisionFromString(tc)
 			require.NoError(t, err)
 
-			require.Equal(t, floatValue, rev.InexactFloat64())
+			require.InEpsilon(t, floatValue, rev.InexactFloat64(), 1e-9)
```
vroldanbet: We don't want to give a % of acceptable deviation here. This is parsing a hybrid logical clock from a string, and we expect the values to be precisely what is provided in the table tests.

internal/caveats/builder_test.go
```diff
@@ -228,7 +228,7 @@ func TestMustCaveatExprForTestingWithContext(t *testing.T) {
 
 	contextMap := result.GetCaveat().Context.AsMap()
 	require.Equal(t, "value1", contextMap["key1"])
-	require.Equal(t, float64(42), contextMap["key2"])
+	require.InEpsilon(t, 42, contextMap["key2"], 0.00001)
```
vroldanbet: I don't understand why using a margin of error here, please revert

internal/dispatch/singleflight/singleflight_test.go
```diff
@@ -187,29 +187,35 @@ func TestSingleFlightDispatcherCancelation(t *testing.T) {
 	disp := New(mockDispatcher{f: f}, &keys.DirectKeyHandler{})
 	wg := sync.WaitGroup{}
 	wg.Add(3)
+	errs := make(chan error, 3)
```
vroldanbet, line 190: Don't understand the reason for the change on this test, but it should be equivalent.

internal/middleware/perfinsights/perfinsights_test.go
```diff
@@ -96,7 +96,7 @@ func TestObserveShapeLatency(t *testing.T) {
 		require.Equal(t, io_prometheus_client.MetricType_HISTOGRAM, metric.GetType())
 
 		require.Equal(t, uint64(1), metric.GetMetric()[0].Histogram.GetSampleCount())
-		require.Equal(t, float64(0.1), metric.GetMetric()[0].Histogram.GetSampleSum())
+		require.InEpsilon(t, 0.1, metric.GetMetric()[0].Histogram.GetSampleSum(), 0.01)
```
vroldanbet: please reframe from using InEpsilon here

pkg/schemadsl/compiler/translator.go
```diff
@@ -453,9 +453,7 @@ func translateExpressionDirect(tctx *translationContext, expressionNode *dslNode
 		if err != nil {
 			return nil, err
 		}
-		leftOps := collapseOps(leftOperation, lookup)
-		rightOps := collapseOps(rightOperation, lookup)
-		ops := append(leftOps, rightOps...)
+		ops := append(collapseOps(leftOperation, lookup), collapseOps(rightOperation, lookup)...)
```
vroldanbet: Please leave it as it was, this is a cosmetic change and it seemed easier to parse visually before

vroldanbet (changes requested):
Thanks for your contribution! ✨ I understand it can get stale easily, but please be patient while we review this low-priority PR; it's a bit lengthy.

internal/datastore/revisions/hlcrevision_test.go
James9074: Noted and reverted.

internal/caveats/builder_test.go
James9074: The changes to InEpsilon came from `float-compare: use require.InEpsilon (or InDelta) (testifylint)`

The implications of this can be [seen the bottom of this post](https://medium.com/pragmatic-programmers/testing-floating-point-numbers-in-go-9872fe6de17f), but in this context, it probably doesn't matter. Apologies, I didn't trace inputs for this test all the way up determine how likely this approximation might actually cause this test to fail from time to time.

I've reverted all InEpsilon changes and marked them as `//nolint:testifylint`

internal/middleware/perfinsights/perfinsights_test.go
James9074: Noted and reverted.

pkg/schemadsl/compiler/translator.go
James9074: I agree that it was easier to parse visually - technically this prevents a [possible memory issue](https://stackoverflow.com/questions/78685528/what-is-the-idiomatic-method-and-style-of-combining-two-slices-in-go) stemming from this lint finding: `appendAssign: append result not assigned to the same slice (gocritic)`

When we do append(leftOps, rightOps...), if leftOps isn't large enough, there would be a new underlying array created. Since we're assigning it back to ops and not leftOps, The original leftOps slice header would still point to the old memory. The linter is upset at that potential reallocation. At least I think that's what's going on here, I haven't tested / profiled this scenario yet.

I don't see this being a huge issue considering the small memory this is using anyways. I'll go ahead and revert it.

internal/dispatch/singleflight/singleflight_test.go
James9074: This stems from `go-require: require must only be used in the goroutine running the test function (testifylint)`. Most of the lint issues in this PR are a result of using require within goroutines.

Some context on why: `vet` warns on using require within goroutines (`require` is the same as `t.FailNow()` and [this](https://go.dev/doc/go1.16#vet-testing-T) is explicitly [discouraged](https://github.com/stretchr/testify/issues/772#issuecomment-945166599) within a goroutine)

To address this, I used the a channel to collect them and then check after the goroutines have completed. 

What do you mean by equivalent here?

internal/datastore/crdb/migrations/zz_migration.0008_add_transaction_metadata_table.go
James9074: Agreed, reverted.

internal/datastore/postgres/xid8.go
James9074: I remember it solving another lint finding, but reverting it back doesn't produce any additional findings, so I must have been mistaken there. My bad.

Reverted!

.golangci.yaml
Meyazhagan, line 1: Hi @James9074 under `linters -> settings` add this, 

```yml
linters:
  settings:
    govet:
      enable-all: true
```

this will enable all checks, and you will see some `govet` linting

James9074: Thanks - auto lint from this this introduced a _large_ number of small cosmetic changes, I'm going to keep it disabled for this PR, let the existing linter resolutions get through, and then address govet in another PR.

internal/datastore/crdb/migrations/zz_migration.0009_add_expiration_support.go
James9074: Reverted

tstirrat15:
@James9074 just wanted to update - this is still something we're interested in and we haven't had the bandwidth to look at it yet. Apologies for the delay.

tstirrat15:
@James9074 we're finally done with the push and should have some more bandwidth now. do you have the time and energy to get this rebased and ready? If not, we can take over the PR and make sure that you're attributed.

James9074:
> @James9074 we're finally done with the push and should have some more bandwidth now. do you have the time and energy to get this rebased and ready? If not, we can take over the PR and make sure that you're attributed.

Awesome! No worries, I'll rebase this week and resubmit for review.

internal/datastore/common/gc_test.go
vroldanbet: While the proposed code simulates unlock on timeout by triggering context cancellation and assumes it is equivalent, the context cancellation is not propagated to the `RunGarbageCollection` function, which uses `context.Background` with a timeout. Thus, we are not testing for a timeout because it is not triggering, and cancellation is handled in a different part of the code.

The test ultimately passes due to chance (the `time.Sleep` allowing enough time for `RunGarbageCollection` to lock and unlock) and the cancellation causing the next retry of `RunGarbageCollection` to be aborted because the context was cancelled on `StartGarbageCollector`.

Therefore:
- The test relies on timing (thus can be flaky)
- It does not give us the guarantee that the unlock happens after a timeout (because it is never given the chance to time out, and cancellation is not propagated).

I'm not sure why we ended up with `RunGarbageCollection` taking a timeout, but I think it should instead take the context, and `StartGarbageCollector` should set the timeout. Then, to definitely assert that unlock happens after cancellation:
- We'd have to synchronize in the `Lock` method, creating a custom implementation of the fake GC that uses a channel to synchronize with the test and block until `Lock` is called
- Once `Lock` is called, we can call `cancel()`, which would propagate all the way up to `RunGarbageCollection`, and give us the guarantee that `UnlockAfterGCRun` is called. Note that the method does not take the context, which is deliberate; however, it should likely have at least a timeout to prevent blocking indefinitely.

One could argue that the current test is equally incorrect as the former test. If you can timebox the proper fix, that would be great; if not, I suggest following up or opening an issue.

pkg/schema/typesystem_test.go
```diff
@@ -36,17 +36,26 @@ func TestTypeSystemConcurrency(t *testing.T) {
...
+						mu.Unlock()
+					}
 				}
 			}
 			wg.Done()
 		}()
 	}
 	wg.Wait()
+
+	require.Empty(errs, "expected no errors in concurrent GetValidatedDefinition calls")
```
vroldanbet, line 60: Why was this necessary? Wouldn't the wait group guarantee the errors are checked with `require.NoError`?

I wrote a quick test, and failures are bubbled up correctly:

```go
func TestName(t *testing.T) {
	wg := sync.WaitGroup{}

	wg.Add(1)
	go func() {
		defer wg.Done()
		require.Fail(t, "should have panicked")
	}()

	wg.Wait()
}
```

pkg/releases/releases_test.go
```diff
@@ -18,8 +18,9 @@ func TestGetSourceRepository(t *testing.T) {
 
 func TestGetLatestRelease(t *testing.T) {
 	t.Run("successful release fetch", func(t *testing.T) {
+		var receivedPath string
 		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			require.Equal(t, "/repos/authzed/spicedb/releases/latest", r.URL.Path)
```
vroldanbet, line 22: Is this being enforced by some new linter? It's fine, but the previous and current code should be equivalent. I'm curious about the rationale. It could be easier to spot the expectations around this test if they are placed in the same spot.

pkg/development/assertions.go
```diff
@@ -16,7 +16,7 @@ const maxDispatchDepth = 25
 // RunAllAssertions runs all assertions found in the given assertions block against the
 // developer context, returning whether any errors occurred.
 func RunAllAssertions(devContext *DevContext, assertions *blocks.Assertions) ([]*devinterface.DeveloperError, error) {
-	trueFailures, err := runAssertions(devContext, assertions.AssertTrue, v1.ResourceCheckResult_MEMBER, "Expected relation or permission %s to exist")
+	failures, err := runAssertions(devContext, assertions.AssertTrue, v1.ResourceCheckResult_MEMBER, "Expected relation or permission %s to exist")
```
vroldanbet, line 19: This leads to an asymmetry with `falseFailures`, resulting in worse readability than before. What's the motivation for the change? A linter?

If using `true` is tripping any linter here, I'd suggest renaming `trueFailures` to `memberFailures` and `falseFailures` to `notMemberFailures`

pkg/datastore/test/revisions.go
```diff
@@ -300,18 +300,33 @@ func ConcurrentRevisionsTest(t *testing.T, tester DatastoreTester) {
...
+
 	for i := 0; i < 10; i++ {
 		go func() {
 			defer wg.Done()
 
 			for i := 0; i < 5; i++ {
 				head, err := ds.HeadRevision(ctx)
-				require.NoError(err)
-				require.NoError(ds.CheckRevision(ctx, head), "expected head revision to be valid in GC Window")
-				require.True(head.GreaterThan(startingRev) || head.Equal(startingRev))
```
vroldanbet, line 311: I would assume this is unnecessary because a wait group is already in place, but perhaps I'm missing something. Do those `require` failures not bubble up when inside of a goroutine and thus would not cause the test to fail?

I don't feel like the `continue` are needed, are they? It could fail fast with a `return`

internal/datastore/proxy/schemacaching/watchingcache_test.go
```diff
@@ -142,30 +143,47 @@ func TestWatchingCacheParallelOperations(t *testing.T) {
...
 		_, _, err := wcache.SnapshotReader(rev("1")).ReadNamespaceByName(t.Context(), "somenamespace")
-		require.ErrorAs(t, err, &datastore.NamespaceNotFoundError{})
-		require.False(t, wcache.namespaceCache.inFallbackMode)
+		var nsNotFoundErr datastore.NamespaceNotFoundError
+		if err == nil || !errors.As(err, &nsNotFoundErr) {
+			t.Errorf("expected NamespaceNotFoundError, got: %v", err)
+		}
+		if wcache.namespaceCache.inFallbackMode {
+			t.Errorf("expected inFallbackMode to be false")
+		}
```
vroldanbet, line 152: What happened here? `require` ultimately causes `t` to fail. And why does not abide by the same recommendations described [here](https://github.com/authzed/spicedb/pull/2455#discussion_r2303688151)?

same for all other instances in this file

internal/testfixtures/datastore.go
```diff
@@ -342,7 +342,7 @@ func (tc RelationshipChecker) VerifyOrderedIteratorResults(iter datastore.Relati
 		expectedStr := tuple.MustString(rel)
 
 		if index > len(expected)-1 {
-			tc.Require.Fail("expected %s, but found no additional results", expectedStr)
+			tc.Require.Fail("expected more results, but found no additional results", expectedStr)
```
vroldanbet, line 345: Could you tell me what you're trying to fix here? How is `expected more results, but found no additional results` a better error message?

pkg/datastore/test/counters.go
```diff
@@ -79,28 +79,34 @@ func RegisterRelationshipCountersInParallelTest(t *testing.T, tester DatastoreTe
 	var numSucceeded atomic.Int32
 	var numFailed atomic.Int32
 	var wg sync.WaitGroup
+	errs := make([]error, 10)
```
vroldanbet, line 82: This looks semantically equivalent to me, but I'd be curious to understand again the rationale behind these type of changes.

pkg/composableschemadsl/compiler/translator.go
vroldanbet, line 516: Could you tell me the reason for this change?

pkg/cmd/util/util.go
vroldanbet: I appended the command at the end, ran locally with `golangci-lint` 2.4.0, and it seemed to work.

internal/dispatch/singleflight/singleflight_test.go
vroldanbet: Sorry, I missed this comment, which explains a bunch of my feedback on my last review.

With equivalent I mean _semantically equivalent_. For a reviewer, the intent behind the change is not apparent because it's not clear what it's attempting to fix.

I wasn't aware of those recommendations, so this makes sense to me.

pkg/schema/typesystem_test.go
vroldanbet: I missed your earlier comment [here](https://github.com/authzed/spicedb/pull/2455#discussion_r2303688151), so I'll resolve this one.

pkg/releases/releases_test.go
vroldanbet: I missed your earlier comment [here](https://github.com/authzed/spicedb/pull/2455#discussion_r2303688151), so I'll resolve this one.

pkg/datastore/test/revisions.go
vroldanbet: I missed your earlier comment [here](https://github.com/authzed/spicedb/pull/2455#discussion_r2303688151), so I'll resolve this one.

pkg/datastore/test/counters.go
vroldanbet: I missed your earlier comment [here](https://github.com/authzed/spicedb/pull/2455#discussion_r2303688151), so I'll resolve this one.

James9074:
@tstirrat15 Feel free to take this one over or close it out, unfortunately I'm short on time this quarter and don't think I'll be able to wrap this up before it's too stale.

tstirrat15:
Sounds good. Thank you so much for the work you put into it!

miparnisari:
Thank you @James9074! Your hard work doesn't go unnoticed. We'll take this over the finish line :)

tstirrat15:
All of the individual pieces of this have been brought in. Thank you for putting this up!

PR #2457, opened 2025-06-17
Implement the "use typechecking" directive, annotating the source, and checking against the schema if the types are present

barakmich:


pkg/schemadsl/parser/tests/permission_type_annotation.zed
```diff
@@ -0,0 +1,7 @@
+definition mydefinition {
+    relation viewer: user
+    
+    permission view: user = viewer
+    permission edit: user = viewer
+    permission admin: group = viewer
```
josephschorr, line 6: Add additional tests for error cases:

- `:` but no type before the `=`
- `: something | `
- `: |`

pkg/schemadsl/compiler/translator.go
```diff
@@ -434,6 +437,70 @@ func extractTypeAnnotations(typeAnnotationNode *dslNode) ([]string, error) {
 	return annotations, nil
 }
 
+func addTypeAnnotationsToMetadata(relation *core.Relation, typeAnnotations []string) error {
+	if relation.Metadata == nil {
+		relation.Metadata = &core.Metadata{}
+	}
+
+	// Find existing PERMISSION RelationMetadata and update it, or create new one
+	for i, metadataAny := range relation.Metadata.MetadataMessage {
```
josephschorr: Move this logic into https://github.com/authzed/spicedb/blob/main/pkg/namespace/metadata.go and follow a similar style to there

proto/internal/impl/v1/impl.proto
```diff
@@ -77,6 +77,7 @@ message RelationMetadata {
   }
 
   RelationKind kind = 1;
+  repeated string type_annotations = 2;
```
josephschorr: Put this under a TypeAnnotations message so we can change it in the future if we want

pkg/schemadsl/compiler/translator.go
```diff
@@ -103,6 +106,24 @@ func translate(tctx *translationContext, root *dslNode) (*CompiledSchema, error)
 
 	// Check for typechecking after all definitions are processed.
 	if slices.Contains(tctx.enabledFlags, "typechecking") {
+		resolver := ResolverForCompiledSchema(compiledSchema)
+		ts := schema.NewTypeSystem(resolver)
```
josephschorr: This should be done as part of the `.Validate` call on the type system, rather than the schema translator

barakmich: one slight caveat there -- we lose the `tctx`. Now, an easy way to fix that is to strip the metadata if it's not enabled, and then check it in Validate, which I'll add as a commit

pkg/schemadsl/lexer/flags.go
```diff
@@ -37,3 +37,5 @@ var Flags = map[string]transformer{
 		return lexeme, false
 	},
 }
+
+var AllUseFlags = []string{FlagExpiration, FlagTypeChecking}
```
josephschorr: have this read the keys from the map via `slices.Collect(maps.Keys(Flags))`

pkg/namespace/metadata.go
```diff
@@ -107,3 +107,89 @@ func SetRelationKind(relation *core.Relation, kind iv1.RelationMetadata_Relation
 	metadata.MetadataMessage = append(metadata.MetadataMessage, encoded)
 	return nil
 }
+
+// GetTypeAnnotations returns the type annotations for a permission relation.
+func GetTypeAnnotations(relation *core.Relation) []string {
+	if relation.Metadata == nil || len(relation.Metadata.MetadataMessage) == 0 {
```
josephschorr: don't need the `len` check here, since below its just a for loop

pkg/schema/type_check_test.go
```diff
@@ -417,6 +417,123 @@ func TestTypecheckingWithSubrelations(t *testing.T) {
...
+			name: "incomplete type annotation",
+			schemaText: `use typechecking
+			definition user {}
+			definition team {}
+
+			definition document {
+				relation viewer: user | team
+				permission view: user = viewer
+			}`,
+			expectedError: "incomplete type annotation on relation view in definition document: team found as reachable type, but not contained in provided set [user]",
```
josephschorr: in the error messages, wrap the names in back ticks, like: "`team` found as reachable..." and "in the provided set: [`user`]"

pkg/schemadsl/compiler/translator.go
```diff
@@ -390,6 +422,22 @@ func translatePermission(tctx *translationContext, permissionNode *dslNode) (*co
 	return permission, nil
 }
 
+func extractTypeAnnotations(typeAnnotationNode *dslNode) ([]string, error) {
```
josephschorr, line 426: doc comment

pkg/schemadsl/compiler/type_annotations_invalid_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/pkg/schemadsl/input"
+)
+
+func TestPermissionTypeAnnotationInvalidSyntax(t *testing.T) {
```
josephschorr: no need for this test suite; instead, add `.zed` files with these parsing errors as test cases

pkg/schemadsl/lexer/flags.go
barakmich: `maps.Keys` [cannot guarantee an order](https://pkg.go.dev/maps#Keys), which can flake the tests. I could do this with slices.Sort() in init() and will do it that way, but doing it as you suggest was why I put this here.

josephschorr: Why should the order of this slice matter?

barakmich: `pkg/schemadsl/parser/tests/invaliduse.zed.expected` -- the error message

josephschorr (approved):
LGTM

PR #2458, opened 2025-06-18
Improve how we select index forcing for CRDB

josephschorr:
This uses a less hand-written method and more structural and well tested method to select which index to force, if any


internal/datastore/crdb/schema/indexutil.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return columns, nil
+}
+
+func forcedIndexForFilter(filter datastore.RelationshipsFilter, indexes []common.IndexDefinition) (*common.IndexDefinition, error) {
+	// Algorithm: Find the index that has the most leading columns matching the filter.
+	// If at any point a column within the index is not in the filter, we stop checking for
+	// that index. We return the index with the most leading columns matched. Resource IDs
+	// are treated as an *immediate* stop, as prefix scanning does not work well with
+	// a following field in the index.
```
tstirrat15, line 65: Is this because we don't know whether a resource ID is prefixed or not at this point in the code?

internal/datastore/crdb/schema/indexes_test.go
```diff
@@ -26,16 +26,16 @@ func TestIndexForFilter(t *testing.T) {
...
+			expectedWithIntegrity:    "ix_relation_tuple_with_integrity",
 		},
 		{
 			name: "filter by resource type and relation",
 			filter: datastore.RelationshipsFilter{
 				OptionalResourceType:     "foo",
 				OptionalResourceRelation: "bar",
 			},
-			expectedWithoutIntegrity: "pk_relation_tuple",
-			expectedWithIntegrity:    "pk_relation_tuple",
```
tstirrat15, line 38: Are these changes reflecting a change in behavior of index selection because the structure of the query doesn't actually match the structure of the index under this new work?

internal/datastore/crdb/schema/indexutil.go
josephschorr: No; its because once we've seen a prefix, the rest of the index cannot be efficiently used

internal/datastore/crdb/schema/indexes_test.go
josephschorr: The changes reflect us picking a better forced index, yes

internal/datastore/crdb/schema/indexutil.go
```diff
@@ -0,0 +1,222 @@
...
+	mustInit()
+}
+
+func mustInit() {
+	for _, idx := range crdbAllIndexes {
+		parsed, err := parseIndexColumns(idx.ColumnsSQL)
+		if err != nil {
+			panic(err)
+		}
+		parsedColumnsPerIndex[idx.Name] = parsed
```
vroldanbet, line 26: please add a check to make sure the same name is not registered twice

internal/datastore/crdb/schema/indexutil.go
```diff
@@ -0,0 +1,222 @@
...
+		return nil, fmt.Errorf("no columns found in parentheses in SQL: %s", columnsSQL)
+	}
+
+	// Split by comma and trim whitespace
+	columnsStr := matches[1]
+	columns := regexp.MustCompile(`\s*,\s*`).Split(columnsStr, -1)
+
+	// Trim any remaining whitespace
+	for i, col := range columns {
+		columns[i] = regexp.MustCompile(`^\s+|\s+$`).ReplaceAllString(col, "")
```
vroldanbet:
```suggestion
		columns[i] =  strings.TrimSpace(col)
```

vroldanbet: Wanted to add a fuzzing test for more peace of mind. Found only 1 issue: index with an empty column name. Easy fix:

```go
	for i, col := range columns {
		trimmed := strings.TrimSpace(col)
		if trimmed == "" {
			return nil, fmt.Errorf("empty column name found in SQL: %s", columnsSQL)
		}

		columns[i] = trimmed
	}
```



Here you have the test:
```go

func FuzzParseIndexColumns(f *testing.F) {
	// Add seed inputs that cover different valid and invalid cases
	f.Add("PRIMARY KEY (namespace, object_id, relation)")
	f.Add("relation_tuple (userset_object_id, userset_namespace)")
	f.Add("PRIMARY KEY ( namespace , object_id )")
	f.Add("PRIMARY KEY ()")
	f.Add("PRIMARY KEY namespace, object_id")
	f.Add("(namespace)")
	f.Add("")
	f.Add(",")
	f.Add(",columnName")
	f.Add("PRIMARY KEY (namespace, object_id, relation")
	f.Add("PRIMARY KEY namespace, object_id, relation)")
	f.Add("PRIMARY KEY (  namespace, object_id,  ")
	f.Add("relation_tuple_with_integrity (namespace, object_id) STORING (integrity_key_id)")

	f.Fuzz(func(t *testing.T, input string) {
		columns, err := parseIndexColumns(input)

		if err == nil {
			// If no error, we should have at least one column
			require.NotEmpty(t, columns, "parseIndexColumns returned no error but empty columns for input: %q", input)

			// All returned columns should be non-empty strings
			for i, col := range columns {
				require.NotEmpty(t, col, "parseIndexColumns returned empty column at index %d for input: %q", i, input)
			}
		} else {
			require.Empty(t, columns, "parseIndexColumns returned no error but empty columns for input: %q", input)
		}
	})
}
```

internal/datastore/crdb/reader.go
```diff
@@ -101,7 +101,11 @@ func (cr *crdbReader) CountRelationships(ctx context.Context, name string) (int,
 		return 0, err
 	}
 
-	index := schema.IndexForFilter(cr.schema, relFilter)
+	index, err := schema.IndexForFilter(cr.schema, relFilter)
```
vroldanbet, line 104: I don't understand how this does not panic. In `internal/datastore/crdb/schema/indexes_test.go` you describe many scenarios that lead to a `nil` index pointer and nil `error` (the change in this PR make that more likely now), but right after this we are calling `index.Name`.

internal/datastore/crdb/schema/indexes_test.go
```diff
@@ -26,16 +26,16 @@ func TestIndexForFilter(t *testing.T) {
...
+			expectedWithoutIntegrity: "",
```
vroldanbet: Non-blocking suggestion: It's not immediately obvious what the intent is - I believe we want to let cockroachDB decide. I think it would help if you created a constant like `const letCockroachDBDecide = ""` would make it easier to follow

internal/datastore/crdb/schema/indexutil_test.go
```diff
@@ -0,0 +1,780 @@
...
+		{
+			name:        "empty parentheses",
+			columnsSQL:  "PRIMARY KEY ()",
+			expectError: true,
+		},
+		{
+			name:         "single column",
+			columnsSQL:   "relation_tuple (namespace)",
+			expectedCols: []string{"namespace"},
+		},
```
vroldanbet, line 63: please add these two extra tests
```suggestion
		},
		{
			name:         "single column with spaces",
			columnsSQL:   "relation_tuple (    namespace     )",
			expectedCols: []string{"namespace"},
		},
		{
			name:         "malformed single column with trailing comma",
			columnsSQL:   "relation_tuple ( ,namespace )",
			expectedCols: []string{"namespace"},
			expectError:  true,
		},
```

internal/datastore/crdb/schema/indexes.go
```diff
@@ -84,100 +92,11 @@ func IndexingHintForQueryShape(schema common.SchemaInformation, qs queryshape.Sh
...
-func (f forcedIndex) FromTable(existingTableName string) (string, error) {
-	// Indexes are forced in CRDB by appending the index name after an @ sign after the table
-	// name in the FROM clause.
-	// Example: FROM relation_tuple@ix_relation_tuple_by_subject
-	if existingTableName == "" {
-		return "", spiceerrors.MustBugf("existing table name is empty")
+func IndexForFilter(schema common.SchemaInformation, filter datastore.RelationshipsFilter) (*common.IndexDefinition, error) {
+	indexesToCheck := crdbWithoutIntegrityIndexes
+	if schema.IntegrityEnabled {
+		indexesToCheck = crdbWithIntegrityIndexes
```
vroldanbet, line 99: Can you elaborate why we are considering only 1 index when integrity is enabled?

internal/datastore/crdb/schema/indexes_test.go
vroldanbet: Validating my understanding: these force no index because we are missing one with `resource_type, resource_relation` that we will eventually add.

internal/datastore/crdb/schema/indexes_test.go
```diff
@@ -217,7 +217,7 @@ func TestIndexForFilter(t *testing.T) {
 				OptionalResourceIDPrefix: "prefix",
 			},
 			expectedWithoutIntegrity: "pk_relation_tuple",
-			expectedWithIntegrity:    "pk_relation_tuple",
+			expectedWithIntegrity:    "ix_relation_tuple_with_integrity",
 		},
 		{
 			name: "filter by resource type, ID prefix and relation",
```
vroldanbet, line 225: Why isn't `"multiple subject selectors with different depths"` selecting the subject's index? Unless I'm missing something, if feels like it should

internal/datastore/crdb/schema/indexes_test.go
vroldanbet, line 1: Not a blocker: Maybe add a fuzzing test to `IndexForfilter`?

internal/datastore/crdb/schema/indexes_test.go
```diff
@@ -337,7 +395,9 @@ func TestIndexForFilter(t *testing.T) {
 		schema := Schema(common.ColumnOptimizationOptionNone, withIntegrity, false)
```
vroldanbet, line 397: Not a blocker: Would it make sense to add more variants to this test that include column optimization turned on? What about `expirationDisabled`?

internal/datastore/crdb/schema/indexutil.go
```diff
@@ -0,0 +1,222 @@
...
+				return nil, nil
+			}
+		}
+	}
+
+	return bestIndex, nil
+}
+
+func checkIfMatchingIndex(filter datastore.RelationshipsFilter, idx common.IndexDefinition) (int, error) {
+	columnNames := parsedColumnsPerIndex[idx.Name]
```
vroldanbet: return `spiceerrors.MustBugf` if no index matches

internal/datastore/crdb/schema/indexutil.go
```diff
@@ -0,0 +1,222 @@
...
+		filterStatus, err := checkFilterColumnMatchesFilter(colName, filter)
+		if err != nil {
+			return -1, err
+		}
+
+		switch filterStatus {
+		case columnFilterNoMatch:
+			if columnIndex == 0 {
+				// If the first column doesn't match, this index is not a match.
+				return -1, nil
```
vroldanbet: nit: magic number - add constant `doesNotMatch` or something along the lines

internal/datastore/crdb/schema/indexutil.go
```diff
@@ -0,0 +1,222 @@
...
+		foundCount := 0
+		for _, sel := range filter.OptionalSubjectsSelectors {
+			if sel.OptionalSubjectType != "" {
+				foundCount++
+			}
+		}
+		if foundCount == 0 {
+			return columnFilterNoMatch, nil
+		} else if foundCount < len(filter.OptionalSubjectsSelectors) {
+			return columnFilterForceNoMatch, nil
```
vroldanbet, line 194: Could you perhaps document this scenario? Imagine a scenario where all resource columns matched. Why wouldn't it be a good idea to match the PK index?

internal/datastore/crdb/reader.go
josephschorr: Weird... I'll try to add a test to reproduce

internal/datastore/crdb/schema/indexes.go
josephschorr: Because it only has a single index defined; we never added the other indexes in the integrity table, because we intended to go back and optimize it after the fact, once we fully migrated the existing indexes for existing non-integrity table. However, since we haven't done that step, we never did the next one. I've added a TODO

internal/datastore/crdb/schema/indexes_test.go
josephschorr: The different depths results in us not being able to use the index in portions of the cases

internal/datastore/crdb/schema/indexes_test.go
josephschorr: They don't impact index selection

internal/datastore/crdb/schema/indexutil.go
josephschorr: Because you're skipping columns and that results in an index mismatch

josephschorr:
Updated

internal/datastore/crdb/schema/indexes_test.go
vroldanbet: ah, you are right

internal/datastore/crdb/schema/indexutil.go
vroldanbet: I was under the impression it's not skipping columns

PR #2460, opened 2025-06-18
test fips

miparnisari:


miparnisari:
opened https://github.com/magefile/mage/pull/526

Issue #2396, opened 2025-05-20
Support relation deprecation in schema


ecordell:
### Problem Statement

Downstream SpiceDB consumers may not be the same groups of people maintaining the schema. It would be useful to have some tooling to help folks roll out phased migrations of the schema itself.

### Solution Brainstorm

This is pure strawman, just getting something on paper:

```
definition document {
   @deprecated(warn)
   relation group: group

   @deprecated(error)
   relation folder: folder
}
```

In this example, writing `document:foo#group@group:bar` would result in a warning on write (i.e. `document#group has been marked deprecated, please migrate away`) but would still write the relationship to the datastore.  Writing to `document:foo#folder@folder:bar` would result in a write error (`document#foo has been disabled`) and would not write to the datastore.

A relation can be phased out by:

- Marking it as `deprecated(warn)`
- Waiting for users to notice / migrate away
- Marking it as `deprecated(error)` so that no new data can be written
- Deleting all existing relationships for that relation
- Removing the relation from the schema

Thoughts?

kartikaysaxena:
I've proposed an approach in #2465. Could you please take a look? It's now open for review.

vroldanbet:
related https://github.com/authzed/spicedb/issues/1535

PR #2465, opened 2025-06-21
feat: support relation and object deprecation

kartikaysaxena:
Related: #2396

This PR adds support for relation-level deprecation in schemas.

- Relations can now be annotated with a `@deprecated(...)` directive to indicate they are deprecated.
- When a relationship is written to a deprecated relation:
  - A **warning** or **error** is generated, depending on the specified deprecation level.
  - In the current implementation, writes fail when an error deprecation is configured.
- Deprecation is associated with the **immediately following relation** in the definition.
- A flag is to be passed to enable this feature, namely `--enable-experimental-relationship-deprecation` along with use directive.

**Example:**
```dsl
use deprecation

definition user {}

definition resource {
  @deprecated(warn)
  relation viewer: user

  @deprecated(error)
  relation admin: user
}
```
> [!NOTE]
>Each relation has an enum `DeprecationType` assigned to them, if unspecified it defaults to `DEPRECATED_TYPE_UNSPECIFIED`


pkg/schemadsl/lexer/lex_def.go
tstirrat15, line 1: A general note on the approach: to the degree that we can, we want to avoid breaking folks' existing schemas. See what we did with `expiration` for an example of how we've worked around it with `use` directives - if we add this feature, it'll be with one of those `use` directives.

tstirrat15: We also need the `composableschemadsl` package to be updated accordingly.

pkg/schemadsl/lexer/lex_def.go
```diff
@@ -153,6 +155,12 @@ Loop:
 
 		case r == '%':
 			l.emit(TokenTypePercent)
+		case r == '@':
+			if l.acceptString("deprecated") {
+				l.emit(TokenTypeKeyword)
```
tstirrat15: This isn't something that should live in the lexer - it should live in the parser.

Or at least my sense is that an `@` should be a marker, and then the token after it should be a keyword, rather than the entire thing being marked as a keyword.

It's also going to relate to what I said above about the `use` directive.

ecordell: it might be worth looking at https://github.com/authzed/spicedb/pull/202, which implemented generic decorators (but we didn't have a use case so never merged it)

kartikaysaxena: reworked the approach a little bit including the use directive just like how it was done for expiration feature, adding more comments for better review.

josephschorr:
I wonder if we want to allow the deprecated annotation on each subject type vs the relation as a whole; I could see benefits to either. @ecordell thoughts?

kartikaysaxena:
we may also support something like
```
use deprecation

definition user {}

definition resource {
  relation viewer: user
  relation admin: user
}

@deprecated(warn, user, "<comments about deprecation>")
@deprecated(error, admin, "<comments about deprecation>")
```
which can include both subjects and relations, this way deprecation can be more flexible and granular also including the comments about the deprecation.

Any thoughts @josephschorr @ecordell @tstirrat15 ?

josephschorr:
This is also missing changes to the schema generator package and associated tests

kartikaysaxena:
> This is also missing changes to the schema generator package and associated tests

Updated the generator to accommodate deprecation and added some tests. I know you all are a bit busy, so feel free to take a look whenever you get the chance, thanks!

kartikaysaxena:
Rebased

josephschorr:
Please rebase again and will continue review from there

josephschorr:
> we may also support something like
> 
> ```
> use deprecation
> 
> definition user {}
> 
> definition resource {
>   relation viewer: user
>   relation admin: user
> }
> 
> @deprecated(warn, user, "<comments about deprecation>")
> @deprecated(error, admin, "<comments about deprecation>")
> ```
> 
> which can include both subjects and relations, this way deprecation can be more flexible and granular also including the comments about the deprecation.
> 
> Any thoughts @josephschorr @ecordell @tstirrat15 ?

I like it; let's add it and have it returned

vroldanbet:
>I like it; let's add it and have it returned

Is the `@` necessary at all?

kartikaysaxena:
requesting for initial review, the deprecation support now looks like
```
use deprecation

definition user {}
definition superuser {}
definition anotheruser {}

definition platform {
	relation admin: user
	relation editor: superuser

	@deprecated(warn, “rel viewer is deprecated”)
	relation viewer: anotheruser
}

@deprecated(error, user, “user object is deprecated”)
@deprecated(warn, platform#editor, “this to deprecate a relation under a namespace”)
```
This adds support for deprecating relation inside a definition namespace for marking the immediate next relation as deprecated, for outside a namespace a relation can be marked deprecated as `@deprecated(warn, platform#editor, “this to deprecate a relation under a namespace”)` and an object by simply changing the relation with the object, example platform#editor -> user will mark user as deprecated, it _works_ if the object is either defined as a subject or a resource in a `WriteRelationship` call.

(also would migrate this pr to a new branch if this branch looks messy after too many rebases)

josephschorr:
> requesting for initial review, the deprecation support now looks like
> 
> ```
> use deprecation
> 
> definition user {}
> definition superuser {}
> definition anotheruser {}
> 
> definition platform {
> 	relation admin: user
> 	relation editor: superuser
> 
> 	@deprecated(warn, “rel viewer is deprecated”)
> 	relation viewer: anotheruser
> }
> 
> @deprecated(error, user, “user object is deprecated”)
> @deprecated(warn, platform#editor, “this to deprecate a relation under a namespace”)
> ```
> 
> This adds support for deprecating relation inside a definition namespace for marking the immediate next relation as deprecated, for outside a namespace a relation can be marked deprecated as `@deprecated(warn, platform#editor, “this to deprecate a relation under a namespace”)` and an object by simply changing the relation with the object, example platform#editor -> user will mark user as deprecated, it _works_ if the object is either defined as a subject or a resource in a `WriteRelationship` call.
> 
> (also would migrate this pr to a new branch if this branch looks messy after too many rebases)

Sorry @kartikaysaxena, this is my bad at not being explicit: I liked the idea of being able to mark definitions as deprecated, but they should be done so by decorating the definition, *not* as a distinct line item:

```
@deprecated(error, “user object is deprecated”)
definition platform { ... }
```

kartikaysaxena:
> > requesting for initial review, the deprecation support now looks like
> > ```
> > use deprecation
> > 
> > definition user {}
> > definition superuser {}
> > definition anotheruser {}
> > 
> > definition platform {
> > 	relation admin: user
> > 	relation editor: superuser
> > 
> > 	@deprecated(warn, “rel viewer is deprecated”)
> > 	relation viewer: anotheruser
> > }
> > 
> > @deprecated(error, user, “user object is deprecated”)
> > @deprecated(warn, platform#editor, “this to deprecate a relation under a namespace”)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This adds support for deprecating relation inside a definition namespace for marking the immediate next relation as deprecated, for outside a namespace a relation can be marked deprecated as `@deprecated(warn, platform#editor, “this to deprecate a relation under a namespace”)` and an object by simply changing the relation with the object, example platform#editor -> user will mark user as deprecated, it _works_ if the object is either defined as a subject or a resource in a `WriteRelationship` call.
> > (also would migrate this pr to a new branch if this branch looks messy after too many rebases)
> 
> Sorry @kartikaysaxena, this is my bad at not being explicit: I liked the idea of being able to mark definitions as deprecated, but they should be done so by decorating the definition, _not_ as a distinct line item:
> 
> ```
> @deprecated(error, “user object is deprecated”)
> definition platform { ... }
> ```

understood, this is implemented and superseded by #2504, mind taking a look?

PR #2467, opened 2025-06-25
default cockroachdb connection jitter to 30m

ecordell:
In cockroach, the custom connection pooler allows us to have larger pools with longer lifetimes. With large pools, it's better to spread out the connection lifecycle with more jitter, so that there's not a stampede of connections closing / re-opening on regular intervals.

### DIFF
--- a/internal/datastore/crdb/options.go
+++ b/internal/datastore/crdb/options.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/utils/ptr"
+
 	"github.com/authzed/spicedb/internal/datastore/common"
 	pgxcommon "github.com/authzed/spicedb/internal/datastore/postgres/common"
 	log "github.com/authzed/spicedb/internal/logging"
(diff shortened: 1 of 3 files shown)


tstirrat15:
Yeah... on one hand this solves a problem we've got now; on the other hand it seems like something that we should fix at the level of the configuration options, not changing default behaviors down in the guts.

ecordell:
The cobra flag defaults affect all databases - if I change them there, we'll need to override the defaults for the non-crdb stores, where a long jitter may not be as safe. 

@vroldanbet specifically I believe you mentioned this was not safe for RDS since jitter would be longer than the connection drain timeout? If we think it's generally safe, I can make this the default everywhere, but I've only tested against crdb so far.

(even if I change them there, it's important to change the defaults "down in the guts" as well, since not everything that instantiates a datastore goes through the cobra CLI)

vroldanbet:
>The cobra flag defaults affect all databases - if I change them there, we'll need to override the defaults for the non-crdb stores, where a long jitter may not be as safe.

That's right, fair point. Taking a step back, the more fundamental problem we are facing is that we can't have datastore-specific flags, see https://github.com/authzed/spicedb/issues/606

>@vroldanbet specifically I believe you mentioned this was not safe for RDS since jitter would be longer than the connection drain timeout? If we think it's generally safe, I can make this the default everywhere, but I've only tested against crdb so far.

That's correct. Databases typically have a connection drain timeout, and the recommended practice is to align the Go SQL connection lifetime with it. You could still use jitter if you take into account connections will live `lifetime +/- jitter`. We could also change the jitter so it becomes `lifetime - jitter`. We could also make it a percentage of the lifetime. With those two things, we could come up with a sensible default that works for all databases out of the box.

For CockroachDB, since we are leveraging `crdbpool`, we can opt for different defaults, with the caveat that these changes will not be evident via the CLI help. Perhaps it's enough to edit the Cobra flag help and indicate the default is different for CRDB. 

> (even if I change them there, it's essential to change the defaults "down in the guts" as well, since not everything that instantiates a datastore goes through the cobra CLI)

agreed

ecordell:
updated the flag description

PR #2468, opened 2025-06-25
add tests for health check service

ecordell:
I wrote these tests because I suspected a bug here, but it looks like it was fine. The extra coverage doesn't hurt, though.

### DIFF
--- a/internal/services/health/health.go
+++ b/internal/services/health/health.go
@@ -108,6 +108,7 @@ func (hm *healthManager) checkIsReady(ctx context.Context) bool {
 	dsReady, err := hm.dsc.ReadyState(ctx)
 	if err != nil {
 		log.Ctx(ctx).Warn().Err(err).Msg("could not check if the datastore was ready")
+		return false
 	}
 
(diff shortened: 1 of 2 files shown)


tstirrat15 (dismissed):
LGTM

internal/services/health/health.go
```diff
@@ -108,6 +108,7 @@ func (hm *healthManager) checkIsReady(ctx context.Context) bool {
 	dsReady, err := hm.dsc.ReadyState(ctx)
 	if err != nil {
 		log.Ctx(ctx).Warn().Err(err).Msg("could not check if the datastore was ready")
+		return false
```
miparnisari, line 108: This change makes sense to me, but why aren't we doing both checks in parallel? Say the datastore takes forever to respond, but the dispatcher already errored out, we will be waiting `datastoreReadyTimeout` when we could have returned "false" very quickly.

It also seems to me that ` (hm *healthManager) Checker` can be simplified by using `backoff.Timer`.

PR #2475, opened 2025-06-30
misc cleanup

jzelinskie:
super low priority, but some of these are low hanging fruit i noticed when i wanted to take stock of why our builds are so large/slow.

each commit can be reviewed individually. if we don't want to bother with the `interface{} -> any` commit, i can drop it.


pkg/genutil/mapz/set.go
```diff
@@ -74,6 +74,18 @@ func (s *Set[T]) Union(other *Set[T]) *Set[T] {
 	return cpy
 }
 
+// Difference returns a new set with all of the values that not in the provided
+// sets.
+func (s *Set[T]) Difference(others ...*Set[T]) *Set[T] {
```
josephschorr, line 79: mind adding a unit test for this?

internal/datastore/mysql/datastore.go
```diff
@@ -92,7 +92,7 @@ func init() {
 }
 
 type sqlFilter interface {
-	ToSql() (string, []interface{}, error)
+	ToSql() (string, []any, error)
```
josephschorr, line 95: I'm surprised there isn't a linter rule that does this automatically...

vroldanbet (commented):
LGTM, pending `Difference()` unit test

internal/datastore/mysql/datastore.go
jzelinskie: i looked and didn't find one

miparnisari: There is, put this inside `.golangci.yaml`

```
formatters:
  enable:
    - "gci"
    - "gofmt"
    - "gofumpt"
    - "goimports"
  settings:
    gofmt:
      rewrite-rules:
        pattern: interface{}
        replacement: any
```

Ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23811#issuecomment-1612717328

jzelinskie: Is there a good way to exclude files? We don't want to rewrite anything with the codegen comment at the top. Maybe golangci-lint is already smart enough.

jzelinskie: actually nvm this just worked! i guess golangci is that smart.

miparnisari: This seems to work

```yaml
formatters:
  enable:
    - "gci"
    - "gofmt"
    - "gofumpt"
    - "goimports"
  settings:
    gofmt:
      rewrite-rules:
        pattern: interface{}
        replacement: any
    gci:
      sections:
        - "standard"
        - "default"
        - "prefix(github.com/authzed)"
        - "localmodule"
    goimports:
      local-prefixes:
        - "github.com/authzed/spicedb"
  exclusions:
    # Source files that have lines matching only the following regular expression will be excluded: `^// Code generated .* DO NOT EDIT\.$`
    generated: "strict"
```

Maybe we should send a separate PR for ^.

.golangci.yaml
```diff
@@ -87,9 +87,16 @@ linters:
...
+    - "gofmt"
     - "gofumpt"
     - "goimports"
   settings:
+    gofmt:
+      rewrite-rules:
+        - pattern: "interface{}"
+          replacement: "any"
+        - pattern: "a[b:len(a)]"
+          replacement: "a[b:]"
```
tstirrat15, line 99: This is nifty

.golangci.yaml
```diff
@@ -87,9 +87,16 @@ linters:
 formatters:
   enable:
     - "gci"
+    - "gofmt"
```
tstirrat15, line 90: Does adding gofmt here mean there's yet another tool that's processing the entire codebase?

Not a reason not to do it in my head but wanna be aware of the tradeoffs.

pkg/genutil/slicez/slicez.go
```diff
@@ -0,0 +1,41 @@
+package slicez
```
tstirrat15, line 1: This is awesome. I'm kinda surprised it's not somewhere in the stdlib yet 🤔

pkg/genutil/slicez/slicez.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	return ys
+}
+
+// Map iterates over a slice and creates a new slice with each element
+// transformed.
+func Map[T any, R any](xs []T, fn func(T) R) []R {
+	ys := make([]R, len(xs))
+	for i, x := range xs {
+		ys[i] = fn(x)
```
tstirrat15, line 20: For my own edification: is there a functional difference between doing a set at an index vs using `append` here?

tstirrat15 (dismissed):
LGTM, see comment

.golangci.yaml
jzelinskie: copypasta from the example on the goreleaser website, lol

.golangci.yaml
jzelinskie: Probably? I'm fairly certain goreleaser runs everything in parallel.

pkg/genutil/slicez/slicez.go
jzelinskie: lots of stuff missing for iterators too: https://github.com/jzelinskie/itz/blob/main/itz.go

pkg/genutil/slicez/slicez.go
jzelinskie: honestly just easier on the optimizer. ideally the compiler would realize that it never has to resize when doing append for that loop and it can optimize it into the same thing as this, but there's no risk of that if you just do what i did

PR #2486, opened 2025-07-03
pkg/cache: update to otter v2

jzelinskie:
Otter v2 has a design significantly more aligned with SpiceDB's cache usage than v1.

You can test otter by adding these hidden flags to `spicedb serve`: `--ns-cache-kind-for-testing otter --dispatch-cache-kind-for-testing otter --dispatch-cluster-cache-kind-for-testing otter`

### DIFF
--- a/pkg/cache/cache_otter.go
+++ b/pkg/cache/cache_otter.go
@@ -5,15 +5,15 @@ package cache
 
 import (
 	"math"
-	"sync"
+	"sync/atomic"
 
 	"github.com/ccoveille/go-safecast"
-	"github.com/maypok86/otter"
+	"github.com/maypok86/otter/v2"
+	"github.com/maypok86/otter/v2/stats"
 	"github.com/rs/zerolog"
 )
 
 func NewOtterCacheWithMetrics[K KeyString, V any](name string, config *Config) (Cache[K, V], error) {
-	// TODO: support metrics
 	return NewOtterCache[K, V](config)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/cache/cache_otter.go
```diff
@@ -23,42 +23,35 @@ type valueAndCost[V any] struct {
...
-	return &otterCache[K, V]{cache, sync.Once{}}, nil
+	cache, err := otter.New(opts)
+	return &otterCache[K, V]{cache, sync.Once{}, otterMetrics{counter}}, err
 }
 
 type otterCache[K KeyString, V any] struct {
-	cache  otter.Cache[string, valueAndCost[V]]
-	closed sync.Once
+	cache   *otter.Cache[string, valueAndCost[V]]
+	closed  sync.Once
```
vroldanbet: It doesn't seem to be used anymore, or have you forgotten to add the `Close` logic?

pkg/cache/cache_otter.go
```diff
@@ -67,27 +60,27 @@ func (wtc *otterCache[K, V]) Get(key K) (V, bool) {
...
+func (wtc *otterCache[K, V]) Close() {}
 
-func (wtc *otterCache[K, V]) Close() {
-	wtc.closed.Do(func() {
-		wtc.cache.Close()
-	})
-}
+type otterMetrics struct{ *stats.Counter }
+
+func (o *otterMetrics) CostAdded() uint64   { return 0 } // Unsupported
```
vroldanbet: idea: we could keep this ourselves in `otherMetrics` as `Set` is called

jzelinskie: great point. added

vroldanbet (approved):

PR #2490, opened 2025-07-07
docs: improve docs of flags

miparnisari:
I went over the newly updated https://authzed.com/docs/spicedb/concepts/commands.

Pending:
```
 --dispatch-cache-enabled                                  enable caching (default true)
 --dispatch-cluster-cache-enabled                     enable caching (default true)
 --dispatch-cluster-enabled                                enable dispatch gRPC server
```

### DIFF
--- a/pkg/cmd/migrate.go
+++ b/pkg/cmd/migrate.go
@@ -159,7 +159,7 @@ func RegisterHeadFlags(cmd *cobra.Command) {
 func NewHeadCommand(programName string) *cobra.Command {
 	return &cobra.Command{
 		Use:     "head",
-		Short:   "compute the head database migration revision",
+		Short:   "compute the head (latest) database migration revision available",
 		PreRunE: server.DefaultPreRunE(programName),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			headRevision, err := HeadRevision(cobrautil.MustGetStringExpanded(cmd, "datastore-engine"))
--- a/pkg/cmd/serve.go
+++ b/pkg/cmd/serve.go
@@ -62,13 +62,10 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	nfs := cobrautil.NewNamedFlagSets(cmd)
 
 	grpcFlagSet := nfs.FlagSet(BoldBlue("gRPC"))
-	// Flags for logging
-	grpcFlagSet.BoolVar(&config.EnableRequestLogs, "grpc-log-requests-enabled", false, "logs API request payloads")
-	grpcFlagSet.BoolVar(&config.EnableResponseLogs, "grpc-log-responses-enabled", false, "logs API response payloads")
 
 	// Flags for the gRPC API server
 	util.RegisterGRPCServerFlags(grpcFlagSet, &config.GRPCServer, "grpc", "gRPC", ":50051", true)
-	grpcFlagSet.StringSliceVar(&config.PresharedSecureKey, PresharedKeyFlag, []string{}, "preshared key(s) to require for authenticated requests")
+	grpcFlagSet.StringSliceVar(&config.PresharedSecureKey, PresharedKeyFlag, []string{}, "(required) preshared key(s) that must be provided by clients to authenticate requests")
 	grpcFlagSet.DurationVar(&config.ShutdownGracePeriod, "grpc-shutdown-grace-period", 0*time.Second, "amount of time after receiving sigint to continue serving")
 	if err := cobra.MarkFlagRequired(grpcFlagSet, PresharedKeyFlag); err != nil {
 		return fmt.Errorf("failed to mark flag as required: %w", err)
@@ -76,7 +73,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 
 	// Flags for HTTP gateway
 	httpFlags := nfs.FlagSet(BoldBlue("HTTP"))
-	util.RegisterHTTPServerFlags(httpFlags, &config.HTTPGateway, "http", "gateway", ":8443", false)
+	util.RegisterHTTPServerFlags(httpFlags, &config.HTTPGateway, "http", "proxy", ":8443", false)
 	httpFlags.StringVar(&config.HTTPGatewayUpstreamAddr, "http-upstream-override-addr", "", "Override the upstream to point to a different gRPC server")
 	if err := httpFlags.MarkHidden("http-upstream-override-addr"); err != nil {
 		return fmt.Errorf("failed to mark flag as hidden: %w", err)
@@ -106,7 +103,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	apiFlags.Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	apiFlags.IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes. A value of zero or less means no limit")
 	apiFlags.IntVar(&config.MaxRelationshipContextSize, "max-relationship-context-size", 25000, "maximum allowed size of the context to be stored in a relationship")
-	apiFlags.DurationVar(&config.StreamingAPITimeout, "streaming-api-response-delay-timeout", 30*time.Second, "max duration time elapsed between messages sent by the server-side to the client (responses) before the stream times out")
+	apiFlags.DurationVar(&config.StreamingAPITimeout, "streaming-api-response-delay-timeout", 30*time.Second, "maximum duration time elapsed between messages sent by the server-side to the client (responses) before the stream times out")
 	apiFlags.DurationVar(&config.WatchHeartbeat, "watch-api-heartbeat", 1*time.Second, "heartbeat time on the watch in the API. 0 means to default to the datastore's minimum.")
 	apiFlags.Uint32Var(&config.MaxReadRelationshipsLimit, "max-read-relationships-limit", 1000, "maximum number of relationships that can be read in a single request")
 	apiFlags.Uint32Var(&config.MaxDeleteRelationshipsLimit, "max-delete-relationships-limit", 1000, "maximum number of relationships that can be deleted in a single request")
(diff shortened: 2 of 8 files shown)


pkg/cmd/server/defaults.go
```diff
@@ -54,12 +54,14 @@ func ServeExample(programName string) string {
 		%[3]s serve --grpc-preshared-key "somerandomkeyhere"
 
 	%[2]s:
-		%[3]s serve --grpc-preshared-key "realkeyhere" --grpc-tls-cert-path path/to/tls/cert --grpc-tls-key-path path/to/tls/key \
-			--http-tls-cert-path path/to/tls/cert --http-tls-key-path path/to/tls/key \
-			--datastore-engine postgres --datastore-conn-uri "postgres-connection-string-here"
+		%[3]s serve --grpc-preshared-key "realkeyhere" \
+		--grpc-tls-cert-path path/to/tls/cert --grpc-tls-key-path path/to/tls/key \
+		--http-enabled http-tls-cert-path path/to/tls/cert --http-tls-key-path path/to/tls/key \
```
miparnisari, line 59: Added `http-enabled`

pkg/cmd/util/util.go
```diff
@@ -68,7 +68,7 @@ func RegisterGRPCServerFlags(flags *pflag.FlagSet, config *GRPCServerConfig, fla
 	flags.StringVar(&config.TLSCertPath, flagPrefix+"-tls-cert-path", "", "local path to the TLS certificate used to serve "+serviceName)
 	flags.StringVar(&config.TLSKeyPath, flagPrefix+"-tls-key-path", "", "local path to the TLS key used to serve "+serviceName)
 	flags.DurationVar(&config.MaxConnAge, flagPrefix+"-max-conn-age", 30*time.Second, "how long a connection serving "+serviceName+" should be able to live")
-	flags.BoolVar(&config.Enabled, flagPrefix+"-enabled", defaultEnabled, "enable "+serviceName+" gRPC server")
+	flags.BoolVar(&config.Enabled, flagPrefix+"-enabled", defaultEnabled, "enable "+serviceName+" server")
```
miparnisari, line 71: Fixes

```
 --grpc-enabled                          enable gRPC gRPC server (default true)
```

to

```
--grpc-enabled                          enable gRPC server (default true)
```

pkg/cmd/migrate.go
```diff
@@ -159,7 +159,7 @@ func RegisterHeadFlags(cmd *cobra.Command) {
 func NewHeadCommand(programName string) *cobra.Command {
 	return &cobra.Command{
 		Use:     "head",
-		Short:   "compute the head database migration revision",
+		Short:   "compute the head (latest) database migration revision available",
```
miparnisari, line 162: I've never heard anyone saying "the _head_ database migration" 🤷‍♀️

pkg/cmd/serve.go
```diff
@@ -62,21 +62,18 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
...
+	grpcFlagSet.StringSliceVar(&config.PresharedSecureKey, PresharedKeyFlag, []string{}, "(required) preshared key(s) that must be provided by clients to authenticate requests")
 	grpcFlagSet.DurationVar(&config.ShutdownGracePeriod, "grpc-shutdown-grace-period", 0*time.Second, "amount of time after receiving sigint to continue serving")
 	if err := cobra.MarkFlagRequired(grpcFlagSet, PresharedKeyFlag); err != nil {
 		return fmt.Errorf("failed to mark flag as required: %w", err)
 	}
 
 	// Flags for HTTP gateway
 	httpFlags := nfs.FlagSet(BoldBlue("HTTP"))
-	util.RegisterHTTPServerFlags(httpFlags, &config.HTTPGateway, "http", "gateway", ":8443", false)
+	util.RegisterHTTPServerFlags(httpFlags, &config.HTTPGateway, "http", "HTTP", ":8443", false)
```
miparnisari: From

```
--http-enabled                enable http gateway server
```

to

```
--http-enabled                enable http HTTP server
```

. `gateway` sounds like a Go term to me

tstirrat15: I'm not sure I like the stutter. "gateway" makes sense to me from a frontend perspective, a la "api gateway", but if you don't like that I'd want to find another word. "wrapper?" "proxy?" "fronting?"

pkg/cmd/serve.go
```diff
@@ -171,23 +168,27 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 			Msg("The old implementation of LookupResources is no longer available, and a `false` value is no longer valid. Please remove this flag.")
 	}
 
-	experimentalFlags.BoolVar(&config.EnableExperimentalRelationshipExpiration, "enable-experimental-relationship-expiration", false, "enables experimental support for first-class relationship expiration")
-	experimentalFlags.BoolVar(&config.EnableExperimentalWatchableSchemaCache, "enable-experimental-watchable-schema-cache", false, "enables the experimental schema cache which makes use of the Watch API for automatic updates")
+	experimentalFlags.BoolVar(&config.EnableExperimentalRelationshipExpiration, "enable-experimental-relationship-expiration", false, "enables experimental support for relationship expiration")
+	experimentalFlags.BoolVar(&config.EnableExperimentalWatchableSchemaCache, "enable-experimental-watchable-schema-cache", false, "enables the experimental schema cache, which uses the Watch API to keep the schema up to date")
```
tstirrat15, line 172: I'm not entirely sure this is correct anymore. I want to say it's polling, and that it's using the datastore's watch API, not the external Watch API. I think @vroldanbet would know better.

pkg/cmd/serve.go
```diff
@@ -171,23 +168,27 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
...
 	// TODO: these two could reasonably be put in either the Dispatch group or the Experimental group. Is there a preference?
 	experimentalFlags.StringToStringVar(&config.DispatchSecondaryUpstreamAddrs, "experimental-dispatch-secondary-upstream-addrs", nil, "secondary upstream addresses for dispatches, each with a name")
 	experimentalFlags.StringToStringVar(&config.DispatchSecondaryUpstreamExprs, "experimental-dispatch-secondary-upstream-exprs", nil, "map from request type to its associated CEL expression, which returns the secondary upstream(s) to be used for the request")
 	experimentalFlags.StringToStringVar(&config.DispatchSecondaryMaximumPrimaryHedgingDelays, "experimental-dispatch-secondary-maximum-primary-hedging-delays", nil, "maximum number of hedging delays to use for each request type to delay the primary request. default is 5ms")
 
-	observabilityFlags := nfs.FlagSet(BoldBlue("Observability"))
+	observabilityFlags := nfs.FlagSet(BoldBlue("Tracing"))
```
tstirrat15: Mmm... We mostly currently use these for tracing, but they're not specifically for tracing.

pkg/cmd/serve.go
```diff
@@ -171,23 +168,27 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
...
 	// Flags for observability and profiling
 	// NOTE: cobraotel.New takes service name as an arg rather than command name.
 	otel := cobraotel.New("spicedb")
 	otel.RegisterFlags(observabilityFlags)
-	runtime.RegisterFlags(observabilityFlags)
```
tstirrat15, line 186: Did we lose this one?

pkg/cmd/migrate.go
jzelinskie: We've used the term internally a decent amount because many folks had experience with alembic, a python migration framework that uses the term. I think alembic might be more powerful than what many other migration frameworks support (diverging trees, etc).

If it's not familiar for folks, we should use different terminology.

pkg/cmd/serve.go
miparnisari: No - instead of having the pprof flags under `observability`, I put them under `metrics`, since pprof is enabled when you turn on metrics: https://github.com/authzed/spicedb/blob/895f2b27e5936e4227936886e2ffa53fb1f59751/pkg/cmd/server/defaults.go#L93-L107

Image

pkg/cmd/migrate.go
miparnisari: I associated it with the git head revision, but it still sounded weird

pkg/cmd/serve.go
miparnisari: Ah! Proxy makes more sense yeah

pkg/cmd/serve.go
miparnisari: Judging by the description of all current flags

```
      --otel-endpoint string           OpenTelemetry collector endpoint - the endpoint can also be set by using enviroment variables
      --otel-insecure                  connect to the OpenTelemetry collector in plaintext
      --otel-provider string           OpenTelemetry provider for tracing ("none", "otlphttp", "otlpgrpc") (default "none")
      --otel-sample-ratio float        ratio of traces that are sampled (default 0.01)
      --otel-service-name string       service name for trace data (default "spicedb")
      --otel-trace-propagator string   OpenTelemetry trace propagation format ("b3", "w3c", "ottrace"). Add multiple propagators separated by comma. (default "w3c")

```

we do use otel just for tracing. I prefer to capture this with a more specific title. It's easy to change this grouping title in the future.

pkg/cmd/server/cacheconfig.go
```diff
@@ -135,7 +135,7 @@ func MustRegisterCacheFlags(flags *pflag.FlagSet, flagPrefix string, config, def
 	config.Name = defaults.Name
 	flagPrefix = stringz.DefaultEmpty(flagPrefix, "cache")
 	flags.StringVar(&config.MaxCost, flagPrefix+"-max-cost", defaults.MaxCost, "upper bound cache size in bytes or percent of available memory")
-	flags.Int64Var(&config.NumCounters, flagPrefix+"-num-counters", defaults.NumCounters, "number of TinyLFU samples to track")
+	flags.Int64Var(&config.NumCounters, flagPrefix+"-num-counters", defaults.NumCounters, "number of TinyLFU samples to track. A higher number means more accurate eviction decisions but more memory usage")
```
miparnisari, line 138: I think in general, if a CLI flag affects behavior, we should give hints as to how to configure them, i.e. trade-offs

PR #2499, opened 2025-07-12
Parse schema from schemafile property

noseworthy:
Support for the `schemaFile` property in the validation file was added in [#2206](https://github.com/authzed/spicedb/pull/2206) but it was never fully integrated in to the bootstrap/import process.

This change adds support similar to the validator from the `zed` cli which was added in [zed#478](https://github.com/authzed/zed/pull/478).

If both `schema` and `schemaFile` are specified, an error is thrown. If only `schemaFile` is specified, the contents of the linked file are read and passed to the compilation function. If `schema` is specified or missing, this function behaves as it always has.

### DIFF
--- a/pkg/validationfile/loader.go
+++ b/pkg/validationfile/loader.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/ccoveille/go-safecast"
 
@@ -21,7 +23,6 @@ import (
 	"github.com/authzed/spicedb/pkg/schemadsl/input"
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/tuple"
-	"github.com/authzed/spicedb/pkg/validationfile/blocks"
 )
 
 // PopulatedValidationFile contains the fully parsed information from a validation file.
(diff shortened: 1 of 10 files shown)


noseworthy:

noseworthy:
recheck

noseworthy:
Hey authzed team, I noticed that I couldn't use the `schemaFile:` property in a validation file when bootstrapping our local development spicedb instance. I'm not a go programmer and don't have much experience with the internals of spicedb but figured I'd give a fix a shot. I shamelessly pulled some of this from the linked `zed` implementation.

Hopefully this is useful. I feel sort of bad that I had to change the `compile()` prototype, but it seemed fairly local/private to the loader so I figured it wasn't _too_ bad a modification.

Let me know what you think!

pkg/validationfile/loader.go
```diff
@@ -94,8 +94,38 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
...
+			return nil, datastore.NoRevision, fmt.Errorf("only one of schema or schemaFile can be specified")
+		}
+
+		var parsedSchemaString string
+		if len(parsed.SchemaFile) > 0 {
+			// Need to join the original filepath with the requested filepath
+			// to construct the path to the referenced schema file.
+			// NOTE: This does not allow for yaml files to transitively reference
+			// each other's schemaFile fields.
+			// TODO: enable this behavior
```
josephschorr: Remove this TODO; YAML loading shouldn't be supported here

pkg/validationfile/loader.go
```diff
@@ -94,8 +94,38 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
...
+			schemaPath := filepath.Join(filepath.Dir(filePath), parsed.SchemaFile)
+
+			if !filepath.IsLocal(schemaPath) {
+				// We want to prevent access of files that are outside of the folder
+				// where the command was originally invoked. This should do that.
+				return nil, datastore.NoRevision, fmt.Errorf("schema file %s is not local", schemaPath)
```
josephschorr: change to %q

pkg/validationfile/loader_test.go
```diff
@@ -74,6 +84,12 @@ func TestPopulateFromFiles(t *testing.T) {
 			want:          nil,
 			expectedError: "object definition `example/project` not found",
 		},
+		{
+			name:          "both schema and schemaFile",
+			filePaths:     []string{"testdata/schema_and_schemafile.yaml"},
+			want:          nil,
+			expectedError: "only one of schema or schemaFile can be specified",
```
josephschorr: please add a test with an error in the schema itself and ensure the returned error has the correct line/col position and filename

josephschorr:
Also please fix lint errors; you can run `mage lint:all` from the root directory to test/fix locally

noseworthy:
Thanks for the review @josephschorr 🙏! I really appreciate it.

Looks like a bunch of relatively minor things, I'll get to them as soon as I can. Might not be until the weekend due to work commitments.

pkg/validationfile/loader_test.go
noseworthy: I've added some more tests. One for the case you mentioned and two more for additional code coverage based on the Codecov report
1. invalid schemaFile. I check that the line numbers are correct, but as for the filename, I feel like that's more of a refactor than I'm comfortable making. The errors only care about the source code, not the file name. I hope that's okay.
2. Checking unable to read schema file error
3. Checking that we can't load a schema file that isn't local to the validation `yaml` file.

PR #2503, opened 2025-07-21
OpenTelemetry Semantic Conventions for Tracing

### LINKED ISSUE #1548: Tracing should follow OpenTelemetry Semantic Conventions

linux, macos, windows

amd64, arm64

### Steps to Reproduce

1. Run SpiceDB with tracing enabled
2. View tracing names and observe that not all follow the [OpenTelemetry Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/)

### Expected Result

The traces follow the well defined conventions so that they combine properly with other systems and are consumed properly by tracing processors.

### Actual Result

SpiceDB has used their own conventions for naming events and attributes.

mazdakb:
Fixes #1548

- Renamed OpenTelemetry tracing event and attribute names to follow [OpenTelemetry Trace semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/trace/)
- Introduced `otelconv` utility package under `telemetry` to centralize OpenTelemetry convention constants
- Replaced hardcoded telemetry event/attribute names with `otelconv` constants across multiple packages
- Added custom linter `telemetryconvcheck` to enforce proper usage of `otelconv` constants
- Consolidated attributes to reduce cardinality
- Where necessary, moved event timing to capture traces after operations complete rather than before they start


internal/datastore/mysql/revisions.go
```diff
@@ -153,7 +156,10 @@ func (mds *Datastore) checkValidTransaction(ctx context.Context, revisionTx uint
 		return false, false, fmt.Errorf(errCheckRevision, err)
 	}
 
-	span.AddEvent("DB returned validTransaction checks")
+	span.AddEvent(otelconv.EventDatastoreMySQLTransactionValidated, trace.WithAttributes(
```
josephschorr, line 159: Can we go one step further and classify events as "internal" vs "external", and name them accordingly, like `spicedb-internal.` vs `spicedb.`? This will allow us to filter traces in a nice way down the road.

internal/telemetry/otelconv/otelconv.go
```diff
@@ -0,0 +1,97 @@
...
+
+// This file contains constants for telemetry instrumentation that follow the OpenTelemetry
+// Semantic Conventions for naming. The naming conventions adhere to the guidelines specified at:
+// https://opentelemetry.io/docs/specs/semconv/general/naming/
+//
+// The events and attributes defined here are used throughout the SpiceDB codebase to provide
+// a consistent naming scheme for spans, events, attributes, etc.
+//
+// IMPORTANT: When adding a new _custom_ name, consider the following conventions:
+//  - Start all names with "spicedb" prefix to identify our application domain
```
josephschorr: Add a string slice containing all of the event names and all the attributes, and then add a unit test enforcing these rules on all names found

internal/datastore/proxy/observable.go
```diff
@@ -297,9 +298,9 @@ func (rwt *observableRWT) UnregisterCounter(ctx context.Context, name string) er
 
 func (rwt *observableRWT) StoreCounterValue(ctx context.Context, name string, value int, computedAtRevision datastore.Revision) error {
 	ctx, closer := observe(ctx, "StoreCounterValue", "", trace.WithAttributes(
-		attribute.String("name", name),
-		attribute.Int("value", value),
-		attribute.String("revision", computedAtRevision.String()),
+		attribute.StringSlice(otelconv.AttrDatastoreNames, []string{name}),
```
miparnisari, line 301: i'm curious, why a slice if it's just one value?

internal/datastore/mysql/revisions.go
mazdakb: That'd be good. The only thing I'd suggest is to have it separated with a dot like `spicedb.internal`, to follow the OTel convention.

internal/datastore/proxy/observable.go
mazdakb: Good point! Initially there were two separate attributes, one for a singular **datastore name** and another for a list of **datastore names** logged in different locations. I've consolidated them into one, in order to avoid having different attributes that represent the same type of source. In other words, to minimize cardinality.

internal/datastore/mysql/revisions.go
mazdakb: Is this distinction made by the fact that, right now, traces originate in `internal` dir?

tools/analyzers/telemetryconvcheck/telemetryconvcheck.go
```diff
@@ -0,0 +1,201 @@
...
+	"go/types"
+	"strings"
+
+	"github.com/samber/lo"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/ast/inspector"
+)
+
+const otelconvPackage = "github.com/authzed/spicedb/internal/telemetry/otelconv"
```
miparnisari: Could we make this a configurable flag with this value as default? Otherwise, if we want to re-use this analyzer in another repo, we would have to come edit SpiceDB

We could make it a list, too, so we could define constants in multiple places

internal/telemetry/otelconv/otelconv.go
mazdakb: Would adding the check to the linter be also fine? That's an easy addition and has the benefit of slightly better dev experience IMO because you won't need to remember to add new constants to the list.

tools/analyzers/telemetryconvcheck/telemetryconvcheck.go
mazdakb: Good idea! I added a `conv-pkg` flag which allows the caller to designate the package where in the constants are defined.
However, I've not made it to accept a list, because IMO the one-target constraint is useful for keeping constants from spreading throughout the codebase. The caller can still exec multiple instances of the analyzer to check multiple paths with different targets in parallel though.

internal/datastore/mysql/revisions.go
josephschorr: No, its more a judgement call on a case by case basis: whether we'd fine it useful outside of debugging

internal/telemetry/otelconv/otelconv.go
josephschorr: Sure

tools/analyzers/telemetryconvcheck/telemetryconvcheck.go
miparnisari: But wouldn't that error out? If you run the analyzer with `analyzer -conv-pkg="a"`, it will fail if the constants are defined in `b`?

internal/datastore/mysql/revisions.go
mazdakb: Now I get what you mean. Added the two prefixes and set them in linter.

internal/telemetry/otelconv/otelconv.go
mazdakb: Done

tools/analyzers/telemetryconvcheck/telemetryconvcheck.go
mazdakb: You are right, that'd error out. But it should still be usable within different directories or by calling the analyzer multiple times with different `skip-pkg` arguments.

PR #2504, opened 2025-07-21
feat: support for rel and namespace deprecation

kartikaysaxena:
This PR adds support for relation and namespace level deprecation in schemas.

- Relations and namespaces can now be annotated with a `@deprecated(...)` directive to indicate they are deprecated.
- When a relationship is written to a deprecated relation or deprecated namespace:
  - A **warning** or **error** is generated, depending on the specified deprecation level.
  - In the current implementation, writes fail when an error deprecation is configured.
- Deprecation is associated with the definition/relation decorated with the deprecation.
- A flag is to be passed to enable this feature, namely `--enable-experimental-deprecation` along with use directive.

**Example:**
```dsl
use deprecation

@deprecated(warn, "comments")
definition user {}

definition resource {
  relation viewer: user | @deprecated(warn, "documents can no longer be public") user:*

  @deprecated(error, "comment goes here")
  relation admin: user
}
```
> [!NOTE]
>Each relation has a object `Deprecation` assigned to them containing an enum `DeprecationType` and optional comments, if unspecified DeprecationType defaults to `DEPRECATED_TYPE_UNSPECIFIED`

> [!NOTE]
>Attempted write with a `DELETE` operation to a deprecated relation/namespace doesn't fail

(related #2465)


internal/services/server.go
```diff
@@ -73,6 +73,7 @@ func RegisterGrpcServices(
 			CaveatTypeSet:                    permSysConfig.CaveatTypeSet,
 			AdditiveOnly:                     schemaServiceOption == V1SchemaServiceAdditiveOnly,
 			ExpiringRelsEnabled:              permSysConfig.ExpiringRelationshipsEnabled,
+			DeprecatedRelAndObjEnabled:       permSysConfig.DeprecatedRelationshipsAndObjectsEnabled,
```
josephschorr: Maybe just `FeatureDeprecationEnabled`?

internal/services/shared/errors.go
```diff
@@ -47,6 +47,33 @@ func NewSchemaWriteDataValidationError(message string, args ...any) SchemaWriteD
...
+type DeprecationError struct {
+	error
+}
+
+func (err DeprecationError) GRPCStatus() *status.Status {
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.Aborted,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_SCHEMA_TYPE_ERROR,
```
josephschorr, line 61: Add a new error reason into the API and use it here

internal/services/shared/errors.go
```diff
@@ -47,6 +47,33 @@ func NewSchemaWriteDataValidationError(message string, args ...any) SchemaWriteD
...
+			map[string]string{},
+		),
+	)
+}
+
+func NewDeprecationError(namespace string, relation string, comments string) DeprecationError {
+	if relation == "" {
+		return DeprecationError{
+			error: fmt.Errorf("object %s is deprecated, comments:%s", namespace, comments),
```
josephschorr: `"object type %s is deprecated: %s"`. If the comments are empty, put "object type %s is marked as deprecated"

internal/services/v1/relationships.go
```diff
@@ -620,3 +632,74 @@ func labelsForFilter(filter *v1.RelationshipFilter) perfinsights.APIShapeLabels
...
 }
+
+func checkForDeprecatedRelationsAndObjects(ctx context.Context, update *v1.RelationshipUpdate, ds datastore.Datastore) error {
+	resource := update.Relationship.Resource
+	headRevision, err := ds.HeadRevision(ctx)
+	if err != nil {
+		return err
+	}
+	reader := ds.SnapshotReader(headRevision)
+	_, relDef, err := namespace.ReadNamespaceAndRelation(ctx, resource.ObjectType, update.Relationship.Relation, reader)
```
josephschorr: rather than reading the namespace again, let's do this check in the existing validation block, just as an additional check if enabled

internal/services/v1/relationships.go
```diff
@@ -620,3 +632,74 @@ func labelsForFilter(filter *v1.RelationshipFilter) perfinsights.APIShapeLabels
...
+	if err != nil {
+		return err
+	}
+
+	if relDef.Deprecation != nil && relDef.Deprecation.DeprecationType != corev1.DeprecationType_DEPRECATED_TYPE_UNSPECIFIED {
```
josephschorr: I think you can add deprecation to the type system and just read it from there

ecordell:
Just thinking about the UX a bit (we don't have to solve it in this PR - just thinking ahead!) it would be nice to be able to do something like:

```
definition document {
  relation viewer: user

  @deprecated(warn, "documents can no longer be public")
  relation viewer: user:*
}
```
which would normally fail validation as a duplicated relation name

josephschorr:
> Just thinking about the UX a bit (we don't have to solve it in this PR - just thinking ahead!) it would be nice to be able to do something like:
> 
> ```
> definition document {
>   relation viewer: user
> 
>   @deprecated(warn, "documents can no longer be public")
>   relation viewer: user:*
> }
> ```
> 
> which would normally fail validation as a duplicated relation name

Not sure... I kind of prefer:

```
relation viewer: user | @deprecated(warn, "documents can no longer be public") user:*
```

internal/services/shared/errors.go
kartikaysaxena: just raising that into the api repo, once that gets in, i will switch to that

internal/services/v1/relationships.go
kartikaysaxena: migrated the check to the relationship validation block to remove unnecessary lookups, also this now uses the typesystem for the read, also added a test

kartikaysaxena:
> > Just thinking about the UX a bit (we don't have to solve it in this PR - just thinking ahead!) it would be nice to be able to do something like:
> > ```
> > definition document {
> >   relation viewer: user
> > 
> >   @deprecated(warn, "documents can no longer be public")
> >   relation viewer: user:*
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which would normally fail validation as a duplicated relation name
> 
> Not sure... I kind of prefer:
> 
> ```
> relation viewer: user | @deprecated(warn, "documents can no longer be public") user:*
> ```

this would require extending the support to the type reference, would look into that, or address do that in a follow up

internal/services/v1/relationships.go
josephschorr: Can we add a method into the type system to retrieve the deprecation status of a namespace and/or relation and then use that? Its good to have it in a single place

josephschorr:
> > > Just thinking about the UX a bit (we don't have to solve it in this PR - just thinking ahead!) it would be nice to be able to do something like:
> > > ```
> > > definition document {
> > >   relation viewer: user
> > > 
> > >   @deprecated(warn, "documents can no longer be public")
> > >   relation viewer: user:*
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > which would normally fail validation as a duplicated relation name
> > 
> > 
> > Not sure... I kind of prefer:
> > ```
> > relation viewer: user | @deprecated(warn, "documents can no longer be public") user:*
> > ```
> 
> this would require extending the support to the type reference, would look into that, or address do that in a follow up

Followup is fine

internal/relationships/validation.go
```diff
@@ -277,3 +284,106 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
...
+		case core.DeprecationType_DEPRECATED_TYPE_ERROR:
+			return shared.NewDeprecationError(rel.Subject.ObjectType, "", subjectNS.Deprecation.Comments)
+		}
+	}
+	// iterate over the allowed relations in the resource definition
+	// and see whether the concrete subject we are writing matches an
+	// entry that has a non-nil deprecation
+	for _, allowed := range relDef.TypeInformation.AllowedDirectRelations {
+		// check namespace
+		if allowed.Namespace != rel.Subject.ObjectType {
```
kartikaysaxena: this iterates over the allowed relation types for a relation to validate the deprecation type, it now supports `relation viewer: user | @deprecated(warn, "documents can no longer be public") user:*`, also added tests for it.

kartikaysaxena:
> > > > Just thinking about the UX a bit (we don't have to solve it in this PR - just thinking ahead!) it would be nice to be able to do something like:
> > > > ```
> > > > definition document {
> > > >   relation viewer: user
> > > > 
> > > >   @deprecated(warn, "documents can no longer be public")
> > > >   relation viewer: user:*
> > > > }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > which would normally fail validation as a duplicated relation name
> > > 
> > > 
> > > Not sure... I kind of prefer:
> > > ```
> > > relation viewer: user | @deprecated(warn, "documents can no longer be public") user:*
> > > ```
> > 
> > 
> > this would require extending the support to the type reference, would look into that, or address do that in a follow up
> 
> Followup is fine

I added it just a while ago 😅, have added tests too, if anything looks sideways, would revert back to previous commit, or if any other suggestion is taken into consideration regarding the type ref approach for relation deprecation, let me know.

pkg/schema/typesystem.go
```diff
@@ -38,6 +38,87 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
...
+		return nil, false, err
+	}
+
+	rel, ok := def.GetRelation(resourceRelation)
+	if !ok {
+		return nil, false, nil
+	}
+
+	for _, allowed := range rel.TypeInformation.GetAllowedDirectRelations() {
+		// check namespace match
```
kartikaysaxena, line 110: could've used `ts.GetFullRecursiveSubjectTypesForRelation()` but that too would've took O(N) time and would want a `*core.AllowedRelation` at other steps as it gives the string types only

internal/services/v1/permissions.go
```diff
@@ -808,7 +808,7 @@ type loadBulkAdapter struct {
 	err          error
 }
 
-func (a *loadBulkAdapter) Next(_ context.Context) (*tuple.Relationship, error) {
+func (a *loadBulkAdapter) Next(ctx context.Context) (*tuple.Relationship, error) {
```
kartikaysaxena, line 811: wired this ctx upstream in a few places as `relationships.ValidateOneRelationship` now requires ctx (for typesystem)

internal/services/v1/relationships.go
kartikaysaxena: done

internal/relationships/validation.go
```diff
@@ -34,6 +36,19 @@ func ValidateRelationshipUpdates(
...
+	ts := schema.NewTypeSystem(schema.ResolverForDatastoreReader(reader))
+
+	// check if the relation is deprecated for create or touch operations
+	var relsToCheck []tuple.Relationship
+	for _, update := range updates {
+		if update.Operation == tuple.UpdateOperationTouch || update.Operation == tuple.UpdateOperationCreate {
+			relsToCheck = append(relsToCheck, update.Relationship)
+		}
+	}
+	if err := CheckDeprecationsOnRelationships(ctx, relsToCheck, ts); err != nil {
```
kartikaysaxena: a little check here as we would not want an attempted write with `DELETE` operation to a deprecated relation or namespace to fail

internal/relationships/validation.go
```diff
@@ -277,3 +299,104 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
...
+	for _, rel := range relationships {
+		if err := checkForDeprecatedRelationsAndObjects(ctx, rel, ts); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func checkForDeprecatedRelationsAndObjects(ctx context.Context, rel tuple.Relationship, ts *schema.TypeSystem) error {
+	// Validate if the resource relation is deprecated
```
kartikaysaxena: this now uses the typesystem with relevant functions

internal/services/v1/relationships.go
kartikaysaxena: done

internal/services/shared/errors.go
kartikaysaxena: here: https://github.com/authzed/api/pull/143

internal/relationships/validation.go
```diff
@@ -277,3 +299,104 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
...
+	relDep, ok, err := ts.GetDeprecationForRelation(ctx, rel.Resource.ObjectType, rel.Resource.Relation)
+	if err != nil {
+		return err
+	}
+	if ok {
+		switch relDep.DeprecationType {
+		case core.DeprecationType_DEPRECATED_TYPE_WARNING:
+			log.Warn().
+				Str("namespace", rel.Resource.ObjectType).
```
josephschorr: namespace -> resource_type

internal/relationships/validation.go
```diff
@@ -277,3 +299,104 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
...
+				Str("relation", rel.Resource.Relation).
+				Str("comments", relDep.Comments).
+				Msg("write to deprecated relation")
+		case core.DeprecationType_DEPRECATED_TYPE_ERROR:
+			return shared.NewDeprecationError(rel.Resource.ObjectType, rel.Resource.Relation, relDep.Comments)
+		}
+	}
+
+	// Validate if the resource namespace is deprecated
+	resDep, ok, err := ts.GetDeprecationForNamespace(ctx, rel.Resource.ObjectType)
```
josephschorr: `GetDeprecationForObjectType`

internal/relationships/validation.go
```diff
@@ -277,3 +299,104 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
...
+		}
+	}
+
+	// Validate if the subject namespace is deprecated
+	subDep, ok, err := ts.GetDeprecationForNamespace(ctx, rel.Subject.ObjectType)
+	if err != nil {
+		return err
+	}
+	if ok {
+		switch subDep.DeprecationType {
```
josephschorr: make this "warning vs error" into a helper method and use here and the other two locations

internal/relationships/validation.go
```diff
@@ -277,3 +299,104 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
...
+				Str("comments", subDep.Comments).
+				Msg("write to deprecated object")
+		case core.DeprecationType_DEPRECATED_TYPE_ERROR:
+			return shared.NewDeprecationError(rel.Subject.ObjectType, "", subDep.Comments)
+		}
+	}
+
+	// check deprecation for allowed relation types
+	dep, ok, err := ts.GetDeprecationForAllowedRelation(
+		ctx,
```
josephschorr: might be worth having a method `GetDeprecationForRelationship` and have it return an error/warning message and what kind it is, if any

internal/relationships/validation_test.go
```diff
@@ -306,6 +306,63 @@ func TestValidateRelationshipOperations(t *testing.T) {
...
+			definition user {}
+
+			definition document {
+				@deprecated(error,"deprecated, migrate away")
+				relation editor: testuser
+				relation viewer: user
+			}`,
+			"document:foo#editor@testuser:tom",
+			core.RelationTupleUpdate_CREATE,
+			"the relation document#editor is deprecated: deprecated, migrate away",
```
josephschorr: drop the `the`, its cleaner :) - `relation ....`

internal/services/shared/errors.go
josephschorr: Add a TODO here

internal/services/shared/errors.go
```diff
@@ -47,6 +47,54 @@ func NewSchemaWriteDataValidationError(message string, args ...any) SchemaWriteD
...
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_SCHEMA_TYPE_ERROR,
+			map[string]string{},
+		),
+	)
+}
+
+func NewDeprecationError(namespace, relation, comments string) DeprecationError {
+	switch {
+	case relation == "*":
```
josephschorr: use the string method in the type system to describe the relation: you need to handle wildcards, expiration and caveats, and all should be done in there

internal/relationships/validation_test.go
```diff
@@ -306,6 +306,63 @@ func TestValidateRelationshipOperations(t *testing.T) {
...
+				relation viewer: user | @deprecated(warn, "comments") testuser
+				relation auditor: user | @deprecated(error, "test") testuser
+			}`,
+			"platform:foo#auditor@testuser:test",
+			core.RelationTupleUpdate_CREATE,
+			"the object type testuser is deprecated: test",
+		},
+		{
+			"deprecated relation same subject type with wildcard",
+			`use deprecation
```
josephschorr, line 354: add tests for deprecation of, say, a subject type without expiration but the one *with* expiration is still valid (and vice versa) and caveats as well

pkg/schema/typesystem.go
```diff
@@ -38,6 +38,89 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
 	return v, err
 }
 
+// GetValidatedDefinition looks up and returns a definition struct, if it has been validated.
+func (ts *TypeSystem) GetDeprecationForNamespace(ctx context.Context, definition string) (*core.Deprecation, bool, error) {
```
josephschorr: Add these methods to the type system tests

pkg/schemadsl/generator/generator.go
```diff
@@ -202,6 +214,18 @@ func (sg *sourceGenerator) emitNamespace(namespace *core.NamespaceDefinition) er
 	sg.markNewScope()
 
 	for _, relation := range namespace.Relation {
+		if relation.Deprecation != nil && relation.Deprecation.DeprecationType != core.DeprecationType_DEPRECATED_TYPE_UNSPECIFIED {
```
josephschorr: extra into a helper method

proto/internal/core/v1/core.proto
```diff
@@ -420,6 +447,11 @@ message AllowedRelation {
    * required_expiration defines the required expiration on this relation.
    */
   ExpirationTrait required_expiration = 7;
+
+  /**
+   * deprecation contains the required deprecation for the relation.
+   */
+  Deprecation required_deprecation = 8;
```
josephschorr: This should not be marked as `required`. The required above means that a caveat (or expiration) must be specified when using the type as part of a subject, which doesn't apply here

internal/relationships/validation.go
kartikaysaxena: but a single relationship might have multiple deprecations (a warning type deprecation at the resource or relation level, and an error type deprecation at the subject level) and it might be necessary to preserve all of them

proto/internal/core/v1/core.proto
kartikaysaxena: renamed

pkg/schema/typesystem.go
kartikaysaxena: done

internal/relationships/validation_test.go
kartikaysaxena: added tests of these scenarios including the log and err messages, example "resource_type user with caveat `somecaveat` and expiration is deprecated:comments here"

pkg/schema/typesystem.go
```diff
@@ -38,6 +41,178 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
...
+			dep.DeprecationType != core.DeprecationType_DEPRECATED_TYPE_UNSPECIFIED {
+			return dep, true, nil
+		}
+	}
+	return nil, false, nil
+}
+
+// CheckRelationshipDeprecation performs a check over the deprecated relationship's resource, relation, subject and allowed relation, if any
+func (ts *TypeSystem) CheckRelationshipDeprecation(ctx context.Context, relationship tuple.Relationship) error {
+	// Validate if the resource relation is deprecated
```
kartikaysaxena, line 148: this method is now invoked during validation of the relationship being written, and is responsible for checking any applicable deprecations on the resource, relation, or subject

pkg/schema/typesystem.go
```diff
@@ -38,6 +41,178 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
...
+func (ts *TypeSystem) GetDeprecationForAllowedRelation(
+	ctx context.Context,
+	resourceNamespace string,
+	resourceRelation string,
+	subjectNamespace string,
+	subjectRelation string,
+	isWildcard bool,
+	hasCaveat bool,
+	hasExpiration bool,
+) (*core.Deprecation, bool, error) {
```
kartikaysaxena: this now checks whether the allowed relation reference has a caveat/expiration or both

internal/relationships/validation.go
kartikaysaxena: these methods would now be invoked in `CheckRelationshipDeprecation`, does this sound about right?

internal/relationships/validation.go
```diff
@@ -84,6 +98,13 @@ func ValidateRelationshipsForCreateOrTouch(
 		}
 	}
 
+	ts := schema.NewTypeSystem(schema.ResolverForDatastoreReader(reader))
```
josephschorr: let's not reconstruct the type system, but rather pass it in

internal/relationships/validation.go
```diff
@@ -277,3 +298,23 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
 		relationship.OptionalCaveat.Context != nil &&
 		len(relationship.OptionalCaveat.Context.GetFields()) > 0
 }
+
+func CheckDeprecationsOnRelationships(
```
josephschorr, line 301: doc comment

internal/relationships/validation_test.go
```diff
@@ -306,6 +306,200 @@ func TestValidateRelationshipOperations(t *testing.T) {
...
+			definition user {}
+			definition testuser {}
+
+			definition platform {
+				relation viewer: user | @deprecated(warn, "comments") testuser
+				relation auditor: testuser | @deprecated(error, "no wildcard please") testuser:*
+			}`,
+			"platform:foo#auditor@testuser:*",
+			core.RelationTupleUpdate_CREATE,
+			"wildcard allowed type testuser:* is deprecated: no wildcard please",
```
josephschorr, line 364: add the same test with writing the non-wildcard and ensure it does not return an error/warning

internal/services/shared/errors.go
```diff
@@ -47,6 +47,29 @@ func NewSchemaWriteDataValidationError(message string, args ...any) SchemaWriteD
...
+		codes.Aborted,
+		spiceerrors.ForReason(
+			// TODO: replace with a deprecation type error reason
+			v1.ErrorReason_ERROR_REASON_SCHEMA_TYPE_ERROR,
+			map[string]string{},
+		),
+	)
+}
+
+func NewDeprecationError(err error) DeprecationError {
```
josephschorr, line 68: doc comment

pkg/namespace/builder.go
```diff
@@ -115,6 +128,65 @@ func AllowedRelationWithExpiration(namespaceName string, relationName string) *c
...
+		RelationOrWildcard: &core.AllowedRelation_Relation{
+			Relation: relationName,
+		},
+		RequiredExpiration: &core.ExpirationTrait{},
+		Deprecation:        deprecation,
+	}
+}
+
+// AllowedRelationWithDeprecationCaveatAndExpiration creates a relation reference to an allowed relation.
+func AllowedRelationWithDeprecationCaveatAndExpiration(namespaceName string, relationName string, withCaveat *core.AllowedCaveat, deprecation *core.Deprecation) *core.AllowedRelation {
```
josephschorr: might be easier to add a `WithDeprecation` that takes in an AllowedRelation, adds it, and returns a new one

pkg/schema/typesystem.go
```diff
@@ -38,6 +41,178 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
...
+	ns, _, err := ts.resolver.LookupDefinition(ctx, definition)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if ns.Deprecation != nil && ns.Deprecation.DeprecationType != core.DeprecationType_DEPRECATED_TYPE_UNSPECIFIED {
+		return ns.Deprecation, true, nil
+	}
+
+	return &core.Deprecation{}, false, nil
```
josephschorr: return `nil` if it isn't present

pkg/schema/typesystem.go
josephschorr: Yeah

pkg/schema/typesystem.go
```diff
@@ -64,3 +239,42 @@ func (ts *TypeSystem) getDefinition(ctx context.Context, definition string) (*De
 	}
 	return d, prevalidated, nil
 }
+
+func handleDeprecation(objectType, relation string, dep *core.Deprecation, logMessage string, caveatName string, hasExpiration bool) error {
```
josephschorr: more descriptive name than `handle...`

pkg/schema/typesystem.go
```diff
@@ -64,3 +239,42 @@ func (ts *TypeSystem) getDefinition(ctx context.Context, definition string) (*De
...
+	extra := ""
+	if caveatName != "" {
+		extra += fmt.Sprintf(" with caveat `%s`", caveatName)
```
josephschorr: switch from fmt.Sprintf to string concat

pkg/schema/typesystem_test.go
```diff
@@ -50,3 +50,338 @@ func TestTypeSystemConcurrency(t *testing.T) {
...
+			args: args{
+				resourceNamespace: "document",
+				resourceRelation:  "viewer",
+				subjectNamespace:  "user",
+				subjectRelation:   "*",
+				isWildcard:        true,
+				hasCaveat:         false,
+				hasExpiration:     false,
+			},
+			expectedDeprecation: ns.Deprecation(core.DeprecationType_DEPRECATED_TYPE_WARNING, "wildcard"),
```
josephschorr, line 1030: make sure to add the get deprecation accessors to the existing type system test accessor suite: https://github.com/authzed/spicedb/blob/main/pkg/schema/definition_test.go#L506

pkg/schemadsl/compiler/translator.go
```diff
@@ -258,7 +285,12 @@ func translateObjectDefinition(tctx *translationContext, defNode *dslNode) (*cor
 		return ns, nil
 	}
 
-	ns := namespace.Namespace(nspath, relationsAndPermissions...)
+	var ns *core.NamespaceDefinition
+	if slices.Contains(tctx.enabledFlags, "deprecation") && slices.Contains(tctx.allowedFlags, "deprecation") && (deprecationForNamespace != &core.Deprecation{}) {
```
josephschorr: if deprecation is not enabled, we should make sure to raise an error during parsing

pkg/schemadsl/compiler/translator.go
```diff
@@ -713,6 +793,17 @@ func translateSpecificTypeReference(tctx *translationContext, typeRefNode *dslNo
 		ref.RequiredExpiration = &core.ExpirationTrait{}
 	}
 
+	// Add the deprecation, if any.
+	if slices.Contains(tctx.enabledFlags, "deprecation") && slices.Contains(tctx.allowedFlags, "deprecation") {
```
josephschorr: move this check into translateDeprecation and have it return an error if not enabled but specified

internal/relationships/validation.go
kartikaysaxena: I initially tried to pass reader to the `CheckDeprecationsOnRelationships` function, then construct a type system there so that it would be centralised for other calls, but this would trigger the hedgingProxy to panic at a point where `HeadRevision()` is called twice, so circled back to this. (Neither `ValidateRelationshipsForCreateOrTouch` nor `ValidateRelationships` gets an argument of type system from upstream)

kartikaysaxena: changed a bit

pkg/schema/typesystem_test.go
kartikaysaxena: done

kartikaysaxena:
Rebased and updated

internal/relationships/validation.go
```diff
@@ -52,6 +54,17 @@ func ValidateRelationshipUpdates(
 		}
 	}
 
+	// check if the relation is deprecated for create or touch operations
+	var relsToCheck []tuple.Relationship
```
josephschorr: `relsToCheck := make([]tuple.Relationship, 0, len(updates))`

internal/relationships/validation.go
```diff
@@ -277,3 +296,24 @@ func hasNonEmptyCaveatContext(relationship tuple.Relationship) bool {
 		relationship.OptionalCaveat.Context != nil &&
 		len(relationship.OptionalCaveat.Context.GetFields()) > 0
 }
+
+// CheckDeprecationsOnRelationships checks the provided relationships for any deprecations.
```
josephschorr: // CheckDeprecationsOnRelationships checks the provided relationships for any deprecations, returning an error if applicable

internal/services/shared/errors.go
josephschorr: `object type, relation, or allowed subject type`

josephschorr: Also: can we add a test such that if an object type is deprecated, and you attempt to write with it as a *subject*, it also fails?

pkg/schema/definition_test.go
```diff
@@ -1041,6 +1041,151 @@ func TestTypeSystemAccessors(t *testing.T) {
 						_, err = vts.PossibleTraitsForSubject("unknown", "user")
```
josephschorr, line 1244: add calls to the deprecation methods in the other tests here, and make sure they all return false/nil/empty

pkg/schema/typesystem.go
josephschorr: Might be worth changing this to take in a traits map

pkg/schema/typesystem.go
```diff
@@ -38,6 +41,178 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
...
+
+		switch w := allowed.RelationOrWildcard.(type) {
+		case *core.AllowedRelation_PublicWildcard_:
+			if !isWildcard {
+				continue
+			}
+		case *core.AllowedRelation_Relation:
+			if isWildcard || w.Relation != subjectRelation {
+				continue
+			}
```
josephschorr, line 133: add a default here that returns a spiceerrors

pkg/schemadsl/compiler/translator.go
```diff
@@ -245,7 +267,12 @@ func translateObjectDefinition(tctx *translationContext, defNode *dslNode) (*cor
 	}
 
 	if len(relationsAndPermissions) == 0 {
-		ns := namespace.Namespace(nspath)
+		var ns *core.NamespaceDefinition
+		if slices.Contains(tctx.enabledFlags, "deprecation") && slices.Contains(tctx.allowedFlags, "deprecation") && (deprecationForNamespace != &core.Deprecation{}) {
```
josephschorr: move `if slices.Contains(tctx.enabledFlags, "deprecation") && slices.Contains(tctx.allowedFlags, "deprecation") && (deprecationForNamespace != &core.Deprecation{})` into a helper method

pkg/schema/typesystem.go
```diff
@@ -38,6 +41,178 @@ func (ts *TypeSystem) GetDefinition(ctx context.Context, definition string) (*De
...
+		if dep := allowed.Deprecation; dep != nil &&
+			dep.DeprecationType != core.DeprecationType_DEPRECATED_TYPE_UNSPECIFIED {
+			return dep, true, nil
+		}
+	}
+	return nil, false, nil
+}
+
+// CheckRelationshipDeprecation performs a check over the deprecated relationship's resource, relation, subject and allowed relation, if any
+func (ts *TypeSystem) CheckRelationshipDeprecation(ctx context.Context, relationship tuple.Relationship) error {
```
barakmich, line 147: A related call to this might also be useful in Validate() (in this package) if the deprecation is an error state.

kartikaysaxena: The `Validate` function is also invoked during `WriteSchema`, so if we call this in `Validate`, any attempt to write a schema with this feature containing deprecated resources/relations/subjects would result in an error at schema write time.

internal/relationships/validation.go
kartikaysaxena: yeah would improve perf, done

pkg/schema/definition_test.go
kartikaysaxena: done

internal/services/shared/errors.go
kartikaysaxena: i think that is already there, something like [this](https://github.com/authzed/spicedb/pull/2504/files#diff-deb1b83fcd53d603a546c08c955c34acd97440279d9235e03305ccd960e616fdR324-R337)?

kartikaysaxena:
checking on this

josephschorr:
@kartikaysaxena Sorry, I thought this merged. Can you rebase and resolve conflicts?

josephschorr:
@kartikaysaxena Checking in on this; I'd like to get it in but it has conflicts

kartikaysaxena:
Hey @josephschorr sorry was busy at work; I have rebased and resolved the the merge conflicts and did a sanity on the changes. Let me know if anything else is needed, happy to help.

josephschorr:
Hi @kartikaysaxena mind rebasing once more? This got lost in the shuffle

PR #2506, opened 2025-07-22
Populate empty caveat_name and caveat_context in MySQL with default values

### LINKED ISSUE #2500: SQL Scan Error on `caveat_context` on MySQL 8.0.40

linux

arm64

v1.45.1

### Steps to Reproduce

The error occurs during a `ReadRelationshipsRequest` grpc call.

### Expected Result

The ReadRelationshipsRequest call succeeds.

### Actual Result

```
rpc error: code = Unknown desc = error when reading tuples: unable to query relationships: scan err: sql: Scan error on column index 3, name \"caveat_context\": unsupported type: []uint8
```

mazdakb:
fixes #2500 

Introduce a new data migration for MySQL that updates rows with `NULL` values as `caveat_name` or `caveat_context` columns in `relation_tuple` table, replacing them with appropriate default values.

### DIFF
--- a/internal/datastore/mysql/readwrite.go
+++ b/internal/datastore/mysql/readwrite.go
@@ -56,13 +56,19 @@ type mysqlReadWriteTXN struct {
 type structpbWrapper map[string]any
 
 func (cc *structpbWrapper) Scan(val any) error {
+	if val == nil {
+		clear(*cc)
+		*cc = nil
+		return nil
+	}
+
 	v, ok := val.([]byte)
 	if !ok {
 		return fmt.Errorf("unsupported type: %T", v)
 	}
 
 	clear(*cc)
-	return json.Unmarshal(v, &cc)
+	return json.Unmarshal(v, cc)
 }
 
 func (cc *structpbWrapper) Value() (driver.Value, error) {
--- a/internal/datastore/mysql/readwrite_test.go
+++ b/internal/datastore/mysql/readwrite_test.go
@@ -1,0 +1,54 @@
+package mysql
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestStructpbWrapperScan(t *testing.T) {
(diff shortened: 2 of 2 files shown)


vroldanbet (commented):
I'm hesitant to add a new data migration because this could cause an incident in large mysql deployments.

Mind decribing what problem is this trying to solve? Are there alternatives that do not involve a migration?

mazdakb:
Data inserted in MySQL's `relation_tuple` table prior to the introduction of `internal/datastore/mysql/migrations/zz_migration.0004_add_caveat.go` has the `caveat_name` and `caveat_context` columns with SQL `NULL` values. In newer SpiceDB versions, `''` and JSON `'null'` literal are used respectively for `caveat_name` and `caveat_context` as null values. 
Here's an example I could reproduce:

- What the user has in their `relation_tuple` table after running the above migration (the last three colums are `caveat_name`, `caveat_context` and `expiration`:
```mysql
(1,'resource','someresource','viewer','user','sarah','...',3,9223372036854775807,NULL,NULL,NULL)
```
- And here's the exact same schema and set of relationships inserted via v1.45.1:
```mysql
(1,'resource','someresource','viewer','user','sarah','...',3,9223372036854775807,'','null',NULL)
```

The SQL null values in older rows causes an error in `ReadRelationships` requests for those rows because the SQL `NULL` cannon be unmarshalled automatically into Go's `nil` for the JSON column: `internal/datastore/mysql/readwrite.go:63`

There are two ways to address this issue:

- Modify the `Scan` function in `internal/datastore/mysql/readwrite.go` and add an additional nil value check.
- Add a MySQL migration to normalize data for all rows that may have been inserted into `relation_tuple` prior to introduction of the above migration.

Here's a relevant thread in MySQL dev forums: https://bugs.mysql.com/bug.php?id=85755

vroldanbet (approved):

PR #2510, opened 2025-07-28
fix: in Watch server, respect config flag StreamingAPITimeout

miparnisari:
## Description

I tested `zed watch` on a server that had no changes going in, and noticed that setting flag `STREAMING_API_RESPONSE_DELAY_TIMEOUT=(averybignumber)` had no effect whatsoever.

An option to this PR would be to clarify in the docs that this flag doesn't affect the Watch API 🤷‍♀️ 

## Testing

**Before:**

```shell
docker run --rm -p 50051:50051 authzed/spicedb:latest serve --grpc-preshared-key=random --streaming-api-response-delay-timeout=1s

zed ctx set local localhost:50051 random
zed watch --insecure

// no errors received
```

**Now:**

```shell
docker build . -t spicedblocal
docker run --rm -p 50051:50051 spicedblocal serve --grpc-preshared-key=random --streaming-api-response-delay-timeout=1s

zed ctx set local localhost:50051 random
zed watch --insecure

//  ERR terminated with errors error="rpc error: code = Canceled desc = watch canceled by user: watch was canceled by the caller"
```

and the server logs this:

```
{"level":"warn","time":"2025-07-28T20:24:58Z","message":"operation took longer than allowed 1s to complete"}
{"level":"info","protocol":"grpc","grpc.component":"server","grpc.service":"authzed.api.v1.WatchService","grpc.method":"Watch","grpc.method_type":"server_stream","requestID":"d23tp6a18gnc73f98fb0","peer.address":"130.211.126.102:53587","grpc.start_time":"2025-07-28T20:24:57Z","grpc.code":"Canceled","grpc.error":"rpc error: code = Canceled desc = watch canceled by user: watch was canceled by the caller","grpc.time_ms":1006,"time":"2025-07-28T20:24:58Z","message":"finished call"}
```

### DIFF
--- a/internal/middleware/streamtimeout/streamtimeout.go
+++ b/internal/middleware/streamtimeout/streamtimeout.go
@@ -5,11 +5,9 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/rs/zerolog/log"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
-
-	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
 // MustStreamServerInterceptor returns a new stream server interceptor that cancels the context
@@ -23,7 +21,9 @@ func MustStreamServerInterceptor(timeout time.Duration) grpc.StreamServerInterce
 		ctx := stream.Context()
 		withCancel, internalCancelFn := context.WithCancelCause(ctx)
 		timer := time.AfterFunc(timeout, func() {
-			internalCancelFn(spiceerrors.WithCodeAndDetailsAsError(fmt.Errorf("operation took longer than allowed %v to complete", timeout), codes.DeadlineExceeded))
+			err := fmt.Errorf("operation took longer than allowed %v to complete", timeout)
+			log.Warn().Msg(err.Error())
+			internalCancelFn(err)
 		})
 		wrapper := &sendWrapper{stream, withCancel, timer, timeout}
 		return handler(srv, wrapper)
--- a/internal/services/v1/watch.go
+++ b/internal/services/v1/watch.go
@@ -12,7 +12,9 @@ import (
 
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 
+	"github.com/authzed/spicedb/internal/middleware"
 	datastoremw "github.com/authzed/spicedb/internal/middleware/datastore"
+	"github.com/authzed/spicedb/internal/middleware/streamtimeout"
 	"github.com/authzed/spicedb/internal/middleware/usagemetrics"
 	"github.com/authzed/spicedb/internal/services/shared"
 	"github.com/authzed/spicedb/pkg/datastore"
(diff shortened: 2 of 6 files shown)


internal/services/v1/watch.go
```diff
@@ -26,16 +28,31 @@ type watchServer struct {
 	v1.UnimplementedWatchServiceServer
 	shared.WithStreamServiceSpecificInterceptor
 
-	heartbeatDuration time.Duration
+	config WatchServerConfig
+}
+
+type WatchServerConfig struct {
```
josephschorr, line 35: doc comments

internal/services/v1/watch.go
```diff
@@ -26,16 +28,31 @@ type watchServer struct {
...
+	// recently received.
+	StreamingAPITimeout time.Duration
 }
 
 // NewWatchServer creates an instance of the watch server.
-func NewWatchServer(heartbeatDuration time.Duration) v1.WatchServiceServer {
+func NewWatchServer(config WatchServerConfig) v1.WatchServiceServer {
+	configWithDefaults := WatchServerConfig{
+		HeartbeatDuration:   config.HeartbeatDuration,
+		StreamingAPITimeout: defaultIfZero(config.StreamingAPITimeout, 30*time.Second),
```
josephschorr: let's see a much higher default for watch timeout; perhaps an hour?

internal/services/v1/watch.go
miparnisari: added

internal/services/v1/watch.go
miparnisari: i'd love to, but the problem with that is that the users may be misled by our docs.

according to ` go run ./cmd/spicedb/main.go serve --help`:

```
 --streaming-api-response-delay-timeout duration           
maximum time until the server will return an error to a caller of a streaming API because no data has been sent (default 30s)
```

(see the (default 30s) at the end)
so unless we split this flag into two, or use the same default everywhere, i'd keep this as is.

internal/middleware/streamtimeout/streamtimeout.go
```diff
@@ -23,7 +21,9 @@ func MustStreamServerInterceptor(timeout time.Duration) grpc.StreamServerInterce
 		ctx := stream.Context()
 		withCancel, internalCancelFn := context.WithCancelCause(ctx)
 		timer := time.AfterFunc(timeout, func() {
-			internalCancelFn(spiceerrors.WithCodeAndDetailsAsError(fmt.Errorf("operation took longer than allowed %v to complete", timeout), codes.DeadlineExceeded))
```
miparnisari, line 26: this error isn't shown to the caller

pkg/cmd/serve.go
```diff
@@ -103,7 +103,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	apiFlags.Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	apiFlags.IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes. A value of zero or less means no limit")
 	apiFlags.IntVar(&config.MaxRelationshipContextSize, "max-relationship-context-size", 25000, "maximum allowed size of the context to be stored in a relationship")
-	apiFlags.DurationVar(&config.StreamingAPITimeout, "streaming-api-response-delay-timeout", 30*time.Second, "maximum duration time elapsed between messages sent by the server-side to the client (responses) before the stream times out")
+	apiFlags.DurationVar(&config.StreamingAPITimeout, "streaming-api-response-delay-timeout", 30*time.Second, "maximum time until the server will return an error to a caller of a streaming API because no data has been sent")
```
miparnisari, line 106: clarified this - the "no data has been sent" is important

internal/services/v1/watch.go
josephschorr: I'd say we split the flag into two and have one for `watch`

miparnisari: I'm not into that idea because the `serve` command already has a lot of flags https://authzed.com/docs/spicedb/concepts/commands#reference-spicedb-serve

we should have a very good reason to add another one.

why do we need this flag in the first place? could we not replace it with guidance that if deploying spicedb in prod, it should be behind a load balancer that puts timeouts in place?

josephschorr: Because watch is intended to be a very long running process and other streams are not; a LookupResources that takes over 45 seconds to return is a problem. A watch that closes before minutes or even hours is up, is unexpected.

miparnisari: Okay, on second thought, we don't need this PR at all 😆 i can just amend the description of the flag to mention the ones it affects

PR #2514, opened 2025-07-30
update goreleaser in nightly build

vroldanbet:
we had success with 2.5.1, so I assume whatever
regression happened in 2.3.0 seems to be fixed now


josephschorr (approved):
LGTM

PR #2519, opened 2025-08-05
fix: description of Stream Timeout flag

miparnisari:
## Description

Stemming from https://github.com/authzed/spicedb/pull/2510

## Testing

```
 $ go run ./cmd/spicedb/main.go serve --help

...
      --streaming-api-response-delay-timeout duration            maximum time that streaming APIs (LookupSubjects, LookupResources, ReadRelationships and ExportBulkRelationships) can be allowed to run but no response be sent to the client before the stream times out (default 30s)
...
```


josephschorr (approved):
LGTM

PR #2520, opened 2025-08-07
Ensure that the proper indexes are used for schema diff operations

josephschorr:
Some indexes were missing in the CRDB indexing map, and some of the queries were not being fully specified, leading to bad index selection and, therefore, a negative performance impact


pkg/datastore/queryshape/queryshape.go
```diff
@@ -98,3 +95,15 @@ const (
 	// ✅ resource_type, *️⃣ resource_id, ✅ resource_relation, ✅ subject_type, 🅿️ subject_id, ✅ subject_relation, *️⃣ caveat, *️⃣ expiration
```
miparnisari, line 95: I'm unclear on the difference between *️⃣ (optional) and 🅿️ (possibly specified). Those are synonyms.

it seems that *️⃣  actually means `not specified`

internal/services/steelthreadtesting/operations.go
```diff
@@ -387,12 +387,31 @@ func bulkCheckPermissions(parameters map[string]any, client v1.PermissionsServic
...
+
+	schemaText := parameters["schema"].(string)
+
+	_, err := clients.SchemaClient.WriteSchema(ctx, &v1.WriteSchemaRequest{
+		Schema: schemaText,
+	})
+	if err != nil {
+		return map[string]any{
+			"error": fmt.Sprintf("failed to write schema: %v", err),
+		}, nil
```
miparnisari, line 402:
```suggestion
		}, err
```

internal/datastore/crdb/schema/indexes.go
```diff
@@ -87,6 +87,15 @@ func IndexingHintForQueryShape(schema common.SchemaInformation, qs queryshape.Sh
 	case queryshape.FindResourceOfType:
 		return forcedIndex{IndexPrimaryKey}
 
+	case queryshape.FindResourceAndSubjectWithRelations:
```
miparnisari, line 90: FYI, not for this PR. There's the `exhaustive` linter that warns if you if the switch doesn't cover all cases. https://golangci-lint.run/usage/linters/#exhaustive

With the right settings it could have found this issue :) (although now, you added a test, so 🤷‍♀️ )

image

pkg/datastore/queryshape/queryshape.go
```diff
@@ -67,15 +67,12 @@ const (
...
+	// given type and relation.
 	//
-	// The query shape selects a resource of a given type and relation, which are
-	// specified by providing the resource type and relation. The other fields are never
-	// specified.
-	//
-	// ✅ resource_type, *️⃣ resource_id, ✅ resource_relation, *️⃣ subject_type, *️⃣ subject_id, *️⃣ subject_relation, *️⃣ caveat, *️⃣ expiration
-	FindResourceOfTypeAndRelation = "find-resource-of-type-and-relation"
+	// ✅ resource_type, *️⃣ resource_id, ✅ resource_relation, ✅ subject_type, *️⃣ subject_id, ✅ subject_relation, *️⃣ caveat, *️⃣ expiration
+	FindResourceAndSubjectWithRelations = "find-resource-and-subject-with-relations"
```
miparnisari:
```suggestion
	FindResourceAndSubjectWithRelations Shape = "find-resource-and-subject-with-relations"
```

here and in all the others. Otherwise the `exhaustive` linter doesn't work

internal/datastore/proxy/indexcheck/queryshapevalidators.go
```diff
@@ -198,6 +214,81 @@ func validateQueryShape(queryShape queryshape.Shape, filter datastore.Relationsh
 		}
 		return nil
 
+	case queryshape.FindResourceAndSubjectWithRelations:
```
miparnisari, line 217: Tests??

internal/services/shared/schema.go
```diff
@@ -435,6 +419,14 @@ func sanityCheckNamespaceChanges(
 	return diff, nil
 }
 
+func subjectRelationForAllowedType(allowedType *core.AllowedRelation) string {
+	// NOTE: wildcards are always stored as `...` in the relationship, so we return that here.
+	if allowedType.GetPublicWildcard() != nil {
+		return tuple.Ellipsis
+	}
+	return allowedType.GetRelation()
+}
```
miparnisari:
```suggestion
func subjectRelationFilterForAllowedType(allowedType *core.AllowedRelation) datastore.SubjectRelationFilter {
	rel := allowedType.GetRelation()
	if allowedType.GetPublicWildcard() != nil {
		// NOTE: wildcards are always stored as `...` in the relationship
		rel = tuple.Ellipsis
	}
	return datastore.SubjectRelationFilter{}.WithRelation(rel)
}
```

pkg/datastore/queryshape/queryshape.go
josephschorr: You're correct; fixed

internal/services/steelthreadtesting/operations.go
josephschorr: This was done on purpose to allow the test to check for expected errors

internal/datastore/crdb/schema/indexes.go
josephschorr: That's a good idea; let's add it as a followup

pkg/datastore/queryshape/queryshape.go
josephschorr: Really?! That's annoying...

internal/datastore/proxy/indexcheck/queryshapevalidators.go
josephschorr: This actually is the test :)

The validator runs to ensure query shapes match their associated queries

josephschorr: Tests added

PR #2526, opened 2025-08-08
Fix missing rows.Err check that could cause too-large writes to silently fail in PG

josephschorr:
This PR improves error handling robustness in the Postgres datastore by adding comprehensive checks for missing `rows.Err()` calls and fixing identified issues.

Changes:
- Added static analysis linter: Introduces a new `iferrafterrowclosecheck` analyzer that detects missing `rows.Err()` checks after `rows.Close()` calls. This helps prevent silent failures when receiving an error after iterating over SQL result sets.
- Fixed missing error checks: Added proper error handling in the Postgres readwrite implementation to check `rows.Err()` after row iteration, ensuring  such database errors aren't silently ignored.
- Enhanced error handling: Added a user-friendly error message if the size of the INSERT query exceeds the bounds allowed by PG

### DIFF
--- a/internal/datastore/common/errors.go
+++ b/internal/datastore/common/errors.go
@@ -43,6 +43,30 @@ func NewSerializationError(err error) error {
 	return SerializationError{err}
 }
 
+type WriteOverLimitError struct {
+	error
+}
+
+func (err WriteOverLimitError) GRPCStatus() *status.Status {
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.Aborted,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_TOO_MANY_UPDATES_IN_REQUEST,
+			map[string]string{},
+		),
+	)
+}
+
+func (err WriteOverLimitError) Unwrap() error {
+	return err.error
+}
+
+// NewWriteOverLimitError creates a new WriteOverLimitError.
+func NewWriteOverLimitError(err error) error {
+	return WriteOverLimitError{err}
+}
+
 // ReadOnlyTransactionError is returned when an otherwise read-write
 // transaction fails on writes with an error indicating that the datastore
 // is currently in a read-only mode.
--- a/internal/datastore/mysql/watch.go
+++ b/internal/datastore/mysql/watch.go
@@ -182,7 +182,7 @@ func (mds *Datastore) loadChanges(
 		}
 	}
 	rows.Close()
-	if err = rows.Err(); err != nil {
+	if rows.Err() != nil {
 		return
 	}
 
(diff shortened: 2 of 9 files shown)


josephschorr:
No idea

miparnisari:
@vroldanbet @josephschorr pretty sure that's because in https://github.com/authzed/spicedb/blob/main/.golangci.yaml#L49-L52 we missed `github.com/jackc/pgx/v5`

if i add that, i see

```
INFO [runner] linters took 1.364939417s with stages: goanalysis_metalinter: 1.346029375s 
internal/datastore/postgres/common/interceptor.go:62:37: rows.Err must be checked (rowserrcheck)
        return q.interceptor.InterceptQuery(ctx, q.delegate, sql, arguments...)
                                           ^
internal/datastore/postgres/common/interceptor.go:122:37: rows.Err must be checked (rowserrcheck)
        return t.interceptor.InterceptQuery(ctx, t.delegate, sql, args...)
                                           ^
internal/datastore/postgres/common/interceptor.go:158:36: rows.Err must be checked (rowserrcheck)
        return i.interceptingQuerier.Query(ctx, sql, optionsAndArgs...)
                                          ^
3 issues:
* rowserrcheck: 3
```

vroldanbet:
@miparnisari does it require specifying the packages, or does it have a default value we can rely on?

josephschorr:
> @vroldanbet @josephschorr pretty sure that's because in https://github.com/authzed/spicedb/blob/main/.golangci.yaml#L49-L52 we missed `github.com/jackc/pgx/v5`
> 
> if i add that, i see
> 
> ```
> INFO [runner] linters took 1.364939417s with stages: goanalysis_metalinter: 1.346029375s 
> internal/datastore/postgres/common/interceptor.go:62:37: rows.Err must be checked (rowserrcheck)
>         return q.interceptor.InterceptQuery(ctx, q.delegate, sql, arguments...)
>                                            ^
> internal/datastore/postgres/common/interceptor.go:122:37: rows.Err must be checked (rowserrcheck)
>         return t.interceptor.InterceptQuery(ctx, t.delegate, sql, args...)
>                                            ^
> internal/datastore/postgres/common/interceptor.go:158:36: rows.Err must be checked (rowserrcheck)
>         return i.interceptingQuerier.Query(ctx, sql, optionsAndArgs...)
>                                           ^
> 3 issues:
> * rowserrcheck: 3
> ```
> 
> so we probably missed a couple more places :/

This seems... wrong. `InterceptQuery` does not need to close the rows because the caller will.

miparnisari:
> @miparnisari does it require specifying the packages, or does it have a default value we can rely on?

no, the default is [] according to https://golangci-lint.run/docs/linters/configuration/#rowserrcheck

> InterceptQuery does not need to close the rows because the caller will.

That's a known issue: https://github.com/jingyugao/rowserrcheck/issues/32. We can use `// nolint` on those

ecordell:
shouldn't this check the type of the variable to see if it's a sql.Rows and not try to guess based on the name of the variable?

josephschorr:
> shouldn't this check the type of the variable to see if it's a sql.Rows and not try to guess based on the name of the variable?

Doing that in a linter is non-trivial, especially with inferred types that go supports

PR #2531, opened 2025-08-14
Implement two new packages: `query` and `schema/v2`

barakmich:

### DIFF
--- a/pkg/query/arrow_test.go
+++ b/pkg/query/arrow_test.go
@@ -4,6 +4,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/pkg/tuple"
 )
 
 func TestArrowIterator(t *testing.T) {
@@ -15,8 +17,8 @@ func TestArrowIterator(t *testing.T) {
 	// Left side: document parent relationships to folders
 	leftRels := NewFolderHierarchyFixedIterator()
 
-	// Right side: folder viewer relationships
-	rightRels := NewDocumentAccessFixedIterator()
+	// Right side: folder viewer relationships (same as left for folder access)
+	rightRels := NewFolderHierarchyFixedIterator()
 
 	arrow := NewArrow(leftRels, rightRels)
 
@@ -28,9 +30,20 @@ func TestArrowIterator(t *testing.T) {
 		relSeq, err := arrow.Check(nil, []string{"spec1", "spec2"}, "alice")
 		require.NoError(err)
 
-		_, err = CollectAll(relSeq)
+		rels, err := CollectAll(relSeq)
 		require.NoError(err)
-		// Results depend on the specific relations in our test data
+
+		// Expected: spec1 should match because alice has viewer access to project1 (spec1's parent)
+		// spec2 should NOT match because alice does not have access to project2 (spec2's parent)
+		expected := []tuple.Relationship{
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "spec1", "parent"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+		}
+		require.Equal(expected, rels)
 	})
 
 	t.Run("Check_EmptyResources", func(t *testing.T) {
@@ -166,11 +179,13 @@ func TestArrowIteratorMultipleResources(t *testing.T) {
 	// - nonexistent doesn't exist -> should NOT match
 
 	// We expect exactly 1 result: spec1 with alice as subject
-	require.Len(rels, 1, "should find exactly one arrow relationship")
-	require.Equal("document", rels[0].Resource.ObjectType)
-	require.Equal("spec1", rels[0].Resource.ObjectID)
-	require.Equal("parent", rels[0].Resource.Relation)
-	require.Equal("user", rels[0].Subject.ObjectType)
-	require.Equal("alice", rels[0].Subject.ObjectID)
-	require.Equal("...", rels[0].Subject.Relation)
+	expected := []tuple.Relationship{
+		{
+			RelationshipReference: tuple.RelationshipReference{
+				Resource: tuple.ONR("document", "spec1", "parent"),
+				Subject:  tuple.ONR("user", "alice", "..."),
+			},
+		},
+	}
+	require.Equal(expected, rels)
 }
--- a/pkg/query/build_tree_test.go
+++ b/pkg/query/build_tree_test.go
@@ -39,9 +39,8 @@ func TestBuildTree(t *testing.T) {
 	relSeq, err := it.Check(ctx, []string{"specialplan"}, "multiroleguy")
 	require.NoError(err)
 
-	rels, err := CollectAll(relSeq)
+	_, err = CollectAll(relSeq)
 	require.NoError(err)
-	t.Log(rels)
 }
 
 func TestBuildTreeMultipleRelations(t *testing.T) {
--- a/pkg/query/quick_e2e_test.go
+++ b/pkg/query/quick_e2e_test.go
@@ -45,9 +45,8 @@ func TestCheck(t *testing.T) {
 	relSeq, err := it.Check(ctx, []string{"specialplan"}, "multiroleguy")
 	require.NoError(err)
 
-	rels, err := CollectAll(relSeq)
+	_, err = CollectAll(relSeq)
 	require.NoError(err)
-	t.Log(rels)
 }
 
 func TestBaseIterSubjects(t *testing.T) {
@@ -75,9 +74,8 @@ func TestBaseIterSubjects(t *testing.T) {
 	relSeq, err := vande.IterSubjects(ctx, "specialplan")
 	require.NoError(err)
 
-	rels, err := CollectAll(relSeq)
+	_, err = CollectAll(relSeq)
 	require.NoError(err)
-	t.Log(rels)
 }
 
 func TestCheckArrow(t *testing.T) {
@@ -108,7 +106,6 @@ func TestCheckArrow(t *testing.T) {
 	relSeq, err := it.Check(ctx, []string{"companyplan"}, "legal")
 	require.NoError(err)
 
-	rels, err := CollectAll(relSeq)
+	_, err = CollectAll(relSeq)
 	require.NoError(err)
-	t.Log(rels)
 }
--- a/pkg/query/union_test.go
+++ b/pkg/query/union_test.go
@@ -4,6 +4,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/pkg/tuple"
 )
 
 func TestUnionIterator(t *testing.T) {
@@ -30,8 +32,37 @@ func TestUnionIterator(t *testing.T) {
 		rels, err := CollectAll(relSeq)
 		require.NoError(err)
 
-		// The result should contain resources where either condition is met
-		require.NotEmpty(rels, "Union should find relations from either iterator")
+		// Union should contain relations from both iterators for alice on doc1 and doc2
+		// DocumentAccess: alice viewer/editor/owner on doc1, alice viewer on doc2
+		// MultiRole: alice viewer/editor/owner on doc1
+		// Union should deduplicate, so we expect: viewer/editor/owner on doc1, viewer on doc2
+		expected := []tuple.Relationship{
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc1", "viewer"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc1", "editor"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc1", "owner"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc2", "viewer"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+		}
+		require.ElementsMatch(expected, rels)
 	})
 
 	t.Run("Check_EmptyUnion", func(t *testing.T) {
@@ -62,8 +93,29 @@ func TestUnionIterator(t *testing.T) {
 		rels, err := CollectAll(relSeq)
 		require.NoError(err)
 
-		// Should return same results as the single iterator alone
-		require.NotEmpty(rels, "Single iterator union should return results")
+		// Should return same results as DocumentAccess iterator for alice on doc1
+		// Alice has viewer, editor, and owner access to doc1
+		expected := []tuple.Relationship{
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc1", "viewer"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc1", "editor"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+			{
+				RelationshipReference: tuple.RelationshipReference{
+					Resource: tuple.ONR("document", "doc1", "owner"),
+					Subject:  tuple.ONR("user", "alice", "..."),
+				},
+			},
+		}
+		require.ElementsMatch(expected, rels)
 	})
 
 	t.Run("Check_EmptyResourceList", func(t *testing.T) {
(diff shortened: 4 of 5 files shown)


pkg/query/arrow.go
```diff
@@ -0,0 +1,92 @@
...
+func (a *Arrow) Check(ctx *Context, resourceIds []string, subjectId string) (RelationSeq, error) {
+	// TODO -- the ordering, directionality, batching, everything can depend on other statistics.
+	//
+	// There are three major strategies:
+	// - LookupSubjects on the left, Check on the right (as per this implementation)
+	// - LookupResources on the right, Check on the left
+	// - LookupSubjects on left, LookupResources on right, and intersect the two iterators here (especially if they are known to be sorted)
+	//
+	// But for now, this is a proof-of-concept, so the first one, one-by-one (no batching).
+	// This is going to be the crux of a lot of statistics optimizations. Many others are
```
tstirrat15: Unfinished comment

pkg/query/arrow.go
```diff
@@ -0,0 +1,92 @@
...
+			if err != nil {
+				yield(Relation{}, err)
+				return
+			}
+			for rel, err := range subit {
+				if err != nil {
+					yield(Relation{}, err)
+					return
+				}
+				checkit, err := a.right.Check(ctx, []string{rel.Subject.ObjectID}, subjectId)
```
tstirrat15: is `Check` in this context shorthand for "does this relation (or equivalent) exist?" Or is it actually checking a permission in the broader sense? If so, where does the name of that permission come from?

pkg/query/arrow.go
```diff
@@ -0,0 +1,92 @@
+package query
+
+import (
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+type Arrow struct {
+	left  Iterator
+	right Iterator
+}
```
tstirrat15, line 15: Is this `Arrow` in the sense of a schema arrow?

pkg/query/quick_e2e_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// This stands in for the step that pre-builds query trees from the schema,
+	// by iterating through the relationships in the schema and then walking them.
+	//
+	// In this case, it's a little contrived.
+	vande := query.NewRelationIterator(dsSchema.Definitions["document"].Relations["viewer_and_editor"].BaseRelations[0])
+	edit := query.NewRelationIterator(dsSchema.Definitions["document"].Relations["editor"].BaseRelations[0])
+	it := query.NewIntersection()
+	it.AddSubIterator(vande)
+	it.AddSubIterator(edit)
```
tstirrat15: Curious about this - what step of the process of doing an API call would set these up?

pkg/query/union.go
```diff
@@ -0,0 +1,83 @@
...
+
+func NewUnion() *Union {
+	return &Union{}
+}
+
+func (u *Union) AddSubIterator(subIt Iterator) {
+	u.subIts = append(u.subIts, subIt)
+}
+
+func (u *Union) Check(ctx *Context, resourceIds []string, subjectIds string) (RelationSeq, error) {
```
tstirrat15: Is the idea that a check of resource/subject pairs returns a sequence of extant pairs? Like it's a function from resources (cross) subject IDs to (resource, subject) tuples?

pkg/query/arrow.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func (a *Arrow) Check(ctx *Context, resourceIds []string, subjectId string) (RelationSeq, error) {
+	// TODO -- the ordering, directionality, batching, everything can depend on other statistics.
+	//
+	// There are three major strategies:
+	// - LookupSubjects on the left, Check on the right (as per this implementation)
+	// - LookupResources on the right, Check on the left
+	// - LookupSubjects on left, LookupResources on right, and intersect the two iterators here (especially if they are known to be sorted)
```
tstirrat15: This comment was helpful for me - will the other Check implementations have similar context, or is this specific to the arrow because it's the only one doing a "walk?"

pkg/query/build_tree.go
```diff
@@ -0,0 +1,139 @@
...
+		return b.buildIteratorFromRelation(r, withSubRelations)
+	}
+	return nil, fmt.Errorf("BuildIteratorFromSchema: couldn't find a relation or permission named `%s` in definition `%s`", relationName, definitionName)
+}
+
+func (b *IteratorBuilder) buildIteratorFromRelation(r *schema.Relation, withSubRelations bool) (Iterator, error) {
+	if len(r.BaseRelations) == 1 {
+		return b.buildBaseRelationIterator(r.BaseRelations[0], withSubRelations)
+	}
+	union := NewUnion()
```
tstirrat15, line 49: If a relation has just a single element, are we treating it as a unary union?

tstirrat15 (dismissed):
See comments, otherwise LGTM

pkg/query/quick_e2e_test.go
barakmich: It's meant to be done ahead of the API call -- by building an abstract tree, cloning it, and then mutating it, the schema side is done (and cacheable)

pkg/query/arrow.go
barakmich: Yes, but a comment here would be worthwhile

pkg/query/arrow.go
barakmich: Yes, it's "does this relation exist" in this set -- but this is also equivalent to CheckPermission

pkg/query/union.go
barakmich: Yep -- but again this should go in an interface comment to clarify

pkg/query/build_tree.go
barakmich: That's what the line above accounts for -- if there's one BaseRelation, it's a simple relation, otherwise it's a union

pkg/query/fixed.go
```diff
@@ -0,0 +1,81 @@
+package query
+
+import "fmt"
+
+// FixedIterator represents a fixed set of pre-computed relations.
+// This is often useful for testing, but can also be used in rare situations
+// where we'd like to force a set of intermediate relations.
+
+// For example: document->folder->ownerGroup->user -- and we'd like to
+// find all documents (LookupResources) that traverse a known folder->ownerGroup relationship
```
tstirrat15: siiick

tstirrat15 (dismissed):
LGTM

pkg/query/build_tree.go
```diff
@@ -0,0 +1,141 @@
...
+			if err != nil {
+				return nil, err
+			}
+			union.AddSubIterator(it)
+		}
+		if len(union.subIts) == 1 {
+			return union.subIts[0], nil
+		}
+		return union, nil
+	case *schema.IntersectionOperation:
```
josephschorr, line 100: newlines before each case statement (except the first)

pkg/query/build_tree.go
```diff
@@ -0,0 +1,141 @@
...
+				return nil, err
+			}
+			inter.AddSubIterator(it)
+		}
+		if len(inter.subIts) == 1 {
+			return inter.subIts[0], nil
+		}
+		return inter, nil
+	case *schema.ExclusionOperation:
+		return nil, ErrUnimplemented
```
josephschorr: have `ErrUnimplemented` be a `spiceerrors.MustBugf` so we panic if it is encountered (during testing)

pkg/query/arrow_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Test with multiple resource IDs
+	relSeq, err := arrow.Check(nil, []string{"spec1", "spec2", "nonexistent"}, "alice")
+	require.NoError(err)
+
+	rels, err := query.CollectAll(relSeq)
+	require.NoError(err)
+
+	t.Logf("Multiple resources results: %+v", rels)
+	// The result should include all valid arrow relationships found across all resources
```
josephschorr: missing check here?

pkg/query/build_tree.go
```diff
@@ -0,0 +1,141 @@
...
+				return nil, err
+			}
+			right, err := b.buildIteratorFromSchemaInternal(br.Type, perm.Right, false)
+			if err != nil {
+				return nil, err
+			}
+			arrow := NewArrow(left, right)
+			union.AddSubIterator(arrow)
+		}
+		if len(union.subIts) == 1 {
```
josephschorr: This is an optimization: let's leave it out for now

pkg/query/build_tree.go
```diff
@@ -0,0 +1,141 @@
...
+	case *schema.UnionOperation:
+		union := NewUnion()
+		for _, op := range perm.Children {
+			it, err := b.buildIteratorFromOperation(p, op)
+			if err != nil {
+				return nil, err
+			}
+			union.AddSubIterator(it)
+		}
+		if len(union.subIts) == 1 {
```
josephschorr: ditto

pkg/query/build_tree.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (b *IteratorBuilder) buildBaseRelationIterator(br *schema.BaseRelation, withSubRelations bool) (Iterator, error) {
+	base := NewRelationIterator(br)
+	if !withSubRelations {
+		return base, nil
+	}
+	if br.Subrelation == "" {
+		return base, nil
+	}
+	// We must check the effective arrow of a subrelation if we have one and subrelations are enabled
```
josephschorr, line 129: newlines around conditional blocks

pkg/query/build_tree_test.go
```diff
@@ -0,0 +1,363 @@
...
+
+	// This stands in for the step of fetching and caching the schema locally.
+	objectDefs := []*corev1.NamespaceDefinition{testfixtures.UserNS.CloneVT(), testfixtures.FolderNS.CloneVT(), testfixtures.DocumentNS.CloneVT()}
+	dsSchema, err := schema.BuildSchemaFromDefinitions(objectDefs, nil)
+	require.NoError(err)
+
+	it, err := query.BuildIteratorFromSchema(dsSchema, "document", "edit")
+	require.NoError(err)
+
+	t.Logf("\n%s", it.Explain().String())
```
josephschorr: remove logs

pkg/query/build_tree_test.go
```diff
@@ -0,0 +1,363 @@
...
+func TestBuildTreeExclusionOperation(t *testing.T) {
+	t.Parallel()
+
+	require := require.New(t)
+
+	// Create a simple schema to test exclusion handling
+	userDef := testfixtures.UserNS.CloneVT()
+
+	// Create a document with an exclusion operation
+	docDef := &corev1.NamespaceDefinition{
```
josephschorr: I'd recommend using the namespace builders here and in the rest of the tests: https://github.com/authzed/spicedb/blob/ebd258d28cb40648490227d6c32ffcf74295bd2f/pkg/namespace/builder.go#L4

pkg/query/quick_e2e_test.go
```diff
@@ -0,0 +1,115 @@
...
```
josephschorr: I wonder if we should have `AddSubIterator` return an *updated* struct and keep all of this as edit-by-copy: its likely worse for mem usage, but far easier to reason about?

pkg/query/types.go
```diff
@@ -0,0 +1,95 @@
...
+// Plan is the external-facing notion of a query plan. These follow the general API for
+// querying anything in the database as well as describing the plan.
+type Plan interface {
+	// Check tests if, for the underlying set of relationships (which may be a full expression or a basic lookup, depending on the iterator)
+	// any of the `resourceIDs` are connected to `subjectID`.
+	// Returns the sequence of matching relations, if they exist, at most `len(resourceIDs)`.
+	Check(ctx *Context, resourceIDs []string, subjectID string) (RelationSeq, error)
+
+	// LookupSubjects returns a sequence of all the relations in this set that match the given resourceID.
+	LookupSubjects(ctx *Context, resourceID string) (RelationSeq, error)
```
josephschorr: I wonder if we should name these `IterSubjects` and `IterResources` to distinguish from the lookups we have today

pkg/query/types.go
```diff
@@ -0,0 +1,95 @@
...
+
+var ErrUnimplemented = errors.New("feature not yet implemented")
+
+// Plan is the external-facing notion of a query plan. These follow the general API for
+// querying anything in the database as well as describing the plan.
+type Plan interface {
+	// Check tests if, for the underlying set of relationships (which may be a full expression or a basic lookup, depending on the iterator)
+	// any of the `resourceIDs` are connected to `subjectID`.
+	// Returns the sequence of matching relations, if they exist, at most `len(resourceIDs)`.
+	Check(ctx *Context, resourceIDs []string, subjectID string) (RelationSeq, error)
```
josephschorr, line 26: I wonder if this is just a `FilteredIterSubjects`?

pkg/schema/v2/operations.go
```diff
@@ -0,0 +1,38 @@
+package schema
```
josephschorr, line 1: mind adding a `docs.go` to each of these new packages, explaining what they do and how they are currently for testing only?

pkg/schema/v2/schema.go
```diff
@@ -0,0 +1,86 @@
...
+	corev1 "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
+)
+
+type Schema struct {
+	Definitions map[string]*Definition
+	Caveats     map[string]*Caveat
+}
+
+type Definition struct {
```
josephschorr, line 15: doc comments on all exported items

pkg/query/build_tree.go
barakmich: ~I'll agree for the explicit union and intersection, but this one I think is worth keeping.~

Nah, actually, fine w/e

pkg/query/quick_e2e_test.go
barakmich: That's a thought, and I've done that in Go code before in other circumstances (mostly when it's something like `type Foo uint64` and I'm tweaking the bits) -- here, building is, aside from schema read time, kinda rare, so I'm not concerned.

One compromise I kinda like and will go ahead and do preemptively is to simply make AddSubIterator private -- only this package can mutate like that, nothing else.

pkg/query/types.go
barakmich: Kind of, but then so is a full table scan 🙃 

The only reason to have a `[]string` for resources is because datastore lets us use it -- it's completely reasonable to make Check do a single check, as an alternative (returning a Relation because we don't know if there's caveats)

pkg/query/arrow_test.go
```diff
@@ -0,0 +1,176 @@
...
+		t.Parallel()
+
+		// Test arrow operation: find resources where left side connects to right side
+		// This looks for documents whose parent folder has viewers
+		relSeq, err := arrow.Check(nil, []string{"spec1", "spec2"}, "alice")
+		require.NoError(err)
+
+		_, err = CollectAll(relSeq)
+		require.NoError(err)
+		// Results depend on the specific relations in our test data
```
josephschorr: remove this comment or check results?

pkg/query/types.go
josephschorr: I meant more the name; passing in the slice makes sense

pkg/query/union_test.go
```diff
@@ -0,0 +1,267 @@
...
+	// Note: Due to union's stateful resource elimination optimization,
+	// we can't reliably test result equivalence without complex setup
+	// The key test is that Clone() creates independent objects with same structure
+
+	// Verify the clone is independent and has same structure
+	require.Equal(originalExplain.String(), clonedExplain.String())
+
+	// Just test that both can execute operations without error
+	// The union's resource elimination optimization makes result comparison tricky
+	t.Log("Clone test passed - both iterators have same structure and can execute")
```
josephschorr: remove `t.Log` here and elsewhere

tstirrat15 (approved):
LGTM

pkg/query/build_tree.go
josephschorr: Add a comment

pkg/query/build_tree_test.go
```diff
@@ -0,0 +1,308 @@
...
+				namespace.Nil(),
+			),
+		),
+	)
+
+	objectDefs := []*corev1.NamespaceDefinition{userDef, docDef}
+	dsSchema, err := schema.BuildSchemaFromDefinitions(objectDefs, nil)
+	require.NoError(err)
+
+	// Test building iterator for exclusion permission - should panic
```
josephschorr, line 242: Add a TODO to remove once we support exclusion

pkg/query/datastore.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	"github.com/authzed/spicedb/pkg/datastore/queryshape"
+	"github.com/authzed/spicedb/pkg/schema/v2"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+const NoSubRel = ""
```
josephschorr, line 13: we usually using `tuple.Ellipsis` internally for the "no subject rel" case

pkg/query/fixed_helpers_test.go
```diff
@@ -0,0 +1,202 @@
+package query
```
josephschorr, line 1: rename this file to not end in `_test`, but rather be something like `testutil.go`

PR #2532, opened 2025-08-14
Improve the errors returned from schema changes

josephschorr:


internal/services/steelthreadtesting/testdata/basic-schema-and-data.yaml
```diff
@@ -16,109 +16,5 @@ schema: |+
...
 
 relationships: |
-  // A single user2:* on a document.
   document:somedoc#editor@user2:*
-
-  // 100 direct users
-  // for userid in range(0, 100)
-  //    document:somedoc#viewer@user:user-{userid}
   document:somedoc#viewer@user:user-0
-  document:somedoc#viewer@user:user-1
```
vroldanbet, line 26: @josephschorr why was all of this removed?

internal/services/shared/schema.go
vroldanbet, line 1: I'd suggest adding tests that cover all possible schema validation errors and check the returned metadata

internal/services/steelthreadtesting/testdata/basic-schema-and-data.yaml
josephschorr: Because it wasn't needed and was making the error message flaky

internal/services/shared/schema.go
josephschorr: We have those; it was just missing a case

PR #2534, opened 2025-08-15
Fix issue with schema write validation

tstirrat15:
## Description
We ran into this issue with one our own schemas/processes. The schema write was failing when it should be succeeding.

## Changes
* Reproduce the issue in a steelthread
* Fix the issue
## Testing
Review.


tstirrat15:
The fix looks good to me. I would approve but I opened the PR.

barakmich (approved):

Verolop (approved):

PR #2537, opened 2025-08-18
ci: Integrate commitizen for conventional commit validation

mazdakb:
Integrate conventional commit validation into lint.yaml workflow

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -49,3 +49,85 @@ jobs:
         with:
           path: ""
           fixup-command: "go run mage.go lint:extra"
+
+  conventional-commits:
+    name: "Lint Commit Messages"
+    runs-on: "depot-ubuntu-24.04-small"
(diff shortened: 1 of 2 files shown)


miparnisari (approved):
I'm curious as to why we preferred a github step instead of a local commit hook? With this approach, you won't find out that your commit msg is wrong until after you push, which is a very late feedback.

But in any case, this LGTM, we can try this out

mazdakb:
@miparnisari I think both are good. It comes down to preference. I like to be able to push commits without caring much about them and then before putting it up for review, cleaning them up or rearranging them. The CI step does not get in my way.

tstirrat15:
One thing I'm curious about: is there potential for this step to trip in `main` when we squash if we don't make the squash commit follow conventional commits? Is there a way to avoid that?

mazdakb:
@tstirrat15 It's possible to add `ruleset` on `main` to enforce commit message patterns. It's not possible to enforce that particular check using Github actions.

PR #2539, opened 2025-08-22
vuln: update go from 1.24.0 to latest 1.25.0

kartikaysaxena:
## Description
Updated to latest go v1.25.0 and `github.com/go-viper/mapstructure` to v2.4.0 to fix a vuln
go v1.24.0 was affected by a vuln in the `database/sql` lib of go due to a race condition in sql.Row.Scan and sql.Rows.Scan

## Testing
see that the tests go green
update: for builds see https://github.com/authzed/actions/pull/98

## References
see: https://pkg.go.dev/vuln/GO-2025-3849 https://github.com/advisories/GHSA-2464-8j7c-4cjm


tstirrat15:
The build failure is because the actions in this repo also need to be updated to use v1.25. The commit lint failure is about making the commit messages use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

tstirrat15 (approved):
LGTM

PR #2543, opened 2025-08-27
fix: improve the cluster error message when errors from all dispatchers

josephschorr:


internal/dispatch/remote/cluster.go
```diff
@@ -737,12 +739,14 @@ func dispatchStreamingRequest[Q streamingRequestMessage, R responseMessage](
 		return nil
 	}
 
-	// Otherwise return a combined error.
-	cerr := fmt.Errorf("no dispatcher returned results")
-	for name, err := range errorsByDispatcherName {
-		cerr = fmt.Errorf("%w; error in dispatcher %s: %w", cerr, name, err)
+	// If there is a primary dispatcher error, return it.
+	if primaryErr, ok := errorsByDispatcherName[primaryDispatcher]; ok {
```
miparnisari, line 743: - Nit: you could simplify this long function (😅) a bit by making `errorsByDispatcherName` a `sync.Map{}` and removing `errorsLock`
- Could we not update the tests? You could have a test with two dispatchers that both throw an error and assert on the error message. You already have the setup to do this it seems

internal/dispatch/remote/cluster.go
```diff
@@ -737,12 +739,14 @@ func dispatchStreamingRequest[Q streamingRequestMessage, R responseMessage](
...
+		log.Warn().Err(primaryErr).Errs("all-errors", slices.Collect(maps.Values(errorsByDispatcherName))).Msg("returning primary dispatcher error as no dispatchers returned results")
```
miparnisari, line 744:
```suggestion
		log.Warn().Err(primaryErr).Errs("all-errors", slices.Collect(maps.Values(errorsByDispatcherName))).Msg("returning primary dispatcher error as no secondary dispatchers returned results")
```

internal/dispatch/remote/cluster.go
josephschorr: We can refactor as a followup, if necessary

internal/dispatch/remote/cluster.go
josephschorr: This is only called if *none* of them returned any results, including them primary

internal/dispatch/remote/cluster.go
josephschorr: Added test

internal/dispatch/remote/cluster_test.go
```diff
@@ -1179,6 +1179,60 @@ func TestDALCount(t *testing.T) {
...
+			Namespace: "foo",
+			ObjectId:  "bar",
+			Relation:  "...",
+		},
+		Metadata: &v1.ResolverMeta{DepthRemaining: 50},
+	}, stream)
+
+	// Should get the primary dispatcher error when no dispatcher returns results
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "primary dispatcher error")
```
miparnisari, line 1233: also `require.NotContains(t, err.Error(), "secondary dispatcher error")`

otherwise this test passes even without your change

miparnisari (approved):

PR #2546, opened 2025-09-02
query: Implement an executor pattern for the iterators

barakmich:


tstirrat15 (approved):
I like the way this separates the concerns. I'd prefer if it was called `ExecutionContext` rather than `Context` to disambiguate it from `context.Context`, but that's not a blocker.

PR #2548, opened 2025-09-03
feat: add exclusion iterator

barakmich:


pkg/query/build_tree.go
```diff
@@ -85,6 +84,9 @@ func (b *iteratorBuilder) buildIteratorFromOperation(p *schema.Permission, op sc
 		return union, nil
 
 	case *schema.RelationReference:
+		if perm.RelationName == "_nil" {
+			return NewEmptyFixedIterator(), nil
```
tstirrat15, line 88: Why is this branch needed?

pkg/query/exclusion.go
```diff
@@ -0,0 +1,102 @@
...
+		}, nil
+	}
+
+	// Get all relations from the excluded set
+	excludedSeq, err := ctx.Check(e.excluded, resourceIDs, subjectID)
+	if err != nil {
+		return nil, err
+	}
+
+	excludedRels, err := CollectAll(excludedSeq)
```
tstirrat15, line 48: Could this be done incrementally? Or is that an optimization that will come later?

pkg/query/exclusion.go
```diff
@@ -0,0 +1,102 @@
...
+	return func(yield func(Relation, error) bool) {
+		for _, mainRel := range mainRels {
+			found := false
+
+			// Check if this relation exists in the excluded set (only compare endpoints: resource and subject object types/IDs)
+			for _, excludedRel := range excludedRels {
+				if mainRel.Resource.ObjectType == excludedRel.Resource.ObjectType &&
+					mainRel.Resource.ObjectID == excludedRel.Resource.ObjectID &&
+					mainRel.Subject.ObjectType == excludedRel.Subject.ObjectType &&
+					mainRel.Subject.ObjectID == excludedRel.Subject.ObjectID {
```
tstirrat15, line 63: This seems like it might be worth making into a `rel.Equal` method or something

pkg/query/union.go
```diff
@@ -23,7 +23,8 @@ func (u *Union) addSubIterator(subIt Iterator) {
 }
 
 func (u *Union) CheckImpl(ctx *Context, resourceIDs []string, subjectID string) (RelationSeq, error) {
-	remaining := resourceIDs
+	remaining := make([]string, len(resourceIDs))
+	copy(remaining, resourceIDs)
```
tstirrat15, line 27: Why is this necessary? Are we mutating this somewhere?

tstirrat15 (commented):
See comments

pkg/query/exclusion.go
barakmich: It could, but it also makes some sense that the maximum length to check is the length of the input slice -- so just materializing it can't overload too much.

pkg/query/build_tree.go
barakmich: It's an addition for handling the `nil` keyword that, while handled in the schema package, was ignored here (and so the fix snuck in while I was looking at some of the other tests)

pkg/query/exclusion.go
barakmich: This is coming in the very next PR because this gets refactored

pkg/query/union.go
barakmich: Yep, right here with slicez.Delete; which causes a bug because it could be used again. Could also change it to create a new incremental list, up to you for what you prefer.

tstirrat15 (approved):
LGTM

PR #2549, opened 2025-09-04
Cursor concurrency improvements

josephschorr:
Implement improvements to the cursoring library used for LR3 to reduce goroutine creation

## Description

1. Iterator concurrency optimization - Added special handling to avoid creating goroutines when concurrency is set to 1
2. Cursored mapping iterator improvement - Modified the iterator to only spawn goroutines when necessary by utilizing TaskRunner
3. LR3 entrypoints optimization - Skip running entrypoints entirely when there aren't any to process

## Testing

`mage test:all`, including new unit tests

### DIFF
--- a/internal/cursorediterator/blocks.go
+++ b/internal/cursorediterator/blocks.go
@@ -157,6 +157,42 @@ func CursoredParallelIterators[I any](
 
 	itersToRun := orderedIterators[currentStartingBranchIndex:]
 
+	// Special case for concurrency=1: execute sequentially without goroutines
+	if concurrency == 1 {
+		return func(yield func(ItemAndCursor[I], error) bool) {
+			defer cancel()
+
+			for collectorOffsetIndex, iter := range itersToRun {
+				iteratorRemainingCursor := remainingCursor
+				if collectorOffsetIndex > 0 {
+					iteratorRemainingCursor = nil
+				}
+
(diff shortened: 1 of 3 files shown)


internal/cursorediterator/blocks.go
```diff
@@ -300,15 +336,88 @@
...
+						return
+					default:
+						if err != nil {
+							if !yield(ItemAndCursor[I]{}, err) {
+								return
+							}
+							return
+						}
+
+						cursorStr, err := cursorToStringConverter(chunk.Follow)
```
mazdakb: Looking at `internal/graph/lookupresources3.go:1303` it seems `cursorToStringConverter` is invariant within a chunk. We can then compute it outside the mapper loop.

internal/cursorediterator/blocks.go
```diff
@@ -300,15 +336,88 @@ func CursoredProducerMapperIterator[C any, P any, I any](
...
+					}
+					return
+				}
+
+				// Skip HoldForMappingComplete in sequential mode since there's no parallelism to synchronize
+				if _, ok := p.(HoldForMappingComplete[P, C]); ok {
+					continue
+				}
+
+				chunk := p.(ChunkAndFollow[P, C])
```
mazdakb: There are two type assertions here, and perhaps it'd be safer to switch over them with such a default that returns a relevant error.

PR #2550, opened 2025-09-04
refactor: Query iterators now operate in Objects and ObjectRelations instead of strings

barakmich:

### DIFF
--- a/pkg/query/datastore.go
+++ b/pkg/query/datastore.go
@@ -35,7 +35,12 @@ func (r *RelationIterator) buildSubjectRelationFilter() datastore.SubjectRelatio
 	return datastore.SubjectRelationFilter{}.WithNonEllipsisRelation(r.base.Subrelation)
 }
 
-func (r *RelationIterator) CheckImpl(ctx *Context, resourceIDs []string, subjectID string) (RelationSeq, error) {
+func (r *RelationIterator) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (RelationSeq, error) {
+	resourceIDs := make([]string, len(resources))
+	for i, res := range resources {
+		resourceIDs[i] = res.ObjectID
+	}
+
 	filter := datastore.RelationshipsFilter{
 		OptionalResourceType:     r.base.DefinitionName(),
 		OptionalResourceIds:      resourceIDs,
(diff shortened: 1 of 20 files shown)


pkg/query/datastore.go
```diff
@@ -35,15 +35,20 @@ func (r *RelationIterator) buildSubjectRelationFilter() datastore.SubjectRelatio
 	return datastore.SubjectRelationFilter{}.WithNonEllipsisRelation(r.base.Subrelation)
 }
 
-func (r *RelationIterator) CheckImpl(ctx *Context, resourceIDs []string, subjectID string) (RelationSeq, error) {
+func (r *RelationIterator) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (RelationSeq, error) {
+	resourceIDs := make([]string, len(resources))
```
tstirrat15, line 39: For my own edification: what's the difference between `make([]string, 0, len(resources))` and what's written here? When would you use one over the other?

pkg/query/union.go
```diff
@@ -37,32 +34,33 @@ func (u *Union) CheckImpl(ctx *Context, resourceIDs []string, subjectID string)
...
-		for _, r := range rels {
-			if idx := slices.Index(remaining, r.Resource.ObjectID); idx != -1 {
-				remaining = slices.Delete(remaining, idx, idx+1)
-			}
-		}
-
-		if len(remaining) == 0 {
-			break
+	// Deduplicate relations
+	seen := make(map[string]bool)
```
tstirrat15, line 40: dunno if it's useful here, but we have the `mapz` package that has a `Set` in it.

pkg/query/datastore.go
barakmich: they're damn close to equivalent, depending on the next lines:
```
for i, res := range resources {
	resourceIDs[i] = res.ObjectID
}
```
In that case, I'm filling in the zero values (empty strings, which I believe under the hood are particular string pointers) with the desired value. you could rewrite this with `make([]string, 0, len(resources))` and use `append()` instead of directly writing the slot with the subscript.

So in this case it makes little to no difference. If you want to do it in a way that's not append-friendly, then this is the better way (say, constructing a binary tree inside an array, or a sort function, that sort of thing) -- so when I can and I know the lengths of both arrays I opt for this form, but it's really not a strong argument either way.

pkg/query/union.go
barakmich: I'll leave it for now because I'm not even sure if dedupe behavior is correct here -- but we'll cross that bridge when we get there (chances are, it becomes a flag, and then we can use mapz)

PR #2553, opened 2025-09-09
feat: add wildcard support to the plan

barakmich:
Dependent upon #2552


internal/services/integrationtesting/query_plan_consistency_test.go
```diff
@@ -127,13 +127,13 @@ func runQueryPlanAssertions(t *testing.T, handle *queryPlanConsistencyHandle) {
 
 							switch entry.expectedPermissionship {
 							case v1.CheckPermissionResponse_PERMISSIONSHIP_CONDITIONAL_PERMISSION:
-								require.Equal(len(rels), 1)
+								require.Equal(1, len(rels))
```
tstirrat15: `require.Len`

pkg/query/datastore.go
```diff
@@ -36,6 +36,13 @@ func (r *RelationIterator) buildSubjectRelationFilter() datastore.SubjectRelatio
 }
 
 func (r *RelationIterator) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (RelationSeq, error) {
+	// If the subject type doesn't match the base relation type, return no results
+	if subject.ObjectType != r.base.Type {
```
tstirrat15, line 40: Why was this branch necessary?

pkg/query/union.go
```diff
@@ -36,17 +36,29 @@ func (u *Union) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
+				seen[key] = rel
+			}
 		}
 	}
 
+	// Convert map to slice
+	deduplicated := make([]Relation, 0)
+	for _, rel := range seen {
+		deduplicated = append(deduplicated, rel)
+	}
```
tstirrat15: https://pkg.go.dev/maps#Values

tstirrat15: Unless a slice is needed specifically?

pkg/query/wildcard.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+	}
+
+	reader := ctx.Datastore.SnapshotReader(ctx.Revision)
+
+	relIter, err := reader.QueryRelationships(ctx, filter,
+		options.WithSkipCaveats(w.base.Caveat == ""),
+		options.WithSkipExpiration(!w.base.Expiration),
+		options.WithQueryShape(queryshape.CheckPermissionSelectDirectSubjects),
+	)
```
tstirrat15: I'm surprised to see a direct reference to the datastore here rather than it being delegated forward to the datastore iterator. Is that because there really isn't anything else that you can do with a wildcard?

pkg/query/datastore.go
barakmich: Technically not, just a quick optimization if there's no way to satisfy the check.

pkg/query/wildcard.go
barakmich: Yeah; the other way to do this is to combine it into the datastore, which may be the better option actually. I'll do that.

Issue #2518, opened 2025-08-05
Using SPICEDB_DATASTORE_READ_REPLICA_CONN_URI and SPICEDB_DATASTORE_CONN_URI environment variables causes duplicate metrics error


shaafekhan:
### What platforms are affected?

linux

### What architectures are affected?

amd64

### What SpiceDB version are you using?

v1.43.0

### Steps to Reproduce

1. Run SpiceDB container with the following environment variables:
- SPICEDB_DATASTORE_READ_REPLICA_CONN_URI=read-replica endpoint  (user:password@tcp(read-only-host:3306)/database?tls=skip-verify&parseTime=true
- SPICEDB_DATASTORE_CONN_URI=write endpoint (user:password@tcp(write-host:3306)/database?tls=skip-verify&parseTime=true
2. Our spicedb config:
  containers:
    - args:
        - spicedb
        - serve
        - '--datastore-engine=mysql'
        - '--grpc-enabled=true'
        - '--http-enabled=true'
        - '--dispatch-cluster-enabled=true'
        - '--metrics-enabled=true'
        - '--otel-endpoint=datadog-statsd.monitoring.svc.cluster.local:4318'
        - '--otel-provider=otlphttp'
        - '--dispatch-upstream-addr=kubernetes://spicedb.spicedb:50053'
        - '--log-format=json'
        - '--otel-service-name=spicedb'
        - '--otel-insecure'
    - envs: (besides the datastore environment variables)
        - name: SPICEDB_GRPC_PRESHARED_KEY
          value: grpckey

We are running SpiceDB in EKS v1.31.1-eks-1b3e656

### Expected Result

``{"level":"info","time":"2025-08-05T08:16:34Z","message":"using mysql datastore engine"}``

### Actual Result

``{"level":"trace","time":"2025-08-03T21:48:27Z","message":"NewMySQLDatastore: unable to instrument connector: unable to register metric: duplicate metrics collector registration attempted"}``

``{"level":"error","error":"failed to create datastore: NewMySQLDatastore: unable to instrument connector. To view details of this error (that may contain sensitive information), please run with --log-level=trace``

shaafekhan:
Just an addendum to testing, it seems that the following setup also leads to the same error when using a docker compose setup. 

```
services:
  spicedb-engine-migrate:
    image: ghcr.io/authzed/spicedb:v1.43.0 #also tried 1.45.1
    command: migrate head
    environment:
[...]

PR #2554, opened 2025-09-09
Fix duplicate metrics registration in mysql datastore

leonnicolas:
When using a replica with the mysql datastore, we register the
same metrics again and again.

This commit ignores AlreadyRegisteredError when registering our
custom metrics and adds a `db_name` label similar to the postgres
instrumentation.

Additionally, it updates the tests for multi instance creation in a way that we
don't create 2 connections with the same index. Otherwise we would
get another registration error when registering the sql collector `sqlstats`.

fixes: https://github.com/authzed/spicedb/issues/2518

Signed-off-by: leonnicolas <leonloechner@gmx.de>

### DIFF
--- a/internal/datastore/mysql/datastore_test.go
+++ b/internal/datastore/mysql/datastore_test.go
@@ -6,6 +6,7 @@ package mysql
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


leonnicolas:

leonnicolas:
I hope I could increase the patch coverage with my last changes and fix the commit message format

miparnisari:
Hi @leonnicolas, I sent this https://github.com/authzed/spicedb/pull/2707 😅 I am sorry that I didn't see this before. I am going to close it, but thanks for spending time looking at it!

PR #2555, opened 2025-09-10
chore: improve rows.Err linter check

miparnisari:
## Description

From https://github.com/authzed/spicedb/pull/2526#issuecomment-3221437946

### DIFF
--- a/internal/datastore/crdb/pool/pool.go
+++ b/internal/datastore/crdb/pool/pool.go
@@ -172,7 +172,11 @@ func (p *RetryPool) QueryFunc(ctx context.Context, rowsFunc func(ctx context.Con
 			return err
 		}
 		defer rows.Close()
-		return rowsFunc(ctx, rows)
+		err = rowsFunc(ctx, rows)
+		if err != nil {
+			return err
+		}
+		return rows.Err()
 	})
 }
 
--- a/internal/datastore/postgres/common/pgx.go
+++ b/internal/datastore/postgres/common/pgx.go
@@ -278,7 +278,11 @@ func (t *QuerierFuncs) QueryFunc(ctx context.Context, rowsFunc func(ctx context.
 		return err
 	}
 	defer rows.Close()
-	return rowsFunc(ctx, rows)
+	err = rowsFunc(ctx, rows)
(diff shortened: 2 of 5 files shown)


internal/datastore/crdb/pool/pool.go
```diff
@@ -172,7 +172,11 @@ func (p *RetryPool) QueryFunc(ctx context.Context, rowsFunc func(ctx context.Con
 			return err
 		}
 		defer rows.Close()
-		return rowsFunc(ctx, rows)
+		err = rowsFunc(ctx, rows)
```
miparnisari, line 175: there is at least one `rowsFunc` that I saw is not handling `rows.Err`. i thought it was sensible to do this check outside so as to safeguard against that _and_ future changes, but i'm open to changing each individual callee.

josephschorr: I think we need to check for the error here first; if we get back an `err`, return that. Otherwise, check the `rows.Err()`

internal/datastore/postgres/common/pgx.go
```diff
@@ -278,7 +278,11 @@ func (t *QuerierFuncs) QueryFunc(ctx context.Context, rowsFunc func(ctx context.
 		return err
 	}
 	defer rows.Close()
-	return rowsFunc(ctx, rows)
+	err = rowsFunc(ctx, rows)
```
josephschorr, line 281: ditto

tools/analyzers/iferrafterrowclosecheck/iferrafterrowclosecheck.go
```diff
@@ -129,40 +123,3 @@ func hasImmediateErrorCheck(closeStmt *ast.ExprStmt, rowsName string, stack []as
 
 	return false
 }
-
-func isStatementRightAfter(first nodeAndStack, second nodeAndStack) bool {
```
josephschorr, line 133: why remove this?

miparnisari: unused

PR #2557, opened 2025-09-10
test: correct usage of goleak.IgnoreCurrent()

miparnisari:
## Description

See https://github.com/uber-go/goleak/issues/137


internal/services/integrationtesting/cert_test.go
```diff
@@ -284,5 +284,4 @@ func TestCertRotation(t *testing.T) {
 	case <-time.After(30 * time.Second):
 		require.Fail(t, "ungraceful server termination")
 	}
-	goleak.VerifyNone(t, goleak.IgnoreCurrent())
```
miparnisari, line 287: this doesn't do anything at the moment

internal/services/integrationtesting/cert_test.go
```diff
@@ -38,7 +38,7 @@ import (
 )
 
 func TestCertRotation(t *testing.T) {
-	t.Parallel()
+	defer goleak.VerifyNone(t, goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"))
```
miparnisari, line 41: See https://github.com/authzed/spicedb/pull/2558

internal/services/integrationtesting/cert_test.go
josephschorr: Wouldn't it verify it at that point?

miparnisari: yes but `goleak.IgnoreCurrent` will blindly mark every goroutine running as ignored 😓 therefore, no leaks will be found

internal/services/v1/permissions_test.go
```diff
@@ -646,10 +646,8 @@ func TestLookupResources(t *testing.T) {
 								tf.StandardDatastoreWithData,
 							)
 							client := v1.NewPermissionsServiceClient(conn)
-							t.Cleanup(func() {
-								goleak.VerifyNone(t, goleak.IgnoreCurrent())
-							})
-							t.Cleanup(cleanup)
+							defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
+							defer cleanup()
```
tstirrat15, line 650: The above change makes sense to me, but why did this get changed from `Cleanup` to `defer`?

miparnisari: if I could do `t.Cleanup(goleak.VerifyNone...)` then i think that would be equivalent to what I did, but i can't do that because `t.Cleanup` expects a function signature that `goleak` can't provide

tstirrat15 (approved):
LGTM

PR #2558, opened 2025-09-10
chore: remove opencensus library

miparnisari:
## Description

While working on https://github.com/authzed/spicedb/pull/2557/files#diff-0d9c5ed01c8a90734943ed1c028bda7f63b67ff60c564a042f69222e892ca962R41 noticed that this opencensus library  was deprecated in 2023.
Also DatastoreMetricsOptionLegacyPrometheus was marked as deprecated in February of this year.

So, killing two birds with one stone


pkg/cmd/datastore/datastore.go
```diff
@@ -272,7 +272,7 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	flagSet.StringVar(&opts.SpannerEmulatorHost, flagName("datastore-spanner-emulator-host"), "", "URI of spanner emulator instance used for development and testing (e.g. localhost:9010)")
 	flagSet.Uint64Var(&opts.SpannerMinSessions, flagName("datastore-spanner-min-sessions"), 100, "minimum number of sessions across all Spanner gRPC connections the client can have at a given time")
 	flagSet.Uint64Var(&opts.SpannerMaxSessions, flagName("datastore-spanner-max-sessions"), 400, "maximum number of sessions across all Spanner gRPC connections the client can have at a given time")
-	flagSet.StringVar(&opts.SpannerDatastoreMetricsOption, flagName("datastore-spanner-metrics"), "otel", `configure the metrics that are emitted by the Spanner datastore ("none", "native", "otel", "deprecated-prometheus")`)
```
tstirrat15, line 275: If someone is currently using this flag, is it a breaking change to remove it for them?

tstirrat15: Not a reason not to merge, but I want to make sure I understand that.

josephschorr: Yes, it would. We should hide the flag and emit a warning if used

miparnisari: to clarify: are y'all saying that i should keep `deprecated-prometheus` as an option here, and then if they pass that value, i say `you are using a deprecated flag; will default to otel metrics`?

miparnisari:
I just noticed that https://github.com/authzed/spicedb/pull/2329 is also open.

pkg/cmd/datastore/datastore.go
jzelinskie: I think we should keep it for a few releases logging a warning and using otel instead. Then we can finally fully remove it.

miparnisari: Please see https://github.com/authzed/spicedb/pull/2329

PR #2556, opened 2025-09-10
chore: bump authzed-go and pin OTEL libraries

miparnisari:
## Description

Also pinning all the OTEL versions to v1.6.0.

PR #2561, opened 2025-09-11
remove deprecated otel interceptors

vroldanbet:
## Description

This commit removes the use of deprecated OTEL Go SDK gRPC interceptors. Those were superfluous since we were already making use of the new StatsHandler installed in the gRPC server.

I also identified that OTel tracing was missing in the test server, so I added it.
I also noticed logs were not showing up because we had not correctly wired up the logger used by the middleware.

## Testing

I verified that traces are being correctly collected via otel using Jaeger, and that the TraceID shows up in the logs. Also tested the grpc Gateway, but did not verify the Dispatch API.

## References

Follow up to https://github.com/authzed/spicedb/pull/2556


internal/gateway/gateway.go
```diff
@@ -155,6 +155,6 @@ func OtelAnnotator(ctx context.Context, r *http.Request) metadata.MD {
 	metadataCopy := requestMetadata.Copy()
 
 	ctx = otel.GetTextMapPropagator().Extract(ctx, propagation.HeaderCarrier(r.Header))
-	otelgrpc.Inject(ctx, &metadataCopy, defaultOtelOpts...)
+	otelgrpc.Inject(ctx, &metadataCopy, defaultOtelOpts...) // nolint:staticcheck
```
tstirrat15, line 158: Why were these necessary?

vroldanbet: with the gRPC update, a bunch of things became deprecated, this is one of them. I didn't want to spend time fixing this.

PR #2564, opened 2025-09-11
perf: Add support in LookupSubjects for skipping caveats/expiration

josephschorr:
Should improve performance when expiration is enabled but unused

### DIFF
--- a/pkg/schema/definition_test.go
+++ b/pkg/schema/definition_test.go
@@ -1263,7 +1263,9 @@ func TestTypeSystemAccessors(t *testing.T) {
 			`use expiration
 
 			definition user {}
-			definition group {}
+			definition group {
+				relation member: user
+			}
 
 			caveat somecaveat(somecondition int) {
 				somecondition == 42
@@ -1274,6 +1276,12 @@ func TestTypeSystemAccessors(t *testing.T) {
 				relation caveated: user with somecaveat | group
 				relation expired: user with expiration | group
 				relation mixed: user | group with somecaveat and expiration
+				relation viewer: user | user:* with expiration
+				relation caveated_viewer: user | user:* with somecaveat
+				relation membered: group#member with somecaveat and expiration
+				relation membered_plain: group#member
+				relation caveated_membered: group#member with somecaveat | group#member
+				relation expired_membered: group#member with expiration | group#member
 			}`,
 			map[string]tsTester{
 				"resource": func(t *testing.T, vts *ValidatedDefinition) {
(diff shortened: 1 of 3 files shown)


pkg/schema/definition_test.go
```diff
@@ -720,6 +735,21 @@ func TestTypeSystemAccessors(t *testing.T) {
...
+					t.Run("PossibleTraitsForAnySubject", func(t *testing.T) {
+						traits, err := vts.PossibleTraitsForAnySubject("editor")
+						require.NoError(t, err)
+						require.False(t, traits.AllowsCaveats)
+						require.False(t, traits.AllowsExpiration)
+
+						traits, err = vts.PossibleTraitsForAnySubject("viewer")
+						require.NoError(t, err)
+						require.False(t, traits.AllowsCaveats)
+						require.False(t, traits.AllowsExpiration)
```
miparnisari, line 748: 1. if you write `relation viewer: user | user:* with expiration` this test still passes. 
2. Same if you do `relation onlycaveated: user:* with somecaveat` in the test `"schema with caveats"`
3. I don't see tests covering something like `relation viewer: group#member with expiration` or `relation viewer: group#member with caveat`?

internal/graph/lookupsubjects.go
```diff
@@ -100,24 +103,52 @@ func subjectsForConcreteIds(subjectIds []string) map[string]*v1.FoundSubjects {
 	return foundSubjects
 }
 
+func (cl *ConcurrentLookupSubjects) queryOptionsForRelation(
+	ctx context.Context,
+	ts *schema.TypeSystem,
+	namespaceName string,
+	relationName string,
+) ([]options.QueryOptionsOption, error) {
+	opts := make([]options.QueryOptionsOption, 0, 1)
```
miparnisari:
```suggestion
	opts := make([]options.QueryOptionsOption, 0, 3)
```

since at most you'll store 3 elements

pkg/schema/definition_test.go
josephschorr: Not sure how you found that, but its a bug in the compiler, not the type system. Good catch and I've added a fix to this PR.

pkg/schemadsl/compiler/translator.go
```diff
@@ -659,11 +659,18 @@ func translateSpecificTypeReference(tctx *translationContext, typeRefNode *dslNo
 			},
 		}
 
+		// Add the caveat(s), if any.
```
miparnisari: I think this function can be simplified to avoid the duplicated blocks:

```go
func translateSpecificTypeReference(tctx *translationContext, typeRefNode *dslNode) (*core.AllowedRelation, error) {
	typePath, err := typeRefNode.GetString(dslshape.NodeSpecificReferencePredicateType)
	if err != nil {
		return nil, typeRefNode.Errorf("invalid type name: %w", err)
	}

	nspath, err := tctx.prefixedPath(typePath)
	if err != nil {
		return nil, typeRefNode.Errorf("%w", err)
	}

	relationName := Ellipsis
	if typeRefNode.Has(dslshape.NodeSpecificReferencePredicateRelation) {
		relationName, err = typeRefNode.GetString(dslshape.NodeSpecificReferencePredicateRelation)
		if err != nil {
			return nil, typeRefNode.Errorf("invalid type relation: %w", err)
		}
	}
	ref := &core.AllowedRelation{
		Namespace: nspath,
		RelationOrWildcard: &core.AllowedRelation_Relation{
			Relation: relationName,
		},
	}

	if typeRefNode.Has(dslshape.NodeSpecificReferencePredicateWildcard) {
		ref.RelationOrWildcard = &core.AllowedRelation_PublicWildcard_{
			PublicWildcard: &core.AllowedRelation_PublicWildcard{},
		}
	}

	// Add the caveat(s), if any.
	err = addWithCaveats(tctx, typeRefNode, ref)
	if err != nil {
		return nil, typeRefNode.Errorf("invalid caveat: %w", err)
	}

	// Add the expiration trait, if any.
	err = addWithExpiration(tctx, typeRefNode, ref)
	if err != nil {
		return nil, typeRefNode.Errorf("invalid expiration: %w", err)
	}

	if !tctx.skipValidate {
		if err := ref.Validate(); err != nil {
			return nil, typeRefNode.Errorf("invalid type relation: %w", err)
		}
	}

	ref.SourcePosition = getSourcePosition(typeRefNode, tctx.mapper)
	return ref, nil
}
```

it passes the tests but please double-check :)

pkg/schema/definition_test.go
miparnisari: is (3) addressed?

josephschorr: Added

PR #2565, opened 2025-09-11
feat: Handle caveats and intersection arrows in the query plan

barakmich:
Dependent on #2553

### DIFF
--- a/pkg/query/simplify_caveat.go
+++ b/pkg/query/simplify_caveat.go
@@ -279,21 +279,26 @@ func simplifyNotOperation(
 		return nil, false, err
 	}
 
+	// Since we're simplifying the NOT operation, the results are basically flipped.
+
 	if simplified == nil && passes {
-		// Child evaluated to true unconditionally - NOT true is false
+		// Child caveat evaluated to true unconditionally - NOT true is false
 		return expr, false, nil // Return original expr to indicate what failed
-	} else if !passes {
-		// Child failed - NOT false is true
-		return nil, true, nil
-	} else {
-		// Child is conditional - NOT conditional is still conditional
-		return &core.CaveatExpression{
-			OperationOrCaveat: &core.CaveatExpression_Operation{
-				Operation: &core.CaveatOperation{
-					Op:       core.CaveatOperation_NOT,
-					Children: []*core.CaveatExpression{simplified},
-				},
-			},
-		}, true, nil
 	}
+
+	if !passes {
+		// Child caveat did NOT pass the caveat check, or simplified to impossible.
+		// Since the NOT of false is true, we DO pass the check, unconditionally.
+		return nil, true, nil
+	}
+
+	// Child is conditional - NOT conditional is still conditional
+	return &core.CaveatExpression{
+		OperationOrCaveat: &core.CaveatExpression_Operation{
+			Operation: &core.CaveatOperation{
+				Op:       core.CaveatOperation_NOT,
+				Children: []*core.CaveatExpression{simplified},
+			},
+		},
+	}, true, nil
 }
--- a/pkg/query/simplify_caveat_test.go
+++ b/pkg/query/simplify_caveat_test.go
@@ -1216,6 +1216,72 @@ func TestSimplifyWithEmptyContext(t *testing.T) {
 	require.NotNil(simplified.GetOperation())
 }
 
+func TestSimplifyNotConditional(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	require := require.New(t)
+
+	// Create test caveat: count < limit
+	env, err := caveats.EnvForVariablesWithDefaultTypeSet(map[string]caveattypes.VariableType{
+		"count": caveattypes.Default.UIntType,
+		"limit": caveattypes.Default.UIntType,
+	})
+	require.NoError(err)
+
+	limitCaveat, err := caveats.CompileCaveatWithName(env, "count < limit", "limit_check")
+	require.NoError(err)
+
+	serialized, err := limitCaveat.Serialize()
+	require.NoError(err)
+
+	caveatDef := &core.CaveatDefinition{
+		Name:                 "limit_check",
+		SerializedExpression: serialized,
+		ParameterTypes:       env.EncodedParametersTypes(),
+	}
+
+	// Create datastore with caveat
+	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	require.NoError(err)
+
+	revision, err := ds.ReadWriteTx(ctx, func(ctx context.Context, tx datastore.ReadWriteTransaction) error {
+		return tx.WriteCaveats(ctx, []*core.CaveatDefinition{caveatDef})
+	})
+	require.NoError(err)
+
+	reader := ds.SnapshotReader(revision)
+	runner := internalcaveats.NewCaveatRunner(caveattypes.Default.TypeSet)
+
+	// Create NOT expression: NOT limit_check(limit=10)
+	// Base case: when child is conditional (partial), NOT conditional should remain conditional
+	notExpr := internalcaveats.Invert(
+		&core.CaveatExpression{
+			OperationOrCaveat: &core.CaveatExpression_Caveat{
+				Caveat: &core.ContextualizedCaveat{
+					CaveatName: "limit_check",
+					Context:    mustToStruct(t, map[string]any{"limit": uint64(10)}),
+				},
+			},
+		},
+	)
+
+	// Test with partial/missing context - child caveat is conditional, so NOT should also be conditional
+	// We provide limit but not count, making the child caveat partial
+	partialContext := map[string]any{} // No count provided, only limit is in relationship context
+
+	simplified, passes, err := SimplifyCaveatExpression(ctx, runner, notExpr, partialContext, reader)
+	require.NoError(err)
+	require.True(passes, "NOT of conditional caveat should pass conditionally")
+	require.NotNil(simplified, "NOT of conditional caveat should remain as NOT expression")
+
+	// Verify the structure: should still be a NOT operation with the child caveat
+	require.NotNil(simplified.GetOperation(), "Result should be an operation")
+	require.Equal(core.CaveatOperation_NOT, simplified.GetOperation().Op, "Should be a NOT operation")
+	require.Len(simplified.GetOperation().Children, 1, "NOT should have one child")
+	require.NotNil(simplified.GetOperation().Children[0].GetCaveat(), "Child should be the original caveat")
+	require.Equal("limit_check", simplified.GetOperation().Children[0].GetCaveat().CaveatName)
+}
+
 func TestSimplifyDeeplyNestedCaveats(t *testing.T) {
 	ctx := context.Background()
 	require := require.New(t)


pkg/query/exclusion.go
```diff
@@ -20,8 +20,56 @@ func NewExclusion(mainSet, excluded Iterator) *Exclusion {
...
+	// Case 3: Main has no caveat, excluded has caveat
+	// Main applies unconditionally, excluded applies conditionally
+	// Result: main path with caveat NOT(excluded_caveat)
+	if mainPath.Caveat == nil && excludedPath.Caveat != nil {
+		// For now, return main with excluded caveat negated
+		// This is complex because we'd need to negate the caveat expression
+		// For simplicity, let's return the main path with the excluded caveat
+		// This represents "main applies, but only when excluded caveat is false"
+		result := *mainPath
+		result.Caveat = excludedPath.Caveat
```
tstirrat15: How does code represent negation of the caveat?

pkg/query/intersection.go
```diff
@@ -22,31 +21,104 @@ func (i *Intersection) addSubIterator(subIt Iterator) {
 }
 
 func (i *Intersection) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
+	ctx.TraceEnterIterator("Intersection", resources, subject)
+
 	validResources := resources
 
-	var paths []*Path
+	// Track paths by resource key for combining with AND logic
+	pathsByKey := make(map[string]*Path)
```
tstirrat15: curious because this isn't something that i understand very well - is there benefit to sizing the map at creation time? how does go do allocation for maps?

pkg/query/intersection.go
```diff
@@ -22,31 +21,104 @@ func (i *Intersection) addSubIterator(subIt Iterator) {
...
 		if err != nil {
 			return nil, err
 		}
 
+		ctx.TraceStep("Intersection", "sub-iterator %d returned %d paths", iterIdx, len(paths))
+
 		if len(paths) == 0 {
+			ctx.TraceStep("Intersection", "sub-iterator %d returned empty, short-circuiting", iterIdx)
+			var emptyPaths []*Path
+			ctx.TraceExitIterator("Intersection", emptyPaths)
```
tstirrat15: Why creating a new object here rather than passing `paths`? Is it because that's a reference that could hang around and then get mutated later in some way?

pkg/query/intersection.go
```diff
@@ -22,31 +21,104 @@ func (i *Intersection) addSubIterator(subIt Iterator) {
...
+			for _, path := range paths {
+				key := path.Resource.ObjectType + ":" + path.Resource.ObjectID
+				if existing, exists := pathsByKey[key]; !exists {
+					pathsByKey[key] = path
+				} else {
+					// If multiple paths for same endpoint in first iterator, merge with OR
+					if err := existing.MergeOr(path); err != nil {
+						return nil, err
+					}
+				}
```
tstirrat15, line 61: This logic is handling when you have multiple distinct paths between a resource and a subject that may have different caveat contexts? Is that why we're talking about merges here?

pkg/query/intersection_test.go
```diff
@@ -285,3 +299,220 @@ func TestIntersectionIteratorEarlyTermination(t *testing.T) {
...
+
+		relSeq, err := ctx.Check(intersect, NewObjects("document", "doc1"), NewObject("user", "alice").WithEllipses())
+		require.NoError(err)
+
+		rels, err := CollectAll(relSeq)
+		require.NoError(err)
+
+		require.Len(rels, 1, "Intersection should return one combined relation")
+		require.NotNil(rels[0].Caveat, "Result should have combined caveat")
+		// The combination logic should combine both caveats with AND logic
```
tstirrat15: Is this a missing assertion?

pkg/query/intersection_test.go
```diff
@@ -285,3 +299,220 @@ func TestIntersectionIteratorEarlyTermination(t *testing.T) {
...
+		}
+
+		pathEditor := MustPathFromString("document:doc1#editor@user:alice")
+		pathEditor.Caveat = &core.CaveatExpression{
+			OperationOrCaveat: &core.CaveatExpression_Caveat{
+				Caveat: &core.ContextualizedCaveat{
+					CaveatName: "caveat2",
+				},
+			},
+		}
```
tstirrat15, line 411: Should one of these be missing the caveat?

pkg/query/intersection_test.go
```diff
@@ -285,3 +299,220 @@ func TestIntersectionIteratorEarlyTermination(t *testing.T) {
...
+
+		relSeq, err := ctx.Check(intersect, NewObjects("document", "doc1"), NewObject("user", "alice").WithEllipses())
+		require.NoError(err)
+
+		rels, err := CollectAll(relSeq)
+		require.NoError(err)
+
+		require.Len(rels, 1, "Should return one intersected path")
+		require.NotNil(rels[0].Caveat, "Final result should have combined caveat")
+		// The final caveat should be some combination of caveat1 and caveat2 (implementation detail)
```
tstirrat15: I'm confused on this point - if one of the potential paths exists and doesn't have a caveat, can't you ignore the caveated paths?

pkg/query/arrow_test.go
```diff
@@ -225,3 +227,233 @@ func TestArrowIteratorMultipleResources(t *testing.T) {
...
+
+		relSeq, err := ctx.Check(arrow, NewObjects("document", "doc1"), NewObject("user", "alice").WithEllipses())
+		require.NoError(err)
+
+		rels, err := CollectAll(relSeq)
+		require.NoError(err)
+
+		require.Len(rels, 1, "Arrow should return one relation")
+		require.NotNil(rels[0].Caveat, "Left caveat should be preserved")
+		// Note: Checking exact caveat content would require parsing the CaveatExpression
```
tstirrat15, line 313: Is this future work? Is it stuff that we have the tools for and it'd just be painful, or is this a project unto itself?

pkg/query/path.go
```diff
@@ -206,6 +206,72 @@ func MustPathFromString(relationshipStr string) *Path {
...
+	}
+
+	return p.Resource.ObjectType == other.Resource.ObjectType &&
+		p.Resource.ObjectID == other.Resource.ObjectID &&
+		p.Subject.ObjectType == other.Subject.ObjectType &&
+		p.Subject.ObjectID == other.Subject.ObjectID
+}
+
+// Equals checks if two paths are fully equal (all fields match)
+func (p *Path) Equals(other *Path) bool {
```
tstirrat15: Oof. Are there libs for this kind of thing? I'm envisioning something that lets you implement something like a python dunder but that has some kind of default for fieldwise comparison.

pkg/query/path_test.go
```diff
@@ -651,3 +651,405 @@ func TestPath_ConversionRoundtrip(t *testing.T) {
...
+		path1 := &Path{
+			Resource: NewObject("document", "doc1"),
+			Relation: "viewer", // Relations don't matter for endpoint comparison
+			Subject:  NewObjectAndRelation("alice", "user", "..."),
+		}
+
+		path2 := &Path{
+			Resource: NewObject("document", "doc1"),
+			Relation: "editor",                                        // Different relation but same endpoints
+			Subject:  NewObjectAndRelation("alice", "user", "member"), // Different subject relation but same endpoint
```
tstirrat15: The subject relation being different but being the same endpoints is kinda surprising to me - it feels like a subject relation is pointing off and away, and so it shouldn't compare as equal, but I'm assuming that's because of a gap in understanding.

pkg/query/union_test.go
```diff
@@ -461,3 +458,179 @@ func TestUnionDeduplicationBugFix(t *testing.T) {
...
+
+		relSeq, err := ctx.Check(union, NewObjects("document", "doc1"), NewObject("user", "alice").WithEllipses())
+		require.NoError(err)
+
+		rels, err := CollectAll(relSeq)
+		require.NoError(err)
+
+		require.Len(rels, 1, "Union should deduplicate to one relation")
+		require.NotNil(rels[0].Caveat, "Result should have combined caveat")
+		// The combination logic should preserve one of the caveats (implementation detail)
```
tstirrat15: Because it's evaluating one of the caveats in the process? Otherwise wouldn't you need to preserve both caveats, because one (or more) could evaluate to false and the overall expression could still be true?

tstirrat15 (commented):
See comments

pkg/query/datastore.go
```diff
@@ -57,8 +57,18 @@ func (r *RelationIterator) buildSubjectRelationFilter() datastore.SubjectRelatio
...
+	// For subrelations, we need to allow type mismatches because the subrelation might bridge different types
+	// For example, group:member -> group:member should find group:everyone#member@group:engineering#member
+	// and then that relationship should be used by the Arrow to check group:engineering#member for user subjects
+	// However, wildcard relations should always enforce strict type checking
+	if subject.ObjectType != r.base.Type && r.base.Subrelation != "" && !r.base.Wildcard {
+		// For non-wildcard subrelations, we proceed with the query even if types don't match
+		// This allows finding intermediate relationships that bridge type gaps
+		ctx.TraceStep(r, "subject type %s doesn't match base type %s, but proceeding due to subrelation %s",
+			subject.ObjectType, r.base.Type, r.base.Subrelation)
+	} else if subject.ObjectType != r.base.Type {
```
tstirrat15: I take it this was a finding in the conformance tests? It's a bummer not to be able to make this optimization... is this related to the conversation about being able to put a permission as a subrelation?

pkg/query/datastore_test.go
```diff
@@ -237,13 +237,14 @@ func TestRelationIteratorSubjectTypeMismatchScenarios(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
-			baseRel := createTestBaseRelation("document", "viewer", tc.expectedSubjectType, tuple.Ellipsis)
+			// Use empty string instead of Ellipsis to test pure type mismatch without subrelation bridging
+			baseRel := createTestBaseRelation("document", "viewer", tc.expectedSubjectType, "")
```
tstirrat15, line 197: Is a non-ellipsis relation technically an allowed variant, or is this something that we're doing for the sake of testing to exercise a particular branch?

pkg/query/intersection_arrow.go
```diff
@@ -0,0 +1,223 @@
...
+			satisfiedCount := 0
+
+			for path, err := range subit {
+				if err != nil {
+					yield(nil, err)
+					return
+				}
+
+				// This is a valid left subject (passed its own caveats)
+				leftSubjectCount++
```
tstirrat15: There's some part of my brain that wants this to be like prometheus metrics where you register and then increment a counter through method calls rather than tracking a local variable. Does that sound useful? I think it's mostly an aesthetic preference on my side so feel free to ignore.

pkg/query/intersection_arrow.go
```diff
@@ -0,0 +1,223 @@
...
+				}
+
+				checkResources := []Object{GetObject(path.Subject)}
+				checkit, err := ctx.Check(ia.right, checkResources, subject)
+				if err != nil {
+					yield(nil, err)
+					return
+				}
+
+				// Check if this left subject satisfies the right condition
```
tstirrat15: "the right condition" being the satisfaction of the relevant relation/permission?

pkg/query/intersection_arrow.go
```diff
@@ -0,0 +1,223 @@
...
+							leftExpr = caveatExpr
+						} else {
+							leftExpr = caveats.And(leftExpr, caveatExpr)
+						}
+					}
+					combinedLeftCaveat = leftExpr
+				}
+
+				// For intersection semantics, we need to create a single path that represents
+				// the AND of all individual conditions: each (leftCaveat AND rightCaveat) must be AND'd together
```
tstirrat15: Curious - isn't this commutative + associative so we don't really care how we're assembling the concatenation of all of these caveats? Or is there something there that I'm missing?

pkg/query/intersection_test.go
tstirrat15: Nah, looks like this test is testing what it should be.

pkg/query/intersection_test.go
tstirrat15: Still curious on this point.

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,303 @@
...
+// Query context takes precedence over relationship context
+func mergeContexts(expr *core.CaveatExpression, queryContext map[string]any) map[string]any {
+	fullContext := make(map[string]any)
+
+	// Start with the relationship context from the caveat expression
+	if expr.GetCaveat() != nil && expr.GetCaveat().Context != nil {
+		for k, v := range expr.GetCaveat().Context.AsMap() {
+			fullContext[k] = v
+		}
+	}
```
tstirrat15, line 21: It's wild to me that golang doesn't have some mechanism for merging two maps or concatenating two sequences so that you can use `maps.Collect`

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,303 @@
...
+	if expr == nil {
+		return
+	}
+
+	// If this is a leaf caveat, collect its context
+	if expr.GetCaveat() != nil && expr.GetCaveat().Context != nil {
+		for k, v := range expr.GetCaveat().Context.AsMap() {
+			if _, exists := contextMap[k]; !exists {
+				contextMap[k] = v
+			}
```
tstirrat15, line 54: When you pass a map, are you passing it by value or by reference? Like does this actually mutate the provided map, or by virtue of passing it are you operating on a copy?

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,303 @@
...
+	// Handle operation nodes
+	cop := expr.GetOperation()
+	if cop == nil {
+		return expr, true, nil // Unknown operation, treat as conditional
+	}
+
+	return simplifyOperation(ctx, runner, expr, cop, context, reader)
+}
+
+func evaluateLeaf(
```
tstirrat15, line 109: Can we either name the returns or add a docstring that explains what the returns are?

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,303 @@
...
+	reader datastore.CaveatReader,
+) (*core.CaveatExpression, bool, error) {
+	if err := runner.PopulateCaveatDefinitionsForExpr(ctx, expr, reader); err != nil {
+		return nil, false, err
+	}
+
+	return simplifyCaveatExpressionInternal(ctx, runner, expr, context, reader)
+}
+
+func simplifyCaveatExpressionInternal(
```
tstirrat15, line 88: Same here

pkg/query/simplify_caveat_test.go
```diff
@@ -0,0 +1,1130 @@
...
+		context := map[string]any{"a": int64(5), "b": int64(15)}
+
+		simplified, passes, err := SimplifyCaveatExpression(ctx, runner, orExpr, context, reader)
+		require.NoError(err)
+		require.NotNil(simplified)
+		require.False(passes) // should not pass
+
+		// Should return original expression since all are false
+		require.NotNil(simplified.GetOperation())
+		require.Equal(core.CaveatOperation_OR, simplified.GetOperation().Op)
```
tstirrat15, line 299: This surprises me a little - if both caveats are false, wouldn't it simplify to returning false?

tstirrat15: Or is the idea that we've failed all caveat assertions, so we retain the context of all caveat assertions that failed?

coverage.out
tstirrat15, line 1: Does this get committed? We should probably add an ignore if not.

pkg/query/caveat_test.go
```diff
@@ -0,0 +1,580 @@
...
+				}
+			}
+
+			if tc.name == "no caveat iterator allows all paths" {
+				require.NoError(t, err)
+				actualPaths, err := CollectAll(seq)
+				require.NoError(t, err)
+				require.Len(t, actualPaths, len(expectedMatchingPaths), "Expected all matching paths to be returned")
+			} else if tc.name == "caveat iterator filters out paths without matching caveat" {
+				// Tests that attempt caveat evaluation should fail since we don't have caveat definitions
```
tstirrat15: Style thing, but for the number of tests here and the fact that we're switching on `tc.name` i'd rather just have separate test cases

pkg/query/caveat.go
tstirrat15, line 1: Is the general idea with the caveat iterator that you can choose when to apply it because it can apply to any type/level of iterator?

pkg/query/caveat.go
```diff
@@ -0,0 +1,418 @@
...
+	subSeq, err := ctx.Check(c.subiterator, resources, subject)
+	if err != nil {
+		return nil, err
+	}
+
+	return func(yield func(*Path, error) bool) {
+		caveatName := "none"
+		if c.caveat != nil {
+			caveatName = c.caveat.CaveatName
+		}
```
tstirrat15: I wish Go had ternary syntax.

pkg/query/caveat.go
```diff
@@ -0,0 +1,418 @@
...
+			if !passes {
+				// Caveat evaluated to false - don't yield the path
+				ctx.TraceStep(c, "path failed caveat evaluation")
+				continue
+			}
+
+			passedCount++
+
+			// Create modified path with simplified caveat
+			modifiedPath := *path // Copy the path
```
tstirrat15: Is a part of the framework that we retain the original paths as we iterate/pass over? I'm thinking that immutability would be lovely from a debugging standpoint, though I suppose it would come with a cost.

pkg/query/caveat.go
```diff
@@ -0,0 +1,418 @@
...
+				}
+
+				// Common real-world patterns
+				if (expected == "membercaveat" && actual == "teamcaveat") ||
+					(expected == "teamcaveat" && actual == "membercaveat") {
+					return true
+				}
+
+				// Add more patterns as needed
+				return false
```
tstirrat15: How big of an optimization is this? This feels like it would be magical behavior from the perspective of a caller and that makes me uneasy. I also don't really like having test-specific behavior in the runtime code if I can avoid it.

tstirrat15: Or was this ported from existing code?

tstirrat15 (commented):
See comments - I have a few questions/comments/concerns

pkg/query/datastore.go
barakmich: Yeah,  it is; and fair -- this could be an optimization by just removing the part at query time. On the other hand, it's less modification at query time. A balance we need to figure out.

pkg/query/caveat.go
barakmich: Yep! That's the plan. Because then an optimizer can be something like "CaveatPushdown"

pkg/schema/v2/operations.go
```diff
@@ -27,15 +27,31 @@ type IntersectionOperation struct {
...
 }
 
-// IntersectionOperation is an Operation that represents `permission foo = a - b`.
+// ExclusionOperation is an Operation that represents `permission foo = a - b`.
 type ExclusionOperation struct {
 	Left  Operation
 	Right Operation
 }
 
+// FunctionedTuplesetOperation is an Operation that represents functioned tuplesets like `permission foo = relation.any(other)` or `permission foo = relation.all(other)`.
```
miparnisari, line 79: What are the pros & cons of having one struct to represent two concepts?

pkg/query/context.go
```diff
@@ -2,20 +2,241 @@ package query
...
+// NewTraceLogger creates a new trace logger
+func NewTraceLogger() *TraceLogger {
+	return &TraceLogger{
+		traces: make([]string, 0),
+		depth:  0,
+		stack:  make([]Iterator, 0),
+	}
+}
+
+// EnterIteratorOld logs entering an iterator
```
miparnisari: This function isn't used except in tests? 🤔

pkg/query/caveat.go
```diff
@@ -0,0 +1,309 @@
...
+// NewCaveatIterator creates a new caveat iterator that wraps the given subiterator
+// and applies the specified caveat conditions.
+func NewCaveatIterator(subiterator Iterator, caveat *core.ContextualizedCaveat) *CaveatIterator {
+	return &CaveatIterator{
+		subiterator: subiterator,
+		caveat:      caveat,
+	}
+}
+
+func (c *CaveatIterator) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
```
miparnisari, line 28: The implementation of `CheckImpl`, `IterSubjectsImpl` and `IterResourcesImp` seem to be exactly the same, except for the very first line. Can we remove the duplication?

pkg/query/simplify_caveat.go
miparnisari: It mutates the map. e.g. https://go.dev/play/p/J4eaapSA2rg

pkg/query/caveat_test.go
miparnisari: +1, this test is very hard to read with lots of nested `if`s

pkg/query/caveat.go
```diff
@@ -0,0 +1,309 @@
...
+	// Use the SimplifyCaveatExpression function to properly handle AND/OR logic
+	simplified, passes, err := SimplifyCaveatExpression(
+		ctx,
+		ctx.CaveatRunner,
+		path.Caveat,
+		ctx.CaveatContext,
+		reader,
+	)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to evaluate caveat: %w", err)
```
miparnisari:
```suggestion
		return nil, false, fmt.Errorf("failed to simplify caveat: %w", err)
```

pkg/query/caveat.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// buildExplainInfo creates detailed explanation information for the caveat iterator
+func (c *CaveatIterator) buildExplainInfo() string {
+	if c.caveat == nil {
+		return "Caveat(none)"
+	}
+
+	// Build basic caveat information
+	info := fmt.Sprintf("Caveat(%s", c.caveat.CaveatName)
```
miparnisari, line 203: FYI, using `strings.Builder` is faster when you don't special need formatting. I'll add a linter :)

pkg/query/path.go
```diff
@@ -206,6 +206,72 @@ func MustPathFromString(relationshipStr string) *Path {
 	return FromRelationship(rel)
 }
 
+// EqualsEndpoints checks if two paths have the same Resource and Subject endpoints (types and IDs only)
```
miparnisari, line 227: can you clarify in godoc why you are not comparing `p.Subject.Relation`? `group:1#member` is a different set than `group:1#admin`

pkg/query/path.go
miparnisari: `reflect.DeepEqual` comes to mind. It passes the tests but we may need a benchmark

pkg/query/intersection_arrow_test.go
```diff
@@ -0,0 +1,503 @@
...
+		intersectionArrow := NewIntersectionArrow(leftIter, rightIter)
+
+		// Create test context
+		ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+		require.NoError(err)
+
+		revision, err := ds.ReadWriteTx(context.Background(), func(ctx context.Context, tx datastore.ReadWriteTransaction) error {
+			return nil
+		})
+		require.NoError(err)
```
miparnisari: What is this revision for? If you comment it out, the test still passes

pkg/query/intersection_arrow_test.go
```diff
@@ -0,0 +1,503 @@
...
+	require.NoError(err)
+
+	ctx := &Context{
+		Context:   context.Background(),
+		Executor:  LocalExecutor{},
+		Datastore: ds,
+		Revision:  revision,
+	}
+
+	t.Run("CombineTwoCaveats_AND_Logic", func(t *testing.T) {
```
miparnisari, line 318: Can you add a test for the scenario where `leftIter` and `rightIter` have more than one paths?

pkg/query/intersection_arrow_test.go
```diff
@@ -0,0 +1,503 @@
...
+		require.NoError(err)
+
+		paths, err := CollectAll(pathSeq)
+		require.NoError(err)
+
+		// Should return a single result since alice is in ALL teams (intersection semantics)
+		require.Len(paths, 1, "Should return single path representing the intersection")
+
+		// Verify the result has alice as the subject
+		require.Equal("alice", paths[0].Subject.ObjectID, "Result should have alice as the subject")
```
miparnisari: if you print `paths[0].Relation` you get `team`, which seems arbitrary (why not `member`?). should this not be an empty string?

pkg/query/simplify_caveat_test.go
```diff
@@ -0,0 +1,1130 @@
...
+						OperationOrCaveat: &core.CaveatExpression_Caveat{
+							Caveat: &core.ContextualizedCaveat{CaveatName: "caveat2"},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	t.Run("BothTrue_ReturnsNil", func(t *testing.T) {
```
miparnisari, line 160: please add a test for both AND and OR with more than 2 caveats

pkg/query/simplify_caveat_test.go
```diff
@@ -0,0 +1,1130 @@
...
+		// count=3: NOT (3 < 4) = NOT true = false
+		context := map[string]any{"count": uint64(3)}
+
+		simplified, passes, err := SimplifyCaveatExpression(ctx, runner, notExpr, context, reader)
+		require.NoError(err)
+		require.False(passes, "Expected NOT expression to fail with count=3 (since 3 < 4 is true, NOT true = false)")
+		require.NotNil(simplified, "Expected simplified expression to contain the failed NOT expression")
+	})
+
+	t.Run("Count5_ShouldPassNegation", func(t *testing.T) {
```
miparnisari, line 648: should we have a test case where `context` is empty?

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,303 @@
...
+	ctx context.Context,
+	runner *caveats.CaveatRunner,
+	expr *core.CaveatExpression,
+	cop *core.CaveatOperation,
+	context map[string]any,
+	reader datastore.CaveatReader,
+) (*core.CaveatExpression, bool, error) {
+	if len(cop.Children) != 1 {
+		// NOT should have exactly one child
+		return expr, true, nil
```
miparnisari: this should be an error no?

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,303 @@
...
+// SimplifyCaveatExpression simplifies a caveat expression by applying AND/OR logic:
+// - For AND: if a caveat evaluates to true, remove it from the expression
+// - For OR: if a caveat evaluates to true, the entire expression becomes true
+// Returns:
+//   - simplified: the simplified expression (nil if unconditionally true)
+//   - passes: true if passes unconditionally or conditionally, false if fails
+//   - error: any error that occurred during simplification
+func SimplifyCaveatExpression(
+	ctx context.Context,
+	runner *caveats.CaveatRunner,
```
miparnisari, line 76: super tiny nitpick but your godoc says "if a caveat evaluates to..", but here you have a `runner`. I suggest renaming `runner->evaluator` for more consistency

pkg/schema/v2/operations.go
barakmich: Mostly, it's trying to follow the schema language. We allow these sorts of function calls, and may allow more -- for the moment, there's only the two. If the function calls are static (which they currently are) and never grow then yeah, might as well make two of them. If the set of function calls ever expands (which the proto is set up for) then this is an operation meaning "function call"

Practically, it's a new coat of paint on the proto as source of truth, so I just followed that structure.

pkg/query/context.go
barakmich: Yeah, this was just leftover unused code from another refactor in the chain. Removed.

pkg/query/caveat.go
barakmich: This doesn't need a linter, I don't think. It's meant for debug printing and maybe a human reads it as part of an explicit EXPLAIN-style command. We can spend a few extra cycles here and not constantly fight a linter.

pkg/query/intersection_arrow_test.go
barakmich: I think that's because the `revision` (set in the context, below) works out as being zero if you do. Fixed to be HeadRevision.

pkg/query/simplify_caveat.go
barakmich: Maybe -- but caveats use a CaveatRunner and that makes the most sense to use. So described it a little better in the godoc.

pkg/query/path.go
tstirrat15: Hmm yeah that's probably the semantic we'd want but also reflection is generally expensive?

pkg/query/intersection_test.go
barakmich: No, that caveat still needs to exist. Because it's intersection, it needs to satisfy all the underlying paths, so if A has a caveat and B does not, and the permission is `permission foo = a & b` then foo is only valid if we pass the caveat (because it's required for `a`).

If there are multiple solutions to A's caveat, those will be taken into account for the subtree for `a` (usually, through a union with `relation bar: type | type with caveat`) -- and union _should_ ignore the caveat, for the reason you're asking.

coverage.out
miparnisari: @barakmich please add an entry in gitignore :)

pkg/query/intersection_test.go
barakmich: Or to put it pithy:
```
true AND maybe = maybe
true OR maybe = true
```

pkg/query/path_test.go
barakmich: Yeah, that was a straight-up mistake; fixed already

pkg/query/union_test.go
barakmich: Extending this test (and others) to exemplify the combine process

pkg/query/arrow_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	corev1 "github.com/authzed/spicedb/pkg/proto/core/v1"
```
miparnisari: same import twice

tstirrat15 (dismissed):
LGTM

pkg/query/arrow_test.go
miparnisari: The test above this one has assertions for the contents of the caveat  🤔

pkg/query/path.go
miparnisari: I think it's fine to leave as-is for now

tstirrat15 (approved):
LGTM

pkg/query/caveat.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	return false
+}
+
+func (c *CaveatIterator) Clone() Iterator {
+	return &CaveatIterator{
+		subiterator: c.subiterator.Clone(),
+		caveat:      c.caveat, // ContextualizedCaveat is immutable, safe to share
```
miparnisari: "ContextualizedCaveat is immutable, safe to share" - no it's not? `c.caveat` is a pointer

pkg/query/caveat.go
```diff
@@ -0,0 +1,218 @@
...
+
+		processedCount := 0
+		passedCount := 0
+
+		for path, err := range subSeq {
+			if err != nil {
+				if !yield(path, err) {
+					return
+				}
+				continue
```
miparnisari: I am confused by the `continue` here and in the error case below - should it not be `return` instead?

pkg/query/testutil.go
```diff
@@ -229,6 +232,30 @@ func NewConflictingPermissionsFixedIterator() *FixedIterator {
 	return NewFixedIterator(paths...)
 }
 
+// AssertRelationsMatchByResource compares two slices of relations by their resource and subject,
+// ignoring the specific relation names. This is useful for testing intersection behavior where
+// we care about which resources have matching subjects, not the specific relation types.
+func AssertRelationsMatchByResource(t *testing.T, expected, actual []tuple.Relationship, msgAndArgs ...any) {
```
miparnisari: this function is unused. also, i don't think its implementation is correct (you are not including `rel.Subject.Relation` in the key?)

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,299 @@
...
+
+	child := cop.Children[0]
+	simplified, passes, err := simplifyCaveatExpressionInternal(ctx, runner, child, context, reader)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if simplified == nil && passes {
+		// Child evaluated to true unconditionally - NOT true is false
+		return expr, false, nil // Return original expr to indicate what failed
```
miparnisari, line 286: "Return original expr to indicate what failed" huh? nothing failed here

pkg/query/simplify_caveat.go
```diff
@@ -0,0 +1,299 @@
...
+	} else if !passes {
+		// Child failed - NOT false is true
+		return nil, true, nil
+	} else {
```
miparnisari: this `else` may run `if simplified != nil || passes` right? in which case, if `simplified != nil ` this code is correct, but if `simplified != nil && passes` should we not be returning `false`? 

i also see that this `else` isn't covered by unit tests, can you add some?

PR #2569, opened 2025-09-15
refactor: switch query plans to use a native `query.Path` type instead of the storage-oriented `tuple.Relationship` type

barakmich:


pkg/query/alias.go
```diff
@@ -43,96 +40,96 @@ func (a *Alias) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
-// rewriteRelation rewrites the Resource's Relation field to the alias relation
-func (a *Alias) rewriteRelation(rel Relation) Relation {
-	rewritten := rel
-	rewritten.Resource.Relation = a.relation
-	return rewritten
+// rewriteRelation rewrites the Path's Relation field to the alias relation
+func (a *Alias) rewriteRelation(path *Path) *Path {
+	rewritten := *path
+	rewritten.Relation = a.relation
+	return &rewritten
```
tstirrat15: Curious about the difference between these two (golang knowledge check for me): I know that in the first one, reassignment creates a copy. Is that the case when you're working with a pointer? Or is it that you're dereffing, which means that you're then creating a copy of the thing?

pkg/query/path.go
```diff
@@ -0,0 +1,215 @@
...
+
+	"github.com/authzed/spicedb/internal/caveats"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+// PathSeq is the intermediate iter closure that any of the planning calls return.
+type PathSeq iter.Seq2[*Path, error]
+
+// Path is an abstract notion of an individual relation. While tuple.Relation is what is stored under the hood,
```
tstirrat15, line 22: Incomplete comment

pkg/query/path.go
```diff
@@ -0,0 +1,215 @@
...
+type Path struct {
+	Resource   Object
+	Relation   string
+	Subject    ObjectAndRelation
```
tstirrat15, line 30: This is an object and relation because of subject relations?

pkg/query/path.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	// Combine caveats based on merge operation
+	switch mergeOp {
+	case pathMergeOpOr:
+		if p.Caveat != nil {
+			if other.Caveat == nil {
+				p.Caveat = nil
+			} else {
+				p.Caveat = caveats.Or(p.Caveat, other.Caveat)
```
tstirrat15: This is the caveat combination that we were talking about?

pkg/query/path.go
```diff
@@ -0,0 +1,215 @@
...
+	pathMergeOpAnd
+	pathMergeOpAndNot
+)
+
+func (p *Path) ResourceOAR() ObjectAndRelation {
+	return p.Resource.WithRelation(p.Relation)
+}
+
+func (p *Path) MergeOr(other *Path) error {
+	return p.mergeFrom(other, pathMergeOpOr)
```
tstirrat15: This is a style nit, but I like passing functions here rather than a flag that goes to a case statement (unless there's a performance reason not to do it). This is very much not a hill that I will die on, though.

pkg/query/union.go
```diff
@@ -22,56 +22,56 @@ func (u *Union) addSubIterator(subIt Iterator) {
...
-			seen[key] = rel
+			seen[key] = path
 		} else {
-			// If we already have a relationship for this resource,
-			// prefer one without caveats (cleaner permission grant)
-			if rel.OptionalCaveat == nil && existing.OptionalCaveat != nil {
-				seen[key] = rel
+			// If we already have a path for this resource,
+			// merge it with the new one using OR semantics
+			if err := existing.MergeOr(path); err != nil {
```
tstirrat15, line 52: Which direction does this merge? Does `existing` get used after this line?

pkg/query/alias.go
barakmich: Deref creates a copy, yep; so rewritten is a full copy, and then we return that reference. Though this can probably just go away, now that I think about it

pkg/query/path.go
barakmich: Yep! Trying to make explicit this difference

pkg/query/path.go
barakmich: Yep; exactly so

pkg/query/path.go
barakmich: Fair; that may make sense. Let's do it that way.

pkg/query/union.go
barakmich: it merges into `existing`, a la a proto. So `existing` stays in the map the whole time.

pkg/query/alias.go
```diff
@@ -43,98 +39,91 @@ func (a *Alias) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
 					if err != nil {
-						yield(rel, err)
+						yield(nil, err)
 						return
 					}
 
-					rewrittenRel := a.rewriteRelation(rel)
-					if !yield(rewrittenRel, nil) {
+					path.Relation = a.relation
+					if !yield(path, nil) {
```
miparnisari, line 56: Can we add a test that covers this block?

pkg/query/path.go
```diff
@@ -0,0 +1,213 @@
...
+func (p *Path) ResourceOAR() ObjectAndRelation {
+	return p.Resource.WithRelation(p.Relation)
+}
+
+// MergeOr combines the paths, ORing the caveats and expiration and metadata together.
+func (p *Path) MergeOr(other *Path) error {
+	return p.mergeFrom(other, func(pCaveat, otherCaveat *core.CaveatExpression) *core.CaveatExpression {
+		if pCaveat != nil {
+			if otherCaveat == nil {
+				return nil
```
miparnisari: why is this `return nil` and not `return pCaveat`?

pkg/query/path.go
```diff
@@ -0,0 +1,213 @@
...
+			} else {
+				return caveats.Or(pCaveat, otherCaveat)
+			}
+		}
+		return pCaveat
```
miparnisari: why this `return pCaveat` and not `return otherCaveat`?

pkg/query/fixed.go
```diff
@@ -36,25 +36,25 @@ func (f *FixedIterator) CheckImpl(ctx *Context, resources []Object, subject Obje
...
-	return func(yield func(Relation, error) bool) {
-		for _, rel := range f.rels {
-			// Check if the relation's subject matches the requested subject
-			if rel.Subject.ObjectID == subject.ObjectID && rel.Subject.ObjectType == subject.ObjectType && rel.Subject.Relation == subject.Relation {
-				if !yield(rel, nil) {
+func (f *FixedIterator) IterResourcesImpl(ctx *Context, subject ObjectAndRelation) (PathSeq, error) {
+	return func(yield func(*Path, error) bool) {
+		for _, path := range f.paths {
+			// Check if the path's subject matches the requested subject
+			if path.Subject.ObjectID == subject.ObjectID && path.Subject.ObjectType == subject.ObjectType && path.Subject.Relation == subject.Relation {
```
miparnisari, line 56: i'm surprised there is no `Equals()` defined for `tuple.ObjectAndRelation` 😆

pkg/query/intersection_test.go
```diff
@@ -77,11 +60,11 @@ func TestIntersectionIterator(t *testing.T) {
 		intersect.addSubIterator(singleUser)
 
 		// Use a subject that doesn't exist in both
-		relSeq, err := ctx.Check(intersect, NewObjects("document", "doc1"), NewObject("user", "bob").WithEllipses())
+		pathSeq, err := ctx.Check(intersect, NewObjects("document", "doc1"), NewObject("user", "bob").WithEllipses())
 		require.NoError(err)
 
-		if relSeq != nil {
-			_, err := CollectAll(relSeq)
+		if pathSeq != nil {
```
miparnisari, line 66: unrelated to this PR, but here and elsewhere, if `pathSeq == nil` we should be failing the test no?

pkg/query/intersection_test.go
```diff
@@ -291,12 +274,12 @@ func TestIntersectionIteratorEarlyTermination(t *testing.T) {
...
 	require.NoError(err)
 
 	// Should return empty results since first iterator has no results
-	if relSeq != nil {
-		rels, err := CollectAll(relSeq)
+	if pathSeq != nil {
+		rels, err := CollectAll(pathSeq)
 		require.NoError(err)
 		// Should be empty due to early termination
 		require.Empty(rels, "Early termination should return no results")
```
miparnisari, line 285: unrelated to this PR: this test doesn't really prove early termination; for that, you'd need to have, for example,`MockFixedIterators` that counts how many times we called `CheckImpl` on them, and then assert that the second iterator had zero calls

or maybe use your existing `FaultyIterator` as a second iterator?

pkg/query/intersection_test.go
```diff
@@ -77,11 +60,11 @@ func TestIntersectionIterator(t *testing.T) {
...
-		relSeq, err := ctx.Check(intersect, NewObjects("document", "doc1"), NewObject("user", "bob").WithEllipses())
+		pathSeq, err := ctx.Check(intersect, NewObjects("document", "doc1"), NewObject("user", "bob").WithEllipses())
 		require.NoError(err)
 
-		if relSeq != nil {
-			_, err := CollectAll(relSeq)
+		if pathSeq != nil {
+			_, err := CollectAll(pathSeq)
 			require.NoError(err)
 			// Should likely be empty since bob isn't in documentAccess
```
miparnisari: "should likely be empty"? why "likely", why not assert?

pkg/query/union_test.go
```diff
@@ -101,12 +101,12 @@ func TestUnionIterator(t *testing.T) {
...
 		require.NoError(err)
 
-		rels, err := CollectAll(relSeq)
+		paths, err := CollectAll(pathSeq)
 		require.NoError(err)
-		require.Empty(rels, "empty resource list should return no results")
+		require.Empty(paths, "empty resource list should return no results")
 	})
 
 	t.Run("Check_EarlyTermination", func(t *testing.T) {
```
miparnisari, line 112: unrelated to this PR. I don't understand this test, where is the early termination happening inside the union iterator? there can't be early termination in a union, right? you need to yield them all

pkg/query/union_test.go
```diff
@@ -33,22 +33,22 @@ func TestUnionIterator(t *testing.T) {
...
-		require.Len(rels, 2, "Union should deduplicate to one relation per resource")
+		require.Len(paths, 2, "Union should deduplicate to one relation per resource")
 
 		// Check that we have relations for both doc1 and doc2
 		resourceIDs := make(map[string]bool)
-		for _, rel := range rels {
+		for _, rel := range paths {
 			require.Equal("alice", rel.Subject.ObjectID, "All relations should be for alice")
 			require.Equal("document", rel.Resource.ObjectType, "All relations should be for documents")
 			resourceIDs[rel.Resource.ObjectID] = true
```
miparnisari:
```suggestion
		for _, path := range paths {
			require.Equal("alice", rel.Subject.ObjectID, "All paths should be for alice")
			require.Equal("document", rel.Resource.ObjectType, "All paths should be for documents")
			resourceIDs[rel.Resource.ObjectID] = true
```

miparnisari: also, unrelated to this PR (and keep in mind i'm still wrapping my head around the overall codebase): i don't understand the logic for deduplication. it is counter-intuitive to me to have deduplication on a union, especially when you are keeping the `relation` as the first relation in the first subiterator?

pkg/query/fixed.go
barakmich: There's not, and I get told not to touch `pkg/tuple`

pkg/query/intersection_test.go
barakmich: It would also return an error, and fail the test. Though a nice little cleanup I'll do right now is a helper to return an empty iterator

pkg/query/union_test.go
barakmich: Yield them all, yes, but if all resource/subjects are found in the first branch, we don't need to call check on the second branch. (This is specific to Check)

PR #2570, opened 2025-09-16
LR3 Fixes and Improvements

josephschorr:
Significant fixes and performance improvements for LR3


internal/cursorediterator/blocks.go
```diff
@@ -234,36 +254,35 @@ func CursoredParallelIterators[I any](
...
 				select {
 				case <-ctx.Done():
+					_ = yield(ItemAndCursor[I]{}, ctx.Err())
 					return
 
-				case r, ok := <-collector:
-					if !ok {
-						continue parent
-					}
+				case err := <-errChannel:
```
miparnisari, line 271: @josephschorr sorry for the late review. I'm seeing that `TestCursoredParallelIterators` is flaky (see https://github.com/authzed/spicedb/actions/runs/18502854529/job/52724196907?pr=2598#step:4:12) and Claude told me what could be the root cause (but disclaimer I haven't confirmed): When a `select` has multiple `case`s, and more than one can be executed, Go chooses one _at random_. So if an iterator yields an item and then an error, we may see the error first (if the consumer isn't fast enough). Again, this is conjecture :)

Ironically, Claude's suggestion was to revert one change that you made:

> The Fix: 
> Modify the item[I] struct to include the error, so errors and items flow through the same channel in order:

internal/cursorediterator/blocks.go
```diff
@@ -107,10 +120,10 @@ func CursoredWithIntegerHeader[I any](
 	return join(cursorAddingHeader, nextWithAdjustedCursor)
 }
 
-// itemOrError is a struct that holds an item or an error.
-type itemOrError[I any] struct {
-	ItemAndCursor ItemAndCursor[I]
-	err           error
+// item is a struct that holds an item or an error.
```
miparnisari, line 123: the godoc is out of date

PR #2572, opened 2025-09-17
perf: add trait filtering support to read relationships

josephschorr:
This allows the read call to skip caveats and/or expiration, where applicable


internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+			expectedCode:          codes.FailedPrecondition,
+		},
+		{
+			name: "invalid relation name",
+			schema: `use expiration
+
+				definition user {}
+				definition document {
+					relation viewer: user
+				}
```
miparnisari, line 2203:
```suggestion
				definition user {}
				definition document {
					relation viewer: user
				}
				definition group {
					relation nonexistent: user
				}
```

internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+	testCases := []struct {
+		name                  string
+		schema                string
+		setupRelationships    []string
+		filter                *v1.RelationshipFilter
+		expectedRelationships []string
+		expectedCode          codes.Code
+	}{
+		{
+			name: "relations with expiration - resource type filter",
```
miparnisari, line 1877: this test doesn't prove that the filter works; add a few relationships for a non-document type

internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+		},
+		{
+			name: "relations with expiration - specific relation filter",
+			schema: `use expiration
+
+				definition user {}
+				definition document {
+					relation viewer: user with expiration
+					relation editor: user
+				}
```
miparnisari, line 1914:
```suggestion
				definition user {}
				definition document {
					relation viewer: user with expiration
					relation editor: user
				}
				definition other {
				   relation viewer: user
				}
```

and add a few relationship for `other#viewer`

internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+
+				definition user {}
+				definition document {
+					relation viewer: user with expiration
+				}
+			`,
+			setupRelationships: []string{
+				`document:doc1#viewer@user:alice[expiration:3000-01-01T00:00:00Z]`,
+				`document:doc2#viewer@user:bob[expiration:3000-01-02T00:00:00Z]`,
+			},
```
miparnisari, line 2044:
```suggestion
			setupRelationships: []string{
				`document:doc1#viewer@user:alice[expiration:3000-01-01T00:00:00Z]`,
				`document:doc2#viewer@user:bob[expiration:3000-01-02T00:00:00Z]`,
				`other:blah#blah@user:alice`,
			},
```

internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+					relation viewer: user with test_caveat
+				}
+				caveat test_caveat(something int) {
+					something > 0
+				}
+			`,
+			setupRelationships: []string{
+				`document:test_doc1#viewer@user:alice[test_caveat:{"something":41}]`,
+				`document:other_doc#viewer@user:bob[test_caveat:{"something":42}]`,
+			},
```
miparnisari, line 2106:
```suggestion
			setupRelationships: []string{
				`document:test_doc1#viewer@user:alice[test_caveat:{"something":41}]`,
				`document:other_doc#viewer@user:bob[test_caveat:{"something":42}]`,
				`other:test_other#blah@blah`,
			},
```

internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+
+				definition user {}
+				definition document {
+					relation viewer: user
+				}
+			`,
+			setupRelationships: []string{},
+			filter: &v1.RelationshipFilter{
+				ResourceType:     "document",
+				OptionalRelation: "nonexistent",
```
miparnisari, line 2212: Add another test for:
```
				ResourceType:     "nonexistent",
				OptionalRelation: "exists_on_some_type",
```

internal/services/v1/relationships_test.go
```diff
@@ -1862,3 +1862,398 @@ func TestManyConcurrentWriteRelationshipsReturnsSerializationErrorOnMemdb(t *tes
...
+				`document:doc1#owner@user:alice[expiration:3000-01-01T00:00:00Z]`,
+				`document:doc1#admin@user:alice[test_caveat:{"something":42}][expiration:3000-01-01T00:00:00Z]`,
+			},
+			filter: &v1.RelationshipFilter{
+				ResourceType: "document",
+				OptionalSubjectFilter: &v1.SubjectFilter{
+					SubjectType:       "user",
+					OptionalSubjectId: "alice",
+				},
+			},
```
miparnisari, line 2171: most of these tests seem to imply that `resourceType` is mandatory but judging by the docs, it's not. 

can you add tests for ONLY OptionalSubjectFilter, ONLY OptionalResourceId, etc?

pkg/schema/definition.go
```diff
@@ -433,3 +440,26 @@ func (def *Definition) PossibleTraitsForAnySubject(relationName string) (Traits,
 
 	return foundTraits, nil
 }
+
+// PossibleTraitsForAnyRelation returns the traits that are possible on *any* relation for the specified resource type.
```
miparnisari, line 444:
```suggestion
// PossibleTraitsForAnyRelation returns the traits that are possible on *any* relation for the specified definition.
```

pkg/datastore/pagination/iterator.go
```diff
@@ -19,14 +19,18 @@ func NewPaginatedIterator(
...
-		ctx,
-		filter,
+	baseOptions := append([]options.QueryOptionsOption{
 		options.WithSort(order),
 		options.WithLimit(&pageSize),
-		options.WithAfter(startCursor),
 		options.WithQueryShape(queryShape),
+	}, opts...)
+
+	iter, err := reader.QueryRelationships(
```
miparnisari, line 30: I'm not very familiar with how iter.seq2 works. Why do we have a call to query here and also down below? is it for performance, so that we fetch the first batch immediately and then if the consumer wants more, we do those calls later ? If so, can you leave a godoc mentioning it?

If not, can we not simplify this function?

```go
func NewPaginatedIterator(
    ctx context.Context,
    reader datastore.Reader,
    filter datastore.RelationshipsFilter,
    limit uint64,
    order options.SortOrder,
    startCursor options.Cursor,
    queryShape queryshape.Shape,
    opts ...options.QueryOptionsOption,
) (datastore.RelationshipIterator, error) {
    baseOptions := append([]options.QueryOptionsOption{
       options.WithSort(order),
       options.WithLimit(&limit),
       options.WithQueryShape(queryShape),
    }, opts...)

    return func(yield func(tuple.Relationship, error) bool) {
       cursor := startCursor
       for {
          iter, err := reader.QueryRelationships(
             ctx,
             filter,
             append(baseOptions, options.WithAfter(cursor))...,
          )
          if err != nil {
             yield(tuple.Relationship{}, err)
             return
          }

          var counter uint64
          for rel, err := range iter {
             if !yield(rel, err) {
                return
             }

             cursor = options.ToCursor(rel)
             counter++

             if counter >= limit {
                break
             }
          }

          if counter < limit {
             return
          }
       }
    }, nil
}
```

pkg/schema/errors.go
```diff
@@ -356,6 +379,14 @@ func NewUnusedCaveatParameterErr(caveatName string, paramName string) error {
 	}
 }
 
+// NewKnownEmptyFilterErr constructs an error indicating that a filter is known to be empty.
+func NewKnownEmptyFilterErr(warningMessage string) error {
+	return asTypeError(KnownEmptyFilterError{
+		error:          fmt.Errorf("filter is known to be empty: %s", warningMessage),
```
miparnisari:
```suggestion
		error:          fmt.Errorf("the filter specified will return an empty set of results because: %s", warningMessage),
```

pkg/schema/typesystem.go
```diff
@@ -64,3 +67,127 @@ func (ts *TypeSystem) getDefinition(ctx context.Context, definition string) (*De
...
+	switch expirationOption {
+	case datastore.ExpirationFilterOptionNone:
+		// No filtering, return traits as-is
+		return traits, nil
+
+	case datastore.ExpirationFilterOptionHasExpiration:
+		// Only relationships WITH expiration are allowed
+		if !traits.AllowsExpiration {
+			// If the relation doesn't support expiration, the filter is empty
+			return Traits{}, NewKnownEmptyFilterErr("filter requires expiration but the schema relation(s) do not support expiration")
```
miparnisari:
```suggestion
			return Traits{}, NewKnownEmptyFilterErr("the filter requested relationships with expiration but the filter relation does not support expiration")
```

also why the plural (s)? you can't specify more than one relation filter right?

internal/services/v1/relationships_test.go
josephschorr: Those should skip the traits system entirely, but added

pkg/schema/definition.go
josephschorr: I want to be explicit here that it is the resource type

pkg/datastore/pagination/iterator.go
josephschorr: We call it for performance; its to avoid holding the DB connection for too long

josephschorr:
Updated

pkg/schema/typesystem.go
josephschorr: Changed

pkg/datastore/pagination/iterator.go
miparnisari: Are you referring to the pagination? I understand that, but then the code I presented does basically the same thing whilst being simpler?

internal/services/v1/relationships_test.go
miparnisari: I don't see this 🤔

pkg/datastore/pagination/iterator.go
josephschorr: Sure; but the goal of this change isn't to significantly modify the rels pagination code. We should do that in a different PR :)

internal/services/v1/relationships_test.go
josephschorr: I added it to the ones below but I'll add it here too

miparnisari (approved):
LGTM

pkg/datastore/pagination/iterator.go
jzelinskie: Did we open up a PR followup for this? @miparnisari

miparnisari: Yes @jzelinskie https://github.com/authzed/spicedb/pull/2597

PR #2574, opened 2025-09-19
fix: span handling in LR3

josephschorr:
Iterators cannot use spans using the traditional `defer span.End()` as it will be called when the chained iterator is returned. Instead, we add a new cter function for wrapping other iterators


tstirrat15 (approved):
Changes make sense to me. It does make me wonder how we'll instrument the new query library.

PR #2575, opened 2025-09-24
metrics: register logical checks metric by default

jzelinskie:
## Description

This change refactors the telemetry code to share a metric with the default registry that contains the number of "logical checks".

This metric is useful for admins and not just reporting back for telemetry.

A nice side effect of this is that we get to eliminate some custom code in favor of Prometheus library utilities for tracking this metric. I suspect there was also a bug with this code because when we loaded this value, we were also [resetting it to 0](https://github.com/authzed/spicedb/blob/7491a826fc13e69b3c5e63a029925890eeb95cad/internal/telemetry/logicalchecks.go#L15), which is semantically wrong.

## Testing

This PR actually deletes tests 😜


squat (dismissed):
Feeding two birds with one scone

internal/telemetry/metrics.go
```diff
@@ -11,15 +11,32 @@ import (
...
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/internal/middleware/usagemetrics"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/promutil"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
+var LogicalChecks = promauto.NewCounter(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "services",
```
miparnisari, line 26:
```suggestion
	Subsystem: "telemetry",
```

otherwise we break this https://github.com/authzed/spicedb/blob/main/TELEMETRY.md#spicedb_telemetry_logical_checks_total-counter

jzelinskie: The metric for telemetry is unchanged because it's provided via this: https://github.com/jzelinskie/spicedb/blob/50506854576e1c833bc374ae5cfc0f65d4d83f69/internal/telemetry/metrics.go#L138-L147

jzelinskie: Just for more context: we explicitly have the normal metric named this way so that folks don't think they've accidentally turned on telemetry.

PR #2577, opened 2025-09-25
fixes breakage of Proposal A6: Retry policies

vroldanbet:
See https://github.com/grpc/proposal/blob/master/A6-client-retries.md#when-retries-are-valid

As I attempted to implement an example client using retry policies, I found that they didn't work.
This was due to the fact SpiceDB's response violated the expectations from that spec, notably:

> In certain cases it is not valid to retry an RPC. These cases occur when the RPC has been committed, and thus it does not make sense to perform the retry.
> The reasoning behind the first scenario is that the Response-Headers include initial metadata
from the server. The metadata (or its absence) it is transmitted to the client application. This may fundamentally change the state of the client, so we cannot safely retry if a failure occurs later in the RPC’s life.

The RequestID middleware was sending the `requestID` as a metadata header, which causes the stream to send a frame just to send the headers before sending the actual response. This, in turn, causes the stream to become committed, which, according to the A6 spec, disables retry policies.

The TL;DR is: use trailers for metadata unless there is an explicit reason to use metadata headers; otherwise, retry policies will become disabled.

This also fixes the version middleware, which was subject to the same problem.

⚠️ This is a breaking change because it moves two values from header to trailer. I'm aware `zed` makes use of this.

### DIFF
--- a/pkg/middleware/requestid/requestid.go
+++ b/pkg/middleware/requestid/requestid.go
@@ -58,14 +58,14 @@ func (r *handleRequestID) ServerReporter(ctx context.Context, _ interceptors.Cal
 	haveRequestID, requestID, ctx := r.fromContextOrGenerate(ctx)
 
 	if haveRequestID {
-		err := responsemeta.SetResponseHeaderMetadata(ctx, map[responsemeta.ResponseMetadataHeaderKey]string{
-			responsemeta.RequestID: requestID,
-		})
-		// if context is cancelled, the stream will be closed, and gRPC will return ErrIllegalHeaderWrite
-		// this prevents logging unnecessary error messages
+		// if context is cancelled, no point in attempting to write any headers, since the client terminated the stream
 		if ctx.Err() != nil {
 			return interceptors.NoopReporter{}, ctx
 		}
+
+		err := responsemeta.SetResponseTrailerMetadata(ctx, map[responsemeta.ResponseMetadataTrailerKey]string{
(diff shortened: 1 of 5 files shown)


pkg/middleware/requestid/requestid.go
```diff
@@ -58,22 +60,37 @@ func (r *handleRequestID) ServerReporter(ctx context.Context, _ interceptors.Cal
...
+
+func (r *requestIDReporter) PostCall(error, time.Duration) {}
+
+func (r *requestIDReporter) PostMsgSend(any, error, time.Duration) {}
+
+func (r *requestIDReporter) PostMsgReceive(any, error, time.Duration) {
+	// if context is cancelled, no point in attempting to write any headers, since the client terminated the stream
+	if errors.Is(r.ctx.Err(), context.Canceled) {
+		return
+	}
```
tstirrat15: Curious here because I still don't understand this all that well - when do you reach for `errors.Is` vs `errors.As`?

tstirrat15 (dismissed):
LGTM; see comments

pkg/middleware/requestid/requestid.go
vroldanbet: `Is` can be safely used when it's basically checking for equality instead of wrapped errors. I'm not sure if this does not hold if you cancel with a cause, but we don't use causes anywhere for now

tstirrat15:
I added a zed issue to update its `version` command accordingly.

PR #2585, opened 2025-10-02
chore: add perfsprint linter

miparnisari:
Follow-up of https://github.com/authzed/spicedb/pull/2565/files#r2400208837

### DIFF
--- a/internal/cursorediterator/blocks_test.go
+++ b/internal/cursorediterator/blocks_test.go
@@ -2,6 +2,7 @@ package cursorediterator
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"iter"
 	"strconv"
@@ -160,7 +161,7 @@ func TestCursoredWithIntegerHeader(t *testing.T) {
 	})
 
 	t.Run("header with range loop processes until error", func(t *testing.T) {
-		headerError := fmt.Errorf("header error")
+		headerError := errors.New("header error")
 
 		header := func(ctx context.Context, startIndex int) iter.Seq2[int, error] {
 			return func(yield func(int, error) bool) {
@@ -251,7 +252,7 @@ func TestCursoredWithIntegerHeader(t *testing.T) {
 	})
 
 	t.Run("header error with consumer early termination", func(t *testing.T) {
-		headerError := fmt.Errorf("header processing error")
+		headerError := errors.New("header processing error")
 
 		header := func(ctx context.Context, startIndex int) iter.Seq2[int, error] {
 			return func(yield func(int, error) bool) {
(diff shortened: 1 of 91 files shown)


internal/cursorediterator/blocks_test.go
```diff
@@ -994,7 +995,7 @@ func TestCursoredProducerMapperIterator(t *testing.T) {
 					return func(yield func(ItemAndCursor[int], error) bool) {
 						for i, item := range chunk {
 							mu.Lock()
-							processingOrder = append(processingOrder, fmt.Sprintf("mapper: processing %s", item))
+							processingOrder = append(processingOrder, "mapper: processing "+item)
```
tstirrat15: This is a preference, but I don't like string concatenation when format strings would work in a given context. How big is the performance difference? Can we disable this particular rule?

tstirrat15 (commented):
See comment, otherwise LGTM

internal/cursorediterator/blocks_test.go
miparnisari: Disabled the linter for test files

internal/datastore/crdb/watch.go
```diff
@@ -93,7 +93,7 @@ func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Rev
 
 		if features.Watch.Status != datastore.FeatureSupported {
 			close(updates)
-			errs <- datastore.NewWatchDisabledErr(fmt.Sprintf("%s. See https://spicedb.dev/d/enable-watch-api-crdb", features.Watch.Reason))
+			errs <- datastore.NewWatchDisabledErr(features.Watch.Reason + ". See https://spicedb.dev/d/enable-watch-api-crdb")
```
miparnisari, line 96: Should the link point to https://authzed.com/docs/spicedb/concepts/datastores#usage-notes ?

internal/cursorediterator/blocks_test.go
miparnisari:
```go
func BenchmarkStringAdditionFormatting(b *testing.B) {
	s1 := "Hello "
	s2 := "world"

	b.Run("fmt.Sprintf", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			_ = fmt.Sprintf("%s%s", s1, s2)
		}
	})

	b.Run("string concatenation", func(b *testing.B) {
		for n := 0; n < b.N; n++ {
			_ = s1 + s2
		}
	})
}
```


```
BenchmarkStringAdditionFormatting/fmt.Sprintf           132447381               45.36 ns/op           16 B/op          1 allocs/op
BenchmarkStringAdditionFormatting/string_concatenation          781756665                7.557 ns/op           0 B/op          0 allocs/op
```

fmt.Sprintf is 6x slower and causes heap allocations, while simple + concatenation is cheap.

internal/datastore/crdb/watch.go
tstirrat15: I think we use the shortener so that we can change what the shortener points at. I don't know where that shortener configuration lives though.

tstirrat15 (approved):
LGTM

PR #2587, opened 2025-10-06
perf: significant improvements around LR3 dispatching

josephschorr:
There was massive overhead associated with dispatching in LR, to the point where an LR2/LR3 call locally with dispatching could be up to 10x (!) slower than without

After much research and testing, this PR works around the vast majority of that overhead. Some of the changes impact LR2, LR3 and LS (such as the middleware changes), but some are solely in LR3, as they required dispatch wire format changes

1) Remove the logging middlewares from streaming dispatch - these middleware run on *every* response message and were adding significant overhead (about 50% of the total) 2) Don't return cursors at all when making an unlimited LR call - this slightly changes the semantics of unlimited LRs, but cursors should not be necessary, and their construction and serialization/deserialization added overhead on gRPC 3) Remove metadata from LR3 - metadata isn't being used in LR right now, but we were paying the construction and serialization/deserialization costs 4) Disabling S2 compression - this added significant overhead (~20%) on streaming calls; unfortunately, we're disabling it for all calls, but that should be fine (checks don't move around a lot of data anyway) 5) Change LR3 cursors and results to not be nested messages - Nested messages incur additional costs in gRPC 6) Switch LR3 to return results in batches of 500: this *significantly* decreases the number of messages being sent, and thus, overhead. Accounts for another ~40% of the overhead

## Testing

Unit tests, Consistency Tests, Steelthread tests and local, manual testing with known datasets


mazdakb:
LGTM when protobuf linter is happy.

Tested with and without limit to execute no cursor path.

PR #2596, opened 2025-10-08
Update buf and use protovalidate

tstirrat15:
## Description
We'd like to use `protovalidate` for protobuf validation instead of `protoc-gen-validate`. This is a first step towards that in SpiceDB, replacing the `validate` generated code with use of `protovalidate`.

This also updates our generation code, using v2 of `buf.gen.yaml`.

## Changes
* Run formatting
* Fix some lint issues around the naming of functions that match a `Printf` signature
* Ran `buf config migrate`
* Removed `protoc-gen-validate` from `buf.yaml` and added `protovalidate` instead
* Use remote generation, like we're doing with the client libraries, instead of invoking protoc locally
* Use `protovalidate.Validate` calls where we were previously calling `.Validate()` on proto messages
## Testing
Review. See that tests pass.

### DIFF
--- a/pkg/schemadsl/compiler/compiler_test.go
+++ b/pkg/schemadsl/compiler/compiler_test.go
@@ -797,7 +797,7 @@ func TestCompile(t *testing.T) {
 			"invalid definition name",
 			nilPrefix,
 			`definition someTenant/fo {}`,
-			"parse error in `invalid definition name`, line 1, column 1: error in object definition someTenant/fo: validation error:\n - name: value does not match regex pattern `^([a-z][a-z0-9_]{1,62}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$` [string.pattern]",
+			"parse error in `invalid definition name`, line 1, column 1: error in object definition someTenant/fo: validation error: name: value does not match regex pattern `^([a-z][a-z0-9_]{1,62}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$`",
 			[]SchemaDefinition{},
 		},
 		{
@@ -806,7 +806,7 @@ func TestCompile(t *testing.T) {
 			`definition some_tenant/foos {
 				relation ab: some_tenant/foos
 			}`,
-			"parse error in `invalid relation name`, line 2, column 5: error in relation ab: validation error:\n - name: value does not match regex pattern `^[a-z][a-z0-9_]{1,62}[a-z0-9]$` [string.pattern]",
+			"parse error in `invalid relation name`, line 2, column 5: error in relation ab: validation error: name: value does not match regex pattern `^[a-z][a-z0-9_]{1,62}[a-z0-9]$`",
 			[]SchemaDefinition{},
 		},
 		{
--- a/pkg/composableschemadsl/compiler/compiler_test.go
+++ b/pkg/composableschemadsl/compiler/compiler_test.go
@@ -769,7 +769,7 @@ func TestCompile(t *testing.T) {
 			"invalid definition name",
 			nilPrefix,
 			`definition someTenant/fo {}`,
-			"parse error in `invalid definition name`, line 1, column 1: error in object definition someTenant/fo: validation error:\n - name: value does not match regex pattern `^([a-z][a-z0-9_]{1,62}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$` [string.pattern]",
+			"parse error in `invalid definition name`, line 1, column 1: error in object definition someTenant/fo: validation error: name: value does not match regex pattern `^([a-z][a-z0-9_]{1,62}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$`",
 			[]SchemaDefinition{},
 		},
 		{
@@ -778,7 +778,7 @@ func TestCompile(t *testing.T) {
 			`definition some_tenant/foos {
 				relation ab: some_tenant/foos
 			}`,
-			"parse error in `invalid relation name`, line 2, column 5: error in relation ab: validation error:\n - name: value does not match regex pattern `^[a-z][a-z0-9_]{1,62}[a-z0-9]$` [string.pattern]",
+			"parse error in `invalid relation name`, line 2, column 5: error in relation ab: validation error: name: value does not match regex pattern `^[a-z][a-z0-9_]{1,62}[a-z0-9]$`",
 			[]SchemaDefinition{},
 		},
 		{


cmd/spicedb/serve_integration_test.go
```diff
@@ -10,8 +10,6 @@ import (
 	"testing"
 	"time"
 
-	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
-	"github.com/authzed/grpcutil"
```
tstirrat15: There's a number of these changes. This is the formatter.

e2e/newenemy/newenemy_test.go
```diff
@@ -277,7 +277,7 @@ func attemptFnsForProbeFns(vulnerableMax int, vulnerableProbe, protectedProbe pr
 			return
 		}
 	}
-	return
+	return vulnerableFn, protectedFn
```
tstirrat15, line 280: There's also several of these changes, which are also the formatter.

internal/datastore/common/schema.go
```diff
@@ -79,7 +79,7 @@ func (si SchemaInformation) expectedIndexesForShape(shape queryshape.Shape) opti
 }
 
 func (si SchemaInformation) debugValidate() {
-	spiceerrors.DebugAssert(func() bool {
+	spiceerrors.DebugAssertf(func() bool {
```
tstirrat15, line 82: This is the new linter complaint - because `DebugAssert` had a `Printf`-like signature, the linter wanted the function to end in `f`

internal/datastore/common/sql.go
```diff
@@ -191,7 +191,7 @@ func (sqf SchemaQueryFilterer) WithIndexingHint(indexingHint IndexingHint) Schem
 }
 
 func (sqf SchemaQueryFilterer) UnderlyingQueryBuilder() sq.SelectBuilder {
-	spiceerrors.DebugAssert(func() bool {
+	spiceerrors.DebugAssertf(func() bool {
```
tstirrat15, line 194: So there's lots of these changes

internal/testfixtures/validating.go
```diff
@@ -61,7 +63,7 @@ func (vsr validatingSnapshotReader) ListAllNamespaces(
 	}
 
 	for _, ns := range read {
-		err := ns.Definition.Validate()
+		err := protovalidate.Validate(ns.Definition)
```
tstirrat15, line 66: These are a part of the substantive changes in this PR - `protovalidate` uses reflection on the messages to figure out what the associated CEL is, and then runs those expressions from the outside, rather than invoking generated code (which is what `protoc-gen-validate` used to do)

pkg/proto/core/v1/core.pb.go
tstirrat15, line 1: These are changes brought on by moving to remote generation + the most recent version of `protoc-gen-go`. They should be functionally equivalent.

pkg/proto/core/v1/core.pb.validate.go
```diff
@@ -1,7023 +0,0 @@
-// Code generated by protoc-gen-validate. DO NOT EDIT.
```
tstirrat15, line 1: All of the generated `.validate.go` files were deleted, as this is the logic that now lives in `protovalidate`.

magefiles/gen.go
```diff
@@ -28,7 +28,7 @@ func (Gen) Go() error {
 func (Gen) Proto() error {
 	fmt.Println("generating buf")
 	return RunSh("go", Tool())("run", "github.com/bufbuild/buf/cmd/buf",
-		"generate", "-o", "../pkg/proto", "../proto/internal", "--template", "../buf.gen.yaml")
+		"generate", "-o", "..", "../proto/internal", "--template", "../buf.gen.yaml")
```
tstirrat15: The prefix is provided by the `buf.gen.yaml` configuration now.

proto/internal/core/v1/core.proto
```diff
@@ -4,22 +4,22 @@ package core.v1;
...
 import "google/protobuf/timestamp.proto";
-import "validate/validate.proto";
+import "buf/validate/validate.proto";
 
 option go_package = "github.com/authzed/spicedb/pkg/proto/core/v1";
 
 message RelationTuple {
   /** resource_and_relation is the resource for the tuple */
-  ObjectAndRelation resource_and_relation = 1 [(validate.rules).message.required = true];
+  ObjectAndRelation resource_and_relation = 1 [(buf.validate.field).required = true];
```
tstirrat15, line 14: These are changes brought by running the [migration tool](https://protovalidate.com/migration-guides/migrate-from-protoc-gen-validate/).

buf.yaml
tstirrat15, line 1: This got moved down and out of the `proto` folder by the migration step. I think this is a fine place for it to be.

buf.work.yaml
```diff
@@ -1,4 +0,0 @@
----
```
tstirrat15, line 1: No longer necessary with `buf.yaml` being in the root.

buf.gen.yaml
```diff
@@ -1,29 +1,26 @@
...
-  - name: 'validate'
-    path: ["go", "run", "github.com/envoyproxy/protoc-gen-validate"]
-    out: '.'
-    opt: 'paths=source_relative,lang=go'
+  - remote: "buf.build/protocolbuffers/go:v1.36.10"
+    out: "pkg/proto"
+    opt: "paths=source_relative"
+  - remote: "buf.build/grpc/go:v1.5.1"
+    out: "pkg/proto"
+    opt: "paths=source_relative"
```
tstirrat15, line 21: We're using the remote plugins now, similar to what we're doing with the client libraries.

proto/internal/core/v1/core.proto
```diff
@@ -439,11 +435,11 @@ message AllowedCaveat {
 
 message UsersetRewrite {
   oneof rewrite_operation {
-    option (validate.required) = true;
+    option (buf.validate.oneof).required = true;
 
-    SetOperation union = 1 [(validate.rules).message.required = true];
-    SetOperation intersection = 2 [(validate.rules).message.required = true];
-    SetOperation exclusion = 3 [(validate.rules).message.required = true];
```
tstirrat15, line 451: This was a change called out by `buf lint` - the `required` fields don't mean anything on the options in a `oneof`, so the linter says they should be excluded.

buf.yaml
```diff
@@ -1,19 +1,26 @@
...
-  - "buf.build/envoyproxy/protoc-gen-validate:bb405eae115246f0b5ccf8997136e3d8"
   - "buf.build/authzed/api:53f3e981ed3346bab8c050c597f74602"
+  - "buf.build/bufbuild/protovalidate:v1.0.0"
   - "buf.build/googleapis/googleapis:4d230ddc6b6d28ad7d2503926a97c6f7fa15483d"
 lint:
   except:
     - "ENUM_VALUE_PREFIX"
     - "ENUM_ZERO_VALUE_SUFFIX"
+    - "FIELD_NOT_REQUIRED"
+    - "PACKAGE_NO_IMPORT_CYCLE"
```
miparnisari: 😨 we have a cycle somewhere?

cmd/spicedb/serve_integration_test.go
miparnisari: I don't understand, which formatter do you mean? You didn't change the Go linter rules..

e2e/newenemy/newenemy_test.go
miparnisari: let's do this in a separate PR https://github.com/authzed/spicedb/pull/2600

pkg/proto/impl/v1/impl.pb.go
```diff
@@ -401,19 +399,19 @@ type V1Cursor struct {
 	// dispatch_version is the version of the dispatcher which created the cursor.
 	DispatchVersion uint32 `protobuf:"varint,4,opt,name=dispatch_version,json=dispatchVersion,proto3" json:"dispatch_version,omitempty"`
 	// flags are flags set by the API caller.
-	Flags map[string]string `protobuf:"bytes,5,rep,name=flags,proto3" json:"flags,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
+	Flags map[string]string `protobuf:"bytes,5,rep,name=flags,proto3" json:"flags,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
```
miparnisari, line 402: ?

pkg/proto/core/v1/core.pb.go
```diff
@@ -597,29 +589,26 @@ func (x *ContextualizedCaveat) GetContext() *structpb.Struct {
...
-	unknownFields protoimpl.UnknownFields
-
+	state protoimpl.MessageState `protogen:"open.v1"`
 	// * name represents the globally-unique identifier of the caveat *
 	Name string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
 	// * serialized_expression is the byte representation of a caveat's logic
 	SerializedExpression []byte `protobuf:"bytes,2,opt,name=serialized_expression,json=serializedExpression,proto3" json:"serialized_expression,omitempty"`
 	// * parameters_and_types is a map from parameter name to its type
-	ParameterTypes map[string]*CaveatTypeReference `protobuf:"bytes,3,rep,name=parameter_types,json=parameterTypes,proto3" json:"parameter_types,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
+	ParameterTypes map[string]*CaveatTypeReference `protobuf:"bytes,3,rep,name=parameter_types,json=parameterTypes,proto3" json:"parameter_types,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
```
miparnisari, line 603: ?

tstirrat15:
Yeah, that's not a bad idea.

pkg/proto/impl/v1/impl.pb.go
tstirrat15: This is a difference between the remote plugin's version and the version we were previously using. I'm not expecting it to have any meaningful effect, but I also can't say for certain because I don't understand this part of protobuf generation very well.

pkg/proto/core/v1/core.pb.go
tstirrat15: Same

buf.yaml
tstirrat15: No - I think this was added defensively by the migration script. I removed both of these exceptions and lints still pass, so I committed that.

tstirrat15:
Some benchmarks from the Buf folks: https://github.com/bufbuild/protovalidate-go/pull/283#discussion_r2363601822

Their sense is that `protovalidate` will be slower than executing the code generated by `protoc-gen-validate`, but not by enough to be a bottleneck concern. It's also using protobuf reflection, which is a separate concept from golang reflection and not as expensive.

pkg/schemadsl/compiler/compiler_test.go
```diff
@@ -509,7 +509,7 @@ func TestCompile(t *testing.T) {
 			"invalid definition name",
 			nilPrefix,
 			`definition someTenant/fo {}`,
-			"parse error in `invalid definition name`, line 1, column 1: error in object definition someTenant/fo: invalid NamespaceDefinition.Name: value does not match regex pattern \"^([a-z][a-z0-9_]{1,62}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$\"",
+			"parse error in `invalid definition name`, line 1, column 1: error in object definition someTenant/fo: validation error:\n - name: value does not match regex pattern `^([a-z][a-z0-9_]{1,62}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$` [string.pattern]",
```
tstirrat15: These changes represent something that's different from the perspective of the user - it no longer names the object being validated, and it includes the `[string.pattern]` bit at the end. I think this is fine.

cmd/spicedb/serve_integration_test.go
tstirrat15: `golangci-lint fmt` is what i'm using locally

PR #2598, opened 2025-10-09
chore: refactor dispatch/remote/cluster.go by using a concurrent map

miparnisari:
Follow-up from [this](https://github.com/authzed/spicedb/pull/2543#discussion_r2308380947) and [this](https://github.com/authzed/internal/pull/8009/files#r2214412749)


miparnisari:
@tstirrat15 to be completely honest, I haven't done benchmarking to compare both implementations

PR #2604, opened 2025-10-10
feat: add granular datastore connection flags

### LINKED ISSUE #2218: Granular Flags/Vars for Datastore Connections
### Problem Statement

Sometimes, people don't store their DB connection info as a connection string, and they choose another method (e.g., a k8s secret with separate fields for the endpoint, password, username, etc.) It can be cumbersome for these people to format this connection data as a connection string.

### Solution Brainstorm

Separate flags/vars for each component of a connection string.

ivanauth:
Fixes #2218

## Summary
Adds individual connection parameter flags as an alternative to `--datastore-conn-uri` for easier configuration from Kubernetes secrets or environment variables.

## New Flags
- `--datastore-host`: database host
- `--datastore-port`: database port (defaults: 5432 for Postgres, 26257 for CockroachDB, 3306 for MySQL)
- `--datastore-username`: database username
- `--datastore-password`: database password
- `--datastore-database`: database name
- `--datastore-conn-param`: repeatable key/value pairs for engine-specific options, appended to the built URI

## Behavior
- Builds the connection URI from the granular params when `--datastore-conn-uri` is not provided
- Supplying both `--datastore-conn-uri` and any granular flag is an **error**, naming the flags that conflict
- Granular flags without `--datastore-host` are an error
- Engines with no granular support (spanner, memory) are an error rather than a silently empty URI
- Supports Postgres, CockroachDB, and MySQL
- Fully backward compatible: `--datastore-conn-uri` on its own is untouched

## Example Usage
```bash
spicedb serve \
  --datastore-engine=postgres \
  --datastore-host=localhost \
  --datastore-port=5432 \
  --datastore-username=spicedb \
  --datastore-password=secret \
  --datastore-database=spicedb \
  --datastore-conn-param=sslmode=require
```

## Notes on URI construction
The URI is built with `net/url` for Postgres/CockroachDB and with go-sql-driver's `mysql.Config.FormatDSN()` for MySQL, rather than by hand, which avoids three encoding bugs:

- `url.QueryEscape` encodes a space as `+`, which the userinfo section reads literally
- A MySQL DSN is not a URL and the driver does not percent-decode it, so encoding the credentials changes them
- `fmt.Sprintf("%s:%s", host, port)` produces an unusable address for IPv6 hosts

`TestBuildConnectionURIParsesForDriver` round-trips the generated URIs through `pgx.ParseConfig` and `mysql.ParseDSN` to confirm the credentials survive.


pkg/cmd/datastore/datastore_test.go
```diff
@@ -125,3 +125,187 @@ func TestLoadDatastoreFromFileAndContents(t *testing.T) {
...
+			engine:   PostgresEngine,
+			host:     "localhost",
+			port:     "5432",
+			username: "testuser",
+			password: "testpass",
+			database: "",
+			expected: "postgres://testuser:testpass@localhost:5432",
+		},
+		{
+			name:     "cockroachdb with all params",
```
miparnisari, line 191: password is unset

pkg/cmd/datastore/datastore.go
```diff
@@ -68,6 +69,14 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 
 	flagSet.StringVar(&opts.Engine, flagName("datastore-engine"), defaults.Engine, fmt.Sprintf(`type of datastore to initialize (%s)`, datastore.EngineOptions()))
 	flagSet.StringVar(&opts.URI, flagName("datastore-conn-uri"), defaults.URI, `connection string used by remote datastores (e.g. "postgres://postgres:password@localhost:5432/spicedb")`)
+
+	// Granular connection parameters, used to build the URI when datastore-conn-uri is not provided.
```
miparnisari: how do you pass extra config options that are database specific?

pkg/cmd/datastore/datastore_test.go
```diff
@@ -125,3 +125,187 @@ func TestLoadDatastoreFromFileAndContents(t *testing.T) {
...
+	}
+
+	require.NotEmpty(t, config.URI)
+	require.Equal(t, "postgres://testuser:testpass@localhost:5432/testdb", config.URI)
+}
+
+func TestNewDatastoreURITakesPrecedence(t *testing.T) {
+	ctx := t.Context()
+
+	// Test that explicit URI takes precedence over granular params
```
miparnisari: i think i prefer to error out if both are set

ivanauth: Agreed - changed to an error.

pkg/cmd/datastore/datastore.go
ivanauth: Good catch - you couldn't, which made the granular flags unusable for anything needing TLS. Added `--datastore-conn-param`, a repeatable key/value flag whose entries are appended to the built UR

pkg/cmd/datastore/datastore_test.go
ivanauth: Fixed

PR #2605, opened 2025-10-10
feat: relationship expiration is now on by default

miparnisari:
🎉 

Do note that `use expiration` is still required on the schema.

## Testing

Flag doesn't show up in the `experimental` section anymore:

```go
 go run ./cmd/spicedb/main.go serve --help
Experimental Flags:
      --enable-experimental-watchable-schema-cache                                      enables the experimental schema cache, which uses the Watch API to keep the schema up to date
      --experimental-dispatch-secondary-maximum-primary-hedging-delays stringToString   maximum number of hedging delays to use for each request type to delay the primary request. default is 5ms (default [])
      --experimental-dispatch-secondary-upstream-addrs stringToString                   secondary upstream addresses for dispatches, each with a name (default [])
      --experimental-dispatch-secondary-upstream-exprs stringToString                   map from request type to its associated CEL expression, which returns the secondary upstream(s) to be used for the request (default [])
      --experimental-lookup-resources-version lr3                                       if non-empty, the version of the experimental lookup resources API to use: lr3 or empty
      --lookup-resources-chunk-cache-enabled                                            enable caching (default true)
      --lookup-resources-chunk-cache-max-cost string                                    upper bound cache size in bytes or percent of available memory (default "50MiB")
      --lookup-resources-chunk-cache-metrics                                            enable cache metrics
      --lookup-resources-chunk-cache-num-counters int                                   number of TinyLFU samples to track. A higher number means more accurate eviction decisions but more memory usage (default 10000)
```

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -362,11 +362,11 @@ func (cds *crdbDatastore) ReadWriteTx(
 		//    a deletion of expired relationships.
 		//
 		//    A transaction is marked as such IF and only IF the operations in the transaction
-		//    consist solely of deletions, as in that scenario, we cannot be certain in the Watc
+		//    consist solely of deletions, as in that scenario, we cannot be certain in the Watch
 		//    changefeed that the transaction is not a deletion of expired relationships performed
 		//    by CRDB itself. This is also only necessary if both expiration and watch are enabled.
 		metadata := config.Metadata.AsMap()
-		requiresMetadata := len(metadata) > 0 || (cds.watchEnabled && !rwt.hasNonExpiredDeletionChange)
+		requiresMetadata := len(metadata) > 0 || (cds.watchEnabled && (config.IncludesExpiredAt || !rwt.hasNonExpiredDeletionChange))
 		if requiresMetadata {
 			// Mark the transaction as coming from SpiceDB. See the comment in watch.go
 			// for why this is necessary.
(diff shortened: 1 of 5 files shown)


pkg/cmd/datastore/datastore.go
```diff
@@ -706,7 +706,7 @@ func newSpannerDatastore(ctx context.Context, opts Config) (datastore.Datastore,
 		spanner.AllowedMigrations(opts.AllowedMigrations),
 		spanner.FilterMaximumIDCount(opts.FilterMaximumIDCount),
 		spanner.WithColumnOptimization(opts.ExperimentalColumnOptimization),
-		spanner.WithExpirationDisabled(!opts.EnableExperimentalRelationshipExpiration),
+		spanner.WithExpirationDisabled(!opts.EnableRelationshipExpiration),
```
josephschorr: We could remove the `WithExpirationDisabled` config on each of the datastores

internal/datastore/crdb/crdb.go
```diff
@@ -368,7 +366,7 @@ func (cds *crdbDatastore) ReadWriteTx(
 		//    changefeed that the transaction is not a deletion of expired relationships performed
 		//    by CRDB itself. This is also only necessary if both expiration and watch are enabled.
 		metadata := config.Metadata.AsMap()
-		requiresMetadata := len(metadata) > 0 || (cds.expirationEnabled && cds.watchEnabled && !rwt.hasNonExpiredDeletionChange)
+		requiresMetadata := len(metadata) > 0 || (cds.watchEnabled && !rwt.hasNonExpiredDeletionChange)
 		if requiresMetadata {
```
josephschorr, line 370: Hmm... this is going to start writing metadata even on installs that don't use expiration at all. We might need to still pass a flag based on whether `use expiration` was specified in the schema

miparnisari: In https://github.com/authzed/spicedb/pull/2605/commits/8f6fcb6358e8e844ba42d726493167ef2cb6dc4f i made it so that we pass a flag to `ReadWriteTx` which has whether or not the Write contains any tuples with `ExpiredAt` set.

josephschorr (approved):
LGTM

PR #2608, opened 2025-10-10
Schema V2 translation, cloning and flattening support

josephschorr:
## Description

This PR adds schema translation capabilities to the v2 schema library to support converting complex permission expressions into a flattened form suitable for translation to core v1 schemas. This is foundational work for enabling schema v2 to v1 translation.

The changes introduce three key capabilities:

1. **Schema Cloning** (`clone.go`): Deep clones schema trees to enable safe mutation during transformation operations without modifying the original schema.
2. **Schema Resolution** (`resolved.go`): Resolves all `RelationReference` and `ArrowReference` nodes in permission expressions to their actual targets within definitions. This creates a fully-linked schema tree where all references point directly to their corresponding relations/permissions, enabling subsequent transformation operations.
3. **Schema Flattening** (`flatten.go`): Recursively flattens nested permission expressions by extracting compound operations (unions, intersections, exclusions) into synthetic permissions. Synthetic permissions are named using the pattern `{permissionName}{separator}{hash}`where the hash is computed via BDD canonicalization using the `rudd` library. This ensures structurally equivalent operations receive identical names, enabling deduplication. Two separator modes are supported: `$` (for internal use) and `__` (valid in DSL identifiers).

### DIFF
--- a/pkg/schema/v2/tocorev1.go
+++ b/pkg/schema/v2/tocorev1.go
@@ -1,0 +1,331 @@
+package schema
+
+import (
+	"errors"
+	"fmt"
+
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// ToNamespaceDefinition converts a Schema to a core.NamespaceDefinition.
+// This is useful for converting schemas back to the protobuf format for serialization.
+func (s *Schema) ToNamespaceDefinition(name string) ([]*core.NamespaceDefinition, []*core.CaveatDefinition, error) {
+	if s == nil {
+		return nil, nil, errors.New("cannot convert nil schema")
+	}
+
+	// Convert each definition
+	definitions := make([]*core.NamespaceDefinition, 0, len(s.definitions))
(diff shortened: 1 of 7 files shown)


josephschorr:
@barakmich was thinking more about this over the weekend, I wonder if instead of naming the synthetic relations `$parent__$hash`, we just do `perm__$hash`? That way, they get reused?

pkg/schema/v2/tocorev1.go
barakmich, line 1: Funnily, I initially tried to avoid this with the idea that protos would remain the source of truth, but I suppose manipulating schemas like this is rather convenient. 

Regardless, this would need some tests, especially round-trip ones

barakmich (dismissed):
Approve with testing comments. Mostly the meat of it (cloning, resolving) looks good

pkg/schema/v2/tocorev1.go
josephschorr: We can always convert back into protos down the road, especially different ones

PR #2614, opened 2025-10-14
feat: disable tracing of health check requests

ivanauth:
## Summary
~Adds `--otel-disable-healthcheck-tracing` flag to suppress OpenTelemetry spans for health check endpoints, reducing trace volume in production.~

No flag provided; they are disabled by default.

## Changes
- Added CLI flag `--otel-disable-healthcheck-tracing` in telemetry section
- Applied gRPC health check filtering using `otelgrpc.WithFilter(filters.Not(filters.HealthCheck()))`
- Applied HTTP `/healthz` filtering using `otelhttp.WithFilter(httpfilters.Not(httpfilters.Path("/healthz")))`
- Filters applied to main server, dispatch server, test servers, and REST gateway
- Added test coverage verifying health checks are not traced while regular API calls remain traced

Fixes #2246

PR #2616, opened 2025-10-15
chore: add docker compose setup with grafana dashboard

miparnisari:
## Description

To test https://github.com/authzed/spicedb/pull/2614 I used https://github.com/authzed/examples/tree/main/tracing. Then decided to copy that setup here because it's very handy to see how the system as a whole behaves.

## Testing

Follow the new steps in the contributing guide. You should be able to see metrics and traces in localhost:3000.


development/schema.yaml
miparnisari, line 1: FYI this has many relationships. Let me know if this is _too many_. The idea was that this will help in case of a load test

Issue #2505, opened 2025-07-21
CockroachDB - EXPERIMENTAL CHANGEFEED is deprecated


jbiel:
In CockroachDB v25.2 the `EXPERIMENTAL CHANGEFEED` feature is deprecated and `CHANGEFEED` should be used instead. SpiceDB currently uses the deprecated functionality.

> The [EXPERIMENTAL CHANGEFEED FOR](https://www.cockroachlabs.com/docs/v25.2/changefeed-for) SQL statement is now deprecated and will be removed in a future release. Instead, create a sinkless changefeed that emits messages directly to a SQL client with the [CREATE CHANGEFEED](https://www.cockroachlabs.com/docs/v25.2/create-changefeed#create-a-sinkless-changefeed) statement.

https://www.cockroachlabs.com/docs/releases/v25.2#v25-2-0-deprecations

miparnisari:
Thank you for the report @jbiel! I've opened a PR for this: https://github.com/authzed/spicedb/pull/2617

PR #2617, opened 2025-10-16
chore: move off experimental changefeed query for CRDB

miparnisari:
Closes https://github.com/authzed/spicedb/issues/2505

### DIFF
--- a/internal/datastore/crdb/crdb_test.go
+++ b/internal/datastore/crdb/crdb_test.go
@@ -251,7 +251,7 @@ func TestWatchFeatureDetection(t *testing.T) {
 		{
 			name: "rangefeeds disabled",
 			postInit: func(ctx context.Context, adminConn *pgx.Conn) {
-				_, err = adminConn.Exec(ctx, `SET CLUSTER SETTING kv.rangefeed.enabled = false;`)
+				_, err := adminConn.Exec(ctx, `SET CLUSTER SETTING kv.rangefeed.enabled = false;`)
 				require.NoError(t, err)
 			},
 			expectEnabled: false,
@@ -260,7 +260,7 @@ func TestWatchFeatureDetection(t *testing.T) {
 		{
 			name: "rangefeeds enabled, user doesn't have permission",
 			postInit: func(ctx context.Context, adminConn *pgx.Conn) {
-				_, err = adminConn.Exec(ctx, `SET CLUSTER SETTING kv.rangefeed.enabled = true;`)
+				_, err := adminConn.Exec(ctx, `SET CLUSTER SETTING kv.rangefeed.enabled = true;`)
 				require.NoError(t, err)
 			},
 			expectEnabled: false,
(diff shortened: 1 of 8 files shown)


internal/datastore/crdb/crdb_test.go
```diff
@@ -251,7 +251,7 @@ func TestWatchFeatureDetection(t *testing.T) {
 		{
 			name: "rangefeeds disabled",
 			postInit: func(ctx context.Context, adminConn *pgx.Conn) {
-				_, err = adminConn.Exec(ctx, `SET CLUSTER SETTING kv.rangefeed.enabled = false;`)
+				_, err := adminConn.Exec(ctx, `SET CLUSTER SETTING kv.rangefeed.enabled = false;`)
```
miparnisari, line 254: When I turned on `-race` this was triggering it 🤦‍♀️

internal/testserver/datastore/crdb.go
```diff
@@ -24,15 +25,19 @@ const (
 	enableRangefeeds = `SET CLUSTER SETTING kv.rangefeed.enabled = true;`
 )
 
+// crdbTester is safe for concurrent use by tests.
```
miparnisari, line 28: Before i added the mutex, `mage testds:crdbver 25.2.0` consistently failed with "conn busy" when I ran in my machine. However, even without the change it works on Github. The mysteries of the universe..

internal/testserver/datastore/crdb.go
```diff
@@ -24,15 +25,19 @@ const (
...
+	hostname  string
+	creds     string
+	port      string
 }
 
+var _ RunningEngineForTest = (*crdbTester)(nil)
+
 // RunCRDBForTesting returns a RunningEngineForTest for CRDB
-func RunCRDBForTesting(t testing.TB, bridgeNetworkName string, crdbVersion string) RunningEngineForTest {
+func RunCRDBForTesting(t testing.TB, bridgeNetworkName string, crdbVersion string) *crdbTester {
```
miparnisari, line 40: [Accept Interfaces, Return Concrete Types](https://github.com/butuzov/ireturn)

internal/datastore/crdb/pool_test.go
```diff
@@ -117,7 +117,7 @@ func TestTxReset(t *testing.T) {
 
 			r, err := ds.ReadyState(ctx)
 			require.NoError(err)
-			require.True(r.IsReady)
+			require.True(r.IsReady, "datastore not ready: %s", r.Message)
```
miparnisari, line 120: This test keeps failing locally but it passes in Github actions and i don't know why 😭 

```
--- FAIL: TestTxReset (4.71s)
    --- FAIL: TestTxReset/resettable (1.17s)
        pool_test.go:120: 
                Error Trace:    /Users/miparnisari/Documents/GitHub/spicedb/internal/datastore/crdb/pool_test.go:120
                Error:          Should be true
                Test:           TestTxReset/resettable
                Messages:       datastore not ready: spicedb does not have the required minimum connection count to the datastore. Read: 16/15, Write: 13/15
```

internal/datastore/crdb/crdb_test.go
tstirrat15: wild. does the linter not yell at you about no new assignments?

internal/datastore/crdb/pool_test.go
tstirrat15: that's bizarre...

internal/datastore/crdb/crdb_test.go
miparnisari: no, because this code is technically correct, the `postInit`s are overriding an `err` that is declared outside of it.

PR #2621, opened 2025-10-16
feat: Add recursive iterators

barakmich:
## Description

Adds a new iterator type to the query plan, specifically Recursive and RecursiveSentinel.

How it works:
- When a recursive definition is encountered walking the schema, we add a RecursiveSentinel.
- At the top level, where the recursive definition was encountered, we add a Recursive iterator. 
- The Recursive iterator is in charge of running the recursion. First, it runs the tree as-is, and then, clones the tree, replaces the sentinel with the clone of the tree, creating a new sentinel that is one iteration deeper than before.
- Continue evaluation, and replacement, until max depth is reached.

### DIFF
--- a/pkg/query/recursive_test.go
+++ b/pkg/query/recursive_test.go
@@ -110,3 +110,112 @@ func TestReplaceSentinels(t *testing.T) {
 	_, isFixed = arrowIter.right.(*FixedIterator)
 	require.True(t, isFixed, "Sentinel in arrow should be replaced with tree")
 }
+
+func TestRecursiveIteratorClone(t *testing.T) {
+	// Create a recursive iterator with a non-trivial tree
+	sentinel := NewRecursiveSentinel("folder", "view", false)
+	union := NewUnion()
+	union.addSubIterator(NewEmptyFixedIterator())
+	union.addSubIterator(sentinel)
+
+	recursive := NewRecursiveIterator(union)
+
+	// Clone it
+	cloned := recursive.Clone()
+
+	// Verify it's a different instance
+	require.NotSame(t, recursive, cloned)
+
+	// Verify the structure is the same
+	clonedRecursive := cloned.(*RecursiveIterator)
+	require.NotNil(t, clonedRecursive.templateTree)
+
+	// Verify the cloned tree is also different instances
+	require.NotSame(t, recursive.templateTree, clonedRecursive.templateTree)
+
+	// But the structure should be equivalent
+	require.Equal(t, recursive.Explain().Name, clonedRecursive.Explain().Name)
+}
+
+func TestRecursiveIteratorSubiteratorsAndReplace(t *testing.T) {
+	// Create a recursive iterator
+	sentinel := NewRecursiveSentinel("folder", "view", false)
+	union := NewUnion()
+	union.addSubIterator(NewEmptyFixedIterator())
+	union.addSubIterator(sentinel)
+
+	recursive := NewRecursiveIterator(union)
+
+	// Test Subiterators
+	subs := recursive.Subiterators()
+	require.Len(t, subs, 1)
+	require.Same(t, union, subs[0])
+
+	// Test ReplaceSubiterators
+	newTree := NewEmptyFixedIterator()
+	replaced, err := recursive.ReplaceSubiterators([]Iterator{newTree})
+	require.NoError(t, err)
+
+	// Verify the replacement worked
+	replacedRecursive := replaced.(*RecursiveIterator)
+	require.Same(t, newTree, replacedRecursive.templateTree)
+
+	// Verify original is unchanged
+	require.Same(t, union, recursive.templateTree)
+}
+
+func TestRecursiveIteratorExecutionError(t *testing.T) {
+	// Test error path: execute() returns an error (line 67-70 in iterativeDeepening)
+	// This tests when CheckImpl/IterSubjects/IterResources fails during execution
+
+	faultyIter := NewFaultyIterator(true, false) // Fails on Check
+	recursive := NewRecursiveIterator(faultyIter)
+
+	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	require.NoError(t, err)
+
+	ctx := &Context{
+		Context:   context.Background(),
+		Executor:  LocalExecutor{},
+		Datastore: ds,
+		Revision:  datastore.NoRevision,
+	}
+
+	// Test CheckImpl with a faulty iterator
+	seq, err := recursive.CheckImpl(ctx, []Object{{ObjectType: "folder", ObjectID: "folder1"}}, ObjectAndRelation{ObjectType: "user", ObjectID: "tom", Relation: "..."})
+	require.NoError(t, err, "CheckImpl should return sequence without error")
+
+	// Error should occur during sequence iteration
+	paths, err := CollectAll(seq)
+	require.Error(t, err, "Should get error from faulty iterator during execution")
+	require.Contains(t, err.Error(), "execution failed at depth", "Error should be wrapped with depth info")
+	require.Empty(t, paths)
+}
+
+func TestRecursiveIteratorCollectionError(t *testing.T) {
+	// Test error path: pathSeq yields an error during iteration (line 78-80 in iterativeDeepening)
+	// This tests when the returned PathSeq fails during collection
+
+	faultyIter := NewFaultyIterator(false, true) // Fails on collection
+	recursive := NewRecursiveIterator(faultyIter)
+
+	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	require.NoError(t, err)
+
+	ctx := &Context{
+		Context:   context.Background(),
+		Executor:  LocalExecutor{},
+		Datastore: ds,
+		Revision:  datastore.NoRevision,
+	}
+
+	// Test CheckImpl with a faulty iterator that fails on collection
+	seq, err := recursive.CheckImpl(ctx, []Object{{ObjectType: "folder", ObjectID: "folder1"}}, ObjectAndRelation{ObjectType: "user", ObjectID: "tom", Relation: "..."})
+	require.NoError(t, err, "CheckImpl should return sequence without error")
+
+	// Error should occur during sequence iteration (collection)
+	paths, err := CollectAll(seq)
+	require.Error(t, err, "Should get error during path collection")
+	require.Contains(t, err.Error(), "faulty iterator collection error", "Should get collection error")
+	require.Empty(t, paths)
+}
--- a/pkg/query/walk_test.go
+++ b/pkg/query/walk_test.go
@@ -1,0 +1,155 @@
+package query
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWalkBasic(t *testing.T) {
+	// Create a simple tree: union with two fixed iterators
+	path1 := MustPathFromString("document:doc1#view@user:alice")
+	fixedIter1 := NewFixedIterator(path1)
+	fixedIter2 := NewEmptyFixedIterator()
+
+	union := NewUnion()
+	union.addSubIterator(fixedIter1)
+	union.addSubIterator(fixedIter2)
+
+	// Walk the tree and count the nodes
+	nodeCount := 0
+	result, err := Walk(union, func(it Iterator) (Iterator, error) {
+		nodeCount++
+		return it, nil
+	})
+
+	require.NoError(t, err)
+	require.NotNil(t, result)
+	require.Equal(t, 3, nodeCount) // 2 fixed iterators + 1 union
+}
+
+func TestWalkNilRoot(t *testing.T) {
+	// Walk with nil root should return nil
+	result, err := Walk(nil, func(it Iterator) (Iterator, error) {
+		t.Fatal("callback should not be called for nil root")
+		return it, nil
+	})
+
+	require.NoError(t, err)
+	require.Nil(t, result)
+}
+
+func TestWalkWithTransformation(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/query/arrow.go
```diff
@@ -138,3 +138,11 @@ func (a *Arrow) Explain() Explain {
...
 	}
 }
+
+func (a *Arrow) Subiterators() []Iterator {
+	return []Iterator{a.left, a.right}
+}
+
+func (a *Arrow) ReplaceSubiterators(newSubs []Iterator) Iterator {
+	return &Arrow{left: newSubs[0], right: newSubs[1]}
+}
```
tstirrat15: Curious about the signature here - this seems like it could be a pure function? Is it more about code organization and ensuring that an interface is satisfied?

Oh. I guess it would be different if golang had overloads.

pkg/query/build_tree_test.go
```diff
@@ -154,10 +154,18 @@ func TestBuildTreeRecursion(t *testing.T) {
 	dsSchema, err := schema.BuildSchemaFromDefinitions(objectDefs, nil)
 	require.NoError(err)
 
-	// This should detect recursion and return an error
-	_, err = BuildIteratorFromSchema(dsSchema, "group", "member")
-	require.Error(err)
-	require.Contains(err.Error(), "recursive schema iterators are as yet unsupported")
+	// This should detect recursion and create a RecursiveIterator
+	it, err := BuildIteratorFromSchema(dsSchema, "group", "member")
```
miparnisari, line 166: the schema in this test seems to be 

```
definition user {} 

definition group {
   relation member: member
}
```

shouldn't it be 

```
definition user {} 

definition group {
   relation member: user | group#member
}
```

?

pkg/query/build_tree_test.go
```diff
@@ -572,7 +580,7 @@ func TestBuildTreeSubrelationHandling(t *testing.T) {
 
 	t.Run("Base Relation with Ellipsis Subrelation", func(t *testing.T) {
```
miparnisari, line 589: This test `"Base Relation with Ellipsis Subrelation"` seems to be exactly the same as `"Base Relation Without Subrelations Disabled"`. I mean, the former has `groupDef` in the schema, but it's not related to `docDef` at all and you don't build an iterator off it, so it's like it doesn't exist.

pkg/query/build_tree_test.go
```diff
@@ -572,7 +580,7 @@ func TestBuildTreeSubrelationHandling(t *testing.T) {
...
 		t.Parallel()
-		// Test base relation with ellipsis - should return just the base relation
+		// Test base relation with ellipsis - creates recursive arrow
```
miparnisari: why is this comment here? the arrow is in line 596

pkg/query/types.go
miparnisari: This Godoc made little sense to me so i asked Claude to provide an alternative. WDYT?

```go
  // Iterator is a Plan that forms a tree structure through its Subiterators, where the tree represents the 
  // query execution plan that can be traversed and optimized.
  // While Plan provides a read-only query interface, Iterator adds methods
  // for cloning, inspecting, and rebuilding iterator trees. This enables
  // query optimization by rewriting the tree.
  //
  // Implementations should form a composite tree structure where leaf nodes
  // (e.g., datastore scans) have no subiterators, and composite nodes
  // (e.g., unions, intersections) combine multiple subiterators.
  //
  // Most tree transformations should use the Walk helper function rather than
  // manually calling Subiterators and ReplaceSubiterators:
  //
  //    transformed := Walk(rootIter, func(subit Iterator) Iterator {
  //        // Transform logic here
  //        return it
  //    })

```

pkg/query/types.go
```diff
@@ -34,6 +34,15 @@ type Iterator interface {
...
 	// Clone does a deep-copy to duplicate the iterator tree at this point.
 	Clone() Iterator
+
+	// Subiterators returns the child iterators of this iterator, if any.
+	// Returns nil or empty slice for leaf iterators.
+	Subiterators() []Iterator
+
+	// ReplaceSubiterators returns a new iterator with the given subiterators replacing the current ones.
+	// For leaf iterators, this returns the iterator unchanged.
+	// For composite iterators, the length of newSubs should match the length of Subiterators().
```
miparnisari, line 51:
```suggestion
	// For composite iterators, the length of newSubs should match the length of Subiterators(). If not, the code will panic
```

do we want that behavior?

pkg/query/recursive.go
```diff
@@ -0,0 +1,184 @@
...
+import (
+	"fmt"
+)
+
+const defaultMaxRecursionDepth = 50
+
+// RecursiveIterator is the root controller that manages iterative deepening for recursive schemas.
+// It wraps an iterator tree that contains RecursiveSentinel sentinels, and executes the tree
+// repeatedly with increasing depth until a fixed point is reached or max depth is exceeded.
+type RecursiveIterator struct {
```
miparnisari, line 14: Are we missing this line? `var _ Iterator = &RecursiveIterator{}`

pkg/query/types.go
```diff
@@ -34,6 +34,15 @@ type Iterator interface {
 
 	// Clone does a deep-copy to duplicate the iterator tree at this point.
 	Clone() Iterator
+
+	// Subiterators returns the child iterators of this iterator, if any.
+	// Returns nil or empty slice for leaf iterators.
+	Subiterators() []Iterator
+
+	// ReplaceSubiterators returns a new iterator with the given subiterators replacing the current ones.
```
miparnisari, line 48: "ReplaceSubiterators returns a new iterator"

nit: this isn't always the case, eg for `f *FixedIterator` it doesn't return a new iterator, it returns `f`

pkg/query/recursive_sentinel.go
```diff
@@ -0,0 +1,83 @@
+package query
+
+import "fmt"
+
+// RecursiveSentinel is a sentinel iterator that marks recursion points during iterator tree construction.
+// It acts as a placeholder that will be replaced during execution by RecursiveIterator.
+type RecursiveSentinel struct {
```
miparnisari, line 13: missing line? `var _ Iterator = (*RecursiveSentinel)(nil)`

pkg/query/recursive_test.go
```diff
@@ -0,0 +1,120 @@
...
+func TestRecursiveIteratorEmptyBaseCase(t *testing.T) {
+	// Create a simple tree with sentinel that will return empty on depth 0
+	sentinel := NewRecursiveSentinel("folder", "view", false)
+	emptyIterator := NewEmptyFixedIterator()
+
+	union := NewUnion()
+	union.addSubIterator(emptyIterator)
+	union.addSubIterator(sentinel)
+
+	recursive := NewRecursiveIterator(union, []*RecursiveSentinel{sentinel})
```
miparnisari: this is confusing to me: `sentinel` is part of the `union` object, but you're also passing it as a separate object to the function 🤔

pkg/query/recursive.go
```diff
@@ -0,0 +1,184 @@
...
+			// Note: We need to unwrap recursively in case buildTreeAtDepth returns another RecursiveIterator
+			return unwrapRecursiveIterators(recIt.buildTreeAtDepth(depth), depth)
+		}
+		return it
+	})
+}
+
+// replaceSentinelsWithTree walks the iterator tree and replaces all RecursiveSentinel instances
+// with a clone of the provided replacement tree
+func (r *RecursiveIterator) replaceSentinelsWithTree(tree Iterator, replacement Iterator) Iterator {
```
miparnisari: this function isn't using `r` at all. also a better name might be `replaceSentinelsINTree`

pkg/query/recursive.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+const defaultMaxRecursionDepth = 50
+
+// RecursiveIterator is the root controller that manages iterative deepening for recursive schemas.
+// It wraps an iterator tree that contains RecursiveSentinel sentinels, and executes the tree
+// repeatedly with increasing depth until a fixed point is reached or max depth is exceeded.
+type RecursiveIterator struct {
+	templateTree Iterator
+	sentinels    []*RecursiveSentinel
```
miparnisari: seems that `r.sentinels` is not used in the more important functions. what am i missing, is this for a future PR?

pkg/query/datastore.go
```diff
@@ -251,3 +251,11 @@ func (r *RelationIterator) Explain() Explain {
 			r.base.Caveat() != "", r.base.Expiration()),
 	}
 }
+
+func (r *RelationIterator) Subiterators() []Iterator {
+	return nil
+}
+
+func (r *RelationIterator) ReplaceSubiterators(newSubs []Iterator) Iterator {
+	return r
```
tstirrat15: This is because the datastore iterator is a leaf iterator?

pkg/query/recursive.go
```diff
@@ -0,0 +1,184 @@
...
+
+// IterResourcesImpl implements iterative deepening for IterResources operations
+func (r *RecursiveIterator) IterResourcesImpl(ctx *Context, subject ObjectAndRelation) (PathSeq, error) {
+	return r.iterativeDeepening(ctx, func(ctx *Context, tree Iterator) (PathSeq, error) {
+		return ctx.IterResources(tree, subject)
+	})
+}
+
+// iterativeDeepening executes the core iterative deepening algorithm
+// It yields results directly, always running to maxDepth to find all valid paths
```
tstirrat15, line 47: Does this mean that it's eager when it's executed?

pkg/query/recursive.go
```diff
@@ -0,0 +1,184 @@
...
+		return it
+	})
+}
+
+// Clone creates a deep copy of the RecursiveIterator
+func (r *RecursiveIterator) Clone() Iterator {
+	// Clone template tree and sentinels
+	clonedSentinels := make([]*RecursiveSentinel, len(r.sentinels))
+	for i, s := range r.sentinels {
+		clonedSentinels[i] = s.Clone().(*RecursiveSentinel)
```
tstirrat15: Why is this cast necessary?

tstirrat15: And does it need to be guarded or do we know something the go compiler doesn't?

pkg/query/build_tree.go
```diff
@@ -33,27 +41,82 @@ func BuildIteratorFromSchema(fullSchema *schema.Schema, definitionName string, r
...
+
 	return result, nil
 }
 
 func (b *iteratorBuilder) buildIteratorFromSchemaInternal(definitionName string, relationName string, withSubRelations bool) (Iterator, error) {
 	id := fmt.Sprintf("%s#%s:%v", definitionName, relationName, withSubRelations)
-	if b.seen[id] {
-		return nil, errors.New("recursive schema iterators are as yet unsupported")
+	// Also check the opposite withSubRelations flag to catch recursion across flags
+	oppositeID := fmt.Sprintf("%s#%s:%v", definitionName, relationName, !withSubRelations)
```
tstirrat15: What situation does this refer to?

pkg/query/build_tree.go
```diff
@@ -33,27 +41,82 @@ func BuildIteratorFromSchema(fullSchema *schema.Schema, definitionName string, r
...
+
+	// Check if we're currently building this (true recursion)
+	// Check both with the same flag and opposite flag, since recursion can cross the boundary
+	if b.building[id] || b.building[oppositeID] {
```
tstirrat15: Is this what you were referring to that would change under mutual recursion?

pkg/query/recursive_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// CheckImpl should return empty
+	seq, err := sentinel.CheckImpl(ctx, []Object{{ObjectType: "folder", ObjectID: "folder1"}}, ObjectAndRelation{ObjectType: "user", ObjectID: "tom", Relation: "..."})
+	require.NoError(t, err)
+
+	count := 0
+	for range seq {
+		count++
+	}
+	require.Equal(t, 0, count)
```
tstirrat15: I wonder whether `require.Empty` would do this for you

pkg/query/build_tree.go
barakmich: Yep!

pkg/query/arrow.go
barakmich: Yep; and the interface will come in handy when we start rewriting trees

pkg/query/datastore.go
barakmich: Yep! In fact, it should probably cause an error.

pkg/query/recursive.go
barakmich: Removed in the future PR!

pkg/query/recursive.go
barakmich: It just creates a new iterator -- the evaluation of which is in the `execute` callback, below, which happens in the greater iteration step... so these only get deepened as needed

pkg/query/types.go
barakmich: Nah, updated to be all errors

pkg/query/build_tree.go
```diff
@@ -1,27 +1,35 @@
...
+	relationName   string
+}
+
 type iteratorBuilder struct {
-	schema           *schema.Schema
-	seen             map[string]bool
-	collectedCaveats []*core.ContextualizedCaveat // Collect caveats to combine with AND logic
+	schema             *schema.Schema
+	building           map[string]bool              // Track what's currently being built (call stack)
+	collectedCaveats   []*core.ContextualizedCaveat // Collect caveats to combine with AND logic
```
miparnisari, line 21: for a future PR: the only place where you update this is line 229. i suggest removing this field and instead having `buildBaseRelationIterator` return also the collected caveats.

otherwise it's hard to figure out _what mutates this field_ and you may assume that it's set for you, but it's not unless you call that function

miparnisari (approved):

PR #2622, opened 2025-10-17
Use prettier for yaml formatting and fix some yaml linting/formatting issues

tstirrat15:
Part of parting out #2455.

## Description
This is one of the pieces of #2455. I'm also very annoyed at `yamllint`, because it gives you a bunch of lint errors that can be easily fixed, but the [authors have no interest in that](https://github.com/adrienverge/yamllint/issues/535).

[prettier](https://prettier.io/), on the other hand, provides both formatting and format checking.

This PR adds prettier checking, runs prettier over the repo, and fixes a couple of other errors that surfaced.

Note that it does not completely replace `yamllint`, because `yamllint` checks things that prettier doesn't; this is mostly intended to have `prettier` act as a yaml formatter and format checker.

## Changes
* Reformat yaml files
* Fix some lint issues
* Add prettier check to magefile
## Testing
Review.


.github/workflows/build-test.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: "Build & Test"
-on:  # yamllint disable-line rule:truthy
```
tstirrat15, line 3: This space was previously requested by yamllint; I turned this off in favor of prettier's convention.

internal/services/integrationtesting/testconfigs/3letterrbac.yaml
```diff
@@ -1,21 +1,21 @@
 ---
 schema: |+
-    definition exm/usr {}
```
tstirrat15, line 3: Prettier indentation changes; shouldn't be semantically different.

internal/services/integrationtesting/testconfigs/arrowtosamesubject.yaml
```diff
@@ -25,5 +25,5 @@ relationships: >-
   folder:folder2#viewer@user:tom
 assertions:
   assertTrue:
-    - 'document:firstdoc#view@user:tom#...'
-    - 'document:seconddoc#view@user:tom#...'
+    - "document:firstdoc#view@user:tom#..."
+    - "document:seconddoc#view@user:tom#..."
```
tstirrat15, line 29: Quotation changes; previously this was something that yamllint yelled about but didn't fix. Prettier happily fixes them.

magefiles/lint.go
```diff
@@ -37,9 +38,13 @@ func (Lint) Yaml() error {
...
 		return err
 	}
-	return sh.RunV("docker", "run", "--rm",
+	// Run prettier for yaml formatting
+	prettierErr := sh.RunV("docker", "run", "--rm", "-w", "/src", "-v", fmt.Sprintf("%s:/src:ro", cwd), "ghcr.io/tmknom/dockerfiles/prettier:3.6.2", "-c", ".")
+	// Run yamllint for content checking
+	yamllintErr := sh.RunV("docker", "run", "--rm",
 		"-v", fmt.Sprintf("%s:/src:ro", cwd),
 		"cytopia/yamllint:1", "-c", "/src/.yamllint", "/src")
+	return errors.Join(prettierErr, yamllintErr)
```
tstirrat15, line 47: Running both of these linters in a single command.

.yamllint
```diff
@@ -4,7 +4,11 @@ yaml-files:
   - "*.yaml"
   - "*.yml"
   - ".yamllint"
+ignore:
+  - "dist"
+  - "steelthreadtesting"
```
tstirrat15, line 9: Steelthread tests are excluded because they're essentially snapshot tests and depend on yaml.v3's yaml generation behavior. We don't actually want to screw with them.

magefiles/lint.go
miparnisari: i'm confused, your PR description made it sound like you were getting rid of `yamllint` altogether..

tstirrat15: Yeah, I was going to and then I remembered that there are things that yamllint checks for and prettier doesn't cover (like malformed anchors). I decided to run them both so that we get as much autofix coverage as possible while still getting the benefit of yamllint catching errors.

PR #2627, opened 2025-10-20
chore: add chunk bytes library in prep for singleton schema work

josephschorr:
## Description

Adds a chunk bytes library to the common datastore code for easy storing and loading of byte data that will be chunked across multiple rows in the database, as some databases have soft or hard limits on the size of BLOB fields.

This will be used by singleton schemas to store schemas and, likely eventually, schema metadata

## Testing

Unit tests. This is very early and not used yet.

### DIFF
--- a/internal/datastore/common/chunkbytes.go
+++ b/internal/datastore/common/chunkbytes.go
@@ -1,0 +1,349 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	sq "github.com/Masterminds/squirrel"
+)
+
+// ChunkedBytesTransaction defines the interface for executing SQL queries within a transaction.
+type ChunkedBytesTransaction interface {
+	// ExecuteWrite executes an INSERT query.
+	ExecuteWrite(ctx context.Context, builder sq.InsertBuilder) error
+
+	// ExecuteDelete executes a DELETE query.
+	ExecuteDelete(ctx context.Context, builder sq.DeleteBuilder) error
+
+	// ExecuteUpdate executes an UPDATE query.
+	ExecuteUpdate(ctx context.Context, builder sq.UpdateBuilder) error
+}
+
+// ChunkedBytesExecutor defines the interface for creating transactions for chunked byte operations.
+type ChunkedBytesExecutor interface {
+	// BeginTransaction starts a new transaction for chunked byte operations.
+	BeginTransaction(ctx context.Context) (ChunkedBytesTransaction, error)
+
+	// ExecuteRead executes a SELECT query and returns the results as a map of chunk index to chunk data.
+	ExecuteRead(ctx context.Context, builder sq.SelectBuilder) (map[int][]byte, error)
+}
+
+// WriteMode defines how chunked data should be written.
+type WriteMode int
+
+const (
+	// WriteModeInsertWithTombstones inserts new chunks and marks old chunks with a tombstone.
+	// Requires TombstoneColumn to be set in config.
+	WriteModeInsertWithTombstones WriteMode = iota
+
+	// WriteModeDeleteAndInsert deletes all existing chunks for the key before inserting new ones.
+	// Useful for replacing data completely.
+	WriteModeDeleteAndInsert
+)
+
+// SQLByteChunkerConfig contains the configuration for creating a SQLByteChunker.
+type SQLByteChunkerConfig[T any] struct {
+	// TableName is the name of the table storing the chunked data.
+	TableName string
+
+	// NameColumn is the column name that stores the identifier for the byte data.
+	NameColumn string
+
+	// ChunkIndexColumn is the column name that stores the chunk index (0-based).
+	ChunkIndexColumn string
+
+	// ChunkDataColumn is the column name that stores the chunk bytes.
+	ChunkDataColumn string
+
+	// MaxChunkSize is the maximum size in bytes for each chunk.
+	MaxChunkSize int
+
+	// PlaceholderFormat is the placeholder format for SQL queries (e.g., sq.Question, sq.Dollar).
+	PlaceholderFormat sq.PlaceholderFormat
+
+	// Executor is the executor for running SQL queries.
+	Executor ChunkedBytesExecutor
+
+	// WriteMode defines how chunked data should be written (insert-with-tombstones or delete-and-insert).
+	WriteMode WriteMode
+
+	// CreatedAtColumn is the column name that stores when a row was created (alive timestamp/transaction ID).
+	// Required when WriteMode is WriteModeInsertWithTombstones.
+	CreatedAtColumn string
+
+	// DeletedAtColumn is the column name that stores when a row was deleted (tombstone timestamp/transaction ID).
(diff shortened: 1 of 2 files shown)


internal/datastore/common/chunkbytes.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// chunkData splits the data into chunks of maxChunkSize.
+func (c *SQLByteChunker[T]) chunkData(data []byte) [][]byte {
+	if len(data) == 0 {
+		return nil
+	}
+
+	numChunks := (len(data) + c.maxChunkSize - 1) / c.maxChunkSize
+	chunks := make([][]byte, 0, numChunks)
```
tstirrat15, line 338: For my own edification: when you allocate a slice of slices, is that essentially a slice of pointers?

internal/datastore/common/chunkbytes_test.go
```diff
@@ -0,0 +1,712 @@
...
+	require.Len(t, txn.updateQueries, 1)
+	require.Len(t, txn.deleteQueries, 0)
+
+	// Check the UPDATE query (tombstone old chunks that are alive)
+	updateSQL := txn.capturedSQL[0]
+	require.Contains(t, updateSQL, "UPDATE test_table")
+	require.Contains(t, updateSQL, "SET deleted_at = ?")
+	require.Contains(t, updateSQL, "WHERE name = ?")
+	require.Contains(t, updateSQL, "AND deleted_at = ?")
+	require.Equal(t, []any{createdAt, "test-key", ^uint64(0)}, txn.capturedArgs[0])
```
tstirrat15, line 243: Why is this an `any`? and what is `^uint64(0)` supposed to represent? Is that a clever way of writing maxint?

internal/datastore/common/chunkbytes_test.go
```diff
@@ -0,0 +1,712 @@
...
+		},
+		{
+			name:          "exact chunk size",
+			data:          []byte("12345"),
+			expectedCount: 1,
+			expectedSizes: []int{5},
+		},
+		{
+			name:          "multiple exact chunks",
+			data:          []byte("1234567890"),
```
tstirrat15, line 611: Do these happen to line up because utf-8 is one byte per char for ascii characters?

tstirrat15 (dismissed):
See comments, looks good to me

miparnisari (commented):
I've amended your PR title from `feat:` to `chore:` because, for customers, this isn't a new feature _yet_

internal/datastore/common/chunkbytes.go
```diff
@@ -0,0 +1,379 @@
...
+	nameColumn        string
+	chunkIndexColumn  string
+	chunkDataColumn   string
+	maxChunkSize      int
+	placeholderFormat sq.PlaceholderFormat
+	executor          ChunkedBytesExecutor
+	writeMode         WriteMode
+	createdAtColumn   string
+	deletedAtColumn   string
+	aliveValue        T
```
miparnisari: all these fields are already defined in `SQLByteChunkerConfig`. why not just hold a pointer to it? this would reduce duplicate code and prevent you from having to write

```
return &SQLByteChunker[T]{
		tableName:         config.TableName,
		nameColumn:        config.NameColumn,
		chunkIndexColumn:  config.ChunkIndexColumn,
		chunkDataColumn:   config.ChunkDataColumn,
		maxChunkSize:      config.MaxChunkSize,
		placeholderFormat: config.PlaceholderFormat,
		executor:          config.Executor,
		writeMode:         config.WriteMode,
		createdAtColumn:   config.CreatedAtColumn,
		deletedAtColumn:   config.DeletedAtColumn,
		aliveValue:        config.AliveValue,
	}
```

which is dangerous, because if you add a new configuration you have to remember to update that code too

internal/datastore/common/chunkbytes.go
```diff
@@ -0,0 +1,379 @@
...
+	DeletedAtColumn string
+
+	// AliveValue is the value used to indicate a row has not been deleted yet (typically max int).
+	// Required when WriteMode is WriteModeInsertWithTombstones.
+	AliveValue T
+}
+
+// SQLByteChunker provides methods for reading and writing byte data
+// that is chunked across multiple rows in a SQL table.
+type SQLByteChunker[T any] struct {
```
miparnisari, line 86:
```suggestion
type IntOrString interface {
	~uint64 | ~string
}
// SQLByteChunker provides methods for reading and writing byte data
// that is chunked across multiple rows in a SQL table.
type SQLByteChunker[T IntOrString] struct {
```

internal/datastore/common/chunkbytes.go
```diff
@@ -0,0 +1,379 @@
...
+func (c *SQLByteChunker[T]) reassembleChunks(chunks map[int][]byte) ([]byte, error) {
+	if len(chunks) == 0 {
+		return nil, errors.New("no chunks found")
+	}
+
+	// Validate that we have all chunks from 0 to N-1
+	maxIndex := -1
+	for index := range chunks {
+		if index > maxIndex {
+			maxIndex = index
```
miparnisari:
```suggestion
		maxIndex = max(maxIndex, index)
```

internal/datastore/common/chunkbytes.go
```diff
@@ -0,0 +1,379 @@
...
+	totalSize := 0
+	for _, chunk := range chunks {
+		totalSize += len(chunk)
+	}
+
+	// Reassemble
+	result := make([]byte, 0, totalSize)
+	for i := 0; i <= maxIndex; i++ {
+		result = append(result, chunks[i]...)
+	}
```
miparnisari:
```suggestion
	// Validate that we have all chunks from 0 to N-1 and calculate total size
	maxIndex := -1
	totalSize := 0
	for index := range chunks {
		maxIndex = max(maxIndex, index)
		totalSize += len(chunks[index])
	}

	// Reassemble, while checking for missing chunks
	result := make([]byte, 0, totalSize)
	for i := 0; i <= maxIndex; i++ {
		if _, exists := chunks[i]; !exists {
			return nil, fmt.Errorf("missing chunk at index %d", i)
		}
		result = append(result, chunks[i]...)
	}
```

internal/datastore/common/chunkbytes.go
josephschorr: Basically, yes

internal/datastore/common/chunkbytes_test.go
josephschorr: Its any because the arguments to the SQL input can be different values of different types. And yes, it is

internal/datastore/common/chunkbytes_test.go
josephschorr: Yes

internal/datastore/common/chunkbytes.go
josephschorr: Its not just int or string; its an `xid8` in Postgres, for example

miparnisari (approved):

PR #2628, opened 2025-10-20
Lint fixes part 5

tstirrat15:
Part of parting out #2455 

## Description
A handful more lint fixes.

## Changes
See commit descriptions

## Testing
Review. See that tests pass.


internal/datastore/proxy/schemacaching/watchingcache_test.go
tstirrat15, line 1: The changes in this file are to address the fact that `require.*` is using `t.FailNow()`, which only terminates the execution of the currently-running goroutine and can cause unpredictable test behavior when it's called from non-test goroutines. This refactors to pull the values we're asserting against out of the goroutines in question and hand them back to the test goroutine using buffered channels.

Issue #2579, opened 2025-09-26
Invalid memory address or nil pointer dereference


joostlawerman:
### What platforms are affected?

linux

### What architectures are affected?

amd64

### What SpiceDB version are you using?

v1.45.4

### Steps to Reproduce

Not sure, we are running spicedb in a docker container (3 instances per environment). Randomly one of the 3 instances crashes, one per +- two weeks. It is happening for some months now.

### Expected Result

No unexpected crashes/ panics, should be running stable.

### Actual Result

We get the following log, which seems to be a memory corruption of some sort.
```
Sep 24 05:54:51 panic: runtime error: invalid memory address or nil pointer dereference
Sep 24 05:54:51 [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x48b5e5]
Sep 24 05:54:51 goroutine 28 [running]:
google.golang.org/protobuf/encoding/protowire.AppendString(...)
    google.golang.org/protobuf/encoding/protowire/wire.go:479
google.golang.org/protobuf/internal/impl.appendStringValidateUTF8(...)
    google.golang.org/protobuf/internal/impl/codec_gen.go:5045 +0x91
google.golang.org/protobuf/internal/impl.(*MessageInfo).initOneofFieldCoders.func4(...)
    google.golang.org/protobuf/internal/impl/codec_field.go:96 +0x5d
google.golang.org/protobuf/internal/impl.appendMessageInfo(...)
    google.golang.org/protobuf/internal/impl/encode.go:215 +0x403
google.golang.org/protobuf/internal/impl.(*MessageInfo).marshalAppendPointer(...)
    google.golang.org/protobuf/internal/impl/codec_field.go:168 +0xa5
google.golang.org/protobuf/proto.Marshal(...)
    google.golang.org/protobuf/proto/encode.go:111 +0x50
go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp.(*client).UploadTraces(...)
    go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp/client.go:134 +0xb3
go.opentelemetry.io/otel/exporters/otlp/otlptrace.(*Exporter).ExportSpans(...)
    go.opentelemetry.io/otel/exporters/otlp/otlptrace/exporter.go:36 +0x63
go.opentelemetry.io/otel/sdk/trace.(*batchSpanProcessor).exportSpans(...)
[...]

PR #2629, opened 2025-10-21
fix: data races that occur when functions do not clone pointer object…

miparnisari:
…s before returning them

This is related to https://github.com/authzed/spicedb/issues/2579. I was able to reproduce one panic with running [this](https://github.com/authzed/spicedb/blob/main/CONTRIBUTING.md#test-a-local-change-with-docker-compose). But I haven't been able to reliably reproduce them, which makes me highly suspicious of the `singleflight` code.

When I ran this PR to reproduce the panic, I couldn't reproduce it anymore. However: **disclaimer**, I don't know if this PR is the end-game solution to those panics.

## Testing
If you undo the changes i made to the code, running the tests in this PR with `-race` should eventually show DATA RACE.

```
$ cd internal/dispatch/caching
$ go test . -race -run=TestConcurrentDebugInfoAccess -count=100
# github.com/authzed/spicedb/internal/dispatch/caching.test
ld: warning: '/private/var/folders/5p/328y159n1334j9czdhzl2ts40000gn/T/go-link-3411654366/000013.o' has malformed LC_DYSYMTAB, expected 98 undefined symbols to start at index 1626, found 95 undefined symbols starting at index 1626
ok      github.com/authzed/spicedb/internal/dispatch/caching    5.425s
```

### DIFF
--- a/internal/dispatch/dispatch.go
+++ b/internal/dispatch/dispatch.go
@@ -37,6 +37,7 @@ type Dispatcher interface {
 // Check interface describes just the methods required to dispatch check requests.
 type Check interface {
 	// DispatchCheck submits a single check request and returns its result.
+	// The result should be safe to access from multiple goroutines.
 	DispatchCheck(ctx context.Context, req *v1.DispatchCheckRequest) (*v1.DispatchCheckResponse, error)
 }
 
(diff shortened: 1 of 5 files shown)


internal/dispatch/dispatch.go
```diff
@@ -37,6 +37,7 @@ type Dispatcher interface {
 // Check interface describes just the methods required to dispatch check requests.
 type Check interface {
 	// DispatchCheck submits a single check request and returns its result.
+	// The result should be safe to access from multiple goroutines.
```
miparnisari, line 40: FYI i believe this should hold, but right now, I've no clue if it does, my gut tells me it doesn't. And if my gut is right, there could be _lots_ of places where we'd need to shove a `.cloneVT()` call 😬

PR #2632, opened 2025-10-21
perf: Have read relationships use dynamic index forcing

josephschorr:
This should ensure that, for recognized query shapes, the best index and sort order is selected for ReadRels

## Testing

Unit, datatstore and steelthread tests


pkg/datastore/options/options.go
```diff
@@ -28,6 +28,9 @@ const (
 	// BySubject might be quite a bit slower than ByResource, as relationships are
 	// indexed by resource.
 	BySubject
+
+	// ChooseEfficient lets the datastore choose the most efficient order based on the query shape and available indexes.
+	ChooseEfficient
```
miparnisari, line 33: why give the caller a new option, when `unsorted: lets the underlying datastore choose the order` exists? or should we update unsorted to just say `no order at all`?

internal/services/v1/relationships.go
```diff
@@ -258,7 +258,7 @@ func (ps *permissionServer) ReadRelationships(req *v1.ReadRelationshipsRequest,
 		ds,
 		dsFilter,
 		pageSize,
-		options.ByResource,
```
miparnisari, line 265: By reading https://buf.build/authzed/api/docs/main:authzed.api.v1#authzed.api.v1.PermissionsService.ReadRelationships it doesn't seem that we ever promised a particular return order. I'm curious why this code was the way it is with `options.ByResource`? was this an undocumented feature or just an accident?

can we update the API to be explicit that callers shouldn't expect any order?

internal/datastore/common/sql_test.go
```diff
@@ -1519,6 +1523,134 @@ func TestIndexHint(t *testing.T) {
...
+						require.Contains(t, sql, test.expectedOrderBy, "Expected ORDER BY clause not found in SQL")
+					}
+
+					wasRun = true
+					return nil, nil
+				},
+			}
+
+			var opts []options.QueryOptionsOption
+			if test.querySortOption != nil {
```
miparnisari, line 1621: nit: why the `if`? all the test cases have this parameter

internal/datastore/common/sql_test.go
```diff
@@ -1519,6 +1523,134 @@ func TestIndexHint(t *testing.T) {
...
+				WithPaginationFilterType(TupleComparison),
+				WithColumnOptimization(ColumnOptimizationOptionStaticValues),
+				WithNowFunction("NOW"),
+				WithExpirationDisabled(true),
+			)
+
+			hint := &fakeIndexingHintWithSort{sortOrder: test.hintSortOrder}
+			filterer := NewSchemaQueryFiltererForRelationshipsSelect(*schema, 100)
+			filterer = filterer.WithIndexingHint(hint)
+			filterer = filterer.FilterToResourceType("sometype")
```
miparnisari, line 1601: nit: if i remove this line the tests still pass

pkg/datastore/options/options.go
josephschorr: Because sometimes the caller knows better and letting it choose is important in those instances

internal/services/v1/relationships.go
josephschorr: Its necessary for cursoring; we need *some* stable order in order to cursor

internal/datastore/common/sql_test.go
josephschorr: they don't:

```
		{
			name:              "No sort option with hint ByResource keeps unsorted",
			hintSortOrder:     options.ByResource,
			querySortOption:   nil,
			expectedNoOrderBy: true,
		},
		```

internal/datastore/common/sql_test.go
josephschorr: Sure; but the SQL produced is technically invalid without it, so I figured leave it in

internal/services/v1/relationships.go
miparnisari: ok; should we update the API docs to say something like "the results will be sorted but the order will depend on the filter passed"?

miparnisari (approved):

internal/services/v1/relationships.go
josephschorr: I think we should just say they won't be stable-y sorted

PR #2635, opened 2025-10-22
chore: Lint fixes part 6

tstirrat15:
Part of parting out #2455 

## Description
More lint fixes. Includes fixes to a test that was incorrectly passing.

## Changes
Will annotate.

## Testing
Review. See that tests pass.

### DIFF
--- a/internal/datastore/postgres/strictreader_test.go
+++ b/internal/datastore/postgres/strictreader_test.go
@@ -28,69 +28,83 @@ func (mc fakeQuerier) ExecFunc(_ context.Context, _ func(ctx context.Context, ta
 	return mc.err
 }
 
-func TestStrictReaderDetectsLagErrors(t *testing.T) {
+/*
+NOTE: these tests aren't in a tabular format because we want to
+assert require.ErrorsAs against different kinds of errors,
+but you can't write a tabular test that takes different kinds of
+errors in a struct because you lose type information.
+*/
+
+func TestStrictReaderDetectsLagErrors_NoErrors(t *testing.T) {
 	mc := fakeQuerier{}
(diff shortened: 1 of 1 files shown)


internal/datastore/proxy/schemacaching/intervaltracker_test.go
```diff
@@ -179,12 +179,12 @@ func TestIntervalTrackerGC(t *testing.T) {
 
 	// GC anything older than 5s, which shouldn't change anything.
 	result := tracker.removeStaleIntervals(5 * time.Second)
-	require.False(t, false, result)
```
miparnisari, line 182: 😆

tstirrat15: 💀

internal/datastore/postgres/strictreader_test.go
```diff
@@ -28,69 +28,83 @@ func (mc fakeQuerier) ExecFunc(_ context.Context, _ func(ctx context.Context, ta
...
+func TestStrictReaderDetectsLagErrors_NoErrors(t *testing.T) {
 	mc := fakeQuerier{}
 	reader := strictReaderQueryFuncs{
 		wrapped: mc,
 	}
-
-	cases := []struct {
-		name string
-		in   error
-		want error
```
tstirrat15, line 40: This being defined as `error` meant that the `&expectedErr` that we passed to `require.ErrorAs` was always a `*error` (as in a pointer to the base error type), which meant that any error would pass the comparison. I was able to verify this by writing a test with two completely different errors that should have failed and it passed.

miparnisari (approved):

PR #2637, opened 2025-10-22
chore: Lint fixes part 7

tstirrat15:
Part of parting out #2455.

## Description
More lint fixes!

## Changes
Will annotate.

## Testing
See that tests pass.


internal/dispatch/graph/check_test.go
```diff
@@ -310,7 +310,7 @@ func TestCheckPermissionOverSchema(t *testing.T) {
 		{
 			"basic union",
 			`definition user {}
-
```
tstirrat15, line 313: I think this is whitespace trimming introduced by OP's editor. It should be fine.

internal/dispatch/singleflight/singleflight_test.go
```diff
@@ -187,29 +187,35 @@ func TestSingleFlightDispatcherCancelation(t *testing.T) {
 	disp := New(mockDispatcher{f: f}, &keys.DirectKeyHandler{})
 	wg := sync.WaitGroup{}
 	wg.Add(3)
+	errs := make(chan error, 3)
```
tstirrat15, line 190: This is another set of testifyrequire fixes around not using require inside of a goroutine. We collect the errors in channels and then examine them.

internal/graph/check_test.go
```diff
@@ -78,7 +78,7 @@ func TestAsyncDispatch(t *testing.T) {
 				defer l.Unlock()
 
 				letFinish.Broadcast()
-				return tc.numRequests == dispatchedCount && tc.numRequests == completedCount
+				return dispatchedCount == tc.numRequests && completedCount == tc.numRequests
```
tstirrat15, line 81: Not sure why the linter didn't like this, but it's equivalent.

miparnisari (approved):

PR #2638, opened 2025-10-22
chore: Enable protogetter linter and run fixes

### LINKED ISSUE #2633: Introduce protogetter linter
It's specifically meant to avoid panics: https://github.com/ghostiam/protogetter

Might help with https://github.com/authzed/spicedb/issues/2496.

tstirrat15:
Fixes #2633 
Part of #2579 

## Description
If you have a deeply nested proto object, calling `a.GetSet().GetOf().GetGetters()` is safer than referencing `some.Chain.Of.Properties`, because the first will return `nil` if there's a nil in the chain, where the second will panic. `protogetter` lints for this and should help address #2579.

## Changes
* Add protogetter linter
* Run fixes
## Testing
Review. See that tests pass.

### DIFF
--- a/pkg/promutil/promutil.go
+++ b/pkg/promutil/promutil.go
@@ -24,5 +24,9 @@ func MustCounterValue(m prometheus.Metric) float64 {
 	if err := m.Write(&written); err != nil {
 		panic("failed to read Prometheus counter: " + err.Error())
 	}
-	return *written.Counter.Value
+	counter := written.GetCounter()
+	if counter == nil {
+		panic("counter was nil")
+	}
+	return counter.GetValue()
 }
--- a/pkg/promutil/promutil_test.go
+++ b/pkg/promutil/promutil_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMustCounterValue(t *testing.T) {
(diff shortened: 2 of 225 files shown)


pkg/promutil/promutil.go
```diff
@@ -24,5 +24,9 @@ func MustCounterValue(m prometheus.Metric) float64 {
 	if err := m.Write(&written); err != nil {
 		panic("failed to read Prometheus counter: " + err.Error())
 	}
-	return *written.Counter.Value
+	counter := written.GetCounter()
+	if counter == nil {
+		panic("counter was nil")
+	}
+	return counter.GetValue()
```
tstirrat15, line 31: This is a meaningful change - we were actually depending on this to panic if `Counter` was nil, so the nil-safety fix kinda broke the functionality.

pkg/promutil/promutil_test.go
```diff
@@ -13,29 +14,23 @@ func TestMustCounterValue(t *testing.T) {
 	})
 
 	result := MustCounterValue(counter)
-	if result != 0.0 {
-		t.Errorf("Initial counter value should be 0.0, got %v", result)
-	}
+	require.Equal(t, 0.0, result, "Initial counter value should be 0.0")
```
tstirrat15, line 17: Refactoring these tests to use `require` to bring them in line with the rest of the codebase.

tstirrat15 (commented):
These are the only two files that I touched manually

PR #2639, opened 2025-10-22
feat: support nested recursive iterators

barakmich:
Stacked on top of #2621

## Description

Fixes the last consistency test that still fails. `walkbackandforth.yaml` is a useful, deeply nested, mutually recursive test. By tagging both ends of the recursive iterator with the name of the recursive rule we're expanding, we can correctly deepen each recursive rule independently.

## Testing

This brings our consistency test suite up to 69/69 passing Check tests. We have reached feature parity!

### DIFF
--- a/pkg/query/build_tree_test.go
+++ b/pkg/query/build_tree_test.go
@@ -936,15 +936,23 @@ func TestBuildTreeMutualRecursionSentinelFiltering(t *testing.T) {
 		}
 		walk(it)
 
-		// We should have at least one RecursiveIterator
-		require.NotEmpty(recursiveIterators, "should find at least one RecursiveIterator")
+		for _, recursiveIterator := range recursiveIterators {
+			require.NotEmpty(recursiveIterator.definitionName)
+			require.NotEmpty(recursiveIterator.relationName)
 
-		// Verify each RecursiveIterator has matching definition/relation names
-		for _, rec := range recursiveIterators {
-			// Each sentinel in this RecursiveIterator should match its definition/relation
-			// or be from a different schema (which is fine - they'll be left alone)
-			require.NotEmpty(rec.definitionName, "RecursiveIterator should have definitionName set")
-			require.NotEmpty(rec.relationName, "RecursiveIterator should have relationName set")
+			var recursiveSentinels []*RecursiveSentinel
+			_, _ = Walk(recursiveIterator.templateTree, func(it Iterator) (Iterator, error) {
+				if sentinel, ok := it.(*RecursiveSentinel); ok {
+					recursiveSentinels = append(recursiveSentinels, sentinel)
+				}
+				return it, nil
+			})
+
+			// Verify each sentinel matches THIS RecursiveIterator
+			for _, recursiveSentinel := range recursiveSentinels {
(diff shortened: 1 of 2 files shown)


pkg/query/recursive.go
```diff
@@ -153,12 +157,18 @@ func unwrapRecursiveIterators(tree Iterator, depth int) (Iterator, error) {
 	})
 }
 
-// replaceSentinelsWithTree walks the iterator tree and replaces all RecursiveSentinel instances
-// with a clone of the provided replacement tree
-func replaceSentinelsInTree(tree Iterator, replacement Iterator) (Iterator, error) {
+// replaceSentinelsWithTree walks the iterator tree and replaces RecursiveSentinel instances
```
miparnisari:
```suggestion
// replaceSentinelsInTree walks the iterator tree and replaces RecursiveSentinel instances
```

pkg/query/build_tree_test.go
```diff
@@ -847,3 +847,104 @@ func TestBuildTreeWildcardIterator(t *testing.T) {
...
+			if rec, ok := it.(*RecursiveIterator); ok {
+				recursiveIterators = append(recursiveIterators, rec)
+			}
+			for _, sub := range it.Subiterators() {
+				walk(sub)
+			}
+		}
+		walk(it)
+
+		// We should have at least one RecursiveIterator
```
miparnisari: why "at least"? can we not assert that it's exactly of length 1?

pkg/query/build_tree_test.go
```diff
@@ -847,3 +847,104 @@ func TestBuildTreeWildcardIterator(t *testing.T) {
...
+		require.NotEmpty(recursiveIterators, "should find at least one RecursiveIterator")
+
+		// Verify each RecursiveIterator has matching definition/relation names
+		for _, rec := range recursiveIterators {
+			// Each sentinel in this RecursiveIterator should match its definition/relation
+			// or be from a different schema (which is fine - they'll be left alone)
+			require.NotEmpty(rec.definitionName, "RecursiveIterator should have definitionName set")
+			require.NotEmpty(rec.relationName, "RecursiveIterator should have relationName set")
+		}
```
miparnisari: are these assertions right? i thought you wanted to compare RecursiveIterators and RecursiveSentinels... how about this?

```go
		for _, recursiveIterator := range recursiveIterators {
			require.NotEmpty(recursiveIterator.definitionName)
			require.NotEmpty(recursiveIterator.relationName)

			var recursiveSentinels []*RecursiveSentinel
			_, _ = Walk(recursiveIterator.templateTree, func(it Iterator) (Iterator, error) {
				if sentinel, ok := it.(*RecursiveSentinel); ok {
					recursiveSentinels = append(recursiveSentinels, sentinel)
				}
				return it, nil
			})

			// Verify each sentinel matches THIS RecursiveIterator
			for _, recursiveSentinel := range recursiveSentinels {
				require.Equal(recursiveIterator.definitionName, recursiveSentinel.DefinitionName())
				require.Equal(recursiveIterator.relationName, recursiveSentinel.RelationName())
			}
		}
```

miparnisari (approved):

PR #2640, opened 2025-10-22
fix: properly rewrite errors for watch api

miparnisari:
## Changes

1. We are seeing these logs: `"grpc.code":"Internal","grpc.error":"rpc error: code = Internal desc = watch error: revision has expired"`. This is wrong: `revision has expired` should NOT be `internal`. It fires alarms.
2. Wrote more tests, which increased coverage (see current coverage is pretty bad: https://app.codecov.io/gh/authzed/spicedb/blob/main/internal%2Fservices%2Fshared%2Ferrors.go)

### DIFF
--- a/internal/services/shared/errors.go
+++ b/internal/services/shared/errors.go
@@ -14,13 +14,11 @@ import (
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 
 	"github.com/authzed/spicedb/internal/datastore/crdb/pool"
-	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/internal/graph"
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/internal/sharederrors"
 	"github.com/authzed/spicedb/pkg/cursor"
 	"github.com/authzed/spicedb/pkg/datastore"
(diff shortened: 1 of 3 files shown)


internal/services/shared/errors.go
```diff
@@ -168,14 +165,6 @@ func rewriteError(ctx context.Context, err error, config *ConfigForErrors) error
 	case errors.As(err, &relationNotFoundError):
 		return spiceerrors.WithCodeAndReason(err, codes.FailedPrecondition, v1.ErrorReason_ERROR_REASON_UNKNOWN_RELATION_OR_PERMISSION)
 
-	case errors.As(err, &maxDepthError):
```
miparnisari, line 171: FYI `dispatch.MaxDepthExceededError` implements GRPCStatus so it gets caught in line 142 above

barakmich (approved):

internal/services/shared/errors.go
```diff
@@ -148,9 +146,8 @@ func rewriteError(ctx context.Context, err error, config *ConfigForErrors) error
 	var relationNotFoundError sharederrors.UnknownRelationError
 
 	var compilerError compiler.BaseCompilerError
-	var sourceError spiceerrors.WithSourceError
+	var sourceError *spiceerrors.WithSourceError
```
miparnisari, line 149: FYI so that `case errors.As(err, &sourceError):` actually works...

PR #2641, opened 2025-10-23
test: add tests for error URLs

miparnisari:
Companion: https://github.com/authzed/spicedb-short-urls/pull/11

### DIFF
--- a/internal/sharederrors/error_help_links_test.go
+++ b/internal/sharederrors/error_help_links_test.go
@@ -1,0 +1,36 @@
+package sharederrors
+
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHelpLinks(t *testing.T) {
+	t.Parallel()
+
+	c := http.Client{}
+	t.Cleanup(c.CloseIdleConnections)
+
+	for _, helpURL := range allErrorHelpUrls {
+		t.Run(helpURL, func(t *testing.T) {
+			t.Parallel()
+
+			require.Eventually(t, func() bool {
+				get, err := c.Get(helpURL)
+				if err != nil {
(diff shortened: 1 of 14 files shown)


internal/sharederrors/error_help_links_test.go
```diff
@@ -0,0 +1,30 @@
...
+		t.Run(helpURL, func(t *testing.T) {
+			t.Parallel()
+
+			require.Eventually(t, func() bool {
+				get, err := c.Get(helpURL)
+				require.NoError(t, err)
+				_ = get.Body.Close()
+				require.Equal(t, http.StatusOK, get.StatusCode)
+				return true
+			}, 30*time.Second, 100*time.Millisecond)
```
tstirrat15: Is `Eventually` spawning a goroutine? Like do we need to worry about nesting `require` calls here, or is `Eventually` just polling until the test completes?

Do we need an `Eventually` here? Or is the idea that it's going to help avoid network request issues?

Last thought: are we gonna get rate limited?

tstirrat15 (dismissed):
LGTM pending questions

internal/sharederrors/error_help_links_test.go
miparnisari: looking deeper, it does spawn a goroutine 👎 

i wanted to prevent flaky tests 😆

tstirrat15: Yeah, that makes sense. I'd also think that it'd be weird if you couldn't make `require` calls inside of `require.Eventually` - I wonder if they're doing special handling here.

tstirrat15: Naw, looks like you can't. That's weird.

tstirrat15 (approved):
LGTM

PR #2643, opened 2025-10-23
Further operations and improvements to Schema V2 library

josephschorr:
This PR enhances the schema v2 package with comprehensive support for arrow operations and traversal capabilities.

- Arrow Traversal and Visiting
- Functioned Arrow Support
- Arrow Flattening

### DIFF
--- a/pkg/schema/v2/operations.go
+++ b/pkg/schema/v2/operations.go
@@ -7,6 +7,23 @@ type Operation interface {
 	clone() Operation
 }
 
+// ArrowOperation is an interface implemented by all arrow-based operations (both standard and functioned arrows).
+// This includes ArrowReference, FunctionedArrowReference, ResolvedArrowReference, and ResolvedFunctionedArrowReference.
+type ArrowOperation interface {
+	Operation
+
+	// Left returns the relation on the resource (left side of the arrow).
+	Left() string
+
+	// Right returns the relation/permission on the subject (right side of the arrow).
+	Right() string
+
+	// Function returns the function type applied to the arrow.
+	// For standard arrows (->), this returns FunctionTypeAny.
+	// For functioned arrows (.any()/.all()), this returns the specific function type.
+	Function() FunctionType
+}
+
 // RelationReference is an Operation that is a simple relation, such as `permission foo = bar`.
 type RelationReference struct {
 	// relationName is the name of the relation or permission being referenced.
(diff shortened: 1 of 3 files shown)


pkg/query/build_tree.go
```diff
@@ -194,20 +194,20 @@ func (b *iteratorBuilder) buildIteratorFromOperation(p *schema.Permission, op sc
 
 		return NewExclusion(mainIt, excludedIt), nil
 
-	case *schema.FunctionedTuplesetOperation:
-		rel, ok := p.Parent().Relations()[perm.TuplesetRelation()]
+	case *schema.FunctionedArrowReference:
```
barakmich, line 197: It wasn't obvious to me that this is always supposed to be an arrow. I took it as a general function call, that happened to return arrows at the moment. Couldn't a functioned tupleset also be some sort of selection function, or multiple arrows (perhaps morphism style) or.... any number of things in the future?

pkg/schema/v2/convert.go
```diff
@@ -234,10 +234,10 @@ func convertChild(child *corev1.SetOperation_Child) (Operation, error) {
 		if err != nil {
 			return nil, err
 		}
-		return &FunctionedTuplesetOperation{
-			tuplesetRelation: childType.FunctionedTupleToUserset.GetTupleset().GetRelation(),
-			function:         functionType,
-			computedRelation: childType.FunctionedTupleToUserset.GetComputedUserset().GetRelation(),
+		return &FunctionedArrowReference{
```
barakmich, line 237: Oh, I see, this is so that it can be a flattened thing and fit with the ArrowOperation interface (which I dig) 

So I guess my suggestion is keeping the FunctionedTupleset construct around, with a way of converting it to the intersection arrow we've got, or simply calling it what it is -- an intersection arrow -- and punting on the function side of the schema language entirely (as we only have two functions at the moment).

pkg/query/build_tree.go
josephschorr: The standard tupleset is what we call arrows externally, so the functioned one was intended to be an arrow with different... functions. If we add another non-arrow construct in the future, we'd likely add a new proto message type

pkg/schema/v2/convert.go
josephschorr: See above

PR #2644, opened 2025-10-23
Lint fixes part 8

tstirrat15:
Part of parting out #2455.

## Description
More lint fixes for when we re-enable linters

## Changes
Will annotate.

## Testing
Review.

### DIFF
--- a/pkg/caveats/types/ipaddress.go
+++ b/pkg/caveats/types/ipaddress.go
@@ -37,8 +37,7 @@ func (ipa IPAddress) SerializedString() string {
 }
 
 func (ipa IPAddress) ConvertToNative(typeDesc reflect.Type) (any, error) {
-	switch typeDesc {
-	case reflect.TypeOf(""):
+	if typeDesc == reflect.TypeOf("") {
 		return ipa.ip.String(), nil
 	}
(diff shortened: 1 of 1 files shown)


internal/services/integrationtesting/cert_test.go
```diff
@@ -180,10 +180,10 @@ func TestCertRotation(t *testing.T) {
 	).Complete(ctx)
 	require.NoError(t, err)
 
-	wait := make(chan struct{}, 1)
+	wait := make(chan error, 1)
 	go func() {
-		require.NoError(t, srv.Run(ctx))
-		wait <- struct{}{}
```
tstirrat15, line 186: We can't make this assertion inside the goroutine, so we instead return the error from the channel and assert about it.

internal/telemetry/reporter_test.go
```diff
@@ -58,14 +58,14 @@ func TestWriteTimeSeries(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			var handlerWriteErr error
```
tstirrat15, line 61: This is similar to the no-require-in-goroutines lint fix, since http handlers are run inside of goroutines.

tstirrat15 (commented):
See comments

PR #2645, opened 2025-10-23
test: refactor GC and GC tests to be more reliable using synctest

tstirrat15:
Parting out #2455

## Description
Part of getting the `testifyrequire` linter enabled.

## TODO
- [x] Address this comment: https://github.com/authzed/spicedb/pull/2455#discussion_r2303544841

## Changes
Will annotate.

## Testing
Review

### DIFF
--- a/internal/datastore/common/gc.go
+++ b/internal/datastore/common/gc.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/cenkalti/backoff/v4"
+	"github.com/cenkalti/backoff/v5"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog"
 
@@ -149,14 +149,20 @@ var MaxGCInterval = 60 * time.Minute
 // StartGarbageCollector loops forever until the context is canceled and
 // performs garbage collection on the provided interval.
 func StartGarbageCollector(ctx context.Context, collectable GarbageCollectableDatastore, interval, window, timeout time.Duration) error {
-	return startGarbageCollectorWithMaxElapsedTime(ctx, collectable, interval, window, 0, timeout, gcFailureCounter)
+	return runOnIntervalWithBackoff(ctx, func() error {
+		// NOTE: we're okay using the parent context here because the
+		// callers of this function create a dedicated garbage collection
+		// context anyway, which is only cancelled when the ds is closed.
+		gcCtx, cancel := context.WithTimeout(ctx, timeout)
+		defer cancel()
+		return RunGarbageCollection(gcCtx, collectable, window)
+	}, interval, timeout, gcFailureCounter)
 }
 
-func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, collectable GarbageCollectableDatastore, interval, window, maxElapsedTime, timeout time.Duration, failureCounter prometheus.Counter) error {
+func runOnIntervalWithBackoff(ctx context.Context, taskFn func() error, interval, timeout time.Duration, failureCounter prometheus.Counter) error {
 	backoffInterval := backoff.NewExponentialBackOff()
 	backoffInterval.InitialInterval = interval
 	backoffInterval.MaxInterval = max(MaxGCInterval, interval)
-	backoffInterval.MaxElapsedTime = maxElapsedTime
 	backoffInterval.Reset()
 
 	nextInterval := interval
@@ -175,11 +181,10 @@ func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, collectable Ga
 		case <-time.After(nextInterval):
 			log.Ctx(ctx).Info().
 				Dur("interval", nextInterval).
-				Dur("window", window).
 				Dur("timeout", timeout).
 				Msg("running garbage collection worker")
 
-			err := RunGarbageCollection(collectable, window, timeout)
+			err := taskFn()
 			if err != nil {
 				failureCounter.Inc()
 				nextInterval = backoffInterval.NextBackOff()
@@ -200,10 +205,7 @@ func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, collectable Ga
 }
 
 // RunGarbageCollection runs garbage collection for the datastore.
-func RunGarbageCollection(collectable GarbageCollectableDatastore, window, timeout time.Duration) error {
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-
+func RunGarbageCollection(ctx context.Context, collectable GarbageCollectableDatastore, window time.Duration) error {
 	ctx, span := tracer.Start(ctx, "RunGarbageCollection")
 	defer span.End()
 
--- a/internal/datastore/common/gc_test.go
+++ b/internal/datastore/common/gc_test.go
@@ -2,15 +2,18 @@ package common
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	promclient "github.com/prometheus/client_model/go"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
 
 	"github.com/authzed/spicedb/internal/datastore/revisions"
 	"github.com/authzed/spicedb/pkg/datastore"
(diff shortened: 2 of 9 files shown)


internal/datastore/common/gc.go
```diff
@@ -200,10 +206,7 @@ func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, collectable Ga
 }
 
 // RunGarbageCollection runs garbage collection for the datastore.
-func RunGarbageCollection(collectable GarbageCollectableDatastore, window, timeout time.Duration) error {
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-
+func RunGarbageCollection(ctx context.Context, collectable GarbageCollectableDatastore, window time.Duration) error {
```
tstirrat15, line 208: This is the big change - passing in a context means that we can test this function's logic using cancellation directly rather than depending on timing. It also means that we're creating the contexts with cancellation in different places than we used to.

internal/datastore/common/gc.go
```diff
@@ -149,14 +149,20 @@ var MaxGCInterval = 60 * time.Minute
...
+		// callers of this function create a dedicated garbage collection
+		// context anyway, which is only cancelled when the ds is closed.
+		gcCtx, cancel := context.WithTimeout(ctx, timeout)
+		defer cancel()
+		return RunGarbageCollection(gcCtx, collectable, window)
+	}, interval, window, timeout, gcFailureCounter)
 }
 
-func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, collectable GarbageCollectableDatastore, interval, window, maxElapsedTime, timeout time.Duration, failureCounter prometheus.Counter) error {
+func runPeriodicallyWithBackoff(ctx context.Context, taskFn func() error, interval, window, timeout time.Duration, failureCounter prometheus.Counter) error {
```
tstirrat15: This is another big change - the `runPeriodicallyWithBackoff` logic is mostly used by GC, but there's nothing GC-specific about it, so we can test it separately and then let them be composed.

internal/datastore/common/gc.go
```diff
@@ -149,14 +149,20 @@ var MaxGCInterval = 60 * time.Minute
 // StartGarbageCollector loops forever until the context is canceled and
 // performs garbage collection on the provided interval.
 func StartGarbageCollector(ctx context.Context, collectable GarbageCollectableDatastore, interval, window, timeout time.Duration) error {
-	return startGarbageCollectorWithMaxElapsedTime(ctx, collectable, interval, window, 0, timeout, gcFailureCounter)
+	return runPeriodicallyWithBackoff(ctx, func() error {
```
tstirrat15: And this is where we're actually using it - we thunk the garbage collection logic.

internal/datastore/common/gc.go
```diff
@@ -149,14 +149,20 @@ var MaxGCInterval = 60 * time.Minute
...
+		return RunGarbageCollection(gcCtx, collectable, window)
+	}, interval, window, timeout, gcFailureCounter)
 }
 
-func startGarbageCollectorWithMaxElapsedTime(ctx context.Context, collectable GarbageCollectableDatastore, interval, window, maxElapsedTime, timeout time.Duration, failureCounter prometheus.Counter) error {
+func runPeriodicallyWithBackoff(ctx context.Context, taskFn func() error, interval, window, timeout time.Duration, failureCounter prometheus.Counter) error {
 	backoffInterval := backoff.NewExponentialBackOff()
 	backoffInterval.InitialInterval = interval
 	backoffInterval.MaxInterval = max(MaxGCInterval, interval)
-	backoffInterval.MaxElapsedTime = maxElapsedTime
```
tstirrat15, line 159: This is no longer an option in the new version of `backoff`. I upgraded to the newer version of `backoff` in part because it plays more nicely with `synctest` than the old version did. It otherwise behaves the same way for our purposes.

internal/datastore/common/gc_test.go
```diff
@@ -181,50 +184,45 @@ func (d revisionErrorDeleter) DeleteExpiredRels() (int64, error) {
...
-	ctx, cancel := context.WithCancel(t.Context())
-	defer cancel()
-	go func() {
-		gc := newFakeGCStore(alwaysErrorDeleter{})
-		require.Error(t, startGarbageCollectorWithMaxElapsedTime(ctx, gc, 100*time.Millisecond, 1*time.Second, 1*time.Nanosecond, 1*time.Minute, localCounter))
-	}()
-	time.Sleep(200 * time.Millisecond)
-	cancel()
+	errCh := make(chan error, 1)
+	synctest.Test(t, func(t *testing.T) {
```
tstirrat15, line 200: This gives us deterministic behavior of this test.

internal/datastore/common/gc_test.go
```diff
@@ -181,50 +184,45 @@ func (d revisionErrorDeleter) DeleteExpiredRels() (int64, error) {
...
+	})
+	require.Error(t, <-errCh)
 
 	metrics, err := reg.Gather()
 	require.NoError(t, err)
 	var mf *promclient.MetricFamily
 	for _, metric := range metrics {
 		if metric.GetName() == "spicedb_datastore_gc_failure_total" {
 			mf = metric
+			break
```
tstirrat15, line 220: Don't need to continue iterating once we find the one that we want.

internal/datastore/common/gc_test.go
```diff
@@ -181,50 +184,45 @@ func (d revisionErrorDeleter) DeleteExpiredRels() (int64, error) {
...
 		}
 	}
-	require.Greater(t, *(mf.GetMetric()[0].Counter.Value), 100.0, "MaxElapsedTime=1ns did not cause backoff to get ignored")
-
-	localCounter = prometheus.NewCounter(gcFailureCounterConfig)
```
tstirrat15, line 208: This logic got pushed upwards.

internal/datastore/common/gc_test.go
```diff
@@ -238,19 +236,25 @@ func TestGCFailureBackoffReset(t *testing.T) {
...
-
-	go func() {
-		interval := 10 * time.Millisecond
-		window := 10 * time.Second
-		timeout := 1 * time.Minute
-
-		require.Error(t, StartGarbageCollector(ctx, gc, interval, window, timeout))
-	}()
+	errCh := make(chan error, 1)
+	synctest.Test(t, func(t *testing.T) {
```
tstirrat15, line 245: Using synctest here as well.

pkg/cmd/datastore.go
```diff
@@ -60,13 +60,14 @@ func NewGCDatastoreCommand(programName string, cfg *datastore.Config) *cobra.Com
 		Long:    "Executes garbage collection against the datastore. Deletes stale relationships, expired relationships, and stale transactions.",
 		PreRunE: server.DefaultPreRunE(programName),
 		RunE: termination.PublishError(func(cmd *cobra.Command, args []string) error {
-			return executeGC(cfg)
+			return executeGC(cmd, cfg)
```
tstirrat15: Passing in the command so that we can use its context.

internal/datastore/common/gc.go
miparnisari: for a followup, can we add godoc for what `window` is?

internal/datastore/common/gc.go
miparnisari: > there's nothing GC-specific about it

the log messages 😛

miparnisari: - `window` is not used except for a log, can we get rid of it?

- also maybe instead of `runPeriodicallyWithBackoff` we could rename it to `runOnIntervalWithBackoff` so that it's clear what `interval` is for

internal/datastore/common/gc_test.go
```diff
@@ -181,50 +184,45 @@ func (d revisionErrorDeleter) DeleteExpiredRels() (int64, error) {
...
-
-	metrics, err = reg.Gather()
-	require.NoError(t, err)
-	for _, metric := range metrics {
-		if metric.GetName() == "spicedb_datastore_gc_failure_total" {
-			mf = metric
-		}
-	}
-	require.Less(t, *(mf.GetMetric()[0].Counter.Value), 3.0, "MaxElapsedTime=0 should have not caused backoff to get ignored")
+	// We expect about 5 failures; the behavior of the library means that there's some wiggle room here.
```
miparnisari, line 223: > // We expect about 5 failures

could you explain the math behind this?

internal/datastore/common/gc_test.go
```diff
@@ -238,19 +236,25 @@ func TestGCFailureBackoffReset(t *testing.T) {
...
+			cancel()
+		})
+		go func() {
+			interval := 10 * time.Millisecond
+			window := 10 * time.Second
+			timeout := 1 * time.Minute
+
+			errCh <- StartGarbageCollector(ctx, gc, interval, window, timeout)
+		}()
+		time.Sleep(500 * time.Millisecond)
```
miparnisari, line 260: magic number?

internal/datastore/common/gc_test.go
```diff
@@ -264,20 +268,29 @@ func TestGCFailureBackoffReset(t *testing.T) {
...
+
+			errCh <- StartGarbageCollector(ctx, gc, interval, window, timeout)
+		}()
+		time.Sleep(30 * time.Millisecond)
+		hasRunChan <- gc.HasGCRun()
+		cancel()
+		synctest.Wait()
+	})
+	require.Error(t, <-errCh)
+	require.False(t, <-hasRunChan, "GC should not have run")
```
miparnisari:
```suggestion
	require.False(t, <-hasRunChan, "GC should not have run because ...")
```

internal/datastore/common/gc.go
tstirrat15: Well sure, but those aren't core to what the function does :P

internal/datastore/common/gc_test.go
tstirrat15: Boy, I had to dig for that one.

miparnisari (approved):

PR #2646, opened 2025-10-23
feat: add Memory Protection Middleware

ecordell:
## Description

This introduces a new memory protection middleware to help prevent out-of-memory conditions in SpiceDB by implementing admission control based on current memory usage.

This is not a perfect solution (doesn't prevent non-traffic-related sources of OOM) and is meant to support other future improvements to resource sharing in a single SpiceDB node.

The middleware is installed both in the main api and in dispatch, but at different thresholds. Memory usage is polled in the background, and if in-flight memory rises above the threshold, backpressure is placed on incoming requests.

The dispatch threshold is higher than the API threshold to preserve already admitted traffic as much as possible.

## Testing

- Unit tests included
- Manual E2E test:
   -  Modify [docker-compose.yaml](https://github.com/authzed/spicedb/blob/main/docker-compose.yaml) to set `mem_limit: "300mb"` to the `spicedb` containers
   - Run `docker-compose up --build`
   - Run this 

```
zed context set example localhost:50051 foobar --insecure
zed import development/schema.yaml
{
    echo '{"items":['
    for i in $(seq 1 200); do
      d=$(( (RANDOM % 9999) + 1 ))
      echo -n "{\"resource\":{\"objectTy  pe\":\"document\",\"objectId\": \"${d}\"}, \"permission\":\"view\",\"subject\":{ \"object\": {\"objectType\": \"user\", \"objectId\": \"1\"}}}"
      [ $i -lt 200 ] && echo -n ","
    done
    echo "], \"with_tracing\": true}"
} > payload.json
ab -n 100000 -c 200 -T 'application/json' -H 'Authorization: Bearer foobar' -p payload.json http://localhost:8443/v1/permissions/checkbulk
```

 you should see logs such as:

```
{
  "level": "warn",
  "traceID": "125b7b37c5775af1f2d9ebf253dcf3d1",
  "protocol": "grpc",
  "grpc.component": "server",
  "grpc.service": "authzed.api.v1.PermissionsService",
  "grpc.method": "CheckBulkPermissions",
  "grpc.method_type": "unary",
  "requestID": "d3vurmuoqu8s73bsdom0",
  "peer.address": "127.0.0.1:60376",
  "grpc.start_time": "2025-10-27T22:10:35Z",
  "grpc.code": "ResourceExhausted",
  "grpc.error": "rpc error: code = ResourceExhausted desc = server rejected the request because memory usage is above configured threshold",
  "grpc.time_ms": 0,
  "time": "2025-10-27T22:10:35Z",
  "message": "finished call"
}
```

and this graph in Grafana:

image

### DIFF
--- a/development/prometheus.yaml
+++ b/development/prometheus.yaml
@@ -1,7 +1,7 @@
 ---
 global:
-  scrape_interval: "15s"
-  evaluation_interval: "15s"
+  scrape_interval: "1s"
+  evaluation_interval: "1s"
 
 scrape_configs:
   - job_name: "spicedb"
--- a/internal/middleware/memoryprotection/memory_sampler_test.go
+++ b/internal/middleware/memoryprotection/memory_sampler_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestMemorySampler(t *testing.T) {
+func TestMemorySamplerOnInterval(t *testing.T) {
 	intervalSeconds := 1000
 	interval := time.Duration(intervalSeconds) * time.Second
 
@@ -19,6 +19,10 @@ func TestMemorySampler(t *testing.T) {
 	}{
 		`positive_limit`: {
 			limit:                     100 * 1024 * 1024,
+			expectGaugeAndUsageUpdate: true,
+		},
+		`very_low_limit`: {
+			limit:                     1,
 			expectGaugeAndUsageUpdate: true,
 		},
 		`negative_limit`: {
@@ -34,7 +38,7 @@ func TestMemorySampler(t *testing.T) {
 	for name, tc := range testcases {
 		t.Run(name, func(t *testing.T) {
 			synctest.Test(t, func(t *testing.T) {
-				sampler := NewMemorySampler(intervalSeconds, &HardCodedMemoryLimitProvider{Hardcodedlimit: tc.limit})
+				sampler := NewMemorySamplerOnInterval(intervalSeconds, &HardCodedMemoryLimitProvider{Hardcodedlimit: tc.limit})
 				t.Cleanup(sampler.Close)
 
 				now := time.Now()
@@ -54,6 +58,7 @@ func TestMemorySampler(t *testing.T) {
 				if tc.expectGaugeAndUsageUpdate {
 					require.Greater(t, gaugeValue, float64(0))
 					require.Greater(t, sampler.GetMemoryUsagePercent(), float64(0))
+					require.LessOrEqual(t, sampler.GetMemoryUsagePercent(), float64(1), "percentage should be between 0 and 1")
 				} else {
 					require.InDelta(t, 0, gaugeValue, 0.001) // near zero
 				}
--- a/pkg/cmd/server/server.go
+++ b/pkg/cmd/server/server.go
@@ -598,7 +598,7 @@ func (c *Config) buildMemoryProtectionConfigs(closeables *closeableStack) (memor
 		dispatchConfig = memoryprotection.Config{
 			ThresholdPercent: c.MemoryProtectionDispatchAPIThresholdPercent,
 		}
-		sampler = memoryprotection.NewMemorySampler(c.MemoryProtectionSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{})
+		sampler = memoryprotection.NewMemorySamplerOnInterval(c.MemoryProtectionSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{})
 		closeables.AddWithoutError(sampler.Close)
 	}
 	return apiConfig, dispatchConfig, sampler, nil
--- a/pkg/cmd/server/server_test.go
+++ b/pkg/cmd/server/server_test.go
@@ -446,7 +446,7 @@ func TestModifyUnaryMiddleware(t *testing.T) {
 		},
 	}}
 
-	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, memoryprotection.Config{ThresholdPercent: 0}, memoryprotection.NewMemorySampler(memoryprotection.DefaultSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{}), nil, nil}
+	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, memoryprotection.Config{ThresholdPercent: 0}, memoryprotection.NewMemorySamplerOnInterval(memoryprotection.DefaultSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{}), nil, nil}
 	opt = opt.WithDatastore(nil)
 
 	defaultMw, err := DefaultUnaryMiddleware(opt)
@@ -474,7 +474,7 @@ func TestModifyStreamingMiddleware(t *testing.T) {
 		},
 	}}
 
-	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, memoryprotection.Config{ThresholdPercent: 0}, memoryprotection.NewMemorySampler(memoryprotection.DefaultSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{}), nil, nil}
+	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, memoryprotection.Config{ThresholdPercent: 0}, memoryprotection.NewMemorySamplerOnInterval(memoryprotection.DefaultSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{}), nil, nil}
 	opt = opt.WithDatastore(nil)
 
 	defaultMw, err := DefaultStreamingMiddleware(opt)
@@ -623,12 +623,23 @@ func TestBuildMemoryProtectionConfig(t *testing.T) {
 			expectedSamplerInterval: 10,
 			expectAddToCloseables:   true,
 		},
-		`err_invalid`: {
+		`err_invalid_above_one_1`: {
 			config: &Config{
 				MemoryProtectionEnabled:                     true,
 				MemoryProtectionSampleIntervalSeconds:       10,
 				MemoryProtectionNormalAPIThresholdPercent:   1.1,
 				MemoryProtectionDispatchAPIThresholdPercent: 0.70,
+			},
+			expectedSamplerInterval: 0,
+			expectedErr:             "invalid memory protection configuration",
+			expectAddToCloseables:   false,
+		},
+		`err_invalid_above_one_2`: {
+			config: &Config{
+				MemoryProtectionEnabled:                     true,
+				MemoryProtectionSampleIntervalSeconds:       10,
+				MemoryProtectionNormalAPIThresholdPercent:   0.70,
+				MemoryProtectionDispatchAPIThresholdPercent: 1.1,
 			},
 			expectedSamplerInterval: 0,
 			expectedErr:             "invalid memory protection configuration",
@@ -697,7 +708,7 @@ func TestBuildDispatchServer(t *testing.T) {
 				_ = closeables.Close()
 			})
 
-			sampler := memoryprotection.NewMemorySampler(memoryprotection.DefaultSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{})
+			sampler := memoryprotection.NewMemorySamplerOnInterval(memoryprotection.DefaultSampleIntervalSeconds, &memoryprotection.DefaultMemoryLimitProvider{})
 			t.Cleanup(sampler.Close)
 
 			srv, err := tc.config.buildDispatchServer(mp, sampler, mockDatastore, mockDispatcher, &closeables, nil)
--- a/development/grafana/dashboards/dashboard.json
+++ b/development/grafana/dashboards/dashboard.json
@@ -124,20 +124,15 @@
             "type": "prometheus",
             "uid": "${DS_PROMETHEUS}"
           },
-          "disableTextWrap": false,
-          "editorMode": "builder",
-          "expr": "histogram_quantile(0.95, sum by(le, grpc_method, grpc_service) (rate(grpc_server_handling_seconds_bucket{job=\"spicedb\"}[$__rate_interval])))",
-          "fullMetaSearch": false,
-          "includeNullMetadata": true,
-          "instant": true,
+          "editorMode": "code",
+          "expr": "histogram_quantile(0.99, sum by(le, grpc_method, grpc_service) (rate(grpc_server_handling_seconds_bucket{job=\"spicedb\"}[$__rate_interval])))",
           "key": "Q-52ee0e33-980f-4a19-9821-39530de9f304-0",
           "legendFormat": "{{grpc_service}}/{{grpc_method}}",
           "range": true,
-          "refId": "A",
-          "useBackend": false
+          "refId": "A"
         }
       ],
-      "title": "p95 Latency by Method",
+      "title": "p99 Latency by Method",
       "type": "timeseries"
     },
     {
@@ -238,7 +233,7 @@
         "type": "prometheus",
         "uid": "${DS_PROMETHEUS}"
       },
-      "description": "5xx errors",
+      "description": "Response codes",
       "fieldConfig": {
         "defaults": {
           "color": {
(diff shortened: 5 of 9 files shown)


development/prometheus.yaml
```diff
@@ -6,6 +6,9 @@ global:
 scrape_configs:
   - job_name: "spicedb"
     static_configs:
-      - targets: ["spicedb:9090"]
+      - targets: ["spicedb-1:9090"]
```
miparnisari, line 9: FYI so that we can verify the new metrics in Grafana

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,281 @@
...
+// In production, use DefaultMemoryLimitProvider.
+// For testing, use HardCodedMemoryLimitProvider.
+type MemoryLimitProvider interface {
+	Get() int64
+	Set(int64)
+}
+
+var (
+	_ MemoryLimitProvider = (*DefaultMemoryLimitProvider)(nil)
+	_ MemoryLimitProvider = (*HardCodedMemoryLimitProvider)(nil)
```
tstirrat15: Why this syntax as opposed to constructing one?

internal/mocks/mock_datastore.go
miparnisari, line 1: FYI i can move the mock generation to a different PR

.gitattributes
```diff
@@ -0,0 +1,4 @@
+internal/mocks/*.go linguist-generated=true
```
miparnisari, line 1: FYI so that when changing these, the "View PR" view on Github says that the files are automatically generated and people can just mark "viewed" on them

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+		{Name: "/memory/classes/heap/objects:bytes"},
+	}
+
+	// Initialize with current memory usage
+	am.sampleMemory()
+
+	// Start background sampling with context
+	am.startBackgroundSampling()
+
+	log.Info().
```
miparnisari: FYI

```
{"level":"info","name":"dispatch-middleware","memory_limit_bytes":15762733056,"threshold_percent":95,"sample_interval_seconds":1,"time":"2025-10-28T18:27:10Z","message":"memory protection middleware initialized with background sampling"}

{"level":"info","name":"unary-middleware","memory_limit_bytes":15762733056,"threshold_percent":90,"sample_interval_seconds":1,"time":"2025-10-28T18:27:10Z","message":"memory protection middleware initialized with background sampling"}

{"level":"info","name":"stream-middleware","memory_limit_bytes":15762733056,"threshold_percent":90,"sample_interval_seconds":1,"time":"2025-10-28T18:27:10Z","message":"memory protection middleware initialized with background sampling"}
```

pkg/cmd/serve.go
```diff
@@ -121,6 +121,12 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	apiFlags.BoolVar(&config.EnablePerformanceInsightMetrics, "enable-performance-insight-metrics", false, "enables performance insight metrics, which are used to track the latency of API calls by shape")
 	apiFlags.StringVar(&config.MismatchZedTokenBehavior, "mismatch-zed-token-behavior", "full-consistency", "behavior to enforce when an API call receives a zedtoken that was originally intended for a different kind of datastore. One of: full-consistency (treat as a full-consistency call, ignoring the zedtoken), min-latency (treat as a min-latency call, ignoring the zedtoken), error (return an error). defaults to full-consistency for safety.")
 
+	// Memory Protection flags
+	apiFlags.BoolVar(&config.MemoryProtectionEnabled, "memory-protection-enabled", true, "enables a memory-based middleware that rejects requests when memory usage is too high")
+	apiFlags.IntVar(&config.MemoryProtectionAPIThresholdPercent, "memory-protection-api-threshold", 90, "memory usage threshold percentage for regular API requests (0-100)")
```
miparnisari: i don't like that we have the default 90 here and also in the struct itself 😭 i'd love a unified approach in the future

internal/middleware/memoryprotection/memoryprotection.go
tstirrat15: What does this mean?

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+		}
+	}()
+}
+
+// sampleMemory samples the current memory usage and updates the cached value
+func (am *MemoryAdmissionMiddleware) sampleMemory() {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Warn().Interface("panic", r).Msg("memory sampling panicked")
```
tstirrat15: Why are we doing this?

internal/mocks/mock_datastore.go
tstirrat15: This is fine.

pkg/cmd/server/server.go
```diff
@@ -431,27 +411,39 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
...
+		Logger:                    log.Logger,
+		AuthFunc:                  c.GRPCAuthFunc,
+		EnableVersionResponse:     !c.DisableVersionResponse,
+		DispatcherForMiddleware:   dispatcher,
+		EnableRequestLog:          c.EnableRequestLogs,
+		EnableResponseLog:         c.EnableResponseLogs,
+		DisableGRPCHistogram:      c.DisableGRPCLatencyHistogram,
+		MiddlewareServiceLabel:    serverName,
+		MismatchingZedTokenOption: mismatchZedTokenOption,
+		MemoryProtectionConfig:    apiMemoryProtectionConfig,
```
tstirrat15, line 430: TY

pkg/cmd/serve.go
```diff
@@ -121,6 +121,12 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	apiFlags.BoolVar(&config.EnablePerformanceInsightMetrics, "enable-performance-insight-metrics", false, "enables performance insight metrics, which are used to track the latency of API calls by shape")
 	apiFlags.StringVar(&config.MismatchZedTokenBehavior, "mismatch-zed-token-behavior", "full-consistency", "behavior to enforce when an API call receives a zedtoken that was originally intended for a different kind of datastore. One of: full-consistency (treat as a full-consistency call, ignoring the zedtoken), min-latency (treat as a min-latency call, ignoring the zedtoken), error (return an error). defaults to full-consistency for safety.")
 
+	// Memory Protection flags
+	apiFlags.BoolVar(&config.MemoryProtectionEnabled, "memory-protection-enabled", true, "enables a memory-based middleware that rejects requests when memory usage is too high")
```
vroldanbet: I understand the appeal to have this enabled by default (it's for the better!), but playing devil's advocate, this behavior may be surprising for folks as they update to the next release.

pkg/cmd/server/defaults.go
```diff
@@ -328,10 +316,16 @@ func DefaultUnaryMiddleware(opts MiddlewareOption) (*MiddlewareChain[grpc.UnaryS
 			WithInterceptor(grpcMetricsUnaryInterceptor).
 			Done(),
 
+		NewUnaryMiddleware().
```
vroldanbet, line 311: I understand why we would want to add it here, and I think realistic and practical place for it, but I'd be remiss if I didn't mention that we miss protection as the early middleware layers are traversed.

But again, this is not meant to be perfect, but good enough ™️

pkg/cmd/server/server_test.go
```diff
@@ -438,10 +444,10 @@ func TestModifyUnaryMiddleware(t *testing.T) {
 		},
 	}}
 
-	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, nil, nil}
+	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, memoryprotection.Config{ThresholdPercent: 0}, nil, nil}
 	opt = opt.WithDatastore(nil)
 
-	defaultMw, err := DefaultUnaryMiddleware(opt)
+	defaultMw, err := DefaultUnaryMiddleware(context.Background(), opt)
```
vroldanbet: nit
```suggestion
	defaultMw, err := DefaultUnaryMiddleware(t.Context(), opt)
```

pkg/cmd/server/server_test.go
```diff
@@ -466,10 +472,10 @@ func TestModifyStreamingMiddleware(t *testing.T) {
 		},
 	}}
 
-	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, nil, nil}
+	opt := MiddlewareOption{logging.Logger, nil, false, nil, false, false, false, "testing", consistency.TreatMismatchingTokensAsFullConsistency, memoryprotection.Config{ThresholdPercent: 0}, nil, nil}
 	opt = opt.WithDatastore(nil)
 
-	defaultMw, err := DefaultStreamingMiddleware(opt)
+	defaultMw, err := DefaultStreamingMiddleware(context.Background(), opt)
```
vroldanbet:
```suggestion
	defaultMw, err := DefaultStreamingMiddleware(t.Context(), opt)
```

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+	"google.golang.org/grpc/status"
+
+	log "github.com/authzed/spicedb/internal/logging"
+)
+
+const DefaultSampleIntervalSeconds = 1
+
+var (
+	// RejectedRequestsCounter tracks requests rejected due to memory pressure
+	RejectedRequestsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
```
vroldanbet: I think it would be easier to craft queries and visualizations if the counter was "self-contained", in that it was `memory admission outcome` rather than just "rejected". You could add a second label with the actual outcome, and that'd make it super easy to visualize ratios. Otherwise folks have to use _another metric_ to craft those ratios.

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+	_ MemoryLimitProvider = (*DefaultMemoryLimitProvider)(nil)
+	_ MemoryLimitProvider = (*HardCodedMemoryLimitProvider)(nil)
+)
+
+type DefaultMemoryLimitProvider struct{}
+
+func (p *DefaultMemoryLimitProvider) Get() int64 {
+	// SetMemoryLimit returns the previously set memory limit.
+	// A negative input does not adjust the limit, and allows for retrieval of the currently set memory limit
+	return debug.SetMemoryLimit(-1)
```
vroldanbet: It's not going to happen because this is called _way late_ in the process' lifecycle, but if automemlimit was called after this, the memory limiter would use the wrong amount of memory.

I know most recent Go versions have native support for cgroups.

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+	am.memoryLimit = memoryLimit
+	am.metricsSamples = []metrics.Sample{
+		{Name: "/memory/classes/heap/objects:bytes"},
+	}
+
+	// Initialize with current memory usage
+	am.sampleMemory()
+
+	// Start background sampling with context
+	am.startBackgroundSampling()
```
vroldanbet: nit: I feel like this should be done in a `Start` method, since it can be surprising to have goroutines spin in a factory method.

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+
+	log.Info().
+		Str("name", name).
+		Int64("memory_limit_bytes", memoryLimit).
```
vroldanbet: Maybe add another field with a humanized memory limit? makes it easier to parse, specially when you start using large machines

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+// checkAdmission returns an error if the request should be denied because memory usage is too high.
+func (am *MemoryAdmissionMiddleware) checkAdmission() error {
+	if am.memoryLimit < 0 || am.config.ThresholdPercent <= 0 {
+		// Memory protection is disabled
+		return nil
+	}
+
+	memoryUsage := am.getLastMemorySampleInBytes()
+
+	usagePercent := float64(memoryUsage) / float64(am.memoryLimit) * 100
```
vroldanbet: `memoryLimit == 0` seems to be allowed, and would cause division by zero, please add a test

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+		config:                    config,
+		lastMemorySampleInBytes:   &atomic.Uint64{},
+		timestampLastMemorySample: &atomic.Pointer[time.Time]{},
+		memoryLimit:               -1, // disabled initially
+		ctx:                       ctx,
+	}
+
+	// Get the current GOMEMLIMIT
+	memoryLimit := limitProvider.Get()
+	if memoryLimit < 0 {
```
vroldanbet: why would we want to allow memory limit zero?

pkg/cmd/serve.go
vroldanbet: We have some percent-based flags that are defined as [0...1] floats. Worth having a look and deciding which approach to commit to.

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+			case <-am.ctx.Done():
+				return
+			}
+		}
+	}()
+}
+
+// sampleMemory samples the current memory usage and updates the cached value
+func (am *MemoryAdmissionMiddleware) sampleMemory() {
+	defer func() {
```
vroldanbet: In which scenarios may this happen? Can the process continue to operate reliably? I don't think we typically recover panics in SpiceDB. Does it, in theory, mean that at some point we could stop sampling and have SpiceDB operate on a stale snapshot of memory?

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+		if r := recover(); r != nil {
+			log.Warn().Interface("panic", r).Msg("memory sampling panicked")
+		}
+	}()
+
+	now := time.Now()
+	metrics.Read(am.metricsSamples)
+	newUsage := am.metricsSamples[0].Value.Uint64()
```
vroldanbet: Indexed array access: Is this guaranteed? Should we add a length check?

internal/middleware/memoryprotection/memoryprotection.go
```diff
@@ -0,0 +1,271 @@
...
+	am.lastMemorySampleInBytes.Store(newUsage)
+	am.timestampLastMemorySample.Store(&now)
```
vroldanbet: Is it a concern that these two guys are not stored atomically as a whole? There is a point in execution where `lastMemorySampleInBytes` would have been updated, but `timestampLastMemorySample` is not yet updated

pkg/cmd/serve.go
miparnisari: I agree in that it would be surprising, but the other side of the coin is that right now if they put a limit on their memory usage and they go above that, they'd be getting an OOM, so this approach is better IMO.

I don't have a strong opinion and i would love other people's opinion on this.

josephschorr: +1 to enabling it by default - the node dying helps no one, and its better to get some requests through than none, IMO

pkg/cmd/server/defaults.go
miparnisari: Looking at the middlewares that run before this one (logging, metrics), it seems right to me, like i wouldn't want this new middleware to run earlier than the others 🤔 do you have a suggestion on a different order?

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: TIL we use `automemlimit`..

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: This is interesting... you are right, but then when I execute the code, the code doesn't panic, it just returns `usagePercent = +Inf`

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: IDK why this was here, i'll remove it

pkg/cmd/serve.go
miparnisari: Made it similar to this

```
 --datastore-revision-quantization-max-staleness-percent float           float percentage (where 1 = 100%)
```


i don't love it but I prefer being consistent

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: I'm not super concerned but i've changed it to use a `RWLock` to be safe 👍

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: Updated

pkg/cmd/serve.go
miparnisari: Funny, I just found that you can pass a large number to that flag and the server accepts it 😆 

```
- "SPICEDB_DATASTORE_REVISION_QUANTIZATION_MAX_STALENESS_PERCENT=10000"
```

internal/middleware/memoryprotection/memory_sampler_test.go
```diff
@@ -0,0 +1,63 @@
...
+				synctest.Wait()
+				t.Log("When we get here, the sampling is guaranteed to NOT have run yet")
+
+				require.False(t, sampler.GetTimestampLastMemorySample().After(now))
+
+				time.Sleep(1 * time.Millisecond)
+				synctest.Wait()
+				t.Log("When we get here, the sampling is guaranteed to have run")
+
+				require.True(t, sampler.GetTimestampLastMemorySample().After(now))
```
tstirrat15, line 55: This is in a goroutine :P

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: WDYM? This is standard go code to assert that a struct satisfies an interface

internal/middleware/memoryprotection/memoryprotection.go
miparnisari: i removed this code

ecordell:
We had a discussion on slack where we looked at some implementation details of `runtime/metrics`. Although it's not stop-the-world like `ReadMemStats` is, it does acquire a metrics lock (`metricsSema`) to read the values. Joey found https://github.com/odigos-io/go-rtml/ which looked promising as a way to get these values faster, _potentially_ even on every request (i.e. without the extra goroutines polling in the background).

pkg/cmd/serve.go
vroldanbet: seems like it's missing some validation :D

pkg/cmd/server/defaults.go
vroldanbet: It's a connundrum. We certainly want the log middleware to show us `ResourceExhausted` error for an API call.

Not suggesting we change it, just wanted to call out this particular scenario

internal/middleware/memoryprotection/memoryprotection.go
vroldanbet: Perhaps a safeguard would be to add a gauge that exposes what is the computed limit?

internal/middleware/memoryprotection/memoryprotection.go
vroldanbet: I learned that panics only happen for integers, but not for floats, as `IEEE-754` specifies +/- infinity as a possible value.

miparnisari:
@ecordell @vroldanbet please see https://github.com/authzed/spicedb/pull/2691

PR #2601, opened 2025-10-09
feat: add support for riscv64 architecture

rodneyosodo:
## Description

Add support for the `riscv64` architecture throughout the goreleaser pipeline by updating build settings, Docker image templates, manifest entries, and nightly release configurations.

- Add `riscv64` to the `goarch` list in both standard and nightly goreleaser build configurations
- Define `riscv64` Docker image builds (release and debug) with templates for Quay, GitHub Container Registry, and Docker Hub
- Include `riscv64` variants in Docker manifest listings for both release and debug tags across all registries
- Add `riscv64` targets in the nightly goreleaser config for both standard and debug image builds

## Testing

I am able to test goreleaser locally with

```bash
goreleaser release --snapshot --clean
```
```bash
  • skipping announce, publish, and validate...
  • cleaning distribution directory
  • loading environment variables
  • getting and validating git state
    • git state                                      commit=1e1e81cbac3d93452ad41eba7f7c9129c89f2f3a branch=feat/riscv-support current_tag=v1.46.0 previous_tag=v1.45.4 dirty=true
    • pipe skipped or partially skipped              reason=disabled during snapshot mode
  • parsing tag
  • setting defaults
    • dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info
    • dockers and docker_manifests are being phased out and will eventually be replaced by dockers_v2, check https://goreleaser.com/deprecations#dockers for more info
    • brews is being phased out in favor of homebrew_casks, check https://goreleaser.com/deprecations#brews for more info
  • snapshotting
    • building snapshot...                           version=1.46.1-next
  • running before hooks
    • running                                        hook=go run mage.go gen:completions
    • running                                        hook=go run mage.go gen:manpages
  • ensuring distribution directory
  • setting up metadata
  • writing release metadata
  • loading go mod information
  • build prerequisites
  • building binaries
    • building                                       binary=dist/spicedb_linux_arm64_v8.0/spicedb
    • building                                       binary=dist/spicedb_linux_riscv64_rva20u64/spicedb
    • building                                       binary=dist/spicedb_darwin_amd64_v1/spicedb
    • building                                       binary=dist/spicedb_darwin_arm64_v8.0/spicedb
    • building                                       binary=dist/spicedb_linux_amd64_v1/spicedb
      • took: 3m38s
  • archives
    • archiving                                      name=dist/spicedb_1.46.1-next_darwin_amd64.tar.gz
[...]

PR #2647, opened 2025-10-23
fix: fix riscv nightly builds

tstirrat15:
Followon to #2601 

## Description
We added support for riscv in #2601, but it broke nightly builds. I found a place where the values meaningfully differed; I also added the correct schema for a yaml LSP to each of the files.

## Changes
* Fix `base_image` key
* Add schema to all files
## Testing
Review. Merge and then check nightly builds.

### DIFF
--- a/.goreleaser.nightly.yml
+++ b/.goreleaser.nightly.yml
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=https://goreleaser.com/static/schema-pro.json
 ---
 version: 2
 builds:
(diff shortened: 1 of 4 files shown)


.goreleaser.nightly.yml
```diff
@@ -133,7 +134,7 @@ kos:
       - "quay.io/authzed/spicedb-git"
       - "ghcr.io/authzed/spicedb-git"
       - "authzed/spicedb-git"
-    base_image: "docker.io/busybox"
+    base_image: "cgr.dev/chainguard/busybox"
```
tstirrat15: This is just updating it to be the same as the other `base_image` keys. I missed this when reviewing.

tstirrat15:
Closed in favor of #2648

PR #2649, opened 2025-10-23
feat: Add a builder pattern to the schema v2 library and fix how caveats are converted

josephschorr:
## Description

Adds a builder pattern for the schema v2 lib.

Example with Done:

```go
	s := schema.NewSchemaBuilder().
		AddDefinition("document").
		AddRelation("owner").
		AllowedDirectRelation("user").
		Done().
		AddRelation("editor").
		AllowedDirectRelation("user").
		Done().
		AddRelation("viewer").
		AllowedDirectRelation("user").
		Done().
		AddPermission("view").
		UnionExpr().
		AddRelationRef("owner").
		AddRelationRef("editor").
		AddRelationRef("viewer").
		Done().
		Done().
		Done().
		Build()
```

Example with constructors:

```go
	s := schema.NewSchemaBuilder().
		Definition(
			schema.NewDefinition("user"),
		).
		Definition(
			schema.NewDefinition("document").
				Relation(schema.NewRelation("owner").AllowedDirectRelation("user")).
				Relation(schema.NewRelation("editor").AllowedDirectRelation("user")).
				Relation(schema.NewRelation("viewer").AllowedDirectRelation("user")).
				Permission(
					schema.NewPermission("view", schema.Union(
						schema.RelRef("owner"),
						schema.RelRef("editor"),
						schema.RelRef("viewer"),
					)),
				),
		).
		Build()
```

## Testing

Local unit tests

### DIFF
--- a/pkg/schema/v2/builder.go
+++ b/pkg/schema/v2/builder.go
@@ -1,0 +1,709 @@
+package schema
+
+import "github.com/authzed/spicedb/pkg/tuple"
+
+// Common caveat parameter type constants for use with CaveatBuilder.
+const (
+	CaveatTypeInt        = "int"
+	CaveatTypeBool       = "bool"
+	CaveatTypeString     = "string"
+	CaveatTypeBytes      = "bytes"
+	CaveatTypeDouble     = "double"
+	CaveatTypeUInt       = "uint"
+	CaveatTypeDuration   = "duration"
+	CaveatTypeTimestamp  = "timestamp"
+	CaveatTypeIPAddress  = "ipaddress"
+	CaveatTypeList       = "list"
+	CaveatTypeMap        = "map"
+	CaveatTypeAny        = "any"
+	CaveatTypeListString = "list<string>"
+	CaveatTypeListInt    = "list<int>"
+	CaveatTypeMapAny     = "map<any>"
+)
+
+// SchemaBuilder provides a fluent API for constructing schemas programmatically.
+// It maintains parent-child relationships automatically and provides method chaining.
+//
+// Example usage:
+//
+//	schema := NewSchemaBuilder().
+//		AddDefinition("user").
+//			Done().
+//		AddDefinition("document").
+//			AddRelation("owner").
+//				AllowedDirectRelation("user").
+//				Done().
+//			AddRelation("viewer").
+//				AllowedDirectRelation("user").
+//				AllowedRelation("group", "member").
+//				Done().
+//			AddPermission("view").
+//				UnionExpr().
+//					Add(RelRef("owner")).
+//					Add(RelRef("viewer")).
+//					Done().
+//				Done().
+//			Done().
+//		AddCaveat("has_access").
+//			Expression("resource == 42").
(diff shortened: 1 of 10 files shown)


pkg/schema/v2/builder.go
```diff
@@ -0,0 +1,605 @@
...
+		db.definition.permissions[name] = permission
+	}
+	return &PermissionBuilder{
+		definitionBuilder: db,
+		permission:        permission,
+	}
+}
+
+// Done completes the definition and returns to the schema builder.
+func (db *DefinitionBuilder) Done() *SchemaBuilder {
```
barakmich, line 203: All these `Done()` functions actually don't "complete the definition" and need to be called, strictly? It may leave things in an intermediate state, and you don't have to call them? The only thing they allow is a really deep (and in my opinion, messy) chain. 

```
schema := NewSchemaBuilder()
schema.Definition("user")
doc := schema.Definition("document")
doc.Relation("owner").AllowedDirectRelation("user")
..etc...
```

josephschorr: Strictly, no. They are more so you can "pop-out" one level and continue using the builder pattern

barakmich: Yeah, the pop-out I get. But then you end up with these ugly super chains. 

And you can keep the super chains. But then the godoc is off.

josephschorr: I still prefer the chains; I feel they are much easier to use as a human

ecordell: IMO the kube applyconfig packages do this well: you can chain things at the same level only.

```go
  applyappsv1.Deployment(name, namespace).
    WithSpec(applyappsv1.DeploymentSpec().
      WithReplicas(replicaCount),
    )
```
(hard to find good examples in the wild but I found this [blog post](https://superorbital.io/blog/server-side-type-tetris/))

translating the example it would be something like:
```go
schema.NewSchema().
  Definition(
    schema.Definition("document").
      Relation(
         schema.Relation("owner").AllowedDirectRelation("user")
      ).
      Relation(
         schema.Relation("editor").AllowedDirectRelation("user")
      ).
      Relation(
         schema.Relation("viewer").AllowedDirectRelation("user")
      ).
      Permission(
        schema.Permission("view", schema.Union(
          schema.RelationRef("owner"), 
          schema.RelationRef("editor"), 
          schema.RelationRef("viewer"),
        ))
      )
   ).
   Definition(schema.Definition("user").Relation(schema.Relation("friend").AllowedDirectRelation("user"))  // added for example
```
but you can be flexible in what stuff is pure-builder vs. what constructors take in (i.e. here I opted to have `Union` take a list of sub expressions vs. building them only)

josephschorr: Yeah, there are a few different designs. I, personally, prefer the `.Done` approach, as it reads how I write (and these are primarily for humans to use), but I could be convinced to change constructors to *also* support taking in the children (we could easily support both)

josephschorr: Added both designs

PR #2651, opened 2025-10-24
chore: change nil references to be their own object type in schema v2

josephschorr:
This is necessary to make resolution easier and to not use a magic string for nil

## Testing

Unit tests


barakmich (approved):
Obviously makes sense as nil is a thing.

_Why_ nil is a thing I don't 100% get, but, sure.

josephschorr:
> Obviously makes sense as nil is a thing.
> 
> _Why_ nil is a thing I don't 100% get, but, sure.

Its to allow for easy slotting during development, i.e.

`permission view = nil // I'll fill this in later`

Also allows for disabling a permission without returning an error

PR #2652, opened 2025-10-27
feat: Add a property-based schema and relationship generator

josephschorr:
Will be used for validation of logic in consistency test suites


pkg/schema/v2/schema.go
```diff
@@ -31,6 +31,13 @@ func (s *Schema) Caveats() map[string]*Caveat {
 	return s.caveats
 }
 
+// GetTypeDefinition returns the type definition with the given name and a boolean
+// indicating whether it exists in the schema.
+func (s *Schema) GetTypeDefinition(name string) (*Definition, bool) {
```
barakmich, line 36: This is good, and I thought it existed before but it doesn't, instead there's `s.Definitions()[name]` in, eg, `pkg/query/build_tree.go` -- so probably use this method there

barakmich (approved):
Approved with cleanup nit

pkg/schema/v2/schema.go
josephschorr: I'll do a followup clean PR

josephschorr: https://github.com/authzed/spicedb/pull/2659

PR #986, opened 2022-11-08
logging and error handling improvements

vroldanbet:
Closes https://github.com/authzed/spicedb/issues/980
Closes https://github.com/authzed/spicedb/issues/903
Some improvements related to https://github.com/authzed/spicedb/issues/753

This PR includes various improvements around:
- handling context cancellation and deadline errors so the API returns the correct gRPC code
- reduce log noise

The commit log contains a detailed description of each change.

PR #2654, opened 2025-10-27
fix: do not warn if requestid middleware errors due to ErrIllegalHeaderWrite

miparnisari:
## Description

Sometimes, we are seeing this:

```

{"level":"warn","error":"rpc error: code = Internal desc = transport: SendHeader called multiple times","time":"2025-10-27T18:15:17Z","message":"requestid: could not report metadata"} |  
```

This is noisy, because `Internal` triggers our 5xx alarms.

## Changes

- Fix `requestid` middleware
- Minor refactor to `usagemetric` middleware (no functional changes)

## References

Similar PR with a similar change was sent in https://github.com/authzed/spicedb/pull/986

### DIFF
--- a/internal/middleware/usagemetrics/usagemetrics.go
+++ b/internal/middleware/usagemetrics/usagemetrics.go
@@ -3,6 +3,7 @@ package usagemetrics
 import (
 	"context"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
@@ -38,7 +39,7 @@ type reporter struct{}
 
 func (r *reporter) ServerReporter(ctx context.Context, callMeta interceptors.CallMeta) (interceptors.Reporter, context.Context) {
 	_, methodName := grpcutil.SplitMethodName(callMeta.FullMethod())
-	ctx = ContextWithHandle(ctx)
+	ctx = contextWithHandle(ctx)
 	return &serverReporter{ctx: ctx, methodName: methodName}, ctx
 }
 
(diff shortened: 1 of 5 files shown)


internal/middleware/usagemetrics/usagemetrics.go
```diff
@@ -48,19 +49,25 @@ type serverReporter struct {
...
 }
 
+// PostCall is invoked after all PostMsgSend operations.
 func (r *serverReporter) PostCall(_ error, _ time.Duration) {
 	responseMeta := FromContext(r.ctx)
 	if responseMeta == nil {
 		responseMeta = &dispatch.ResponseMeta{}
 	}
 
-	err := annotateAndReportForMetadata(r.ctx, r.methodName, responseMeta)
```
miparnisari, line 57: FYI inlined function

internal/middleware/usagemetrics/usagemetrics.go
```diff
@@ -48,19 +49,25 @@ type serverReporter struct {
...
-		return
-	}
+	DispatchedCountHistogram.WithLabelValues(r.methodName, "false").Observe(float64(responseMeta.DispatchCount))
+	DispatchedCountHistogram.WithLabelValues(r.methodName, "true").Observe(float64(responseMeta.CachedDispatchCount))
+	err := responsemeta.SetResponseTrailerMetadata(r.ctx, map[responsemeta.ResponseMetadataTrailerKey]string{
+		responsemeta.DispatchedOperationsCount: strconv.Itoa(int(responseMeta.DispatchCount)),
+		responsemeta.CachedOperationsCount:     strconv.Itoa(int(responseMeta.CachedDispatchCount)),
+	})
 	if err != nil {
+		// if context is cancelled, the stream will be closed, and gRPC will return ErrIllegalHeaderWrite
```
miparnisari: FYI `ErrIllegalHeaderWrite` is not public so we cannot reference it from here :(

pkg/middleware/requestid/requestid.go
```diff
@@ -54,6 +55,7 @@ func (r *handleRequestID) ClientReporter(ctx context.Context, meta interceptors.
 	return interceptors.NoopReporter{}, ctx
 }
 
+// ServerReporter is invoked before the request begins processing.
```
miparnisari, line 58: FYI this is what makes it tricky to test - if this code was inside `PostCall` like the other middleware above, it would be easier to simulate a context cancellation which would cause `PostCall` to _notice it_

internal/middleware/usagemetrics/usagemetrics.go
barakmich: Perhaps add this note to this comment? (and below?)

miparnisari: done

pkg/middleware/requestid/requestid.go
```diff
@@ -67,7 +69,11 @@ func (r *handleRequestID) ServerReporter(ctx context.Context, _ interceptors.Cal
 			responsemeta.ResponseMetadataTrailerKey(responsemeta.RequestID): requestID,
 		})
 		if err != nil {
-			log.Ctx(ctx).Warn().Err(err).Msg("requestid: could not report metadata")
+			// if context is cancelled, the stream will be closed, and gRPC will return ErrIllegalHeaderWrite (which is private)
+			// this prevents logging unnecessary error messages
+			if !strings.Contains(err.Error(), "SendHeader called multiple times") {
```
tstirrat15, line 74: Is there a way to check whether the stream is already closed before we attempt to set the trailer metadata, rather than attempting and then having to read the error?

pkg/middleware/requestid/requestid_test.go
```diff
@@ -358,17 +359,29 @@ func TestRequestIDMiddlewareTimeout(t *testing.T) {
...
+
+			// context will be cancelled in the middle of the middleware (if we are lucky)
+			ctx, cancel := context.WithTimeout(s.T().Context(), 1*time.Millisecond)
+			defer cancel()
+
+			var trailer metadata.MD
+			_, err := s.Client.PingEmpty(ctx, &testpb.PingEmptyRequest{}, grpc.Trailer(&trailer))
+
+			// The RPC should fail due to context cancellation, but middleware should handle it gracefully
+			require.Error(s.T(), err)
```
tstirrat15, line 382: this is in a goroutine

pkg/middleware/requestid/requestid.go
miparnisari: Not that i know of.. we can get the stream with `grpc.ServerTransportStreamFromContext` but we can't get its state

PR #2655, opened 2025-10-27
Lint fixes part 10

tstirrat15:
Parting out #2455 

## Description
Another set of lint fixes

## Changes
Will annotate

## Testing
Review. See that tests pass.

### DIFF
--- a/pkg/datastore/test/revisions.go
+++ b/pkg/datastore/test/revisions.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"testing"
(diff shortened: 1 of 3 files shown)


pkg/datastore/test/revisions.go
```diff
@@ -300,18 +300,33 @@ func ConcurrentRevisionsTest(t *testing.T, tester DatastoreTester) {
...
+					errCh <- fmt.Errorf("HeadRevision error: %w", err)
+					continue
+				}
+				if err := ds.CheckRevision(ctx, head); err != nil {
+					errCh <- fmt.Errorf("CheckRevision error: %w", err)
+					continue
+				}
+				if !(head.GreaterThan(startingRev) || head.Equal(startingRev)) {
+					errCh <- fmt.Errorf("head revision is not greater than or equal to startingRev")
+				}
```
tstirrat15, line 322: The semantics of this test are a little unclear - I guess it seems like it should be "if any of these error conditions are met at any point the test fails," which matches the way it was rewritten. I'm not quite sure what an alternative would be.

pkg/releases/releases_test.go
```diff
@@ -24,8 +24,9 @@ func TestGetLatestRelease(t *testing.T) {
 	})
 
 	t.Run("successful release fetch", func(t *testing.T) {
+		var receivedPath string
 		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			require.Equal(t, "/repos/authzed/spicedb/releases/latest", r.URL.Path)
+			receivedPath = r.URL.Path
```
tstirrat15, line 29: Not technically thread-safe, but shouldn't be getting invoked more than once.

pkg/schema/typesystem_test.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 func TestTypeSystemConcurrency(t *testing.T) {
+	fmt.Println("starting test")
```
miparnisari:
```suggestion
```

miparnisari (dismissed):

miparnisari (approved):

PR #2656, opened 2025-10-28
fix: properly rewrite errors for watch api, part 2

miparnisari:
This is a continuation of https://github.com/authzed/spicedb/pull/2640

### DIFF
--- a/internal/services/shared/errors.go
+++ b/internal/services/shared/errors.go
@@ -177,6 +177,9 @@ func rewriteError(ctx context.Context, err error, config *ConfigForErrors) error
 		return status.Errorf(codes.Canceled, "watch canceled by user: %s", err)
 	case errors.As(err, &datastore.WatchDisconnectedError{}):
 		return status.Errorf(codes.ResourceExhausted, "watch disconnected: %s", err)
+	case errors.As(err, &datastore.WatchRetryableError{}):
+		// Unavailable is safe to retry
+		return status.Error(codes.Unavailable, err.Error())
 	case errors.As(err, &datastore.CounterAlreadyRegisteredError{}):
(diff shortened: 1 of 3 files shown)


internal/services/shared/errors.go
```diff
@@ -177,6 +177,9 @@ func rewriteError(ctx context.Context, err error, config *ConfigForErrors) error
 		return status.Errorf(codes.Canceled, "watch canceled by user: %s", err)
 	case errors.As(err, &datastore.WatchDisconnectedError{}):
 		return status.Errorf(codes.ResourceExhausted, "watch disconnected: %s", err)
+	case errors.As(err, &datastore.WatchRetryableError{}):
+		// FailedPrecondition is safe to retry
+		return status.Error(codes.FailedPrecondition, err.Error())
```
vroldanbet: It would seem like `Unavailable` is a better error condition

From https://grpc.io/docs/guides/status-codes/

>The service is currently unavailable. This is most likely a transient condition, which can be corrected by retrying with a backoff. Note that it is not always safe to retry non-idempotent operations.

Whereas FailedPrecondition says this

>The operation was rejected because the system is not in a state required for the operation’s execution. For example, the directory to be deleted is non-empty, an rmdir operation is applied to a non-directory, etc. Service implementors can use the following guidelines to decide between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher level (e.g., when a client-specified test-and-set fails, indicating the client should restart a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an “rmdir” fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless the files are deleted from the directory.

miparnisari: Yeah i only had one reason for using FailedPrecondition, which was that we didn't alarm on it 😆 i've updated to use Unavailable now.

PR #2663, opened 2025-10-28
feat: add internal flag for running CheckPermission with a query plan

barakmich:


pkg/cmd/serve.go
```diff
@@ -167,6 +167,10 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 
 	experimentalFlags := nfs.FlagSet(BoldBlue("Experimental"))
 	// Flags for experimental features
+	experimentalFlags.BoolVar(&config.ExperimentalQueryPlan, "experimental-query-plan", false, "enables experimental query plan within the API")
```
josephschorr: I recommend making this take a string instead of a bool; that way, we can put options or other flags in it, if need be

PR #2664, opened 2025-10-28
chore: remove NodeID middleware and replace with static function

miparnisari:
When we invoke the dispatcher out of band and in parallel with the normal request handlers, there can be a race wherein we have two pieces of code trying to read the `nodeid` from the context, and one of them may be attempting to write it.


internal/testserver/server.go
```diff
@@ -131,6 +132,10 @@ func NewTestServerWithConfigAndDatastore(require *require.Assertions,
 						Name:       "servicespecific",
 						Middleware: servicespecific.UnaryServerInterceptor,
 					},
+					{
```
miparnisari: This is suspicious. Why do we have this custom server setup here? It seems it's only used by the steelthread tests. Why can't those tests use a default server, why do they need a special arrangement of middleware?

josephschorr: Because we don't need all the other middleware running and all it does it increase the test time to have them running

miparnisari: How could middleware add so much test time that we decide to not include them?

pkg/middleware/nodeid/nodeid.go
vroldanbet, line 1: This is no longer middleware, but it still lives in `pkg/middleware`?

This is btw a breaking change, since it's deleting exported packages. Not that we have directives in place to honor that kind of compatibility, we only focus on gRPC / HTTP API compatibility.

pkg/middleware/nodeid/nodeid.go
```diff
@@ -37,66 +26,10 @@ func init() {
...
-
-// ContextWithHandle adds a placeholder to a context that will later be
-// filled by the Node ID.
-func ContextWithHandle(ctx context.Context) context.Context {
-	return context.WithValue(ctx, nodeIDKey, &nodeIDHandle{})
+	computedNodeID = spiceDBPrefix + hex.EncodeToString(hasher.Sum(nil))
 }
 
-// FromContext reads the node's ID out of a context.Context.
-func FromContext(ctx context.Context) (string, error) {
```
vroldanbet, line 50: @josephschorr do you remember why you turned this into a middleware in the first place? The nodeID is not request specific, yet it was designed as if it was. Is there perhaps something we are missing? I don't see the nodeID showing up in logs, so I can't find an explanation for original design.

pkg/middleware/nodeid/nodeid.go
vroldanbet: alternatively, if we find having it as a middleware is still necessary (I can't find a reason), I'd suggest reimplementing it using `authzed/ctxkey` module

vroldanbet (commented):
This seems to make sense, although I'm worried we are missing a reason why this was designed as a middleware. As far as I can tell, there is no reason for it, since the nodeID is deterministic, and based on the host.

Only remark would be to move this outside of the middleware package

pkg/middleware/nodeid/nodeid.go
miparnisari: >This is no longer middleware, but it still lives in pkg/middleware

yes, i wasn't sure where to move it to; do you have a suggestion?

pkg/middleware/nodeid/nodeid.go
josephschorr: I think the idea was to provide custom node ID support, but then it was never requested

pkg/middleware/nodeid/nodeid.go
miparnisari: @vroldanbet I will merge as is but please do let me know if you have a suggestion for a better place and I can move it in a follow-up PR

PR #2667, opened 2025-10-29
Lint fixes part 11

tstirrat15:
Part of parting out #2455 

## Description
More lint fixes for the pile

## Changes
Will annotate.

## Testing
Review. See that tests pass.

### DIFF
--- a/pkg/composableschemadsl/input/sourcepositionmapper_test.go
+++ b/pkg/composableschemadsl/input/sourcepositionmapper_test.go
@@ -5,27 +5,22 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/development/assertions.go
```diff
@@ -31,9 +33,10 @@ func RunAllAssertions(devContext *DevContext, assertions *blocks.Assertions) ([]
 		return nil, err
 	}
 
-	failures := append(trueFailures, caveatedFailures...)
-	failures = append(failures, falseFailures...)
-	return failures, nil
```
tstirrat15, line 36: The linter was complaining here that `append` was called with different slices as the base. Adding an `allFailures` at the start retains the semantic while making it a bit easier to read.

tstirrat15 (commented):
See comment

miparnisari (approved):

PR #2668, opened 2025-10-29
Lint fixes part 12

tstirrat15:
Part of parting out #2455

## Description
This is the last component of #2455, with the exception of enabling the linters and fixing the garbage collection test. We're getting very close.

## Changes
See comments

## Testing
Review. See that tests pass.

### DIFF
--- a/internal/datastore/postgres/xid8.go
+++ b/internal/datastore/postgres/xid8.go
@@ -71,8 +71,7 @@ func (Uint64Codec) PlanEncode(_ *pgtype.Map, _ uint32, format int16, value any) 
 			return encodePlanUint64CodecBinaryUint64Valuer{}
 		}
 	case pgtype.TextFormatCode:
-		switch value.(type) { //nolint:gocritic
-		case uint64:
+		if _, ok := value.(uint64); ok {
 			return encodePlanUint64CodecTextUint64{}
 		}
 	}


pkg/cmd/server/server_test.go
```diff
@@ -113,8 +113,9 @@ func TestOTelReporting(t *testing.T) {
 
 	schemaSrv := v1.NewSchemaServiceClient(conn)
 
+	runErrCh := make(chan error, 1)
 	go func() {
-		require.NoError(t, srv.Run(ctx))
```
tstirrat15, line 117: I checked to make sure that this test still runs, and it appears to.

pkg/composableschemadsl/compiler/compiler_test.go
```diff
@@ -1353,24 +1353,23 @@ func TestCompile(t *testing.T) {
...
-					filterSourcePositions(l.Get(i).Message())
-				}
-			} else if fd.IsMap() {
-				m := v.Map()
-				m.Range(func(k protoreflect.MapKey, v protoreflect.Value) bool {
-					filterSourcePositions(v.Message())
-					return true
-				})
+		switch {
+		case fd.Kind() == protoreflect.MessageKind && fd.IsList():
```
tstirrat15, line 1357: This satisfies the linter but also reads a little weird to me - I'm happy to try for a refactor if desired.

tstirrat15 (commented):
See comments

pkg/cmd/server/server_test.go
miparnisari: what do you mean by that? why would this test not be running?

miparnisari (approved):

PR #2670, opened 2025-10-30
test: fix flaky TestErrCtx and TestContextError

miparnisari:


pkg/migrate/migrate_test.go
```diff
@@ -46,7 +46,7 @@ type fakeTx struct{}
 
 func TestContextError(t *testing.T) {
 	req := require.New(t)
-	ctx, cancelFunc := context.WithDeadline(t.Context(), time.Now().Add(1*time.Millisecond))
```
tstirrat15, line 49: Do you have a sense for why this ever had a timeout?

Also is there time involved in this test? Would synctest help at all, or does using a normal cancellation context solve the underlying problem?

tstirrat15 (commented):
See comment, otherwise LGTM

pkg/migrate/migrate_test.go
miparnisari: i don't know why this ctx had a timeout. There's an explicit cancellation a few lines below 🤷‍♀️

tstirrat15 (approved):
LGTM

PR #2674, opened 2025-10-31
chore: reduce chance of flakiness on the migrate integration test

josephschorr:
Adds retries and increases the wait timeout for the migration call

### DIFF
--- a/cmd/spicedb/migrate_integration_test.go
+++ b/cmd/spicedb/migrate_integration_test.go
@@ -43,57 +43,71 @@ func TestMigrate(t *testing.T) {
 		}
 
 		t.Run(engineKey, func(t *testing.T) {
-			engineKey := engineKey
+			for retryCount := range 5 {
+				engineKey := engineKey
 
-			r := testdatastore.RunDatastoreEngineWithBridge(t, engineKey, bridgeNetworkName)
-			db := r.NewDatabase(t)
+				r := testdatastore.RunDatastoreEngineWithBridge(t, engineKey, bridgeNetworkName)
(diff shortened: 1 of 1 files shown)


cmd/spicedb/migrate_integration_test.go
```diff
@@ -43,57 +43,71 @@ func TestMigrate(t *testing.T) {
...
+					// On error, retry a few times before failing
+					t.Logf("retrying migration for engine %s due to error: %v", engineKey, err)
+					continue
+				}
 
-				require.Fail(t, "Got non-zero exit code", stream.String())
-			}
+				require.NoError(t, err)
+
+				waitCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
miparnisari, line 77: why do we even have a timeout here..? i say let the thing take as long as it wants: remove the retry stuff and the `waitCtx`.

josephschorr: I don't want the test to hang

miparnisari: it won't hang, ~that's what `go test -timeout XXX` is for: https://github.com/authzed/actions/blob/main/go-test/action.yaml#L11~

ugh we don't use that action for these. can you add `-timeout 20m` to https://github.com/authzed/spicedb/blob/03a31daa2c6ce49dc40a47353f6d43a048595394/magefiles/util.go#L42-L51?

Also FYI, github actions have a default timeout of 360 minutes per step.

josephschorr:
Closing in favor of https://github.com/authzed/spicedb/pull/2696

PR #2675, opened 2025-10-31
ci: update dependabot rules

miparnisari:


.github/dependabot.yml
```diff
@@ -2,9 +2,13 @@
 version: 2
 updates:
   - package-ecosystem: "gomod"
-    directory: "/"
+    directories:
+      - "/"
+      - "/magefiles"
+      - "e2e"
```
miparnisari: No wonder the spicedb dependency in `e2e` is so old 😮‍💨 

https://github.com/authzed/spicedb/blob/8c66fc6880adee174593d5b3882fac7f2eadb39b/e2e/go.mod#L13

tstirrat15 (approved):
yesss

PR #2676, opened 2025-10-31
test: disable revision heartbeat in PG revision tests

josephschorr:
They may be interfering with the expected revisions and therefore causing flakiness


vroldanbet (approved):

miparnisari:
@josephschorr could you explain the reasoning behind this change?

josephschorr:
> @josephschorr could you explain the reasoning behind this change?

The heartbeat moves the revision on a cadence but the tests expect it to not change while checking

PR #2677, opened 2025-10-31
test: fix flaky TestWatchIntegrityFailureDueToInvalidHashSignature

josephschorr:


pkg/datastore/datastore.go
```diff
@@ -710,7 +710,9 @@ type ReadOnlyDatastore interface {
 
 	// Watch notifies the caller about changes to the datastore, based on the specified options.
 	//
-	// All events following afterRevision will be sent to the caller.
+	// All events following afterRevision will be sent to the caller. Changes made *in* afterRevision will not be included.
+	//
+	// When the changes channel is closed, callers MUST discard any changes received (they will be the zero value).
```
miparnisari, line 715: We could also make the channel return pointers so that closure returns `nil`, which is more clear. But this is more work.

PR #2678, opened 2025-10-31
chore: fix new govet issues

tstirrat15:
## Description
Part of getting linting sorted. These came with the most recent version of golangci-lint.

## Changes
* Autofix `buildtag: +build line is no longer needed (govet)`
* Fix `concat-loop: string concatenation in a loop (perfsprint)`
## Testing
Review. See that tests pass.

### DIFF
--- a/cmd/spicedb/migrate_integration_test.go
+++ b/cmd/spicedb/migrate_integration_test.go
@@ -1,5 +1,4 @@
 //go:build docker && image
-// +build docker,image
 
 package main
 
--- a/internal/datastore/common/relationships.go
+++ b/internal/datastore/common/relationships.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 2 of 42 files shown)


cmd/spicedb/migrate_integration_test.go
```diff
@@ -1,5 +1,4 @@
 //go:build docker && image
-// +build docker,image
```
tstirrat15, line 2: These are no longer needed according to govet.

internal/datastore/common/relationships.go
```diff
@@ -71,14 +72,16 @@ func runExplainIfNecessary[R Rows](ctx context.Context, builder RelationshipsQue
 	}
 
 	err = tx.QueryFunc(ctx, func(ctx context.Context, rows R) error {
-		explainString := ""
+		var explainBuilder strings.Builder
```
tstirrat15, line 75: These were the `perfsprint` changes - basically that `string += otherString` results in an allocation, where `strings.Builder` lets you use a buffer.

miparnisari (approved):

pkg/composableschemadsl/compiler/development.go
```diff
@@ -46,11 +48,11 @@ func (nc *NodeChain) FindNodeOfType(nodeType dslshape.NodeType) DSLNode {
 }
 
 func (nc *NodeChain) String() string {
-	var out string
+	var builder strings.Builder
```
josephschorr, line 51: Can we add a test for this as a followup?

PR #2682, opened 2025-10-31
chore: enable gocritic

tstirrat15:
## Description
Getting close to done with linting things! This enables `gocritic` across the repo.

## Changes
Will annotate.

## Testing
Review. See that tests pass.

### DIFF
--- a/internal/cursorediterator/blocks_test.go
+++ b/internal/cursorediterator/blocks_test.go
@@ -639,9 +639,7 @@ func TestCursoredProducerMapperIterator(t *testing.T) {
 			}
 
 			// Cursor converter functions for int
-			intFromString := func(s string) (int, error) {
-				return strconv.Atoi(s)
-			}
+			intFromString := strconv.Atoi
 
 			intToString := func(i int) (string, error) {
 				return strconv.Itoa(i), nil
@@ -1084,9 +1082,7 @@ func TestCursoredProducerMapperIterator(t *testing.T) {
 		}
 
 		// Cursor converter functions for int
-		intFromString := func(s string) (int, error) {
-			return strconv.Atoi(s)
-		}
+		intFromString := strconv.Atoi
 
 		intToString := func(i int) (string, error) {
 			return strconv.Itoa(i), nil
@@ -1383,9 +1379,7 @@ func TestCursoredProducerMapperIteratorConcurrency1SpecialCase(t *testing.T) {
 	ctx := t.Context()
 
 	// Cursor converter functions for int
-	intFromString := func(s string) (int, error) {
-		return strconv.Atoi(s)
-	}
+	intFromString := strconv.Atoi
 
 	intToString := func(i int) (string, error) {
 		return strconv.Itoa(i), nil
@@ -1718,9 +1712,7 @@ func TestConcurrencyConsistency(t *testing.T) {
 
 	t.Run("CursoredProducerMapperIterator consistency", func(t *testing.T) {
 		// Cursor converter functions
-		intFromString := func(s string) (int, error) {
-			return strconv.Atoi(s)
-		}
+		intFromString := strconv.Atoi
 		intToString := func(i int) (string, error) {
 			return strconv.Itoa(i), nil
 		}
@@ -1972,9 +1964,7 @@ func TestDisableCursorsInContext(t *testing.T) {
 	t.Run("CursoredProducerMapperIterator with cursors disabled", func(t *testing.T) {
 		ctxWithoutCursors := DisableCursorsInContext(ctx)
 
-		intFromString := func(s string) (int, error) {
-			return strconv.Atoi(s)
-		}
+		intFromString := strconv.Atoi
 
 		intToString := func(i int) (string, error) {
 			return strconv.Itoa(i), nil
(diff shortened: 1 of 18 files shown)


internal/cursorediterator/blocks_test.go
```diff
@@ -639,9 +639,7 @@ func TestCursoredProducerMapperIterator(t *testing.T) {
 			}
 
 			// Cursor converter functions for int
-			intFromString := func(s string) (int, error) {
-				return strconv.Atoi(s)
-			}
```
tstirrat15, line 644: This is the `unlambda` rule - basically don't wrap a function in a lambda if the lambda is exactly the same as the function.

internal/datastore/crdb/schema/indexutil.go
```diff
@@ -188,11 +188,12 @@ func checkFilterColumnMatchesFilter(colName string, filter datastore.Relationshi
 				foundCount++
 			}
 		}
-		if foundCount == 0 {
+		switch {
+		case foundCount == 0:
```
tstirrat15, line 192: Some if/elseif/else rewriting to cases

internal/graph/lookupresources3.go
```diff
@@ -1399,7 +1399,8 @@ func mustDatastoreIndexToString(index datastoreIndex) (string, error) {
 		return dsIndexPrefix + index.chunkID, nil
 	}
 
-	return dsIndexPrefix + index.chunkID + ":" + (*index.dbCursor).String(), nil
+	cursor := (*index.dbCursor)
+	return dsIndexPrefix + index.chunkID + ":" + cursor.String(), nil
```
tstirrat15: `gocritic` was complaining that you could get rid of the deref here, except that you can't because the compiler doesn't correctly recognize the interface. Separating the assignment fixes things.

internal/services/shared/schema.go
```diff
@@ -489,7 +489,9 @@ func errorIfTupleIteratorReturnsTuples(_ context.Context, qy datastore.Relations
 			fullMetadata = make(map[string]string)
 		}
 		fullMetadata["relationship"] = strValue
-		newArgs := append(args, strValue)
+		newArgs := make([]any, 0, len(args)+len(strValue))
+		newArgs = append(newArgs, args)
+		newArgs = append(newArgs, strValue)
```
tstirrat15: I think these are functionally equivalent, but `gocritic` doesn't like it when you use append and assign to a new slice, so this makes it more explicit.

pkg/cmd/server/server_test.go
```diff
@@ -276,7 +275,7 @@ func TestRetryPolicy(t *testing.T) {
 		datastore.WithRequestHedgingEnabled(false),
 	)
 	if err != nil {
-		log.Fatalf("unable to start memdb datastore: %s", err)
```
tstirrat15, line 279: It was complaining here that `log.Fatalf` prevents `defer` blocks from running. `t.Fatalf` should be equivalent and also work around that.

pkg/datastore/test/pagination.go
```diff
@@ -406,9 +406,7 @@ func foreachTxType(
 }
 
 func sortedStandardData(resourceType string, order options.SortOrder) []tuple.Relationship {
-	asTuples := slicez.Map(testfixtures.StandardRelationships, func(item string) tuple.Relationship {
-		return tuple.MustParse(item)
-	})
```
tstirrat15, line 411: Same unlambda thing here.

pkg/validationfile/fileformat.go
```diff
@@ -33,10 +33,12 @@ type ValidationFile struct {
 	ExpectedRelations blocks.ParsedExpectedRelations `yaml:"validation"`
 
 	// NamespaceConfigs are the namespace configuration protos, in text format.
+	//
```
tstirrat15, line 36: `gocritic` wants `Deprecated:` blocks to be separated from the rest of the doc comment.

internal/graph/lookupresources3.go
```diff
@@ -1399,7 +1399,7 @@ func mustDatastoreIndexToString(index datastoreIndex) (string, error) {
 		return dsIndexPrefix + index.chunkID, nil
 	}
 
-	return dsIndexPrefix + index.chunkID + ":" + (*index.dbCursor).String(), nil
+	return dsIndexPrefix + index.chunkID + ":" + index.dbCursor.RelationshipReference.String(), nil
```
tstirrat15, line 1402: I'm not entirely sure how this deref worked previously - this both makes things a little clearer and also satisfies gocritic.

tstirrat15 (commented):
See comment

internal/graph/lookupresources3.go
tstirrat15: Except that it doesn't work 🤔

internal/cursorediterator/blocks_test.go
miparnisari: I don't like lambdas in general because they make reading stacktraces a pain in the behind

internal/datastore/postgres/postgres_shared_test.go
```diff
@@ -1981,11 +1981,9 @@ func RevisionTimestampAndTransactionIDTest(t *testing.T, ds datastore.Datastore)
 				require.True(transactionIDPresent, "expected transactionID to be present in revision")
 
 				checkedUpdate = true
-			} else {
 				// we wait for a checkpoint right after the update. Checkpoints could happen at any time off band.
```
miparnisari: the comment should be moved too

internal/graph/lookupresources3.go
miparnisari: does this explain the broken tests or are you talking about something else?

tstirrat15: I thought this was the test failure but it's not 🤔

internal/services/shared/schema.go
tstirrat15: This change is the problem. I'm not sure why it's semantically different.

PR #2685, opened 2025-11-04
build(deps): migrate to github.com/ccoveille/go-safecast v2

ccoVeille:
## Description

I'm the owner and maintainer of [ccoVeille/go-safecast](https://github.com/ccoVeille/go-safecast)

I have just published my v[2.0.0](https://github.com/ccoVeille/go-safecast/releases/tag/v2.0.0)

Please take a look at the change I introduced to understand the changes I did in your code.

This is somehow a follow-up of https://github.com/authzed/zed/pull/574

- https://github.com/authzed/zed/pull/574

## Testing

```console
$ go test -v ./...
```

### DIFF
--- a/internal/datastore/postgres/revisions_test.go
+++ b/internal/datastore/postgres/revisions_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ccoveille/go-safecast"
+	"github.com/ccoveille/go-safecast/v2"
 	"github.com/stretchr/testify/require"
 )
 
@@ -52,7 +52,7 @@ func TestRevisionSerDe(t *testing.T) {
 	maxSizeList := make([]uint64, 20)
 	for i := range maxSizeList {
 		// i should be nonnegative
-		index, _ := safecast.Convert[uint64](i)
+		index := safecast.RequireConvert[uint64](t, i)
 		maxSizeList[i] = maxInt - uint64(len(maxSizeList)) + index
 	}
 
@@ -97,7 +97,7 @@ func TestRevisionSerDe(t *testing.T) {
 
 func TestTxIDTimestampAvailable(t *testing.T) {
 	// Timestamps should be non-negative
-	testTimestamp, _ := safecast.Convert[uint64](time.Now().Unix())
+	testTimestamp := safecast.RequireConvert[uint64](t, time.Now().Unix())
 	snapshot := snap(0, 5, 1)
 	pgr := postgresRevision{snapshot: snapshot, optionalTxID: NewXid8(1), optionalNanosTimestamp: testTimestamp}
 	receivedTimestamp, ok := pgr.OptionalNanosTimestamp()
--- a/internal/datastore/postgres/snapshot.go
+++ b/internal/datastore/postgres/snapshot.go
@@ -7,7 +7,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/ccoveille/go-safecast"
+	"github.com/ccoveille/go-safecast/v2"
 	"github.com/jackc/pgx/v5/pgtype"
 )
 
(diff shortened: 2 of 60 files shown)


ccoVeille:
I have a few things so fix apparently. No problem.

ccoVeille:
I'm unsure how to deal with issues reported by code coverage.

As you stated in your comment, many of the values cannot be negative even they were defined as signed integer.

go-safecast v2 no longer returns 0 on error, that's why I added this piece of code.

Should I ignore the test ?

I would prefer not to use safecast.MustConvert that panic if there is a conversion issue. 

So I'm interested by feedback here.

miparnisari:
> I'm unsure how to deal with issues reported by code coverage.

We can ignore this. The codecov check isn't mandatory.

miparnisari:
@ccoVeille thank you for sending this! Could you run `go mod tidy` and `mage lint:all`?

internal/datastore/postgres/snapshot.go
```diff
@@ -279,7 +279,11 @@ func (s pgSnapshot) markInProgress(txid uint64) pgSnapshot {
 	startingXipLen := len(newSnapshot.xipList)
 	for numToDrop = 0; numToDrop < startingXipLen; numToDrop++ {
 		// numToDrop should be nonnegative
-		uintNumToDrop, _ := safecast.Convert[uint64](numToDrop)
+		uintNumToDrop, err := safecast.Convert[uint64](numToDrop)
+		if err != nil {
+			uintNumToDrop = 0
```
josephschorr, line 284: I think it might be a good idea to add a new wrapper method for this in spiceerrors called `MustSafecast`. Have it take in the generic type, call the safecast here, then if an error occurs, have it call `spiceerrors.DebugAssert` to fail with a panic at test time (but no-op at prod time) and then return the default value given to the function. That way, we have documented behavior everywhere.

ccoVeille: I can see methods like what you describe here in spiceerrors.

So I could implement something, but first, I would like to check things with you guys, at least the maintainers:

- is there anyone else than Josh who wants to do it the way he described?
- do you expect me to do this ?

I'm asking because I'm unsure I'm the right person to update this.

Just in case, my lib has two methods you could consider:
- MustConvert that panics on failure
- RequireConvert that calls t.Error on failure and so invalidate the tests

Here I feel none of these matches what you described.

Also, I don't understand the need to panic on tests and passes on "production"
If the code differs, it cannot be tested with the same way. I dislike the idea.

That said, I can adapt my PR. But it goes a bit too far from what I expected to do on your project as the owner and maintainer of go-safecast.

I created this PR to facilitate the migration from v1 to v2 for you. I didn't plan to invest more.

Please let me know.

josephschorr: We have a general rule of not allowing panics at production time; instead, we prefer to return errors and panic only during tests to identify bugs but not cause production crashes.

We can change the behavior in a follow, I suppose

ccoVeille: I would prefer you to apply something that suits your habit instead of having a PR that I have to update possibly iteratively because my changes might not face what you are expecting in term of changes.

Please note that go-safecast Convert method doesn't panics either in v1 or v2

So if you have an issue with the way you use my lib with v2, you had it in v1

tstirrat15:
I'm okay with merging this without the aforementioned refactors. We can take that as a follow-on.

One thing that isn't currently working is the e2e tests - they're `go.work`ed to the parent module and they need some updates as well. Is that something that you have time for? Otherwise I can pick this up and keep it moving.

internal/datastore/postgres/revisions_test.go
```diff
@@ -52,7 +52,7 @@ func TestRevisionSerDe(t *testing.T) {
 	maxSizeList := make([]uint64, 20)
 	for i := range maxSizeList {
 		// i should be nonnegative
-		index, _ := safecast.Convert[uint64](i)
+		index := safecast.RequireConvert[uint64](t, i)
```
tstirrat15, line 55: Oooh I like this

PR #2691, opened 2025-11-11
feat: add Memory Protection Middleware

miparnisari:
## Pre-reqs

- https://github.com/authzed/actions/pull/105

## Description

This PR introduces a new memory protection middleware to help prevent out-of-memory conditions in SpiceDB by implementing admission control based on current memory usage.

This is not a perfect solution (doesn't prevent non-traffic-related sources of OOM) and is meant to support other future improvements to resource sharing in a single SpiceDB node.

The middleware is installed both in the main API and in Dispatch APIs, and it uses this lib: https://github.com/odigos-io/go-rtml

⚠️ It does NOT reject health check requests based on memory usage, 
⚠️ It needs a special build flag.

## References

Alternative implementation: https://github.com/authzed/spicedb/pull/2646. 

Upon testing both implementations, it was found that:

1. this PR rejects _less_ requests.
2. this PR didn't OOM the docker container when `mem_limit` was hit.

image

### DIFF
--- a/pkg/cmd/server/defaults.go
+++ b/pkg/cmd/server/defaults.go
@@ -309,15 +309,15 @@ func DefaultUnaryMiddleware(opts MiddlewareOption) (*MiddlewareChain[grpc.UnaryS
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareMemoryProtection).
 			WithInterceptor(selector.UnaryServerInterceptor(
-				memoryProtectionUnaryInterceptor.UnaryServerInterceptor(),
-				selector.MatchFunc(doesNotMatchRoute(healthCheckRoute)))).
-			EnsureAlreadyExecuted(DefaultMiddlewareGRPCProm).         // so that prom middleware reports OOM status
+										memoryProtectionUnaryInterceptor.UnaryServerInterceptor(),
+										selector.MatchFunc(doesNotMatchRoute(healthCheckRoute)))).
+			EnsureAlreadyExecuted(DefaultMiddlewareGRPCProm). // so that prom middleware reports OOM status
 			Done(),
 
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareGRPCAuth).
 			WithInterceptor(grpcauth.UnaryServerInterceptor(opts.AuthFunc)).
-			EnsureAlreadyExecuted(DefaultMiddlewareGRPCProm).         // so that prom middleware reports auth failures
+			EnsureAlreadyExecuted(DefaultMiddlewareGRPCProm). // so that prom middleware reports auth failures
 			Done(),
 
 		NewUnaryMiddleware().
@@ -384,13 +384,13 @@ func DefaultStreamingMiddleware(opts MiddlewareOption) (*MiddlewareChain[grpc.St
 		NewStreamMiddleware().
 			WithName(DefaultMiddlewareMemoryProtection).
 			WithInterceptor(memoryProtectionStreamInterceptor.StreamServerInterceptor()).
-			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareGRPCProm).         // so that prom middleware reports OOM backpressure
+			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareGRPCProm). // so that prom middleware reports OOM backpressure
 			Done(),
 
 		NewStreamMiddleware().
 			WithName(DefaultMiddlewareGRPCAuth).
 			WithInterceptor(grpcauth.StreamServerInterceptor(opts.AuthFunc)).
-			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareGRPCProm).         // so that prom middleware reports auth failures
+			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareGRPCProm). // so that prom middleware reports auth failures
 			Done(),
 
 		NewStreamMiddleware().


magefiles/util.go
```diff
@@ -45,6 +45,8 @@ func testWithArgs(ctx context.Context, args ...string) ([]string, error) {
 		"-failfast",
 		"-count=1",
 		"-race",
+		"-timeout=20m",
```
vroldanbet, line 50: why increasing from 10m to 20m?

pkg/cmd/server/defaults.go
```diff
@@ -319,10 +306,16 @@ func DefaultUnaryMiddleware(opts MiddlewareOption) (*MiddlewareChain[grpc.UnaryS
...
+			WithName(DefaultMiddlewareMemoryProtection).
+			WithInterceptor(memoryProtectionUnaryInterceptor.UnaryServerInterceptor()).
+			Done(),
+
 		NewUnaryMiddleware().
 			WithName(DefaultMiddlewareGRPCAuth).
 			WithInterceptor(grpcauth.UnaryServerInterceptor(opts.AuthFunc)).
-			EnsureAlreadyExecuted(DefaultMiddlewareGRPCProm). // so that prom middleware reports auth failures
+			EnsureAlreadyExecuted(DefaultMiddlewareGRPCProm).         // so that prom middleware reports auth failures
+			EnsureAlreadyExecuted(DefaultMiddlewareMemoryProtection). // so that memory protection happens before auth
```
vroldanbet: Why does it have to run before auth?

The rationale for `EnsureAlreadyExecuted` is to declare a dependency between middlewares explicitly: "I need this to run before me because I depend on its side effects". So why does the auth middleware depend on memory protection?

If not, you can remove it.

If anything, I'd argue that memory protection depends on all the previous middlewares, because we want observability to take place _before we kick memory protection_, otherwise we would be running blind when `ResourceExhausted` is returned.

pkg/cmd/server/defaults.go
```diff
@@ -385,10 +379,16 @@ func DefaultStreamingMiddleware(opts MiddlewareOption) (*MiddlewareChain[grpc.St
...
+			WithName(DefaultMiddlewareMemoryProtection).
+			WithInterceptor(memoryProtectionStreamInterceptor.StreamServerInterceptor()).
+			Done(),
+
 		NewStreamMiddleware().
 			WithName(DefaultMiddlewareGRPCAuth).
 			WithInterceptor(grpcauth.StreamServerInterceptor(opts.AuthFunc)).
-			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareGRPCProm). // so that prom middleware reports auth failures
+			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareGRPCProm).         // so that prom middleware reports auth failures
+			EnsureInterceptorAlreadyExecuted(DefaultMiddlewareMemoryProtection). // so that memory protection happens before auth
```
vroldanbet: same comment

internal/middleware/memoryprotection/memory_protection_test.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			mockCurrentUsageProvider := mocks.NewMockMemoryUsageProvider(ctrl)
+			mockCurrentUsageProvider.EXPECT().IsMemLimitReached().Return(tt.memoryLimitReached).AnyTimes()
```
ecordell: I don't see what the mock is giving you here - you could use the HardCoded provider and get the same effect?

pkg/cmd/server/server.go
```diff
@@ -587,6 +576,37 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 	}, nil
 }
 
+func (c *Config) buildMemoryUsageProvider() memoryprotection.MemoryUsageProvider {
+	if c.MemoryProtectionEnabled {
+		return memoryprotection.NewRealTimeMemoryUsageProvider()
+	}
+	return &memoryprotection.HarcodedMemoryLimitProvider{AcceptAllRequests: true}
```
ecordell, line 583: I would've expected that if the feature was disabled we would just not install the middleware at all, instead of installing it with a no-op limiter?

pkg/cmd/server/server_test.go
```diff
@@ -578,3 +585,81 @@ func TestSupportOldAndNewReadReplicaConnectionPoolFlags(t *testing.T) {
...
+		},
+	}
+
+	for name, tc := range testcases {
+		t.Run(name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			mockDatastore := mocks.NewMockDatastore(ctrl)
+			mockDispatcher := mocks.NewMockDispatcher(ctrl)
```
ecordell: I know mocks are in general contentious, but I don't think you're getting value out of having generated mocks for these vs.  just passing dummy objects of the right type or using the real objects. They're not really being used like mocks.

internal/middleware/memoryprotection/memory_usage_provider.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (n *HarcodedMemoryLimitProvider) IsMemLimitReached() bool {
+	return !n.AcceptAllRequests
+}
+
+type GoRealTimeMemoryLimiter struct{}
+
+func (g GoRealTimeMemoryLimiter) IsMemLimitReached() bool {
+	return rtml.IsMemLimitReached()
```
ecordell, line 29: I wish we had a way to set a threshold here. Since this will get called both for dispatch and primary api calls, it will start rejecting them both equally, when really we want to reject primary api before we start rejecting dispatch.

I don't think this is worth blocking on, but I do wonder if we should contribute a threshold % to the upstream library.

internal/middleware/memoryprotection/memory_protection.go
```diff
@@ -0,0 +1,91 @@
...
+	log "github.com/authzed/spicedb/internal/logging"
+)
+
+// RequestsProcessed tracks requests that were processed by this middleware.
+var RequestsProcessed = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "memory_middleware",
+	Name:      "requests_processed_total",
+	Help:      "Total requests processed by the memory protection middleware (flag --memory-protection-enabled)",
+}, []string{"endpoint", "accepted"})
```
tstirrat15, line 24: Does this give us anything that a request counter doesn't? Should we change this to be "requests rejected" or something?

internal/middleware/memoryprotection/memory_protection.go
```diff
@@ -0,0 +1,91 @@
...
+// recordMetric updates the RequestsProcessed metric and returns the endpoint type for the input method.
+func (am *MemoryProtectionMiddleware) recordMetric(fullMethod string, accepted bool) string {
+	endpointType := "api"
+	if strings.HasPrefix(fullMethod, "/dispatch.v1.DispatchService") {
+		endpointType = "dispatch"
+	}
+
+	acceptedStr := strconv.FormatBool(accepted)
+
+	RequestsProcessed.WithLabelValues(endpointType, acceptedStr).Inc()
```
tstirrat15, line 89: Ah, we're doing it with a label? Hmm...

tstirrat15 (commented):
See comments, otherwise looks good.

pkg/cmd/server/server.go
miparnisari: I'm wondering now why do we have the `MemoryProtectionEnabled` flag at all 🤔

pkg/cmd/server/server_test.go
miparnisari: Yeah, i know. I'm not using the mocks at all, i just needed something to put in the function calls, and I know how to instantiate mocks, so I used them 🤷‍♀️

internal/middleware/memoryprotection/memory_usage_provider.go
miparnisari: Good idea. Opened https://github.com/odigos-io/go-rtml/issues/3

magefiles/util.go
miparnisari: Ah sorry, that was a leftover from a different PR. The answer is https://github.com/authzed/spicedb/pull/2674/files#r2505759338

Tl;dr instead of putting timeouts in the test code itself, i think it's better to put it in `go test -timeout`

pkg/cmd/server/defaults.go
miparnisari: Oh this was a leftover from @ecordell's original PR. I don't have a strong opinion around this, I didn't know that `EnsureAlreadyExecuted` meant an explicit dependency.

@ecordell do you have any thoughts?

pkg/cmd/server/server.go
tstirrat15: I'd be in favor of having it be always-on, and then we just don't install the middleware in tests (if that's possible).

tstirrat15 (commented):
I'd be keen to see what @ecordell has to say, and I'm down with the idea of making this memory middleware non-optional.

pkg/cmd/server/server.go
miparnisari: I looked at our existing middlewares to see if we have an example of one that is conditionally installed, and it doesn't exist... if a middleware isn't enabled, it still runs, but it just no-ops. Example: https://github.com/authzed/spicedb/blob/d4ff660514fb61eea9d8d9d58520d9f00eaefd4f/internal/middleware/perfinsights/perfinsights.go#L157-L160

pkg/cmd/server/defaults.go
ecordell: I don't have a strong opinion here, I just thought that the earlier the memory protection could run the better.

pkg/cmd/server/server.go
tstirrat15: Captured in #2723

tstirrat15: And I think I'm going to split the difference and not expose it as a user-facing flag, but leave the ability to configure it. My thought is that if (e.g.) our steelthread tests start hitting OOM conditions that trigger this middleware, it'll be good to disable the middleware so that we get failures instead of `ResourceExhausted`s

magefiles/build.go
```diff
@@ -13,9 +13,9 @@ type Build mg.Namespace
 
 // Binary builds the binary
 func (Build) Binary() error {
-	return sh.RunWithV(
-		map[string]string{},
```
tstirrat15, line 17: No vars, so don't need `With`

tstirrat15 (dismissed):
This looks like it should be good.

PR #2696, opened 2025-11-13
test: remove timeouts in flaky tests

miparnisari:
- Remove timeouts in `TestMigrate/cockroachdb`
- Remove timeouts in  `TestSchemaWatch/cockroachdb`
- Adds a default `-timeout=20m` to all tests

### DIFF
--- a/cmd/spicedb/migrate_integration_test.go
+++ b/cmd/spicedb/migrate_integration_test.go
@@ -4,11 +4,9 @@ package main
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"slices"
 	"testing"
-	"time"
 
 	"github.com/google/uuid"
 	"github.com/ory/dockertest/v3"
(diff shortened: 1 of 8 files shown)


cmd/spicedb/migrate_integration_test.go
```diff
@@ -88,11 +86,6 @@ func TestMigrate(t *testing.T) {
 
 				require.Fail(t, "Got non-zero exit code", stream.String())
 			}
-
-			t.Cleanup(func() {
-				// When you're done, kill and remove the container
-				_ = pool.Purge(resource)
-			})
```
miparnisari, line 95: FYI cleanups should be written right after the resource was created

cmd/spicedb/restgateway_integration_test.go
```diff
@@ -26,7 +26,6 @@ func TestRESTGateway(t *testing.T) {
 		false,
 	)
 	require.NoError(err)
-	defer tester.cleanup()
```
miparnisari, line 29: FYI now the `newTester` code does the cleanup automatically

PR #2698, opened 2025-11-13
test: create mage target to run e2e tests

miparnisari:
# Description

Followup of https://github.com/authzed/spicedb/pull/2665

- create mage targets
- prevent `TestNoNewEnemy/protected_from_data_newenemy_with_standard_overlap` from flaking out because the `-timeout=10m` is too short

## Testing

@tstirrat15 could you test the mage target on your machine?

### DIFF
--- a/magefiles/test.go
+++ b/magefiles/test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/magefile/mage/mg"
 	"github.com/magefile/mage/sh"
+	"golang.org/x/sync/errgroup"
 )
 
 type Test mg.Namespace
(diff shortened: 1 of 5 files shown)


magefiles/test.go
```diff
@@ -62,6 +63,35 @@ func (Test) Integration(ctx context.Context) error {
 	return goTest(ctx, "./internal/services/integrationtesting/...", "-tags", "ci,docker", "-timeout", "15m")
 }
 
+// e2e Runs e2e tests (new enemy)
+func (Test) E2e(ctx context.Context, crdbVersion string) error {
+	if crdbVersion == "" {
+		// TODO 25.x don't support manually overriding the clock
+		// so you will see ERROR: request timestamp 1763064549.980401596,0 too far in future
+		return fmt.Errorf("no crdb version provided. try 25.2.0 or 25.3.0")
```
miparnisari, line 71: this `if` doesn't actually work lol

PR #2699, opened 2025-11-13
chore: unflake sleeper tests using synctest

tstirrat15:
## Description
These are a prime candidate for synctest - all of the logic is based around sleeping and waiting. This makes these tests deterministic and (relatively) instantaneous.

## Changes
* Use `synctest` to make these tests deterministic
## Testing
Review.

### DIFF
--- a/internal/dispatch/remote/primarysleeper_test.go
+++ b/internal/dispatch/remote/primarysleeper_test.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/require"
 )
 
 func TestPrimarySleeper_SleepWithZeroWaitTime(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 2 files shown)


internal/dispatch/remote/primarysleeper_test.go
```diff
@@ -253,25 +291,15 @@ func TestPrimarySleeper_EdgeCaseZeroDurationContext(t *testing.T) {
...
+	synctest.Test(t, func(t *testing.T) {
+		ctx, cancel := context.WithDeadline(t.Context(), time.Now())
+		defer cancel()
 
-	require.Less(t, elapsed, 5*time.Millisecond)
-}
-
-func TestPrimarySleeper_MetricsAreObserved(t *testing.T) {
-	waitTime := 5 * time.Millisecond
-	sleeper := &primarySleeper{
```
tstirrat15, line 268: As far as I can tell, this test isn't actually asserting anything, so I removed it.

tstirrat15 (commented):
See comments

PR #2702, opened 2025-11-14
fix: schema/relationship generator does not conform to regex 

vroldanbet:
## Description

The Schema/Relationship Generator in the `schema/v2/testing` package does not conform to the regular expression defined for namespaces, object types, object IDs and relations. 

## Testing

A test was added that generates 1K relationships and parses it using the reference regular expression


vroldanbet:
Still caused it to fail, I believe it's due to lack of limit in length. Will look into it.

PR #2703, opened 2025-11-14
chore: add docs command to be used by mage

miparnisari:
In https://github.com/authzed/spicedb/pull/2688 I introduced a `mage` command that generates docs, which is great, except I now realize that adding a dependency inside `magefiles/go.mod` to `spicedb` will be problematic for https://github.com/authzed/spicedb/pull/2691. This PR needs a build flag, and because of this dependency, the build flag will also be needed to run `mage`, which I don't want.


tstirrat15 (approved):
LGTM

PR #2704, opened 2025-11-14
feat: implement a static statistics-based optimizer

barakmich:

### DIFF
--- a/pkg/query/statistics_optimizer.go
+++ b/pkg/query/statistics_optimizer.go
@@ -57,9 +57,8 @@ func (s StatisticsOptimizer) reorderIntersection(it Iterator) (Iterator, bool, e
 }
 
 // rebalanceArrow rebalances arrow operators to minimize total cost.
-// If an arrow contains nested arrows on either side (possibly through wrapper
-// iterators like Alias or CaveatIterator), we can restructure them to reduce
-// the overall computation cost.
+// If an arrow contains nested arrows on either side, we can restructure them
+// to reduce the overall computation cost.
 //
 // For example: (A->B)->C can be rebalanced to A->(B->C) if that's cheaper.
 // The key insight is that arrow operators are left-associative but we can
@@ -70,16 +69,16 @@ func (s StatisticsOptimizer) rebalanceArrow(it Iterator) (Iterator, bool, error)
 		return it, false, nil
 	}
 
-	// Check if left side contains an arrow (possibly wrapped): wrapped(A->B)->C
-	if leftArrow, leftWrappers := unwrapToArrow(arrow.left); leftArrow != nil {
-		if alternative, changed, err := s.tryRebalance(arrow, leftArrow, leftWrappers, true); err != nil || changed {
+	// Check if left side is an arrow: (A->B)->C
+	if leftArrow, ok := arrow.left.(*Arrow); ok {
+		if alternative, changed, err := s.tryRebalance(arrow, leftArrow, true); err != nil || changed {
 			return alternative, changed, err
 		}
 	}
 
-	// Check if right side contains an arrow (possibly wrapped): A->wrapped(B->C)
-	if rightArrow, rightWrappers := unwrapToArrow(arrow.right); rightArrow != nil {
-		if alternative, changed, err := s.tryRebalance(arrow, rightArrow, rightWrappers, false); err != nil || changed {
+	// Check if right side is an arrow: A->(B->C)
+	if rightArrow, ok := arrow.right.(*Arrow); ok {
+		if alternative, changed, err := s.tryRebalance(arrow, rightArrow, false); err != nil || changed {
 			return alternative, changed, err
 		}
 	}
@@ -87,75 +86,9 @@ func (s StatisticsOptimizer) rebalanceArrow(it Iterator) (Iterator, bool, error)
 	return it, false, nil
 }
 
-// unwrapToArrow looks through single-subiterator wrapper iterators
-// (Alias, CaveatIterator, etc.) to find an underlying Arrow.
-// Returns the arrow and the path of wrappers, or nil if no arrow is found.
-func unwrapToArrow(it Iterator) (*Arrow, []Iterator) {
-	var wrappers []Iterator
-	current := it
-
-	// Keep unwrapping single-subiterator wrappers
-	for {
-		subs := current.Subiterators()
-		if len(subs) != 1 {
-			// Not a single-subiterator wrapper
-			break
-		}
-
-		// Check if current is an arrow
-		if arrow, ok := current.(*Arrow); ok {
-			return arrow, wrappers
-		}
-
-		// It's a wrapper, record it and continue
-		wrappers = append(wrappers, current)
-		current = subs[0]
-	}
-
-	// Check the final iterator
-	if arrow, ok := current.(*Arrow); ok {
-		return arrow, wrappers
-	}
-
-	return nil, nil
-}
-
-// rewrapIterator wraps an iterator with the given wrapper chain,
-// properly handling CaveatIterators by checking which branch they should apply to.
-// The wrappers are applied in order (innermost to outermost).
-func rewrapIterator(inner Iterator, wrappers []Iterator, leftBranch, rightBranch Iterator) (Iterator, error) {
-	result := inner
-	for _, wrapper := range wrappers {
-		// Special handling for CaveatIterator - only apply if the caveat is relevant to the branches
-		if caveatIt, ok := wrapper.(*CaveatIterator); ok {
-			// Check if the caveat applies to left or right branch
-			leftHasCaveat := containsCaveat(leftBranch, caveatIt.caveat)
-			rightHasCaveat := containsCaveat(rightBranch, caveatIt.caveat)
-
-			// Only wrap if the caveat applies to at least one branch
-			if !leftHasCaveat && !rightHasCaveat {
-				// Caveat doesn't apply to either branch, skip it
-				continue
-			}
-
-			// If caveat applies to both branches, we need to wrap the whole thing
-			// If it only applies to one, we still wrap the whole thing since we're
-			// at the level where the arrow combines them
-		}
-
-		// Replace the wrapper's subiterator with our current result
-		newWrapper, err := wrapper.ReplaceSubiterators([]Iterator{result})
-		if err != nil {
-			return nil, err
-		}
-		result = newWrapper
-	}
-	return result, nil
-}
-
 // tryRebalance attempts to rebalance an arrow with a nested arrow on one side.
 // Returns the rebalanced iterator and true if the rebalancing was cheaper, nil and false otherwise.
-func (s StatisticsOptimizer) tryRebalance(original *Arrow, nestedArrow *Arrow, wrappers []Iterator, isLeftSide bool) (Iterator, bool, error) {
+func (s StatisticsOptimizer) tryRebalance(original *Arrow, nestedArrow *Arrow, isLeftSide bool) (Iterator, bool, error) {
 	// Calculate cost of original
 	originalCost, err := s.Source.Cost(original)
 	if err != nil {
@@ -163,36 +96,18 @@ func (s StatisticsOptimizer) tryRebalance(original *Arrow, nestedArrow *Arrow, w
 	}
 
 	// Build the alternative based on which side the nested arrow is on
-	var innerArrow *Arrow
 	var alternative *Arrow
-	var leftBranch, rightBranch Iterator
 
 	if isLeftSide {
-		// Original: wrapped(A->B)->C
-		// Alternative: A->wrapped(B->C)
-		leftBranch = nestedArrow.right
-		rightBranch = original.right
-		innerArrow = NewArrow(leftBranch, rightBranch)
-
-		wrappedInner, err := rewrapIterator(innerArrow, wrappers, leftBranch, rightBranch)
-		if err != nil {
-			return nil, false, err
-		}
-
-		alternative = NewArrow(nestedArrow.left, wrappedInner)
+		// Original: (A->B)->C
+		// Alternative: A->(B->C)
+		innerArrow := NewArrow(nestedArrow.right, original.right)
+		alternative = NewArrow(nestedArrow.left, innerArrow)
 	} else {
-		// Original: A->wrapped(B->C)
-		// Alternative: wrapped(A->B)->C
-		leftBranch = original.left
-		rightBranch = nestedArrow.left
-		innerArrow = NewArrow(leftBranch, rightBranch)
-
-		wrappedInner, err := rewrapIterator(innerArrow, wrappers, leftBranch, rightBranch)
-		if err != nil {
-			return nil, false, err
-		}
-
-		alternative = NewArrow(wrappedInner, nestedArrow.right)
+		// Original: A->(B->C)
+		// Alternative: (A->B)->C
+		innerArrow := NewArrow(original.left, nestedArrow.left)
+		alternative = NewArrow(innerArrow, nestedArrow.right)
 	}
 
 	// Calculate cost of alternative
--- a/pkg/query/statistics_optimizer_test.go
+++ b/pkg/query/statistics_optimizer_test.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/stretchr/testify/require"
 
-	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/schema/v2"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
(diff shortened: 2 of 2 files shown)


pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+	// Cost returns a cost estimate for the given iterator.
+	Cost(it Iterator) (Estimate, error)
+}
+
+// StaticStatistics provides static cost estimates for iterators based on
+// configurable parameters. This is useful for basic query planning and
+// when dynamic statistics are not available.
+type StaticStatistics struct {
+	// DatastoreSize is the assumed number of tuples in a relation.
+	DatastoreSize int
```
miparnisari:
```suggestion
	NumberOfTuplesInRelation int
```

also for _what_ relation? or is it an average like `fanout`?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+	// Fanout is the assumed average number of subjects per resource or
+	// resources per subject.
+	Fanout int
+
+	// CheckSelectivity is the default probability (0.0-1.0) that a Check
+	// operation will return true.
+	CheckSelectivity float64
+
+	// IntersectionArrowReduction is an additional reduction factor for
+	// IntersectionArrow selectivity to account for "all subjects" semantics.
```
miparnisari: i'm not sure why `IntersectionArrowReduction` is a customizable field. can we not calculate it ourselves based on some other data?

if the check is for `group.all(member)` (ie `N.all(M)`), then the chance of a check returning true is directly related to the size of `N`, right?

put differently: given two sets of size N and M. 

- what are the chances that **one** element in N is also present in M? (this is the probability of `check(arrow)` returning `true`)
- what are the chances that **all** elements in N are also present in M? (this is the probability of `check(intersection arrow)` returning `true`)

can we not derive formulas for that?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+// Estimate represents the estimated cost and selectivity metrics for an iterator.
+// These estimates are used by the query optimizer to make decisions about query plan structure.
+type Estimate struct {
+	Cardinality int // Cardinality is the estimated number of results this iterator will produce.
+
+	// CheckSelectivity is the estimated probability (0.0-1.0) that a Check operation
+	// will return true. Higher values mean the check is more likely to pass.
+	CheckSelectivity float64
+
+	// CheckCost is the estimated cost to perform a Check operation on this iterator.
```
miparnisari, line 24: what unit is used and what is the range? (zero-math.max?)

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+// StaticStatistics provides static cost estimates for iterators based on
+// configurable parameters. This is useful for basic query planning and
+// when dynamic statistics are not available.
+type StaticStatistics struct {
+	// DatastoreSize is the assumed number of tuples in a relation.
+	DatastoreSize int
+
+	// Fanout is the assumed average number of subjects per resource or
+	// resources per subject.
+	Fanout int
```
miparnisari, line 58: why is this not two distinct values `FanoutSubjectsPerResource` and `FanoutResourcesPerSubject`? they can vary wildly

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+		ls, err := s.Cost(it.left)
+		if err != nil {
+			return Estimate{}, err
+		}
+		rs, err := s.Cost(it.right)
+		if err != nil {
+			return Estimate{}, err
+		}
+		// Arrow does IterSubjects on left (ls.Cardinality iterations),
+		// then Check on right for each subject (rs.CheckCost per iteration)
```
miparnisari: why this comment here? this is the current implementation of arrow resolution but it will change in the future

or is this a worst-case scenario estimation, which is based on the naive implementation of arrow that has zero concurrency?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+		return Estimate{
+			Cardinality:       ls.Cardinality * rs.Cardinality,
```
miparnisari, line 111: Why is this a `*` and not a `+`? if the schema has `permission view = parent->read` and you have 3 parent folders and the user can read 5 folders, you're not doing 5*3=15 database reads, you're doing 8 at most

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
+package query
+
+import (
+	"errors"
+
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+// Estimate represents the estimated cost and selectivity metrics for an iterator.
```
miparnisari:
```suggestion
// Estimate represents the estimated worst-case scenario cost and selectivity metrics for an iterator.
```

also might want to define "selectivity"

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		rs, err := s.Cost(it.right)
+		if err != nil {
+			return Estimate{}, err
+		}
+		// IntersectionArrow also does IterSubjects on left, then Check on right,
+		// but only yields results when ALL subjects satisfy (more selective)
+		selectivity := ls.CheckSelectivity * rs.CheckSelectivity * s.IntersectionArrowReduction // Additional reduction for "all" semantics
+		return Estimate{
+			Cardinality:       int(float64(ls.Cardinality*rs.Cardinality) * selectivity),
```
miparnisari, line 130: i don't understand why you are multiplying by selectivity which is a probability. in the worst case scenario you will have to read everything

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+		// Arrow does IterSubjects on left (ls.Cardinality iterations),
+		// then Check on right for each subject (rs.CheckCost per iteration)
+		return Estimate{
+			Cardinality:       ls.Cardinality * rs.Cardinality,
+			CheckCost:         ls.IterSubjectsCost + (ls.Cardinality * rs.CheckCost),
+			CheckSelectivity:  ls.CheckSelectivity * rs.CheckSelectivity,
+			IterResourcesCost: ls.IterResourcesCost + (ls.Cardinality * rs.IterResourcesCost),
+			IterSubjectsCost:  ls.IterSubjectsCost + (ls.Cardinality * rs.IterSubjectsCost),
+		}, nil
+	case *IntersectionArrow:
```
miparnisari, line 117: i just had the random realization that in the case of intersection arrow `N.all(M)`, if the set `M` has less results than the set `N` we can return `check=false` immediately, no? is this captured in our planned optimizations? @josephschorr  does Check today take this shortcut?

pkg/query/statistics.go
tstirrat15: Is this always the case, or is this one strategy? What happens if/when this gets rewritten?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+
+// Cost returns a cost estimate for the given iterator using static assumptions.
+// It recursively estimates costs for composite iterators by combining the costs
+// of their subiterators according to their operational semantics.
+func (s StaticStatistics) Cost(iterator Iterator) (Estimate, error) {
+	switch it := iterator.(type) {
+	case *FixedIterator:
+		return Estimate{
+			Cardinality:       len(it.paths),
+			CheckCost:         10,
```
tstirrat15: What does this cost being 10 mean? Or is a cost unitless and only makes sense when comparing two costs?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+			if err != nil {
+				return Estimate{}, err
+			}
+			result.Cardinality += est.Cardinality
+			result.CheckCost += est.CheckCost
+			result.IterResourcesCost += est.IterResourcesCost
+			result.IterSubjectsCost += est.IterSubjectsCost
+			if est.CheckSelectivity > result.CheckSelectivity {
+				result.CheckSelectivity = est.CheckSelectivity
+			}
```
miparnisari: nit

```suggestion
			result.CheckSelectivity = max(result.CheckSelectivity, est.CheckSelectivity)
```

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+			Cardinality:       len(it.paths),
+			CheckCost:         10,
+			CheckSelectivity:  s.CheckSelectivity,
+			IterResourcesCost: len(it.paths),
+			IterSubjectsCost:  len(it.paths),
+		}, nil
+	case *RelationIterator:
+		return Estimate{
+			Cardinality:       s.DatastoreSize,
+			CheckCost:         s.DatastoreSize,
```
tstirrat15: Like here - what does it mean that the check cost is the datastore size?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+			est, err := s.Cost(subIt)
+			if err != nil {
+				return Estimate{}, err
+			}
+
+			if idx == 0 {
+				result.Cardinality = est.Cardinality
+			} else {
+				// Each subsequent intersection filters down the results
+				result.Cardinality = int(float64(result.Cardinality) * est.CheckSelectivity)
```
miparnisari: i don't know if i'm misunderstanding this line or the concept of `cardinality` as a whole. But in my mind, comparing a union and an intersection, the cardinality is the same because in the worst case you will have to read from all iterators involved

pkg/query/statistics_optimizer.go
```diff
@@ -0,0 +1,268 @@
...
+	if !ok {
+		return it, false, nil
+	}
+
+	newSubs, changed, err := s.reorderBySelectivity(union.subIts, false)
+	if err != nil || !changed {
+		return it, changed, err
+	}
+
+	return NewUnion(newSubs...), true, nil
```
tstirrat15, line 39: Don't we usually parallelize these dispatches? Does reordering still work in that case? Or are we leaving that as a later exercise?

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+
+		// Cardinality: main minus the excluded portion
+		// Approximate as: main.Cardinality * (1 - excluded.CheckSelectivity)
+		exclusionFactor := 1.0 - excludedEst.CheckSelectivity
+		if exclusionFactor < 0 {
+			exclusionFactor = 0
+		}
+
+		return Estimate{
+			Cardinality:       int(float64(mainEst.Cardinality) * exclusionFactor),
```
miparnisari, line 192: similar to my other comment:  why is this not the sum of `main` and `excluded`? you are mixing counts with probabilities which doesn't make sense to me

pkg/query/statistics_optimizer.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+
+		// Check if current is an arrow
+		if arrow, ok := current.(*Arrow); ok {
+			return arrow, wrappers
+		}
+
+		// It's a wrapper, record it and continue
+		wrappers = append(wrappers, current)
+		current = subs[0]
```
tstirrat15: Sanity check: we're guaranteed to have a finite tree here?

pkg/query/statistics_optimizer.go
```diff
@@ -0,0 +1,268 @@
...
+	// Return alternative if cheaper
+	if alternativeCost.CheckCost < originalCost.CheckCost {
+		return alternative, true, nil
+	}
+
+	return nil, false, nil
+}
+
+// subWithCost pairs an iterator with its selectivity for sorting.
+type subWithCost struct {
```
miparnisari:
```suggestion
type iteratorWithSelectivity struct {
```

pkg/query/statistics_optimizer_test.go
```diff
@@ -0,0 +1,512 @@
...
+			MustPathFromString("document:doc4#viewer@user:dave"),
+			MustPathFromString("document:doc5#viewer@user:eve"),
+			MustPathFromString("document:doc6#viewer@user:frank"),
+		)
+
+		union := NewUnion(sub1, sub2, sub3)
+		_, changed, err := optimizer.Optimize(union)
+		require.NoError(t, err)
+
+		// Since all have same selectivity, order shouldn't change
```
miparnisari, line 54: the test is named `reorder by selectivity`. i expected to see some reordering. you already have a test case for the no-reordering case.

put differently, where is the case with `require.True(t, changed)`? (same comment applies for reordering intersection)

pkg/query/statistics_optimizer_test.go
```diff
@@ -0,0 +1,512 @@
...
+
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/schema/v2"
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+func TestStatisticsOptimizer_ReorderUnion(t *testing.T) {
+	t.Parallel()
+
+	stats := DefaultStaticStatistics()
```
miparnisari, line 15: why not use `subWithCost` here? so you can give different static selectivities to different iterators

pkg/query/statistics.go
```diff
@@ -0,0 +1,210 @@
...
+	IterSubjectsCost int
+}
+
+// StatisticsSource provides cost estimates for iterators.
+// Implementations can provide static estimates or dynamic estimates based on
+// actual datastore statistics.
+type StatisticsSource interface {
+	// Cost returns a cost estimate for the given iterator.
+	Cost(it Iterator) (Estimate, error)
+}
```
miparnisari, line 44: i think i would have designed this differently, with the decorator pattern:

```go
type IteratorWithCost interface {
    Iterator
    Cost(statisticsAroundTheDatabase) (Estimate, error)
}
```

i.e. you _ask_ the iterator what its cost is. This makes it easy to mock out iterators with specific estimates.

Then you could change this 

```go
func (s StatisticsOptimizer) reorderBySelectivity(subs []Iterator
```

to

```go
func (s StatisticsOptimizer) reorderBySelectivity(subs []IteratorWithCost
```

pkg/query/statistics_optimizer_test.go
```diff
@@ -0,0 +1,512 @@
...
+		innerArrow := NewArrow(b, c)
+		alternative := NewArrow(a, innerArrow)
+		alternativeCost, err := stats.Cost(alternative)
+		require.NoError(t, err)
+
+		result, changed, err := optimizer.Optimize(originalArrow)
+		require.NoError(t, err)
+
+		// Check if optimization occurred when expected
+		if alternativeCost.CheckCost < originalCost.CheckCost {
```
miparnisari: I don't like having `if`s in tests. Why not mock out the cost of each arrow to specific values?

pkg/query/statistics_optimizer_test.go
```diff
@@ -0,0 +1,512 @@
...
+			require.True(t, changed, "Should have rebalanced when alternative is cheaper")
+			// Verify structure
+			resultArrow, ok := result.(*Arrow)
+			require.True(t, ok)
+			_, isArrow := resultArrow.right.(*Arrow)
+			require.True(t, isArrow, "Right side should be an arrow after rebalancing")
```
miparnisari: i don't think this assertion is enough. don't you need to unwrap the returned arrows and compare against `a, b, c`?

miparnisari (changes requested):

pkg/query/statistics_optimizer.go
```diff
@@ -0,0 +1,268 @@
...
+// If ascending is false, sorts by higher selectivity first (for unions).
+func (s StatisticsOptimizer) reorderBySelectivity(subs []Iterator, ascending bool) ([]Iterator, bool, error) {
+	if len(subs) <= 1 {
+		return subs, false, nil
+	}
+
+	// Get cost estimates for each subiterator
+	subsWithCost := make([]subWithCost, len(subs))
+	for i, sub := range subs {
+		est, err := s.Source.Cost(sub)
```
tstirrat15, line 144: General thought: should we treat these cost calculations as expensive? i.e. would it be worth memoizing them?

pkg/query/statistics_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+		// IntersectionArrow has reduced selectivity due to "all" semantics
+		// selectivity = left.CheckSelectivity * right.CheckSelectivity * IntersectionArrowReduction
+		// = 0.9 * 0.9 * 0.5 = 0.405
+		expectedSelectivity := 0.9 * 0.9 * 0.5
+		require.InDelta(t, expectedSelectivity, est.CheckSelectivity, 0.001)
+
+		// Cardinality: int(left.Cardinality * right.Cardinality * selectivity)
+		// = int(3 * 1 * 0.405) = 1
+		require.Equal(t, 1, est.Cardinality)
```
tstirrat15, line 129: Does floating point arithmetic work such that this won't flake? or are we coercing to an int so it doesn't matter?

pkg/query/statistics.go
barakmich: The units are pretty arbitrary. But I added some commentary to improve this.

pkg/query/statistics.go
barakmich: They can vary wildly! But by simply assuming that there is some amount of fanout, in either direction, it's a way of taking the added cost of doing the operation into account. These statistics are meant to be really dumb 🙃

pkg/query/statistics.go
barakmich: expanded

pkg/query/statistics.go
barakmich: Worst case; the size of the arrow is the size of the full outer join, which is the `*` of the cardinality. It's probably smaller, true. Adding a comment to this effect.

pkg/query/statistics.go
barakmich: So I definitely considered this, and in fact, was planning it the way you describe. The reason I did it the way I did here is that I was working on the plan that there are multiple statistics sources, which may have different ways of computing the statistics. You can't easily ask an iterator what it's costs are if there are parts of `statisticsAroundTheDatabase` that exist in some sources and not others. We could try to surface the right interface so that the iterator covered all possibilities, but I decided to go the other way.

pkg/query/statistics.go
miparnisari: hmm. okay. our implementation would have to be super broken to get to that worst case scenario 😄

pkg/query/statistics.go
miparnisari: i see this was resolved but i don't understand why

pkg/query/statistics_optimizer_test.go
miparnisari: this is unresolved?

pkg/query/statistics.go
```diff
@@ -0,0 +1,207 @@
...
+			IterResourcesCost: len(it.paths),
+			IterSubjectsCost:  len(it.paths),
+		}, nil
+	case *RelationIterator:
+		return Estimate{
+			Cardinality:       s.NumberOfTuplesInRelation,
+			CheckCost:         1,
+			CheckSelectivity:  s.CheckSelectivity,
+			IterResourcesCost: s.NumberOfTuplesInRelation * s.Fanout,
+			IterSubjectsCost:  s.NumberOfTuplesInRelation * s.Fanout,
```
tstirrat15: `s.Fanout`

pkg/query/statistics_optimizer.go
```diff
@@ -0,0 +1,268 @@
...
+		return arrow, wrappers
+	}
+
+	return nil, nil
+}
+
+// rewrapIterator wraps an iterator with the given wrapper chain,
+// properly handling CaveatIterators by checking which branch they should apply to.
+// The wrappers are applied in order (innermost to outermost).
+func rewrapIterator(inner Iterator, wrappers []Iterator, leftBranch, rightBranch Iterator) (Iterator, error) {
```
tstirrat15: We chatted about lifting this out of this PR and having arrow rebalancing be its own PR.

pkg/query/statistics_optimizer.go
barakmich: Later exercise

PR #2705, opened 2025-11-14
chore: introduce mockgen

miparnisari:
This will be used in https://github.com/authzed/spicedb/pull/2691


tstirrat15 (approved):
Yep, this looks good to me.

miparnisari:
@barakmich i checked with Joey and he's fine with this :) 

The mocks are automatically generated on the existing `mage gen:go` command.

PR #2719, opened 2025-11-20
Feat/ds last observed revision

ostafen:


ostafen:
Changes are no more needed (switching to different approach).

PR #2722, opened 2025-11-21
chore: unflake cursored parallel iterators test by refactoring iterator logic

tstirrat15:
## Description
This is a flake that we were seeing relatively frequently: https://github.com/authzed/spicedb/actions/runs/19578836216/job/56071339937?pr=2691

The issue was that the test [expected that the error iterator would produce a result before the error](https://github.com/authzed/spicedb/blob/b55a9f65aaacba19ec0b3eb004529a0d0bca4530/internal/cursorediterator/blocks_test.go#L397-L400), but the code was effectively racing the result and the error between a pair of channels that held results and errors.

This refactors it so that the `item` that's passed through the channel holds the error as well, so we're no longer racing channels.

## Changes
* Refactor `CursoredParallelIterators` to remove `errChannels`
## Testing
Review. See that tests pass and no longer flake.

### DIFF
--- a/internal/cursorediterator/blocks.go
+++ b/internal/cursorediterator/blocks.go
@@ -146,6 +146,7 @@ func CursoredWithIntegerHeader[I any](
 // item is a struct that holds an item or an error.
 type item[I any] struct {
 	itemAndCursor ItemAndCursor[I]
+	err           error
 	completed     bool
 }
 
@@ -239,15 +240,11 @@ func CursoredParallelIterators[I any](
 
 	tr := taskrunner.NewPreloadedTaskRunner(ctx, concurrency, len(itersToRun))
 	collectors := make([]chan item[I], len(itersToRun))
-	errChannels := make([]chan error, len(itersToRun))
 	for i, iter := range itersToRun {
 		i := i
 		iter := iter
 		collector := make(chan item[I], parallelIteratorResultsBufferSize)
 		collectors[i] = collector
-
-		errChannel := make(chan error, 1)
-		errChannels[i] = errChannel
 
 		// If not the first iterator being executed by this invocation, then we set
 		// the remaining Cursor to nil, as it cannot apply to subsequent iterators,
@@ -260,17 +257,17 @@ func CursoredParallelIterators[I any](
 
 		tr.Add(func(ctx context.Context) error {
 			for r, err := range iter(ctx, iteratorRemainingCursor) {
-				if err != nil {
-					errChannel <- err
-					return nil
-				}
-
 				select {
 				case <-ctx.Done():
-					errChannel <- ctx.Err()
-					return nil
+					return ctx.Err()
 
-				case collector <- item[I]{r, false}:
+				// Propagate the result and the error up to the collector
+				case collector <- item[I]{r, err, false}:
+					if err != nil {
+						// Stop processing if an error was encountered
+						return nil
+					}
+					// Otherwise continue
 					continue
 				}
 			}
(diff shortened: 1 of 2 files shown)


internal/cursorediterator/blocks.go
```diff
@@ -290,19 +286,19 @@ func CursoredParallelIterators[I any](
...
 
 			for {
 				select {
 				case <-ctx.Done():
 					_ = yield(ItemAndCursor[I]{}, ctx.Err())
 					return
 
-				case err := <-errChannel:
-					_ = yield(ItemAndCursor[I]{}, err)
-					return
```
tstirrat15, line 303: The tests were flaking because there was a race between a result coming through the collector and an error coming through the `errChannel`. Removing the `errChannel` from the picture fixes the problem.

internal/cursorediterator/blocks.go
```diff
@@ -261,16 +258,15 @@ func CursoredParallelIterators[I any](
...
 				if err != nil {
-					errChannel <- err
+					collector <- item[I]{r, err, true}
 					return nil
 				}
 
 				select {
 				case <-ctx.Done():
-					errChannel <- ctx.Err()
-					return nil
```
tstirrat15, line 271: The `ctx.Done()` is also handled in the collector logic, so we can terminate the task/goroutine when we've encountered the `Done()`.

internal/cursorediterator/blocks.go
```diff
@@ -260,17 +257,17 @@ func CursoredParallelIterators[I any](
...
-
 				select {
 				case <-ctx.Done():
-					errChannel <- ctx.Err()
-					return nil
+					return ctx.Err()
 
-				case collector <- item[I]{r, false}:
+				// Propagate the result and the error up to the collector
+				case collector <- item[I]{r, err, false}:
```
tstirrat15, line 265: I moved this down to prevent the case where context gets cancelled and the collector attempts to produce while the consumer is shutting down.

I think there's still a potential for a race here if the consumer shuts down in between ticks, so to speak. I think the only guarantee we'd get here is if we were somehow able to ensure that the producers were cancelled before the consumer was cancelled.

tstirrat15 (commented):
See comments

internal/taskrunner/preloadedtaskrunner.go
```diff
@@ -54,9 +54,9 @@ func (tr *PreloadedTaskRunner) Start() {
 	}
 }
 
-// StartAndWait starts running the tasks in the task runner and waits for them to complete.
-func (tr *PreloadedTaskRunner) StartAndWait() error {
-	tr.Start()
+// Wait calls .Wait() on the underlying waitgroup, allowing for e.g.
+// cleanup of existing tasks to happen.
+func (tr *PreloadedTaskRunner) Wait() error {
```
tstirrat15, line 59: Splitting out this logic so that we can `Wait()` on a taskrunner to finish all of its goroutines (e.g. during context cancellation) before tearing down the goroutine that spawned the taskrunner

internal/cursorediterator/blocks.go
```diff
@@ -290,19 +287,23 @@ func CursoredParallelIterators[I any](
...
 			collector := collectors[collectorOffsetIndex]
-			errChannel := errChannels[collectorOffsetIndex]
 
 			for {
 				select {
 				case <-ctx.Done():
+					// Wait for taskrunner tasks to complete, so that
+					// we tear down producers to the collector channels
+					// before we tear down the consumer
+					_ = tr.Wait()
```
tstirrat15, line 297: This ensures that we never have a producer goroutine hanging on trying to produce to a channel that is no longer being consumed and which would hit its buffer limit and hang indefinitely.

ecordell (approved):

PR #2724, opened 2025-11-24
chore: fix flaky ProducerMapperIterator test

tstirrat15:
## Description
Similar refactor to #2722. This should make these tests deterministic and simplify the producer logic a bit.

## Changes
See comments

## Testing
See that tests pass. Review.

### DIFF
--- a/internal/cursorediterator/blocks.go
+++ b/internal/cursorediterator/blocks.go
@@ -291,10 +291,6 @@ func CursoredParallelIterators[I any](
 			for {
 				select {
 				case <-ctx.Done():
-					// Wait for taskrunner tasks to complete, so that
-					// we tear down producers to the collector channels
-					// before we tear down the consumer
-					_ = tr.Wait()
 					_ = yield(ItemAndCursor[I]{}, ctx.Err())
 					return
 
@@ -587,9 +583,6 @@ func CursoredProducerMapperIterator[C any, P any, I any](
 
 				select {
 				case <-ctx.Done():
-					// If we cancel, we wait for the taskrunner to complete
-					// so that there are no hanging goroutines.
-					_ = tr.Wait()
 					_ = yield(ItemAndCursor[I]{}, ctx.Err())
 					return
 
--- a/internal/cursorediterator/blocks_test.go
+++ b/internal/cursorediterator/blocks_test.go
@@ -758,6 +758,37 @@ func TestCursoredProducerMapperIterator(t *testing.T) {
 				require.Equal(t, 4, items[0].Item) // len("good")
 			})
 
+			// NOTE: this test exists because we saw a periodic flake with this test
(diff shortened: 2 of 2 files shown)


internal/cursorediterator/blocks.go
```diff
@@ -349,10 +349,11 @@ type HoldForMappingComplete[K any, C any] struct{}
 
 func (HoldForMappingComplete[K, C]) chunkOrHold() {}
 
-// producerChunk is a struct that holds a chunk of items or indicates that the chunking has completed.
+// producerChunkOrError is a struct that holds a chunk of items or an error or indicates that the chunking has completed.
 type producerChunkOrError[P any, C any, I any] struct {
 	producerCompleted bool
 
+	err error
```
tstirrat15: Putting the error in here rather than in a separate channel means the ordering is deterministic.

internal/cursorediterator/blocks.go
```diff
@@ -486,14 +486,14 @@ func CursoredProducerMapperIterator[C any, P any, I any](
 		}
 	}
 
-	producerFunc := func(ctx context.Context) {
+	producerFunc := func(ctx context.Context) error {
```
tstirrat15, line 485: Need to return the error here since we're no longer sending it to the error channel.

internal/cursorediterator/blocks.go
```diff
@@ -573,11 +570,11 @@ func CursoredProducerMapperIterator[C any, P any, I any](
...
 				return
 
-			case err := <-producerErrChan:
-				_ = yield(ItemAndCursor[I]{}, err)
-				return
-
 			case chunk := <-orderedProducerChunks:
+				if chunk.err != nil {
+					_ = yield(ItemAndCursor[I]{}, chunk.err)
+					return
```
tstirrat15, line 572: This is where we now handle the error that comes out of the iterator.

internal/cursorediterator/blocks.go
```diff
@@ -590,6 +587,9 @@ func CursoredProducerMapperIterator[C any, P any, I any](
 
 				select {
 				case <-ctx.Done():
+					// If we cancel, we wait for the taskrunner to complete
+					// so that there are no hanging goroutines.
+					_ = tr.Wait()
```
tstirrat15: This ensures that if the context is canceled, the producing goroutines are torn down before the consuming goroutine, which helps ensure that we don't get a goroutine leak.

tstirrat15 (commented):
See comments

internal/cursorediterator/blocks.go
josephschorr: I'd rather not wait here; the goroutines will eventually shut down but we don't want to block the caller

josephschorr (dismissed):
LGTM

josephschorr (approved):

PR #2727, opened 2025-11-24
Use action-conventional-commits instead of bespoke logic

tstirrat15:
## Description
Trying out https://github.com/webiny/action-conventional-commits, which should be an easier thing to replicate across our other repositories.

## Changes
* Replace bespoke logic with existing action

## Testing
Review.


tstirrat15:
I validated that it failed with a bad commit message.

PR #2729, opened 2025-11-25
fix: use FAILED_PRECONDITION for recursion depth errors

### LINKED ISSUE #2728: Change recursion depth error to `FAILED_PRECONDITION`
From the [GRPC docs](https://grpc.io/docs/guides/status-codes/#full-list-of-status-codes):

> The operation was rejected because the system is not in a state required for the operation’s execution. For example, the directory to be deleted is non-empty, an rmdir operation is applied to a non-directory, etc. Service implementors can use the following guidelines to decide between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher level (e.g., when a client-specified test-and-set fails, indicating the client should restart a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an “rmdir” fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless the files are deleted from the directory.

We currently return a `RESOURCE_EXHAUSTED`, but we want to reserve `RESOURCE_EXHAUSTED` for operations that can safely be retried. A recursion depth error indicates a data cycle that requires the client to fix the cycle before the operation can be safely retried.

It seems as though `FAILED_PRECONDITION` is a better fit here.

tstirrat15:
Fixes #2728 

## Description
See #2728. We want to reserve `RESOURCE_EXHAUSTED` for things that can be safely retried, and a recursion depth error cannot generally be safely retried. The docs indicate that `FAILED_PRECONDITION` is more appropriate here.

## Changes
* Change recursion depth error to have a `ResourceExhausted` code
* Update tests
## Testing
Review.

### DIFF
--- a/internal/services/shared/errors_test.go
+++ b/internal/services/shared/errors_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/internal/graph"
 	"github.com/authzed/spicedb/internal/namespace"
+	"github.com/authzed/spicedb/internal/sharederrors"
 	"github.com/authzed/spicedb/pkg/cursor"
 	"github.com/authzed/spicedb/pkg/datastore"
 	corev1 "github.com/authzed/spicedb/pkg/proto/core/v1"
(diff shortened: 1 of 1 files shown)


tstirrat15:
I've added an issue to that repo.

internal/services/shared/errors_test.go
```diff
@@ -77,7 +77,7 @@ func TestRewriteError(t *testing.T) {
 			config: &ConfigForErrors{
 				MaximumAPIDepth: 50,
 			},
-			expectedCode:     codes.ResourceExhausted,
+			expectedCode:     codes.FailedPrecondition,
 			expectedContains: "max depth exceeded: this usually indicates a recursive or too deep data dependency. See https://spicedb.dev/d/debug-max-depth",
```
miparnisari:
```suggestion
			expectedContains: "max depth exceeded: this usually indicates a recursive or too deep data dependency. See " + sharederrors.MaxDepthErrorLink,
```

miparnisari (dismissed):

miparnisari (approved):

PR #2732, opened 2025-11-26
fix(generator): do not generate duplicated relations/perms

vroldanbet:
## Description

The generator was drawing from rapid's pool without uniqueness guarantee. This led to invalid schemas.

## Testing

Updated the test to call `Validate`


josephschorr (approved):
LGTM

PR #2733, opened 2025-11-26
feat: allow underscore prefix for private identifiers

### LINKED ISSUE #2066: Establish a convention for public/private
### Problem Statement

I often want to signal to my fellow developers that a particular permission is "private". This is very common with "synthetic permissions" that exist only to further nest arrows.

This problem will only get exacerbated once there is a module system for composing schemas via library.

Ideally there is an officially sanctioned syntax for representing a private definition.

### Solution Brainstorm

Enable identifiers to be able to begin with an underscore

ivanauth:
## Summary
- Enable underscore prefix (`_`) for definition, relation, and permission identifiers
- Establishes a convention for marking identifiers as "private" or "internal"

## Description
This PR implements the feature requested in issue #2066 by updating the identifier regex patterns from `[a-z]` to `[a-z_]` to allow identifiers to begin with an underscore.

This is useful for:
- **Synthetic permissions**: Permissions that exist only to compose other permissions
- **Internal relations**: Relations not meant to be directly referenced by application code
- **Implementation details**: Parts of your schema that may change without affecting the public API

## Changes
- Updated regex patterns in proto validation to allow underscore-prefixed identifiers
- Updated corresponding test expectations

## Example Usage
```zed
definition document {
    relation viewer: user
    relation _internal_viewer: user
    
    // Private synthetic permission
    permission _can_view = viewer + _internal_viewer
    
    // Public permission
    permission view = _can_view
}
```

## Note
End-user documentation for this feature should be added to [authzed/docs](https://github.com/authzed/docs) at `pages/spicedb/concepts/schema.mdx`.

Fixes #2066

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [1.56.0] - 2026-07-24
 ### Added
 - New metric `check_permissionship_total` for CheckPermission and CheckBulkPermissions that counts the number of requests that returned HAS_PERMISSION. Also, `write_relationships_updates` also includes BulkImport calls (https://github.com/authzed/spicedb/pull/3240)
+- Allow underscore prefix for private identifiers in definitions, relations, and permissions (https://github.com/authzed/spicedb/pull/2733)
 
 ### Changed
 - Schema: reads inside write transactions now use a cheap hash-only lookup (`schema_revision`) to check the cache before loading the full schema blob, reducing DB round-trips on cache hits (https://github.com/authzed/spicedb/pull/3160)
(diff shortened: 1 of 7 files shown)


docs/schema-language.md
miparnisari, line 1: Why are we writing this doc here? This will not get published anywhere. Maybe we should update https://authzed.com/docs/spicedb/concepts/schema instead?

ivanauth: Thank you, i removed docs from this pr and created a new docs pr.

josephschorr:
I think if we introduce this, we should *enforce* that private identifiers cannot be accessed outside the scope of either the definition *or* the file (in composable schemas) and perhaps even enforce that they cannot be used in API calls

ivanauth:
@josephschorr good point - do you want me to try adding definition-scoped + API rejection by default?

ivanauth:
I think we need to revert https://github.com/authzed/docs/pull/421 until this PR is merged.

Mr-Pepe:
> I think we need to revert https://github.com/authzed/docs/pull/421 until this PR is merged.

Or merge this 😋 

The discrepancy is indeed a bit confusing.

1993alexey:
I was a bit confused by the updated documentation without code support, as I kept getting schema validation errors. Agree on rolling back the docs change until this MR is merged

PR #2734, opened 2025-11-27
fix(pkg): isolates dependency with github.com/odigos-io/go-rtml

vroldanbet:
## Description

This PR isolates dependency with `github.com/odigos-io/go-rtml`

The introduction of `github.com/odigos-io/go-rtml` as a dependency has caused many go toolchain operations to fail because it now requires special flags to resolve a runtime symbol. This is not ideal because it breaks the developer experience for folks using IDEs with native Go support. Something as simple as `go test ./...` no longer works without flags. But most importantly, it **becomes sticky** for any go module that imports SpiceDB.

This commit isolates the memory protection usage by:
- adding a global variable that acts as the registry
- isolating the `MemoryUsageProvider` into its own package
- introducing a new `NoopMemoryLimiter`, and using it everywhere we explicitly used the go-rtml package
- moving tests into isolated packages
- initializing the global default provider in `main.go`, using a function controlled by build tags. Without the `memoryprotection` build tag, it defaults to using the noop implementation.
- moving all tests in the main package to make sure one can run them without the need of passing the special flags.
  
Consequence:
- now all tests run without special flags, except for those where we explicitly test for this behavior.
- go modules importing SpiceDB do won't require the flags.
- Folks can run SpiceDB in their IDEs quickly without tinkering with toolchain flags

## Testing

- Run `main.go` without tag should show a warning indicating it fell back to noop provider
- Run `main.go` with `ldflags` and `memoryprotection` tag, no warn log is printed, and log shows the type used
- run all mage commands that tounch something that uses ldflags and the tags
- build docker container and check logs, making sure the rtml limiter is in place


internal/middleware/memoryprotection/memory_protection.go
```diff
@@ -34,6 +35,7 @@ func New(usageProvider MemoryUsageProvider, name string) *MemoryProtectionMiddle
 
 	log.Info().
 		Str("name", name).
+		Str("provider", reflect.TypeOf(usageProvider).String()).
```
josephschorr, line 38: Maybe add a Name() to the interface?

vroldanbet: Seems overkill

pkg/cmd/server/server.go
```diff
@@ -576,9 +578,14 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 	}, nil
 }
 
-func (c *Config) buildMemoryUsageProvider() memoryprotection.MemoryUsageProvider {
+func (c *Config) MustBuildMemoryUsageProvider() memoryprotection.MemoryUsageProvider {
```
tstirrat15, line 581: Why is this a `Must`? Don't we usually use the `Must` prefix to indicate that the function will panic if it can't do the thing that it's supposed to do?

vroldanbet: This is an oversight. At the beginning, I was panicking when `DefaultMemoryUsageProvider` was `nil`, but then I found that it would force me to configure a lot of tests with the specific provider, which I found annoying, so instead I added a fallback. I can open a quick follow-up PR

vroldanbet: opened https://github.com/authzed/spicedb/pull/2735

PR #2739, opened 2025-12-02
ci: fix docs workflow

miparnisari:


.github/workflows/docs.yaml
```diff
@@ -50,7 +50,8 @@ jobs:
           path: "docs-repo"
           title: "Auto-generated PR: Update spicedb docs"
           body: "This PR was auto-generated by GitHub Actions."
-          branch: "auto-update-branch"
+          branch: "docs-spicedb-update"
```
miparnisari, line 53: so as not to conflict with other automated PRs

.github/workflows/docs.yaml
```diff
@@ -79,5 +80,6 @@ jobs:
                 text:
                   type: "mrkdwn"
                   text: |
+                    :x: @eng-oss Could not sync docs from spicedb repo to the docs repo. Please take a look.
```
miparnisari, line 83: so that the slack notification has _some_ text 😅

squat (approved):

PR #2741, opened 2025-12-02
chore: reduce t.Parallel() fanout

barakmich:
## Description

Reorders and removes some overly-aggressive uses of `t.Parallel()` to improve CI

## Testing

Run `go run mage.go test:integrationCover` like CI
Local timing before change: 419.87s
Local timing after change: 437.459s

## References
https://github.com/authzed/spicedb/actions/runs/19870343724/job/56944204925

### DIFF
--- a/internal/services/integrationtesting/query_plan_consistency_test.go
+++ b/internal/services/integrationtesting/query_plan_consistency_test.go
@@ -22,7 +22,7 @@ import (
 	"github.com/authzed/spicedb/pkg/validationfile/blocks"
 )
 
-func TestQueryPlanConsistency(t *testing.T) {
+func TestQueryPlanConsistency(t *testing.T) { // nolint:tparallel
 	t.Parallel()
 	consistencyTestFiles, err := consistencytestutil.ListTestConfigs()
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


internal/services/integrationtesting/query_plan_consistency_test.go
```diff
@@ -30,7 +30,7 @@ func TestQueryPlanConsistency(t *testing.T) {
 		filePath := filePath
 
 		t.Run(filepath.Base(filePath), func(t *testing.T) {
-			t.Parallel()
+			//	t.Parallel()
```
tstirrat15: Is this something we would re-enable? if not, I'd rather remove it, and if it's about being explicit about not putting it here, i'd rather put a comment that says so.

miparnisari: the linter is failing anyway. add a comment to it explaining why it got removed referencing this PR

barakmich:
> For my own edification could you explain how you arrived to the conclusion that the segfault was caused by having too many tests running at once?

Yeah; so the crash was from the Go runtime segfaulting while trying to grow a stack. It's technically a bug in Go -- but it would happen on CI under memory pressure. Go will grow the stacks as necessary but also needs to keep a bunch of partial stacks around for defers and closures (like subtests) as well. Combine with the fact that there were an absurd number of goroutines present in the crash, and the culprits to trip it were in fact the consistency tests -- which created a goroutine per assertion per optimization mode per file -- and so they're the ones most likely to build a bunch of stacks. 

So we can take a small hit in the parallelism and just do some of them linearly (like, one per file) and reduce the pressure.

PR #2747, opened 2025-12-05
feat: add a proxy for counting datastore accesses and expose them through prometheus

barakmich:

### DIFF
--- a/internal/datastore/proxy/counting.go
+++ b/internal/datastore/proxy/counting.go
@@ -119,6 +119,30 @@ func NewCountingDatastoreProxy(d datastore.ReadOnlyDatastore) (datastore.ReadOnl
 	}, counts
 }
 
+// NewCountingDatastoreProxyForDatastore wraps a full Datastore with counting proxy.
+// Like NewCountingDatastoreProxy, but returns a full Datastore interface for use with middleware.
+// The ReadWriteTx method is passed through without counting (only Reader methods are counted).
+func NewCountingDatastoreProxyForDatastore(d datastore.Datastore) (datastore.Datastore, *DatastoreReaderMethodCounts) {
+	counts := &DatastoreReaderMethodCounts{}
+	return &countingDatastoreProxy{
+		countingProxy: countingProxy{
+			delegate: d,
+			counts:   counts,
+		},
+		fullDatastore: d,
+	}, counts
+}
+
+type countingDatastoreProxy struct {
+	countingProxy
+	fullDatastore datastore.Datastore
+}
+
+// ReadWriteTx delegates to the underlying full datastore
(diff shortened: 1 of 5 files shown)


internal/datastore/proxy/counting.go
```diff
@@ -0,0 +1,270 @@
...
+	countingLookupNamespacesWithNamesTotal prometheus.Counter
+)
+
+func init() {
+	countingQueryRelationshipsTotal = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: "spicedb",
+		Subsystem: "datastore",
+		Name:      "query_relationships_total",
+		Help:      "total number of QueryRelationships calls",
+	})
```
tstirrat15, line 29: Nice. I can see these being useful metrics in their own right outside of this use case.

internal/datastore/proxy/counting.go
```diff
@@ -0,0 +1,270 @@
...
+	return m.lookupNamespacesWithNames.Load()
+}
+
+// WriteMethodCounts writes the counts to Prometheus metrics as additive counters.
+// The DatastoreReaderMethodCounts struct is ephemeral (typically per-request),
+// and this function adds its counts to the global Prometheus counters.
+// Multiple counts can be combined by calling this function on each.
+// Tests can skip calling this and just access the DatastoreReaderMethodCounts directly.
+func WriteMethodCounts(counts *DatastoreReaderMethodCounts) {
+	countingQueryRelationshipsTotal.Add(float64(counts.QueryRelationships()))
```
tstirrat15, line 103: Them being floats surprises me - is that just a prom thing?

internal/datastore/proxy/counting.go
```diff
@@ -0,0 +1,270 @@
...
+	counts := &DatastoreReaderMethodCounts{}
+	return &countingProxy{
+		delegate: d,
+		counts:   counts,
+	}, counts
+}
+
+// NewCountingDatastoreProxyForDatastore wraps a full Datastore with counting proxy.
+// Like NewCountingDatastoreProxy, but returns a full Datastore interface for use with middleware.
+// The ReadWriteTx method is passed through without counting (only Reader methods are counted).
```
tstirrat15, line 124: Will we potentially miss reads here? Or is this really about read API paths so we don't care about them?

internal/datastore/proxy/counting.go
barakmich: It is a prom thing!

internal/datastore/proxy/counting.go
barakmich: We might miss reads in read/write txns, fair, but yep, this is about the read API paths (and if we want read/write ones, simple addition to this code)

PR #2750, opened 2025-12-06
test: fix flaky TestExecuteRepair

miparnisari:
Should prevent this

```
--- FAIL: TestExecuteRepair (0.00s)
    --- FAIL: TestExecuteRepair/cockroachdb_does_not_support_repair (13.51s)
        datastore_test.go:89: 
            	Error Trace:	/home/runner/work/spicedb/spicedb/pkg/cmd/datastore_test.go:89
            	Error:      	Error "failed to create datastore: duplicate metrics collector registration attempted" does not contain "datastore of type 'cockroachdb' does not support the repair operation"
            	Test:       	TestExecuteRepair/cockroachdb_does_not_support_repair
FAIL
coverage: 3.8% of statements in ./...

```

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -465,7 +465,10 @@ func (cds *crdbDatastore) Close() error {
 	cds.readPool.Close()
 	cds.writePool.Close()
 	for _, collector := range cds.collectors {
-		_ = prometheus.Unregister(collector)
+		ok := prometheus.Unregister(collector)
+		if !ok {
+			return errors.New("could not unregister collector for CRDB datastore")
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


internal/datastore/crdb/crdb.go
```diff
@@ -644,7 +647,7 @@ func (cds *crdbDatastore) registerPrometheusCollectors(enablePrometheusStats boo
 	})
 
 	if err := prometheus.Register(readCollector); err != nil {
-		return err
+		return errors.New("failed to register prometheus read collector")
```
tstirrat15: Should we wrap the old error here with `%w`?

tstirrat15: (also the other spots)

PR #2754, opened 2025-12-09
test: use synctest in some CRDB pool retry tests

### LINKED ISSUE #2713: Rewrite CRDB pool retry tests to use synctest
https://github.com/authzed/spicedb/pull/2576/files#diff-1d476df7947c1d448571ef7d5f104475648c4faee1c799821c19d8af17cd716f

These are a good example of tests that benefit from the use of `test/synctest`.

ivanauth:
## Summary

Rewrites the CRDB pool retry tests to use Go's `testing/synctest` package for better time simulation and test reliability.

Fixes #2713

## Changes

- Wrap all retry tests in `synctest.Test(t, func(t *testing.T) { ... })`
- Use `t.Context()` instead of `context.Background()`
- Add `t.Parallel()` for concurrent test execution

## Benefits

- Tests no longer rely on real time delays
- More deterministic test execution
- Faster test runs with simulated time
- Better parallel execution support

### DIFF
--- a/internal/datastore/crdb/pool/pool_test.go
+++ b/internal/datastore/crdb/pool/pool_test.go
@@ -139,44 +139,40 @@ func TestAcquireTimeoutReturnsErrAcquire(t *testing.T) {
 
 func TestAcquireSucceedsButTopLevelContextCancelled(t *testing.T) {
 	t.Parallel()
-	synctest.Test(t, func(t *testing.T) {
-		testPool := NewTestPool()
+	testPool := NewTestPool()
 
-		retryPool := createTestRetryPool(testPool)
-		ctx, cancel := context.WithCancel(t.Context())
-		cancel() // Cancel immediately
+	retryPool := createTestRetryPool(testPool)
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // Cancel immediately
 
-		err := retryPool.withRetries(ctx, 0, func(conn *pgxpool.Conn) error {
-			t.Fatal("function should not be called when context is cancelled")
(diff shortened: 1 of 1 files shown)


internal/datastore/crdb/pool/pool_test.go
```diff
@@ -79,204 +80,231 @@ func createTestRetryPool(testPool *TestPool) *RetryPool {
...
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "error acquiring connection from pool")
+		require.ErrorIs(t, errors.Unwrap(err), ErrAcquire)
+	})
 }
 
 func TestAcquireSucceedsButTopLevelContextCancelled(t *testing.T) {
-	testPool := NewTestPool()
+	t.Parallel()
+	synctest.Test(t, func(t *testing.T) {
```
miparnisari: why do we need synctest here? there is no synchronization involved

a good rule of thumb i am just discovering is: if you checkout `main` and change all the `time.Milliseconds` to `time.Seconds`, what tests slow down considerably? those are tests that we want to use `synctest` on. 

following that: i'd only put `synctest` on
- TestContextCancelledDuringBlockingAcquire
- TestAcquireTimeoutReturnsErrAcquire
- TestAcquireSucceedsWithinTimeout
- TestAcquireTimeoutContextCausePreserved

the rest don't need it

ivanauth: Thanks! I removed unnecessary pieces.

miparnisari (approved):

PR #2757, opened 2025-12-09
fix: goroutine leaks

miparnisari:
For reviewers: please see changes in
- `pkg/testutil/leakdetection.go`
- `internal/dispatch/combined/combined.go`

### DIFF
--- a/internal/datastore/proxy/schemacaching/watchingcache.go
+++ b/internal/datastore/proxy/schemacaching/watchingcache.go
@@ -136,8 +136,9 @@ func (p *watchingCachingProxy) ReadWriteTx(
 	return p.fallbackCache.ReadWriteTx(ctx, f, opts...)
 }
 
+// Start is async so that prepopulating doesn't block the server start.
+// The caller must cancel the context before calling Close.
 func (p *watchingCachingProxy) Start(ctx context.Context) error {
-	// Start async so that prepopulating doesn't block the server start.
 	go func() {
 		_ = p.startSync(ctx)
 	}()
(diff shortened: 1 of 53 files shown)


internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -89,8 +89,7 @@ func createWatchingCacheProxy(delegate datastore.Datastore, c cache.Cache[cache.
 
 		gcWindow:       gcWindow,
 		watchHeartbeat: watchHeartbeat,
-		closed:         make(chan bool, 2),
```
miparnisari: FYI making this channel buffered means that the sender will _not_ block. So `Close()` isn't waiting until the receiver receive and the goroutines close. 

But, funnily enough, removing the buffer makes the tests never finish. So i'm replacing with a `waitgroup`.

internal/dispatch/combined/combined.go
```diff
@@ -196,51 +196,51 @@ func NewDispatcher(options ...Option) (dispatch.Dispatcher, error) {
...
 
-	chunkSize := opts.dispatchChunkSize
-	if chunkSize == 0 {
-		chunkSize = 100
-		log.Warn().Msgf("CombinedDispatcher: dispatchChunkSize not set, defaulting to %d", chunkSize)
-	}
+	var redispatch dispatch.Dispatcher
 
-	cts := caveattypes.TypeSetOrDefault(opts.caveatTypeSet)
+	if opts.upstreamAddr == "" {
```
miparnisari, line 201: FYI major change on the setup. Notice how before, `redispatch`'s content was getting overwritten on line 307

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -136,8 +136,9 @@ func (p *watchingCachingProxy) ReadWriteTx(
 	return p.fallbackCache.ReadWriteTx(ctx, f, opts...)
 }
 
+// Start is async so that prepopulating doesn't block the server start.
+// The caller must cancel the context before calling Close.
```
tstirrat15, line 140: Oof. This seems like a footgun.

PR #2761, opened 2025-12-11
feat: introduce support for schema PostOrder traversal

vroldanbet:
## Summary

Introduces comprehensive post-order traversal support for the schema walker, including automatic arrow target traversal and parent tracking. This removes significant burden from client-side code by automatically traversing arrow references and providing parent context during schema walks.

### 1. PostOrder Traversal Strategy
Add support for both PreOrder and PostOrder traversal strategies to the schema walker, while maintaining backward compatibility with existing code.

**PreOrder (current behavior):**
1. Visit node first (call visitor method)
2. If continue flag is true, visit children
3. Return accumulated value

**PostOrder (new behavior):**
1. Visit children first (unconditionally)
2. Then visit node (call visitor method)
3. Return accumulated value

### 2. Parent Concept in Schema V2
Introduces parent tracking throughout the schema hierarchy, allowing visitors to access parent context during traversal. This enables more sophisticated schema analysis and transformation patterns.

### 3. Arrow Traversal in Post-Order
Automatically traverses arrow targets during post-order walks, eliminating the need for clients to manually follow arrow references. The options system was refactored to support stateful traversal, with a new fluent factory pattern for constructing walker options.

## Design Decision: Options Pattern with WithOptions Functions

**Selected Approach:** Create new `*WithOptions` variants for all Walk functions that accept a `WalkOptions` struct, following the pattern already established in `pkg/schema/v2/flatten.go`.

**Rationale:**
 - Backward Compatible: Existing callers don't need any changes
 - Follows Codebase Pattern: Matches FlattenSchema/FlattenSchemaWithOptions pattern in flatten.go
 - Explicit Control: New code can opt into different strategies using WithOptions variants
 - Extensible: Options struct can be extended with future traversal options without breaking APIs
 - Type Safety: No confusing variadic parameters with generics
 - Clean Separation: Strategy is configuration, not part of visitor logic

## Important Notes

- In PostOrder, children are visited unconditionally (the continue flag from the visitor is checked after children are already visited)
- Terminal visitors (those returning (T, error)) work the same in both orders since they have no children
- Options must be threaded through all recursive calls
- For terminal nodes (Caveat, BaseRelation), the internal implementation can be simple passthrough since there's no difference between orders

## Operation Tree Traversal

The `walkOperationWithOptions` function is the most complex due to its type-switch structure:

- For compound operations (Union, Intersection, Exclusion): visit children based on strategy order
- For terminal operations (RelationReference, ArrowReference, etc.): strategy doesn't affect behavior
- Must handle both generic OperationVisitor[T] and specific visitors appropriately
- Double-dispatch for arrow operations: call both ArrowOperationVisitor and specific arrow visitor

## Continue Flag Semantics

Continue flag behavior differs between strategies:
- PreOrder: cont=false means "skip visiting children"
- PostOrder: Children are already visited before checking the flag, so cont=false has limited effect

### DIFF
--- a/pkg/query/build_tree.go
+++ b/pkg/query/build_tree.go
@@ -163,7 +163,7 @@ func (b *iteratorBuilder) buildIteratorFromPermission(p *schema.Permission) (Ite
 }
 
 func (b *iteratorBuilder) buildIteratorFromOperation(p *schema.Permission, op schema.Operation) (Iterator, error) {
-	parentDef := p.TypedParent()
+	parentDef := p.Definition()
 
 	switch perm := op.(type) {
 	case *schema.ArrowReference:
--- a/pkg/schema/v2/schema.go
+++ b/pkg/schema/v2/schema.go
@@ -292,8 +292,8 @@ func (p *Permission) Parent() Parented {
 	return p.parent
 }
 
-// TypedParent returns the parent definition with its concrete type.
-func (p *Permission) TypedParent() *Definition {
+// Definition returns the parent definition with its concrete type.
+func (p *Permission) Definition() *Definition {
 	return p.parent
 }
 
@@ -369,8 +369,8 @@ func (r *Relation) Parent() Parented {
 	return r.parent
 }
 
-// TypedParent returns the parent definition with its concrete type.
-func (r *Relation) TypedParent() *Definition {
+// Definition returns the parent definition with its concrete type.
+func (r *Relation) Definition() *Definition {
 	return r.parent
 }
 
--- a/pkg/schema/v2/walk_arrow_test.go
+++ b/pkg/schema/v2/walk_arrow_test.go
@@ -16,13 +16,13 @@ type arrowTraversalVisitor struct {
 }
 
 func (v *arrowTraversalVisitor) VisitPermission(p *Permission, value struct{}) (struct{}, bool, error) {
-	defName := p.TypedParent().Name()
+	defName := p.Definition().Name()
 	v.visitOrder = append(v.visitOrder, defName+"#"+p.Name())
 	return value, true, nil
 }
 
 func (v *arrowTraversalVisitor) VisitRelation(r *Relation, value struct{}) (struct{}, bool, error) {
-	defName := r.TypedParent().Name()
+	defName := r.Definition().Name()
 	v.visitOrder = append(v.visitOrder, defName+"#"+r.Name())
 	return value, true, nil
 }


pkg/query/build_tree.go
```diff
@@ -162,22 +162,24 @@ func (b *iteratorBuilder) buildIteratorFromPermission(p *schema.Permission) (Ite
 }
 
 func (b *iteratorBuilder) buildIteratorFromOperation(p *schema.Permission, op schema.Operation) (Iterator, error) {
+	parentDef := p.Parent().(*schema.Definition)
```
tstirrat15: Is this a safe cast? Should it be `ok`ing and then returning an error if not?

pkg/schema/v2/operations.go
```diff
@@ -134,10 +228,32 @@ func (u *UnionOperation) clone() Operation {
...
+func (u *UnionOperation) cloneWithParent(parent Operation) Operation {
+	if u == nil {
+		return nil
+	}
+	cloned := &UnionOperation{
+		parent:   parent,
+		children: make([]Operation, len(u.children)),
+	}
+	for i, child := range u.children {
+		if childCloner, ok := child.(interface{ cloneWithParent(Operation) Operation }); ok {
```
tstirrat15: Why is this interface defined inline vs being defined statically?

pkg/schema/v2/operations.go
```diff
@@ -134,10 +228,32 @@ func (u *UnionOperation) clone() Operation {
...
+			cloned.children[i] = childCloner.cloneWithParent(cloned)
```
tstirrat15: Is this because not every possible schema entity implements `cloneWithParent`?

pkg/schema/v2/operations.go
```diff
@@ -363,6 +627,38 @@ func (f *FunctionedArrowReference) clone() Operation {
...
+		left:     f.left,
+		right:    f.right,
+		function: f.function,
+	}
+}
+
+// setParent sets the parent pointer for a Parented element.
+// The parent parameter can be either an Operation (for child nodes) or a *Permission (for the root node).
+// This is a simple setter that delegates to the element's setParent method.
+func setParent(elem Parented, parent Parented) {
```
tstirrat15: Why this pattern rather than invoking the setter directly? Is it just about nil handling?

pkg/schema/v2/schema.go
```diff
@@ -74,10 +126,16 @@ type Definition struct {
 }
 
 // Parent returns the parent schema.
-func (d *Definition) Parent() *Schema {
+func (d *Definition) Parent() Parented {
 	return d.parent
 }
 
+func (d *Definition) setParent(p Parented) {
+	if s, ok := p.(*Schema); ok {
```
tstirrat15, line 134: Worth adding `MustBugf` for the not-`ok` case?

pkg/schema/v2/schema.go
```diff
@@ -244,11 +319,26 @@ func (p *Permission) cloneWithParent(parentDefinition *Definition) *Permission {
 		return nil
 	}
 
-	return &Permission{
+	var clonedOp Operation
+	if p.operation != nil {
+		if opCloner, ok := p.operation.(interface{ cloneWithParent(Operation) Operation }); ok {
```
tstirrat15: Same deal here - should this use a concrete interface? When would we expect the interface to not be satisfied?

pkg/schema/v2/schema_test.go
```diff
@@ -212,3 +212,152 @@ func TestGetPermission(t *testing.T) {
...
+
+	// Test finding Definition parent
+	definition := FindParent[*Definition](firstChild)
+	require.NotNil(t, definition, "should find definition parent")
+	require.Equal(t, "document", definition.Name(), "should find the correct definition")
+
+	// Test finding Schema parent
+	schema := FindParent[*Schema](firstChild)
+	require.NotNil(t, schema, "should find schema parent")
+	require.Equal(t, s, schema, "should find the same schema instance")
```
tstirrat15, line 250: Is this testing that the pointers are the same or that the objects are structurally equivalent?

pkg/schema/v2/schema_test.go
```diff
@@ -212,3 +212,152 @@ func TestGetPermission(t *testing.T) {
...
+		RelationRef("owner").
+		Done().
+		Done().
+		Build()
+
+	def, _ := s.GetTypeDefinition("document")
+	viewPerm, _ := def.GetPermission("view")
+	op := viewPerm.Operation()
+
+	// The immediate parent of the operation should be the permission
```
tstirrat15: There isn't another permission for it to end up at - what is this test asserting?

pkg/schema/v2/walk.go
```diff
@@ -1,5 +1,78 @@
...
+func (opts WalkOptions) WithStrategy(strategy WalkStrategy) WalkOptions {
+	opts.strategy = strategy
+	return opts
+}
+
+// WithTraverseArrowTargets enables automatic traversal of arrow reference targets.
+//
+// Requirements:
+//   - PostOrder strategy (set via WithStrategy)
+//   - Resolved schema (call ResolveSchema() before walking)
```
tstirrat15: What is a "resolved" schema?

pkg/schema/v2/walk.go
```diff
@@ -245,45 +453,396 @@ func WalkBaseRelation[T any](br *BaseRelation, v Visitor[T], value T) (T, error)
...
+	if options.traverseArrowTargets && options.strategy != WalkPostOrder {
+		return value, errors.New("TraverseArrowTargets requires PostOrder strategy")
+	}
+
+	// If arrow traversal is enabled, ensure visited set is available.
+	// This is the central initialization point for all operation walks to avoid duplication.
+	if options.traverseArrowTargets {
+		if options.visitedArrowTargets == nil {
+			options.visitedArrowTargets = make(map[string]bool)
+		}
```
tstirrat15: Is there a reason this is lazily constructed rather than being constructed in the option setter that sets `traverseArrowTargets`?

pkg/schema/v2/walk.go
```diff
@@ -245,45 +453,396 @@ func WalkBaseRelation[T any](br *BaseRelation, v Visitor[T], value T) (T, error)
...
+	}
+
+	if options.strategy == WalkPostOrder {
+		return walkOperationPostOrder(op, v, value, options)
+	}
+	return walkOperationPreOrder(op, v, value, options)
+}
+
+// walkOperationPostOrder handles PostOrder traversal of operation trees.
+func walkOperationPostOrder[T any](op Operation, v Visitor[T], value T, options WalkOptions) (T, error) {
```
tstirrat15, line 565: 👍

pkg/schema/v2/walk.go
```diff
@@ -245,45 +453,396 @@ func WalkBaseRelation[T any](br *BaseRelation, v Visitor[T], value T) (T, error)
...
+			if err != nil {
+				return currentValue, err
+			}
+			currentValue = newValue
+		}
+
+	case *ArrowReference:
+		// If arrow traversal is enabled but schema is not resolved, provide helpful error
+		if options.traverseArrowTargets {
+			return currentValue, errors.New(
```
tstirrat15, line 593: 👍

pkg/schema/v2/walk.go
```diff
@@ -245,45 +453,396 @@ func WalkBaseRelation[T any](br *BaseRelation, v Visitor[T], value T) (T, error)
...
+			newValue, err := rrrv.VisitResolvedRelationReference(o, currentValue)
+			if err != nil {
+				return currentValue, err
+			}
+			currentValue = newValue
+		}
+
+	case *ResolvedArrowReference:
+		// If arrow traversal is enabled, walk target permissions first (PostOrder)
+		if options.traverseArrowTargets {
```
tstirrat15, line 673: Is the idea here that previously you'd just have a marker saying "this is a reference to this permission on this other definition," but now we have a handle on the actual permission itself, which lets us walk through?

pkg/schema/v2/walk.go
```diff
@@ -245,45 +453,396 @@ func WalkBaseRelation[T any](br *BaseRelation, v Visitor[T], value T) (T, error)
...
+			resolvedLeft := o.ResolvedLeft()
+			if resolvedLeft != nil {
+				// For each base relation (allowed subject type), walk the target permission
+				for _, br := range resolvedLeft.BaseRelations() {
+					targetDefName := br.Type()
+					targetPermName := o.Right()
+					targetKey := targetDefName + "#" + targetPermName
+
+					// Skip if already visited to prevent infinite recursion
+					if options.visitedArrowTargets[targetKey] {
```
tstirrat15: Is this registry local to walking a particular subset of the tree, or is it global? Is it a problem if different arrow instances reference the same relations in different places in the schema?

pkg/schema/v2/walk_arrow_test.go
```diff
@@ -0,0 +1,693 @@
...
+
+	// With arrow traversal enabled, folder#view should be visited BEFORE document's parent->view arrow
+	// because the arrow targets folder#view
+	t.Logf("Visit order: %v", visitor.visitOrder)
+
+	firstFolderViewIndex := -1
+	arrowRefIndex := -1
+
+	// Find the FIRST occurrence of folder#view (via arrow traversal) and the arrow itself
+	for i, visit := range visitor.visitOrder {
```
tstirrat15: https://pkg.go.dev/slices#Index?

pkg/schema/v2/walk_arrow_test.go
```diff
@@ -0,0 +1,693 @@
...
+		if visit == "folder#view" && firstFolderViewIndex == -1 {
+			firstFolderViewIndex = i
+		}
+		if visit == "document#parent->view" {
+			arrowRefIndex = i
+		}
+	}
+
+	require.NotEqual(t, -1, firstFolderViewIndex, "folder#view should be visited")
+	require.NotEqual(t, -1, arrowRefIndex, "document arrow should be visited, but got visits: %v", visitor.visitOrder)
```
tstirrat15: `require.Contains`?

pkg/schema/v2/parented_example_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// Build a simple schema
+	s := schema.NewSchemaBuilder().
+		AddDefinition("document").
+		AddRelation("owner").
+		AllowedDirectRelation("user").
+		Done().
+		AddPermission("view").
+		Union(
+			schema.RelRef("owner"),
+			schema.ArrowRef("parent", "view"),
```
miparnisari, line 20: where is `parent` defined? i guess it doesn't matter for the purposes of this test but it's weird that `Build()` didn't throw an error 🤔

pkg/schema/v2/walk_arrow_test.go
```diff
@@ -0,0 +1,693 @@
...
+			WithTraverseArrowTargets(resolved.Schema()),
+	)
+	require.NoError(t, err)
+
+	// With arrow traversal enabled, folder#view should be visited BEFORE document's parent->view arrow
+	// because the arrow targets folder#view
+	t.Logf("Visit order: %v", visitor.visitOrder)
+
+	firstFolderViewIndex := -1
+	arrowRefIndex := -1
```
tstirrat15: Same comment

pkg/schema/v2/parented_example_test.go
tstirrat15: It doesn't need to be - that's not actually enforced by the structure of the schema :P It will be caught by subsequent typechecking, such as on validation or in the WriteSchema API, but it's not present at this level.

pkg/schema/v2/operations.go
```diff
@@ -3,6 +3,7 @@ package schema
 // Operation is a closed enum of things that can exist on the right-hand-side of a permission.
 // It forms a tree of unions, intersections and exclusions, until the leaves are things like references to other permissions or relations, or are arrows.
 type Operation interface {
+	Parented
 	isOperation()
 	clone() Operation
```
miparnisari, line 8: if `Operations` satisfy `Parented`, why is `cloneWithParent(parent Operation)` not part of this interface?

also: why does `SelfReference` not implement `cloneWithParent`?

also: why do we have `clone()` in `Operation` AND in `schemaUnit`?

pkg/schema/v2/operations.go
miparnisari: how can this be? the children of a union will always have a parent. You can even see line 244 is never executed in tests; i feel line 244 should panic or something.

pkg/schema/v2/walk_test.go
```diff
@@ -1106,3 +1107,690 @@ func TestNilReferenceVisitor_ErrorHandling(t *testing.T) {
...
+	editorRel.parent = docDef
+	viewPerm.parent = docDef
+	editPerm.parent = docDef
+
+	// Create schema
+	schema := &Schema{
+		definitions: map[string]*Definition{
+			"document": docDef,
+		},
+		caveats: make(map[string]*Caveat),
```
miparnisari: can we add one caveat here? to increase test coverage (lines 244-249)

pkg/query/build_tree.go
vroldanbet: doh I could have used `p.TypedParent()`. Pushed fix.

pkg/schema/v2/operations.go
vroldanbet: good idea! pushed

pkg/schema/v2/walk.go
```diff
@@ -1,5 +1,78 @@
...
+// unresolved ArrowReference nodes, the walk will fail with a clear error message.
+//
+// When enabled, arrow references (e.g., "parent->view") will automatically
+// traverse into their target permissions before visiting the arrow reference itself.
+// This enables single-pass validation of cross-definition references.
+//
+// Example:
+//
+//	schema, _ := BuildSchemaFromCompiledSchema(compiled)
+//	resolved, _ := ResolveSchema(schema)  // Required!
```
miparnisari: nit: this feels like an anti-pattern.. why can't we take a `ResolvedSchema` as a parameter? or why can't `opts` call `ResolveSchema` if it needs it so much?

pkg/schema/v2/schema.go
```diff
@@ -21,6 +63,16 @@ type Schema struct {
...
 }
 
+// Parent returns nil for Schema as it's the top-level element.
+func (s *Schema) Parent() Parented {
+	return nil
+}
+
+// setParent is a no-op for Schema since it's the top-level element.
+func (s *Schema) setParent(p Parented) {
+	// Schema has no parent
```
miparnisari, line 73: if schema cannot have a parent why does it implement the `Parented` interface?

i think this is related to this other code smell (in my view):

```go
type Parented interface {
   Parent() Parented
   setParent(Parented)
}
```

this seems to imply that the parent of an object _must also have a parent_ - why not do this?

```go
type Parented interface {
	Parent() any
	setParent(any)
}
```

then the change that you will have to make is during `FindParent`:

```go
		if parented, ok := current.(Parented); ok {
			current = parented.Parent()
		} else {
			// Reached a non-Parented element (e.g., Schema), stop traversal
			break
		}
```

pkg/schema/v2/operations.go
vroldanbet: This is accidental complexity, and in fact, these methods are not really used; they are only needed in anything that is above an operation. I've cleaned up in https://github.com/authzed/spicedb/pull/2761/changes/cb055dde3b00a784aa5993c9cebcd4bd374cfe62

pkg/schema/v2/schema.go
vroldanbet: cleaned this up

pkg/schema/v2/walk.go
vroldanbet: When you call `ResolveSchema()`, the `ArrowReference` nodes are replaced with `ResolvedArrowReference` nodes that contain a direct pointer to the actual target `Relation` or `Permission` objects. For my use case, this allows the walker to automatically traverse into the target namespace and relation, enabling single-pass validation and analysis of cross-definition references.

pkg/schema/v2/walk.go
vroldanbet: Good catch! I didn't have the factory method for the options at the beginning.

Moved it there!

pkg/schema/v2/walk_arrow_test.go
vroldanbet: done!

pkg/schema/v2/walk_arrow_test.go
vroldanbet: done

pkg/schema/v2/walk_arrow_test.go
vroldanbet: done

pkg/schema/v2/walk.go
vroldanbet: That's right! Since that was not the previous behavior, I made it optional to avoid breaking

vroldanbet:
>Where we had Parent() as a method before, let's call them by their name; so
>parentDef := p.Parent().(*schema.Definition)
>for example should really be
>parentDef := p.Definition()
>and then the implementation of (*schema.Permission).Parent() is simply return p.Definition(). Fits the interface. Maintains >cleanliness when you know the type.

@barakmich, did you notice `TypedParent()`? It gives you that, the actual typed parent. I just found it easier to use to have a single name that just has the actual return type, but I'm not strongly opinionated; it would be just a bit annoying to update all the callsites, but can do it.

barakmich:
> @barakmich, did you notice `TypedParent()`? It gives you that, the actual typed parent. I just found it easier to use to have a single name that just has the actual return type, but I'm not strongly opinionated; it would be just a bit annoying to update all the callsites, but can do it.

I did, but that name is less helpful IMHO; if you want to keep it that way, I guess I don't mind

pkg/schema/v2/schema.go
vroldanbet: I think it would require returning an error, and perhaps that's the best way to go.

pkg/schema/v2/operations.go
vroldanbet: I believe I added it for convenience when the element is nil, but that shouldn't happen often enough to justify it. I'll replace all the calls.

vroldanbet: done in https://github.com/authzed/spicedb/pull/2761/commits/9b3af9ae520c1b186dbf9412118169876488b395

pkg/schema/v2/schema_test.go
vroldanbet: correct! both things point to the same value

pkg/schema/v2/schema_test.go
vroldanbet:
```go
	// Test that FindParent works for immediate parents (no traversal needed).
	// The operation's parent is the permission itself, so FindParent should return it directly.
	perm := FindParent[*Permission](op)
	require.NotNil(t, perm, "should find permission parent")
	require.Equal(t, viewPerm, perm, "should find the immediate parent permission")
```

This test asserts that `FindParent` properly returns the parent of the provided argument, i.e., the immediate parent, and that it does not have to "hop" to other parents. It's a bit silly of a test, in retrospect 😅

pkg/schema/v2/walk.go
vroldanbet: This is a bit of a hack; I don't feel like `options` was the right place for this, but it was the place of least resistance, since the walker infrastructure doesn't natively provide a "traversal context".

So this registry would be local to the traversal journey itself. If you start with the schema, you'll see everything. But if you start at a relation or permission, it would only observe the relevant subset of the graph, and so it may only see some arrows, or none at all.

You raise a really good point: if you traverse the same arrow target through different parts of the schema, subsequent calls would not traverse. This was fine for my two use cases, because I didn't need to traverse the same subgraph again, but I wonder if it could be surprising in other use cases. What do you think? 

`visitedArrowTargets` was only added to prevent recursion during traversal of arrows.

pkg/schema/v2/operations.go
vroldanbet: - I ended up removing `cloneWithParent`, it leaked from the items _above operation_.
- SelfReference did not exist when I wrote this, so good find! Now it's not needed to add it
- `schemaUnit` represents something different to `Operation`. are you suggesting something like a "clonable" interface that is parameterized with generics?

pkg/schema/v2/schema.go
vroldanbet: I agree that the schema has no parent. But if I remove it, it has downstream effects: `Definition` can no longer return `Schema` in `Parent()`. It means that on traversal up, the schema is now _different_ and cannot be handled in the same type-agnostic way.

I personally don't think it's worth it; it just makes handling the top-level node of the tree annoying. Would it help if I clarified that we use schema Parented for convenience, so we don't treat the top node differently during traversal?

pkg/schema/v2/walk.go
vroldanbet: I thought `ResolvedSchema` had side-effects, but it actually creates a completely new tree, so yes, this is possible! good idea!

pkg/schema/v2/walk_test.go
vroldanbet: done!

pkg/schema/v2/walk.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/2761/commits/1ddddd1c8194ca4f613e50e8df84d5e81daed73f

pkg/schema/v2/walk.go
tstirrat15: Yeah, I don't have a sense of whether it's a correctness issue or just a missed potential optimization issue. If you're comfortable with it for now then I'm not too worried about it.

PR #2766, opened 2025-12-12
fix: change crdb ready check to measure idle conns

### LINKED ISSUE #2765: Add logic to healthchecks to wait until the specified minimum number of DB connections are available in the connection pool
We've observed cases where there's a large number of `ResourceExhausted` errors when a new pod comes up.

The `--datastore-conn-pool-read-min-open` and `--datastore-conn-pool-write-min-open` flags specify the minimum number of connections that are supposed to be available. Even with these flags in place, the pod is still acting as if it's waiting on connections to become available.

PGX should theoretically not report the pool as available until the minimum number of connections are available, but this doesn't match the observed behavior.

We'll want to get a reproduction of this behavior before calling it fixed.

This is where we're checking the readystate for CRDB: https://github.com/authzed/spicedb/blob/70cd75785eea27d6be697db13952a4745270b199/internal/datastore/crdb/crdb.go#L448-L459

`TotalConns` describes `ConstructingConns + IdleConns + AcquiredConns`. It may be that we need to check `IdleConns` specifically, since `ConstructingConns` aren't actually available for acquisition.

We should likely also check this in Postgres.

tstirrat15:
Fixes #2765 

## Description
We've seen cases in deployments where a SpiceDB rollout backed by CRDB will have many `ResourceExhausted` errors caused by a write request being  being unable to acquire a write pool connection within the timeout. This was surprising because we wait until the pool is filled to the number of minimum connections before a pod starts accepting traffic.

We noticed that we were counting `TotalConns`, which according to the docs is the sum of `ConstructingConns`, `IdleConns`, and `AcquiredConns`. When the system is starting up and new connections are being established, all (or most) of the connections would be in a constructing state, at which point they're not able to accept traffic.

Changing this to check for `IdleConns` should mean that we're waiting until the number of connections ready to accept traffic is above our minimum threshold, which should mitigate this issue.

## Changes
* Add docstring
* Check `IdleConns` instead of `TotalConns`
## Testing
Review.

### DIFF
--- a/internal/datastore/crdb/crdb_test.go
+++ b/internal/datastore/crdb/crdb_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/ory/dockertest/v3"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -138,7 +139,6 @@ func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
 	for _, quantization := range quantizationDurations {
 		t.Run(fmt.Sprintf("Quantization%s", quantization), func(t *testing.T) {
 			t.Parallel()
-			require := require.New(t)
 			ctx := context.Background()
 
 			ds := engine.NewDatastore(t, func(engine, uri string) datastore.Datastore {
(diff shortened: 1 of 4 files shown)


internal/datastore/crdb/crdb_test.go
```diff
@@ -138,7 +139,6 @@ func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
 	for _, quantization := range quantizationDurations {
 		t.Run(fmt.Sprintf("Quantization%s", quantization), func(t *testing.T) {
 			t.Parallel()
-			require := require.New(t)
```
tstirrat15, line 141: I had to do this in a couple of places because I wanted to use `require.EventuallyWithT`, which passes a `c` object to its child `require.*` calls, but if we're using an aliased `require` we can't do that.

internal/datastore/crdb/crdb_test.go
```diff
@@ -151,27 +151,29 @@ func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
...
 			})
 
-			r, err := ds.ReadyState(ctx)
-			require.NoError(err)
-			require.True(r.IsReady)
+			require.EventuallyWithT(t, func(c *assert.CollectT) {
+				r, err := ds.ReadyState(ctx)
+				require.NoError(c, err)
+				require.True(c, r.IsReady, "datastore not ready: %s", r.Message)
+			}, 3*time.Second, 50*time.Millisecond)
```
tstirrat15, line 165: `EventuallyWithT` is kinda nifty - it lets you write a set of assertions using `require` and keep asserting them until they pass. In this case, we're using it because it takes time for the CRDB connections to become both available and idle - previously this test was passing because there was a single available idle connection and the rest were constructing.

This also provides a bit of anecdotal support for the motivation of this work.

miparnisari: the fact that you had to add this `EventuallyWithT` everywhere makes me think that we should push that down to `ds.ReadyState` with a for loop or something

tstirrat15: Maybe? It's happening at the layer above: https://github.com/authzed/spicedb/blob/70cd75785eea27d6be697db13952a4745270b199/internal/services/health/health.go#L63

I think it's fine for ReadyState to be a query rather than something that waits for something to be healthy, but that's also not a hill I'll die on.

PR #2767, opened 2025-12-12
fix: Translate and preserve metadata in schema/v2

### LINKED ISSUE #2738: pkg/schema/v2:  schemas created with `schema.NewSchemaBuilder()` miss `core.Metadata`

macos

arm64

1.47.1

_No response_

### Steps to Reproduce

1. Create new schema with schema/v2
2. Call `.ToDefinitions()`

### Expected Result

The `core.NamespaceDefinition` and `core.CaveatDefinition` schema V1 types include the same `core.Metadata` that a schema V1 would have. Without it, reachability functions like `Definition.IsPermission(relName)` would fail to correctly identify a relation as a permission.

### Actual Result

Schemas created with V2 `schema.NewSchemaBuilder()` and converted to V1 using `.ToDefinitions()` have the same metadata as if the V1 compiler was used.

barakmich:
Fixes #2738

### DIFF
--- a/pkg/schema/v2/convert.go
+++ b/pkg/schema/v2/convert.go
@@ -9,11 +9,17 @@ import (
 )
 
 func convertDefinition(def *corev1.NamespaceDefinition) (*Definition, error) {
+	// Parse metadata
+	metadata, err := parseMetadata(def.GetMetadata())
+	if err != nil {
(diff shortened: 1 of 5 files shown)


tstirrat15 (approved):
This makes sense to me. Is this something that we'll get to discard with the new schema representation?

PR #2774, opened 2025-12-15
feat: introduce evaluation analysis for query plans

barakmich:

### DIFF
--- a/pkg/query/optimize_test.go
+++ b/pkg/query/optimize_test.go
@@ -158,8 +158,8 @@ func TestRemoveNullIterators(t *testing.T) {
 		require.True(t, changed)
 
 		// Should remove both empty iterators, leaving a union with 2 elements
-		resultUnion, ok := result.(*Union)
-		require.True(t, ok)
+		require.IsType(t, &Union{}, result)
+		resultUnion := result.(*Union)
 		require.Len(t, resultUnion.subIts, 2)
 		require.Equal(t, fixed1, resultUnion.subIts[0])
 		require.Equal(t, fixed2, resultUnion.subIts[1])
@@ -177,8 +177,8 @@ func TestRemoveNullIterators(t *testing.T) {
 		require.True(t, changed)
 
 		// Should return an empty fixed iterator
-		resultFixed, ok := result.(*FixedIterator)
-		require.True(t, ok)
+		require.IsType(t, &FixedIterator{}, result)
+		resultFixed := result.(*FixedIterator)
 		require.Empty(t, resultFixed.paths)
 	})
 
@@ -221,8 +221,8 @@ func TestRemoveNullIterators(t *testing.T) {
 		require.True(t, changed)
 
 		// Should return an empty fixed iterator
-		resultFixed, ok := result.(*FixedIterator)
-		require.True(t, ok)
+		require.IsType(t, &FixedIterator{}, result)
+		resultFixed := result.(*FixedIterator)
 		require.Empty(t, resultFixed.paths)
 	})
 }
@@ -244,8 +244,8 @@ func TestApplyOptimizations(t *testing.T) {
 
 		// The outer union should still be a union (has 2 elements)
 		// but the inner singleton union should be collapsed
-		resultUnion, ok := result.(*Union)
-		require.True(t, ok)
+		require.IsType(t, &Union{}, result)
+		resultUnion := result.(*Union)
 		require.Len(t, resultUnion.subIts, 2)
 		require.Equal(t, fixed, resultUnion.subIts[0])
 	})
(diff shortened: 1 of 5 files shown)


pkg/query/analyze_test.go
```diff
@@ -0,0 +1,541 @@
...
+	)
+
+	// Create a context with analysis enabled
+	analyze := make(map[string]AnalyzeStats)
+	ctx := &Context{
+		Context:  context.Background(),
+		Executor: LocalExecutor{},
+		Reader:   ds.SnapshotReader(datastore.NoRevision),
+		Analyze:  analyze,
+	}
```
miparnisari, line 147: nit: time for `NewContext(...)` ?

pkg/query/analyze_test.go
```diff
@@ -0,0 +1,541 @@
...
+	// Create a datastore
+	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	require.NoError(t, err)
+
+	// Create a simple iterator
+	fixed := NewFixedIterator(
+		Path{
+			Resource: Object{ObjectType: "document", ObjectID: "doc1"},
+			Subject:  ObjectAndRelation{ObjectType: "user", ObjectID: "alice", Relation: "..."},
+		},
```
miparnisari, line 133: nit: it probably doesn't matter in this test, but wouldn't `Relation` be mandatory when building a Path? in which case `NewPath(res, rel, sub, ....optionals)` may be needed

pkg/query/context.go
```diff
@@ -197,6 +245,75 @@ func (ctx *Context) wrapPathSeqForTracing(it Iterator, pathSeq PathSeq) PathSeq
...
+				// Record partial results and timing before early exit
+				elapsed := time.Since(startTime)
+				ctx.recordAnalysisResults(iterID, opType, resultCount, elapsed)
+				return
+			}
+		}
+
+		// Record final result count and timing
+		elapsed := time.Since(startTime)
+		ctx.recordAnalysisResults(iterID, opType, resultCount, elapsed)
```
miparnisari:
```suggestion
		startTime := time.Now()
		resultCount := 0

		defer func() {
			elapsed := time.Since(startTime)
			ctx.recordAnalysisResults(iterID, opType, resultCount, elapsed)
		}()

		for path, err := range pathSeq {
			if err == nil {
				resultCount++
			}
			if !yield(path, err) {
				return
			}
		}
```

miparnisari (dismissed):

pkg/query/analyze_test.go
```diff
@@ -0,0 +1,541 @@
...
+			"Optimized tree should have fewer Check calls (unoptimized: %d, optimized: %d)",
+			unoptimizedStats.CheckCalls, optimizedStats.CheckCalls)
+
+		t.Logf("Unoptimized tree had %d Check calls, optimized tree has %d Check calls (%.1f%% reduction)",
+			unoptimizedStats.CheckCalls, optimizedStats.CheckCalls,
+			100.0*float64(unoptimizedStats.CheckCalls-optimizedStats.CheckCalls)/float64(unoptimizedStats.CheckCalls))
+
+		// Verify that timing was tracked
+		require.Greater(t, unoptimizedStats.CheckTime, time.Duration(0), "Unoptimized tree should have non-zero Check time")
+		require.Greater(t, optimizedStats.CheckTime, time.Duration(0), "Optimized tree should have non-zero Check time")
```
tstirrat15: require.Positive

pkg/query/analyze_test.go
```diff
@@ -0,0 +1,541 @@
...
+
+		t.Logf("Timing - Unoptimized: %v, Optimized: %v", unoptimizedStats.CheckTime, optimizedStats.CheckTime)
+	})
+
+	t.Run("collapsing singleton unions reduces overhead", func(t *testing.T) {
```
tstirrat15, line 357: I like this test, but is it more about the optimizer or the analyzer? Or does that matter too much?

pkg/query/context.go
```diff
@@ -36,12 +37,23 @@ func (t *TraceLogger) EnterIterator(it Iterator, resources []Object, subject Obj
 	}
 
 	indent := strings.Repeat("  ", t.depth)
+	var idPrefix string
+	if id := it.ID(); id != "" {
```
tstirrat15: Worth pulling into a utility?

pkg/query/context.go
```diff
@@ -197,6 +245,75 @@ func (ctx *Context) wrapPathSeqForTracing(it Iterator, pathSeq PathSeq) PathSeq
...
+
+	// Increment call counter based on operation type
+	stats := ctx.Analyze[iterID]
+	switch opType {
+	case "Check":
+		stats.CheckCalls++
+	case "IterSubjects":
+		stats.IterSubjectsCalls++
+	case "IterResources":
+		stats.IterResourcesCalls++
```
tstirrat15: Sanity check: do these operations need to be thread-safe?

pkg/query/optimize_test.go
```diff
@@ -412,8 +412,12 @@ func TestApplyOptimizations(t *testing.T) {
 		// Both should result in an empty fixed iterator
 		// After removing all empties, we get Union[], which then gets optimized
 		// to an empty fixed iterator by CollapseSingletonUnionAndIntersection
-		require.Equal(t, result1, result2)
-		_, ok := result1.(*FixedIterator)
-		require.True(t, ok, "result1 should be a FixedIterator (empty), got %T", result1)
+		fixed1, ok1 := result1.(*FixedIterator)
+		require.True(t, ok1, "result1 should be a FixedIterator (empty), got %T", result1)
```
tstirrat15: https://pkg.go.dev/github.com/stretchr/testify/require#IsType or https://pkg.go.dev/github.com/stretchr/testify/require#Implements?

tstirrat15 (dismissed):
See comments, generally LGTM

pkg/query/analyze_test.go
barakmich: filed an issue

pkg/query/analyze_test.go
barakmich: Eh, kinda all one package

pkg/query/context.go
barakmich: good call; created a thread-safe Collector to handle that at the map level (the individual structs are still simple)

tstirrat15 (approved):
LGTM

PR #2775, opened 2025-12-16
fix: improve error messages for missing database tables

### LINKED ISSUE #2749: update error messages about missing tables in the database to say "run spicedb migrate"
Many users on Discord have reported this friction: [1](https://discord.com/channels/844600078504951838/844600078948630559/1293436935284461639), [2](https://discord.com/channels/844600078504951838/844600078948630559/1275862772567244851), [3](https://discord.com/channels/844600078504951838/844600078948630559/1189148009078988871), [4](https://discord.com/channels/844600078504951838/844600078948630559/1012488411900301443), to name a few.

To reproduce: If you update https://github.com/authzed/spicedb/blob/main/docker-compose.yaml to remove the `migrate` service, and then run `docker-compose up --build`, and then try to write a schema, you get this:

```
// http 500
{
  "code": 2,
  "message": "unable to list caveat: ERROR: relation \"caveat\" does not exist (SQLSTATE 42P01)",
  "details": []
}
```

which makes a very poor experience since it doesn't tell the operator how to fix it.

Let's update all the error messages of type `relation X does not exist` for every datastore to include `please run spicedb datastore migrate`.

ivanauth:
When users start SpiceDB without running migrations first, they get a confusing error like:

```
unable to list namespaces: ERROR: relation "namespace_config" does not exist (SQLSTATE 42P01)
```

This doesn't tell them what went wrong or how to fix it.

Now the error message is clear and actionable:

```
datastore error: the database schema has not been initialized; please run "spicedb datastore migrate": ERROR: relation "namespace_config" does not exist (SQLSTATE 42P01)
```

This applies to all SQL datastores (postgres, mysql, crdb) and covers read operations, write operations, and statistics queries.

Closes #2749

### DIFF
--- a/internal/datastore/common/errors.go
+++ b/internal/datastore/common/errors.go
@@ -121,3 +121,33 @@ type RevisionUnavailableError struct {
 func NewRevisionUnavailableError(err error) error {
 	return RevisionUnavailableError{err}
 }
+
+// SchemaNotInitializedError is returned when a datastore operation fails because the
+// required database tables do not exist. This typically means that migrations have not been run.
+type SchemaNotInitializedError struct {
+	error
+}
+
+func (err SchemaNotInitializedError) GRPCStatus() *status.Status {
+	// TODO: Update to use ERROR_REASON_DATASTORE_NOT_MIGRATED once authzed/api#159 is merged
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.FailedPrecondition,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_UNSPECIFIED,
+			map[string]string{},
+		),
+	)
+}
+
+func (err SchemaNotInitializedError) Unwrap() error {
+	return err.error
+}
+
+// NewSchemaNotInitializedError creates a new SchemaNotInitializedError with a helpful message
+// instructing the user to run migrations.
+func NewSchemaNotInitializedError(underlying error) error {
+	return SchemaNotInitializedError{
+		fmt.Errorf("%w. please run \"spicedb datastore migrate\"", underlying),
+	}
+}
--- a/internal/datastore/common/errors.go
+++ b/internal/datastore/common/errors.go
@@ -129,12 +129,11 @@ type SchemaNotInitializedError struct {
 }
 
 func (err SchemaNotInitializedError) GRPCStatus() *status.Status {
-	// TODO: Update to use ERROR_REASON_DATASTORE_NOT_MIGRATED once authzed/api#159 is merged
 	return spiceerrors.WithCodeAndDetails(
 		err,
 		codes.FailedPrecondition,
 		spiceerrors.ForReason(
-			v1.ErrorReason_ERROR_REASON_UNSPECIFIED,
+			v1.ErrorReason_ERROR_REASON_DATASTORE_NOT_MIGRATED,
 			map[string]string{},
 		),
 	)
--- a/internal/datastore/crdb/caveat.go
+++ b/internal/datastore/crdb/caveat.go
@@ -9,7 +9,9 @@ import (
 	sq "github.com/Masterminds/squirrel"
 	"github.com/jackc/pgx/v5"
 
+	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/internal/datastore/crdb/schema"
+	pgxcommon "github.com/authzed/spicedb/internal/datastore/postgres/common"
 	"github.com/authzed/spicedb/internal/datastore/revisions"
 	"github.com/authzed/spicedb/pkg/datastore"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
@@ -52,6 +54,9 @@ func (cr *crdbReader) LegacyReadCaveatByName(ctx context.Context, name string) (
 	if err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
 			err = datastore.NewCaveatNameNotFoundErr(name)
+		}
+		if pgxcommon.IsMissingTableError(err) {
+			err = common.NewSchemaNotInitializedError(err)
 		}
 		return nil, datastore.NoRevision, fmt.Errorf(errReadCaveat, name, err)
 	}
@@ -109,6 +114,9 @@ func (cr *crdbReader) lookupCaveats(ctx context.Context, caveatNames []string) (
 		return nil
 	}, sql, args...)
 	if err != nil {
+		if pgxcommon.IsMissingTableError(err) {
+			err = common.NewSchemaNotInitializedError(err)
+		}
 		return nil, fmt.Errorf(errListCaveats, err)
 	}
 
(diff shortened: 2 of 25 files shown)


internal/datastore/common/errors.go
```diff
@@ -176,3 +176,32 @@ type RevisionUnavailableError struct {
...
+type SchemaNotInitializedError struct {
+	error
+}
+
+func (err SchemaNotInitializedError) GRPCStatus() *status.Status {
+	return spiceerrors.WithCodeAndDetails(
+		err,
+		codes.FailedPrecondition,
+		spiceerrors.ForReason(
+			v1.ErrorReason_ERROR_REASON_UNSPECIFIED,
```
miparnisari: nit: please create the right error code

internal/datastore/common/errors.go
```diff
@@ -176,3 +176,32 @@ type RevisionUnavailableError struct {
...
+
+func (err SchemaNotInitializedError) Unwrap() error {
+	return err.error
+}
+
+// NewSchemaNotInitializedError creates a new SchemaNotInitializedError with a helpful message
+// instructing the user to run migrations.
+func NewSchemaNotInitializedError(underlying error) error {
+	return SchemaNotInitializedError{
+		fmt.Errorf("datastore error: the database schema has not been initialized; please run \"spicedb datastore migrate\": %w", underlying),
```
miparnisari:
```suggestion
		fmt.Errorf("%w. please run \"spicedb datastore migrate\", underlying),
```

internal/datastore/crdb/caveat.go
```diff
@@ -53,6 +54,9 @@ func (cr *crdbReader) ReadCaveatByName(ctx context.Context, name string) (*core.
 		if errors.Is(err, pgx.ErrNoRows) {
 			err = datastore.NewCaveatNameNotFoundErr(name)
 		}
+		if wrappedErr := pgxcommon.WrapMissingTableError(err); wrappedErr != nil {
+			return nil, datastore.NoRevision, wrappedErr
```
miparnisari:
```suggestion
		if pgxcommon.IsMissingTableError(err) {
			err = common.NewSchemaNotInitializedError(err)
		}
```

to keep the same pattern as the lines before

internal/datastore/postgres/common/errors.go
```diff
@@ -19,6 +19,7 @@ const (
 	pgReadOnlyTransaction       = "25006"
 	pgQueryCanceled             = "57014"
 	pgInvalidArgument           = "22023"
+	pgMissingTable              = "42P01"
```
miparnisari: there are two constants with the same value defined elsewhere: `postgresMissingTableErrorCode`

please remove them in favor of this one

internal/datastore/postgres/revisions.go
```diff
@@ -312,6 +312,9 @@ func createNewTransaction(ctx context.Context, tx pgx.Tx, metadata map[string]an
 
 	cterr := tx.QueryRow(ctx, sql, args...).Scan(&newXID, &newSnapshot, &timestamp)
 	if cterr != nil {
+		if wrappedErr := common.WrapMissingTableError(cterr); wrappedErr != nil {
+			return newXID, newSnapshot, timestamp, wrappedErr
```
miparnisari:
```suggestion
			cterr = ....
```

internal/datastore/postgres/reader.go
```diff
@@ -252,6 +264,9 @@ func (r *pgReader) LookupNamespacesWithNames(ctx context.Context, nsNames []stri
 		return r.aliveFilter(original).Where(clause)
 	})
 	if err != nil {
+		if wrappedErr := pgxcommon.WrapMissingTableError(err); wrappedErr != nil {
+			return nil, wrappedErr
```
miparnisari:
```suggestion
			err = wrappedErr
```

otherwise you lose the `errUnableToListNamespaces` below

miparnisari (changes requested):
Used Postman and tried all the APIs against CRDB and Spanner. Notes:
- we're missing this change for Spanner. You can use `docker-compose -f docker-compose.spanner.yaml up --build` to test. E.g. 
    - call to WriteSchema I get `"unable to list caveats: spanner: code = \"NotFound\", desc = \"Table not found: caveat\", requestID = \"1.23ff9b5081a896f8.1.1.70.1\"",`.
    - call to DeleteRelationships I get `"object definition doc not found"`. This line has to be updated to account for when the error description says "table" https://github.com/authzed/spicedb/blob/f41f64166fee87eb65077120e50f8ea6ebdc16f6/internal/datastore/spanner/reader.go#L300
- in many cases, the errors returned to the user are prefixed with `error encoding zedtoken: unable to read unique ID` which makes no sense to me, and probably not to the end user 😆  It comes from https://github.com/authzed/spicedb/blob/01d72a1387ad154c0d268e161e5fdc37788eca2a/pkg/zedtoken/zedtoken.go#L71-L74. i'm very tempted to remove that `error encoding zedtoken` and to clarify what the "unique ID" is
- in spanner, if you didn't run migrations, calling Watch api gives a HTTP 400. But against CRDB, the watch service is completely disabled: `unknown service authzed.api.v1.WatchService`, which means that you need to re-boot the server after you ran the migrations. I wonder if the server is supposed to be hable to handle migrations running at the same time ?

Anyway. Could you update https://github.com/authzed/spicedb/blob/3370198fda1678280ff81477b1d3eca2d3cbdecd/internal/datastore/crdb/crdb.go#L576 to set a different error message in this case?

- when you call ImportBulk, you get this

```json
{
    "code": 2,
    "message": "statement description failed: ERROR: relation \"relation_tuple\" does not exist (SQLSTATE 42P01)",
    "details": []
}
```

, seems like we missed an error handling here: https://github.com/authzed/spicedb/blob/3370198fda1678280ff81477b1d3eca2d3cbdecd/internal/datastore/postgres/common/bulk.go#L79

miparnisari:
I also wonder if instead of littering all the datastore implementations with error handling, we could just prevent the server from coming up online when you have never ran `migrate`.

E.g. here

https://github.com/authzed/spicedb/blob/70cd75785eea27d6be697db13952a4745270b199/internal/services/health/health.go#L112-L115

if `dsReady.message` includes `datastore is not migrated: currently at revision \"\"`, then we could tell people to run `migrate`. we'd need to write a generic test ensuring that all datastores return empty revision when they haven't been migrated. 

I'm sure there are more changes required than this, perhaps an update to `Server.Complete` to include this check, but @josephschorr does this sound reasonable?

ivanauth:
I replaced scattered `IsMissingTableError` checks with a single gRPC readiness middleware that blocks ALL requests until the datastore is migrated.

Before: 15 files with scattered IsMissingTableError checks (easy to miss spots)
After: One middleware gates ALL requests until migrated

ivanauth:
Squashed to 2 commits and addressed all feedback:

* Replaced `WrapMissingTableError` with `IsMissingTableError` and `NewSchemaNotInitializedError` across datastores to preserve caller context
* Added case-insensitive “table not found” matching for Spanner
* Switched readiness gate to `resenje.org/singleflight` to fix lint
* Error code remains `UNSPECIFIED` until api#159 lands

Zedtoken prefix and Watch API behavior seem like separate concerns - happy to tackle in follow-up PRs.

Unit test failure (`TestCertWatcherStart` in `pkg/x509util`) is flaky and unrelated.

PR #2779, opened 2025-12-17
fix: re-enable async logging with proper shutdown flushing

### LINKED ISSUE #889: re-enable async logging
`cobrautil/v2` async logger was introduced in https://github.com/authzed/spicedb/pull/844 and then disabled in https://github.com/authzed/spicedb/pull/888. Currently there is no means to prevent log from being lost on process termination because the corresponding log buffer is not being flushed.

Reintroduce async log, making sure we can flush out when the process is terminating.

ivanauth:
Fixes #889

Async logging was disabled because nothing was flushing the diode.Writer buffer on shutdown - logs were getting lost. The cobrazerolog library hides the closer so we couldn't fix it there.

This creates our own logging setup that exposes the closer, stores it globally, and flushes it via `defer log.Close()` in main. Also restructured main() to return an int so defers actually run before exit.


tstirrat15:
We own `cobrautil` - it might make sense to fix it there.

ivanauth:
Update: cobrautil fix submitted

Instead of duplicating cobrautil's logging setup, I've submitted a fix upstream: https://github.com/jzelinskie/cobrautil/pull/66

That PR:
- Exposes `Builder.Close()` to flush the async diode writer on shutdown
- Fixes `WithAsync(size, pollInterval)` params being ignored (hardcoded to 1000/10ms)
- Wraps `os.Stderr` to prevent `diode.Writer.Close()` from closing the fd
- Guards against double `RunE()` leaking poll goroutines
- Defaults size to 1000 if <= 0 to prevent panic

Once that's merged and tagged, this PR can be significantly simplified:
- **Delete** `logging_run_e.go` and `logging_run_e_test.go` (no need to duplicate cobrautil's setup)
- **Delete** `noCloseWriter` (cobrautil now handles this)
- **Simplify** `internal/logging/logger.go` — store the `*Builder` ref, call `builder.Close()`
- **Keep** the `main()` → `run() int` restructuring (so defers run before `os.Exit`)
- **Bump** cobrautil dependency

cc @tstirrat15 per your suggestion to fix it in cobrautil

PR #2780, opened 2025-12-17
Fix composable schema interactions with expiration feature

### LINKED ISSUE #2778: Multiple similar `use` directives across different imported files in Composable Schemas cause errors
If you've got:

```
use expiration

import "./foo.zed"

definition resource {}
```

in one file and then this in `foo.zed`:

```
use expiration

definition user {}
```

The compiler will yell at you about multiple `use expiration` directives. The "normal" behavior for use directives in other languages is that they apply to the file, not to the entire bundled whole, but that's not the way our system behaves.

The simplest way to do this is likely to allow for duplicates and then merge them.

tstirrat15:
Fixes #2778 

## Description
There were two problems in #2778:

1. `with expiration` wasn't correctly being applied to wildcard relations. This was fixed in #2564 but wasn't ported to the `composableschemadsl` package.
2. The composable schema compiler was barking about multiple `use` directives. In a monolithic schema that might make sense; in a multi-file schema it doesn't really. They're also a little weird because they're interpreted at the level of the entire schema rather than on a per-file basis, but normally `use` directives (e.g. in JS) are expected to behave on a per-file basis.

This fixes both of those problems. I'll get this into zed shortly thereafter.

## Changes
Will annotate.

## Testing
Review. See that tests pass.


pkg/composableschemadsl/compiler/importer-test/expiration-usage/expected.zed
tstirrat15, line 1: This particular issue wasn't related to the import functionality, but it showed up in composable schema stuff so I'm going to leave this in as a sort of regression test.

pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -55,7 +54,7 @@ func (cs CompiledSchema) SourcePositionToRunePosition(source input.Source, posit
 type config struct {
 	skipValidation   bool
 	objectTypePrefix *string
-	allowedFlags     []string
+	allowedFlags     *mapz.Set[string]
```
tstirrat15, line 57: Changed `enabledFlags` and `allowedFlags` to be sets since we don't actually care about duplicates.

pkg/composableschemadsl/compiler/compiler_test.go
```diff
@@ -1254,17 +1254,23 @@ func TestCompile(t *testing.T) {
 			},
 		},
 		{
-			"duplicate use pragmas",
```
tstirrat15, line 1257: Got rid of this test, since we don't want to enforce this anymore; added a test from `schemadsl` that captured the wildcard + expiration issue.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -668,29 +667,6 @@ func translateSpecificTypeReference(tctx *translationContext, typeRefNode *dslNo
 		return nil, typeRefNode.Errorf("%w", err)
 	}
 
-	if typeRefNode.Has(dslshape.NodeSpecificReferencePredicateWildcard) {
-		ref := &core.AllowedRelation{
-			Namespace: nspath,
-			RelationOrWildcard: &core.AllowedRelation_PublicWildcard_{
-				PublicWildcard: &core.AllowedRelation_PublicWildcard{},
-			},
```
tstirrat15, line 676: This change and the changes immediately below are transliterating the changes from #2564 into this file.

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -946,9 +937,9 @@ func translateUseFlag(tctx *translationContext, useFlagNode *dslNode) error {
...
 		return err
 	}
-	if slices.Contains(tctx.enabledFlags, flagName) {
-		return useFlagNode.Errorf("found duplicate use flag: %s", flagName)
-	}
-	tctx.enabledFlags = append(tctx.enabledFlags, flagName)
+	// NOTE: we're okay with multiple instances of a given `use` directive in
+	// composable schemas, because each file may declare it separately
+	// and that should be valid.
+	tctx.enabledFlags.Add(flagName)
```
tstirrat15, line 943: This is the other substantive change.

tstirrat15 (commented):
See comment

PR #2782, opened 2025-12-18
chore: replace ory/dockertest with testcontainers

miparnisari:
(this is @tstirrat15)

## Description
`testcontainers` is a more modern take on `dockertest` that provides a number of nice modules and handles on test docker containers. It also does a much better job of cleaning up after itself - it's much harder to end up with orphaned test containers when a test fails or panics or otherwise crashes. This reworks our test code to use `testcontainers`.

## Changes
See annotations.

## Testing
Review. See that tests pass


cmd/spicedb/integration/migrate_integration_test.go
```diff
@@ -20,18 +23,14 @@ import (
...
-
 	// Create a bridge network for testing.
-	network, err := pool.Client.CreateNetwork(docker.CreateNetworkOptions{
-		Name: bridgeNetworkName,
-	})
+	net, err := network.New(ctx, network.WithDriver("bridge"), network.WithLabels(map[string]string{"name": bridgeNetworkName}))
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		_ = pool.Client.RemoveNetwork(network.ID)
+		_ = net.Remove(ctx)
```
mdelapenya: suggestion: use CleanupNetwork, and call it before require, as it handle nils

```go
nw, err := network.New(ctx, network.WithDriver("bridge"), network.WithLabels(map[string]string{"name": bridgeNetworkName}))
testcontainers.CleanupNetwork(t, nw)
require.NoError(t, err)
    
```

cmd/spicedb/integration/migrate_integration_test.go
```diff
@@ -45,43 +44,43 @@ func TestMigrate(t *testing.T) {
...
-				Repository: "authzed/spicedb",
-				Tag:        "ci",
-				Cmd:        []string{"migrate", "head", "--datastore-engine", engineKey, "--datastore-conn-uri", db},
-				NetworkID:  bridgeNetworkName,
-				Env:        envVars,
-			}, func(config *docker.HostConfig) {
-				config.RestartPolicy = docker.RestartPolicy{
-					Name: "no",
-				}
+			container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
```
mdelapenya: question: why not using the module? https://github.com/Mariscal6/testcontainers-spicedb-go

If it does not provide the APIs need by these tests, we can contribute them to the upstream. I'm pretty sure @Mariscal6 will be happy to review that

cmd/spicedb/integration/migrate_integration_test.go
```diff
@@ -45,43 +44,43 @@ func TestMigrate(t *testing.T) {
...
+					Networks: []string{bridgeNetworkName},
+					Env:      envVars,
+				},
+				Started: true,
 			})
 			require.NoError(t, err)
 			t.Cleanup(func() {
-				_ = pool.Purge(resource)
+				_ = container.Terminate(ctx)
 			})
```
mdelapenya: suggestion: use CleanupContainer, as for networks.

```go
ctr, err := ... // module's Run function OR the GenericContainer() one.
testcontainers.CleanupContainer(t, ctr)
require.NoError(t, err)

File: cmd/spicedb/integration/migrate_integration_test.go [hunk 4]
```diff
@@ -45,43 +44,43 @@ func TestMigrate(t *testing.T) {
...
 
 			// Ensure the command completed successfully.
-			status, err := pool.Client.WaitContainerWithContext(resource.Container.ID, t.Context())
+			state, err := container.State(ctx)
 			require.NoError(t, err)
 
-			if status != 0 {
+			if state.ExitCode != 0 {
```
mdelapenya: suggestion: you could define a log consumer at container creation, and print it here. Please see https://golang.testcontainers.org/features/follow_logs/#following-container-logs

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,83 +35,75 @@ func TestSchemaWatch(t *testing.T) {
...
-				Repository: "authzed/spicedb",
-				Tag:        "ci",
-				Cmd:        []string{"migrate", "head", "--datastore-engine", driverName, "--datastore-conn-uri", db},
-				NetworkID:  bridgeNetworkName,
-				Env:        envVars,
-			}, func(config *docker.HostConfig) {
-				config.RestartPolicy = docker.RestartPolicy{
-					Name: "no",
-				}
+			migrateContainer, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
```
mdelapenya: suggestion: given there are a few of this, you could abstract the container creation building your own test "framework", something like this:

```go
// pseudocode
func NewTestSpiceDB(ctx Context, opts ...ContainerOpts) {...}
```

cmd/spicedb/integration/serve_integration_test.go
```diff
@@ -110,31 +106,28 @@ func gracefulShutdown(pool *dockertest.Pool, serveResource *dockertest.Resource)
 		return true
 
 	case <-time.After(10 * time.Second):
-		_ = pool.Purge(serveResource)
+		_ = container.Terminate(ctx)
 		return false
 	}
 }
 
 func TestGracefulShutdownInMemory(t *testing.T) {
```
mdelapenya: I think this tests is testing the library, you probably don't want to test that here

cmd/spicedb/integration/serve_integration_test.go
```diff
@@ -161,84 +154,76 @@ func TestGracefulShutdown(t *testing.T) {
...
 			bridgeNetworkName := fmt.Sprintf("bridge-%s", uuid.New().String())
 
-			pool, err := dockertest.NewPool("")
-			require.NoError(t, err)
-
 			// Create a bridge network for testing.
-			network, err := pool.Client.CreateNetwork(docker.CreateNetworkOptions{
-				Name: bridgeNetworkName,
-			})
+			net, err := network.New(ctx, network.WithDriver("bridge"), network.WithLabels(map[string]string{"name": bridgeNetworkName}))
```
mdelapenya: question: is the network really needed? I think it's not, as I do not see any other container being attached to it. If testing the container in a network is required, then it's fine, else, I would remove it to simplify the test scenarios and avoid noise in the case the network creation failed.

cmd/spicedb/integration/serve_integration_test.go
```diff
@@ -110,31 +106,28 @@ func gracefulShutdown(pool *dockertest.Pool, serveResource *dockertest.Resource)
...
 	t.Cleanup(func() {
-		_ = pool.Purge(serveResource)
+		_ = container.Terminate(ctx)
 	})
 
-	require.True(t, gracefulShutdown(pool, serveResource))
+	require.True(t, gracefulShutdown(ctx, container))
 }
 
 type watchingWriter struct {
```
mdelapenya: I think this watchingWriter struct is the log consumer from testcontainers. You could simply replace it

cmd/spicedb/integration/servetesting_integration_test.go
```diff
@@ -189,38 +187,59 @@ const retryCount = 8
 // newTester spins up a SpiceDB server running against a specific datastore with a specific access token.
 // It also writes or reads a schema.
 // On test termination it cleans up all resources.
-func newTester(t *testing.T, containerOpts *dockertest.RunOptions, token string, withExistingSchema bool) (*spicedbHandle, error) {
+func newTester(t *testing.T, containerReq testcontainers.ContainerRequest, token string, withExistingSchema bool) (*spicedbHandle, error) {
```
mdelapenya: suggestion: we advocate for migrating to the newer APIs using the Run function and passing container customisers as functional options

internal/datastore/crdb/crdb_test.go
```diff
@@ -455,39 +453,53 @@ func newCRDBWithUser(t *testing.T, pool *dockertest.Pool) (adminConn *pgx.Conn,
...
-	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository: "mirror.gcr.io/cockroachdb/cockroach",
-		Tag:        "v" + crdbTestVersion(),
-		Cmd:        []string{"start-single-node", "--certs-dir", "/certs", "--accept-sql-without-tls"},
-		Mounts:     []string{certDir + ":/certs"},
-	})
+	container, err := cockroachdb.Run(t.Context(),
+		"cockroachdb/cockroach:v"+crdbTestVersion(),
+		cockroachdb.WithInsecure(),
+		testcontainers.CustomizeRequest(testcontainers.GenericContainerRequest{
```
mdelapenya: suggestion: use [WithCmd](https://golang.testcontainers.org/features/creating_container/#withcmd), and [WithMounts](https://golang.testcontainers.org/features/creating_container/#withmounts) (although we recommend [WIthFiles](https://golang.testcontainers.org/features/creating_container/#withfiles) to avoid sharing the filesystem with the host == test isolation)

internal/datastore/crdb/crdb_test.go
```diff
@@ -455,39 +453,53 @@ func newCRDBWithUser(t *testing.T, pool *dockertest.Pool) (adminConn *pgx.Conn,
...
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		require.NoError(t, pool.Purge(resource))
+		require.NoError(t, container.Terminate(context.Background()))
 	})
 
-	port := resource.GetPort(fmt.Sprintf("%d/tcp", 26257))
-	require.NoError(t, pool.Retry(func() error {
-		var err error
+	mappedPort, err := container.MappedPort(t.Context(), "26257")
```
mdelapenya: suggestion: the cockroachDB module already waits for an internal state, including the `/health` endpoint and a SQL connection to the 26257  `defaultSQLPort`. Please see https://github.com/testcontainers/testcontainers-go/blob/main/modules/cockroachdb/cockroachdb.go#L194-L208

This entire block would go away

internal/testserver/datastore/crdb.go
```diff
@@ -3,122 +3,107 @@ package datastore
...
 var _ RunningEngineForTest = (*crdbTester)(nil)
 
 // RunCRDBForTesting returns a RunningEngineForTest for CRDB
 func RunCRDBForTesting(t testing.TB, bridgeNetworkName string, crdbVersion string) *crdbTester {
-	pool, err := dockertest.NewPool("")
+	ctx := context.Background()
+	name := "crds-" + uuid.New().String()
+
+	containerReq := testcontainers.ContainerRequest{
+		Name: name,
```
mdelapenya: suggestion: we advice against using names, as it makes tests flaky in the case multiple parallel tests call this method, creating a container with the same name.

internal/testserver/datastore/crdb.go
```diff
@@ -3,122 +3,107 @@ package datastore
...
+		Cmd:  []string{"--insecure", "--max-offset=50ms"},
+	}
+
+	if bridgeNetworkName != "" {
+		containerReq.Networks = []string{bridgeNetworkName}
+	}
+
+	opts := []testcontainers.ContainerCustomizer{
+		cockroachdb.WithInsecure(),
+		testcontainers.CustomizeRequest(testcontainers.GenericContainerRequest{
```
mdelapenya: suggestion: do not use the customizeRequest option, just use the WithCmd one, as this is what you modify in https://github.com/authzed/spicedb/pull/2782/files#diff-17664b2ee7777f7a101b83c18a1feb2d0a7ef997d595b28e98fb0c7d440e7c06R33

internal/testserver/datastore/crdb.go
```diff
@@ -3,122 +3,107 @@ package datastore
...
-	}, func(config *docker.HostConfig) {
-		// set AutoRemove to true so that stopped container goes away by itself
-		config.AutoRemove = true
-		config.RestartPolicy = docker.RestartPolicy{Name: "no"}
+	t.Cleanup(func() {
+		require.NoError(t, container.Terminate(ctx))
 	})
+
+	// enable changefeeds
+	code, _, _ := container.Exec(ctx, []string{
```
mdelapenya: suggestion: check for errors here

internal/testserver/datastore/crdb.go
```diff
@@ -3,122 +3,107 @@ package datastore
...
-	port      string
+	databaseConnectStr string
 }
 
 var _ RunningEngineForTest = (*crdbTester)(nil)
 
 // RunCRDBForTesting returns a RunningEngineForTest for CRDB
 func RunCRDBForTesting(t testing.TB, bridgeNetworkName string, crdbVersion string) *crdbTester {
-	pool, err := dockertest.NewPool("")
+	ctx := context.Background()
```
mdelapenya: suggestion: start any test helper function with `t.Helper()`

internal/testserver/datastore/crdb.go
mdelapenya: suggestion: you probably want to take a look at the container lifecycle hooks, and create this right after the container is created. https://golang.testcontainers.org/features/creating_container/#withadditionallifecyclehooks and https://golang.testcontainers.org/features/creating_container/#__tabbed_4_1

internal/testserver/datastore/crdb.go
```diff
@@ -3,122 +3,107 @@ package datastore
...
+
+	// enable changefeeds
+	code, _, _ := container.Exec(ctx, []string{
+		"cockroach", "sql",
+		"--insecure",
+		"-e", "SET CLUSTER SETTING kv.rangefeed.enabled = true;",
+	})
+	require.Equal(t, 0, code)
+
+	// create DB
```
mdelapenya: suggestion: there is a `cockroachdb.WithDatabase` option. See https://github.com/testcontainers/testcontainers-go/blob/main/modules/cockroachdb/options.go#L20

You can also get the ConnectionString to the instance for that database and credentials. See https://golang.testcontainers.org/modules/cockroachdb/#connectionstring

internal/testserver/datastore/mysql.go
```diff
@@ -50,59 +50,74 @@ func RunMySQLForTesting(t testing.TB, bridgeNetworkName string) RunningEngineFor
...
 	containerImageTag := version.MinimumSupportedMySQLVersion
 
 	name := "mysql-" + uuid.New().String()
-	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Name:       name,
-		Repository: "mirror.gcr.io/library/mysql",
-		Tag:        containerImageTag,
-		Env:        []string{"MYSQL_ROOT_PASSWORD=secret"},
+
+	req := testcontainers.ContainerRequest{
```
mdelapenya: suggestion: use the [MySQL module](https://golang.testcontainers.org/modules/mysql/), passing your custom image.

internal/testserver/datastore/mysql.go
```diff
@@ -50,59 +50,74 @@ func RunMySQLForTesting(t testing.TB, bridgeNetworkName string) RunningEngineFor
...
 		builder.port = strconv.Itoa(mysqlPort)
 	} else {
-		builder.port = port
+		builder.port = mappedPort.Port()
 	}
 
-	dsn := fmt.Sprintf("%s@(localhost:%s)/mysql?parseTime=true", builder.creds, port)
-	require.NoError(t, pool.Retry(func() error {
-		var err error
+	dsn := fmt.Sprintf("%s@(localhost:%s)/mysql?parseTime=true", builder.creds, mappedPort.Port())
```
mdelapenya: suggestion: the module give you all this for free, connection string, and wait for readiness

internal/testserver/datastore/postgres.go
```diff
@@ -64,47 +65,59 @@ func RunPostgresForTestingWithCommitTimestamps(t testing.TB, bridgeNetworkName s
...
 	}
 
-	postgres, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Name:       postgresContainerHostname,
-		Repository: "mirror.gcr.io/library/postgres",
-		Tag:        pgVersion,
-		Env: []string{
-			"POSTGRES_USER=" + PostgresTestUser,
-			"POSTGRES_PASSWORD=" + PostgresTestPassword,
+	req := testcontainers.ContainerRequest{
```
mdelapenya: suggestion: same for the [Postgres module](https://golang.testcontainers.org/modules/postgres/)

internal/testserver/datastore/postgres.go
```diff
@@ -168,47 +181,60 @@ func (b *postgresTester) NewDatastore(t testing.TB, initFunc InitFunc) datastore
 	return nil
 }
 
-func createNetworkBridge(t testing.TB, pool *dockertest.Pool) string {
+func createNetworkBridge(t testing.TB) (string, *testcontainers.DockerNetwork) {
```
mdelapenya: suggestion: you could expose this in the "framework" and use it everywhere else that you create a network. Adding the CleanupNetwork here would automatically clean up resources with every call.

internal/testserver/datastore/postgres.go
```diff
@@ -168,47 +181,60 @@ func (b *postgresTester) NewDatastore(t testing.TB, initFunc InitFunc) datastore
...
-			"DB_PASSWORD=" + PostgresTestPassword,
-			"DB_HOST=" + b.containerHostname,
-			"DB_PORT=" + b.containerPort,
-			"LISTEN_PORT=" + PgbouncerTestPort,
-			"DB_NAME=*",     // Needed to make pgbouncer okay with the randomly named databases generated by the test suite
-			"AUTH_TYPE=md5", // use the same auth type as postgres
-			"MAX_CLIENT_CONN=" + PostgresTestMaxConnections,
-		},
+	req := testcontainers.ContainerRequest{
+		Name:         pgbouncerContainerHostname,
```
mdelapenya: suggestion: remember the advice against names :)

internal/testserver/datastore/postgres.go
```diff
@@ -168,47 +181,60 @@ func (b *postgresTester) NewDatastore(t testing.TB, initFunc InitFunc) datastore
...
+			"LISTEN_PORT":      PgbouncerTestPort,
+			"DB_NAME":          "*",   // Needed to make pgbouncer okay with the randomly named databases generated by the test suite
+			"AUTH_TYPE":        "md5", // use the same auth type as postgres
+			"MAX_CLIENT_CONN":  PostgresTestMaxConnections,
+		},
+		WaitingFor: wait.ForListeningPort(PgbouncerTestPort + "/tcp").
+			WithStartupTimeout(dockerBootTimeout),
+	}
+
+	if bridgeNetworkName != "" {
```
mdelapenya: suggestion: if moving to the Run function with container customisers as functional options, then you could be passing any of the `WithNetworkXXX` options. See https://golang.testcontainers.org/features/creating_container/#networking-options

internal/testserver/datastore/postgres.go
```diff
@@ -168,47 +181,60 @@ func (b *postgresTester) NewDatastore(t testing.TB, initFunc InitFunc) datastore
...
-			"DB_HOST=" + b.containerHostname,
-			"DB_PORT=" + b.containerPort,
-			"LISTEN_PORT=" + PgbouncerTestPort,
-			"DB_NAME=*",     // Needed to make pgbouncer okay with the randomly named databases generated by the test suite
-			"AUTH_TYPE=md5", // use the same auth type as postgres
-			"MAX_CLIENT_CONN=" + PostgresTestMaxConnections,
-		},
+	req := testcontainers.ContainerRequest{
+		Name:         pgbouncerContainerHostname,
+		Image:        "mirror.gcr.io/edoburu/pgbouncer:latest",
```
mdelapenya: question: is this a postgres db instance? Try using the Postgres module with your pgBouncer image, so you get the Postgres module's APIs. Then you could remove all the wait code below

internal/testserver/datastore/spanner.go
```diff
@@ -29,54 +29,66 @@ type spannerTest struct {
...
 // RunSpannerForTesting returns a RunningEngineForTest for spanner
 func RunSpannerForTesting(t testing.TB, bridgeNetworkName string, targetMigration string) RunningEngineForTest {
-	pool, err := dockertest.NewPool("")
-	require.NoError(t, err)
-
+	ctx := context.Background()
 	name := "spanner-" + uuid.New().String()
-	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
+
+	req := testcontainers.ContainerRequest{
```
mdelapenya: suggestion: we have a module for [Spanner](https://golang.testcontainers.org/modules/gcloud/#spanner) :)

mdelapenya (commented):
This looks really promising! I added some comments regarding best practices when using the library. Feel free to ping me for anything you need, I'm more than happy to help you with this 😊

internal/testserver/datastore/spanner.go
tstirrat15: Yeah, the issue we've found is that the spanner emulator doesn't replicate the behaviors that we need faithfully enough to be of much use.

mdelapenya: Oh that's good to know! Is this something we can add to the upstream module?

cmd/spicedb/integration/restgateway_integration_test.go
```diff
@@ -8,26 +8,27 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/ory/dockertest/v3"
 	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+
+	"github.com/authzed/spicedb/pkg/testutil/sdbtestcontainer"
```
tstirrat15, line 14: this is an internal package for now, but this is probably something that we can upstream to testcontainers.

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,96 +36,85 @@ func TestSchemaWatch(t *testing.T) {
 		}
 
 		t.Run(driverName, func(t *testing.T) {
-			bridgeNetworkName := fmt.Sprintf("bridge-%s", uuid.New().String())
+			ctx := t.Context()
 
-			pool, err := dockertest.NewPool("")
+			// Create an internal network
+			net, err := network.New(ctx)
+			testcontainers.CleanupNetwork(t, net)
```
tstirrat15, line 42: There's several of these throughout these changes. The idea is that rather than using a bridge network to the host, or having test containers loop back through the host to communicate with each other, we create a dedicated docker network for them to talk to each other on (similar to how docker-compose works). It requires some finagling around connection strings, because most of the connection string constructors assume that you want to talk to the container from the host over a mapped port, but that seems preferable to bridge/host network hacks.

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,96 +35,74 @@ func TestSchemaWatch(t *testing.T) {
...
+				network.WithNetwork([]string{driverName}, net))
 
-			engine := testdatastore.RunDatastoreEngineWithBridge(t, driverName, bridgeNetworkName)
+			db := engine.NewDatabase(t)
 
-			envVars := []string{}
-			if wev, ok := engine.(testdatastore.RunningEngineForTestWithEnvVars); ok {
-				envVars = wev.ExternalEnvVars()
-			}
+			connectionVars, err := testutil.InternalConnectionEnvVars(db, driverName)
```
tstirrat15, line 53: This is a helper I added to provide SpiceDB connection vars that point at an internal network. It takes care of the host and port as well as other concerns (like how spanner wants an additional env var for host/port).

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,96 +35,74 @@ func TestSchemaWatch(t *testing.T) {
...
-				Cmd:        []string{"migrate", "head", "--datastore-engine", driverName, "--datastore-conn-uri", db},
-				NetworkID:  bridgeNetworkName,
-				Env:        envVars,
-			}, func(config *docker.HostConfig) {
-				config.RestartPolicy = docker.RestartPolicy{
-					Name: "no",
-				}
-			})
+			migrateContainer, err := testcontainers.Run(ctx, ciImage,
+				network.WithNetwork([]string{"migrate"}, net),
```
tstirrat15, line 58: This is another thing that you'll see throughout: we use `testcontainers.Run` for migration instead of `sdbtestcontainers.Run`, because we don't want the command or waitstrategy that comes from SpiceDB.

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,96 +35,74 @@ func TestSchemaWatch(t *testing.T) {
...
 			require.NoError(t, err)
-			t.Cleanup(func() {
-				_ = pool.Purge(migrateResource)
-			})
+			testcontainers.CleanupContainer(t, migrateContainer)
 
 			// Ensure the command completed successfully.
-			status, err := pool.Client.WaitContainerWithContext(migrateResource.Container.ID, t.Context())
+			containerState, err := migrateContainer.State(ctx)
+			if containerState.ExitCode != 0 {
```
tstirrat15, line 69: There's a few of these blocks - they're mostly for test debugging. It's how I figured out that I needed the rewrites for internal connection strings. I left them in because they're still useful for debugging failures.

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,96 +35,74 @@ func TestSchemaWatch(t *testing.T) {
...
+
+			spicedbEnvVars["SPICEDB_DATASTORE_GC_INTERVAL"] = "1s"
+			spicedbEnvVars["SPICEDB_LOG_LEVEL"] = "trace"
+			spicedbEnvVars["SPICEDB_ENABLE_EXPERIMENTAL_WATCHABLE_SCHEMA_CACHE"] = "true"
 
 			// Run a serve and immediately close, ensuring it shuts down gracefully.
-			serveResource, err := pool.RunWithOptions(&dockertest.RunOptions{
-				Repository: "authzed/spicedb",
-				Tag:        "ci",
-				Cmd:        []string{"serve", "--grpc-preshared-key", "firstkey", "--datastore-engine", driverName, "--datastore-conn-uri", db, "--datastore-gc-interval", "1s", "--telemetry-endpoint", "", "--log-level", "trace", "--enable-experimental-watchable-schema-cache"},
```
tstirrat15, line 84: I generally moved command flags into env vars for container configuration because maps are easier to manipulate than string slices.

cmd/spicedb/integration/schemawatch_integration_test.go
```diff
@@ -33,96 +35,74 @@ func TestSchemaWatch(t *testing.T) {
...
 
 			select {
 			case <-ww.c:
-				break
-
 			case <-time.After(10 * time.Second):
 				require.Fail(t, "timed out waiting for schema watch to run")
 			}
-
-			require.True(t, gracefulShutdown(pool, serveResource))
```
tstirrat15, line 125: Don't need shutdown stuff because testcontainers takes care of it for us

cmd/spicedb/integration/serve_integration_test.go
```diff
@@ -25,51 +26,46 @@ import (
...
-		&dockertest.RunOptions{
-			Repository:   "authzed/spicedb",
-			Tag:          "ci",
-			Cmd:          []string{"serve", "--log-level", "debug", "--grpc-preshared-key", "firstkey", "--grpc-preshared-key", "secondkey"},
-			ExposedPorts: []string{"50051/tcp"},
-		},
-		"firstkey",
-		false,
+	container, err := sdbtestcontainer.Run(t.Context(), sdbtestcontainer.DefaultImageReference,
+		sdbtestcontainer.WithBootstrapSchema(defaultSchema),
```
tstirrat15, line 36: This seems like it'll be useful to downstream consumers eventually

cmd/spicedb/integration/serve_integration_test.go
```diff
@@ -25,51 +26,46 @@ import (
...
-			conn, err := grpc.NewClient(fmt.Sprintf("localhost:%s", tester.port), opts...)
+			conn, err := grpc.NewClient(container.GRPCEndpoint(), opts...)
 
 			require.NoError(err)
 			t.Cleanup(func() {
 				_ = conn.Close()
 			})
 
-			require.Eventually(func() bool {
+			require.EventuallyWithT(func(collect *assert.CollectT) {
```
tstirrat15, line 62: Made a few of these refactors to get better test output.

cmd/spicedb/integration/serve_integration_test.go
```diff
@@ -93,61 +89,22 @@ func TestServe(t *testing.T) {
...
-	case <-closed:
-		return true
-
-	case <-time.After(10 * time.Second):
-		_ = pool.Purge(serveResource)
-		return false
-	}
-}
-
-func TestGracefulShutdownInMemory(t *testing.T) {
```
tstirrat15, line 118: Agreed with other reviewer that this is testing docker behavior, not binary bheavior.

cmd/spicedb/integration/servetesting_integration_test.go
```diff
@@ -23,52 +23,77 @@ import (
...
 	require.NoError(err)
-	defer conn.Close()
+	t.Cleanup(func() {
+		_ = conn.Close()
+	})
 
-	roConn, err := grpc.NewClient(fmt.Sprintf("localhost:%s", tester.readonlyPort), options...)
+	containerHost, err := container.Host(t.Context())
+	require.NoError(err)
+	roMapped, err := container.MappedPort(t.Context(), "50052/tcp")
```
tstirrat15, line 73: This is an example of some faffing we have to do to get access to a port that the container doesn't already expose - there's a few places where we have to do this.

cmd/spicedb/integration/servetesting_integration_test.go
```diff
@@ -160,119 +187,23 @@ func TestTestServer(t *testing.T) {
 	require.Contains(string(body), "schemaText")
 	require.Contains(string(body), "definition resource")
 
-	/*
-		 * TODO(jschorr): Re-enable once we figure out why this makes the test flaky
```
tstirrat15, line 164: I re-enabled this and it seems to be fine.

cmd/spicedb/integration/servetesting_integration_test.go
```diff
@@ -160,119 +187,23 @@ func TestTestServer(t *testing.T) {
...
-	HTTPPort         string
-	readonlyHTTPPort string
-}
-
-const retryCount = 8
-
-// newTester spins up a SpiceDB server running against a specific datastore with a specific access token.
-// It also writes or reads a schema.
-// On test termination it cleans up all resources.
-func newTester(t *testing.T, containerOpts *dockertest.RunOptions, token string, withExistingSchema bool) (*spicedbHandle, error) {
```
tstirrat15, line 191: All of this is now encompassed by the sdbtestcontainer module

cmd/spicedb/integration/servetesting_race_test.go
```diff
@@ -3,54 +3,45 @@
...
 	"google.golang.org/grpc/credentials/insecure"
-	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+
+	"github.com/authzed/spicedb/pkg/testutil/sdbtestcontainer"
 )
 
+//go:embed testdata/bootstrap.yaml
+var bootstrapContents string
```
tstirrat15, line 21: Using a reader over an embed means that we don't have to faff with figuring out the filepath to the module during the test.

cmd/spicedb/integration/servetesting_race_test.go
```diff
@@ -3,54 +3,45 @@
...
-	for i := 0; i < 1000; i++ {
-		conn, err := grpc.NewClient(fmt.Sprintf("localhost:%s", tester.port),
+	for i := range 1000 {
+		conn, err := grpc.NewClient(container.GRPCEndpoint(),
 			grpc.WithTransportCredentials(insecure.NewCredentials()))
 		require.NoError(t, err)
 
-		require.Eventually(t, func() bool {
-			resp, err := healthpb.NewHealthClient(conn).Check(t.Context(), &healthpb.HealthCheckRequest{Service: "authzed.api.v1.SchemaService"})
-			if err != nil || resp.GetStatus() != healthpb.HealthCheckResponse_SERVING {
```
tstirrat15, line 47: This is covered by the spicedb waits, because it's already connecting to the container to write the schema.

internal/datastore/crdb/crdb_test.go
```diff
@@ -198,18 +192,19 @@ func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
 
 var defaultKeyForTesting = proxy.KeyConfig{
 	ID: "defaultfortest",
-	Bytes: (func() []byte {
+	Bytes: func() []byte {
```
tstirrat15, line 195: Format change

internal/datastore/crdb/crdb_test.go
```diff
@@ -148,7 +142,7 @@ func TestCRDBDatastoreWithFollowerReads(t *testing.T) {
 	followerReadDelay := time.Duration(4.8 * float64(time.Second))
 	gcWindow := 100 * time.Second
 
-	engine := testdatastore.RunCRDBForTesting(t, "", crdbTestVersion())
```
tstirrat15, line 151: The string arg was `bridgeNetworkName`; we're handling networking differently so we drop it.

internal/datastore/crdb/crdb_test.go
```diff
@@ -310,12 +304,9 @@ func TestWatchFeatureDetection(t *testing.T) {
 		},
 	}
 	for _, tt := range cases {
-		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
-			ctx, cancel := context.WithCancel(t.Context())
-			t.Cleanup(cancel)
```
tstirrat15, line 316: wat

internal/datastore/crdb/crdb_test.go
```diff
@@ -358,156 +349,40 @@ const (
 	unprivileged provisionedUser = "unprivileged"
 )
 
-func newCRDBWithUser(t *testing.T, pool *dockertest.Pool) (adminConn *pgx.Conn, connStrings map[provisionedUser]string) {
-	// in order to create users, cockroach must be running with
-	// real certs, and the root user must be authenticated with
-	// client certs.
-	certDir := t.TempDir()
-
-	ca := &x509.Certificate{
```
tstirrat15, line 367: The CRDB testcontainer already does all of this

internal/datastore/crdb/reader.go
```diff
@@ -339,7 +339,7 @@ func (cr crdbReader) loadNamespace(ctx context.Context, tx pgxcommon.DBFuncQueri
 }
 
 func (cr crdbReader) lookupNamespaces(ctx context.Context, tx pgxcommon.DBFuncQuerier, nsNames []string) ([]datastore.RevisionedNamespace, error) {
-	clause := sq.Or{}
+	clause := make(sq.Or, 0, len(nsNames))
```
tstirrat15, line 342: Drive-by to silence a prealloc complaint

internal/datastore/postgres/common/cancelation.go
```diff
@@ -1,18 +0,0 @@
...
-import (
-	"time"
-
-	"github.com/jackc/pgx/v5/pgconn"
-	"github.com/jackc/pgx/v5/pgconn/ctxwatch"
-)
-
-// CancelationContextHandler returns a context watcher handler for canceling requests
-// when a context is canceled, rather than closing connections.
-func CancelationContextHandler(pgConn *pgconn.PgConn) ctxwatch.Handler {
```
tstirrat15, line 12: This was identified as dead code

internal/datastore/postgres/postgres.go
```diff
@@ -674,6 +675,31 @@ func errorRetryable(err error) bool {
 	return false
 }
 
+// readErrorRetryable returns whether an error from an idempotent, read-only query may
+// be retried. Unlike errorRetryable, it also covers connection-level failures such as
+// i/o timeouts: those surface when a pooled connection is torn down underneath the
+// query, and pgx discards the connection afterward, so a retry acquires a fresh one.
+// It never retries once the calling context itself is done.
+func readErrorRetryable(ctx context.Context, err error) bool {
```
tstirrat15: This is net-new - there were test flakes because of net errors under the pgx driver. We ought to be able to retry those transparently rather than bubbling the error.

internal/datastore/postgres/postgres_shared_test.go
```diff
@@ -71,11 +73,11 @@ func testPostgresDatastore(t *testing.T, config postgresTestConfig) {
 	}
 
 	t.Run(fmt.Sprintf("%spostgres-%s-%s-%s-gc", pgbouncerStr, config.pgVersion, config.targetMigration, config.migrationPhase), func(t *testing.T) {
-		b := testdatastore.RunPostgresForTesting(t, "", config.targetMigration, config.pgVersion, config.pgbouncer)
+		b := testdatastore.RunPostgresForTesting(t, config.targetMigration, config.pgVersion, config.pgbouncer)
 		ctx := t.Context()
 
 		// NOTE: gc tests take exclusive locks, so they are run under non-parallel.
-		test.OnlyGCTests(t, test.DatastoreTesterFunc(func(_ testing.TB, revisionQuantization, gcInterval, gcWindow time.Duration, watchBufferLength uint16) (datastore.Datastore, error) {
+		test.OnlyGCTests(t, test.DatastoreTesterFunc(func(t testing.TB, revisionQuantization, gcInterval, gcWindow time.Duration, watchBufferLength uint16) (datastore.Datastore, error) {
```
tstirrat15, line 80: Not using the nested `t` was screwing up test reporting when these tests failed.

internal/datastore/postgres/revisions.go
```diff
@@ -128,13 +127,31 @@ const (
...
 )
 
+// maxTransientReadRetries bounds the retries performed by queryRowWithRetries.
+const maxTransientReadRetries = 3
+
+// queryRowWithRetries runs a single-row query on the read pool, retrying a bounded
+// number of times when the failure is a transient connection error (e.g. a pooled
+// connection torn down underneath the query). Callers must only pass idempotent,
+// read-only queries.
+func (pgd *pgDatastore) queryRowWithRetries(ctx context.Context, sql string, scanFn func(row pgx.Row) error) error {
```
tstirrat15: I'd also appreciate eyes on this.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -549,22 +544,12 @@ definition document {
 			runEndToEndTest(t, tc)
 		})
 	}
-
-	t.Run("lookupresources cursor pagination", func(t *testing.T) {
-		runLookupResourcesCursorPaginationTest(t)
-	})
```
tstirrat15, line 555: I didn't see a reason for these to be nested here - I renamed the referenced functions to turn them into test functions.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -690,14 +672,9 @@ definition document {
...
-	pgconn := runPostgres(t)
+	pgconn := runPostgres(t, pgServerPort)
 
-	_, filename, _, _ := runtime.Caller(0)
-	dat, err := os.ReadFile(path.Join(path.Dir(filename), "../../configuration.sql"))
-	require.NoError(t, err)
-	createCommands := strings.ReplaceAll(string(dat), "port '5433'", fmt.Sprintf("port '%d'", pgServerPort))
-
-	_, err = pgconn.Exec(t.Context(), createCommands)
+	_, err = pgconn.Exec(t.Context(), readCreateCommands(t, pgServerPort))
```
tstirrat15, line 677: `readCreateCommands` uses the same embed trick as before to avoid needing to mess with reading files under indirection.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -881,80 +853,73 @@ func runEndToEndTest(t *testing.T, tc e2eTestCase) {
...
 	// Invoke initial Postgres commands.
-	_, err = pgconn.Exec(t.Context(), createCommands)
+	_, err = pgconn.Exec(t.Context(), readCreateCommands(t, pgServerPort))
 	require.NoError(t, err)
 	t.Log("Initial Postgres commands executed")
 
 	// Run queries.
 	var lastZedToken string
 	for _, q := range tc.queries {
-		t.Run(q.name, func(t *testing.T) {
```
tstirrat15, line 904: This nesting didn't seem to do anything useful for us.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -881,80 +853,73 @@ func runEndToEndTest(t *testing.T, tc e2eTestCase) {
...
+			if len(q.expectedRows) > 0 {
+				require.Equal(t, q.expectedRows, foundRows)
+			} else if q.query != queryToRun {
+				// For consistency tests that inject parameters, we just verify we got results
+				require.NotEmpty(t, foundRows)
+			}
+			require.Equal(t, q.expectedResponseTag, rows.CommandTag().String())
+		}
+		// Close the query for this query in the loop
+		rows.Close()
```
tstirrat15, line 922: Needed to move this because of that nesting change - there wasn't a defer boundary.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -1063,52 +1029,46 @@ func runSpiceDB(t *testing.T) *authzed.Client {
...
-	pool, err := dockertest.NewPool("")
+func runPostgres(t *testing.T, pgServerPort int) (conn *pgx.Conn) {
+	t.Helper()
+	t.Log("postgres fdw server port: ", pgServerPort)
+	// TODO: use alpine?
+	logger := log.TestLogger(t)
+	image := fmt.Sprintf("mirror.gcr.io/library/postgres:%s", pgVersion)
+	container, err := postgres.Run(t.Context(), image,
+		testcontainers.WithLogger(logger),
+		testcontainers.WithHostPortAccess(pgServerPort),
```
tstirrat15, line 1040: This is how it's possible for the postgres server running in the container to reach out to SpiceDB running on the host - testcontainers gives us an abstraction that uses a tunnel sidecar container to make all of the hostnames line up.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -1063,52 +1029,46 @@ func runSpiceDB(t *testing.T) *authzed.Client {
...
-		if err != nil {
-			return err
-		}
-		return nil
-	})
+// readCreateCommands takes the `configuration.sql` defined in the root of this directory
+// and templates in the variables necessary to run in an integration test context
+func readCreateCommands(t testing.TB, pgServerPort int) string {
+	_, filename, _, _ := runtime.Caller(0)
+	dat, err := os.ReadFile(path.Join(path.Dir(filename), "../../configuration.sql"))
```
tstirrat15, line 1064: Nvm it's not embed - I just abstracted it into a function. The issue with using embed is that you can't embed a file outside of the file's directory for security reasons, and that configuration file is defined in the root of the repository. We might be able to revisit that decision.

internal/testserver/datastore/crdb.go
```diff
@@ -1,119 +1,85 @@
...
+	"github.com/testcontainers/testcontainers-go/modules/cockroachdb"
 
 	crdbmigrations "github.com/authzed/spicedb/internal/datastore/crdb/migrations"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/migrate"
 	"github.com/authzed/spicedb/pkg/secrets"
 )
 
-const (
-	enableRangefeeds = `SET CLUSTER SETTING kv.rangefeed.enabled = true;`
```
tstirrat15, line 22: I inlined this one.

internal/testserver/datastore/crdb.go
```diff
@@ -1,119 +1,85 @@
...
+	// endpoint is the host:port of the cockroach container.
+	endpoint string
 }
 
 var _ RunningEngineForTest = (*crdbTester)(nil)
 
 // RunCRDBForTesting returns a RunningEngineForTest for CRDB
-func RunCRDBForTesting(t testing.TB, bridgeNetworkName string, crdbVersion string) *crdbTester {
-	pool, err := dockertest.NewPool("")
+func RunCRDBForTesting(t testing.TB, crdbVersion string, opts ...testcontainers.ContainerCustomizer) *crdbTester {
```
tstirrat15, line 28: You'll see this change on all of the testers - I added `opts` here so that calling code can pass in things like a network, which means that we don't have to completely refactor this code in order to make it possible for containers to talk to datastore containers.

internal/testserver/datastore/datastore.go
```diff
@@ -3,18 +3,15 @@ package datastore
...
-	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
 
 	crdbversion "github.com/authzed/spicedb/internal/datastore/crdb/version"
 	pgversion "github.com/authzed/spicedb/internal/datastore/postgres/version"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/migrate"
 )
 
-const dockerBootTimeout = 10 * time.Second
```
tstirrat15, line 16: This is handled by `testcontainers`

internal/testserver/datastore/datastore.go
```diff
@@ -33,48 +30,31 @@ type RunningEngineForTest interface {
 	NewDatastore(t testing.TB, initFunc InitFunc) datastore.Datastore
 }
 
-// RunningEngineForTestWithEnvVars represents an instance of a datastore engine running, that also
-// requires env vars set to use from an external source (like a container).
-type RunningEngineForTestWithEnvVars interface {
-	RunningEngineForTest
-
-	// ExternalEnvVars are the environment variables to be set.
-	ExternalEnvVars() []string
```
tstirrat15, line 42: This was entirely for dealing with the env vars that are needed for spanner. That's now a special case in the function that provides env vars, so we don't need this interface.

internal/testserver/datastore/datastore.go
```diff
@@ -33,48 +30,31 @@ type RunningEngineForTest interface {
...
 // the Cleanup of the testing object.
-func RunDatastoreEngine(t testing.TB, engine string) RunningEngineForTest {
-	return RunDatastoreEngineWithBridge(t, engine, "")
-}
-
-// RunDatastoreEngineWithBridge runs a datastore engine on a specific bridge. If a bridge is
-// specified, then the hostnames returned by the engines are those to be called from another
-// container on the bridge.
-func RunDatastoreEngineWithBridge(t testing.TB, engine string, bridgeNetworkName string) RunningEngineForTest {
+func RunDatastoreEngine(t testing.TB, engine string, opts ...testcontainers.ContainerCustomizer) RunningEngineForTest {
```
tstirrat15, line 38: We get to drop the bridge part because we're using native networks when and where we need them.

internal/testserver/datastore/datastore.go
```diff
@@ -33,48 +30,31 @@ type RunningEngineForTest interface {
...
 	case "memory":
-		require.Empty(t, bridgeNetworkName, "memory datastore does not support bridge networking")
 		return RunMemoryForTesting(t)
 	case "cockroachdb":
 		ver := os.Getenv("CRDB_TEST_VERSION")
 		if ver == "" {
 			ver = crdbversion.LatestTestedCockroachDBVersion
 		}
-		return RunCRDBForTesting(t, bridgeNetworkName, ver)
+		return RunCRDBForTesting(t, ver, opts...)
```
tstirrat15, line 47: Note the passthrough here

internal/testserver/datastore/postgres.go
```diff
@@ -19,247 +26,268 @@ import (
...
-	useContainerHostname bool
+	targetMigration string
+	pgbouncerProxy  *postgres.PostgresContainer
+	pgContainer     *postgres.PostgresContainer
 }
 
 // RunPostgresForTesting returns a RunningEngineForTest for postgres
-func RunPostgresForTesting(t testing.TB, bridgeNetworkName string, targetMigration string, pgVersion string, enablePgbouncer bool) RunningEngineForTest {
-	return RunPostgresForTestingWithCommitTimestamps(t, bridgeNetworkName, targetMigration, true, pgVersion, enablePgbouncer)
+func RunPostgresForTesting(t testing.TB, targetMigration string, pgVersion string, enablePgbouncer bool, opts ...testcontainers.ContainerCustomizer) RunningEngineForTest {
```
tstirrat15, line 42: It might eventually be worth splitting out the pg and pgbouncer variants of this - the datastore interface is the same, but setting up pgbouncer is significantly different

internal/testserver/datastore/postgres.go
```diff
@@ -19,247 +26,264 @@ import (
...
+		postgresOptions...,
+	)
+	require.NoError(t, err)
+	testcontainers.CleanupContainer(t, pgContainer)
+	b.pgContainer = pgContainer
+
+	// set up the bouncer container
+	// NOTE: this is a bit of a bodge; a pgbouncer container is not the same as a postgres container,
+	// so we have to undo some of what the postgres container wrapper is doing.
+	bouncerContainer, err := postgres.Run(ctx, "mirror.gcr.io/edoburu/pgbouncer:latest",
```
tstirrat15, line 200: We use `postgres` here instead of a generic `testcontainer` so that we get the same behavior around `ConnectionString` etc

internal/testserver/cluster.go
```diff
@@ -74,8 +74,7 @@ func init() {
 // SafeManualResolverBuilder is a resolver builder that builds SafeManualResolvers
 // it is similar to manual.Resolver in grpc, but is thread safe
 type SafeManualResolverBuilder struct {
-	resolvers sync.Map
```
tstirrat15, line 77: This was identified as something that doesn't really help with the `NewClient` approach of setting up gRPC connections. Everything seems to work without it.

pkg/cmd/migrate_test.go
```diff
@@ -178,45 +180,39 @@ func TestExecuteMigrateWithNoDataSucceeds(t *testing.T) {
...
 		if engineKey == "memory" {
 			continue
 		}
 
+		ctx := t.Context()
+
+		// Create an internal network
+		net, err := network.New(ctx)
+		testcontainers.CleanupNetwork(t, net)
+		require.NoError(t, err)
```
tstirrat15, line 193: It's a little annoying that we have to do this same setup stuff as in the `cmd/spicedb/integration` files, as well as that it's an integration test that uses docker, which seems strange.

pkg/testutil/sdbtestcontainer/options.go
tstirrat15, line 1: This is what I'd like to upstream, although not now, because I don't want to block this PR.

pkg/testutil/internal_container_connnection.go
```diff
@@ -0,0 +1,66 @@
...
+		dbURL.Host = fmt.Sprintf("%s:%s", driverName, defaultPort)
+	}
+	return dbURL.String(), nil
+}
+
+// InternalConnectionEnvVars takes a db connection URI and a database driver name
+// and returns the environment variables necessary for a SpiceDB pod to talk to a container
+// running that datastore over the internal docker network. this is necessary because otherwise
+// the db URI makes the assumption that you're connecting on the host.
+func InternalConnectionEnvVars(dbConnURI, driverName string) (map[string]string, error) {
```
tstirrat15, line 44: This is what gives us connection vars for a SpiceDB container. I'm open to different ways of organizing this code; I don't love that `driverName` is something that we're passing around like this. Maybe a type alias would help?

Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build --mount=type=cache,target=/g
 FROM cgr.dev/chainguard/static@sha256:77d8b8925dc27970ec2f48243f44c7a260d52c49cd778288e4ee97566e0cb75b
 # NOTE: the copy target location differs from Dockerfile.release for historical reasons. It's referenced in
 # compose files and elsewhere so we're keeping it the way it is.
-COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.52 /ko-app/grpc-health-probe /bin/grpc_health_probe
+COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.53 /ko-app/grpc-health-probe /bin/grpc_health_probe
```
tstirrat15, line 13: marked CVE

go.mod
```diff
@@ -69,12 +66,12 @@ require (
 	github.com/lithammer/fuzzysearch v1.1.8
 	github.com/mattn/go-isatty v0.0.22
 	github.com/maypok86/otter/v2 v2.3.0
+	github.com/moby/moby/api v1.54.2
 	github.com/mostynb/go-grpc-compression v1.2.3
 	github.com/muesli/mango-cobra v1.3.0
 	github.com/muesli/roff v0.1.0
 	github.com/ngrok/sqlmw v0.0.0-20220520173518-97c9c04efc79
 	github.com/odigos-io/go-rtml v0.0.1
-	github.com/ory/dockertest/v3 v3.12.0
```
tstirrat15, line 77: Finally

tstirrat15 (commented):
See comments

pkg/consistency/test/consistency.go
```diff
@@ -124,7 +124,6 @@ func ConsistencyForEngine(t *testing.T, engineID string, tester dstest.Datastore
 
 			for _, chunkSize := range []uint16{5, 10} {
 				t.Run(fmt.Sprintf("chunk-size-%d", chunkSize), func(t *testing.T) {
-					t.Parallel()
```
tstirrat15, line 127: Reducing parallelism to (hopefully) reduce postgres flakes here

tstirrat15 (commented):
See comment

.github/workflows/build-test.yaml
```diff
@@ -226,7 +226,7 @@ jobs:
 
   pgdatastoreconstest:
     name: "Datastore Consistency Tests"
-    runs-on: "depot-ubuntu-24.04-arm-4"
+    runs-on: "depot-ubuntu-24.04-arm-8"
```
tstirrat15, line 229: Postgres consistency tests are failing with errors that are consistent with resource pressure; adding some more resources might help.

tstirrat15: It appears to have worked 😎

pkg/testutil/sdbtestcontainer/options.go
miparnisari: let's wait a bit and stabilize this code before moving it to a separate repo

pkg/testutil/sdbtestcontainer/spicedb.go
```diff
@@ -0,0 +1,241 @@
...
+// from the configuration. Every SpiceDB flag is bound to an env var named
+// SPICEDB_ + the flag in upper snake case (e.g. --grpc-preshared-key maps to
+// SPICEDB_GRPC_PRESHARED_KEY).
+func buildEnv(cfg *options) map[string]string {
+	env := map[string]string{
+		"SPICEDB_GRPC_PRESHARED_KEY": strings.Join(cfg.presharedKeys, ","),
+		"SPICEDB_DATASTORE_ENGINE":   cfg.datastoreEngine,
+		"SPICEDB_LOG_LEVEL":          cfg.logLevel,
+		// Avoid phoning home from tests.
+		"SPICEDB_TELEMETRY_ENDPOINT": "",
```
miparnisari, line 178: nit: should we turn on request & response logging?

pkg/testutil/sdbtestcontainer/spicedb.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	if cfg.httpEnabled {
+		env["SPICEDB_HTTP_ENABLED"] = "true"
+		env["SPICEDB_HTTP_ADDR"] = ":8443"
+	}
+	return env
+}
+
+// bootstrap writes the configured schema and relationships over the gRPC API.
+func (c *Container) bootstrap(ctx context.Context, cfg *options) error {
```
miparnisari, line 191: another option is to accept a new env var `--datastore-bootstrap-files` and let our code handle this

internal/testserver/datastore/spanner.go
```diff
@@ -29,44 +28,39 @@ type spannerTest struct {
 }
 
 // RunSpannerForTesting returns a RunningEngineForTest for spanner
-func RunSpannerForTesting(t testing.TB, bridgeNetworkName string, targetMigration string) RunningEngineForTest {
-	pool, err := dockertest.NewPool("")
+func RunSpannerForTesting(t testing.TB, targetMigration string, opts ...testcontainers.ContainerCustomizer) RunningEngineForTest {
```
miparnisari, line 31: nit: `targetMigration` is always `head` :)

internal/testserver/datastore/spanner.go
```diff
@@ -29,44 +28,39 @@ type spannerTest struct {
...
+	ctx := t.Context()
+
+	options := make([]testcontainers.ContainerCustomizer, 0, len(opts)+2)
+	options = append(options,
+		testcontainers.WithExposedPorts("9010/tcp"),
+		testcontainers.WithWaitStrategy(wait.ForListeningPort("9010/tcp").WithStartupTimeout(time.Minute)),
+	)
+	options = append(options, opts...)
+
+	container, err := testcontainers.Run(ctx, "gcr.io/cloud-spanner-emulator/emulator:1.5.41",
```
miparnisari, line 41: not `latest`?

also i'm curious, does this attempt to pull down the image _every single time this function is invoked_ or is it smart enough to cache it?

internal/testserver/datastore/spanner.go
```diff
@@ -76,27 +70,17 @@ func RunSpannerForTesting(t testing.TB, bridgeNetworkName string, targetMigratio
 				NodeCount:   1,
 			},
 		})
-		return err
-	}))
+		return err == nil
+	}, time.Minute, 500*time.Millisecond)
 
 	builder := &spannerTest{
+		hostname:        host,
```
miparnisari, line 77: nit: these two fields are written but never read. I think you can remove them and set `instancesClient`, which is used by `NewDatabase`. (notice how right now `NewDatabase` isn't using any field of `b`)

internal/testserver/datastore/config/postgres-with-timestamps.conf
```diff
@@ -0,0 +1,3 @@
+listen_addresses = '*'
+max_connections = 3000
+track_commit_timestamp = 1
```
miparnisari, line 3: extensions=pg_hint_plan? :D

internal/datastore/postgres/postgres_shared_test.go
```diff
@@ -1497,10 +1504,49 @@ func WatchNotEnabledTest(t *testing.T, _ testdatastore.RunningEngineForTest, pgV
 	require.Contains(err.Error(), "track_commit_timestamp=on")
 }
 
+func BenchmarkPostgresQuery(b *testing.B) {
```
miparnisari, line 1507: this is new?

cmd/spicedb/integration/servetesting_race_test.go
```diff
@@ -3,54 +3,45 @@
...
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+
+	"github.com/authzed/spicedb/pkg/testutil/sdbtestcontainer"
 )
 
+//go:embed testdata/bootstrap.yaml
+var bootstrapContents string
+
 // Based on a test originally written by https://github.com/wscalf
 func TestCheckPermissionOnTesterNoFlakes(t *testing.T) {
```
miparnisari, line 24: this is a really weird test, i wonder what the original issue was

internal/testserver/datastore/postgres.go
```diff
@@ -19,247 +26,264 @@ import (
...
+		builder.runPostgresForTesting(t, pgVersion, withCommitTimestamps, opts...)
 	}
 
 	return builder
 }
 
 func (b *postgresTester) NewDatabase(t testing.TB) string {
-	uniquePortion, err := secrets.TokenHex(4)
+	t.Helper()
+	uri, err := b.newDatabase(t.Context())
```
miparnisari, line 65: nit: inline `newDatabase`. `NewDatabase` isn't doing anything

miparnisari (commented):
LGTM! 🚀 

None of my comments are blocking but it would be nice to address them soon. 

Important followups
- opened https://github.com/authzed/spicedb/issues/3224
- let's wait a bit and stabilize our testcontainer before moving it to a separate repo
- can we turn on `pg_hint_plan` by default for the postgres container?

pkg/testutil/internal_container_connnection.go
miparnisari: i don't see what's bad about the `driverName string`?

internal/testserver/datastore/spanner.go
tstirrat15: The docker engine on the thing running the commands should cache it

internal/datastore/postgres/postgres_shared_test.go
tstirrat15: I thought you added it :thinking: I can remove otherwise

internal/testserver/datastore/postgres.go
tstirrat15: It's used in other places, and `NewDatabase` satisfies the interface

cmd/spicedb/integration/servetesting_race_test.go
tstirrat15: I went and dug it up: https://github.com/authzed/spicedb/pull/1263

pkg/testutil/sdbtestcontainer/spicedb.go
tstirrat15: I think it'd make sense to provide an external option for that but not do it by default

internal/testserver/datastore/spanner.go
tstirrat15: Huh, also true of postgres

tstirrat15 (approved):
Approving since Maria opened the PR

PR #2785, opened 2025-12-19
feat: Add support for `self` keyword in schema for referencing a resource as a subject

### LINKED ISSUE #338: Support for `this` to allow checks on resources with themselves as the subject
There are some certain scenarios where we want to check a subject against itself as the resource, for example:

```
definition user {
  relation manager: user
  permission view_paystubs = manager + <the user themselves>
}
```

We would want to have checks for this pass:
`check(user:user_id, view_paystubs, user:user_id)`

We should discuss adding some kind of keyword to the schema language which allows a permission to refer to the subject itself without an explicit relationship.

As a workaround, for now, we can write an extra relationship between a resource and itself.

```
definition user {
  relation manager: user
  relation myself: user
  permission view_paystubs = manager + myself
}
```

`user:user_id#myself@user:user_id`

tstirrat15:
Reverts #1515
Fixes #338 
Continues #1511

## Description
The reason we reverted #1511 is that we'd break schemas where folks were using `self` as an actual relation name. We now have a mechanism to work around that, which is `use` directives. This updates the code from #1511 to introduce a `use self` directive which enables this feature at the schema level, disallows the use of `self` as a normal relation name, and then compiles accordingly.

## TODO
- [x] Figure out where this code moved: https://github.com/authzed/spicedb/pull/1511/changes#diff-b443ad8517c2657f0460e20d33e05f0f0f2a746c059f0e7db67f79f6ba16af7b
- [x] Implement parser flag
- [x] Ensure it works with composable schemas
- [x] Update the integration test
- [x] Move the consistency test out of the consistency test framework so that we can get this through

## Changes
Will annotate. See commits after the first one for the changes required for the `use` directive.

## Testing
Review. See that tests pass.

### DIFF
--- a/pkg/query/self.go
+++ b/pkg/query/self.go
@@ -10,7 +10,10 @@ import (
 // Resource in the subiterator that connects it to
 // streamed from the sub-iterator to a specified alias relation.
 type Self struct {
-	id       string
+	id string
+	// relation is the name of the permission that uses the self keyword,
+	// and is used because otherwise "self" doesn't have an associated relation,
+	// as it's purely checking whether the subject and object are the same.
 	relation string
 }
 
@@ -27,8 +30,11 @@ func (s *Self) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRela
 	// for Self, the check returns the reflexive relation if it's in the set of resources
 	// or else returns an empty set.
 	for _, resource := range resources {
-		// TODO: does the subject relation need to be considered here?
-		if resource.Equals(GetObject(subject)) {
+		// NOTE: we assert that the subject and the object are the same and that
+		// both are ellipses because self is checking whether the subject and the object
+		// are the same, and a subject with a non-ellipsis relation can't semantically
+		// match the object.
+		if subject.Equals(resource.WithEllipses()) {
 			return func(yield func(Path, error) bool) {
 				yield(Path{
 					Resource: resource,
@@ -47,7 +53,7 @@ func (s *Self) IterSubjectsImpl(ctx *Context, resource Object) (PathSeq, error) 
 		yield(Path{
 			Resource: resource,
 			Relation: s.relation,
-			// TODO: is this correct?
+			// NOTE: this is WithEllipses() for the same reason as the comment above.
 			Subject: resource.WithEllipses(),
 		}, nil)
 	}, nil
--- a/pkg/tuple/structs.go
+++ b/pkg/tuple/structs.go
@@ -60,6 +60,10 @@ func (onr ObjectAndRelation) String() string {
 	return onr.ObjectType + ":" + onr.ObjectID + "#" + onr.Relation
 }
 
+func (onr ObjectAndRelation) Equals(other ObjectAndRelation) bool {
+	return onr.ObjectType == other.ObjectType && onr.ObjectID == other.ObjectID && onr.Relation == other.Relation
+}
+
 // RelationshipReference represents a reference to a relationship, i.e. those portions
 // of a relationship that are not the integrity or caveat and thus form the unique
 // identifier of the relationship.


internal/graph/lookupsubjects.go
```diff
@@ -555,6 +555,28 @@ func (cl *ConcurrentLookupSubjects) lookupSetOperation(
 			// Purposely do nothing.
 			continue
 
+		case *core.SetOperation_Child_XSelf:
+			// Return the current resource(s) as the found subject(s).
+			foundResources := map[string]*v1.FoundSubjects{}
```
miparnisari: we need to add consistency tests to exercise this

proto/internal/core/v1/core.proto
```diff
@@ -463,6 +473,7 @@ message SetOperation {
       UsersetRewrite userset_rewrite = 4 [(validate.rules).message.required = true];
       FunctionedTupleToUserset functioned_tuple_to_userset = 8 [(validate.rules).message.required = true];
       Nil _nil = 6;
+      Self _self = 9;
```
miparnisari: why are the numbers not in consecutive order?

internal/graph/lookupresources2.go
```diff
@@ -154,6 +154,22 @@ func (crr *CursoredLookupResources2) afterSameType(
 			defer span.End()
 
 			switch entrypoint.EntrypointKind() {
+			case core.ReachabilityEntrypoint_SELF_ENTRYPOINT:
+				containingRelation := entrypoint.ContainingRelationOrPermission()
```
tstirrat15, line 158: I don't completely understand this part of the implementation. Where is the self logic here? Or is it actually implemented elsewhere?

internal/services/integrationtesting/testconfigs/self.yaml
```diff
@@ -0,0 +1,24 @@
+---
+schema: >-
+  use self
+  definition user {
+    relation related: user
+    permission me_or_related = related + self
+  }
+
+  definition resource {
+    relation viewer: user#me_or_related | user
```
tstirrat15: Should we refactor this to use an arrow, or is there something important about this being a userset?

pkg/composableschemadsl/compiler/compiler_test.go
```diff
@@ -1299,6 +1299,118 @@ func TestCompile(t *testing.T) {
 				),
 			},
 		},
+		{
+			"self is allowed when used in arrow and interpreted as relation name",
+			withTenantPrefix,
+			`definition expressioned {
```
tstirrat15, line 1305: Why don't these need the use flags?

pkg/composableschemadsl/lexer/flaggablelexer.go
```diff
@@ -26,7 +26,7 @@ func (l *FlaggableLexer) NextToken() Lexeme {
 	nextToken := l.lex.nextToken()
 
 	// Look for `use somefeature`
-	if nextToken.Kind == TokenTypeIdentifier {
+	if nextToken.Kind == TokenTypeIdentifier || nextToken.Kind == TokenTypeKeyword {
```
tstirrat15, line 29: Why was this necessary? Is it because we've already reserved `self`?

We'll probably refactor this when we go and collapse composableschemadsl and schemadsl.

pkg/composableschemadsl/lexer/flags.go
```diff
@@ -21,6 +25,16 @@ var Flags = map[string]transformer{
 			return lexeme, true
 		}
 
+		return lexeme, false
+	},
+	// TODO add typechecking flag?
```
tstirrat15, line 30: context?

pkg/composableschemadsl/parser/parser.go
```diff
@@ -705,6 +709,17 @@ func (p *sourceParser) tryConsumeNilExpression() (AstNode, bool) {
 	return node, true
 }
 
+func (p *sourceParser) tryConsumeSelfExpression() (AstNode, bool) {
+	if !p.isKeyword("self") {
+		return nil, false
```
tstirrat15, line 714: This seems like it's not necessary based on how this function is invoked. Is it worth keeping just so that the caller doesn't have to know that?

pkg/composableschemadsl/parser/parser.go
```diff
@@ -705,6 +709,17 @@ func (p *sourceParser) tryConsumeNilExpression() (AstNode, bool) {
...
+	}
+
+	node := p.startNode(dslshape.NodeTypeSelfExpression)
+	p.consumeKeyword("self")
+	defer p.mustFinishNode()
```
tstirrat15, line 719: Nit, move up

pkg/query/build_tree.go
```diff
@@ -172,6 +172,8 @@ func (b *iteratorBuilder) buildIteratorFromOperation(p *schema.Permission, op sc
 
 	case *schema.NilReference:
 		return NewEmptyFixedIterator(), nil
+	case *schema.SelfReference:
+		return NewEmptyFixedIterator(), nil // TODO is this ok?
```
tstirrat15: @barakmich

internal/services/integrationtesting/testconfigs/self.yaml.skip
tstirrat15, line 1: Need to update this - we're not actually checking the `self` relation because we're not asserting that a user is `me_or_related` to themselves.

pkg/query/self.go
```diff
@@ -0,0 +1,81 @@
...
+				yield(Path{
+					Resource: resource,
+					Relation: s.relation,
+					Subject: subject,
+				}, nil)
+			}, nil
+		}
+	}
+	// Return the empty set if none of the resources matches
+	return func(yield func(Path, error) bool) {}, nil
```
barakmich: There's a helper for this (EmptyPathSeq()) but I went to find it and found more copies of this; so this is fine and I'll do a chore on it.

barakmich (approved):
Looks good from my end -- I checked `pkg/query` and `pkg/schema/v2` thoroughly, and gave the parser a glance from my recollection of it

tstirrat15:
For anyone following along:

The integration tests are broken because the expected relations portion of the integration test harness makes an assumption that it can calculate all accessibility information based on the test relations defined in the validation file. That no longer holds for `self`-defined permissions, because they're fully synthetic and aren't represented by a relation.

Barak and I talked about how we might fix it both in the short term and the long term; I'd like to revisit that conversation on Monday and figure out whether we want to just get this PR through or come up with a better way of structuring the integration tests.

internal/services/integrationtesting/consistency_test.go
```diff
@@ -70,6 +70,114 @@ func TestConsistency(t *testing.T) {
...
+		for _, nsDef := range cad.Populated.NamespaceDefinitions {
+			_, err := ts.GetValidatedDefinition(cad.Ctx, nsDef.Name)
+			require.NoError(t, err)
+		}
+
+		testers := consistencytestutil.ServiceTesters(cad.Conn)
+		tester := testers[0]
+			runConsistencyTestsWithServiceTester(t, cad, tester, headRevision, false)
+	// Build an accessibility set.
+		accessibilitySet := consistencytestutil.BuildAccessibilitySet(t, cad.Ctx, cad.Populated, cad.DataStore)
```
miparnisari, line 101: nit: fix indentation please

internal/services/integrationtesting/consistency_test.go
```diff
@@ -892,7 +1004,18 @@ func validateDevelopmentExpectedRels(t *testing.T, devContext *development.DevCo
 
 			// For non-wildcard subjects, ensure they are accessible.
 			if subjectWithExceptions.Subject.Subject.ObjectID != tuple.PublicWildcard {
-				accessibility, permissionship, ok := vctx.accessibilitySet.AccessibiliyAndPermissionshipFor(resourceAndRelation, subjectWithExceptions.Subject.Subject)
+				accessibility, permissionship, ok := vctx.accessibilitySet.AccessibilityAndPermissionshipFor(resourceAndRelation, subjectWithExceptions.Subject.Subject)
+				if !ok {
+					fmt.Println("we are not ok")
```
miparnisari: remove?

internal/services/integrationtesting/consistency_test.go
```diff
@@ -70,6 +70,114 @@ func TestConsistency(t *testing.T) {
...
+		require.NoError(t, err)
+		resourceIds := make([]string, 0, len(foundResources))
+		for _, resource := range foundResources {
+			resourceIds = append(resourceIds, resource.ResourceObjectId)
+		}
+
+		// We expect that tom is related to himself and to tommy
+		require.ElementsMatch(t, []string{"tom", "tommyboy"}, resourceIds, "expected both tom and tommyboy as resources")
+
+		// Look for resources for tommyboy
```
miparnisari: nit: can we use alice/bob/charles for names?

internal/services/integrationtesting/consistency_test.go
```diff
@@ -70,6 +70,114 @@ func TestConsistency(t *testing.T) {
 			}
 		})
 	}
+	// NOTE: this test is defined separately because the integration test harness
+	// can't deal with logic that doesn't have a concrete relation associated with it
+	// as of time of writing.
+	// TODO: remove this test and rename self.yaml.skip to self.yaml
```
miparnisari:
```suggestion
	// NOTE: this test is mostly a copy of runConsistencyTestSuiteForFile 
	// but it is defined separately because the integration test harness
	// can't deal with logic that doesn't have a concrete relation associated with it
	// as of time of writing.
	// TODO: remove this test and rename self.yaml.skip to self.yaml
```

internal/services/integrationtesting/consistency_test.go
tstirrat15: Yeah, linter got it

pkg/query/self.go
```diff
@@ -0,0 +1,89 @@
...
+		id:       uuid.NewString(),
+		relation: relation,
+	}
+}
+
+func (s *Self) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
+	// for Self, the check returns the reflexive relation if it's in the set of resources
+	// or else returns an empty set.
+	for _, resource := range resources {
+		// TODO: does the subject relation need to be considered here?
```
barakmich: It has to be tuple.Ellipses -- the reason being that we're comparing a resource (has no relation) to a subject, which might, but the only if it's Ellipses.

Contrast to Alias, which is written as such _because_ it renames the relation in question.

We do have `s.relation` but it's not clear that that's useful when considering how the self keyword generally works. 

```
permission foo = (bar & baz) + self
```

One can ask, what is the relation for the subquery `bar & baz`, since we're ANDing two paths? The same applies to self here. Then Alias' job is to take `(bar & baz) + self` and make sure things in this set that match all constraints all are now relation `foo`.

So it's fine (useful, potentially) to fill the yielded paths with s.relation like this does, but it doesn't strictly need to be considered

pkg/query/self.go
```diff
@@ -0,0 +1,89 @@
...
+	// Return the empty set if none of the resources matches
+	return EmptyPathSeq(), nil
+}
+
+func (s *Self) IterSubjectsImpl(ctx *Context, resource Object) (PathSeq, error) {
+	return func(yield func(Path, error) bool) {
+		yield(Path{
+			Resource: resource,
+			Relation: s.relation,
+			// TODO: is this correct?
```
barakmich: Yes. For the above reason. Which may make a good comment on the file!

barakmich (approved):
Looks good; maybe an added comment block

barakmich (dismissed):

barakmich (approved):

pkg/composableschemadsl/lexer/flags.go
miparnisari: `use typechecking` doesn't work with composable schemas

PR #2786, opened 2025-12-19
feat: add warning for mixed operators without parentheses

### LINKED ISSUE #598: add warning to use parenthesis for precedence
Stemming from this comment https://github.com/authzed/docs/issues/363#issuecomment-2984469813, we want to issue a warning on schemas where folks write something like

```
permission view = foo - bar & baz
```

and they don't specify parenthesis for precedence.

ivanauth:
When writing permission expressions like `viewer - blocked & editor`, it's easy to assume arithmetic-style precedence. But SpiceDB's actual precedence (exclusion binds loosest, then intersection, then union) can lead to unexpected behavior.

This adds a lint warning that flags mixed operators at the same scope, nudging users to add parentheses and make their intent explicit.

**Example:**
```
permission view = viewer - blocked & editor  // warns
permission view = (viewer - blocked) & editor  // ok
permission view = viewer + editor + blocked  // ok (same operator)
```

The warning can be suppressed with:
```
// spicedb-ignore-warning: mixed-operators-without-parentheses
permission view = viewer - blocked & editor
```

Fixes https://github.com/authzed/zed/issues/598

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [1.56.0] - 2026-07-24
 ### Added
 - New metric `check_permissionship_total` for CheckPermission and CheckBulkPermissions that counts the number of requests that returned HAS_PERMISSION. Also, `write_relationships_updates` also includes BulkImport calls (https://github.com/authzed/spicedb/pull/3240)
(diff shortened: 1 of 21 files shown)


josephschorr:
@ivanauth this won't actually work in all cases, because expressions can cross multiple lines. Instead, this should likely be done by marking the expression during compilation with a piece of metadata, and having the warnings generator read that metadata

ivanauth:
Switched from source-scanning to compile-time metadata detection per review feedback. Added `NodeTypeParenthesizedExpression` wrapper in parser to track explicit grouping, then detect mixed operators during AST translation. Also fixed edge case where top-level parentheses like `(a + b - c)` would skip the warning.

PR #2794, opened 2025-12-30
chore: refactor resolvers

miparnisari:
No functional changes.

- A few renames
- Simplified the setup of some unit tests that _don't_ need a full blown datastore to pass

### DIFF
--- a/pkg/development/schema_position_mapper.go
+++ b/pkg/development/schema_position_mapper.go
@@ -62,21 +62,21 @@ type SchemaReference struct {
 	TargetNamePositionOffset int
 }
 
-// Resolver resolves references to schema nodes from source positions.
-type Resolver struct {
+// SchemaPositionMapper maps source positions to references.
+type SchemaPositionMapper struct {
 	schema     *compiler.CompiledSchema
 	typeSystem *schema.TypeSystem
 }
 
-// NewResolver creates a new resolver for the given schema.
-func NewResolver(compiledSchema *compiler.CompiledSchema) (*Resolver, error) {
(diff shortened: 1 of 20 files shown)


pkg/development/schema_position_mapper.go
```diff
@@ -62,21 +62,21 @@ type SchemaReference struct {
 	TargetNamePositionOffset int
 }
 
-// Resolver resolves references to schema nodes from source positions.
-type Resolver struct {
```
miparnisari, line 66: i was confused by the existence of this struct and `pkg/schema/resolver.go` so i renamed both

pkg/schema/typesystem_validation.go
```diff
@@ -277,7 +277,7 @@ func (def *Definition) Validate(ctx context.Context) (*ValidatedDefinition, erro
 
 			// Check the caveat, if any.
 			if allowedRelation.GetRequiredCaveat() != nil {
-				_, err := def.TypeSystem().resolver.LookupCaveat(ctx, allowedRelation.GetRequiredCaveat().CaveatName)
+				_, err := def.TypeSystem().GetCaveat(ctx, allowedRelation.GetRequiredCaveat().CaveatName)
```
miparnisari, line 280: fixed this code smell: https://refactoring.guru/smells/message-chains

tstirrat15: Like in the sense that it prevents the caller from needing to reach into the struct? I agree with the refactor but I don't really understand what the link is asserting.

miparnisari: >Like in the sense that it prevents the caller from needing to reach into the struct?

yes

PR #2796, opened 2025-12-31
fix(mysql,crdb): touch without expiration should remove the expiration

miparnisari:
## Description

On the MySQL and CRDB datastores, a touch without an expiration will leave the existing expiration intact if the relationship already has one. This is a bug.


josephschorr (approved):
LGTM

PR #2799, opened 2025-12-31
ci: add more entries to codecov ignores

miparnisari:


internal/datastore/postgres/postgres.go
```diff
@@ -708,32 +708,23 @@ func (pgd *pgDatastore) OfflineFeatures() (*datastore.Features, error) {
...
+		},
+		WatchEmitsImmediately: datastore.Feature{
 			Status: datastore.FeatureUnsupported,
 		},
 		IntegrityData: datastore.Feature{
 			Status: datastore.FeatureUnsupported,
 		},
 		ContinuousCheckpointing: datastore.Feature{
-			Status: datastore.FeatureUnsupported,
+			Status: continuousCheckpointing,
```
miparnisari, line 727: this is technically changing behavior because previously, this was *always* unsupported if `watchEnabled` was `false`. but i think this is a bug so i changed it

PR #2804, opened 2026-01-06
fix: issue with multiple imports leading to empty output

tstirrat15:
## Description
We had a report that a particular schema structure wasn't working for a user. This adds a test for that structure and fixes the issue.

The tl;dr: is that schema content was being dropped when a previously-encountered import was seen again, because we were `return`ing instead of `continue`ing.

## Changes
* Fix log message
* Change `return` to `continue`
* Add test and fix test
## Testing
Review. See that tests pass


barakmich (approved):
Yay new tests too

PR #2805, opened 2026-01-06
Prepare for single stored schema

josephschorr:
Refactors the code in preparation for storing schema as a singleton and accessing it in that manner

## Testing

Unit tests

### DIFF
--- a/internal/datastore/proxy/observable.go
+++ b/internal/datastore/proxy/observable.go
@@ -370,6 +370,10 @@ func (rwt *observableRWT) LegacyDeleteNamespaces(ctx context.Context, nsNames []
 	return rwt.delegate.LegacyDeleteNamespaces(ctx, nsNames, delOption)
 }
 
+func (rwt *observableRWT) SchemaWriter() (datastore.SchemaWriter, error) {
+	return rwt.delegate.SchemaWriter()
+}
+
 func (rwt *observableRWT) DeleteRelationships(ctx context.Context, filter *v1.RelationshipFilter, options ...options.DeleteOptionsOption) (uint64, bool, error) {
 	ctx, closer := observe(ctx, "DeleteRelationships", "", trace.WithAttributes(
 		filterToAttributes(filter)...,
--- a/internal/datastore/schema/schema.go
+++ b/internal/datastore/schema/schema.go
@@ -4,11 +4,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"testing"
 
 	"github.com/authzed/spicedb/pkg/caveats/types"
 	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/schemadsl/generator"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
 // LegacySchemaReaderAdapter is a common implementation of SchemaReader that uses the legacy
@@ -36,15 +40,15 @@ func (l *LegacySchemaReaderAdapter) SchemaText() (string, error) {
 		return "", err
 	}
 
+	// Check if there is any schema defined
+	if len(namespaces) == 0 {
+		return "", datastore.NewSchemaNotDefinedErr()
+	}
+
 	// Read all caveats
 	caveats, err := l.ListAllCaveatDefinitions(ctx)
 	if err != nil {
 		return "", err
-	}
-
-	// Check if there is any schema defined
-	if len(namespaces) == 0 {
-		return "", datastore.NewSchemaNotDefinedErr()
 	}
 
 	// Build a list of all schema definitions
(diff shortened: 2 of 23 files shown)


pkg/datastore/schema.go
```diff
@@ -0,0 +1,65 @@
...
+type SchemaWriter interface {
+	// WriteSchema writes the full set of schema definitions. The schema string is provided for
+	// future use but is currently ignored by implementations. The method validates that no
+	// definition names overlap, loads existing definitions, replaces changed ones, and deletes
+	// definitions no longer present.
+	WriteSchema(ctx context.Context, definitions []SchemaDefinition, schemaString string, caveatTypeSet *caveattypes.TypeSet) error
+
+	// AddDefinitionsForTesting adds or overwrites the given schema definitions. This method is
+	// only for use in testing and requires a testing.TB instance to enforce this constraint.
+	AddDefinitionsForTesting(ctx context.Context, tb testing.TB, definitions ...SchemaDefinition) error
```
tstirrat15, line 64: This seems like a sane way to do this that i wouldn't have thought of. Is there a reason that it's a part of the interface rather than something that each concrete type defines its own instance of?

pkg/datastore/legacyschema.go
```diff
@@ -0,0 +1,62 @@
...
+	// It is therefore the caller's responsibility to delete any relationships in those namespaces.
+	DeleteNamespacesOnly DeleteNamespacesRelationshipsOption = iota
+
+	// DeleteNamespacesAndRelationships indicates that namespaces and all relationships
+	// in those namespaces should be deleted.
+	DeleteNamespacesAndRelationships
+)
+
+// RevisionedCaveat is a revisioned version of a caveat definition.
+type RevisionedCaveat = RevisionedDefinition[*core.CaveatDefinition]
```
tstirrat15, line 23: Why is this one in `legacy` but the other one is in `schema`? Will caveats be handled differently under the new system?

pkg/datastore/errors.go
```diff
@@ -111,6 +111,34 @@ func NewNamespaceNotFoundErr(nsName string) error {
...
+func (err SchemaNotDefinedError) DetailsMetadata() map[string]string {
+	return map[string]string{}
+}
+
+// NewSchemaNotDefinedErr constructs a new schema not defined error.
+func NewSchemaNotDefinedErr() error {
+	return SchemaNotDefinedError{
+		error: errors.New("no schema has been defined; please call WriteSchema to start"),
+	}
+}
```
tstirrat15, line 140: Nice.

internal/services/v1/schema.go
```diff
@@ -92,41 +89,18 @@ func (ss *schemaServer) ReadSchema(ctx context.Context, _ *v1.ReadSchemaRequest)
 
 	reader := ds.SnapshotReader(headRevision)
 
-	nsDefs, err := reader.ListAllNamespaces(ctx)
+	schemaReader, err := reader.SchemaReader()
 	if err != nil {
 		return nil, ss.rewriteError(ctx, err)
 	}
 
-	caveatDefs, err := reader.ListAllCaveats(ctx)
```
tstirrat15, line 100: I like the way this simplifies (knowing that there's more of this to come)

internal/services/v1/permissions.go
```diff
@@ -1072,32 +1072,34 @@ func (ps *permissionServer) ImportBulkRelationships(stream grpc.ClientStreamingS
...
+				for _, def := range foundDefs {
+					if nsDef, ok := def.(*core.NamespaceDefinition); ok {
+						newDef, err := schema.NewDefinition(ts, nsDef)
+						if err != nil {
+							return err
+						}
+						loadedNamespaces[nsDef.Name] = newDef
+					} else if caveatDef, ok := def.(*core.CaveatDefinition); ok {
+						loadedCaveats[caveatDef.Name] = caveatDef
+					}
```
tstirrat15, line 1100: This probably isn't a "now" problem, but if we're going to upstream things like partials into the main schema DSL, will this logic need to change? Or are we only storing materialized definitions in the database and we'll say that users can't write a schema with partials?

internal/services/shared/schema.go
```diff
@@ -190,35 +205,44 @@ func ApplySchemaChangesOverExisting(
...
-			return nil, err
-		}
-	}
-
-	if !validated.additiveOnly {
-		// Delete the removed namespaces. Note that we don't need to delete relationships here,
-		// as that is handled by the ensureNoRelationshipsExistWithResourceType call above.
-		if removedObjectDefNames.Len() > 0 {
-			if err := rwt.DeleteNamespaces(ctx, removedObjectDefNames.AsSlice(), datastore.DeleteNamespacesOnly); err != nil {
+		// Write the new/changes caveats.
```
tstirrat15, line 212:
```suggestion
		// Write the new/changed caveats.
```

internal/services/shared/schema.go
```diff
@@ -190,35 +205,44 @@ func ApplySchemaChangesOverExisting(
...
 
-	// Write the new/changes caveats.
-	if len(caveatDefsWithChanges) > 0 {
-		if err := rwt.WriteCaveats(ctx, caveatDefsWithChanges); err != nil {
-			return nil, err
-		}
-	}
+	if validated.additiveOnly {
+		// DEPRECATED: Use of legacy methods for additive-only schema changes is deprecated.
+		// This path is maintained for backwards compatibility but will be removed in a future version.
```
tstirrat15, line 210: What's the backwards compatibility in question here?

internal/datastore/schema/schema.go
```diff
@@ -0,0 +1,344 @@
...
+	// Read all caveats
+	caveats, err := l.ListAllCaveatDefinitions(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	// Check if there is any schema defined
+	if len(namespaces) == 0 {
+		return "", datastore.NewSchemaNotDefinedErr()
+	}
```
tstirrat15: Nit, but this can go above the above block

internal/datastore/schema/schema.go
```diff
@@ -0,0 +1,344 @@
...
+	return &LegacySchemaWriterAdapter{
+		legacyWriter: legacyWriter,
+		legacyReader: legacyReader,
+	}
+}
+
+// WriteSchema writes the full set of schema definitions. The schema string is provided for
+// future use but is currently ignored by implementations. The method validates that no
+// definition names overlap, loads existing definitions, replaces changed ones, and deletes
+// definitions no longer present.
```
tstirrat15, line 200: Is it implicit that all of this is happening within the same transaction?

pkg/datastore/schema.go
josephschorr: So that tests can operate on it without having to know to which impl they are talking, e.g. for the generic datastore tests

pkg/datastore/legacyschema.go
josephschorr: History - I plan to move over the legacy one after this PR when the new impl goes in

internal/services/v1/permissions.go
josephschorr: We're not going to allow partials in "final" schemas - it will have to be complied before hand

internal/services/shared/schema.go
josephschorr: Serverless

internal/datastore/schema/schema.go
josephschorr: Moved

internal/datastore/schema/schema.go
josephschorr: The writer is defined on the transaction, so yes

tstirrat15 (approved):
LGTM

internal/datastore/proxy/strictreplicated.go
```diff
@@ -260,3 +260,7 @@ func (rr *strictReadReplicatedReader) LookupCounters(ctx context.Context) ([]dat
 	}
 	return counters, err
 }
+
+func (rr *strictReadReplicatedReader) SchemaReader() (datastore.SchemaReader, error) {
+	return rr.replica.SnapshotReader(rr.rev).SchemaReader()
```
miparnisari, line 265: why is this not falling back to the primary if it fails, like the other methods?

internal/datastore/proxy/observable.go
```diff
@@ -277,6 +277,10 @@ func (r *observableReader) ReverseQueryRelationships(ctx context.Context, subjec
 	}, nil
 }
 
+func (r *observableReader) SchemaReader() (datastore.SchemaReader, error) {
+	return r.delegate.SchemaReader()
```
miparnisari, line 281: why is this method not being wrapped with `observe` like the other methods?

PR #2806, opened 2026-01-07
feat: Initial implementation of the PostgresFDW

josephschorr:
The PostgresFDW allows SpiceDB to be used from within Postgres via the Postgres Foreign Data Wrapper protocol, as-if Postgres was speaking to another Postgres.

It supports querying (Check, LookupResources, LookupSubjects), as well as reads (RealRelationships), writes (WriteRelationships) and schema updates (ReadSchema/WriteSchema).

These operations are performed against "virtual" tables (`permissions`, `relationships` and `schema`) that are placed into the Postgres and translated by this proxy.

For more information, see the README.md

This is currently experimental and subject to changes

### DIFF
--- a/docs/spicedb.md
+++ b/docs/spicedb.md
@@ -30,6 +30,7 @@ A database that stores and computes permissions
 - [spicedb datastore](#reference-spicedb-datastore)	 - datastore operations
 - [spicedb lsp](#reference-spicedb-lsp)	 - serve language server protocol
 - [spicedb man](#reference-spicedb-man)	 - Generate man page
+- [spicedb postgres-fdw](#reference-spicedb-postgres-fdw)	 - serve a Postgres Foreign Data Wrapper for SpiceDB (EXPERIMENTAL)
 - [spicedb serve](#reference-spicedb-serve)	 - serve the permissions database
 - [spicedb serve-testing](#reference-spicedb-serve-testing)	 - test server with an in-memory datastore
 - [spicedb version](#reference-spicedb-version)	 - displays the version of SpiceDB
@@ -358,6 +359,36 @@ Generate a man page for SpiceDB.
 
 ```
 spicedb man
+```
+
+### Options Inherited From Parent Flags
+
+```
+      --log-format string    format of logs ("auto", "console", "json") (default "auto")
+      --log-level string     verbosity of logging ("trace", "debug", "info", "warn", "error") (default "info")
+      --skip-release-check   if true, skips checking for new SpiceDB releases
+```
+
+
+
+## Reference: `spicedb postgres-fdw`
+
+EXPERIMENTAL: Serves a Postgres-compatible interface for querying SpiceDB data using foreign data wrappers. This feature is experimental and subject to change.
+
+```
+spicedb postgres-fdw [flags]
+```
+
+### Options
+
+```
+      --postgres-access-token-secret string   (required) The password that Postgres will use to authenticate to the FDW proxy (configured in the Postgres FDW extension's OPTIONS)
+      --postgres-endpoint string              The endpoint at which to serve the Postgres protocol (default ":5432")
+      --postgres-username string              The username that Postgres will use to connect to the FDW proxy (default "postgres")
+      --shutdown-grace-period duration        The duration to wait for the server to shutdown gracefully
+      --spicedb-access-token-secret string    (required) Access token for calling the SpiceDB API
+      --spicedb-api-endpoint string           SpiceDB API endpoint (default "localhost:50051")
+      --spicedb-insecure                      Use insecure connection to SpiceDB API
 ```
 
 ### Options Inherited From Parent Flags
--- a/internal/fdw/README.md
+++ b/internal/fdw/README.md
@@ -1,0 +1,333 @@
+# SpiceDB Postgres Foreign Data Wrapper (FDW)
+
+> **EXPERIMENTAL**: This feature is experimental and subject to change. The API, behavior, and configuration options may change in future releases without notice.
+
+The SpiceDB Postgres FDW provides a Postgres-compatible interface for querying SpiceDB permissions and relationships data using standard SQL. This allows you to integrate SpiceDB with existing Postgres-based tools and applications through Foreign Data Wrapper support.
+
+## Overview
+
+The FDW proxy implements the Postgres wire protocol and translates SQL queries into SpiceDB API calls. It exposes virtual SQL tables that map to SpiceDB concepts:
+
+- **`permissions`**: Query and check permissions (CheckPermission, LookupResources, LookupSubjects)
+- **`relationships`**: Read, insert, and delete relationships
+- **`schema`**: Read SpiceDB schema definitions
+
+## Quick Start
+
+### 1. Start the FDW Proxy
+
+#### Using Docker (Recommended)
+
+```bash
+docker run --rm -p 5432:5432 \
+  authzed/spicedb \
+  postgres-fdw \
+  --spicedb-api-endpoint localhost:50051 \
+  --spicedb-access-token-secret "your-spicedb-token" \
+  --postgres-endpoint ":5432" \
+  --postgres-username "postgres" \
+  --postgres-access-token-secret "your-fdw-password"
+```
+
+#### Using the Binary
+
+```bash
+spicedb postgres-fdw \
+  --spicedb-api-endpoint localhost:50051 \
+  --spicedb-access-token-secret "your-spicedb-token" \
+  --postgres-endpoint ":5432" \
+  --postgres-username "postgres" \
+  --postgres-access-token-secret "your-fdw-password"
+```
+
+Or using environment variables:
+
+```bash
+export SPICEDB_SPICEDB_API_ENDPOINT="localhost:50051"
+export SPICEDB_SPICEDB_ACCESS_TOKEN_SECRET="your-spicedb-token"
+export SPICEDB_POSTGRES_ENDPOINT=":5432"
+export SPICEDB_POSTGRES_USERNAME="postgres"
+export SPICEDB_POSTGRES_ACCESS_TOKEN_SECRET="your-fdw-password"
+
+spicedb postgres-fdw
+```
+
+### 2. Configure Postgres Foreign Data Wrapper
+
+In your Postgres database, install and configure the FDW extension. See [`configuration.sql`](../../configuration.sql) for a complete setup example:
+
+```sql
+-- Install the postgres_fdw extension
+CREATE EXTENSION IF NOT EXISTS postgres_fdw;
+
+-- Create a foreign server pointing to the FDW proxy
+CREATE SERVER spicedb_server
+  FOREIGN DATA WRAPPER postgres_fdw
+  OPTIONS (
+    host 'localhost',
+    port '5432',
+    dbname 'ignored'
+  );
+
+-- Create user mapping with authentication credentials
+CREATE USER MAPPING FOR CURRENT_USER
+  SERVER spicedb_server
+  OPTIONS (
+    user 'postgres',
+    password 'your-fdw-password'
+  );
+
+-- Import foreign tables
+IMPORT FOREIGN SCHEMA public
+  LIMIT TO (permissions, relationships, schema)
+  FROM SERVER spicedb_server
+  INTO public;
+```
+
+### 3. Query SpiceDB Data
+
+Once configured, you can query SpiceDB using standard SQL:
+
+#### Check Permissions
+
+```sql
+-- Check if user:alice has permission to view document:readme
+SELECT has_permission
+FROM permissions
+WHERE resource_type = 'document'
+  AND resource_id = 'readme'
+  AND permission = 'view'
+  AND subject_type = 'user'
+  AND subject_id = 'alice';
+```
+
+#### Lookup Resources
+
+```sql
+-- Find all documents that user:alice can view
+SELECT resource_id
+FROM permissions
+WHERE resource_type = 'document'
+  AND permission = 'view'
+  AND subject_type = 'user'
+  AND subject_id = 'alice'
+  AND has_permission = true;
+```
+
+#### Lookup Subjects
+
+```sql
+-- Find all users who can view document:readme
+SELECT subject_id
+FROM permissions
+WHERE resource_type = 'document'
+  AND resource_id = 'readme'
+  AND permission = 'view'
+  AND subject_type = 'user'
+  AND has_permission = true;
+```
+
+#### Query Relationships
+
+```sql
+-- Read relationships for a specific resource
+SELECT resource_type, resource_id, relation, subject_type, subject_id
+FROM relationships
+WHERE resource_type = 'document'
+  AND resource_id = 'readme';
+```
+
+#### Insert Relationships
+
+```sql
+-- Add a new relationship
+INSERT INTO relationships (resource_type, resource_id, relation, subject_type, subject_id)
+VALUES ('document', 'readme', 'viewer', 'user', 'alice');
+```
+
+#### Delete Relationships
+
+```sql
+-- Remove a relationship
+DELETE FROM relationships
+WHERE resource_type = 'document'
+  AND resource_id = 'readme'
+  AND relation = 'viewer'
+  AND subject_type = 'user'
+  AND subject_id = 'alice';
+```
+
+#### Read Schema
+
+```sql
+-- Get all schema definitions
+SELECT definition FROM schema;
+```
+
+## Configuration Options
+
+### Command Line Flags
+
+**SpiceDB Connection:**
+
+- `--spicedb-api-endpoint`: SpiceDB API endpoint (default: `localhost:50051`)
+- `--spicedb-access-token-secret`: **(required)** SpiceDB API access token
+- `--spicedb-insecure`: Use insecure connection to SpiceDB
+
+**FDW Server:**
+
+- `--postgres-endpoint`: FDW server listen address (default: `:5432`)
+- `--postgres-username`: Username for Postgres authentication (default: `postgres`)
+- `--postgres-access-token-secret`: **(required)** Password for Postgres authentication
+- `--shutdown-grace-period`: Graceful shutdown timeout (default: `0s`)
+
+### Environment Variables
+
+All flags can be set via environment variables with the `SPICEDB_` prefix:
+
+```bash
+SPICEDB_SPICEDB_API_ENDPOINT
+SPICEDB_SPICEDB_ACCESS_TOKEN_SECRET
+SPICEDB_SPICEDB_INSECURE
+SPICEDB_POSTGRES_ENDPOINT
+SPICEDB_POSTGRES_USERNAME
+SPICEDB_POSTGRES_ACCESS_TOKEN_SECRET
+SPICEDB_SHUTDOWN_GRACE_PERIOD
+```
+
+## Table Schemas
+
+### `permissions` Table
+
+| Column | Type | Description |
+|--------|------|-------------|
+| `resource_type` | text | Resource type (e.g., 'document') |
+| `resource_id` | text | Resource ID |
+| `permission` | text | Permission name |
+| `subject_type` | text | Subject type (e.g., 'user') |
+| `subject_id` | text | Subject ID |
+| `optional_subject_relation` | text | Optional subject relation |
+| `has_permission` | boolean | Whether permission is granted |
+| `consistency` | text | Consistency token (ZedToken) |
+
+**Supported Operations:**
+
+- SELECT (CheckPermission, LookupResources, LookupSubjects)
+
+### `relationships` Table
+
+| Column | Type | Description |
+|--------|------|-------------|
+| `resource_type` | text | Resource type |
+| `resource_id` | text | Resource ID |
+| `relation` | text | Relation name |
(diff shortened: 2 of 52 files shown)


docs/spicedb.md
```diff
@@ -30,6 +30,7 @@ A database that stores and computes permissions
 - [spicedb datastore](#reference-spicedb-datastore)	 - datastore operations
 - [spicedb lsp](#reference-spicedb-lsp)	 - serve language server protocol
 - [spicedb man](#reference-spicedb-man)	 - Generate man page
+- [spicedb postgres-fdw](#reference-spicedb-postgres-fdw)	 - serve a Postgres Foreign Data Wrapper for SpiceDB
```
tstirrat15:
```suggestion
- [spicedb postgres-fdw](#reference-spicedb-postgres-fdw)  - serve a Postgres Foreign Data Wrapper for SpiceDB
```

whitespace change

docs/spicedb.md
```diff
@@ -370,6 +371,36 @@ spicedb man
 
 
 
+## Reference: `spicedb postgres-fdw`
+
+Serves a Postgres-compatible interface for querying SpiceDB data using foreign data wrappers
+
+```
+spicedb postgres-fdw [flags]
```
tstirrat15, line 379: General question: does this mean the `postgres-fdw` is now a part of the SpiceDB binary? When do we start to get concerned about the size of the binary?

internal/fdw/common/errors_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := NewUnsupportedError(tc.baseErr)
+			require.Error(t, err)
+			require.ErrorContains(t, err, tc.expectedPrefix)
+			require.ErrorIs(t, err, tc.baseErr)
+
+			// Verify error is properly wrapped with psql codes
+			require.Error(t, err)
```
tstirrat15: I don't see these assertions

internal/fdw/common/errors_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := NewSemanticsError(tc.baseErr)
+			require.Error(t, err)
+			require.ErrorContains(t, err, tc.expectedPrefix)
+			require.ErrorIs(t, err, tc.baseErr)
+
+			// Verify error is properly wrapped with psql codes
+			require.Error(t, err)
```
tstirrat15: Same

internal/fdw/common/errors_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := NewQueryError(tc.baseErr)
+			require.Error(t, err)
+			require.ErrorContains(t, err, tc.expectedPrefix)
+			require.ErrorIs(t, err, tc.baseErr)
+
+			// Verify error is properly wrapped with psql codes
+			require.Error(t, err)
```
tstirrat15: Same

internal/fdw/stats/stats_test.go
```diff
@@ -0,0 +1,370 @@
...
+			samples: []uint{100},
+		},
+		{
+			name:    "multiple samples",
+			samples: []uint{50, 100, 150, 200},
+		},
+		{
+			name:    "varying widths",
+			samples: []uint{10, 500, 250, 100, 300},
+		},
```
tstirrat15, line 71: How do these differ?

internal/fdw/stats/stats_test.go
```diff
@@ -0,0 +1,370 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			pkg := NewPackage("test_table", "Test Scan")
+			for _, sample := range tc.samples {
+				pkg.AddWidthSample(sample)
+			}
+
+			// Verify the package still works after adding samples
```
tstirrat15, line 83: "Works" meaning "can be explained?" Is there another way to make an assertion about whether this struct is working as expected? Or is it a simple enough entity that it doesn't really make sense?

internal/fdw/stats/stats_test.go
```diff
@@ -0,0 +1,370 @@
...
+		done <- true
+	}()
+
+	// Wait for all goroutines to complete
+	<-done
+	<-done
+	<-done
+
+	// Verify the package is still in a valid state
+	result := pkg.Explain()
```
tstirrat15, line 347: How would it fail to be in a valid state? Like what would the failure mode for this test be?

internal/fdw/tables/consistency.go
```diff
@@ -0,0 +1,84 @@
...
+	// Consistency mode values
+	consistencyMinimizeLatency = "minimize_latency"
+	consistencyFullyConsistent = "fully_consistent"
+	consistencyExactPrefix     = "@"
+
+	fieldConsistency = "consistency"
+)
+
+var consistencyField = wire.Column{
+	Table: 0,
```
tstirrat15, line 22: What does 0 mean here?

internal/fdw/tables/consistency.go
```diff
@@ -0,0 +1,84 @@
...
+	Name:  fieldConsistency,
+	Oid:   uint32(oid.T_text),
+	Width: 256,
```
tstirrat15, line 25: Similar - what does width: 256 mean in this context?

internal/fdw/configuration.sql
```diff
@@ -0,0 +1,34 @@
+CREATE EXTENSION IF NOT EXISTS postgres_fdw;
+
+CREATE SERVER spicedb_fdw FOREIGN DATA WRAPPER postgres_fdw OPTIONS (host 'hostofproxy', dbname 'default', port '5432', use_remote_estimate 'true');
+
+CREATE USER MAPPING FOR postgres SERVER spicedb_fdw OPTIONS (user 'proxyusername', password 'proxypassword');
+
+CREATE SCHEMA spicedb;
+
+-- IMPORT FOREIGN SCHEMA spicedb FROM SERVER spicedb_fdw INTO spicedb;
```
tstirrat15, line 9: Is this example usage? Should it be removed?

internal/fdw/configuration.sql
```diff
@@ -0,0 +1,34 @@
+CREATE EXTENSION IF NOT EXISTS postgres_fdw;
```
tstirrat15, line 1: Does the `if not exists` short-circuit the rest of the file if the thing already exists, or do the rest of the statements also need `if not exists`?

internal/fdw/configuration.sql
```diff
@@ -0,0 +1,34 @@
...
+
+CREATE SERVER spicedb_fdw FOREIGN DATA WRAPPER postgres_fdw OPTIONS (host 'hostofproxy', dbname 'default', port '5432', use_remote_estimate 'true');
```
tstirrat15, line 3: Curious - why are we using remote estimates? Is it because postgres estimation makes assumptions that don't hold in a SpiceDB context?

internal/fdw/session.go
```diff
@@ -0,0 +1,96 @@
...
+// It tracks transaction state and maintains cursors for the current session.
+// All methods are thread-safe.
+type session struct {
+	mu sync.Mutex
+
+	// withinTransaction indicates whether the session is currently inside a transaction.
+	withinTransaction bool // GUARDED_BY(mu)
+
+	// cursors maps cursor names to their corresponding cursor objects.
+	cursors map[string]*cursor // GUARDED_BY(mu)
```
tstirrat15, line 26: Worth using atomics for this?

internal/fdw/pgserver.go
```diff
@@ -0,0 +1,125 @@
...
+// appropriate table handlers.
+type PgBackend struct {
+	client   *authzed.Client
+	server   *wire.Server
+	username string
+	password string
+}
+
+// NewPgBackend creates a new Postgres FDW backend server.
+// The username and password are used for Postgres wire protocol authentication.
```
tstirrat15, line 31: I'm realizing that I don't know this about postgres - does everyone use username/pass or are there other mechanisms? Is that something that we'd have to support eventually?

internal/fdw/pgserver.go
```diff
@@ -0,0 +1,125 @@
...
+// It blocks until the context is cancelled or an error occurs.
+func (p *PgBackend) Run(ctx context.Context, endpoint string) error {
+	server, err := wire.NewServer(p.handler, wire.SessionMiddleware(sessionMiddleware))
+	if err != nil {
+		return err
+	}
+
+	server.Auth = wire.ClearTextPassword(p.validateAuth)
+
+	// slog.SetLogLoggerLevel(slog.LevelDebug)
```
tstirrat15, line 48:
```suggestion
```

internal/fdw/README.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+#### Query Relationships
+
+```sql
+-- Read relationships for a specific resource
+SELECT resource_type, resource_id, relation, subject_type, subject_id
+FROM relationships
+WHERE resource_type = 'document'
+  AND resource_id = 'readme';
```
tstirrat15, line 137: I'm not sure this is a "now" issue, but is there a way that we can help ensure that we don't have index misses on ReadRels?

internal/fdw/README.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+#### Insert Relationships
+
+```sql
+-- Add a new relationship
+INSERT INTO relationships (resource_type, resource_id, relation, subject_type, subject_id)
+VALUES ('document', 'readme', 'viewer', 'user', 'alice');
+```
```
tstirrat15, line 146: Something I wanna make sure we clarify in external documentation if I've got it right: the transaction support here doesn't sidestep the dual-write problem, right?

internal/fdw/pgserver.go
```diff
@@ -0,0 +1,125 @@
...
+// NewPgBackend creates a new Postgres FDW backend server.
+// The username and password are used for Postgres wire protocol authentication.
+func NewPgBackend(client *authzed.Client, username, password string) *PgBackend {
+	connHandler := &PgBackend{client: client, username: username, password: password}
+	return connHandler
+}
+
+// Run starts the Postgres wire protocol server on the specified endpoint.
+// It blocks until the context is cancelled or an error occurs.
+func (p *PgBackend) Run(ctx context.Context, endpoint string) error {
```
tstirrat15, line 39: It doesn't have to be in this PR, but I'd wanna see observability and metrics for this as a fast follow

internal/fdw/pgserver_e2e_test.go
```diff
@@ -0,0 +1,817 @@
...
+		cancel()
+	})
+
+	runnableServer, err := config.Complete(ctx)
+	require.NoError(t, err)
+
+	serverReady := make(chan bool)
+	go (func() {
+		serverReady <- true
+		_ = runnableServer.Run(ctx)
```
tstirrat15, line 748: can use `assert.NoError` here if desired

internal/fdw/pgserver_e2e_test.go
```diff
@@ -0,0 +1,817 @@
...
+	})()
+
+	// Wait for server goroutine to start
+	<-serverReady
+
+	// Verify server is ready
+	retryCount := 0
+	for retryCount < 10 {
```
tstirrat15: Can use https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithT for this if desired

pkg/cmd/postgresfdw.go
```diff
@@ -0,0 +1,177 @@
...
+	Close() error
+}
+
+func RegisterPostgresFDWFlags(cmd *cobra.Command, config *PostgresFDWConfig) error {
+	nfs := cobrautil.NewNamedFlagSets(cmd)
+
+	spicedbFlagSet := nfs.FlagSet(BoldBlue("spicedb"))
+	spicedbFlagSet.StringVar(&config.SpiceDBEndpoint, "spicedb-api-endpoint", "localhost:50051", "SpiceDB API endpoint")
+	spicedbFlagSet.StringVar(&config.SecureSpiceDBAccessToken, "spicedb-access-token-secret", "", "(required) Access token for calling the SpiceDB API")
+	spicedbFlagSet.BoolVar(&config.SpiceDBInsecure, "spicedb-insecure", false, "Use insecure connection to SpiceDB API")
```
tstirrat15: Oh wait. It didn't occur to me that these flags are scoped to this command specifically. Disregard previous comments to this effect.

pkg/cmd/postgresfdw_test.go
```diff
@@ -0,0 +1,232 @@
...
+	confdir := prepareTempConfigDir(t)
+	confFile, err := os.Create(filepath.Join(confdir, "spicedb.env"))
+	require.NoError(t, err)
+	_, err = confFile.WriteString(config)
+	require.NoError(t, err)
+	require.NoError(t, confFile.Close())
+	prevEnv := os.Environ()
+	restore := func() {
+		restoreEnv(prevEnv)
+	}
```
tstirrat15, line 43: Is this necessary with `t.Setenv`?

internal/fdw/tables/permissions.go
```diff
@@ -0,0 +1,478 @@
...
+		}
+
+		optionalSubjectRelation, err := optionalStringValue(fields[permissionsOptionalSubjectRelationField], parameters)
+		if err != nil {
+			return nil, fmt.Errorf("error getting optional_subject_relation: %w", err)
+		}
+
+		var trailerMD metadata.MD
+
+		checkResult, err := client.CheckPermission(ctx, &v1.CheckPermissionRequest{
```
tstirrat15, line 165: Is the caveat context also a todo here?

internal/fdw/tables/relationships.go
```diff
@@ -0,0 +1,388 @@
...
+			ObjectType: valuesByColumnName["resource_type"],
+			ObjectId:   valuesByColumnName["resource_id"],
+		},
+		Relation: valuesByColumnName["relation"],
+		Subject: &v1.SubjectReference{
+			Object: &v1.ObjectReference{
+				ObjectType: valuesByColumnName["subject_type"],
+				ObjectId:   valuesByColumnName["subject_id"],
+			},
+			OptionalRelation: valuesByColumnName["optional_subject_relation"],
```
tstirrat15, line 288: Does this mean that the optional subject relation has to be defined, or is the assumption that it'll typically come through as an empty string and be handled as an unset value?

internal/fdw/tables/select.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	if selectStatement.GetLimitCount() != nil {
+		return errors.New("LIMIT not supported")
+	}
+
+	return nil
+}
+
+type valueOrRef struct {
```
tstirrat15, line 258: Docstring on this one? Also maybe pull it towards the top or into a separate file.

internal/fdw/tables/statement_test.go
```diff
@@ -0,0 +1,147 @@
...
+			expectedName: "permissions",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			require.Equal(t, tc.expectedName, tc.stmt.TableName())
+			require.Equal(t, tc.stmt.isUnsupported, tc.stmt.IsUnsupported())
```
tstirrat15, line 40: What is this asserting? Is it functionally asserting that the interface is satisfied?

tstirrat15 (approved):
See comments; otherwise this is looking good to me for the first implementation.

docs/spicedb.md
josephschorr: Yes and the plan is any SpiceDB-adjacent systems will be as well.

I think, for now, we won't be worrying about it

internal/fdw/common/errors_test.go
josephschorr: Fixed

internal/fdw/common/errors_test.go
josephschorr: Fixed

internal/fdw/common/errors_test.go
josephschorr: Fixed

internal/fdw/stats/stats_test.go
josephschorr: Just more information in the widths tracking

internal/fdw/stats/stats_test.go
josephschorr: It just verifies that an EXPLAIN plan is returned. Most of this is temporary currently, with the goal to eventually replace it with real stats we get from the query planner on the SpiceDB side

internal/fdw/stats/stats_test.go
josephschorr: Mostly that it panics or does not contain the expected string

internal/fdw/tables/consistency.go
josephschorr: Here width is the max size of the data in the column

internal/fdw/configuration.sql
josephschorr: Yes its example usage and its referenced by the README

internal/fdw/configuration.sql
josephschorr: No, it does not. This is the standard way to ensure that if it already exists, it doesn't error and break loading the rest of the file

internal/fdw/configuration.sql
josephschorr: Postgres won't have any good estimates and, in a followup, the estimates will be wired to SpiceDB's query planner and/or Materialize, which *are* correct

internal/fdw/session.go
josephschorr: Nah; I think its fine to use a mutex on the session now - its only really shared by a single caller and the mutex is just for safety

internal/fdw/pgserver.go
josephschorr: I believe there might be a way to use a cert, but this is an initial impl, so I'm sticking with what is easy to use

internal/fdw/pgserver.go
josephschorr: Added comment

internal/fdw/README.md
josephschorr: No; We do not have full coverage on all shapes of read rels that are possible (and likely never will). The best we could do is warn, but that's unrelated to these changes

internal/fdw/README.md
josephschorr: No, because postgres_fdw sadly does not support two-phase commit

internal/fdw/pgserver.go
josephschorr: okay

internal/fdw/pgserver_e2e_test.go
josephschorr: Nah; I don't like error checking after cleanup in tests. It likely *is* returning an error (context canceled) when its forcibly shutdown at the end of the time

internal/fdw/tables/permissions.go
josephschorr: Yes, it has TODOs saying so

internal/fdw/tables/relationships.go
josephschorr: Empty string

internal/fdw/tables/select.go
josephschorr: Why? Its used internally

internal/fdw/tables/statement_test.go
josephschorr: Its just making sure the accessors return what was defined

internal/fdw/configuration.sql
tstirrat15: Well right, but won't the other statements in the file fail if the file has already been run once?

internal/fdw/README.md
tstirrat15: Yeah. My concern is that this implementation detail is kinda hidden, and that it might be relatively easy to write a postgres query that accidentally misses the indices and then it leaves the user wondering why SpiceDB is slow. I'm okay with saying that's a future problem, though.

internal/fdw/tables/select.go
tstirrat15: Just cuz it's used widely enough in this package. I ended up inferring it from usage as I was reading and then finding it in one of the last files I looked at and thinking that it would have been nicer to have it up front. I guess that's a review-time concern, though, so it's probably fine.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -0,0 +1,821 @@
...
+		cancel()
+		require.NoError(t, pgserver.Close())
+	})
+
+	go func() {
+		_ = pgserver.Run(ctx, fmt.Sprintf("localhost:%d", port))
+	}()
+
+	// Give PGServer time to start
+	time.Sleep(50 * time.Millisecond)
```
tstirrat15, line 697: Is there a condition that we can explicitly poll here? I don't want this to become a flake source.

tstirrat15 (approved):
LGTM. I'm curious to see what the ramifications of CGO are.

Issue #2807, opened 2026-01-08
Support fs.FS Composable Schemas


drew-richardson:
### Problem Statement

In some situations, such as build pipelines in CI, or when deploying, the schema may be distributed in other formats like a compressed archive. Although it's possible to first extract the archive to disk, then compile it, it may also make sense to support directly compiling the schema from such an archive 

### Solution Brainstorm

We can reuse go's [io.FS](https://pkg.go.dev/io/fs#FS) support. Then, for example, we could pass a [archive/zip.Reader](https://pkg.go.dev/archive/zip#Reader) (which implements io.FS) and compile the schema directly bypassing the filesystem. Here's a [RFC implementation](https://github.com/authzed/spicedb/compare/main...drew-richardson:spicedb:schemafs), if this is satisfactory, I'm happy to open a PR to discuss further.

tstirrat15:
I like it in theory 🤔 

I'm not that familiar with this part of golang. Is the idea that these approaches are more-or-less interchangeable, or would this change how the commands need to be invoked?

drew-richardson:
> I like it in theory 🤔
> 
> I'm not that familiar with this part of golang. Is the idea that these approaches are more-or-less interchangeable, or would this change how the commands need to be invoked?

I expect this to be fully backwards compatible with current usage. `go test ./...` passed locally with the changes I propose in my [RFC implementation](https://github.com/authzed/spicedb/compare/main...drew-richardson:spicedb:schemafs), I assume that runs the composable schema unit tests as well, right?

PR #2808, opened 2026-01-12
Support reading Composable Schemas from fs.FS

drew-richardson:
## Description

Enhance the `composableschemadsl/compiler.Compile` API by adding a new option `SourceFS` which instead of reading from the filesystem, reads from a virtual go filesystem. This makes no changes to the zed command line, but opens up the possibility to add something like `zed compile --using-archive schemapackage.zip root.zed` in the future.

## Testing

I ran the unit tests and verified they passed, and I also added a trivial test for `SourceFS`.

## References

https://github.com/authzed/spicedb/issues/2807

### DIFF
--- a/pkg/composableschemadsl/compiler/compiler.go
+++ b/pkg/composableschemadsl/compiler/compiler.go
@@ -3,6 +3,8 @@ package compiler
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


drew-richardson:

PR #2809, opened 2026-01-12
feat: finish initial implementation of IterSubjects

barakmich:

### DIFF
--- a/pkg/query/exclusion.go
+++ b/pkg/query/exclusion.go
@@ -69,27 +69,7 @@ func combineExclusionCaveats(mainPath, excludedPath Path) (Path, bool) {
 }
 
 func (e *Exclusion) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
-	// Get all paths from the main set
-	ctx.TraceStep(e, "getting paths from main set")
-	mainSeq, err := ctx.Check(e.mainSet, resources, subject)
-	if err != nil {
-		return nil, err
-	}
-
-	mainPaths, err := CollectAll(mainSeq)
-	if err != nil {
-		return nil, err
-	}
-
-	ctx.TraceStep(e, "main set returned %d paths", len(mainPaths))
-
-	// If main set is empty, return empty result
-	if len(mainPaths) == 0 {
-		ctx.TraceStep(e, "main set empty, returning empty")
-		return EmptyPathSeq(), nil
-	}
-
-	// Get all paths from the excluded set
+	// Get all paths from the excluded set first and build a lookup map
 	ctx.TraceStep(e, "getting paths from excluded set")
 	excludedSeq, err := ctx.Check(e.excluded, resources, subject)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/query/arrow.go
```diff
@@ -122,11 +122,82 @@ func (a *Arrow) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
+				}
+
+				// Create combined path with resource from left and subject from right
+				combinedPath := Path{
+					Resource:   leftPath.Resource,
+					Relation:   leftPath.Relation,
+					Subject:    rightPath.Subject,
+					Caveat:     combinedCaveat,
+					Expiration: rightPath.Expiration,
+					Integrity:  rightPath.Integrity,
```
tstirrat15, line 182: Would these not need to be combined somehow? Or does the arrow semantic mean that the right path takes precedence?

pkg/query/exclusion.go
```diff
@@ -138,11 +138,77 @@ func (e *Exclusion) CheckImpl(ctx *Context, resources []Object, subject ObjectAn
...
+		if shouldInclude {
+			finalPaths = append(finalPaths, resultPath)
+		} else {
+			ctx.TraceStep(e, "subject completely excluded")
+		}
+	}
+
+	return func(yield func(Path, error) bool) {
+		for _, path := range finalPaths {
+			if !yield(path, nil) {
```
tstirrat15: This is probably premature optimization, but would it make sense to yield the paths as they're determined rather than collecting them into a slice and then yielding the slice?

pkg/query/intersection_test.go
```diff
@@ -522,3 +535,139 @@ func TestIntersectionIteratorCaveatCombination(t *testing.T) {
...
+
+		intersect := NewIntersection(iter1, iter2, iter3)
+
+		pathSeq, err := ctx.IterSubjects(intersect, NewObject("document", "doc1"))
+		require.NoError(err)
+
+		paths, err := CollectAll(pathSeq)
+		require.NoError(err)
+
+		require.Len(paths, 1, "Should find alice in all three")
```
tstirrat15, line 631: This kind of surprises me - what is the concrete single path in this case? Or should this be 3?

tstirrat15: Or is the path really just pointing at Alice and that's the thing we care about because we're iterating on subjects?

pkg/query/union_test.go
```diff
@@ -637,3 +683,169 @@ func TestUnionIteratorCaveatCombination(t *testing.T) {
...
+		union.addSubIterator(iter2)
+
+		pathSeq, err := ctx.IterResources(union, NewObject("user", "alice").WithEllipses())
+		require.NoError(err)
+
+		paths, err := CollectAll(pathSeq)
+		require.NoError(err)
+
+		// Should be deduplicated to one path per resource
+		require.Len(paths, 1, "Union should deduplicate resources")
```
tstirrat15, line 801: Yeah I'm still a little hung up on this. Why doesn't the relation matter in this context? Or do I have the wrong conception of a path?

tstirrat15 (approved):
LGTM. see comments

pkg/query/arrow.go
barakmich: Good question, filing a follow-up ticket

pkg/query/intersection_test.go
barakmich: yep, that's the reason -- it's in all three, so it shows up as one (the path through to alice)

tstirrat15 (approved):
LGTM

PR #2819, opened 2026-01-14
chore: remove datastore hedging

tstirrat15:
## Description
This is a feature that we originally implemented because it was a part of the [Zanzibar paper](https://authzed.com/zanzibar/2PxynOFLlE:2c:3R), but in practice we found that it didn't give significant performance gains and reduced system stability. We've since discouraged its use. Lately it just makes our test harness less stable.

## Changes
* Make flags noop and deprecate them
* Remove proxy

## Testing
Review. See that tests pass.


ecordell:
I generally agree with removing this _but_ I would add that the lack of effectiveness has a lot to do with not being able to retry a query on a different database node. We have that ability now with the crdbpool client, but haven't ever tried to wire it up to the hedger to get behavior that's more similar to what the paper describes.

PR #2820, opened 2026-01-14
test: reduce flakiness of RegisterRelationshipCountersInParallelTest

tstirrat15:
## Description
See https://github.com/authzed/spicedb/actions/runs/21007920564/job/60396619281?pr=2818. We were getting test flakes from write failures that were the result of too much load on the memdb instance. For this reason, we're allowing the set of Memdb tests to run in parallel with other tests, but the test itself will not be parallelized to ensure that we aren't hammering the instance.

## Changes
* Turn off parallelism for memdb tests
## Testing
Review.

### DIFF
--- a/internal/datastore/memdb/memdb_test.go
+++ b/internal/datastore/memdb/memdb_test.go
@@ -27,10 +27,7 @@ func (mdbt memDBTest) New(revisionQuantization, _, gcWindow time.Duration, watch
 
 func TestMemdbDatastore(t *testing.T) {
 	t.Parallel()
-	// NOTE: The individual tests are not parallelized because enough write traffic will cause
-	// the memdb instance to reject requests with serialization errors, which results in
-	// flaky tests.
-	test.All(t, memDBTest{}, false)
+	test.All(t, memDBTest{}, true)
 }
 
 func TestConcurrentWritePanic(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tstirrat15:
Hmm... The memdb tests are already spinning up a separate memdb instance per test, so it's not likely resource contention within a single database.

internal/datastore/memdb/memdb_test.go
```diff
@@ -27,7 +27,10 @@ func (mdbt memDBTest) New(revisionQuantization, _, gcWindow time.Duration, watch
 
 func TestMemdbDatastore(t *testing.T) {
 	t.Parallel()
-	test.All(t, memDBTest{}, true)
+	// NOTE: The individual tests are not parallelized because enough write traffic will cause
```
miparnisari: i don't understand how this fixes things. Each individual test is creating its own memdb instance. For example:

```go
func RegisterRelationshipCountersInParallelTest(t *testing.T, tester DatastoreTester) {
	rawDS, err := tester.New(0, veryLargeGCInterval, veryLargeGCWindow, 1)
```

could it be that the default number of retries is 10 and this test is doing 10 parallel writes?

tstirrat15: oh hmmm... yeah I suppose we could reduce it to 9 and see if that helps?

miparnisari: yeah. but don't harcode to 9, use `numAttempts - 1` and maybe rename `numAttempts` to `defaultNumAttempts` :P

miparnisari: done

miparnisari (approved):

PR #2821, opened 2026-01-14
feat: add arrow reversal optimization

barakmich:

### DIFF
--- a/pkg/query/arrow.go
+++ b/pkg/query/arrow.go
@@ -1,10 +1,23 @@
 package query
 
 import (
+	"fmt"
+
 	"github.com/google/uuid"
 
 	"github.com/authzed/spicedb/internal/caveats"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+// arrowDirection specifies which direction to execute the arrow check
+type arrowDirection int
+
+const (
+	// leftToRight executes IterSubjects on left, Check on right
+	leftToRight arrowDirection = iota
+	// rightToLeft executes IterResources on right, Check on left
+	rightToLeft
 )
 
 // Arrow is an iterator that represents the set of paths that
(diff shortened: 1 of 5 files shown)


pkg/query/arrow.go
```diff
@@ -121,6 +118,95 @@ func (a *Arrow) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
+				return
+			}
+			rightPathCount++
+
+			// rightPath.Resource is an intermediate object from the right side
+			// Now check if any of our input resources connect to this intermediate via left
+			intermediateAsSubject := ObjectAndRelation{
+				ObjectType: rightPath.Resource.ObjectType,
+				ObjectID:   rightPath.Resource.ObjectID,
+				Relation:   "",
```
tstirrat15, line 150: What does an empty relation here mean? Is it different from an ellipsis?

pkg/query/arrow.go
```diff
@@ -121,6 +118,95 @@ func (a *Arrow) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
+		combinedCaveat = rightPath.Caveat
+	}
+
+	return Path{
+		Resource:   leftPath.Resource,
+		Relation:   leftPath.Relation,
+		Subject:    rightPath.Subject,
+		Caveat:     combinedCaveat,
+		Expiration: rightPath.Expiration,
+		Integrity:  rightPath.Integrity,
```
tstirrat15, line 205: Still curious on this point - are we arbitrarily choosing the right for expiration and integrity, or is there a reason that it makes sense semantically?

pkg/query/optimize_arrow_direction.go
```diff
@@ -0,0 +1,44 @@
...
+		if rightToLeftCost < leftToRightCost && a.direction != rightToLeft {
+			// Create new arrow with inverted direction
+			newArrow := &Arrow{
+				id:        a.id, // Keep same ID
+				left:      a.left,
+				right:     a.right,
+				direction: rightToLeft,
+			}
+			return newArrow, true, nil // Changed
+		}
```
tstirrat15, line 39: Is there a case where we'd need to switch it the other way, or is that implicit because LTR is the default?

barakmich: implicit default, yeah

PR #2823, opened 2026-01-15
chore: intersection arrow LR implementation

tstirrat15:
## Description
This is the last unimplemented LR. After this I'll move to checking and fixing the integration tests.

## Changes
Will annotate.

## Testing
Review.

### DIFF
--- a/pkg/query/intersection_arrow.go
+++ b/pkg/query/intersection_arrow.go
@@ -4,8 +4,8 @@ import (
 	"github.com/google/uuid"
 
 	"github.com/authzed/spicedb/internal/caveats"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
-	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
 // IntersectionArrow is an iterator that represents the set of relations that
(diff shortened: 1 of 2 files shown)


pkg/query/intersection_arrow.go
```diff
@@ -224,7 +224,83 @@ func (ia *IntersectionArrow) IterSubjectsImpl(ctx *Context, resource Object) (Pa
...
+	// Get all right resources
+	rightSeq, err := ctx.IterResources(ia.right, subject)
+	if err != nil {
+		return nil, err
+	}
+
+	rightPaths, err := CollectAll(rightSeq)
+	if err != nil {
+		return nil, err
+	}
```
tstirrat15, line 239: I'm curious whether we'd be able to do iterators top-to-bottom in a future refactor

pkg/query/intersection_arrow.go
```diff
@@ -224,7 +224,83 @@ func (ia *IntersectionArrow) IterSubjectsImpl(ctx *Context, resource Object) (Pa
...
+
+	ctx.TraceStep(ia, "right side returned %d resources", len(rightPaths))
+
+	if len(rightPaths) == 0 {
+		ctx.TraceStep(ia, "no right resources, returning empty")
+		return EmptyPathSeq(), nil
+	}
+
+	// seenResources is used to avoid rechecking resources that we've already seen
+	seenResources := mapz.NewSet[string]()
```
tstirrat15, line 249: This is an optimization because I'm treating invoking `CheckImpl` as expensive. I can back it out if there's a different way we want to approach this.

pkg/query/intersection_arrow.go
```diff
@@ -224,7 +224,83 @@ func (ia *IntersectionArrow) IterSubjectsImpl(ctx *Context, resource Object) (Pa
...
+	validResults := make([]Path, 0)
+
+	for _, rightPath := range rightPaths {
+		rightResourceAsSubject := rightPath.Resource.WithEllipses()
+		ctx.TraceStep(ia, "checking left side for right resource %s:%s", rightResourceAsSubject.ObjectType, rightResourceAsSubject.ObjectID)
+
+		// TODO: should we be cloning here?
+		leftSeq, err := ctx.IterResources(ia.left, rightResourceAsSubject)
```
tstirrat15: Curious on this point - it seems like we'd be iterating over `left` as many times as `len(rightPaths)`

pkg/query/intersection_arrow.go
```diff
@@ -224,7 +224,83 @@ func (ia *IntersectionArrow) IterSubjectsImpl(ctx *Context, resource Object) (Pa
...
+
+		ctx.TraceStep(ia, "right subject %s:%s returned %d left resources", rightPath.Resource.ObjectType, rightPath.Resource.ObjectID, len(leftPaths))
+
+		// Make a list of the leftPaths that we haven't seen
+		// before that we need to check
+		leftResources := make([]Object, 0, len(leftPaths))
+		for _, path := range leftPaths {
+			resource := path.Resource
+			key := resource.Key()
+			if notSeen := seenResources.Add(key); !notSeen {
```
tstirrat15, line 274: I don't like the double negative here but this is the semantic that `Set` uses

pkg/query/intersection_arrow_test.go
tstirrat15, line 1: The noise in this file is me moving some things around and getting rid of the datastore pieces

pkg/query/intersection_arrow_test.go
```diff
@@ -563,52 +563,50 @@ func TestIntersectionArrowIteratorUnimplementedMethods(t *testing.T) {
 	rightIter := NewFixedIterator()
 	intersectionArrow := NewIntersectionArrow(leftIter, rightIter)
 
-	// Create test context
-	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
```
tstirrat15, line 567: This isn't actually used by the test, so I got rid of it

pkg/query/intersection_arrow_test.go
```diff
@@ -563,52 +563,50 @@ func TestIntersectionArrowIteratorUnimplementedMethods(t *testing.T) {
...
-	require := require.New(t)
-
 	ctx := &Context{
 		Context:  t.Context(),
 		Executor: LocalExecutor{},
 	}
 
 	t.Run("AllLeftSubjectsSatisfyRight", func(t *testing.T) {
 		t.Parallel()
+		require := require.New(t)
```
tstirrat15, line 609: The `require` instance needs to be associated with the most proximate `t` to prevent test failures from returning panic stacktraces.

pkg/query/intersection_arrow.go
barakmich: Yeah, that makes sense. In fact, one thing to ponder is a double-negative construction of intersection arrows (you can construct an "all" with `(P & !(!P))`)

pkg/query/intersection_arrow.go
barakmich: Yeah, it's not ideal. If you want to help highlight this, an interesting change (in this, or follow-up) is to take that cost into consideration in the statistics part of the package.

pkg/query/intersection_arrow.go
```diff
@@ -224,7 +224,83 @@ func (ia *IntersectionArrow) IterSubjectsImpl(ctx *Context, resource Object) (Pa
...
+
+	// seenResources is used to avoid rechecking resources that we've already seen
+	seenResources := mapz.NewSet[string]()
+	validResults := make([]Path, 0)
+
+	for _, rightPath := range rightPaths {
+		rightResourceAsSubject := rightPath.Resource.WithEllipses()
+		ctx.TraceStep(ia, "checking left side for right resource %s:%s", rightResourceAsSubject.ObjectType, rightResourceAsSubject.ObjectID)
+
+		// TODO: should we be cloning here?
```
barakmich: Cloning goes more towards building a query rather than evaluating it; we should be fine

pkg/query/intersection_arrow.go
```diff
@@ -224,7 +224,83 @@ func (ia *IntersectionArrow) IterSubjectsImpl(ctx *Context, resource Object) (Pa
...
+		for _, path := range leftPaths {
+			resource := path.Resource
+			key := resource.Key()
+			if notSeen := seenResources.Add(key); !notSeen {
+				leftResources = append(leftResources, path.Resource)
+			}
+		}
+
+		// We check each of the left resources with the right resource as the subject
+		checkSeq, err := ia.CheckImpl(ctx, leftResources, rightResourceAsSubject)
```
barakmich: Each of the left resources -- yes, but since we're checking this arrow itself, the right subject should be the original `subject`.

Plus, this is worth a long comment. Something like "first we're collecting any potential left resources; but in order to check that any left resource can satisfy ALL its paths to the right ones, we have to test them individually"

PR #2832, opened 2026-01-19
makes clearer pg timestamps should not be used for ordering txs

vroldanbet:
## Description

Captured some feedback from @jakedt, where it's possible engineers may end up overlooking the fact pg timestamps cannot be used to order transactions.

## Testing

No need to test, it's just a variable rename and comments


tstirrat15 (approved):
LGTM

PR #2838, opened 2026-01-21
feat: implement a proper BFS when doing IterSubjects/IterResources for recursive nodes

barakmich:

### DIFF
--- a/pkg/query/recursive.go
+++ b/pkg/query/recursive.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 
 	"github.com/google/uuid"
+
+	"github.com/authzed/spicedb/pkg/tuple"
 )
 
 const defaultMaxRecursionDepth = 50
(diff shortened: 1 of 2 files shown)


pkg/query/recursive_benchmark_test.go
```diff
@@ -0,0 +1,347 @@
...
+			Resource: Object{ObjectType: "folder", ObjectID: "folder3"},
+			Relation: "viewer",
+			Subject:  ObjectAndRelation{ObjectType: "user", ObjectID: "alice", Relation: "..."},
+		},
+	}
+
+	iter := NewFixedIterator(paths...)
+	recursive := NewRecursiveIterator(iter, "folder", "parent")
+
+	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
```
tstirrat15, line 40: I was wondering about this in other tests too - if we're using fixed iterators, do we need the datastore at all? When I removed it from some of the tests it seemed to work just fine.

barakmich: I'll file a ticket to check and simplify

PR #2844, opened 2026-01-23
feat(schema): add support for parent namespace in arrow information

mazdakb:
Enhance the `ArrowSet` utility in `pkg/schema` to provide more granular information about arrows `TupleToUserset`s defined in a schema. It adds parent namespace to `ArrowInformation` and a new lookup method to retrieve arrows by their computed userset.

### Changes
- Added a `ParentNamespace` field to `ArrowInformation`. This allows each arrow reference to be fully qualified by its parent relation's location (namepace + relation name).
- Added `LookupArrowsWithComputedUserset` on `ArrowSet` that returns all `ArrowInformation` objects pointing to a specific namespace and relation.

### Why
Currently, `ArrowSet` only provides a boolean check (`HasPossibleArrowWithComputedUserset`) to see if any arrow points to a relation. This forces consumers who need to perform filtered analysis i.e., checking if an arrow is reachable from a specific set of entry points, to manually walk the schema (using `graph.WalkRewrite`) or parse userset rewrites.
By including the parent context and providing a direct lookup, consumers can now efficiently retrieve and filter arrows without duplicating the traversal logic already performed by `ArrowSet`.

### DIFF
--- a/pkg/schema/arrows_test.go
+++ b/pkg/schema/arrows_test.go
@@ -271,3 +271,275 @@ func TestAllReachableRelations(t *testing.T) {
 		})
 	}
 }
+
+func TestLookupArrowsWithComputedUserset(t *testing.T) {
+	t.Parallel()
+
+	type expectedArrow struct {
+		parentNamespace string
+		parentRelation  string
+		arrowExpression string
+	}
+
+	type testcase struct {
+		name       string
+		schemaText string
+
+		targetNamespace string
+		targetRelation  string
+		expected        []expectedArrow
+	}
+
+	tcs := []testcase{
+		{
+			name: "basic arrow",
+			schemaText: `
+			definition user {}
+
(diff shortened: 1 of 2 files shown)


pkg/schema/arrows_test.go
```diff
@@ -271,3 +271,157 @@ func TestAllReachableRelations(t *testing.T) {
...
+			schemaText: `
+			definition user {}
+
+			definition organization {
+				relation member: user
+			}
+
+			definition resource {
+				relation org: organization
+				permission view = org->member
```
miparnisari, line 305: can we write a test case where `view` is a complex expression like `view = a + b - c + org->member`?

mazdakb: Added three more cases with a complex expression, nesting and two arrows in one expression.

PR #2846, opened 2026-01-23
chore: use fork of jitterbug

miparnisari:
This will eliminate the need to exclude this library from the goroutine detector `goleak`.

### DIFF
--- a/internal/datastore/crdb/pool/health.go
+++ b/internal/datastore/crdb/pool/health.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v5"
-	"github.com/lthibault/jitterbug"
+	"github.com/miparnisari/jitterbug"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/time/rate"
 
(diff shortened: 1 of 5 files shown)


internal/datastore/crdb/pool/health.go
```diff
@@ -63,7 +63,6 @@ func (t *NodeHealthTracker) Poll(ctx context.Context, interval time.Duration) {
 		// G404 use of non cryptographically secure random number generator is not concern here,
 		// as it's used for jittering the interval for health checks.
 		Source: rand.New(rand.NewSource(time.Now().Unix())),
-		Min:    interval,
```
miparnisari, line 66: In the current library and version v2.0.0, `Uniform.Jitter()` draws a random value from [0, interval). Then it returns `max(random_value, Min)`. Since `Min = interval` and the random value is always <         
  interval, the result will always be exactly interval. 

This gives us zero jitter - the ticker will tick at exactly `interval` every time, defeating the purpose of using jitterbug 😆 

My fork is based on v2.2.2 which panics when this happens.

PR #2849, opened 2026-01-23
feat: each iterator can report its types on either end now

barakmich:

### DIFF
--- a/pkg/query/alias.go
+++ b/pkg/query/alias.go
@@ -200,17 +200,9 @@ func (a *Alias) ID() string {
 }
 
 func (a *Alias) ResourceType() (ObjectType, error) {
-	// Get the type from the wrapped iterator but use the alias relation
-	subType, err := a.subIt.ResourceType()
-	if err != nil {
-		return ObjectType{}, err
-	}
-	return ObjectType{
-		Type: subType.Type,
-	}, nil
+	return a.subIt.ResourceType()
 }
 
 func (a *Alias) SubjectTypes() ([]ObjectType, error) {
-	// Subject types are unchanged by aliasing
 	return a.subIt.SubjectTypes()
 }
--- a/pkg/query/self.go
+++ b/pkg/query/self.go
@@ -10,21 +10,20 @@ import (
 // Resource in the subiterator that connects it to
 // streamed from the sub-iterator to a specified alias relation.
 type Self struct {
-	id string
-	// relation is the name of the permission that uses the self keyword,
-	// and is used because otherwise "self" doesn't have an associated relation,
-	// as it's purely checking whether the subject and object are the same.
-	relation     string
-	resourceType ObjectType
+	id       string
+	relation string
+	// typeName is the name of the type associated with the self definition.
+	// It's used to generate the ObjectType for ResourceType() and SubjectTypes().
+	typeName string
 }
 
 var _ Iterator = &Self{}
 
-func NewSelf(relation string, resourceType ObjectType) *Self {
+func NewSelf(relation string, typeName string) *Self {
 	return &Self{
-		id:           uuid.NewString(),
-		relation:     relation,
-		resourceType: resourceType,
+		id:       uuid.NewString(),
+		relation: relation,
+		typeName: typeName,
 	}
 }
 
@@ -73,9 +72,9 @@ func (s *Self) IterResourcesImpl(ctx *Context, subject ObjectAndRelation) (PathS
 
 func (s *Self) Clone() Iterator {
 	return &Self{
-		id:           uuid.NewString(),
-		relation:     s.relation,
-		resourceType: s.resourceType,
+		id:       uuid.NewString(),
+		relation: s.relation,
+		typeName: s.typeName,
 	}
 }
 
@@ -99,10 +98,10 @@ func (s *Self) ID() string {
 }
 
 func (s *Self) ResourceType() (ObjectType, error) {
-	return s.resourceType, nil
+	return ObjectType{Type: s.typeName}, nil
 }
 
 func (s *Self) SubjectTypes() ([]ObjectType, error) {
 	// Self is self-referential - subjects are the same type as resources
-	return []ObjectType{s.resourceType}, nil
+	return []ObjectType{{Type: s.typeName}}, nil
 }
--- a/pkg/query/types.go
+++ b/pkg/query/types.go
@@ -114,7 +114,6 @@ func collectAndDeduplicateSubjectTypes(iterators []Iterator) ([]ObjectType, erro
 	}
 	set := mapz.NewSet[string]()
 
-	seen := make(map[string]bool)
 	var result []ObjectType
 
 	for _, it := range iterators {
@@ -123,10 +122,9 @@ func collectAndDeduplicateSubjectTypes(iterators []Iterator) ([]ObjectType, erro
 			return nil, err
 		}
 		for _, st := range subjectTypes {
-			set.Add(st.String())
-			key := st.Type + "#" + st.Subrelation
-			if !seen[key] {
-				seen[key] = true
+			key := st.String()
+			if !set.Has(key) {
+				set.Add(st.String())
 				result = append(result, st)
 			}
 		}
--- a/pkg/query/union_test.go
+++ b/pkg/query/union_test.go
@@ -19,15 +19,12 @@ func TestUnionIterator(t *testing.T) {
 	t.Run("Check_Union", func(t *testing.T) {
 		t.Parallel()
 
-		// Create a union of different access patterns
-		union := NewUnion()
-
 		// Add different iterators with distinct data sets
 		documentAccess := NewDocumentAccessFixedIterator()
 		multiRole := NewMultiRoleFixedIterator()
 
-		union.addSubIterator(documentAccess)
-		union.addSubIterator(multiRole)
+		// Create a union of different access patterns
+		union := NewUnion(documentAccess, multiRole)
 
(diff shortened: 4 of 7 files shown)


pkg/query/exclusion_test.go
```diff
@@ -854,3 +854,50 @@ func TestExclusionIterSubjects(t *testing.T) {
...
+
+		// Check that both user and group types are present
+		typeMap := make(map[string]bool)
+		for _, st := range subjectTypes {
+			key := st.Type + "#" + st.Subrelation
+			typeMap[key] = true
+		}
+
+		require.True(typeMap["user#..."] || typeMap["user#"])
+		require.True(typeMap["group#member"])
```
tstirrat15: https://pkg.go.dev/github.com/stretchr/testify/require#Contains?

pkg/query/exclusion_test.go
```diff
@@ -854,3 +854,50 @@ func TestExclusionIterSubjects(t *testing.T) {
...
+		require.Len(subjectTypes, 2) // Only from mainSet, not excludedSet
+
+		// Check that both user and group types are present
+		typeMap := make(map[string]bool)
+		for _, st := range subjectTypes {
+			key := st.Type + "#" + st.Subrelation
+			typeMap[key] = true
+		}
+
+		require.True(typeMap["user#..."] || typeMap["user#"])
```
tstirrat15: Why the or here? Is it nondeterministic?

pkg/query/fixed.go
```diff
@@ -15,16 +15,49 @@ import (
...
+			key := subjectType.Type + "#" + subjectType.Subrelation
+			subjectTypeMap[key] = subjectType
+		}
+	}
+
+	// Convert subject types map to slice
+	subjectTypes := make([]ObjectType, 0, len(subjectTypeMap))
+	for _, st := range subjectTypeMap {
+		subjectTypes = append(subjectTypes, st)
+	}
```
tstirrat15, line 54: `slices.Collect(maps.Keys())`

tstirrat15: Or use a `mapz.Set`

pkg/query/fixed_test.go
```diff
@@ -108,3 +108,57 @@ func TestFixedIterator(t *testing.T) {
...
+
+		// Check that both user and group types are present
+		typeMap := make(map[string]bool)
+		for _, st := range subjectTypes {
+			key := st.Type + "#" + st.Subrelation
+			typeMap[key] = true
+		}
+
+		require.True(typeMap["user#..."] || typeMap["user#"])
+		require.True(typeMap["group#member"])
```
tstirrat15: Same comments

pkg/query/intersection.go
```diff
@@ -341,3 +341,15 @@ func (i *Intersection) ReplaceSubiterators(newSubs []Iterator) (Iterator, error)
 func (i *Intersection) ID() string {
 	return i.id
 }
+
+func (i *Intersection) ResourceType() ObjectType {
+	if len(i.subIts) == 0 {
+		return ObjectType{}
+	}
+	// All subiterators should have the same resource type
+	return i.subIts[0].ResourceType()
```
tstirrat15: Is this enforced somewhere, implicitly or explicitly?

pkg/query/intersection_test.go
```diff
@@ -672,3 +672,94 @@ func TestIntersectionIterSubjects(t *testing.T) {
...
+
+		// Check that both user and group types are present
+		typeMap := make(map[string]bool)
+		for _, st := range subjectTypes {
+			key := st.Type + "#" + st.Subrelation
+			typeMap[key] = true
+		}
+
+		require.True(typeMap["user#..."] || typeMap["user#"])
+		require.True(typeMap["group#member"])
```
tstirrat15: Sames

pkg/query/types.go
```diff
@@ -55,6 +55,24 @@ type Iterator interface {
...
+	// SubjectTypes returns all the ObjectTypes for this iterator tree.
+	SubjectTypes() []ObjectType
+}
+
+type ObjectType struct {
+	Type        string
+	Subrelation string
+}
+
+func NewType(typename string, subrelation ...string) ObjectType {
```
tstirrat15, line 76: Oh, this is how you do optional args in golang. I was trying to figure out why you'd ignore every arg after the first.

pkg/query/types.go
```diff
@@ -79,3 +97,25 @@ func (e Explain) IndentString(depth int) string {
...
 	return fmt.Sprintf("%s%s\n%s", strings.Repeat("\t", depth), e.Info, sb.String())
 }
+
+// collectAndDeduplicateSubjectTypes collects subject types from multiple iterators and deduplicates
+func collectAndDeduplicateSubjectTypes(iterators []Iterator) []ObjectType {
+	if len(iterators) == 0 {
+		return []ObjectType{}
+	}
+
+	seen := make(map[string]bool)
```
tstirrat15: Could also use `mapz.Set` here as well

pkg/query/alias.go
```diff
@@ -198,3 +198,20 @@ func (a *Alias) ReplaceSubiterators(newSubs []Iterator) (Iterator, error) {
...
+
+func (a *Alias) ResourceType() (ObjectType, error) {
+	// Get the type from the wrapped iterator but use the alias relation
+	subType, err := a.subIt.ResourceType()
+	if err != nil {
+		return ObjectType{}, err
+	}
+	return ObjectType{
+		Type:        subType.Type,
+		Subrelation: a.relation,
```
tstirrat15: Wait, what does a subrelation on a resource mean?

pkg/query/intersection.go
```diff
@@ -341,3 +343,33 @@ func (i *Intersection) ReplaceSubiterators(newSubs []Iterator) (Iterator, error)
...
+	for idx, subIt := range i.subIts[1:] {
+		subType, err := subIt.ResourceType()
+		if err != nil {
+			return ObjectType{}, err
+		}
+		if firstType != subType {
+			return ObjectType{}, spiceerrors.MustBugf("intersection resource type mismatch: subiterator 0 has type %s, subiterator %d has type %s",
+				firstType.String(), idx+1, subType.String())
+		}
+	}
```
tstirrat15, line 368: Is this potentially expensive?

barakmich: Not terribly, no.

pkg/query/types.go
```diff
@@ -79,3 +106,31 @@ func (e Explain) IndentString(depth int) string {
...
+	var result []ObjectType
+
+	for _, it := range iterators {
+		subjectTypes, err := it.SubjectTypes()
+		if err != nil {
+			return nil, err
+		}
+		for _, st := range subjectTypes {
+			set.Add(st.String())
+			key := st.Type + "#" + st.Subrelation
```
miparnisari:
```suggestion
			key := st.String()
```

pkg/query/union_test.go
```diff
@@ -834,3 +834,119 @@ func TestUnionIterResourcesDeduplication(t *testing.T) {
...
+
+		// Create union with fixed iterators
+		path1 := MustPathFromString("document:doc1#viewer@user:alice")
+		path2 := MustPathFromString("document:doc2#editor@user:bob")
+		iter1 := NewFixedIterator(path1)
+		iter2 := NewFixedIterator(path2)
+
+		union := NewUnion()
+		union.addSubIterator(iter1)
+		union.addSubIterator(iter2)
```
miparnisari:
```suggestion
		union := NewUnion(iter1, iter2)
```

pkg/query/alias.go
```diff
@@ -198,3 +198,19 @@ func (a *Alias) ReplaceSubiterators(newSubs []Iterator) (Iterator, error) {
 func (a *Alias) ID() string {
 	return a.id
 }
+
+func (a *Alias) ResourceType() (ObjectType, error) {
+	// Get the type from the wrapped iterator but use the alias relation
```
miparnisari: ? the alias relation isn't being used

pkg/query/arrow.go
```diff
@@ -358,3 +358,13 @@ func (a *Arrow) ReplaceSubiterators(newSubs []Iterator) (Iterator, error) {
 func (a *Arrow) ID() string {
 	return a.id
 }
+
+func (a *Arrow) ResourceType() (ObjectType, error) {
+	// Arrow's resources come from the left side
```
miparnisari, line 363: i don't yet understand the reasoning behind this (why left and not right?) i suppose it will become apparent in a follow up PR where we _use_ these functions

but in any case i'd move this to be a proper godoc of the function

pkg/query/self.go
```diff
@@ -14,15 +14,17 @@ type Self struct {
...
 	// as it's purely checking whether the subject and object are the same.
-	relation string
+	relation     string
+	resourceType ObjectType
 }
 
 var _ Iterator = &Self{}
 
-func NewSelf(relation string) *Self {
+func NewSelf(relation string, resourceType ObjectType) *Self {
```
miparnisari: i can't find a caller of this function where `relation != resourceType.subrelation`. Do you need that first parameter at all?

pkg/query/arrow.go
barakmich:
```
definition user {}

definition group {
 relation member: user
}

definition document {
 relation group: group
 permission view = group->member
}
```

for `view`, then, the `document` is clearly the type of the resource -- it's also the type of the resource for the `group` relation (so this is the left side). Then the right side is the right side of member, which is `user`

PR #2850, opened 2026-01-24
feat: add type filtering to the query plan

barakmich:
Depends on #2849

### DIFF
--- a/pkg/query/executor_local.go
+++ b/pkg/query/executor_local.go
@@ -14,11 +14,21 @@ func (l LocalExecutor) Check(ctx *Context, it Iterator, resources []Object, subj
 }
 
 // IterSubjects returns a sequence of all the paths in this set that match the given resource.
-func (l LocalExecutor) IterSubjects(ctx *Context, it Iterator, resource Object) (PathSeq, error) {
-	return it.IterSubjectsImpl(ctx, resource)
+func (l LocalExecutor) IterSubjects(ctx *Context, it Iterator, resource Object, filterSubjectType ObjectType) (PathSeq, error) {
+	pathSeq, err := it.IterSubjectsImpl(ctx, resource, filterSubjectType)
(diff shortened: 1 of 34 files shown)


pkg/query/executor_local.go
```diff
@@ -14,11 +14,21 @@ func (l LocalExecutor) Check(ctx *Context, it Iterator, resources []Object, subj
...
 // IterSubjects returns a sequence of all the paths in this set that match the given resource.
-func (l LocalExecutor) IterSubjects(ctx *Context, it Iterator, resource Object) (PathSeq, error) {
-	return it.IterSubjectsImpl(ctx, resource)
+func (l LocalExecutor) IterSubjects(ctx *Context, it Iterator, resource Object, filterSubjectType ObjectType) (PathSeq, error) {
+	pathSeq, err := it.IterSubjectsImpl(ctx, resource, filterSubjectType)
+	if err != nil {
+		return nil, err
+	}
+	// Apply filtering wrapper - this is where the actual filtering happens
+	return FilterSubjectsByType(pathSeq, filterSubjectType), nil
```
tstirrat15, line 23: I think this is the only part that's surprising to me - why does this need to be plumbed all the way through the tree if we're doing the filtering at the outermost layer? Or am I misinterpreting that?

miparnisari: according to Claude:

> Why the redundancy?                                                          
>                                                                                
>   This appears to be a defense-in-depth pattern where:                         
>   - Iterator implementations could optimize by filtering early (reducing    unnecessary computation)                                                     
>   - But the executor guarantees filtering happens as a safety wrapper

tstirrat15 (approved):

pkg/query/filter_test.go
miparnisari, line 1: there is a `filter_test.go` but not `filter.go`?

PR #2852, opened 2026-01-27
fix: make schemautil.ApplySchemaChangesOverExisting preserve existing stuff

tstirrat15:
## Description
`ApplySchemaChangesOverExisting` was intended to allow you to patch a subset of a schema. When we refactored the `WriteSchema` logic in #2805, we didn't account for definitions in the datastore that remained unchanged - the new schema and its patch were treated as the entirety of the schema that we were operating on.

This changes the logic so that it accounts for the unchanged portion of the schema and writes it along with the rest of the changes into the written schema.

## Changes
* Add a failing test
* Fix that failing test by reading out what's in the datastore
## Testing
Review. Tell me things that don't make sense so that I can clarify them.

### DIFF
--- a/internal/services/shared/schema.go
+++ b/internal/services/shared/schema.go
@@ -26,8 +26,10 @@ type ValidatedSchemaChanges struct {
 	validatedDefinitions map[string]*schema.ValidatedDefinition
 	newCaveatDefNames    *mapz.Set[string]
 	newObjectDefNames    *mapz.Set[string]
-	additiveOnly         bool
-	schemaText           string
+	// additiveOnly indicates whether all operations should be applied
+	// or only those operations which are additive (i.e. not changes or deletions)
+	additiveOnly bool
+	schemaText   string
 }
 
 // ValidateSchemaChanges validates the schema found in the compiled schema and returns a
@@ -120,7 +122,7 @@ func ApplySchemaChanges(ctx context.Context, rwt datastore.ReadWriteTransaction,
 // ApplySchemaChangesOverExisting applies schema changes found in the validated changes struct, against
 // existing caveat and object definitions given.
 // The idea is that the given schema will be diffed against the existing objects given, and any
-// objects not named in one of those two will be preserved.
+// objects not named in one of those two which are present in the datastore will be preserved.
 func ApplySchemaChangesOverExisting(
 	ctx context.Context,
 	rwt datastore.ReadWriteTransaction,
--- a/internal/services/shared/schema_test.go
+++ b/internal/services/shared/schema_test.go
@@ -640,7 +640,6 @@ func TestApplySchemaChangesOverExisting(t *testing.T) {
 		startingSchema               string
 		patchSchema                  string
 		expectedSchema               string
-		relationships                []string
 		expectedAppliedSchemaChanges AppliedSchemaChanges
 		expectedError                string
 	}{
@@ -673,6 +672,7 @@ func TestApplySchemaChangesOverExisting(t *testing.T) {
 			`,
 			expectedSchema: "caveat catchTwentyTwo(value int) {\n\tvalue == 22\n}\n\ndefinition organization {\n\trelation member: user\n\tpermission admin = member\n}\n\ndefinition user {}",
 			expectedAppliedSchemaChanges: AppliedSchemaChanges{
+				// NOTE: this is 5 because the `user` definition is written even though it's unchanged
 				TotalOperationCount:   5,
 				NewObjectDefNames:     []string{"organization"},
 				RemovedObjectDefNames: []string{"document"},
@@ -712,6 +712,7 @@ func TestApplySchemaChangesOverExisting(t *testing.T) {
 			expectedSchema: "caveat catchTwentyTwo(value int) {\n\tvalue == 22\n}\n\ndefinition admin {}\n\ndefinition organization {\n\trelation member: user\n\tpermission admin = member\n}\n\ndefinition user {}",
 			// NOTE: we're expecting that the `admin` part of the schema stays there.
 			expectedAppliedSchemaChanges: AppliedSchemaChanges{
+				// NOTE: this is 5 because the `user` definition is written even though it's unchanged
 				TotalOperationCount:   5,
 				NewObjectDefNames:     []string{"organization"},
 				RemovedObjectDefNames: []string{"document"},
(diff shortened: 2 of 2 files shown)


internal/services/shared/schema_test.go
```diff
@@ -632,3 +632,148 @@ definition resource {
...
+				definition organization {
+					relation member: user
+					permission admin = member
+				}
+
+				caveat catchTwentyTwo(value int) {
+				value == 22
+				}
+			`,
+			expectedSchema: "caveat catchTwentyTwo(value int) {\n\tvalue == 22\n}\n\ndefinition organization {\n\trelation member: user\n\tpermission admin = member\n}\n\ndefinition user {}",
```
josephschorr, line 673: shouldn't this be the same after the patch?

tstirrat15: It is, but because it's generated it's annoying to have the same formatting. I can reformat for readability.

internal/services/shared/schema_test.go
```diff
@@ -632,3 +632,148 @@ definition resource {
...
+			expectedAppliedSchemaChanges: AppliedSchemaChanges{
+				TotalOperationCount:   5,
```
miparnisari, line 676: hmm how is this 5 and not 4? is there an operation not accounted for in the fields below?

internal/services/shared/schema_test.go
```diff
@@ -632,3 +632,148 @@ definition resource {
...
+			rawDS, err := dsfortesting.NewMemDBDatastoreForTesting(t, 0, 0, memdb.DisableGC)
+			require.NoError(err)
+
+			// Write the initial schema.
+			relationships := make([]tuple.Relationship, 0, len(tc.relationships))
+			for _, rel := range tc.relationships {
+				relationships = append(relationships, tuple.MustParse(rel))
+			}
+
+			schemaInDB := tc.staticSchema + "\n\n" + tc.startingSchema
```
miparnisari, line 735: can you clarify with a comment why this line is like this? why didn't you just write `tc.startingSchema` in the DB and get rid of `tc.staticSchema`?

internal/services/shared/schema_test.go
```diff
@@ -632,3 +632,148 @@ definition resource {
...
+			compiled, err := compiler.Compile(compiler.InputSchema{
+				Source:       input.Source("schema"),
+				SchemaString: tc.patchSchema,
+			}, compiler.AllowUnprefixedObjectType())
+			require.NoError(err)
+
+			validated, err := ValidateSchemaChanges(t.Context(), compiled, caveattypes.Default.TypeSet, false, tc.patchSchema)
+			if tc.expectedError != "" && err != nil && tc.expectedError == err.Error() {
+				return
+			}
```
miparnisari:
```suggestion
			if tc.expectedError != "" {
				require.EqualError(err, tc.expectedError) 
				return
			}
```

internal/services/shared/schema_test.go
```diff
@@ -632,3 +632,148 @@ definition resource {
...
+
+			schemaInDB := tc.staticSchema + "\n\n" + tc.startingSchema
+
+			compiledStartingSchema, err := compiler.Compile(compiler.InputSchema{
+				Source:       input.Source("schema"),
+				SchemaString: tc.startingSchema,
+			}, compiler.AllowUnprefixedObjectType())
+			require.NoError(err)
+
+			ds, _ := testfixtures.DatastoreFromSchemaAndTestRelationships(rawDS, schemaInDB, relationships, require)
```
miparnisari: when the unit test for a function has a lot of setup required i'm starting to think there is a code smell somewhere 😅 

in this case, it's unclear to me why you need to write relationships at all in this test, or why you need to write the schema in the db.

i understand that down below you need a `rwt` and the easiest way to get one is through the `ds`, but couldn't you pass a `ds` with no data for that?

internal/services/shared/schema_test.go
```diff
@@ -632,3 +632,148 @@ definition resource {
...
+
+			validated, err := ValidateSchemaChanges(t.Context(), compiled, caveattypes.Default.TypeSet, false, tc.patchSchema)
+			if tc.expectedError != "" && err != nil && tc.expectedError == err.Error() {
+				return
+			}
+
+			require.NoError(err)
+
+			_, err = ds.ReadWriteTx(t.Context(), func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+				applied, err := ApplySchemaChangesOverExisting(t.Context(), rwt, caveattypes.Default.TypeSet, validated, compiledStartingSchema.CaveatDefinitions, compiledStartingSchema.ObjectDefinitions)
```
miparnisari: hehe, this line is giving me a headache just by reading it. i read the go doc and i still cannot visualize what it does. too many parameters 😄

internal/services/shared/schema_test.go
josephschorr: Relationships being added means the validation is running, which is a nice side test

miparnisari: but in this PR i don't care about validation of relationships, i care about the diffing. can we keep the test focused on one thing?

internal/services/shared/schema_test.go
miparnisari:
```
// TotalOperationCount holds the total number of "dispatch" operations performed by the schema
// being applied.
```

why would a WriteSchema imply a dispatch? 🫠 
this godoc is wrong, and also this seems wrong: https://github.com/authzed/spicedb/blob/d8f4e8fd88ed543c134c80972ea6aa6d1b30ccdf/internal/services/v1/schema.go#L154-L161

i'd remove this field altogether..

internal/services/shared/schema_test.go
tstirrat15: The datastore holds the schema, and in this case we need there to be stuff in the database that isn't being explicitly written by the call. I've removed the relationships part as it's not relevant to this test.

internal/services/shared/schema_test.go
tstirrat15: I agree; I get the sense that it arrived at its current state gradually. It's going to be removed soon :tm:  so I'm disinclined to do the work to make it pretty.

internal/services/shared/schema.go
```diff
@@ -26,6 +26,9 @@ type ValidatedSchemaChanges struct {
 	validatedDefinitions map[string]*schema.ValidatedDefinition
 	newCaveatDefNames    *mapz.Set[string]
 	newObjectDefNames    *mapz.Set[string]
+	// additiveOnly holds a marker of whether the set of changes described by
+	// this object is only additive (i.e. no deletions or changes) which can
+	// be used to make calling code easier to write.
```
miparnisari: this seems contradictory to what i see in the code. when you call `ValidateSchemaChanges(...)` you HAVE to pass in the value of `additiveOnly`, so it's not like this field is computed from the other fields, it just falls from the ~sky~ caller...

it seems to be a configuration flag that affects the behavior of `ApplySchemaChangesOverExisting`, to decide what to apply in the datastore

tstirrat15: Done

miparnisari (approved):

PR #2853, opened 2026-01-27
temp: write test for ApplySchemaChangesOverExisting but before #2805

tstirrat15:
## Description
This is mostly a proof of concept that shows that the same test as #2852 works the same way in the logic before #2805 was merged. This is not intended to be merged.

## Changes
* Add the same test (up to the schemareader usage) pre-#2805
## Testing
Check this out locally and see that tests pass.


tstirrat15:
Gonna close this - it's mostly for reference.

PR #2859, opened 2026-01-28
feat: plumb through datastore buffer margin

tstirrat15:
## Description
We have a buffer size defined down low in the watch implementation, but we don't have a way of controlling it from the outside. This provides that by adding a new flag and then plumbing it in.

## Notes
This felt like a wild number of changes for the description of the task. It feels like we ought to be able to reduce the number of intermediate objects between flag and DS implementation.

## Changes
Will annotate.

## Testing
See that tests pass. See that you can pass this flag.


internal/datastore/common/helpers.go
```diff
@@ -47,3 +59,45 @@ func ContextualizedCaveatFrom(name string, context map[string]any) (*core.Contex
...
+	return freeMem / 100 * parsedPercent, nil
+}
+
+// WatchBufferSize takes a string and interprets it as
+// either a percentage of memory (as a percentage of
+// 75% of free memory as measured on startup)
+// or a humanized byte string and returns the number of
+// bytes or an error if the value cannot be interpreted.
+// Returns 0 on an empty string.
+func WatchBufferSize(sizeString string) (size uint64, err error) {
```
tstirrat15, line 85: This takes the flag string and returns the uint64 that gets plumbed down.

internal/datastore/crdb/crdb.go
```diff
@@ -265,6 +265,7 @@ type crdbDatastore struct {
 	readPool, writePool     *pool.RetryPool
 	collectors              []prometheus.Collector
 	watchBufferLength       uint16
+	watchBufferMaximumSize  uint64
```
tstirrat15: The config value hangs off of the struct for all of these.

internal/datastore/crdb/options.go
```diff
@@ -16,6 +16,7 @@ type crdbOptions struct {
 	connectRate                 time.Duration
 
 	watchBufferLength              uint16
+	watchBufferMaximumSize         uint64
```
tstirrat15: But we also define it on the config

internal/datastore/crdb/watch.go
```diff
@@ -70,7 +70,7 @@ type changeDetails struct {
 
 func (cds *crdbDatastore) Watch(ctx context.Context, afterRevision datastore.Revision, options datastore.WatchOptions) (<-chan datastore.RevisionChanges, <-chan error) {
 	watchBufferLength := options.WatchBufferLength
-	if watchBufferLength <= 0 {
+	if watchBufferLength == 0 {
```
tstirrat15, line 73: This just bugged me. It's a `uint64` so `<=` isn't technically wrong, but it's cognitive dissonance.

internal/datastore/crdb/watch.go
```diff
@@ -344,7 +344,12 @@ func (cds *crdbDatastore) processChanges(ctx context.Context, changes pgx.Rows,
 			content:    opts.Content,
 		}
 	} else {
-		tracked = common.NewChanges(revisions.HLCKeyFunc, opts.Content, opts.MaximumBufferedChangesByteSize)
+		watchBufferSize := opts.MaximumBufferedChangesByteSize
+		if watchBufferSize == 0 {
+			watchBufferSize = cds.watchBufferMaximumSize
+		}
```
tstirrat15, line 350: This is where we actually grab the value that's passed down and in. There's an `opts.MaximumBufferedChangesByteSize` value, but I can't see where it's actually set.

internal/datastore/mysql/datastore.go
```diff
@@ -126,7 +126,7 @@ func NewReadOnlyMySQLDatastore(
 	return datastoreinternal.NewSeparatingContextDatastoreProxy(ds), nil
 }
 
-func newMySQLDatastore(ctx context.Context, uri string, replicaIndex int, options ...Option) (*Datastore, error) {
+func newMySQLDatastore(ctx context.Context, uri string, replicaIndex int, options ...Option) (*mysqlDatastore, error) {
```
tstirrat15, line 129: This also was a bit of a drive-by thing - it bugged me that this was the only implementation with a public `Datastore` object, so I renamed it to match the naming convention in other files.

internal/datastore/mysql/datastore.go
```diff
@@ -472,8 +472,10 @@ func newMySQLExecutor(tx querier, explainable datastore.Explainable) common.Exec
 }
 
 // Datastore is a MySQL-based implementation of the datastore.Datastore interface
-type Datastore struct {
+type mysqlDatastore struct {
+	*revisions.CachedOptimizedRevisions
```
tstirrat15, line 477: Just moving this to the top.

internal/datastore/mysql/datastore_test.go
```diff
@@ -152,8 +152,8 @@ func additionalMySQLTests(t *testing.T, b testdatastore.RunningEngineForTest) {
 }
 
 func LockingTest(t *testing.T, ds datastore.Datastore, ds2 datastore.Datastore) {
-	mds := ds.(*Datastore)
-	mds2 := ds2.(*Datastore)
+	mds := ds.(*mysqlDatastore)
+	mds2 := ds2.(*mysqlDatastore)
```
tstirrat15, line 156: Why are these casts necessary? I don't understand it.

pkg/cmd/datastore/datastore.go
```diff
@@ -533,6 +536,11 @@ func newCRDBDatastore(ctx context.Context, opts Config) (datastore.Datastore, er
 		return nil, errors.New("max-retries could not be cast to uint8")
 	}
 
+	watchBufferMaximumSize, err := common.WatchBufferSize(opts.WatchBufferMaximumSize)
```
tstirrat15: We use our helper up at the top so that we can invoke it once and use the value everywhere

pkg/cmd/datastore/datastore.go
```diff
@@ -739,11 +760,19 @@ func commonMySQLDatastoreOptions(opts Config) ([]mysql.Option, error) {
...
+	}
+
 	return []mysql.Option{
 		mysql.TablePrefix(opts.TablePrefix),
 		mysql.MaxRetries(maxRetries),
 		mysql.OverrideLockWaitTimeout(1),
 		mysql.WithEnablePrometheusStats(opts.EnableDatastoreMetrics),
+		mysql.WatchBufferLength(opts.WatchBufferLength),
+		mysql.WatchBufferWriteTimeout(opts.WatchBufferWriteTimeout),
+		mysql.WatchBufferMaximumSize(watchBufferMaximumSize),
```
tstirrat15: I don't know why these common bits weren't added to the common options. I'm just moving them around.

pkg/datastore/datastore.go
```diff
@@ -606,6 +606,11 @@ type WatchOptions struct {
 	// given the zero value, the datastore's default will be used.
 	WatchBufferLength uint16
 
+	// WatchBufferMaximumSize is the size of the buffer for the watch channel,
+	// provided as a humanized byte string or as a percentage. If set, overrides
+	// the value set for WatchBufferLength.
+	WatchBufferMaximumSize string
```
tstirrat15: I guess I'm not entirely sure that this is necessary - I'd previously assumed that `WatchOptions` was somehow plumbed in from the flags, but that doesn't appear to be the case. It does reflect the way that the other options are defined, though.

pkg/schema/definition_test.go
```diff
@@ -450,7 +450,7 @@ func TestDefinition(t *testing.T) {
 
 			ctx := t.Context()
 
-			predefinedNamespaces := make([]*core.NamespaceDefinition, 0)
+			predefinedNamespaces := make([]*core.NamespaceDefinition, 0, 1+len(tc.otherNamespaces))
```
tstirrat15, line 453: The `prealloc` linter was complaining at me locally about this.

internal/datastore/mysql/datastore.go
```diff
@@ -472,8 +472,10 @@ func newMySQLExecutor(tx querier, explainable datastore.Explainable) common.Exec
 }
 
 // Datastore is a MySQL-based implementation of the datastore.Datastore interface
-type Datastore struct {
+type mysqlDatastore struct {
```
vroldanbet, line 476: why the rename? it stutters. `mysql.mysqlDatastore`

pkg/datastore/datastore.go
vroldanbet: isn't this WatchBufferLength?

pkg/cmd/datastore/datastore.go
```diff
@@ -300,6 +302,7 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	flagSet.StringVar(&opts.MigrationPhase, flagName("datastore-migration-phase"), "", "datastore-specific flag that should be used to signal to a datastore which phase of a multi-step migration it is in")
 	flagSet.StringArrayVar(&opts.AllowedMigrations, flagName("datastore-allowed-migrations"), []string{}, "migration levels that will not fail the health check (in addition to the current head migration)")
 	flagSet.Uint16Var(&opts.WatchBufferLength, flagName("datastore-watch-buffer-length"), 1024, "how large the watch buffer should be before blocking")
```
vroldanbet, line 304: These two flags read the same. I'd suggest naming the new one `WatchChangeBufferMaximumSize`

pkg/cmd/datastore/datastore.go
```diff
@@ -300,6 +302,7 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
...
+	flagSet.StringVar(&opts.WatchBufferMaximumSize, flagName("datastore-watch-buffer-maximum-size"), "", "how much memory to reserve for the watch buffer, either as a humanized byte string value or a percentage")
```
vroldanbet: not sure "humanized byte string" is clear to the reader, also not sure that's the correct term. Perhaps add some examples? (128Mi, 20%)

vroldanbet: I'm torn about adding a default. It's a breaking change, but customers are also facing potential OOMKills when it goes wrong. I lean toward setting a 15% memory default here to add a sane default and protect the service.

docs/spicedb.md
```diff
@@ -123,6 +123,7 @@ spicedb datastore gc [flags]
       --datastore-tx-overlap-key string                                       static key to touch when writing to ensure transactions overlap (only used if --datastore-tx-overlap-strategy=static is set; CockroachDB driver only) (default "key")
       --datastore-tx-overlap-strategy string                                  strategy to generate transaction overlap keys ("request", "prefix", "static", "insecure") (CockroachDB driver only - see https://spicedb.dev/d/crdb-overlap for details) (default "static")
       --datastore-watch-buffer-length uint16                                  how large the watch buffer should be before blocking (default 1024)
+      --datastore-watch-buffer-maximum-size string                            how much memory to reserve for the watch buffer, either as a humanized byte string value or a percentage
```
miparnisari: this isn't clear to me, this flag and the one above seem to control the same thing, why add a new one?

(it seems like `watchBufferLength` is useless, or i don't understand as a customer how i would set it)

also do these flags apply to all datastores?

internal/datastore/crdb/options.go
```diff
@@ -264,6 +265,12 @@ func WatchBufferWriteTimeout(watchBufferWriteTimeout time.Duration) Option {
 	return func(po *crdbOptions) { po.watchBufferWriteTimeout = watchBufferWriteTimeout }
 }
 
+// WatchBufferMaximumSize is the maximum size in bytes of the watch buffer.
+// If this value is exceeded the caller will receive an error.
```
miparnisari, line 269: lets document this as part of the flag description so users know the consequence of setting this value

internal/datastore/crdb/watch.go
miparnisari: This is silly. I looked around and i also can't find where it is set, neither `MaximumBufferedChangesByteSize` nor a bunch of others. I think the point of `datastore.WatchOptions` might have been to specify client-side options, but we've been blindly adding also server options for no reason 😭

internal/datastore/mysql/datastore.go
miparnisari: one upside is that if you are looking for implementations of the `datastore` interface, this implementation shows up more nicely and you aren't forced to look at the name of the file to know which one it is. This is an issue because there are a dozen implementations 😅 

image

tstirrat15: Because we do the same in other packages - postgres has `pgDatastore`, crdb has `crdbDatastore`, etc.

There isn't a nice way to shorten `mysql` that I can think of.

You're also not going to be referencing this from outside of this package.

internal/datastore/mysql/datastore_test.go
miparnisari: probably because we are using internal methods of `mysqlDatastore`

pkg/cmd/datastore/datastore.go
tstirrat15: Yeah, I'm happy to add some examples, and I wasn't happy with the phrase I came up with. I haven't come up with a better name for `20Gi` though.

docs/spicedb.md
```diff
@@ -123,6 +123,7 @@ spicedb datastore gc [flags]
       --datastore-tx-overlap-key string                                       static key to touch when writing to ensure transactions overlap (only used if --datastore-tx-overlap-strategy=static is set; CockroachDB driver only) (default "key")
       --datastore-tx-overlap-strategy string                                  strategy to generate transaction overlap keys ("request", "prefix", "static", "insecure") (CockroachDB driver only - see https://spicedb.dev/d/crdb-overlap for details) (default "static")
       --datastore-watch-buffer-length uint16                                  how large the watch buffer should be before blocking (default 1024)
```
miparnisari, line 125: before blocking what?

docs/spicedb.md
```diff
@@ -123,6 +123,7 @@ spicedb datastore gc [flags]
...
+      --datastore-watch-buffer-maximum-size string                            how much memory to reserve for the watch buffer, either as a humanized byte string value or a percentage
       --datastore-watch-buffer-write-timeout duration                         how long the watch buffer should queue before forcefully disconnecting the reader (default 1s)
       --datastore-watch-connect-timeout duration                              how long the watch connection should wait before timing out (CockroachDB driver only) (default 1s)
```
miparnisari:
```suggestion
      --datastore-watch-connect-timeout duration                              how long the watch connection to the underlying datastore should wait before timing out (CockroachDB driver only) (default 1s)
```

miparnisari (commented):
not your fault, and not for this PR, but there's a lot of this going on: (😓) 

```
    var := options.var
	if var != something {
		var = cds.var
	}
```

which is funny because, most of these aren't really options as in "client-specifiable" options, these are server options, so presumably you will always enter the "if" in production. 

i think a refactor is due..
- split `datastore.WatchOptions` struct to `ServerWatchOptions` and `ClientWatchOptions` 
- add a new `datastore.WatchOptionsDefaults` interface method that must be overriden per datastore
- introduce a common fn `merge(serverOptions, defaultDsOptions, clientWatchOptions) -> ClientAndServerOptions`. also, `serverWatchOptions` are the same no matter the datastore, so they don't need to live in the `cds` or `memdb` implementations, they could live inside `watchServer`, 
- have `server.Watch()` call the merge function
- then we change the `datastore.Watch` interface to take `ClientAndServerOptions`

pkg/cmd/datastore/datastore.go
tstirrat15: Done

internal/datastore/mysql/datastore.go
vroldanbet: fair!

PR #2860, opened 2026-01-29
Remove CGO Dependency from PostgreSQL FDW

ecordell:
## Description

This PR removes the CGO dependency from SpiceDB's PostgreSQL Foreign Data Wrapper (FDW) feature by replacing `pganalyze/pg_query_go` with `wasilibs/go-pgquery`, a WebAssembly-based drop-in replacement. This change simplifies builds across all platforms, eliminates C compiler requirements, and enables true static binary compilation.

I started looking into this when I tried to pull in another dependency that had a conflict at the C layer (you can't import two CGO libraries that import the same 3rd party C lib, the linker gives up).

I found wasilibs/go-pgquery which compiles pg_query to wasm and loads it that way. If you do build with CGO though, it links in the same C library, so this just gives flexibility.

I would expect the parsing to be marginally slower via wasm, but that should not be the bottleneck in the fdw.

The wasilib seems maintained but doesn't follow any sort of release or versioning, so I'm presenting this for discussion.

## Testing

I didn't see any failures after swapping the library out, but we may want to consider further testing in real envs.

## References

https://github.com/wasilibs/go-pgquery

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,10 @@
 # use `docker buildx imagetools inspect <image>` to get the multi-platform sha256
-# Note: Using Debian-based golang image (not alpine) to build with glibc for postgres-fdw support
-FROM golang:1.25.6@sha256:ce63a16e0f7063787ebb4eb28e72d477b00b4726f79874b3205a965ffd797ab2 AS spicedb-builder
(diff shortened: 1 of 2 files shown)


vroldanbet (approved):
LGTM

tstirrat15:
Looking at the benchmarks, it seems like we can expect about a 5x slowdown, but that doesn't seem terrible given the context.

PR #2863, opened 2026-01-29
chore: use protovalidate interceptor instead of protoc-gen-validate interceptor

tstirrat15:
## Description
Development on `protoc-gen-validate` has basically stopped in favor of [protovalidate](https://protovalidate.com/). It's intended to be a more consistent and flexible validation framework.

This does mean that it will be slightly slower, but not slower enough to be meaningful: https://bufbuild.slack.com/archives/CRZ680FUH/p1760037391133719?thread_ts=1759933966.935909&cid=CRZ680FUH

## Changes
* Use protovalidate middleware

## Testing
Review. See that validation still happens as expected.


internal/services/v1/experimental.go
```diff
@@ -93,16 +94,18 @@ func NewExperimentalServer(dispatch dispatch.Dispatcher, permServerConfig Permis
 		chunkSize = 100
 	}
 
+	validator := genutil.MustNewProtoValidator()
+
 	return &experimentalServer{
 		WithServiceSpecificInterceptors: shared.WithServiceSpecificInterceptors{
 			Unary: middleware.ChainUnaryServer(
-				grpcvalidate.UnaryServerInterceptor(),
+				grpcvalidate.UnaryServerInterceptor(validator),
```
miparnisari, line 102: please run a couple of manual tests to ensure that the validation still works as expected (`docker-compose up --build` for example). Try both HTTP and gRPC endpoints

i'm also wary of the `handwrittenvalidation` below. i don't know what it does but i'd want to make sure we didn't accidentally break it

PR #2864, opened 2026-01-29
fix: update IterSubjects for wildcards and Alias iterators for confomance

barakmich:

### DIFF
--- a/pkg/query/alias.go
+++ b/pkg/query/alias.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/datastore/options"
-	"github.com/google/uuid"
 )
 
 // Alias is an iterator that rewrites the Resource's Relation field of all paths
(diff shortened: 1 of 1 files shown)


internal/datastore/memdb/readonly.go
```diff
@@ -505,15 +505,12 @@ func newSubjectSortedIterator(now time.Time, it memdb.ResultIterator, limit *uin
...
-		if rt.OptionalExpiration != nil && rt.OptionalExpiration.Before(now) {
-			continue
-		}
-
 		if skipCaveats && rt.OptionalCaveat != nil {
 			continue
 		}
 
-		if skipExpiration && rt.OptionalExpiration != nil {
+		// Only check expiration if skipExpiration is false
```
tstirrat15, line 512: Oh that's kinda fun

pkg/query/alias.go
```diff
@@ -24,60 +26,46 @@ func NewAlias(relation string, subIt Iterator) *Alias {
...
-				return nil, err
-			}
-
-			// Create combined sequence with self-edge and rewritten paths
-			combined := func(yield func(Path, error) bool) {
-				// Yield the self-edge first
-				if !yield(selfPath, nil) {
+// maybePrependSelfEdge checks if a self-edge should be added and combines it with the given sequence.
+// A self-edge is added when the resource with the alias relation matches the subject.
+func (a *Alias) maybePrependSelfEdge(resource Object, subSeq PathSeq, shouldAddSelfEdge bool) PathSeq {
```
tstirrat15, line 32: This is the helper that we were talking about splitting out?

pkg/query/alias.go
```diff
@@ -91,84 +79,111 @@ func (a *Alias) CheckImpl(ctx *Context, resources []Object, subject ObjectAndRel
...
 
-	// else we don't have a subpath and do the normal logic
-	return func(yield func(Path, error) bool) {
-		for path, err := range subSeq {
-			if err != nil {
-				yield(Path{}, err)
-				return
-			}
+	iter, err := ctx.Reader.QueryRelationships(ctx, filter,
+		options.WithLimit(options.LimitOne),
```
tstirrat15, line 160: Wow, so the check logic is conditionally running to the datastore like this? Or is this more an artifact of the query planner having a different structure?

pkg/query/datastore.go
```diff
@@ -231,13 +231,18 @@ func (r *RelationIterator) iterSubjectsWildcardImpl(ctx *Context, resource Objec
...
-	// Note: This may return some of the same subjects as the non-wildcard branch
-	// (when both wildcard and concrete relationships exist), but the Union will
-	// deduplicate them.
+	// Wildcard exists, so enumerate all concrete subjects of the appropriate type.
+	// A wildcard (e.g., user:*) means "all subjects of that type", so we need to enumerate
+	// all defined subjects of that type in the datastore. This may return some of the same
+	// subjects as the non-wildcard branch (when both wildcard and concrete relationships exist),
+	// but the Union will deduplicate them.
+	//
+	// Note: We query for all subjects of the appropriate type, not just those with a relationship
```
tstirrat15, line 240: Wow... this feels kinda wild.

tstirrat15 (commented):
Boy these are weird. I appreciate the way you documented them and tested them.

pkg/query/alias.go
barakmich: yep!

pkg/query/alias.go
barakmich: More the different structure. Whereas in other LS implementations we just fetch where necessary, and in QP we've tried to keep it at the leaves. Unfortunately, this is something where the semantics break the expectation 😞

tstirrat15: sadface

PR #2865, opened 2026-01-29
ci: compare Go benchmarks properly

miparnisari:
Fixing an issue where PR's are not comparing against `main`.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -417,7 +417,7 @@ jobs:
 
   benchmarks:
     name: "Run Go benchmarks"
-    runs-on: "depot-ubuntu-24.04-arm-8"
+    runs-on: "depot-ubuntu-24.04-arm-8" # same as benchmark.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yaml
```diff
@@ -456,7 +452,7 @@ jobs:
         id: "downloadbenchmark"
         with:
           path: "./cache/benchmark-data.json"
-          key: "${{ steps.get-main-branch-sha.outputs.sha }}-${{ runner.os }}-${{ steps.system-info.outputs.cpu-model }}-go-benchmark"
+          key: "${{ steps.get-main-branch-sha.outputs.sha }}-${{ runner.os }}-go-benchmark"
```
miparnisari, line 455: this was returning `unknown`, https://github.com/authzed/spicedb/actions/runs/21220346253/job/61055195969?pr=2805#step:7:13

which caused the cache key to not be found, therefore the comparison wasn't being done

image

PR #2868, opened 2026-01-30
fix: hook up schema caching proxies with schemareader

tstirrat15:
## Description
In #2805 we refactored the caching proxies, and a part of that refactor made `SchemaReader()` return the underlying `Reader` instance rather than a reference to the current proxy. This meant that the proxy was effectively disabled for anything that interacted with schema reading/writing, because there wasn't a way to get a handle on the proxy that exercised the schema reading logic.

This remedies that by returning a wrapped version of the proxy that adds adapters for the new methods.

It also breaks out the `LookupSchemaDefinitions`'s definition into two other functions that are caveat and typedef specific, which means that we can use the specific version at the callsite and prevent cache misses and overfetching.

## Changes
Will annotate.

## Testing
Review. Load-test this against the previous commit and see that the schema cache is actually populated and used.

### DIFF
--- a/pkg/datastore/schema.go
+++ b/pkg/datastore/schema.go
@@ -46,7 +46,7 @@ type SchemaReader interface {
 	// LookupTypeDefinitionsByNames looks up type definitions by name.
 	LookupTypeDefinitionsByNames(ctx context.Context, names []string) (map[string]SchemaDefinition, error)
 
-	// LookupCaveatDefinitionsByNames looks up type definitions by name.
+	// LookupCaveatDefinitionsByNames looks up caveat definitions by name.
 	LookupCaveatDefinitionsByNames(ctx context.Context, names []string) (map[string]SchemaDefinition, error)
 }
 


internal/datastore/proxy/schemacaching/standardcache.go
```diff
@@ -99,8 +100,11 @@ func (r *definitionCachingReader) LegacyLookupCaveatsWithNames(
 		estimatedCaveatDefinitionSize)
 }
 
+// SchemaReader returns a reference to this reader, but wrapped in such a way
+// that the legacy methods are used to drive the new methods, which ensures
+// that caching logic stays in place.
 func (r *definitionCachingReader) SchemaReader() (datastore.SchemaReader, error) {
-	return r.Reader.SchemaReader()
+	return schemautil.NewLegacySchemaReaderAdapter(r), nil
```
tstirrat15, line 107: I took the expedient approach in this file under the understanding that this code will be completely discarded once new schema storage is in place.

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -603,6 +603,116 @@ type watchingCachingReader struct {
 	p   *watchingCachingProxy
 }
 
+var _ datastore.SchemaReader = (*watchingCachingReader)(nil)
```
tstirrat15: I was slightly more thorough in this file under the understanding that the new schema storage mechanism will reuse some of this logic.

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -603,6 +603,116 @@ type watchingCachingReader struct {
...
+	}
+	return datastore.RevisionedCaveat{
+		Definition: def,
+		LastWrittenRevision: revision,
+	}, true, nil
+}
+
+// ListAllTypeDefinitions delegates to the underlying reader, as the cache doesn't currently
+// support reading all definitions directly.
+// TODO: implement this in the cache
```
tstirrat15: Question for reviewers: does this seem sane? Listing all will bypass the cache, but I don't have a sense of whether that's actually used in places where caching matters.

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -603,6 +603,116 @@ type watchingCachingReader struct {
...
+	reader, err := w.Reader.SchemaReader()
+	if err != nil {
+		return nil, err
+	}
+	return reader.ListAllSchemaDefinitions(ctx)
+}
+
+// LookupSchemaDefinitionsByNames looks up type and caveat definitions by name.
+// TODO: implement this directly in the cache, rather than overfetching from the caches
+func (w *watchingCachingReader) LookupSchemaDefinitionsByNames(ctx context.Context, names []string) (map[string]datastore.SchemaDefinition, error) {
```
tstirrat15: This function is slightly inefficient because it's always going to make twice as many requests as necessary, but the caches should hold a `notFound` after their first request so it should be amortized (?).

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -53,7 +58,7 @@ func TestNilUnmarshal(t *testing.T) {
 	require.Equal(t, nsDef, newDef)
 }
 
-type testerDef struct {
+type oldTesterDef struct {
```
tstirrat15, line 60: Renaming here was mostly about ensuring that I didn't miss a test.

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -208,8 +213,162 @@ func TestSnapshotCaching(t *testing.T) {
 	}
 }
 
-func TestRWTCaching(t *testing.T) {
-	for _, tester := range testers {
+func TestSnapshotCaching(t *testing.T) {
```
tstirrat15, line 215: Most of the refactors in these two files look roughly like this - flattening the tables and refactoring to use the new methods.

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -208,8 +213,162 @@ func TestSnapshotCaching(t *testing.T) {
...
+	dsMock := &proxy_test.MockDatastore{}
+
+	oneReader := &proxy_test.MockReader{}
+	dsMock.On("SnapshotReader", one).Return(oneReader)
+	oneReader.On("LegacyReadNamespaceByName", nsA).Return(nil, old, nil).Once()
```
tstirrat15, line 220: Note that the mock still references the legacy method - this is because the wrapper is just invoking the legacy method.

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -331,7 +544,62 @@ func TestSingleFlight(t *testing.T) {
...
-func TestSnapshotCachingRealDatastore(t *testing.T) {
+func TestSingleFlight(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		dsMock := &proxy_test.MockDatastore{}
+
+		oneReader := &proxy_test.MockReader{}
+		dsMock.On("SnapshotReader", one).Return(oneReader)
+		oneReader.
+			On("LegacyReadNamespaceByName", nsA).
+			WaitUntil(time.After(50*time.Millisecond)).
```
tstirrat15, line 553: Using `synctest` here to avoid needing to actually sleep for the given period of time.

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -409,28 +677,138 @@ func TestSnapshotCachingRealDatastore(t *testing.T) {
...
+			caveatRevDef2, _, err := schemaReader.LookupCaveatDefByName(ctx, tc.caveatName)
+			require.NoError(t, err)
+			testutil.RequireProtoEqual(t, tc.caveatDef, caveatRevDef2.Definition, "found different caveats")
+		})
+	}
+}
+
+// singleflightReader is used to test singleflight context cancellation
+// behavior.
+type singleflightReader struct {
```
tstirrat15, line 770: Renaming this to make its use a bit clearer.

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -527,9 +949,94 @@ func TestMixedCaching(t *testing.T) {
 	}
 }
 
+// TODO
+func TestMixedCaching(t *testing.T) {
```
tstirrat15, line 959: This is the test whose semantics differ the most from the test it's based on.

Most of it comes down to the `LookupSchemaDefinitionsByNames` function taking namespace *or* caveat names as args, which means that it needs to look up the names as both caveats and namespaces.

Because the caches don't hold negative results, you can't make the assertion that the lookup function doesn't go to the underlying reader, because the lookup function is always going to be looking in the wrong place for a name half of the time.

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -603,34 +604,34 @@ type watchingCachingReader struct {
 	p   *watchingCachingProxy
 }
 
-func (r *watchingCachingReader) LegacyReadNamespaceByName(
+func (w *watchingCachingReader) LegacyReadNamespaceByName(
```
tstirrat15, line 607: This bugged me so I updated it.

internal/datastore/proxy/schemacaching/watchingcache_test.go
```diff
@@ -123,7 +124,142 @@ func TestWatchingCacheBasicOperation(t *testing.T) {
...
+	reader = wcache.SnapshotReader(rev("3.0000000005"))
+	schemaReader, err = reader.SchemaReader()
+	require.NoError(t, err)
+
+	// Read again, which should now be via the cache.
+	nsRevDef, _, err = schemaReader.LookupTypeDefByName(t.Context(), "somenamespace")
+	require.NoError(t, err)
+	require.Equal(t, "somenamespace", nsRevDef.Definition.Name)
+
+	// NOTE: we don't read via lookup because we can't retain the semantic of the test
```
tstirrat15: This test's semantic also differs from the previous one, for the same reason as previously - we can't assert that the Lookup doesn't go to the underlying datastore because it's always going to miss the cache half the time.

internal/datastore/proxy/schemacaching/standardcache.go
miparnisari: >this code will be completely discarded

this entire file?

josephschorr: Yes

internal/datastore/proxy/schemacaching/standardcaching_test.go
miparnisari, line 1: if you undo the changes in `internal/datastore/proxy/schemacaching/standardcache.go`, do any of these tests fail?

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -331,7 +543,68 @@ func TestSingleFlight(t *testing.T) {
...
+		}
+
+		// NOTE: if the singleflight isn't working, the second call to land will invoke
+		// the mock a second time and panic.
+		go readNamespace()
+		go readNamespace()
+		go readCaveat()
+		go readCaveat()
+
+		// Let the timers elapse
```
miparnisari, line 596: what timer?

internal/datastore/proxy/schemacaching/standardcaching_test.go
```diff
@@ -466,8 +849,55 @@ func TestSingleFlightCancelled(t *testing.T) {
...
 }
 
-func TestMixedCaching(t *testing.T) {
-	for _, tester := range testers {
+func TestSingleFlightCancelled(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		dsMock := &proxy_test.MockDatastore{}
+		ctx1, cancel1 := context.WithCancel(t.Context())
+		defer cancel1()
+		// Note that ctx2 will not be cancelled
```
miparnisari, line 856: ?
> t.Context returns a context that is canceled just before Cleanup-registered functions are called.

internal/datastore/proxy/schemacaching/standardcaching_test.go
miparnisari: nit: don't name things based on how they are expected to be used, name them after what they do. in this case, it's a slow reader that reacts to context cancellation

internal/datastore/proxy/schemacaching/standardcaching_test.go
tstirrat15: Yes - I verified that as a part of the work. None of them work because they're going directly to the underlying mock all the time.

internal/caveats/run.go
```diff
@@ -112,7 +112,7 @@ func (cr *CaveatRunner) PopulateCaveatDefinitionsForExpr(ctx context.Context, ex
 	if err != nil {
 		return err
 	}
-	foundDefs, err := schemaReader.LookupSchemaDefinitionsByNames(ctx, caveatNames.AsSlice())
+	foundDefs, err := schemaReader.LookupCaveatDefinitionsByNames(ctx, caveatNames.AsSlice())
```
tstirrat15, line 115: Using the specific lookup should prevent an NS lookup.

internal/datastore/proxy/schemacaching/standardcaching_test.go
tstirrat15: The timers on the singleflight reader mock

internal/datastore/proxy/schemacaching/standardcaching_test.go
tstirrat15: In the context of the test :P I meant that the second context isn't imperatively cancelled during test execution

internal/datastore/proxy/schemacaching/standardcaching_test.go
tstirrat15: I was able to remedy this by adding separate `Lookup` methods that return just caveats and namespaces.

internal/datastore/proxy/schemacaching/watchingcache_test.go
```diff
@@ -123,7 +124,155 @@ func TestWatchingCacheBasicOperation(t *testing.T) {
...
+	schemaReader, err = reader.SchemaReader()
+	require.NoError(t, err)
+
+	// Read again, which should now be via the cache.
+	nsRevDef, _, err = schemaReader.LookupTypeDefByName(t.Context(), "somenamespace")
+	require.NoError(t, err)
+	require.Equal(t, "somenamespace", nsRevDef.Definition.Name)
+
+	// Repeat with lookup
+	nsDefMap, err := schemaReader.LookupTypeDefinitionsByNames(t.Context(), []string{"somenamespace"})
```
tstirrat15, line 218: Same in this file - using the new methods means that the new test better reflects the old one.

internal/datastore/schema/schema.go
```diff
@@ -151,6 +153,30 @@ func (l *LegacySchemaReaderAdapter) ListAllSchemaDefinitions(ctx context.Context
 
 // LookupSchemaDefinitionsByNames looks up type and caveat definitions by name.
 func (l *LegacySchemaReaderAdapter) LookupSchemaDefinitionsByNames(ctx context.Context, names []string) (map[string]datastore.SchemaDefinition, error) {
```
tstirrat15, line 155: I rewrote this function in terms of the new lookup functions.

internal/namespace/util.go
```diff
@@ -73,7 +73,7 @@ func CheckNamespaceAndRelations(ctx context.Context, checks []TypeAndRelationToC
 	if err != nil {
 		return err
 	}
-	foundDefs, err := schemaReader.LookupSchemaDefinitionsByNames(ctx, nsNames.AsSlice())
+	foundDefs, err := schemaReader.LookupTypeDefinitionsByNames(ctx, nsNames.AsSlice())
```
tstirrat15, line 76: This only handles typedefs so we can use the more specific function.

internal/relationships/validation.go
```diff
@@ -101,37 +101,42 @@ func loadNamespacesAndCaveats(ctx context.Context, rels []tuple.Relationship, re
 	var referencedNamespaceMap map[string]*schema.Definition
 	var referencedCaveatMap map[string]*core.CaveatDefinition
 
-	if !referencedNamespaceNames.IsEmpty() || !referencedCaveatNamesWithContext.IsEmpty() {
```
tstirrat15, line 104: I'd welcome another pair of eyes on this file.

The old logic combined the namespace and caveat names and then made a single Lookup call. The new one keeps them separate and handles them separately, which should reduce the work and also makes the code slightly easier to follow at the expense of some duplication.

internal/services/v1/experimental.go
```diff
@@ -265,17 +265,12 @@ func (es *experimentalServer) BulkImportRelationships(stream v1.ExperimentalServ
 			numWritten += streamWritten
 
 			// The stream has terminated because we're awaiting namespace and/or caveat information
-			if len(adapter.awaitingNamespaces) > 0 || len(adapter.awaitingCaveats) > 0 {
```
tstirrat15, line 268: This is the exact same refactor.

internal/services/v1/permissions.go
```diff
@@ -1073,17 +1073,12 @@ func (ps *permissionServer) ImportBulkRelationships(stream grpc.ClientStreamingS
 			numWritten += streamWritten
 
 			// The stream has terminated because we're awaiting namespace and/or caveat information
-			if len(adapter.awaitingNamespaces) > 0 || len(adapter.awaitingCaveats) > 0 {
```
tstirrat15, line 1076: And this too.

pkg/datastore/schema.go
```diff
@@ -42,6 +42,12 @@ type SchemaReader interface {
 
 	// LookupSchemaDefinitionsByNames looks up type and caveat definitions by name.
 	LookupSchemaDefinitionsByNames(ctx context.Context, names []string) (map[string]SchemaDefinition, error)
+
+	// LookupTypeDefinitionsByNames looks up type definitions by name.
+	LookupTypeDefinitionsByNames(ctx context.Context, names []string) (map[string]SchemaDefinition, error)
```
tstirrat15, line 47: These are the new methods added.

josephschorr (approved):
LGTM

pkg/datastore/schema.go
```diff
@@ -42,6 +42,12 @@ type SchemaReader interface {
...
+
+	// LookupCaveatDefinitionsByNames looks up type definitions by name.
```
miparnisari:
```suggestion
	// LookupCaveatDefinitionsByNames looks up caveat definitions by name.
```

pkg/datastore/schema.go
```diff
@@ -42,6 +42,12 @@ type SchemaReader interface {
 
 	// LookupSchemaDefinitionsByNames looks up type and caveat definitions by name.
 	LookupSchemaDefinitionsByNames(ctx context.Context, names []string) (map[string]SchemaDefinition, error)
```
miparnisari, line 44: `LookupSchemaDefinitionsByNames` isn't used anymore (well except in tests). Can we mark it as deprecated with pointers to the new ones?

internal/relationships/validation.go
```diff
@@ -101,37 +101,42 @@ func loadNamespacesAndCaveats(ctx context.Context, rels []tuple.Relationship, re
...
 
-		foundDefs, err := schemaReader.LookupSchemaDefinitionsByNames(ctx, allNames)
+	if !referencedNamespaceNames.IsEmpty() {
+		foundNamespaceDefs, err := schemaReader.LookupTypeDefinitionsByNames(ctx, referencedNamespaceNames.AsSlice())
 		if err != nil {
 			return nil, nil, err
 		}
-
+		referencedNamespaceMap = make(map[string]*schema.Definition, len(foundNamespaceDefs))
 		ts := schema.NewTypeSystem(schema.ResolverForDatastoreReader(reader))
```
miparnisari, line 115: this is funny.. if you remove `ts` and pass `schema.NewDefinition(nil,...)` the unit tests still pass 💀 

not saying you change that but i think the `schema.Definition` struct needs a review

internal/relationships/validation.go
miparnisari: This looks good to me

pkg/datastore/schema.go
tstirrat15: We'll use it when we move to the new schema storage mechanism, because it will make sense when the cost is the same to go look at the entire schema.

PR #2874, opened 2026-02-03
 chore: add metrics and tests to all cache implementations

miparnisari:


pkg/cache/cache_ristretto.go
miparnisari, line 1: FYI this is the one used in prod today

pkg/cache/cache_ristretto.go
```diff
@@ -43,32 +43,32 @@ func NewRistrettoCacheWithMetrics[K KeyString, V any](name string, config *Confi
 		return nil, err
 	}
 
-	cache := wrapped[K, V]{name, config, config.DefaultTTL, rcache}
+	cache := ristretoCache[K, V]{name, config, config.DefaultTTL, rcache}
```
tstirrat15, line 46:
```suggestion
	cache := ristrettoCache[K, V]{name, config, config.DefaultTTL, rcache}
```
Plus other updates

pkg/cache/cache_otter.go
```diff
@@ -67,13 +81,23 @@ func (wtc *otterCache[K, V]) Set(key K, value V, cost int64) bool {
 		// was too big, so we set to maxint in that case.
 		uintCost = math.MaxUint32
 	}
+
 	wtc.metrics.costAdded.Add(uint64(uintCost))
+	_, ok := wtc.Get(key)
```
tstirrat15, line 86: I wonder if there's a meaningful difference between Otter and Ristretto such that this bookkeeping isn't required?

pkg/cache/cache_ristretto.go
miparnisari: follow-up PR :D

pkg/cache/cache_otter.go
miparnisari: More generally i think the semantics of this entire Cache interface are unclear. What is Set supposed to return if the key exists? 

Theine overwrites.
Otter fails if it already exists.
Ristretto overwrites, i think.

I don't know how the callers expect Set to behave to say how each impl _should_ behave.

PR #2883, opened 2026-02-05
fix: server not dying when streaming api is still running

miparnisari:
## Description

Have the server die when you kill it when a streaming api (e.g. Watch) is still running.

## Testing

1. On tab 1, run `go run ./cmd/spicedb/main.go serve --grpc-preshared-key foobar`
2. On tab 2, run `zed watch`
3. On tab 1, ctrl+c the server. The server should die.


miparnisari:
Closing because we need to fix the `--grpc-shutdown-grace-period` instead.

PR #2884, opened 2026-02-05
fix: handle `self` keyword in warnings checks, and check these warnings are error-free in consistency

barakmich:


internal/services/integrationtesting/consistency_test.go
```diff
@@ -250,6 +250,12 @@ func runConsistencyTestSuiteForFile(t *testing.T, filePath string, useCachingDis
 	for _, nsDef := range cad.Populated.NamespaceDefinitions {
 		_, err := ts.GetValidatedDefinition(cad.Ctx, nsDef.Name)
 		require.NoError(t, err)
+
+		_, err = development.AddDefinitionWarnings(cad.Ctx, nsDef, ts)
```
miparnisari, line 260: should we also have this check after line 95?

tstirrat15 (approved):
LGTM

PR #2886, opened 2026-02-06
build: add new mustcallcheck analyzer and fix instances

tstirrat15:
## Description
Follow-on to #2878. We realized we could write an analyzer that checks for that condition explicitly, so we did that and fixed the findings.

## Changes
* Add `mustcallcheck` analyzer
* Add tests for the analyzer
* Fix what the analyzer found
## Testing
Review. See that analyzers and tests pass.

### DIFF
--- a/internal/datasets/subjectsetbytype.go
+++ b/internal/datasets/subjectsetbytype.go
@@ -58,7 +58,10 @@ func (s *SubjectByTypeSet) ForEachType(handler func(rr *core.RelationReference, 
 func (s *SubjectByTypeSet) Map(mapper func(rr *core.RelationReference) (*core.RelationReference, error)) (*SubjectByTypeSet, error) {
 	mapped := NewSubjectByTypeSet()
 	for key, subjectset := range s.byType {
-		ns, rel := tuple.MustSplitRelRef(key)
+		ns, rel, err := tuple.SplitRelRef(key)
+		if err != nil {
+			return nil, err
+		}
 		updatedType, err := mapper(&core.RelationReference{
 			Namespace: ns,
 			Relation:  rel,
--- a/pkg/cmd/server/cacheconfig.go
+++ b/pkg/cmd/server/cacheconfig.go
@@ -137,9 +137,8 @@ func parsePercent(str string, freeMem uint64) (uint64, error) {
 	return freeMem / 100 * parsedPercent, nil
 }
 
-// MustRegisterCacheFlags registers flags used to configure SpiceDB's various
-// caches.
-func MustRegisterCacheFlags(flags *pflag.FlagSet, flagPrefix, flagDescription string, config, defaults *CacheConfig) {
+// RegisterCacheFlags registers flags used to configure SpiceDB's various caches.
+func RegisterCacheFlags(flags *pflag.FlagSet, flagPrefix, flagDescription string, config, defaults *CacheConfig) error {
 	config.Name = defaults.Name
 	flagPrefix = cmp.Or(flagPrefix, "cache")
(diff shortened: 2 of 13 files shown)


internal/datasets/subjectsetbytype.go
```diff
@@ -58,7 +58,10 @@ func (s *SubjectByTypeSet) ForEachType(handler func(rr *core.RelationReference,
 func (s *SubjectByTypeSet) Map(mapper func(rr *core.RelationReference) (*core.RelationReference, error)) (*SubjectByTypeSet, error) {
 	mapped := NewSubjectByTypeSet()
 	for key, subjectset := range s.byType {
-		ns, rel := tuple.MustSplitRelRef(key)
+		ns, rel, err := tuple.SplitRelRef(key)
```
tstirrat15, line 61: I added this new function and then used it where it was appropriate.

pkg/cmd/server/cacheconfig.go
```diff
@@ -137,9 +129,8 @@ func parsePercent(str string, freeMem uint64) (uint64, error) {
 	return freeMem / 100 * parsedPercent, nil
 }
 
-// MustRegisterCacheFlags registers flags used to configure SpiceDB's various
-// caches.
-func MustRegisterCacheFlags(flags *pflag.FlagSet, flagPrefix, flagDescription string, config, defaults *CacheConfig) {
+// RegisterCacheFlags registers flags used to configure SpiceDB's various caches.
+func RegisterCacheFlags(flags *pflag.FlagSet, flagPrefix, flagDescription string, config, defaults *CacheConfig) error {
```
tstirrat15, line 141: This probably would have been fine as an exception because the error case is hard to hit, but it also wasn't too bad to make it return an error.

pkg/development/devcontext.go
```diff
@@ -148,12 +148,17 @@ func newDevContextWithDatastore(ctx context.Context, requestContext *devinterfac
 		RelationshipChunkCache: nil, // Disable caching for devcontext
 	}
 
+	dispatcher, err := graph.NewLocalOnlyDispatcher(params)
+	if err != nil {
+		return nil, nil, err
+	}
```
tstirrat15, line 154: Similar here - hard to hit, but also probably not a problem.

tools/analyzers/mustcallcheck/mustcallcheck.go
```diff
@@ -0,0 +1,178 @@
...
+		// MustNewInterceptorPooler comes from pgx and doesn't have an alternative
+		"MustNewInterceptorPooler",
+		// MustRequirePresharedKey is intended to be a guardrail if the preshared key isn't
+		// present at the callsite.
+		"MustRequirePresharedKey",
+	}
+	packageAllowList = []string{
+		// cobrautil has many Must* functions but doesn't have non-Must variants
+		"cobrautil",
+	}
```
tstirrat15, line 38: This is the list of exceptions - take a look and lemme know what you think.

tools/analyzers/mustcallcheck/mustcallcheck.go
```diff
@@ -0,0 +1,178 @@
...
+		// present, it means there's probably a bug and a panic is appropriate.
+		"MustValue",
+		// MustNewInterceptorPooler comes from pgx and doesn't have an alternative
+		"MustNewInterceptorPooler",
+		// MustRequirePresharedKey is intended to be a guardrail if the preshared key isn't
+		// present at the callsite.
+		"MustRequirePresharedKey",
+	}
+	packageAllowList = []string{
+		// cobrautil has many Must* functions but doesn't have non-Must variants
```
miparnisari, line 36: 😭

tstirrat15: Yeah, but I'm also not too worried about it because they typically happen during process startup and if they panic it indicates a bug in the code rather than invalid input.

PR #2891, opened 2026-02-09
feat: add recursive direction strategies, and fix IS BFS

barakmich:

### DIFF
--- a/pkg/query/benchmarks/check_deep_arrow_benchmark_test.go
+++ b/pkg/query/benchmarks/check_deep_arrow_benchmark_test.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/internal/datastore/memdb"
 	"github.com/authzed/spicedb/pkg/datastore"
@@ -13,7 +15,6 @@ import (
 	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/schemadsl/input"
 	"github.com/authzed/spicedb/pkg/tuple"
-	"github.com/stretchr/testify/require"
 )
 
 // BenchmarkCheckDeepArrow benchmarks permission checking through a deep recursive chain.
@@ -111,6 +112,6 @@ func BenchmarkCheckDeepArrow(b *testing.B) {
 
 		// Verify we found the expected result
 		require.Len(b, paths, 1)
-		require.Equal(b, paths[0].Subject.ObjectID, "slow")
+		require.Equal(b, "slow", paths[0].Subject.ObjectID)
 	}
 }
--- a/pkg/query/recursive_strategies_test.go
+++ b/pkg/query/recursive_strategies_test.go
@@ -54,8 +54,6 @@ func TestRecursiveCheckStrategies(t *testing.T) {
 		{"Deepening", recursiveCheckDeepening},
 	}
 
-	var allResults [][]Path
-
 	for _, tc := range strategies {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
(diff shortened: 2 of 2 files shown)


pkg/query/context.go
```diff
@@ -3,21 +3,25 @@ package query
...
 	"github.com/authzed/spicedb/pkg/spiceerrors"
+	"github.com/authzed/spicedb/pkg/tuple"
 )
 
 // TraceLogger is used for debugging iterator execution
 type TraceLogger struct {
 	traces []string
 	depth  int
 	stack  []Iterator // Stack of iterator pointers for proper indentation context
+	writer io.Writer  // Optional writer to output traces in real-time
```
tstirrat15, line 24: oooo

pkg/query/context.go
```diff
@@ -258,6 +281,17 @@ type Context struct {
 	TraceLogger       *TraceLogger      // For debugging iterator execution
 	Analyze           *AnalyzeCollector // Thread-safe collector for query analysis stats
 	MaxRecursionDepth int               // Maximum depth for recursive iterators (0 = use default of 10)
+
+	// Pagination options for IterSubjects and IterResources
+	PaginationCursors map[string]*tuple.Relationship // Cursors for pagination, keyed by iterator ID
+	PaginationLimit   *uint64                        // Limit for pagination (max number of results to return)
+	PaginationSort    options.SortOrder              // Sort order for pagination
```
tstirrat15, line 288: Is this the beginning of cursoring/pagination support?

pkg/query/datastore.go
```diff
@@ -113,7 +113,17 @@ func (r *RelationIterator) checkNormalImpl(ctx *Context, resources []Object, sub
 		return nil, err
 	}
 
-	return convertRelationSeqToPathSeq(iter.Seq2[tuple.Relationship, error](relIter)), nil
+	// Convert to PathSeq
+	pathSeq := convertRelationSeqToPathSeq(iter.Seq2[tuple.Relationship, error](relIter))
+
+	// Eagerly collect all results to terminate the database query immediately
```
tstirrat15, line 119: This is what prevents it from holding onto the collection in the recursive case?

That makes sense, but it also makes me wonder if there's another place to put the eagerness that still retains some of the nicety of laziness. Or are pgx queries not particularly streamy?

pkg/query/datastore.go
```diff
@@ -168,18 +188,96 @@ func (r *RelationIterator) iterSubjectsNormalImpl(ctx *Context, resource Object)
...
+			// If no results, we're done
+			if len(paths) == 0 {
+				return
+			}
+
+			// Update cursor for next page
+			lastPath := paths[len(paths)-1]
+			if rel, err := lastPath.ToRelationship(); err == nil {
+				cursor = &rel
+				ctx.SetPaginationCursor(iteratorID, cursor)
```
tstirrat15, line 265: Is this pagination currently only within a given request, or will this eventually be the externally-facing cursor mechanism? Do we know how we'll ensure that IDs are stable across requests, or is that a tomorrow problem?

pkg/query/recursive_coverage_test.go
```diff
@@ -97,38 +97,6 @@ func TestReplaceRecursiveSentinel_DeepNesting(t *testing.T) {
 	require.True(isFixed, "Deeply nested sentinel should be replaced")
 }
 
-// TestIsRecursiveSubject_NonMatching tests isRecursiveSubject with non-matching types
-func TestIsRecursiveSubject_NonMatching(t *testing.T) {
```
tstirrat15, line 101: Why did this test go away?

tstirrat15 (approved):
LGTM, see questions

pkg/query/benchmarks/check_deep_arrow_benchmark_test.go
```diff
@@ -0,0 +1,136 @@
...
+func BenchmarkCheckDeepArrow(b *testing.B) {
+	// Create an in-memory datastore
+	rawDS, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	ctx := context.Background()
+
+	// Schema definition matching testharness/schema.zed
```
miparnisari: ?

pkg/query/benchmarks/check_deep_arrow_benchmark_test.go
```diff
@@ -0,0 +1,136 @@
...
+// - Checking if user:slow has viewer permission on document:target
+//
+// The permission viewer = view + parent->viewer creates a recursive traversal through
+// all 30+ levels to find the view relationship at the end of the chain.
+func BenchmarkCheckDeepArrow(b *testing.B) {
+	// Create an in-memory datastore
+	rawDS, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	if err != nil {
+		b.Fatal(err)
+	}
```
miparnisari:
```suggestion
	require.NoError(b, err)
```

pkg/query/context.go
barakmich: This is for paginated datastore fetching in the short term. However, in a sense yes, because this is part of the serialization needed for wherever the datastore iterators are currently at in the iteration. But in time, there will need to be storage for the state of the other iterators.

pkg/query/datastore.go
barakmich: Yeah, pgx queries aren't that streamy. I took a cue from how Joey did it in the various LRs (read a page at a time)

pkg/query/recursive_coverage_test.go
barakmich: The function it was testing went away too!

pkg/query/datastore.go
barakmich: It's currently only within a given request. When we want to get to cursoring, we'll be saving state from around the tree into some serializable form. It'll include these cursors, but for now, this is a future-looking way for datastores to save some intermediate data, namely, the high-water ID that marks the page boundary. Across requests... it probably doesn't make too much sense.

PR #2895, opened 2026-02-11
chore: add partial support to schemadsl

tstirrat15:
## Description
Part of unifying `composableschemadsl` and `schemadsl`. This will enable us to better-support composable schema syntax in tooling, from `zed` to `spicedb-vscode` to the playground.

## Changes
Will annotate.

## Testing
Review; see that tests pass.

### DIFF
--- a/pkg/query/recursive_strategies_test.go
+++ b/pkg/query/recursive_strategies_test.go
@@ -178,7 +178,7 @@ func TestRecursiveCheckStrategiesMultipleResources(t *testing.T) {
 	}
 	subject := ObjectAndRelation{ObjectType: "user", ObjectID: "alice", Relation: "..."}
 
-	allResults := make([][]Path, 0)
+	allResults := make([][]Path, 0, len(strategies))
 
 	for _, strategy := range strategies {
 		recursive := NewRecursiveIterator(union, "folder", "view")
--- a/pkg/schemadsl/compiler/compiler.go
+++ b/pkg/schemadsl/compiler/compiler.go
@@ -7,7 +7,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	caveattypes "github.com/authzed/spicedb/pkg/caveats/types"
-	"github.com/authzed/spicedb/pkg/genutil/slicez"
+	"github.com/authzed/spicedb/pkg/genutil/mapz"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/schemadsl/dslshape"
 	"github.com/authzed/spicedb/pkg/schemadsl/input"
@@ -54,7 +54,7 @@ func (cs CompiledSchema) SourcePositionToRunePosition(source input.Source, posit
 type config struct {
 	skipValidation   bool
 	objectTypePrefix *string
-	allowedFlags     []string
+	allowedFlags     *mapz.Set[string]
 	caveatTypeSet    *caveattypes.TypeSet
 }
 
@@ -77,15 +77,17 @@ func CaveatTypeSet(cts *caveattypes.TypeSet) Option {
 }
 
 const (
-	expirationFlag = "expiration"
-	selfFlag       = "self"
+	expirationFlag   = "expiration"
+	selfFlag         = "self"
+	typeCheckingFlag = "typechecking"
+	partialFlag      = "partial"
 )
+
+var allowedFlags = mapz.NewSet(expirationFlag, selfFlag, typeCheckingFlag, partialFlag)
 
 func DisallowExpirationFlag() Option {
 	return func(cfg *config) {
-		cfg.allowedFlags = slicez.Filter(cfg.allowedFlags, func(s string) bool {
-			return s != expirationFlag
-		})
+		cfg.allowedFlags.Delete(expirationFlag)
 	}
 }
 
@@ -96,7 +98,7 @@ type ObjectPrefixOption func(*config)
 // Compile compilers the input schema into a set of namespace definition protos.
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
 	cfg := &config{
-		allowedFlags: []string{expirationFlag, selfFlag},
+		allowedFlags: allowedFlags,
 	}
 
 	prefix(cfg) // required option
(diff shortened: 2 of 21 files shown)


pkg/query/recursive_strategies_test.go
```diff
@@ -174,7 +174,7 @@ func TestRecursiveCheckStrategiesMultipleResources(t *testing.T) {
 	}
 	subject := ObjectAndRelation{ObjectType: "user", ObjectID: "alice", Relation: "..."}
 
-	allResults := make([][]Path, 0)
+	allResults := make([][]Path, 0, len(strategies))
```
tstirrat15, line 181: My linter complained about this.

pkg/schemadsl/compiler/compiler.go
```diff
@@ -96,7 +98,7 @@ type ObjectPrefixOption func(*config)
 // Compile compilers the input schema into a set of namespace definition protos.
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
 	cfg := &config{
-		allowedFlags: []string{expirationFlag, selfFlag},
```
tstirrat15, line 99: I think this means that `use typechecking` wouldn't have worked? I'm curious whether anyone has used it in the real world.

pkg/schemadsl/compiler/compiler.go
```diff
@@ -77,15 +77,17 @@ func CaveatTypeSet(cts *caveattypes.TypeSet) Option {
...
 const (
-	expirationFlag = "expiration"
-	selfFlag       = "self"
+	expirationFlag   = "expiration"
+	selfFlag         = "self"
+	typeCheckingFlag = "typechecking"
+	partialFlag      = "partial"
 )
 
+var allowedFlags = mapz.NewSet(expirationFlag, selfFlag, typeCheckingFlag, partialFlag)
```
tstirrat15, line 86: Will need to keep this updated.

pkg/schemadsl/compiler/compiler.go
```diff
@@ -113,14 +115,19 @@ func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*Co
 		return nil, err
 	}
 
-	cts := caveattypes.TypeSetOrDefault(cfg.caveatTypeSet)
+	initialCompiledPartials := make(map[string][]*core.Relation)
+	caveatTypeSet := caveattypes.TypeSetOrDefault(cfg.caveatTypeSet)
```
tstirrat15, line 119: All of this is ported over from `composableschemadsl`

pkg/schemadsl/compiler/compiler_test.go
```diff
@@ -90,6 +90,294 @@ func TestCompile(t *testing.T) {
 				),
 			},
 		},
+		{
```
tstirrat15, line 93: Tests are ported over directly, with the exception of adding `use partial` at the top of them.

pkg/schemadsl/compiler/translator.go
tstirrat15, line 1: All of these changes reflect the structure of the code in `composableschemadsl`

pkg/schemadsl/dslshape/dslshape.go
```diff
@@ -218,4 +222,19 @@ const (
 	//
 	NodeExpressionPredicateLeftExpr  = "left-expr"
 	NodeExpressionPredicateRightExpr = "right-expr"
+
+	//
+	// NodeTypeImport
+	//
+	NodeImportPredicatePath = "import-path"
```
tstirrat15, line 229: Bringing in import stuff now because why not

pkg/schemadsl/lexer/flaggablelexer_test.go
```diff
@@ -60,6 +60,12 @@ var flaggableLexerTests = []lexerTest{
 		{TokenTypeIdentifier, 0, "self", ""},
 		tEOF,
 	}},
+	{"use partial", "use partial", []Lexeme{
```
tstirrat15, line 63: This is the only thing I'd want feedback on - `use partial` or `use partials`? If it's the latter I'll need to change this test a bit and update a few code locations. It also might not matter because most people will get it via `use v2026.1`

pkg/schemadsl/parser/tests/nonsense_top_level_block.zed
tstirrat15, line 1: All of these tests are copied over from the `composableschemadsl` package.

pkg/schemadsl/parser/tests/partials.zed
```diff
@@ -0,0 +1,12 @@
+use partial
```
tstirrat15, line 1: I added `use partial` to the top of each

pkg/schemadsl/parser/tests/partials.zed.expected
```diff
@@ -0,0 +1,59 @@
+NodeTypeFile
+  end-rune = 147
```
tstirrat15, line 2: Which changed the rune offsets in these files as compared to their `composableschemadsl` counterparts. I went through the failing tests one-by-one and that seemed to be the only meaningful difference.

pkg/schemadsl/compiler/compiler.go
miparnisari: 🤔 it works on the playground

image

although, without the `use`, i don't get any errors.. it is parsing it, but it's ignoring it?

image

pkg/schemadsl/compiler/translator.go
```diff
@@ -52,49 +60,69 @@ func translate(tctx *translationContext, root *dslNode) (*CompiledSchema, error)
...
 			if err != nil {
 				return nil, err
 			}
 
-			definition = def
 			caveatDefinitions = append(caveatDefinitions, def)
 
+			name := def.GetName()
+			if !names.Add(name) {
+				return nil, topLevelNode.WithSourceErrorf(name, "found name reused between multiple definitions and/or caveats: %s", name)
```
miparnisari, line 94: no test for this line?

pkg/schemadsl/compiler/translator.go
```diff
@@ -748,15 +801,96 @@ func addWithCaveats(tctx *translationContext, typeRefNode *dslNode, ref *core.Al
...
 	flagName, err := useFlagNode.GetString(dslshape.NodeUseFlagPredicateName)
 	if err != nil {
 		return err
 	}
-	if slices.Contains(tctx.enabledFlags, flagName) {
+	if tctx.enabledFlags.Has(flagName) {
 		return useFlagNode.Errorf("found duplicate use flag: %s", flagName)
 	}
-	tctx.enabledFlags = append(tctx.enabledFlags, flagName)
+	// TODO: should this be checking for disallowed use flags?
```
miparnisari: why not? pros/cons? (i'm not an expert of this part of the codebase)

pkg/schemadsl/lexer/flaggablelexer_test.go
miparnisari: `use partial` is fine to me

pkg/schemadsl/compiler/compiler.go
tstirrat15: Yeah, without the use the checking is turned off. It could also be that typechecking happened inside of the compilation? Or was otherwise ignored? I was a bit confused by this.

pkg/schemadsl/compiler/translator.go
tstirrat15: It's being exercised: https://github.com/authzed/spicedb/blob/b5350b2c6a348e741c81e6f94facc85476cc3900/pkg/schemadsl/compiler/compiler_test.go#L577-L584

tstirrat15: NFI why coverage isn't catching it

pkg/schemadsl/compiler/translator.go
tstirrat15: I probably will need to address this before adding `use import`, since it will be allowed in most places but not in a `WriteSchema` call

pkg/schemadsl/compiler/translator.go
miparnisari: No, that test is hitting line 113. Add this test:

```go
{
			"duplicate caveat",
			withTenantPrefix,
			`caveat foo(someParam int) {
				someParam == 42
			}
			caveat foo(someParam int) {
				someParam == 42
			}`,
			"parse error in `duplicate caveat`, line 2, column 4: found name reused between multiple definitions and/or caveats: sometenant/foo",
			[]SchemaDefinition{},
		},
``

tstirrat15: Doh. Will do.

PR #2901, opened 2026-02-13
feat: introduce query plan outlines and canonicalization

barakmich:

### DIFF
--- a/pkg/query/outline_fuzz_test.go
+++ b/pkg/query/outline_fuzz_test.go
@@ -293,7 +293,7 @@ func FuzzOutlineCompare(f *testing.F) {
 		case cmp12 > 0:
 			require.Negative(t, cmp21, "Antisymmetry violated: compare(a,b)>0 but compare(b,a)>=0")
 		default:
-			require.Greater(t, cmp21, 0, "Antisymmetry violated: compare(a,b)<0 but compare(b,a)<=0")
+			require.Positive(t, cmp21, "Antisymmetry violated: compare(a,b)<0 but compare(b,a)<=0")
 		}
 
 		// Property: compare(a, a) == 0 (reflexivity)
--- a/pkg/query/path_fuzz_test.go
+++ b/pkg/query/path_fuzz_test.go
@@ -36,11 +36,13 @@ func FuzzPathOrder(f *testing.F) {
 		// Test antisymmetry: if PathOrder(x, y) < 0, then PathOrder(y, x) > 0
 		ab := PathOrder(a, b)
 		ba := PathOrder(b, a)
-		if ab < 0 {
+
+		switch {
+		case ab < 0:
 			require.Positive(t, ba, "PathOrder should be antisymmetric: if PathOrder(a, b) < 0, then PathOrder(b, a) > 0")
-		} else if ab > 0 {
+		case ab > 0:
 			require.Negative(t, ba, "PathOrder should be antisymmetric: if PathOrder(a, b) > 0, then PathOrder(b, a) < 0")
-		} else {
+		default:
 			require.Equal(t, 0, ba, "PathOrder should be antisymmetric: if PathOrder(a, b) == 0, then PathOrder(b, a) == 0")
 		}
 
@@ -48,18 +50,19 @@ func FuzzPathOrder(f *testing.F) {
 		bc := PathOrder(b, c)
 		ac := PathOrder(a, c)
 
-		if ab < 0 && bc < 0 {
+		switch {
+		case ab < 0 && bc < 0:
 			// Case 1: a < b and b < c => a < c
 			require.Negative(t, ac, "PathOrder transitivity: if a < b and b < c, then a < c")
-		} else if ab > 0 && bc > 0 {
+		case ab > 0 && bc > 0:
 			// Case 2: a > b and b > c => a > c
 			require.Positive(t, ac, "PathOrder transitivity: if a > b and b > c, then a > c")
-		} else if ab > 0 {
+		case ab > 0:
 			// Otherwise, b is least element.
 			paths := []Path{a, b, c}
 			slices.SortFunc(paths, PathOrder)
 			require.Equal(t, 0, PathOrder(paths[0], b))
-		} else if ab < 0 {
+		case ab < 0:
 			// Or b is the greatest element
 			paths := []Path{a, b, c}
 			slices.SortFunc(paths, PathOrder)


pkg/query/canonicalize.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// Sort caveats for deterministic ordering
+	slices.SortFunc(caveats, caveatCompare)
+
+	return caveatlessTree, caveats
+}
+
+// extractCaveats recursively extracts all CaveatIteratorType nodes from the tree.
+// Returns the tree with caveats removed and a list of all extracted caveats.
+func extractCaveats(outline Outline) (Outline, []*core.ContextualizedCaveat) {
```
tstirrat15: More for my own edification than something that needs to change: what's the tradeoff between using iterators for this kind of work and using slices? Like I had the thought that `extractCaveats` could be written as an iterator and then you could use `slices.SortedFunc` above, but I don't know whether one approach is better than the other.

pkg/query/canonicalize.go
```diff
@@ -0,0 +1,272 @@
...
+	// If this is a caveat node, extract it
+	if outline.Type == CaveatIteratorType {
+		if outline.Args == nil || outline.Args.Caveat == nil {
+			// Malformed caveat node, skip it
```
tstirrat15: Does this indicate a bug? i.e. should it be a `MustBugf`?

pkg/query/canonicalize.go
```diff
@@ -0,0 +1,272 @@
...
+// Some steps are applied multiple times to ensure convergence (e.g., collapseSingleChild
+// after propagateNull, since propagateNull can create single-child composites).
+func applyBottomUpCanonicalization(outline Outline) Outline {
+	steps := []OutlineMutation{
+		replaceEmptyComposites,
+		propagateNull,
+		flattenComposites,
+		collapseSingleChild,
+		sortCompositeChildren,
+	}
```
tstirrat15: Pull out of the function?

pkg/query/path.go
```diff
@@ -315,6 +315,149 @@ func (p Path) Equals(other Path) bool {
 	return true
 }
 
+// PathOrder defines ordering for Path objects
+// Returns -1 if a < b, 0 if a == b, 1 if a > b
+// PathOrder defines a total ordering on Path objects for canonicalization.
+// Returns -1 if a < b, 0 if a == b, 1 if a > b.
+// Compatible with slices.SortFunc.
+func PathOrder(a, b Path) int {
```
tstirrat15, line 323: Is there a fuzz for this guy? I could see this being a function that we forget to update if we hang something new off of `Path`

pkg/schema/v2/schema.go
```diff
@@ -392,6 +392,83 @@ func (b *BaseRelation) cloneWithParent(parentRelation *Relation) *BaseRelation {
 	}
 }
 
+// Compare compares two BaseRelations, returning -1 if b < other, 0 if b == other, 1 if b > other.
+// Comparison is done lexicographically by: definition name, relation name, type, subrelation, caveat, expiration.
+func (b *BaseRelation) Compare(other *BaseRelation) int {
```
tstirrat15, line 500: Same q here

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+// IsEmpty returns true if the key is empty (not yet computed)
+func (k CanonicalKey) IsEmpty() bool {
+	return string(k) == ""
+}
+
+// Hash returns a hash fingerprint of the key for use in maps
+func (k CanonicalKey) Hash() uint64 {
+	// Use xxhash for fast, non-cryptographic hashing
+	return xxhash.Sum64String(string(k))
+}
```
tstirrat15, line 56: Fast is good, but does using this approach mean that we might have more of a chance of collision? Or is that not the way that fast hashes work?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+
+// Outline is a single type representing the tree of yet-to-be-compiled Iterators.
+type Outline struct {
+	Type         IteratorType
+	Args         *IteratorArgs
+	Subiterators []Outline
```
tstirrat15: Should this be `SubOutlines`?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+		argsStr := serializeArgs(outline.Args)
+		if argsStr != "" {
+			result += "(" + argsStr + ")"
+		}
+	}
+
+	// Add subiterators if present
+	if len(outline.Subiterators) > 0 {
+		result += "["
+		var resultSb522 strings.Builder
```
tstirrat15: naming - I kinda wish this wasn't actually an autofix

pkg/query/canonicalize.go
```diff
@@ -0,0 +1,272 @@
...
+// CanonicalizeOutline transforms an Outline into canonical form.
+// Canonicalization standardizes the representation of logically equivalent queries
+// to enable efficient comparison and deduplication.
+//
+// The canonicalization process has two phases:
+// 1. Filter Lifting: Extract all caveats, sort them, and nest them at the top
+// 2. Bottom-Up Canonicalization: Apply five transformation steps sequentially
+//
+// The function is idempotent: applying it multiple times produces the same result.
+func CanonicalizeOutline(outline Outline) (Outline, error) {
```
miparnisari, line 19: nitpick: why not `func (outline *Outline) CanonicalizeOutline() (Outline, error)`? makes discovering the `func` easier

and why return `Outline` and not a new `CanonicalizedOutline`?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+	// Use xxhash for fast, non-cryptographic hashing
+	return xxhash.Sum64String(string(k))
+}
+
+// Outline is a single type representing the tree of yet-to-be-compiled Iterators.
+type Outline struct {
+	Type         IteratorType
+	Args         *IteratorArgs
+	Subiterators []Outline
+	CanonicalKey CanonicalKey // Populated only after canonicalization
```
miparnisari, line 63: same as my other comment - code smell here? why not make a new type `CanonicalizedOutline`?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+		result += "["
+		var resultSb522 strings.Builder
+		for i, sub := range outline.Subiterators {
+			if i > 0 {
+				resultSb522.WriteString(",")
+			}
+			resultSb522.WriteString(SerializeOutline(sub).String())
+		}
+		result += resultSb522.String()
+		result += "]"
```
tstirrat15: And the static parts should also probably live in the builder

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+	// If names are equal, use EqualVT to check if they're truly equal
+	// If not equal, we need some deterministic ordering for the context
+	if a.EqualVT(b) {
+		return 0
+	}
+
+	// For different contexts with same name, compare context string representation
+	// This ensures a stable (though arbitrary) ordering
+	aCtx := a.Context.String()
+	bCtx := b.Context.String()
```
tstirrat15, line 495: Is this actually guaranteed in golang? I don't know that it would be in JS or Python other than that the main distributions of either happen to use ordered hashmaps

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		rng:       rand.New(rand.NewSource(seed)), //nolint:gosec // weak random is fine for fuzz testing
+		maxDepth:  5,                              // Limit depth to avoid extremely large structures
+		typeNames: []string{"document", "folder", "user", "group", "organization"},
+		relations: []string{"viewer", "editor", "owner", "member", "parent"},
+	}
+}
+
+// Generate creates a random valid Outline
+func (g *outlineGenerator) Generate() Outline {
+	return g.generateWithDepth(0)
```
tstirrat15: What does depth 0 mean semantically?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+		return Outline{
+			Type: SelfIteratorType,
+			Args: &IteratorArgs{
+				RelationName:   typed.relation,
+				DefinitionName: typed.typeName,
+			},
+			Subiterators: decompSubs,
+		}, nil
+
+	default:
```
miparnisari, line 330: i'm starting to think that we should bring this linter https://golangci-lint.run/docs/linters/configuration/#exhaustive to prevent errors of the form "we added a new type of iterator but we forgot to update a few `switch`es"

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		FixedIteratorType,
+		ArrowIteratorType,
+		ExclusionIteratorType,
+		CaveatIteratorType,
+		AliasIteratorType,
+		RecursiveIteratorType,
+		RecursiveSentinelIteratorType,
+		IntersectionArrowIteratorType,
+		SelfIteratorType,
+	}
```
tstirrat15: Pull out of the function? Also would this be a useful static definition in the package generally?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+}
+
+// generateLeafIterator generates a leaf iterator (no subiterators)
+func (g *outlineGenerator) generateLeafIterator() Outline {
+	leafTypes := []IteratorType{
+		NullIteratorType,
+		DatastoreIteratorType,
+		FixedIteratorType,
+		RecursiveSentinelIteratorType,
+		SelfIteratorType,
```
tstirrat15: Same

pkg/query/path.go
tstirrat15: Also is this necessary because of the need to canonicalize a FixedIterator?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		outline.Args = &IteratorArgs{
+			DefinitionName: g.randomTypeName(),
+			RelationName:   g.randomRelation(),
+		}
+
+	case SelfIteratorType:
+		outline.Args = &IteratorArgs{
+			DefinitionName: g.randomTypeName(),
+			RelationName:   g.randomRelation(),
+		}
```
tstirrat15, line 172: Is there a package for "generate random structs?" Or are we doing it manually because we want to constrain the shapes of things?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+	return schema.NewTestBaseRelation(
+		g.randomTypeName(),
+		g.randomRelation(),
+		g.randomTypeName(),
+		tuple.Ellipsis,
+	)
+}
+
+func (g *outlineGenerator) randomCaveat() *core.ContextualizedCaveat {
+	caveats := []string{"caveat1", "caveat2", "caveat3", "test_caveat"}
```
tstirrat15: I'd pull this out as well, up to the top of the file

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+}
+
+// FuzzOutlineCompileDecompile tests that compiling and decompiling Outlines
+// preserves their structure
+func FuzzOutlineCompileDecompile(f *testing.F) {
+	// Add some seed corpus
+	f.Add([]byte{0})
+	f.Add([]byte{1, 2, 3, 4, 5, 6, 7, 8})
+	f.Add([]byte{255, 255, 255, 255, 255, 255, 255, 255})
+	f.Add(bytes.Repeat([]byte{42}, 100))
```
tstirrat15: Does this imply that every test run is identical? I don't really understand what this seed corpus is.

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		}
+
+		// Generate a random outline
+		gen := newOutlineGenerator(data)
+		original := gen.Generate()
+
+		// Canonicalize first (required before compilation)
+		canonical, err := CanonicalizeOutline(original)
+		if err != nil {
+			// Some randomly generated outlines might be invalid
```
tstirrat15: What would invalid mean?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		SelfIteratorType,
+	}
+
+	iterType := iteratorTypes[g.rng.Intn(len(iteratorTypes))]
+
+	return g.generateOutlineOfType(iterType, depth)
+}
+
+// generateLeafIterator generates a leaf iterator (no subiterators)
+func (g *outlineGenerator) generateLeafIterator() Outline {
```
miparnisari:
```suggestion
// generateLeafIterator generates a leaf outline (no subiterators)
func (g *outlineGenerator) generateLeafOutline() Outline {
```

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+			// Some randomly generated outlines might be invalid
+			t.Skip()
+		}
+
+		// Try to compile it
+		compiled, err := canonical.Compile()
+		if err != nil {
+			// Some randomly generated outlines might be invalid
+			// (e.g., wrong number of subiterators after generation)
+			// This is acceptable - just skip
```
tstirrat15: Ah, here's an answer, but it's still surprising to me. What would that look like?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+
+		// Decompile it back
+		roundtrip, err := Decompile(compiled)
+		if err != nil {
+			t.Fatalf("Decompile failed: %v", err)
+		}
+
+		// Check that they're equal to canonical form
+		if !canonical.Equals(roundtrip) {
+			t.Errorf("Roundtrip failed:\nCanonical: %+v\nRoundtrip: %+v", canonical, roundtrip)
```
tstirrat15: Can you use `assert` here?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		}
+
+		// Generate a random outline
+		gen := newOutlineGenerator(data)
+		original := gen.Generate()
+
+		// Canonicalize it once
+		canonical1, err := CanonicalizeOutline(original)
+		if err != nil {
+			t.Fatalf("First canonicalization failed: %v", err)
```
miparnisari: disclaimer: i'm not super familiar with fuzzing.

why is that here, an error aborts the test, but in the tests above, you put this

```go
canonical, err := CanonicalizeOutline(original)
if err != nil {
			// Some randomly generated outlines might be invalid
			t.Skip()
		}
```

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+			return nil, err
+		}
+		compiledSubs[i] = compiled
+	}
+
+	// Now construct the iterator based on type and set canonical key
+	switch c.Type {
+	case NullIteratorType:
+		it := NewFixedIterator()
+		it.canonicalKey = c.CanonicalKey
```
miparnisari: why this line? I searched and it doesn't seem that you are using `it.canonicalKey` for anything after this. also if you remove this line, the tests still pass.

pkg/query/path.go
```diff
@@ -315,6 +315,149 @@ func (p Path) Equals(other Path) bool {
 	return true
 }
 
+// PathOrder defines ordering for Path objects
+// Returns -1 if a < b, 0 if a == b, 1 if a > b
```
miparnisari, line 320:
```suggestion
```

pkg/query/path.go
miparnisari, line 323: in this PR or in a follow-up PR can you please add unit tests for `PathOrder`?

pkg/query/path.go
```diff
@@ -315,6 +315,149 @@ func (p Path) Equals(other Path) bool {
...
+			}
+			return 1
+		}
+
+		// For equal-length maps, use maps.Equal for equality check
+		// If not equal, we need some deterministic ordering
+		// Compare string representations as a fallback
+		if !maps.Equal(a.Metadata, b.Metadata) {
+			aStr := fmt.Sprintf("%v", a.Metadata)
+			bStr := fmt.Sprintf("%v", b.Metadata)
```
miparnisari: when you format a map like this, do you have a guarantee that the keys will be printed in the same order every time? i don't think so.

how about we make `Metadata` a proper struct that has a comparison method of its own? then you can put all this logic there.

pkg/schema/v2/schema.go
miparnisari, line 500: same as the other comment; please include unit tests for this

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+	}
+
+	return 0
+}
+
+// SerializeOutline generates a compact, deterministic string representation
+// of an Outline tree, excluding the CanonicalKey field itself.
+// Format: <Type>(<Args>)[<Sub1>,<Sub2>,...]
+// Returns a CanonicalKey wrapping the serialized string.
+func SerializeOutline(outline Outline) CanonicalKey {
```
miparnisari:
```suggestion
func (outline *Outline) Serialize() CanonicalKey {
```

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+		if !a.Subiterators[i].Equals(b.Subiterators[i]) {
+			return OutlineCompare(a.Subiterators[i], b.Subiterators[i])
+		}
+	}
+
+	return 0 // Equal
+}
+
+// argsCompare returns -1 if a < b, 0 if a == b, 1 if a > b
+func argsCompare(a, b *IteratorArgs) int {
```
tstirrat15, line 403: Same question here - will we capture it if we add new args?

pkg/query/outline_fuzz_test.go
```diff
@@ -0,0 +1,588 @@
...
+		if canonical1.Equals(canonical2) {
+			cmp := OutlineCompare(canonical1, canonical2)
+			if cmp != 0 {
+				t.Errorf("Equal canonical forms don't compare as equal: compare=%d", cmp)
+			}
+		}
+
+		// Property: OutlineCompare should be consistent with Equals
+		cmp := OutlineCompare(canonical1, canonical2)
+		equals := canonical1.Equals(canonical2)
```
tstirrat15: Isn't this definitionally true? Or is the concern that the implementations could diverge at some point?

pkg/query/outline_test.go
```diff
@@ -0,0 +1,1583 @@
...
+		}
+
+		canonical, err := CanonicalizeOutline(outline)
+		require.NoError(err)
+
+		it, err := canonical.Compile()
+		require.NoError(err)
+
+		union, ok := it.(*UnionIterator)
+		require.True(ok, "should be a UnionIterator")
```
tstirrat15: https://pkg.go.dev/github.com/stretchr/testify/require#IsType ?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+		}
+		resultSb574.WriteString(part)
+	}
+	result += resultSb574.String()
+	return result
+}
+
+// serializeRelation converts BaseRelation to compact string
+// Format: defName/relName/subjectType/subrel with flags /c (caveat), /e (expiration), /w (wildcard)
+func serializeRelation(rel *schema.BaseRelation) string {
```
tstirrat15, line 597: Worth using a string builder here as well?

pkg/query/outline.go
```diff
@@ -0,0 +1,627 @@
...
+		result += "]"
+	}
+
+	return CanonicalKey(result)
+}
+
+// serializeArgs converts IteratorArgs to compact string representation
+// Format: field1:val1,field2:val2,...
+// Fields in deterministic order: DefinitionName, RelationName, Relation, Caveat, FixedPaths
+func serializeArgs(args *IteratorArgs) string {
```
tstirrat15, line 544: And I guess on the others too

pkg/query/canonicalize.go
tstirrat15: Where `CanonicalizedOutline` is an alias for `Outline`?

pkg/query/outline.go
tstirrat15: Whyso?

pkg/query/canonicalize.go
barakmich: So the cool thing about Outlines is all in the function signature.

Iterators, for doing the Check/IR/IS, make sense to be purely abstracted. They're a set, and we just want to fetch/evaluate some subset of that set.

But when we're manipulating them directly, like this, the signature `func(Iterator) (Iterator, []*core.ContextualizedCaveat)` here requires a lot of type switching, especially when replacing subiterators. `.ReplaceSubiterators()` does some of that, but hopefully that can go away!

So you totally _can_ write extractCaveats this way -- the first version of canonicalization did exactly that -- but it's far more convenient to work with outlines.

(But this is also a good note to clean up the Iterator interface a little in a followup!)

pkg/query/path.go
barakmich: Yep, and yep. I dig the idea of a fuzzer for this, ensuring a total order 🙂 Lemme do that real quick

pkg/schema/v2/schema.go
barakmich: This one I feel more confident that we won't be changing, unless there's something drastic to change in the schema. But! We are testing it at a higher level because otherwise the fuzz for the canonicalization will fail

pkg/query/outline.go
barakmich: It technically would. But that requires a collision in UINT64_MAX; which would be hard to generate on demand when the number of things we're comparing against is in the dozens, maybe a few hundred at most (based on the size and shape of a schema). 

Plus, we can account for it if we want to when storing these. Hash them to a uint64, pull that from the map, and if the actual string doesn't match (held in the value).... add one. You're in the next bucket over.

Or, from _HMS Pinafore_:

"What never?"
"No never!"
"What never??"
"Well, hardly ever!"

pkg/query/outline.go
barakmich: It would be, and in the other languages too. Strings are totally ordered. If the strings aren't equal, they must have an order. We can't know what the string will be; so we can't say with certainty if a or b is first; that depends on the proto representation. But, we know that the strings will compare.

This is the fallback in general though.

pkg/query/outline_fuzz_test.go
barakmich: Improved this.

pkg/query/outline_fuzz_test.go
barakmich: I guess one could talk about generating random structs -- not as easy in Go -- but doing it by hand means that we're testing only valid (if nonsensical, from an evaluation standpoint) outlines

pkg/query/outline_fuzz_test.go
barakmich: An arrow of three suboutlines? It _shouldn't_ generate one but if it does, it's just a skip.

pkg/query/outline.go
barakmich: Yep; fuzzing on this sort order should find args that wouldn't stably sort

pkg/query/outline_fuzz_test.go
barakmich: Seed corpus creates a starting place for the fuzzer. It can add bytes, remove bytes, tweak the bytes, and try to find more interesting things. It can be more targeted if there's, say, strings we're parsing, but we're just using it as more or less an RNG seed.

pkg/query/outline.go
miparnisari: IMO static functions with no receivers are hard  to discover

pkg/query/canonicalize.go
tstirrat15: Oh! When I said "iterator" I meant golang iter seqs or whatever they're called. Like lazy evaluation and yielding instead of operating on slices.

pkg/query/outline.go
tstirrat15: Ohhh this is a `String()` that you defined rather than some automatic stringification function. Doh.

pkg/query/outline_fuzz_test.go
tstirrat15: Right - isn't that a bug in our fuzzing logic? Or am I misinterpreting how fuzzing goes?

pkg/query/outline.go
tstirrat15: Ah, that's fair. I'm coming from a place where I prefer normal functions in most cases because I tend to prefer functional programming to OOP. I think in golang i think the difference between a method and a function that takes one argument aren't that huge, though.

pkg/query/outline.go
barakmich: I'm open to the idea, but only in explicit annotation mode. And then annotate places like this. Or, better, if it supports annotating specific type alias enums which, if they are switched over, should be exhaustive; and this would certainly be one of them (then it will enforce new code easier)

pkg/query/path.go
barakmich: Fuzzing added!

pkg/query/outline.go
barakmich: I might be getting ahead of myself there, but wanted to add it ahead of time

pkg/query/outline.go
barakmich: So I did think about that. The thing is, we need to keep track of the CanonicalKey for a given node in the outline/iterator tree. But at the same time, we want to modify the Outline, restructuring it as we see fit, so that it's not canonical but has a CanonicalKey. 

And so that's exactly how it's written. CanonicalOutline is a good idea, but it's not obvious to me how to do that without some indirection; how to keep a CanonicalKey attached to a node in a tree, without keeping it on the node.

It's true that one could generate UUIDs, and then this field is a UUID, and the CanonicalOutline contains the outline, and a map from UUID to CanonicalKey. (Or anything else we want to track). I don't mind this if you prefer!

pkg/query/path_fuzz_test.go
tstirrat15, line 1: I dig

pkg/query/path.go
tstirrat15: I think the spec of the language doesn't guarantee it, but the implementation of the language does, similar to javascript and python.

pkg/query/canonicalize.go
miparnisari: no, a new struct altogether

Issue #2525, opened 2025-08-07
Using Read Replica on AWS Aurora Postgres causing errors


epbensimpson:
### What platforms are affected?

linux

### What architectures are affected?

amd64

### What SpiceDB version are you using?

v1.44.4

Also ran into this problem back on `1.42.x` and was hoping it had been fixed in the interim.

### Steps to Reproduce

* AWS Aurora Postgres cluster with single writer and single reader
* SpiceDB configured to use the Reader Endpoint of the cluster via `SPICEDB_DATASTORE_CONN_URI`
* Follower delay `SPICEDB_DATASTORE_FOLLOWER_READ_DELAY_DURATION=2000ms`
* Multiple SpiceDB tasks running (2-5)
* Dispatch enabled

I attempted to reproduce the issue on a test environment using Thumper but was unable to see any errors after 1m+ requests 😓 My script was both checking permissions (weight 10) and writing relations (weight 1) to ensure new revisions were being created.

### Expected Result

No errors due to using a read replica when the follower delay is greater than the maximum replication latency

### Actual Result

Example logs for a single request:

```json
{
    "requestID": "d2aie7vv6t0moif2fshg",
    "time": "2025-08-07T22:22:23Z",
    "level": "error",
    "source": "stderr",
    "error": {
        "error": "object definition `auth/platform` not found",
        "namespace": "auth/platform"
    },
    "message": "received unexpected graph error"
}
{
    "requestID": "d2aie7vv6t0moif2fshg",
    "time": "2025-08-07T22:22:23Z",
    "level": "error",
    "source": "stderr",
    "error": {
        "error": "object definition `auth/platform` not found",
        "namespace": "auth/platform"
    },
    "message": "unexpected dispatch graph error"
}
{
    "grpc.method": "DispatchCheck",
    "grpc.method_type": "unary",
    "peer.address": "10.0.2.197:57466",
    "grpc.start_time": "2025-08-07T22:22:23Z",
    "grpc.request.deadline": "2025-08-07T22:23:23Z",
    "source": "stderr",
    "grpc.code": "Unknown",
    "protocol": "grpc",
    "grpc.error": "object definition `auth/platform` not found",
    "grpc.time_ms": 3,
    "requestID": "d2aie7vv6t0moif2fshg",
    "time": "2025-08-07T22:22:23Z",
    "level": "error",
    "traceID": "8550e95f4f8d5a26eeeb9ad8900ac67f",
    "message": "finished call",
[...]

PR #2903, opened 2026-02-17
chore: add postgres docker-compose setup

miparnisari:
This was me trying to understand the various primary/replica configs and this issue https://github.com/authzed/spicedb/issues/2525.


development/postgres/primary-init.sh
tstirrat15, line 1: replication is always a pain. if it appears to work in practice then i'm good with it.

tstirrat15 (approved):
This looks good to me

PR #2906, opened 2026-02-18
style: run new go fix with modernizers

jzelinskie:
## Description

This change modernizes the Go codebase with newer idioms that we might have missed updating ourselves.

## Testing

These should all be equivalent in behavior. The only thing to be careful about is the functionality and the minimum Go version SpiceDB is built with (in our go.mod).

## References

https://go.dev/blog/gofix

### DIFF
--- a/pkg/validationfile/blocks/expectedrelations.go
+++ b/pkg/validationfile/blocks/expectedrelations.go
@@ -209,8 +209,8 @@ func (vs ValidationString) Subject() (*SubjectWithExceptions, *spiceerrors.WithS
 		exceptions = make([]SubjectAndCaveat, 0, len(exceptionsStringsSlice))
 		for _, exceptionString := range exceptionsStringsSlice {
 			isCaveated := false
-			if strings.HasSuffix(exceptionString, "[...]") {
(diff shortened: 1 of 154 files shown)


tstirrat15 (dismissed):
Yep, I like it

tstirrat15:
I imagine that at some point these analyzers will be a part of `golangci-lint`; the blog post also mentions that they'll be a part of `gopls`, though I'm not sure what that looks like.

tstirrat15 (approved):
Yep, I like it

pkg/validationfile/blocks/expectedrelations.go
```diff
@@ -209,8 +209,8 @@ func (vs ValidationString) Subject() (*SubjectWithExceptions, *spiceerrors.WithS
 		exceptions = make([]SubjectAndCaveat, 0, len(exceptionsStringsSlice))
 		for _, exceptionString := range exceptionsStringsSlice {
 			isCaveated := false
-			if strings.HasSuffix(exceptionString, "[...]") {
```
miparnisari, line 212: hm this strikes me as less readable

PR #2907, opened 2026-02-18
chore: add cockroachdb 26.1 to the matrix and modify version parsing

tstirrat15:
## Description
Cockroach 26.1 has been released and we want to make sure that SpiceDB plays nice with it. This moves the matrix so that we test against 26.1.

We found that Cockroach now restricts access to its internal table, so we can't grab the version object directly. This changes the logic so that we rely on string parsing. This is less safe/direct, but it also works with v26.1.

We also found that checkpointing behavior changed in 26.1, so we updated `min_checkpoint_interval`.

We also 

## Changes
* Move matrix so that we test against 26.1
* Add a test for version checking behavior
* Fix version checking behavior for 26.1
* Fix an issue with `min_checkpoint_interval` in 26.1

## Testing
Review. See that tests pass.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -262,7 +262,11 @@ jobs:
       fail-fast: false
       matrix:
         datastore: ["crdb"]
-        crdbversion: ["25.2.0", "25.3.0"]
+        crdbversion:
+          # the version many folks are on
+          - "25.2.0"
+          # the most recent version
+          - "26.1.0"
     steps:
       - uses: "actions/checkout@v6"
         if: |
@@ -297,7 +301,11 @@ jobs:
       fail-fast: false
       matrix:
         datastore: ["crdb"]
-        crdbversion: ["25.2.0", "25.3.0"]
+        crdbversion:
+          # the version many folks are on
+          - "25.2.0"
+          # the most recent version
+          - "26.1.0"
     steps:
       - uses: "actions/checkout@v6"
         if: |
--- a/internal/datastore/crdb/crdb_test.go
+++ b/internal/datastore/crdb/crdb_test.go
@@ -17,6 +17,8 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"testing"
 	"time"
 
@@ -118,7 +120,7 @@ type datastoreTestFunc func(t *testing.T, ds datastore.Datastore)
 
 func createDatastoreTest(b testdatastore.RunningEngineForTest, tf datastoreTestFunc, options ...Option) func(*testing.T) {
 	return func(t *testing.T) {
-		ctx := context.Background()
+		ctx := t.Context()
 		ds := b.NewDatastore(t, func(engine, uri string) datastore.Datastore {
 			ds, err := NewCRDBDatastore(ctx, uri, options...)
 			require.NoError(t, err)
@@ -959,4 +961,40 @@ func TestRegisterPrometheusCollectors(t *testing.T) {
 	require.Equal(t, float64(writeMaxConns), poolWriteMetric.GetGauge().GetValue()) //nolint:testifylint // we expect exact values
 	require.NotNil(t, poolReadMetric)
 	require.Equal(t, float64(readMaxConns), poolReadMetric.GetGauge().GetValue()) //nolint:testifylint // we expect exact values
+}
+
+func TestVersionReading(t *testing.T) {
+	require := require.New(t)
+
+	expectedVersionList := strings.Split(crdbTestVersion(), ".")
+	expectedMajor, err := strconv.Atoi(expectedVersionList[0])
+	require.NoError(err)
+	expectedMinor, err := strconv.Atoi(expectedVersionList[1])
+	require.NoError(err)
+	expectedPatch, err := strconv.Atoi(expectedVersionList[2])
+	require.NoError(err)
+
+	var version crdbVersion
+
+	b := testdatastore.RunCRDBForTesting(t, "", crdbTestVersion())
+	uri := b.NewDatabase(t)
+
+	// Set up a raw connection to the DB
+	initPoolConfig, err := pgxpool.ParseConfig(uri)
+	require.NoError(err)
+	checker, err := pool.NewNodeHealthChecker(uri)
+	require.NoError(err)
+	initPool, err := pool.NewRetryPool(t.Context(), "pool", initPoolConfig, checker, 18, 20)
+	require.NoError(err)
+	t.Cleanup(func() {
+		initPool.Close()
+	})
+
+	// Make the query for the server version
+	err = queryServerVersion(t.Context(), initPool, &version)
+	require.NoError(err)
+
+	require.Equal(expectedMajor, version.Major)
+	require.Equal(expectedMinor, version.Minor)
+	require.Equal(expectedPatch, version.Patch)
 }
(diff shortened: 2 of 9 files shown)


.github/workflows/build-test.yaml
```diff
@@ -262,7 +262,7 @@ jobs:
       fail-fast: false
       matrix:
         datastore: ["crdb"]
-        crdbversion: ["25.2.0", "25.3.0"]
+        crdbversion: ["25.3.0", "26.1.0"]
```
miparnisari: please don't remove 25.2.0

internal/datastore/common/errors.go
```diff
@@ -163,6 +163,7 @@ func RedactAndLogSensitiveConnString(ctx context.Context, baseErr string, err er
 	filtered = strings.ReplaceAll(filtered, pgURL, "(redacted)")
 	filtered = portMatchRegex.ReplaceAllString(filtered, "(redacted)")
 	filtered = parseMatchRegex.ReplaceAllString(filtered, "(redacted)")
+	fmt.Println(baseErr + ": " + filtered)
```
miparnisari: ?

tstirrat15: Need to remove this.

internal/datastore/crdb/crdb_test.go
```diff
@@ -960,3 +970,32 @@ func TestRegisterPrometheusCollectors(t *testing.T) {
 	require.NotNil(t, poolReadMetric)
 	require.Equal(t, float64(readMaxConns), poolReadMetric.GetGauge().GetValue()) //nolint:testifylint // we expect exact values
 }
+
+func VersionReadingTest(t *testing.T, rawDS datastore.Datastore) {
```
miparnisari: `rawDS` is unused?

tstirrat15: Yeah, it's because the function is being passed to a function that expects it

internal/datastore/crdb/crdb_test.go
```diff
@@ -112,6 +113,15 @@ func TestCRDBDatastoreWithoutIntegrity(t *testing.T) {
 		GCWindow(veryLargeGCWindow),
 		WithAcquireTimeout(5*time.Second),
 	))
+
+	t.Run("TestVersionReading", createDatastoreTest(
```
miparnisari: why wrap `VersionReadingTest` at all with `createDatastoreTest`? just write it as a normal test `func TestVersionReading(t *testing.T) { .. }`

internal/datastore/crdb/crdb_test.go
```diff
@@ -960,3 +970,32 @@ func TestRegisterPrometheusCollectors(t *testing.T) {
...
+
+	expectedVersionList := strings.Split(crdbTestVersion(), ".")
+	expectedMajor := expectedVersionList[0]
+	expectedMinor := expectedVersionList[1]
+	expectedPatch := expectedVersionList[2]
+
+	const (
+		readMaxConns = 10
+	)
+	// Set up a raw connection to the DB
```
miparnisari, line 982: to _what_ DB? where was the DB created?

internal/datastore/crdb/crdb_test.go
tstirrat15: Because the `createDatastoreTest` is parametrized over CRDB versions and takes care of setting up the datastore. I want to be sure that we're correctly reading the version for all CRDB versions that we're testing against.

internal/datastore/crdb/crdb_test.go
tstirrat15: The DB instance in the test? Or are you wanting a more descriptive comment?

miparnisari: the DB is created in `createDatastoreTest`, and the url that we hardcoded in line 986 just... works? is that it?

i wonder if we could structure this better. e.g. if the DB created points to port 5433, this test will break.

tstirrat15: I think so - I copied this test from the structure of a preceding test. I haven't dug into how they work internally. I kinda wonder if this is something with a pgx test connection or a detail of how we're using test containers.

tstirrat15:
Maybe refactor to use `SHOW CLUSTER SETTING version;` instead; this should give `major.minor`

tstirrat15:
^ That would require admin, which would probably break OSS users of CRDB, so we won't go that route.

internal/datastore/crdb/version.go
tstirrat15, line 1: These are the salient changes.

internal/datastore/crdb/crdb_test.go
```diff
@@ -960,3 +970,32 @@ func TestRegisterPrometheusCollectors(t *testing.T) {
 	require.NotNil(t, poolReadMetric)
 	require.Equal(t, float64(readMaxConns), poolReadMetric.GetGauge().GetValue()) //nolint:testifylint // we expect exact values
 }
+
+func VersionReadingTest(t *testing.T, _ datastore.Datastore) {
```
tstirrat15: This is a net-new integration test that asserts that the version requested by the env var is correctly parsed by the internal logic.

pkg/datastore/test/revisions.go
```diff
@@ -20,7 +20,6 @@ import (
 
 // RevisionQuantizationTest tests whether or not the requirements for revisions hold
 // for a particular datastore.
-// TODO: rewrite using synctest
```
tstirrat15, line 23: I removed this TODO because I figured out that this is interacting with an external database whose time state can't be captured by a synctest bubble.

pkg/datastore/test/revisions.go
```diff
@@ -198,9 +198,14 @@ func RevisionGCTest(t *testing.T, tester DatastoreTester) {
 	// require.Error(err, "expected previously written schema to exist at out-of-GC window head")
 
 	// check freshly fetched head revision is valid after GC window elapsed
-	head, err = ds.HeadRevision(ctx)
 	require.NoError(err)
 
+	// assert that recent call to head revision is also valid, even after a GC window cycle without writes elapsed
+	require.NoError(ds.CheckRevision(ctx, head), "expected freshly obtained head revision to be valid")
```
tstirrat15, line 205: Moved this up, since we don't want the length of time that the namespace reads to cause this assertion to flap (which is what was happening to me locally)

pkg/datastore/test/revisions.go
```diff
@@ -198,9 +198,14 @@ func RevisionGCTest(t *testing.T, tester DatastoreTester) {
...
+
+	// TODO: these reads are taking a significant amount of time on CRDB, on the order
+	// of 100ms for a row read. We need to ascertain whether this is a test artifact
+	// or a performance regression.
```
tstirrat15, line 209: This is something that we should figure out but isn't related to the changes in this PR.

pkg/datastore/test/watch.go
```diff
@@ -69,8 +69,11 @@ func WatchTest(t *testing.T, tester DatastoreTester) {
 			require.NoError(err)
 
 			opts := datastore.WatchOptions{
-				Content:                 datastore.WatchRelationships,
-				WatchBufferLength:       50,
+				Content:           datastore.WatchRelationships,
+				WatchBufferLength: 50,
+				// This should mean that we can wait until the checkpoint is emitted and then
+				// make our writes.
+				// CheckpointInterval: 1*time.Second,
```
ecordell:
```suggestion
```

ecordell (approved):

PR #2909, opened 2026-02-19
chore: more lint fixes

tstirrat15:
## Description
Things that my local `golangci-lint` was picking up, I think slightly ahead of the version we're using in CI.

## Changes
See annotations.

## Testing
Review.

### DIFF
--- a/internal/datastore/common/migrations.go
+++ b/internal/datastore/common/migrations.go
@@ -29,10 +29,10 @@ func (mv *MigrationValidator) MigrationReadyState(version string) datastore.Read
 		return datastore.ReadyState{IsReady: true}
 	}
 	var msgBuilder strings.Builder
-	msgBuilder.WriteString(fmt.Sprintf("datastore is not migrated: currently at revision %q, but requires %q", version, mv.headMigration))
+	fmt.Fprintf(&msgBuilder, "datastore is not migrated: currently at revision %q, but requires %q", version, mv.headMigration)
 
 	if len(mv.additionalAllowedMigrations) > 0 {
-		msgBuilder.WriteString(fmt.Sprintf(" (additional allowed migrations: %v)", mv.additionalAllowedMigrations))
+		fmt.Fprintf(&msgBuilder, " (additional allowed migrations: %v)", mv.additionalAllowedMigrations)
 	}
 	msgBuilder.WriteString(". Please run \"spicedb datastore migrate\".")
 	return datastore.ReadyState{
--- a/internal/digests/digestmap_test.go
+++ b/internal/digests/digestmap_test.go
@@ -79,39 +79,36 @@ func TestDigestMap_MultipleKeys(t *testing.T) {
 
 func TestDigestMap_ConcurrentAccess(t *testing.T) {
 	dm := NewDigestMap()
-	numGoroutines := 10
-	numOperations := 100
+	numGoroutines := int32(10)
+	numOperations := int32(100)
 
 	var wg sync.WaitGroup
-	wg.Add(numGoroutines * 2) // writers and readers
 
 	// Concurrent writers
 	for i := range numGoroutines {
-		go func(id int) {
-			defer wg.Done()
-			key := "key-" + string(rune('0'+id))
+		wg.Go(func() {
+			key := "key-" + string('0'+i)
 			for j := range numOperations {
 				dm.Add(key, float64(j))
 			}
-		}(i)
+		})
 	}
 
 	// Concurrent readers
 	for i := range numGoroutines {
-		go func(id int) {
-			defer wg.Done()
-			key := "key-" + string(rune('0'+id))
+		wg.Go(func() {
+			key := "key-" + string('0'+i)
 			for range numOperations {
 				dm.CDF(key, 50.0)
 			}
-		}(i)
+		})
 	}
 
 	wg.Wait()
 
 	// Verify all keys have data
 	for i := range numGoroutines {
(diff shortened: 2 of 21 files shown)


internal/datastore/common/migrations.go
```diff
@@ -29,10 +29,10 @@ func (mv *MigrationValidator) MigrationReadyState(version string) datastore.Read
 		return datastore.ReadyState{IsReady: true}
 	}
 	var msgBuilder strings.Builder
-	msgBuilder.WriteString(fmt.Sprintf("datastore is not migrated: currently at revision %q, but requires %q", version, mv.headMigration))
+	fmt.Fprintf(&msgBuilder, "datastore is not migrated: currently at revision %q, but requires %q", version, mv.headMigration)
```
tstirrat15, line 32: autofixed

internal/digests/digestmap_test.go
```diff
@@ -79,39 +79,36 @@ func TestDigestMap_MultipleKeys(t *testing.T) {
...
 
 	var wg sync.WaitGroup
-	wg.Add(numGoroutines * 2) // writers and readers
 
 	// Concurrent writers
 	for i := range numGoroutines {
-		go func(id int) {
-			defer wg.Done()
-			key := "key-" + string(rune('0'+id))
+		wg.Go(func() {
```
tstirrat15, line 89: Not sure why `go fix` didn't pick this up, but this should be equivalent.

internal/digests/digestmap_test.go
```diff
@@ -79,39 +79,36 @@ func TestDigestMap_MultipleKeys(t *testing.T) {
 
 func TestDigestMap_ConcurrentAccess(t *testing.T) {
 	dm := NewDigestMap()
-	numGoroutines := 10
-	numOperations := 100
+	numGoroutines := int32(10)
+	numOperations := int32(100)
```
tstirrat15, line 83: `rune` is an alias for `int32`, so by iterating over an `int32` we don't have to worry about unsafe casts that weren't happening in this case but that the linter complained about.

internal/dispatch/keys/hasher_ristretto.go
```diff
@@ -70,7 +71,9 @@ func memhash(p unsafe.Pointer, h, s uintptr) uintptr
 
 func runMemHash(seed uint64, data []byte) uint64 {
 	ss := (*stringStruct)(unsafe.Pointer(&data))
-	return uint64(memhash(ss.str, uintptr(seed), uintptr(ss.len)))
+	// NOTE: s.len represents a length which should never be negative, so we ignore the error.
+	length, _ := safecast.Convert[uintptr](ss.len)
```
tstirrat15, line 75: Check me on this - is this too much in the hot path?

internal/fdw/tables/consistency_test.go
```diff
@@ -37,15 +37,15 @@ func TestConsistencyFromFields(t *testing.T) {
 			},
 			expectedConsistencyType: "FullyConsistent",
 		},
-		{
+		{ //nolint:gosec  // this is a test file and test logic
```
tstirrat15, line 40: It doesn't like that there's a base64-encoded value that could be a token

internal/telemetry/reporter.go
```diff
@@ -62,7 +62,7 @@ func writeTimeSeries(ctx context.Context, client *http.Client, endpoint string,
 	r.Header.Add("Content-Encoding", "snappy")
 	r.Header.Set("Content-Type", "application/x-protobuf")
 
-	resp, err := client.Do(r)
+	resp, err := client.Do(r) //nolint:gosec  // re: SSRF: the only thing that's user-controllable is the endpoint, and that's through a config.
```
tstirrat15, line 65: Check me on this. gosec was yelling about SSRF here based on taint analysis

pkg/cmd/postgresfdw_test.go
```diff
@@ -166,7 +166,7 @@ func TestPostgresFDWCompleteValidation(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("missing SpiceDB endpoint", func(t *testing.T) {
-		config := &PostgresFDWConfig{
+		config := &PostgresFDWConfig{ //nolint:gosec  // this is a test file and test logic
```
tstirrat15, line 169: Yelling about tokens here

pkg/releases/releases.go
```diff
@@ -48,7 +48,7 @@ func getLatestReleaseWithClient(ctx context.Context, httpClient *http.Client) (*
 
 	req.Header.Set("Accept", "application/vnd.github+json")
 
-	resp, err := httpClient.Do(req)
+	resp, err := httpClient.Do(req) //nolint:gosec  // re: SSRF: nothing is actually user-controllable that i can see.
```
tstirrat15, line 51: Check me here as well.

pkg/zedtoken/zedtoken_test.go
```diff
@@ -75,78 +75,78 @@ var decodeTests = []struct {
 	expectedStatus    TokenStatus
 	expectError       bool
 }{
-	{
+	{ //nolint:gosec  // this is a test file and test logic
```
tstirrat15, line 78: It doesn't like the `token` field 🙄

internal/dispatch/keys/hasher_ristretto.go
miparnisari: yes 
image

PR #2910, opened 2026-02-19
test: fixup synctest usage in crdbpool

tstirrat15:
Follow-on to #2754

## Description
Synctest runs its tests inside special goroutines, which means we need to use `assert` inside of them (at least as I understand it; I might be wrong on this point).

We also want to `Wait()` to ensure that all goroutines have settled.

## Changes
* synctest refactors
## Testing
Review.

### DIFF
--- a/internal/datastore/crdb/pool/pool_test.go
+++ b/internal/datastore/crdb/pool/pool_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/time/rate"
 )
@@ -103,8 +104,10 @@ func TestContextCancelledDuringBlockingAcquire(t *testing.T) {
 			return nil
 		})
 
-		require.Error(t, err)
-		require.ErrorIs(t, err, context.Canceled)
+		synctest.Wait()
+
+		assert.Error(t, err) //nolint:testifylint  // we're inside a goroutine so this is appropriate
+		assert.ErrorIs(t, err, context.Canceled)
 	})
 }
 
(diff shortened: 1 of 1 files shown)


internal/datastore/crdb/pool/pool_test.go
```diff
@@ -142,37 +147,41 @@ func TestAcquireSucceedsButTopLevelContextCancelled(t *testing.T) {
 	testPool := NewTestPool()
 
 	retryPool := createTestRetryPool(testPool)
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(t.Context())
```
tstirrat15, line 150: Not sure why `testifylint` didn't catch this

internal/datastore/crdb/pool/pool_test.go
```diff
@@ -227,31 +238,33 @@ func TestNoAcquireTimeoutUsesOriginalContext(t *testing.T) {
...
-	// Verify the timeout context has the expected deadline
-	deadline, hasDeadline := acquireContext.Deadline()
-	require.True(t, hasDeadline, "acquire context should have a deadline")
-	expectedDeadline := startTime.Add(acquireTimeout)
-	require.WithinDuration(t, expectedDeadline, deadline, 10*time.Millisecond, "deadline should be approximately correct")
+		// Verify the timeout context has the expected deadline
+		deadline, hasDeadline := acquireContext.Deadline()
+		assert.True(t, hasDeadline, "acquire context should have a deadline")
+		expectedDeadline := startTime.Add(acquireTimeout)
+		assert.Equal(t, expectedDeadline, deadline, "deadline should be correct")
```
tstirrat15, line 266: This is the only change that isn't rewrapping or adding `synctest.Wait` - with `synctest` the time values are exact, so we can assert directly on them without relying on bounds.

internal/datastore/crdb/pool/pool_test.go
```diff
@@ -142,37 +147,41 @@ func TestAcquireSucceedsButTopLevelContextCancelled(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(t.Context())
 	cancel() // Cancel immediately
 
 	err := retryPool.withRetries(ctx, 0, func(conn *pgxpool.Conn) error {
 		t.Fatal("function should not be called when context is cancelled")
 		return nil
 	})
 
-	require.Error(t, err)
+	require.Error(t, err) //nolint:testifylint  // we're inside a goroutine so this is appropriate
```
miparnisari: you wrote the same comment on line 109 but with `assert`???

tstirrat15: Yep, that was incorrect.

miparnisari (approved):

PR #2912, opened 2026-02-19
refactor: introduce datalayer abstraction, remove intermediate interfaces

josephschorr:
Replace direct datastore.Reader usage with datalayer.DataLayer and datalayer.RevisionedReader throughout the codebase. This hides Legacy* methods behind clean interfaces (SchemaReader, RevisionedReader, ReadWriteTransaction) and routes all schema/relationship access through the datalayer package.

Key changes:
- Add pkg/datalayer with DataLayer, RevisionedReader, SchemaReader, and ReadWriteTransaction interfaces that wrap the raw datastore
- Replace manual context key boilerplate with ctxkey.NewBoxedWithDefault
- Rename internal/middleware/datastore and pkg/middleware/datastore packages to datalayer to match the abstraction they serve
- Remove DefinitionLookup interface from pkg/schema/resolver.go; use datalayer.SchemaReader directly
- Remove RelationshipQuerier interface from pagination/iterator.go; use datalayer.RevisionedReader directly
- Remove RelationshipReader interface from pkg/query/context.go; use datalayer.RevisionedReader directly
- Delete internal/datastore/schema/schema.go (consolidated into datalayer)
- Update all datastore implementations (postgres, crdb, mysql, spanner, memdb) and proxies to work through the new interfaces


pkg/datalayer/datalayer.go
```diff
@@ -0,0 +1,139 @@
+package datalayer
```
miparnisari, line 1:
```suggestion
//go:generate go run go.uber.org/mock/mockgen -source datalayer.go -destination ./mocks/mock_datalayer.go
package datalayer
```

internal/caveats/run_test.go
```diff
@@ -469,7 +470,9 @@ func TestRunCaveatExpressions(t *testing.T) {
 			headRevision, err := ds.HeadRevision(t.Context())
 			req.NoError(err)
 
-			reader := ds.SnapshotReader(headRevision)
+			dl := datalayer.NewDataLayer(ds)
+			sr, err := dl.SnapshotReader(headRevision).ReadSchema()
+			req.NoError(err)
```
miparnisari, line 475: For a follow up PR:

`RunSingleCaveatExpression` takes a `datastore.SchemaReadable` — an interface with 1 method. These tests don't need a real datastore at all. Spinning up a full in-memory datastore, parsing a schema string, writing it to storage, fetching a head revision, and getting a snapshot reader is a sign of "excessive arrange phase" code smell ([patent pending](https://frinkiac.com/img/S12E10/766766.jpg) 😄 ).

There are downsides to this:
1. it distracts me from what is actually being tested: caveat expression evaluation, not datastore behavior.
2. it points to a developer issue: we don't have a production-grade helper to create a caveat proto
3. it points to a code issue: when i was refactoring this, i noticed this - why does `LookupCaveatDefinitionsByNames` store `datastore.SchemaDefinition` and not `*core.CaveatDefinition` directly? i see no less than 4 places where callers are forced to do `if caveatDef, ok := def.(*core.CaveatDefinition)`.

This could be simplified as follows:

```go
defer ctrl.Finish()

sr := mock_datalayer.NewMockSchemaReader(ctrl)
sr.EXPECT().LookupCaveatDefinitionsByNames(gomock.Any(), gomock.Any()).Return(map[string]datastore.SchemaDefinition{
	"firstCaveat": createTestCaveat(t, "firstCaveat", "first", types.Default.IntType, "first == 42"),
	"secondCaveat":  createTestCaveat(t, "secondCaveat", "second", types.Default.StringType, "second == 'hello'"),
	"thirdCaveat":  createTestCaveat(t, "thirdCaveat", "third", types.Default.BooleanType, "third"),
}, nil).AnyTimes()
```

and you already wrote `createTestCaveat` in a different file 😄 

```go
func createTestCaveat(t *testing.T, name, paramName string, paramType types.VariableType, expr string) *core.CaveatDefinition {
	env, err := pkgcaveats.EnvForVariablesWithDefaultTypeSet(map[string]types.VariableType{
		paramName: paramType,
	})
	require.NoError(t, err)

	c, err := pkgcaveats.CompileCaveatWithName(env, expr, name)
	require.NoError(t, err)

	cBytes, err := c.Serialize()
	require.NoError(t, err)

	return &core.CaveatDefinition{
		Name:                 name,
		SerializedExpression: cBytes,
		ParameterTypes:       env.EncodedParametersTypes(),
	}
}
```

A similar change could be made in `TestValidateRelationshipOperations`, `TestCheckNamespaceAndRelations`

internal/middleware/datalayer/datastore.go
```diff
@@ -0,0 +1,48 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/authzed/spicedb/pkg/datalayer"
+)
+
+// ContextWithHandle adds a placeholder to a context that will later be
+// filled by the datalayer.
+func ContextWithHandle(ctx context.Context) context.Context {
+	return datalayer.ContextWithHandle(ctx)
+}
```
miparnisari: - `adds a placeholder` ?? why do we need a placeholder? `datalayerKey = ctxkey.NewBoxedWithDefault` is already setting a default. why do we need this method at all? every usage i see (e.g. in tests), they call `ContextWithHandle` and then immediately call `SetInContext`. just make one method... 
- why do we need this file? it is delegating everything to `pkg/datalayer`. either remove this file, or put all the logic here and then `pkg` calls into `internal`.  right now this is backwards (internal calling into pkg)

pkg/validationfile/loader.go
```diff
@@ -165,21 +166,19 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
 	}
 
 	// Load the definitions and relationships into the datastore.
-	revision, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
-		// Write the caveat definitions.
-		err := rwt.LegacyWriteCaveats(ctx, caveatDefs)
+	revision, err := dl.ReadWriteTx(ctx, func(ctx context.Context, rwt datalayer.ReadWriteTransaction) error {
+		ctx = datalayermw.ContextWithDataLayer(ctx, dl)
```
miparnisari: if i remove this line, the tests still pass

pkg/validationfile/loader.go
```diff
@@ -49,7 +50,7 @@ type PopulatedValidationFile struct {
 
 // PopulateFromFiles populates the given datastore with the namespaces and tuples found in
 // the validation file(s) specified.
-func PopulateFromFiles(ctx context.Context, ds datastore.Datastore, caveatTypeSet *caveattypes.TypeSet, filePaths []string) (*PopulatedValidationFile, datastore.Revision, error) {
+func PopulateFromFiles(ctx context.Context, dl datalayer.DataLayer, caveatTypeSet *caveattypes.TypeSet, filePaths []string) (*PopulatedValidationFile, datastore.Revision, error) {
```
miparnisari, line 52: for another PR: this feels like an API that should be part of the `dataLayer` implementation, instead of being here which is very hard to discover. 

You could add these methods to `DataLayer`:
- bulk export(To file)
- bulk export(to response)
- bulk import(from files)
- bulk import(from request)

pkg/validationfile/loader.go
```diff
@@ -189,13 +188,23 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
 			if aerr != nil {
 				return aerr
 			}
```
miparnisari, line 187: if i remove from line 185 to 188, the tests still pass

pkg/validationfile/loader.go
```diff
@@ -165,21 +166,19 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
...
-
-		res := schema.ResolverForDatastoreReader(rwt).WithPredefinedElements(schema.PredefinedElements{
+		ts := schema.NewTypeSystem(resolver.WithPredefinedElements(schema.PredefinedElements{
 			Definitions: objectDefs,
 			Caveats:     caveatDefs,
-		})
-		ts := schema.NewTypeSystem(res)
-		// Validate and write the object definitions.
+		}))
+		// Validate the object definitions.
```
miparnisari, line 177: we don't validate caveats?

pkg/validationfile/loader.go
```diff
@@ -165,21 +166,19 @@ func PopulateFromFilesContents(ctx context.Context, ds datastore.Datastore, cave
...
+	revision, err := dl.ReadWriteTx(ctx, func(ctx context.Context, rwt datalayer.ReadWriteTransaction) error {
+		ctx = datalayermw.ContextWithDataLayer(ctx, dl)
+
+		resolver, err := schema.ResolverForSchemaReader(rwt)
 		if err != nil {
 			return err
 		}
-
-		res := schema.ResolverForDatastoreReader(rwt).WithPredefinedElements(schema.PredefinedElements{
+		ts := schema.NewTypeSystem(resolver.WithPredefinedElements(schema.PredefinedElements{
```
miparnisari, line 173:
```suggestion
		resolver := schema.ResolverForPredefinedDefinitions(schema.PredefinedElements{
			Definitions: objectDefs,
			Caveats:     caveatDefs,
		})
		ts := schema.NewTypeSystem(resolver)
```

pkg/query/context.go
```diff
@@ -10,12 +10,17 @@ import (
...
-	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/datalayer"
 	"github.com/authzed/spicedb/pkg/datastore/options"
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
+// getSchemaReaderFromReader extracts a CaveatDefinitionLookup from a RevisionedReader.
+func getSchemaReaderFromReader(reader datalayer.RevisionedReader) (caveats.CaveatDefinitionLookup, error) {
+	return reader.ReadSchema()
```
miparnisari: nitpick: i don't like these functions that do nothing except delegate the call to something else. they add one layer of indirection that confuse me for no reason

internal/middleware/datalayer/counting.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"google.golang.org/grpc"
+
+	"github.com/authzed/spicedb/pkg/datalayer"
+)
+
+// UnaryCountingInterceptor wraps the datalayer in a counting proxy for each unary request.
+// After the request completes, it exports the counts to Prometheus metrics.
+func UnaryCountingInterceptor() grpc.UnaryServerInterceptor {
+	return datalayer.UnaryCountingInterceptor(nil)
```
miparnisari: similar to my other comment; this middleware being in `internal` and delegating to `pkg` is weird

pkg/cmd/datastore/datastore.go
```diff
@@ -445,17 +446,25 @@ func NewDatastore(ctx context.Context, options ...ConfigOption) (datastore.Datas
 
 		log.Ctx(ctx).Info().Strs("files", opts.BootstrapFiles).Msg("initializing datastore from bootstrap files")
 
-		if len(opts.BootstrapFiles) > 0 {
-			_, _, err = validationfile.PopulateFromFiles(ctx, ds, opts.CaveatTypeSet, opts.BootstrapFiles)
-			if err != nil {
-				return nil, fmt.Errorf("failed to load bootstrap files: %w", err)
+		// Combine bootstrap files and direct contents into a single set so that
+		// all definitions are written together (WriteSchema replaces the full schema).
+		bootstrapContents := make(map[string][]byte, len(opts.BootstrapFiles)+len(opts.BootstrapFileContents))
```
miparnisari, line 451: am i right in understanding that `opts.BootstrapFileContents` is not really an `opt` for users; it's to make testing easier?

internal/services/integrationtesting/consistencytestutil/clusteranddata.go
```diff
@@ -48,15 +49,19 @@ func LoadDataAndCreateClusterForTesting(t *testing.T, consistencyTestFilePath st
...
-	dsCtx := datastoremw.ContextWithHandle(t.Context())
-	require.NoError(datastoremw.SetInContext(dsCtx, ds))
-	res := schema.ResolverForDatastoreReader(ds.SnapshotReader(revision))
+	dl := datalayer.NewDataLayer(ds)
+	dsCtx := datalayermw.ContextWithHandle(t.Context())
+	require.NoError(datalayermw.SetInContext(dsCtx, dl))
+	reader := dl.SnapshotReader(revision)
+	sr, err := reader.ReadSchema()
+	require.NoError(err)
+	res := schema.ResolverFor(sr)
```
miparnisari: i think you can remove these lines. you need them to perform validation in 68-71, but `validationfile.PopulateFromFiles` is already doing validation (well, for types anyway, not for caveats.. see my other comment https://github.com/authzed/spicedb/pull/2912#discussion_r2835006327)

pkg/datalayer/datalayer.go
```diff
@@ -0,0 +1,139 @@
...
+	RevisionFromString(serialized string) (datastore.Revision, error)
+	Watch(ctx context.Context, afterRevision datastore.Revision, options datastore.WatchOptions) (<-chan datastore.RevisionChanges, <-chan error)
+	ReadyState(ctx context.Context) (datastore.ReadyState, error)
+	Features(ctx context.Context) (*datastore.Features, error)
+	OfflineFeatures() (*datastore.Features, error)
+	Statistics(ctx context.Context) (datastore.Stats, error)
+	UniqueID(ctx context.Context) (string, error)
+	MetricsID() (string, error)
+	Close() error
+}
```
miparnisari, line 50:
```suggestion
    // BulkExport
    // BulkImport
    // DeleteAllData (i saw the convenience method in datastore/util.go...)
}
```

and any other APIs this DataLayer should eventually have, in your mind?

i'd like to add them as comments here for future PRs

pkg/datalayer/datalayer.go
```diff
@@ -0,0 +1,139 @@
...
+	LookupCaveatDefByName(ctx context.Context, name string) (datastore.RevisionedCaveat, bool, error)
+
+	// ListAllTypeDefinitions lists all type definitions.
+	ListAllTypeDefinitions(ctx context.Context) ([]datastore.RevisionedTypeDefinition, error)
+
+	// ListAllCaveatDefinitions lists all caveat definitions.
+	ListAllCaveatDefinitions(ctx context.Context) ([]datastore.RevisionedCaveat, error)
+
+	// ListAllSchemaDefinitions lists all type and caveat definitions.
+	ListAllSchemaDefinitions(ctx context.Context) (map[string]datastore.SchemaDefinition, error)
```
miparnisari, line 70: there is cognitive dissonance here. as i was reading this list of functions, all the ones above return a `datastore.Revisioned<something>`. This one and the ones below break this rule. Why?

pkg/datalayer/datalayer.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// LookupSchemaDefinitionsByNames looks up type and caveat definitions by name.
+	LookupSchemaDefinitionsByNames(ctx context.Context, names []string) (map[string]datastore.SchemaDefinition, error)
+
+	// LookupTypeDefinitionsByNames looks up type definitions by name.
+	LookupTypeDefinitionsByNames(ctx context.Context, names []string) (map[string]datastore.SchemaDefinition, error)
```
miparnisari: why does `LookupTypeDefinitionsByNames ` return a map of SCHEMA definitions and not a map of `RevisionedNamespace`

same question for `LookupCaveatDefinitionsByNames` - it could returned `RevisionedCaveats`

pkg/datalayer/datalayer.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// LookupCaveatDefinitionsByNames looks up caveat definitions by name.
+	LookupCaveatDefinitionsByNames(ctx context.Context, names []string) (map[string]datastore.SchemaDefinition, error)
+}
+
+// RevisionedReader reads data at a specific revision.
+type RevisionedReader interface {
```
miparnisari, line 83: 1. this is called `Revisioned`Reader yet none of the methods below take a `revision` as parameter. I mean, I get it, in order to get a `RevisionedReader` you must do `SnapshotReader(revision).ReadSchema()`. But: what does having the `SnapshotReader` method buy you? as far as i can tell, it buys you one extra hop for no good reason. you had to write `defaultDataLayer` and `revisionedReader` structs when it could have been all in `defaultDataLayer`, you had to write `countingDataLayer` and `countingRevisionedReader`, etc.. i guess it saves you from passing the `revision` parameter around, but i wouldn't call this necessarily a bad thing if it makes it super obvious what revision you are using.
2.  funny: `Watch` in line 36, takes a revision but is not part of `RevisionedReader` 😄

internal/services/v1/experimental.go
```diff
@@ -833,9 +852,40 @@ func (es *experimentalServer) ExperimentalCountRelationships(ctx context.Context
 	}, nil
 }
 
+// validateRelationshipsFilterWithSR validates a relationship filter using a datalayer.SchemaReader
+// instead of a datastore.Reader. This is used when the caller has a datalayer.ReadWriteTransaction
+// which provides a SchemaReader via ReadSchema() rather than implementing datastore.Reader directly.
+func validateRelationshipsFilterWithSR(ctx context.Context, filter *v1.RelationshipFilter, sr datalayer.SchemaReader) error {
```
miparnisari: this function looks identical to `func validateRelationshipsFilter(` in `relationships.go`

internal/services/v1/expreflection_test.go
```diff
@@ -550,11 +550,11 @@ func TestExpConvertDiff(t *testing.T) {
 			diff, err := diff.DiffSchemas(es, cs, caveattypes.Default.TypeSet)
 			require.NoError(t, err)
 
-			ds, err := dsfortesting.NewMemDBDatastoreForTesting(t, 100, 1*time.Second, 100*time.Minute)
+			dl, err := dsfortesting.DataLayerForTesting(t, 100, 1*time.Second, 100*time.Minute)
```
miparnisari, line 553: same as https://github.com/authzed/spicedb/pull/2912#discussion_r2833435248. it's incredible that in the test setup you are building a full blown datastore _per test!_ when all you need is something that responds to `.UniqueID(ctx)`.

pkg/datalayer/datalayer.go
josephschorr: As discussed, I highly dislike mocks, so will keep as-is

internal/caveats/run_test.go
josephschorr: Again, no mocks.

If we wanted to simplify here, we could use a fake implementation of the interface, but I don't really have a strong opinion on that vs keeping as-is, so I'd prefer to keep as-is.

internal/middleware/datalayer/datastore.go
josephschorr: Removed and simplified

pkg/validationfile/loader.go
josephschorr: Removed

pkg/validationfile/loader.go
josephschorr: Maybe... I can see it both ways, as I'm unsure we want the file format to be defined in the data layer itself

pkg/validationfile/loader.go
josephschorr: Added a test, although other tests do break if this is removed (any test using the file loader outside of this package)

pkg/validationfile/loader.go
josephschorr: There is no validation on caveats - they don't have dependent types across things

pkg/validationfile/loader.go
josephschorr: That's a semantics change: we want the existing types (if any) in the datastore to be used as well

pkg/query/context.go
josephschorr: It used to do more; it was refactored away

internal/middleware/datalayer/counting.go
josephschorr: Removed

pkg/cmd/datastore/datastore.go
josephschorr: Yes

internal/services/integrationtesting/consistencytestutil/clusteranddata.go
josephschorr: Removed

pkg/datalayer/datalayer.go
josephschorr: Nothing comes to mind

pkg/datalayer/datalayer.go
josephschorr: Because some callers don't need the revision and they usually just need a map

pkg/datalayer/datalayer.go
josephschorr: Because the callers didn't need the revisions

pkg/datalayer/datalayer.go
josephschorr: it is reading a specific revision, hence revisioned reader. I could just name it SnapshotReader like we do in datatstore, but I wanted a different interface name to see it at a glance

internal/services/v1/experimental.go
josephschorr: It was different before a refactoring; must have missed it

internal/services/v1/expreflection_test.go
josephschorr: Not really - again, we could use a fake here, but we have an in-memory DS, might as well use it (and it allows us to test for dependencies within the DS as well)

josephschorr:
Addressed feedback

pkg/datalayer/datalayer.go
miparnisari: The point i was trying to make is that the name `RevisionedReader` doesn't communicate how the revision got there. When someone holds a `RevisionedReader` and calls methods on it, the revision is invisible at the call site. You can't tell, without tracing back to where it was constructed, what revision the data is being read at. The revision is implicit state hidden inside the object.

So what I was proposing is adding a new parameter `revision` to every method in `RevisionedReader`. Then, it is obvious what revision the data is being read at.

This will have to be done in a follow-up PR though.

Can you address (2)?

pkg/datalayer/datalayer.go
miparnisari: that doesn't answer the question. What if you returned `Namespaces`? that is more type-safe than returning `SchemaDefinition`, which forces callers to disambiguate between `Namespaces` and `Caveats`

pkg/datalayer/datalayer.go
miparnisari: is that indicative that these methods do _not_ belong inside of `RevisionedReader` then?

pkg/datalayer/datalayer.go
miparnisari: can you add a TODO here with the list of methods that we need to add in a follow-up PR?

pkg/validationfile/loader.go
miparnisari: it could be `bulkExport(io.Writer)` and `bulkImport(io.Reader)`

pkg/datalayer/datalayer.go
josephschorr: No - ALL datastore access MUST be revisioned; the caller may not need the revision of when the rel or schema *changed*, however (most callers do not)

pkg/datalayer/datalayer.go
josephschorr: Oh, I thought you were asking why it wasn't revisioned. Changed to the specific type

pkg/datalayer/datalayer.go
josephschorr: > The point i was trying to make is that the name RevisionedReader doesn't communicate how the revision got there

Which is fine... the caller just asked for a specific point-in-time. This is no different from the current datastore interface.

> Can you address (2)?

Watch doesn't operate *on* a revision, it starts at one; it should not be on the reader

pkg/datalayer/datalayer.go
josephschorr: Okay

internal/middleware/datalayer/datastore.go
miparnisari: would you be so kind as to update every place that is doing `ContextWithHandle+SetInContext` with `ContextWithDataLayer`? 😄

pkg/datalayer/datalayer.go
miparnisari: I was asking both things 😄 i'm still not happy that you have a `RevisionedReader` interface where some methods don't actually return `RevisionedThings`

pkg/datalayer/datalayer.go
miparnisari: > Which is fine... the caller just asked for a specific point-in-time. 

right, but the function A that creates the `SnapshotReader(rev)` and the function B that actually does `ReadSchema` may not be the same. So if you are reading the code at B and you wanna know what `rev` you are currently using, you need to find the function A.

> Watch doesn't operate on a revision, it starts at one;

how is this different?

pkg/datalayer/datalayer.go
josephschorr: Most things are not going to be revisioned on it - see the `SnapshotReader`, which this more or less duplicates. For example, we don't return the revisions at which the relationships were written - no code needs it and loading it would be extra work for nothing. Ditto for *most* schema operations - the callers don't need the revision at which the schema was written, it simply needs to be grabbing the schema *at* a specific revision.

pkg/datalayer/datalayer.go
josephschorr: > So if you are reading the code at B and you wanna know what rev you are currently using, you need to find the function A.

This has been a long-standing "issue" (I disagree its an issue, as the "code at B" rarely needs to know over which specific revision it is operating). I'd rather not change it right now, unless there is a Very Good Reason

> how is this different?

It isn't working off of a single snapshot/revision, as this reader is. Instead, its explicitly starting at one and then receiving updates from that point forward, which is very much *not* what a "snapshot/revision reader" would do

PR #2916, opened 2026-02-20
test: add unit tests for consistency/forcefull.go

### LINKED ISSUE #2900: add unit tests to `consistency/forcefull.go`
It currently has no tests https://app.codecov.io/gh/authzed/spicedb/blob/main/pkg%2Fmiddleware%2Fconsistency%2Fforcefull.go

ivanauth:
Fixes #2900

Adds 21 test cases covering all exported functions in `pkg/middleware/consistency/forcefull.go`: unary and stream interceptors, bypass whitelist, error propagation, and readonly error rewriting.


pkg/middleware/consistency/forcefull_test.go
```diff
@@ -0,0 +1,508 @@
...
+		require.NoError(err)
+
+		rev, _, err := RevisionFromContext(ctx)
+		require.NoError(err)
+		require.True(head.Equal(rev))
+
+		ds.AssertExpectations(t)
+	})
+
+	t.Run("returns nil for request without consistency interface", func(t *testing.T) {
```
miparnisari, line 99: there is nothing returning `nil` here. you should change the assertion - call `RevisionFromContext` and then assert that you got nothing back.

pkg/middleware/consistency/forcefull_test.go
```diff
@@ -0,0 +1,508 @@
...
+		t.Parallel()
+		ctx := ContextWithHandle(t.Context())
+		err := setFullConsistencyRevisionToContext(ctx, "not-a-consistency-request", nil, "", TreatMismatchingTokensAsFullConsistency)
```
miparnisari: you are passing a request without consistency, but you are also passing `nil` as the datastore. as a reader of this test, i have no proof that the result that you got back is because of the former and not because of the latter. please pass a valid `ds`.

pkg/middleware/consistency/forcefull_test.go
```diff
@@ -0,0 +1,508 @@
...
+		handlerCalled := false
+		handler := func(ctx context.Context, req any) (any, error) {
+			handlerCalled = true
+			return "response", nil
+		}
+
+		resp, err := interceptor(
+			context.Background(),
+			&v1.ReadRelationshipsRequest{},
+			&grpc.UnaryServerInfo{FullMethod: "/grpc.reflection.v1.ServerReflection/ServerReflectionInfo"},
```
miparnisari: @josephschorr could you explain why the ForceFull interceptors use the `bypassServiceWhitelist`? what is the point? none of the methods in those services will satisfy the `hasConsistency` check here https://github.com/authzed/spicedb/blob/8c66fc6880adee174593d5b3882fac7f2eadb39b/pkg/middleware/consistency/forcefull.go#L52 anyway. So it seems like we're just wasting CPU for no good reason.

pkg/middleware/consistency/forcefull_test.go
```diff
@@ -0,0 +1,508 @@
...
+			handler,
+		)
+		require.NoError(err)
+		require.Equal("response", resp)
+		require.True(handlerCalled)
```
miparnisari: these assertions don't show that the `bypassServiceWhitelist` is being used, right?  you need to show that `setFullConsistencyRevisionToContext` didn't produce any side effects

pkg/middleware/consistency/forcefull_test.go
```diff
@@ -0,0 +1,508 @@
...
+		require.Equal("response", resp)
+
+		rev, _, err := RevisionFromContext(capturedCtx)
+		require.NoError(err)
+		require.True(head.Equal(rev))
+
+		ds.AssertExpectations(t)
+	})
+
+	t.Run("returns error when HeadRevision fails", func(t *testing.T) {
```
miparnisari, line 165: this test is almost the same as the previous one (line 76), but you are testing more things here, which suggests that the former is not really needed.

could you re-structure this test suite via using `testpb.InterceptorTestSuite`, like this? https://github.com/authzed/spicedb/blob/757d0dbb1c8b12c407c7c5e96055a6b7b511f077/internal/middleware/usagemetrics/usagemetrics_test.go#L52-L64

ivanauth:
I addressed all the comments. Thank you.

miparnisari:
can you fix the compilation errors?

also, were you not able to use `testpb.InterceptorTestSuite`?

ivanauth:
Fixed the compilation errors.

 Also added `testpb.InterceptorTestSuite` tests.

PR #2919, opened 2026-02-24
feat: Implement use import

tstirrat15:
## Description
Part of the puzzle of folding `composableschemadsl` into `schemadsl`. This implements `import` logic behind a `use` flag.

## Changes
* Port over code from `composableschemadsl`
* Remove test from importer tests that asserted nonlocal `use` behavior (`expiration-usage` or whatever it's called)
* Add test that asserts that `import` statements are disallowed in a schema written to `WriteSchema`
* Fix that test
## Testing
Review. See that tests pass.

### DIFF
--- a/internal/services/v1/schema.go
+++ b/internal/services/v1/schema.go
@@ -121,7 +121,7 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
 	dl := datalayer.MustFromContext(ctx)
 
 	// Compile the schema into the namespace definitions.
-	opts := make([]compiler.Option, 0, 4)
+	opts := make([]compiler.Option, 0, 3)
 	if !ss.expiringRelsEnabled {
 		opts = append(opts, compiler.DisallowExpirationFlag())
 	}
--- a/pkg/composableschemadsl/compiler/importer_test.go
+++ b/pkg/composableschemadsl/compiler/importer_test.go
@@ -168,23 +168,3 @@ func TestEscapeAttemptCausesError(t *testing.T) {
 
 	require.ErrorContains(t, err, "must stay within")
 }
-
-func TestConflictingDefinitionsCausesError(t *testing.T) {
-	t.Parallel()
-
-	workingDir, err := os.Getwd()
-	require.NoError(t, err)
-	test := importerTest{"", "conflicting-definitions"}
-
-	sourceFolder := path.Join(workingDir, test.relativePath())
-
-	inputSchema := test.input()
-
-	_, err = compiler.Compile(compiler.InputSchema{
-		Source:       input.Source("schema"),
-		SchemaString: inputSchema,
-	}, compiler.AllowUnprefixedObjectType(),
-		compiler.SourceFolder(sourceFolder))
-
-	require.ErrorContains(t, err, "AAAAA")
-}
--- a/pkg/schemadsl/compiler/compiler.go
+++ b/pkg/schemadsl/compiler/compiler.go
@@ -102,7 +102,9 @@ const (
 	importFlag       = "import"
 )
 
-var allowedFlags = mapz.NewSet(expirationFlag, selfFlag, typeCheckingFlag, partialFlag, importFlag)
+func allowedFlags() *mapz.Set[string] {
+	return mapz.NewSet(expirationFlag, selfFlag, typeCheckingFlag, partialFlag, importFlag)
+}
 
 func DisallowExpirationFlag() Option {
 	return func(cfg *config) {
@@ -123,7 +125,7 @@ type ObjectPrefixOption func(*config)
 // Compile compilers the input schema into a set of namespace definition protos.
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
 	cfg := &config{
-		allowedFlags: allowedFlags,
+		allowedFlags: allowedFlags(),
 	}
 
 	prefix(cfg) // required option
--- a/pkg/schemadsl/compiler/importer-test/conflicting-definitions/left.zed
+++ b/pkg/schemadsl/compiler/importer-test/conflicting-definitions/left.zed
@@ -1,0 +1,1 @@
+definition bar {}
--- a/pkg/schemadsl/compiler/importer.go
+++ b/pkg/schemadsl/compiler/importer.go
@@ -19,7 +19,7 @@ type CircularImportError struct {
 func importFile(fsys fs.FS, filePath string) (*dslNode, error) {
 	schemaBytes, err := fs.ReadFile(fsys, filePath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read import in schema file: %w", err)
+		return nil, fmt.Errorf("failed to read import in schema file %q: %w", filePath, err)
 	}
 	log.Trace().Str("schema", string(schemaBytes)).Str("file", filePath).Msg("read schema from file")
 
--- a/pkg/schemadsl/compiler/importer_test.go
+++ b/pkg/schemadsl/compiler/importer_test.go
@@ -167,3 +167,23 @@ func TestEscapeAttemptCausesError(t *testing.T) {
 
 	require.ErrorContains(t, err, "must stay within")
 }
+
+func TestConflictingDefinitionsCausesError(t *testing.T) {
+	t.Parallel()
+
+	workingDir, err := os.Getwd()
+	require.NoError(t, err)
+	test := importerTest{"", "conflicting-definitions"}
+
+	sourceFolder := path.Join(workingDir, test.relativePath())
+
+	inputSchema := test.input()
+
+	_, err = compiler.Compile(compiler.InputSchema{
+		Source:       input.Source("schema"),
+		SchemaString: inputSchema,
+	}, compiler.AllowUnprefixedObjectType(),
+		compiler.SourceFolder(sourceFolder))
+
+	require.ErrorContains(t, err, "found name reused between multiple definitions and/or caveats")
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Begin deprecation of library "github.com/dlmiddlecote/sqlstats" (https://github.com/authzed/spicedb/pull/2904).
   NOTE: in a future release, MySQL metrics will change.
+- Add support for imports and partials to the schemadsl package that drives the LSP and development server (https://github.com/authzed/spicedb/pull/2919).
 
 ### Fixed
 - enforce graceful shutdown on serve and serve-testing (https://github.com/authzed/spicedb/pull/2888)
--- a/pkg/schemadsl/compiler/importer-test/conflicting-definitions/root.zed
+++ b/pkg/schemadsl/compiler/importer-test/conflicting-definitions/root.zed
@@ -1,3 +1,5 @@
+use import
+
 import "left.zed"
 import "right.zed"
 


internal/services/shared/errors.go
```diff
@@ -146,7 +146,7 @@ func rewriteError(ctx context.Context, err error, config *ConfigForErrors) error
 	var relationNotFoundError sharederrors.UnknownRelationError
 
 	var compilerError compiler.BaseCompilerError
-	var sourceError *spiceerrors.WithSourceError
```
tstirrat15, line 149: I'm pretty sure this was wrong?

pkg/composableschemadsl/compiler/compiler.go
```diff
@@ -117,7 +117,7 @@ type ObjectPrefixOption func(*config)
 // Compile compilers the input schema into a set of namespace definition protos.
 func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*CompiledSchema, error) {
 	cfg := &config{
-		allowedFlags: mapz.NewSet[string](expirationFlag, selfFlag),
```
tstirrat15, line 120: My LSP marked this as unnecessary

pkg/schemadsl/parser/parser.go
```diff
@@ -262,7 +267,8 @@ func (p *sourceParser) consumeUseFlag(afterDefinition bool) AstNode {
 	}
 
 	if _, ok := lexer.Flags[useFlag]; !ok {
-		p.emitErrorf("Unknown use flag: `%s`. Options are: %s", useFlag, strings.Join(lexer.AllUseFlags, ", "))
+		opts := strings.Join(slices.Sorted(maps.Keys(lexer.Flags)), ", ")
+		p.emitErrorf("Unknown use flag: `%s`. Options are: %s", useFlag, opts)
```
tstirrat15, line 271: Porting this over from `composableschemadsl` - should give us a stable output order.

pkg/schemadsl/compiler/compiler_test.go
```diff
@@ -1327,19 +1327,6 @@ func TestCompile(t *testing.T) {
 				),
 			},
 		},
-		{
-			"duplicate use of expiration pragmas",
```
tstirrat15, line 1331: `use`s should be declared per-file, and should be treated as file-local, so the compiler shouldn't take issue with having multiple of them in the same file.

pkg/schemadsl/compiler/compiler.go
```diff
@@ -107,14 +132,35 @@ func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*Co
...
 
+	present, err := validateImportPresence(cfg.allowedFlags.Has(importFlag), root)
+	if err != nil {
+		// This condition should basically always be satisfied (we trigger errors off of the node),
+		// but we're defensive here in case the implementation changes.
+		var withNodeError withNodeError
+		if errors.As(err, &withNodeError) {
+			return nil, toContextError(withNodeError.Error(), withNodeError.errorSourceCode, withNodeError.node, mapper)
+		}
+	}
```
tstirrat15, line 151: This is net-new - this is how we ensure that import compilation isn't attempted if it's disallowed.

pkg/schemadsl/compiler/compiler.go
```diff
@@ -107,14 +132,35 @@ func Compile(schema InputSchema, prefix ObjectPrefixOption, opts ...Option) (*Co
...
+
+	if present {
```
tstirrat15, line 153: And then because we have information about whether imports are present, we can either run this or not - it's a small optimization but why not

internal/services/v1/schema_test.go
```diff
@@ -61,6 +61,27 @@ func TestSchemaWriteInvalidNamespace(t *testing.T) {
...
+// NOTE: imports must be handled by precompilation;
+// a write of a schema with an import statement is an error.
+func TestSchemaWriteImportsDisallowed(t *testing.T) {
+	conn, cleanup, _, _ := testserver.NewTestServer(require.New(t), 0, memdb.DisableGC, true, tf.EmptyDatastore)
+	t.Cleanup(cleanup)
+	client := v1.NewSchemaServiceClient(conn)
+
+	_, err := client.WriteSchema(t.Context(), &v1.WriteSchemaRequest{
+		Schema: `
+		use import
```
barakmich, line 73: This is correct, in the context of the `WriteSchema` API -- however, I think that part of the overarching problem is that we don't have a `WriteCompiledSchema` API or similar

pkg/schemadsl/compiler/importer-test/README.md
```diff
@@ -0,0 +1,5 @@
+# Test Structure
+
+Every folder should have a `root.zed`. This is the target for compilation.
+
+Every folder will have an `expected.zed`, which is the output of the compilation process.
```
barakmich, line 5: The corollary to a `WriteCompiledSchema` API is that every folder would have a generated `expected.proto`. Again, farther in the future...

pkg/schemadsl/compiler/compiler.go
```diff
@@ -56,6 +58,10 @@ type config struct {
 	objectTypePrefix *string
 	allowedFlags     *mapz.Set[string]
 	caveatTypeSet    *caveattypes.TypeSet
+
+	// In an import context, this is the FS containing
+	// the importing schema (as opposed to imported schemas)
+	sourceFS fs.FS
```
barakmich, line 64: nice use of fs.FS

pkg/schemadsl/compiler/importer_test.go
```diff
@@ -0,0 +1,169 @@
...
+	importerTests := []importerTest{
+		{"simple local import", "simple-local"},
+		{"simple local import with transitive hop", "simple-local-with-hop"},
+		{"nested local import", "nested-local"},
+		{"nested local import with transitive hop", "nested-local-with-hop"},
+		{"nested local two layers deep import", "nested-two-layer-local"},
+		{"diamond-shaped imports are fine", "diamond-shaped"},
+		{"multiple use directives are fine", "multiple-use-directives"},
+		{"many imports are correctly resolved", "many-imports"},
+	}
```
barakmich, line 72: If you're feeling saucy, this could be collected by walking the directory (and omitting the name field), but I don't feel strongly to do so either way.

It does match `pkg/schemadsl/parser/parser_test.go` though so there's that too

internal/services/shared/errors.go
tstirrat15: Nvm it was right. I'm pretty confused as to why it is the way it is.

internal/services/v1/schema.go
```diff
@@ -121,12 +121,15 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
 	dl := datalayer.MustFromContext(ctx)
 
 	// Compile the schema into the namespace definitions.
-	opts := make([]compiler.Option, 0, 3)
+	opts := make([]compiler.Option, 0, 4)
```
miparnisari: where is the 4th option that can be added?

internal/services/v1/schema_test.go
```diff
@@ -61,6 +61,27 @@ func TestSchemaWriteInvalidNamespace(t *testing.T) {
...
+
+		import "foo/bar/baz.zed"
```
miparnisari, line 75: which of these two lines triggers the error?

i suggest splitting this test into two

pkg/schemadsl/compiler/importer-test/many-imports/caveat.zed
```diff
@@ -0,0 +1,2 @@
+// caveat
```
miparnisari, line 1: if i import a file that has one caveat, does it work? or is imports only for definitions 🤔

pkg/schemadsl/compiler/importer-test/nested-two-layer-local/root.zed
```diff
@@ -0,0 +1,8 @@
+use import
+
+import "definitions/user/user.zed"
```
miparnisari, line 3: do we have a test somewhere for a `root` with two imports, and both have the same `definition sometype` with different relations? it should error out

pkg/schemadsl/compiler/importer.go
```diff
@@ -0,0 +1,47 @@
...
+
+type CircularImportError struct {
+	error
+	filePath string
+}
+
+func importFile(fsys fs.FS, filePath string) (*dslNode, error) {
+	schemaBytes, err := fs.ReadFile(fsys, filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read import in schema file: %w", err)
```
miparnisari:
```suggestion
		return nil, fmt.Errorf("failed to read import in schema file %q: %w", filePath, err)
```

pkg/schemadsl/compiler/importer_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+			generated, _, err := generator.GenerateSchema(compiled.OrderedDefinitions)
+			require.NoError(t, err)
+
+			if os.Getenv("REGEN") == "true" {
+				test.writeExpected(generated)
+			} else {
+				expected := test.expected()
+				if !assert.Equal(t, expected, generated, test.name) {
+					t.Log(generated)
```
miparnisari, line 97: no `t.Fail`? here and in line 124?

internal/services/v1/schema_test.go
tstirrat15: It's the latter, and you need both. If you don't have `use import` it will fail to parse, which is captured in the compiler tests themselves. If you don't have `import`, there isn't anything in the schema that's technically disallowed.

pkg/schemadsl/compiler/importer-test/many-imports/caveat.zed
tstirrat15: Imports is for anything - caveats, definitions, partials

pkg/schemadsl/compiler/importer-test/nested-two-layer-local/root.zed
tstirrat15: No, but I can add that

pkg/schemadsl/compiler/importer_test.go
tstirrat15: `assert.Equal` calls `t.Fail` under the hood

pkg/schemadsl/compiler/importer-test/many-imports/caveat.zed
miparnisari: oh ok; i don't think i've seen a test where the imports is for caveats

internal/services/v1/schema.go
tstirrat15: Not entirely sure; I put it down to 3 and things worked.

pkg/composableschemadsl/compiler/importer_test.go
```diff
@@ -168,3 +168,23 @@ func TestEscapeAttemptCausesError(t *testing.T) {
...
+
+	inputSchema := test.input()
+
+	_, err = compiler.Compile(compiler.InputSchema{
+		Source:       input.Source("schema"),
+		SchemaString: inputSchema,
+	}, compiler.AllowUnprefixedObjectType(),
+		compiler.SourceFolder(sourceFolder))
+
+	require.ErrorContains(t, err, "AAAAA")
```
miparnisari: 🤣

pkg/schemadsl/compiler/importer-test/conflicting-definitions/left.zed
miparnisari, line 1: this file is empty?

pkg/schemadsl/compiler/compiler.go
tstirrat15: I think I ended up here accidentally :sweat_smile:

pkg/schemadsl/compiler/importer_test.go
tstirrat15: Yeah, and the other part is that I'd then need a convention for negative tests (of which there are two or three). I found that it was easier to be explicit.

PR #2920, opened 2026-02-24
chore: followup for canonical outlines

barakmich:

### DIFF
--- a/pkg/query/canonicalize.go
+++ b/pkg/query/canonicalize.go
@@ -258,6 +258,8 @@ func isNullOutline(outline Outline) bool {
 
 // assignNodeIDs recursively assigns a unique OutlineNodeID to each node bottom-up,
 // recording each node's serialized CanonicalKey in keys.
+// Filter operations (see isFilterOperation) are left with ID == 0 and are not
+// added to the map; they derive their canonical key on demand via Serialize().
 func assignNodeIDs(outline Outline, keys map[OutlineNodeID]CanonicalKey) Outline {
 	// Recurse on children first (bottom-up)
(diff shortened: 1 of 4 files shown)


pkg/query/canonicalize.go
```diff
@@ -2,25 +2,30 @@ package query
...
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 )
 
-// CanonicalizeOutline transforms an Outline into canonical form.
-// Canonicalization standardizes the representation of logically equivalent queries
-// to enable efficient comparison and deduplication.
+// nodeIDCounter is a package-level counter for assigning OutlineNodeIDs.
+// Its zero value means "unset"; the first Add(1) call returns 1.
+var nodeIDCounter atomic.Uint64
```
tstirrat15, line 13: I think this makes sense to me; is there a reason to use this approach over UUIDs? I'd have probably reached for a UUID here but that's maybe out of ignorance.

pkg/query/outline.go
```diff
@@ -55,35 +55,53 @@ func (k CanonicalKey) Hash() uint64 {
 	return xxhash.Sum64String(string(k))
 }
 
+// OutlineNodeID is a numeric identifier assigned to each node in a CanonicalOutline.
+// It is populated by CanonicalizeOutline; plain (non-canonical) Outlines have a
+// zero-valued ID and cannot be compiled.
+type OutlineNodeID int
```
tstirrat15: Should be uint64

tstirrat15 (dismissed):
See comments; otherwise LGTM

PR #2923, opened 2026-02-25
chore(query): replace UUIDs with canonical keys and add observers

barakmich:


pkg/query/analyze.go
```diff
@@ -65,7 +65,7 @@ func formatNode(it Iterator, analyze map[string]AnalyzeStats, sb *strings.Builde
 	}
 
 	// Write iterator name and ID (truncated for readability)
-	fmt.Fprintf(sb, "%s (ID: %s)\n", explain.Info, iterID[:8])
+	fmt.Fprintf(sb, "%s (Hash: %016x)\n", explain.Info, iterID)
```
tstirrat15: is this the format string for "base 16?"

pkg/query/datastore.go
```diff
@@ -216,7 +212,7 @@ func (r *DatastoreIterator) iterSubjectsNormalImpl(ctx *Context, resource Object
 	// Pagination is configured - return a PathSeq that fetches pages as needed
 	return func(yield func(Path, error) bool) {
 		var cursor *tuple.Relationship
-		iteratorID := r.ID() + ":iter_subjects"
+		iteratorID := fmt.Sprintf("%x:iter_subjects", r.Hash())
```
tstirrat15: What's `x` vs `016x`?

pkg/query/optimize_arrow_direction.go
```diff
@@ -30,10 +30,11 @@ func optimizeArrowDirection(source StatisticsSource) TypedOptimizerFunc[*ArrowIt
 		if rightToLeftCost < leftToRightCost && a.direction != rightToLeft {
 			// Create new arrow with inverted direction
 			newArrow := &ArrowIterator{
-				id:        a.id, // Keep same ID
-				left:      a.left,
-				right:     a.right,
-				direction: rightToLeft,
+				canonicalKey:  a.canonicalKey,
```
tstirrat15, line 33: Is the direction of the iterator not a part of the set of args for the purposes of canonicalization?

tstirrat15 (dismissed):
The question about the reversed arrow iterator is the only thing I'm concerned about

pkg/query/optimize_arrow_direction.go
barakmich: Correct. We don't care which way the arrow points to evaluate the set, the set is the same either way (and so canonical)

pkg/query/analyze.go
barakmich: 016x == 0-padded 16-length (which, for hex bytes is correct; 8 bytes of two hex characters) and the x is hexadecimal

pkg/query/datastore.go
barakmich: yeah, let's make this one match

pkg/query/optimize_arrow_direction.go
tstirrat15: Makes sense!

tstirrat15 (approved):

pkg/query/observer_analyze.go
```diff
@@ -0,0 +1,213 @@
...
+	CheckTime            time.Duration
+	IterSubjectsTime     time.Duration
+	IterResourcesTime    time.Duration
+}
+
+// AnalyzeObserver is a thread-safe Observer that collects execution statistics keyed by CanonicalKey.
+type AnalyzeObserver struct {
+	mu         sync.Mutex
+	stats      map[CanonicalKey]AnalyzeStats                    // GUARDED_BY(mu)
+	startTimes map[ObserverOperation]map[CanonicalKey]time.Time // GUARDED_BY(mu)
```
tstirrat15, line 28: Any benefit to using `sync.Map`?

pkg/query/observer_analyze.go
```diff
@@ -0,0 +1,213 @@
...
+	IterResourcesCalls   int
+	CheckResults         int
+	IterSubjectsResults  int
+	IterResourcesResults int
+	CheckTime            time.Duration
+	IterSubjectsTime     time.Duration
+	IterResourcesTime    time.Duration
+}
+
+// AnalyzeObserver is a thread-safe Observer that collects execution statistics keyed by CanonicalKey.
```
tstirrat15, line 24: Curious - why is thread-safe needed here? Or is it getting ahead of parallelizing execution using goroutines?

pkg/query/observer_analyze.go
```diff
@@ -0,0 +1,213 @@
...
+type AnalyzeStats struct {
+	CheckCalls           int
+	IterSubjectsCalls    int
+	IterResourcesCalls   int
+	CheckResults         int
+	IterSubjectsResults  int
+	IterResourcesResults int
+	CheckTime            time.Duration
+	IterSubjectsTime     time.Duration
+	IterResourcesTime    time.Duration
```
tstirrat15, line 21: And/or use sync primitives here and make it more granular

tstirrat15: Less about microoptimization and more about using things where we can't forget a mutex call. manual mutex management makes me uncomfortable.

pkg/query/observer_analyze.go
```diff
@@ -0,0 +1,213 @@
...
+// GetStats returns a copy of all collected stats.
+func (a *AnalyzeObserver) GetStats() map[CanonicalKey]AnalyzeStats {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	result := make(map[CanonicalKey]AnalyzeStats, len(a.stats))
+	maps.Copy(result, a.stats)
+	return result
+}
+
```
tstirrat15, line 110: Do we know that this can only be called when the stats are completely finished collecting, or could a call to this result in partial statistics?

pkg/query/trace.go
```diff
@@ -0,0 +1,166 @@
...
+	"strings"
+)
+
+// TraceLogger is used for debugging iterator execution
+type TraceLogger struct {
+	traces []string
+	depth  int
+	stack  []Iterator // Stack of iterator pointers for proper indentation context
+	writer io.Writer  // Optional writer to output traces in real-time
+}
```
tstirrat15, line 15: Do we need synchronization primitives here as well? Or is it less of a concern because you'd force an execution of something being `EXPLAIN ANALYZE`d to be single threaded anyway?

pkg/query/observer_analyze.go
barakmich: Yep, getting ahead of the game

pkg/query/trace.go
barakmich: there's `EXPLAIN ANALYZE` which should hand you the tree and the timings (threadsafe) and then there's doing a trace which you can't even really do in SQL -- it would be like setting `log_statement = all` in the Postgres config -- except we can do it per request, which can be useful for debugging, and yes, that would almost certainly be single-threaded. (Debugging correctness, as opposed to performance with EXPLAIN ANALYZE)

pkg/query/observer_analyze.go
barakmich: You could get partial statistics, but the idea is you create an observer alongside the plan, stuff it in the context, and run the query. Then, when the query returns, it's done (and that's when you call `GetStats()`)

You could spawn a goroutine to call this periodically and report partial statistics, I suppose. Maybe if the query were spinning indefinitely? But either way, no, we don't know this is complete; it's only complete because you've orchestrated it that way.

pkg/query/observer_analyze.go
tstirrat15: Chatted with barak on this point and i'm satisfied

barakmich: You'd have to do both the sync.Map part (so that entries in the map are created with locks) _and_ use all the sync primitives. Not to mention sync.Map isn't generic 😭. 

Which just feels onerous, compared to locking a mutex that exists within a single file. It's pretty standard Go.

PR #2924, opened 2026-02-25
feat: add unified schema storage with ReadStoredSchema/WriteStoredSchema

josephschorr:
Introduce the foundation for unified schema storage, where the entire compiled schema (definitions + schema text + hash) is stored as a single serialized proto blob, chunked across rows for SQL datastores.

Datastore interface changes:
- Add ReadStoredSchema(ctx, SchemaHash) to Reader interface
- Add WriteStoredSchema(ctx, *StoredSchema) to ReadWriteTransaction
- Add SchemaHash type with sentinel values for cache bypass

DataLayer interface changes:
- SnapshotReader now takes (Revision, SchemaHash) to thread cache keys
- OptimizedRevision/HeadRevision return SchemaHash alongside Revision
- Add SchemaMode configuration for migration path (legacy → dual → new)
- Add storedSchemaReaderAdapter for reading from StoredSchema protos
- Add writeSchemaViaStoredSchema for building and writing StoredSchema

Storage implementation:
- Add SQLByteChunker generic chunked blob storage for SQL datastores
- Add SQLSingleStoreSchemaReaderWriter for read/write of StoredSchema
- Add per-datastore ReadStoredSchema/WriteStoredSchema implementations for postgres, crdb, mysql, spanner, and memdb
- Add schema table migrations for all SQL datastores
- Add populate migrations to backfill from legacy namespace/caveat tables
- Rename MySQL schema table to stored_schema (schema is a reserved word)

Caching:
- Add SchemaHashCache with LRU + singleflight for schema-by-hash lookups

Proxy/middleware updates:
- Add ReadStoredSchema/WriteStoredSchema pass-through to all datastore proxies (observable, counting, readonly, singleflight, indexcheck, strictreplicated, checkingreplicated, relationshipintegrity, hashcache)
- Update consistency middleware for new HeadRevision/OptimizedRevision signatures

Proto changes:
- Add StoredSchema message to core.proto with V1StoredSchema containing schema_text, schema_hash, namespace_definitions, caveat_definitions


internal/datastore/common/hashcache.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// NewSchemaHashCache creates a new hash-based schema cache wrapping the given cache.
+func NewSchemaHashCache(c SchemaCache) *SchemaHashCache {
+	return &SchemaHashCache{
+		cache: c,
+	}
+}
+
+// Get retrieves a schema from the cache by revision and hash.
```
miparnisari: `rev` is unused

internal/datastore/common/hashcache.go
```diff
@@ -0,0 +1,238 @@
...
+	// Nil revision indicates a bug - should use NoRevision for transaction cases
+	if rev == nil {
+		return spiceerrors.MustBugf("nil revision passed to cache.Set() - use datastore.NoRevision for transaction cases")
+	}
+
+	// Update atomic latest if this is newer or if no revision check (txn case)
+	shouldUpdateLatest := rev == datastore.NoRevision
+	if !shouldUpdateLatest {
+		// Check if this revision is newer than the current latest
+		if latest := c.latest.Load(); latest != nil {
```
miparnisari: this isn't doing a compare-and-swap. i wonder if it should

pkg/datalayer/hashcache.go
```diff
@@ -0,0 +1,238 @@
...
+			if schema != nil {
+				return schema, nil
+			}
+
+			return loader(ctx)
+		}
+		return nil, err
+	}
+
+	return result, nil
```
miparnisari, line 208: we should do `result.CloneVT()`, similar to https://github.com/authzed/spicedb/pull/2629

and add a test please

internal/datastore/common/sqlschema.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	for _, caveat := range caveats {
+		definitions = append(definitions, caveat.Definition)
+	}
+
+	sort.Slice(definitions, func(i, j int) bool {
+		return definitions[i].GetName() < definitions[j].GetName()
+	})
+
+	schemaText, _, err := generator.GenerateSchema(definitions)
```
miparnisari: i can't find a single place where the _ return value is actually used. for a follow up PR: remove it

internal/datastore/common/sqlschema.go
```diff
@@ -0,0 +1,170 @@
...
+	UnifiedSchemaName = "unified_schema"
+
+	// LiveDeletedTxnID is the transaction ID value used to indicate a row has not been deleted yet.
+	// This is the maximum value for uint64, used by databases with tombstone-based deletion.
+	LiveDeletedTxnID = uint64(math.MaxInt64)
+)
+
+// ComputeSchemaHashFromDefinitions computes a SHA256 schema hash from the given namespace and caveat
+// definitions. Definitions are sorted by name before generating the schema text for consistent ordering.
+func ComputeSchemaHashFromDefinitions(namespaces []datastore.RevisionedNamespace, caveats []datastore.RevisionedCaveat) (string, error) {
```
miparnisari: this function isn't used except in tests

internal/datastore/crdb/schema_chunker.go
```diff
@@ -0,0 +1,212 @@
...
+	TableName:         "schema",
+	NameColumn:        "name",
+	ChunkIndexColumn:  "chunk_index",
+	ChunkDataColumn:   "chunk_data",
+	MaxChunkSize:      crdbMaxChunkSize,
+	PlaceholderFormat: sq.Dollar,
+	WriteMode:         common.WriteModeDeleteAndInsert,
+}
+
+// crdbChunkedBytesExecutor implements common.ChunkedBytesExecutor for CockroachDB.
```
miparnisari: `var _ common.ChunkedBytesExecutor = (*crdbChunkedBytesExecutor)(nil)`

internal/datastore/crdb/schema_chunker.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	return t.pool.BeginFunc(ctx, func(tx pgx.Tx) error {
+		_, err := tx.Exec(ctx, sql, args...)
+		return err
+	})
+}
+
+// GetSchemaChunker returns a SQLByteChunker for the schema table.
+// This is exported for testing purposes.
```
miparnisari: you're not using it in tests

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
```diff
@@ -0,0 +1,167 @@
...
+
+		// Compute SHA256 hash
+		hashBytes := sha256.Sum256([]byte(canonicalSchemaText))
+		schemaHash := hashBytes[:]
+
+		// Generate user-facing schema text
+		schemaText, _, err := generator.GenerateSchema(allDefs)
+		if err != nil {
+			return fmt.Errorf("failed to generate schema: %w", err)
+		}
```
miparnisari, line 103: why not use `ComputeSchemaHashFromDefinitions` here? also, you are generating the schema text twice

miparnisari (commented):
what happened to the flags?

internal/datastore/common/hashcache.go
josephschorr: I kept going back and forth on it, but I've changed it to CAS

internal/datastore/common/hashcache.go
josephschorr: Made into `_`, as I want to keep it there for the future

pkg/datalayer/hashcache.go
josephschorr: I don't believe so, as schema should not be modified after loading

internal/datastore/common/sqlschema.go
josephschorr: Follow up

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
josephschorr: Fixed; I must have missed it

internal/datastore/common/sqlschema.go
josephschorr: It was... fixed

internal/datastore/crdb/schema_chunker.go
josephschorr: Fixed; it was before

josephschorr:
Updated

pkg/query/observer_analyze_test.go
```diff
@@ -335,3 +335,254 @@ func TestAnalyzeObserver(t *testing.T) {
 		require.Equal(t, stats1, stats2)
 	})
 }
+
+func TestOptimizationImprovements(t *testing.T) {
```
miparnisari: where did this test come from??

pkg/datalayer/datalayer.go
```diff
@@ -27,10 +27,10 @@ type LegacySchemaWriter interface {
 // It abstracts the underlying datastore, hiding Legacy* methods and
 // providing clean access to schema, relationships, and metadata.
 type DataLayer interface {
-	SnapshotReader(datastore.Revision) RevisionedReader
+	SnapshotReader(datastore.Revision, datastore.SchemaHash) RevisionedReader
```
miparnisari: can you add a godoc to this method? let callers know that they need to call `consistency.RevisionFromContext(ctx)` or `OptimizedRevision` or `HeadRevision` to get both parameters

i also don't know how i feel about this change. i feel like you could have put the `schemaHash` parameter inside of the `revision` struct and rename `revision` to ... some thing else? `pointInTime`?

i would like to hear thoughts from @barakmich and @tstirrat15  since this affects many call sites

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
miparnisari: you're still not using it

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
```diff
@@ -0,0 +1,167 @@
...
+		for chunkIndex := 0; chunkIndex*schemaChunkSize < len(schemaData); chunkIndex++ {
+			start := chunkIndex * schemaChunkSize
+			end := start + schemaChunkSize
+			if end > len(schemaData) {
+				end = len(schemaData)
+			}
+			chunk := schemaData[start:end]
+
+			_, err = tx.Exec(ctx, `
+					INSERT INTO schema (name, chunk_index, chunk_data)
```
miparnisari, line 134: i'm not sure i understand why you are populating these 2 tables. if i run `spicedb migrate head` i will now have the latest schema in them, but because i'm in mode "read-legacy-write-legacy", if i remove a namespace, there will be drift between `schema` and `namespace config`. it SEEMS okay because when i move to "read-legacy-write-new" it will be overwritten. so why bother writing in the first place?

internal/datastore/common/hashcache.go
miparnisari: mind leaving a TODO then?

pkg/datalayer/hashcache.go
miparnisari: well no, it **shouldn't**, but i had the same assumption about CheckResponses and i was wrong

internal/dispatch/remote/cluster_test.go
```diff
@@ -208,8 +209,11 @@ func TestDispatchTimeout(t *testing.T) {
 			resp, err := dispatcher.DispatchCheck(t.Context(), &v1.DispatchCheckRequest{
 				ResourceRelation: &corev1.RelationReference{Namespace: "sometype", Relation: "somerel"},
 				ResourceIds:      []string{"foo"},
-				Metadata:         &v1.ResolverMeta{DepthRemaining: 50},
-				Subject:          &corev1.ObjectAndRelation{Namespace: "foo", ObjectId: "bar", Relation: "..."},
+				Metadata: &v1.ResolverMeta{
+					DepthRemaining: 50,
+					SchemaHash:     []byte(datastore.NoSchemaHashForTesting),
```
miparnisari: here and below, if i remove this line, the tests still pass. Why bother writing what seems to be a default value?

internal/graph/check.go
```diff
@@ -326,7 +333,7 @@ func (cc *ConcurrentChecker) checkDirect(ctx context.Context, crc currentRequest
 		}
 	}()
 	log.Ctx(ctx).Trace().Object("direct", crc.parentReq).Send()
-	dl := datalayer.MustFromContext(ctx).SnapshotReader(crc.parentReq.Revision)
+	dl := datalayer.MustFromContext(ctx).SnapshotReader(crc.parentReq.Revision, datastore.SchemaHash(crc.parentReq.Metadata.GetSchemaHash()))
```
miparnisari: why is `schemaHash` inside of `metadata` but `revision` isn't? to me, `metadata` implies something that is not really required to resolve the request. But in this case, both fields are required. They should be at the same level.

This is also interesting:

```go
type ValidatedCheckRequest struct {
	*v1.DispatchCheckRequest
	Revision datastore.Revision
```

we are duplicating data: the first field _also_ contains the revision.

internal/datastore/proxy/strictreplicated.go
```diff
@@ -260,3 +260,13 @@ func (rr *strictReadReplicatedReader) LookupCounters(ctx context.Context) ([]dat
 	}
 	return counters, err
 }
+
+func (rr *strictReadReplicatedReader) ReadStoredSchema(ctx context.Context, schemaHash datastore.SchemaHash) (*core.StoredSchema, error) {
+	sr := rr.replica.SnapshotReader(rr.rev)
+	schema, err := sr.ReadStoredSchema(ctx, schemaHash)
+	if err != nil && errors.As(err, &common.RevisionUnavailableError{}) {
+		log.Trace().Str("revision", rr.rev.String()).Msg("replica does not contain the requested revision, using primary")
```
miparnisari, line 268: here and elsewhere in this file, instead of printing a log here i'd prefer to attach an attribute to the span: "source=primary". can be a follow up PR

pkg/datastore/datastore.go
```diff
@@ -520,6 +520,12 @@ type Reader interface {
 	LegacySchemaReader
 	CounterReader
 
+	// ReadStoredSchema reads the unified stored schema from the datastore.
+	// The schemaHash parameter is used by caching layers to determine if a cached
+	// version can be used; datastore implementations should ignore it and always
```
miparnisari: re: "datastore implementations should ignore it"

this feels like a major design flaw. You are leaking caching details to the datastore, which should remain unaware of that. Caching should be at the `datalayer` level. `ReadStoredSchema` at the datalayer should accept the hash, but `ReadStoredSchema` at the datastore layer should not.

internal/datastore/proxy/hashcache/hashcache.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// NewHashCacheProxy wraps a datastore with a caching layer for ReadStoredSchema.
+// The cache uses SchemaHash as the key: when a ReadStoredSchema call provides a
+// non-sentinel SchemaHash, the cached result is returned if available.
+// All other datastore operations pass through to the delegate unchanged.
+func NewHashCacheProxy(delegate datastore.Datastore, c cache.Cache[common.SchemaCacheKey, *core.StoredSchema]) datastore.Datastore {
```
miparnisari: this method is never called.

also — unit tests?

miparnisari:
⚠️ 

1. what is the future for both of these flags?
- `--ns-cache-enabled`
- `--enable-experimental-watchable-schema-cache`

2. where is the end-to-end test with all the various datastores that asserts that `spicedb migrate` works and that the server can successfully move through the experiments? from "read-legacy-write-legacy" to "read-legacy-write-both" and so on? it should assert that you can read the schema from the previous migration, write a schema and get it back. you could also throw in a huge schema in the loop, and verify that the chunking is working.

pkg/datastore/datastore.go
```diff
@@ -983,3 +992,50 @@ func (nilRevision) String() string {
...
+	// NoSchemaHashInDevelopment is a sentinel value used when operating in
+	// development mode, where caching of schema is not desired.
+	NoSchemaHashInDevelopment SchemaHash = "no-schema-hash-in-development"
+
+	// NoSchemaHashForTesting is a sentinel value used in tests, where schema
+	// caching is not needed.
+	NoSchemaHashForTesting SchemaHash = "no-schema-hash-for-testing"
+
+	// NoSchemaHashForWatch is a sentinel value used when reading schema for
+	// watch operations, where the hash is not yet available.
```
miparnisari: 1. this is confusing. why would the hash not be available in a Watch call?
2. also, the SchemHash should belong at the datalayer level IMO

pkg/datastore/test/storedschema.go
```diff
@@ -0,0 +1,520 @@
...
+	"github.com/authzed/spicedb/pkg/datastore"
+	ns "github.com/authzed/spicedb/pkg/namespace"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
+	"github.com/authzed/spicedb/pkg/schemadsl/generator"
+	"github.com/authzed/spicedb/pkg/schemadsl/input"
+	"github.com/authzed/spicedb/pkg/testutil"
+)
+
+// buildStoredSchema is a test helper that builds a StoredSchema proto from definitions and schema text.
```
miparnisari: i dislike test helpers. prod code or nothing 😄 
in this case, you already have `writeSchemaViaStoredSchema` that is prod code that does the same thing and more. 
just extract the part of that function that doesn't use `rwt` into a separate function, and then use THAT in these tests

pkg/datastore/datastore.go
```diff
@@ -983,3 +992,50 @@ func (nilRevision) String() string {
...
+	// read-write transaction where the schema revision is not yet stable.
+	NoSchemaHashInTransaction SchemaHash = "no-schema-hash-in-transaction"
+
+	// NoSchemaHashInDevelopment is a sentinel value used when operating in
+	// development mode, where caching of schema is not desired.
+	NoSchemaHashInDevelopment SchemaHash = "no-schema-hash-in-development"
+
+	// NoSchemaHashForTesting is a sentinel value used in tests, where schema
+	// caching is not needed.
+	NoSchemaHashForTesting SchemaHash = "no-schema-hash-for-testing"
```
miparnisari: i dislike code made for dev or for testing. merge these two into one sentinel: `SkipSchemaCache` and call it a day. the implementation won't need to know WHY we are skipping the cache.

pkg/datastore/test/storedschema.go
```diff
@@ -0,0 +1,520 @@
...
+
+	allDefs := make([]compiler.SchemaDefinition, 0, len(compiled.ObjectDefinitions)+len(compiled.CaveatDefinitions))
+	for _, caveatDef := range compiled.CaveatDefinitions {
+		allDefs = append(allDefs, caveatDef)
+	}
+	for _, objDef := range compiled.ObjectDefinitions {
+		allDefs = append(allDefs, objDef)
+	}
+
+	storedSchema := buildStoredSchema(t, allDefs, schemaString)
```
miparnisari: i'm noticing a pattern in the codebase where first we put all caveats+defs into one array, and then we split the array back into caveats and defs (in this case inside `buildStoredSchema`). i wish this pattern wasn't so pervasive :(

pkg/datastore/test/storedschema.go
```diff
@@ -0,0 +1,520 @@
...
+// StoredSchemaWriteReadTest tests basic write and read of a stored schema.
+func StoredSchemaWriteReadTest(t *testing.T, tester DatastoreTester) {
+	require := require.New(t)
+
+	ds, err := tester.New(t, 0, veryLargeGCInterval, veryLargeGCWindow, 1)
+	require.NoError(err)
+
+	ctx := t.Context()
+
+	definitions := []compiler.SchemaDefinition{
```
miparnisari: add caveats to this list and then get rid of `StoredSchemaWithCaveatsTest`

pkg/datastore/test/storedschema.go
```diff
@@ -0,0 +1,520 @@
...
+		require.Equal(schemaText, readSchema.GetV1().SchemaText)
+		return nil
+	})
+	require.NoError(err)
+}
+
+// StoredSchemaStableTextTest verifies that a schema with multiple namespaces and caveats
+// produces stable text when definitions are sorted before generation, regardless of the
+// initial ordering of definitions.
+func StoredSchemaStableTextTest(t *testing.T, tester DatastoreTester) {
```
miparnisari, line 341: does this test belong to the datastore suite of tests?

this test is conflating two concerns:

1. Lines 498-506: Write a schema to the datastore and read it back; already covered by other tests like StoredSchemaRoundTripTest.
2. Lines 508-519: Sort definitions in different orders, run generator.GenerateSchema, and assert the output is identical. This doesn't touch the datastore at all.

internal/datastore/common/hashcache.go
josephschorr: Added

pkg/datalayer/hashcache.go
josephschorr: Sure, but I immediately realized that. I'll change it to return a readonly-wrapper to make it clear to the caller to never change it - cloning adds a bunch of memory pressure

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
josephschorr: Moved into the generator

pkg/query/observer_analyze_test.go
josephschorr: Claude keeps adding it for some reason and I keep removing it, but I must have forgotten this time. I have no idea why it keeps doing so... ghost in the machine

pkg/datalayer/datalayer.go
josephschorr: > i also don't know how i feel about this change. i feel like you could have put the schemaHash parameter inside of the revision struct and rename revision to ... some thing else? pointInTime?

This was my original proposal, but I had pushback that it made things read less nice. Personally, I'd prefer to have the middleware return a combined struct and then pass that, but it does make the caller more constrained

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
josephschorr: For people who want to try the new flag immediately

internal/dispatch/remote/cluster_test.go
josephschorr: The test still passes because its not actually reaching the schema load portion. If it did, without that value, it would explicitly fail

internal/graph/check.go
josephschorr: I can move it out of the metadata and next to the revision, but we generally decided against adding more top-level fields to the dispatch. I'm fine with either.

pkg/datastore/datastore.go
josephschorr: There is no other way if you want the hashcache to be a proxy - it needs access to the value, but it cannot add new parameters to the datastore's interface. If you want to have it supported *without* passing to the underlying datastore, then the hashcache needs to be implemented inside of the datalayer itself (which is fine) OR we need to pass it via context, which is super ugly.

internal/datastore/proxy/strictreplicated.go
josephschorr: Sure, although this is unchanged from previous approach

josephschorr:
> ⚠️
> 
> 1. what is the future for both of these flags?
> 
> * `--ns-cache-enabled`
> * `--enable-experimental-watchable-schema-cache`
> 

Changed to be no-ops and hidden. They will become useless flags and, eventually, removed.

> 2. where is the end-to-end test with all the various datastores that asserts that `spicedb migrate` works and that the server can successfully move through the experiments? from "read-legacy-write-legacy" to "read-legacy-write-both" and so on? it should assert that you can read the schema from the previous migration, write a schema and get it back. you could also throw in a huge schema in the loop, and verify that the chunking is working.

There is a revision-based multi-phase test for each phase in the datastore tests package. We don't run a a migration-per-phase test; the expectation is that checking each phase works is sufficient.

pkg/datastore/datastore.go
josephschorr: 1. as mentioned above, watch does not operate on a single point-in-time; it has no need for the schema hash for a specific revision
2. That's fine, but you previously asked me to put it as a proxy, so this would be going "back" on that decision. I discussed this in a comment above; let me know which approach you prefer

internal/datastore/proxy/hashcache/hashcache.go
josephschorr: Added

pkg/datastore/datastore.go
josephschorr: I *HIGHLY* disagree - when you have sentinels like this, it is VERY important to know why certain things are being skipped, and having that intention *in* the code at the call site is important for understanding.

`SkipSchemaCache` is way too simple and could be misinterpreted

pkg/datastore/test/storedschema.go
josephschorr: Its a relic of storing them as distinct items in the protos and in the namespace and caveat tables. Once we have single-store schema fully in place, we can stop doing it everywhere

pkg/datastore/test/storedschema.go
josephschorr: Combined, although I usually prefer distinct caveat tests

pkg/datastore/test/storedschema.go
josephschorr: Yes, it does, because returning the schema text in a stable manner is now going to be a part of the contract.

pkg/query/observer_analyze_test.go
miparnisari: 🤣

pkg/datastore/datastore.go
miparnisari: >  having that intention in the code at the call

there could be one or a million reasons to skip the cache. are you going to create a sentinel for each? this is what i'm trying to avoid. Why not leave a comment at each call site?

josephschorr: There could be, but in reality there are like... 5. If it grows much beyond that, I'll likely switch down to 2: one for testing and one for prod.

pkg/cmd/serve.go
```diff
@@ -188,6 +188,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 		return fmt.Errorf("failed to mark flag as deprecated: %w", err)
 	}
 	experimentalFlags.BoolVar(&config.EnableExperimentalWatchableSchemaCache, "enable-experimental-watchable-schema-cache", false, "enables the experimental schema cache, which uses the Watch API to keep the schema up to date")
+	experimentalFlags.StringVar(&config.ExperimentalSchemaMode, "experimental-schema-mode", "read-legacy-write-legacy", "schema storage mode for migration to unified schema: read-legacy-write-legacy, read-legacy-write-both, read-new-write-both, read-new-write-new")
```
miparnisari, line 200: can you hide the flags that you said were no longer applicable? or amend their explanation saying so?

internal/services/v1/experimental.go
```diff
@@ -348,9 +349,10 @@ func BulkExport(ctx context.Context, dl datalayer.DataLayer, batchSize uint64, r
 		if err != nil {
 			return shared.RewriteErrorWithoutConfig(ctx, err)
 		}
+		schemaHash = datalayer.NoSchemaHashForLegacyCursor
```
miparnisari: suggestion: instead of doing this here, update `decodeCursor` to return this sentinel value

pkg/middleware/consistency/consistency.go
```diff
@@ -274,51 +283,51 @@ func (s *recvWrapper) RecvMsg(m any) error {
...
 			}
 		}
 
 		if databaseRev.GreaterThan(requestedRev) {
-			return databaseRev, false, nil
+			return databaseRev, hash, false, nil
 		}
 
-		return requestedRev, true, nil
+		return requestedRev, datalayer.NoSchemaHashForLegacyCursor, true, nil
```
miparnisari, line 327: similar to https://github.com/authzed/spicedb/pull/2924#discussion_r2898666519, instead of hardcoding this sentinel here,  update `zedtoken.DecodeRevision` to return it. (The downside is this will be a breaking change for users of SpiceDB as a library)

pkg/cmd/server/server.go
```diff
@@ -224,6 +229,12 @@ func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
 		cachingMode = schemacaching.WatchIfSupported
 	}
 
+	storedSchemaCache, err := CompleteCache[datalayer.SchemaCacheKey, *datastore.ReadOnlyStoredSchema](&c.StoredSchemaCacheConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create stored schema cache: %w", err)
+	}
+	log.Ctx(ctx).Info().EmbedObject(storedSchemaCache).Msg("configured stored schema cache")
```
miparnisari, line 223: 1. missing `closeables.AddWithoutError(storedSchemaCache.Close)`
2. if i do `go run ./cmd/spicedb/main.go serve --grpc-preshared-key foobar` i see this in the logs, which seems wrong:

```go
StoredSchemaCacheConfig={"CacheKindForTesting":"","Enabled":false,"MaxCost":"","Metrics":false,"Name":"","NumCounters":0}
```

i assume we're missing the flags that map to these configs, but should we _add_ flags to configure this? in general i dislike having too many knobs to give customers and i'd prefer to set some sane defaults (i.e. always on, max 32mb, etc...)

pkg/datalayer/schemahash.go
```diff
@@ -0,0 +1,48 @@
...
+	// caching is not needed.
+	NoSchemaHashForTesting SchemaHash = "no-schema-hash-for-testing"
+
+	// NoSchemaHashForWatch is a sentinel value used when reading schema for
+	// watch operations, where the hash is not yet available.
+	NoSchemaHashForWatch SchemaHash = "no-schema-hash-for-watch"
+
+	// NoSchemaHashForLegacyCursor is a sentinel value for decoding legacy cursors
+	// that do not contain a schema hash field.
+	NoSchemaHashForLegacyCursor SchemaHash = "no-schema-hash-for-legacy-cursor"
```
miparnisari, line 27: nit - i'm confused by the "legacy cursor". are they really legacy? are we going to update them? or should this say:
> // NoSchemaHashUnknown is a sentinel value used when the schema hash is not available from the source data (e.g. a cursor)

pkg/datastore/test/storedschema.go
miparnisari: i understand that the goal is to ensure that the call to ReadSchema calls return the same order no matter what order it went in. What I'm saying is that _this test doesn't prove that_. In line 414 you are calling WriteSchema with sorted text and asserting it comes back sorted, gotcha, but then you never call WriteSchema again. I think in line 429 you should call WriteSchema too.

pkg/datastore/test/storedschema.go
```diff
@@ -0,0 +1,652 @@
...
+			})
+			require.NoError(t, err)
+			lastRev = rev
+
+			// Verify legacy storage state.
+			legacyNsDefs, err := ds.SnapshotReader(rev).LegacyListAllNamespaces(ctx)
+			require.NoError(t, err)
+			if phase.hasLegacy {
+				require.NotEmpty(t, legacyNsDefs, "phase %d: expected legacy data", i+1)
+			}
```
miparnisari, line 596: missing:

```go
legacyCaveatDefs, err := ds.SnapshotReader(rev).LegacyListAllCaveats(ctx)
require.NoError(t, err)
if phase.hasLegacy {
     require.NotEmpty(t, legacyCaveatDefs, "phase %d: expected legacy caveat data", i+1)
}
```

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
miparnisari: for our own sanity I would much rather forbid people from skipping phases. it's a lot easier to test that you can move from A to B to C to D than to test that the jump A->C or A->D or .... works correctly.

can we add a startup check that prevents skipping phases? we could store what phase you're in in the DB itself.

pkg/datalayer/impl.go
```diff
@@ -9,35 +9,105 @@ import (
...
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
+// storedSchemaCache caches stored schemas by hash.
+type storedSchemaCache interface {
+	GetOrLoad(ctx context.Context, rev datastore.Revision, schemaHash SchemaHash,
+		loader func(ctx context.Context) (*datastore.ReadOnlyStoredSchema, error)) (*datastore.ReadOnlyStoredSchema, error)
+	Set(rev datastore.Revision, schemaHash SchemaHash, schema *datastore.ReadOnlyStoredSchema) error
```
miparnisari: why make `rev` a parameter? you're always setting it to `datastore.NoRevision`.

if this is for a future optimization, please remove it. If we need it, we can add it later.

pkg/datalayer/hashcache.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// get retrieves a schema from the cache by revision and hash.
+// Fast path: If the hash matches the atomic latest entry, return immediately.
+// Slow path: Check the cache.
+// Returns (nil, nil) if the hash is a bypass sentinel (NoSchemaHashInTransaction or NoSchemaHashForTesting).
+// Returns error if hash is empty string (indicates a bug where schema hash wasn't properly provided).
+func (c *schemaHashCache) get(_ datastore.Revision, schemaHash SchemaHash) (*datastore.ReadOnlyStoredSchema, error) {
+	// TODO(jschorr): Further optimize if we can use the revision in the case when the schema hash does not match,
+	// and perhaps have a cache-by-revision. For now we just ignore the revision in Get since the cache is keyed by hash,
```
miparnisari: > perhaps have a cache by revision

why? revisions change constanly, but the schema changes rarely. i can't think of a scenario where changing the key to include the revision would help.

pkg/datalayer/hashcache.go
```diff
@@ -0,0 +1,226 @@
...
+		schema, err := c.get(rev, schemaHash)
+		if err != nil {
+			return nil, err
+		}
+		if schema != nil {
+			return schema, nil
+		}
+
+		// Load from datastore
+		schema, err = loader(sfCtx)
```
miparnisari: shouldn't this use `ctx` instead of `sfCtx`?

pkg/datalayer/hashcache_test.go
```diff
@@ -0,0 +1,302 @@
...
+		VersionOneof: &core.StoredSchema_V1{
+			V1: &core.StoredSchema_V1StoredSchema{
+				SchemaText: text,
+			},
+		},
+	})
+}
+
+func TestSchemaHashCache_BasicGetSet(t *testing.T) {
+	shc := newSchemaHashCache(newTestSchemaCache()).(*schemaHashCache)
```
miparnisari: instead of doing this ugly casting here, have `newSchemaHashCache` return a `*schemaHashCache` directly and write `var _ storedSchemaCache = (*schemaHashCache)(nil)`

pkg/datalayer/hashcache_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	rev2 := revisions.NewForTransactionID(2)
+
+	err := shc.Set(rev2, SchemaHash("hashRev"), makeTestSchema("definition rev2 {}"))
+	require.NoError(t, err)
+
+	latest := shc.latest.Load()
+	require.Equal(t, SchemaHash("hashRev"), latest.hash)
+
+	// NoRevision should always update latest (transaction case)
```
miparnisari: if you're in the middle of a tx you'd be updating the `latest` pointer, but what if the tx rolls back? `latest` would be incorrect, no?

internal/datastore/spanner/schema_chunker.go
```diff
@@ -0,0 +1,167 @@
...
+
+// ExecuteUpdate converts an UPDATE builder to Spanner mutations.
+func (t *spannerChunkedBytesTransaction) ExecuteUpdate(ctx context.Context, builder sq.UpdateBuilder) error {
+	return spiceerrors.MustBugf("ExecuteUpdate not implemented for Spanner chunked bytes")
+}
+
+// convertInsertToMutation converts an INSERT SQL statement to a Spanner mutation.
+// This assumes the chunker generates INSERT statements in the expected format:
+// INSERT INTO schema (name, chunk_index, chunk_data) VALUES (@p1, @p2, @p3)
+// The format is validated in schema_chunker_test.go.
```
miparnisari: schema_chunker_test.go doesn't exist. can we add one? (and for every other datastore?). i would like to ensure that we don't have any off by one errors

miparnisari:
Things that are missing, IMO:
- update the existing flags, any deprecations?
- observability? e.g. add spans around the new chunking handlers, and around the cache layer. Also maybe attributes, e.g. `SchemaReadFromcached=true/false`

pkg/cmd/serve.go
josephschorr: Not yet - as this is just an experiment, the existing flags need to remain open and visible until such time as *this* implementation is no longer experimental

pkg/middleware/consistency/consistency.go
josephschorr: Its not directly related to zedtoken handling though: it only applies to cursors

pkg/cmd/server/server.go
josephschorr: If we want it to use the cache flags, it needs the cache flags

pkg/datalayer/schemahash.go
josephschorr: They are legacy, as new cursors should have the schema hash in them

internal/services/v1/experimental.go
josephschorr: added

pkg/datalayer/schemahash.go
josephschorr: Added the missing schema hash for the new cursors

pkg/datalayer/impl.go
josephschorr: It is for a followup optimization; I've removed it

pkg/datalayer/hashcache.go
josephschorr: > in the case when the schema hash does not match

Because if we have it watching, we can likely still answer from the cache IF the watch is running fast and we have a revision. I'm just unsure if it is worth the overhead for the rare cache miss

pkg/datalayer/hashcache.go
josephschorr: Doesn't really matter but changed

pkg/datalayer/hashcache_test.go
josephschorr: Changed

pkg/datalayer/hashcache_test.go
josephschorr: Sure, and then the cache would not hit and we'd reload it; its not great, but the chances of that happening are small and its likely not worth the overhead to add txn tracking

internal/datastore/spanner/schema_chunker.go
josephschorr: The unified datastore schema tests should catch these issues, but sure

internal/datastore/crdb/migrations/zz_migration.0011_populate_schema_tables.go
josephschorr: I don't want to store the phase itself in the DB; the operator is designed to automatically move through phases

miparnisari: but that assumes that you're using the k8s operator. what if you're not?

josephschorr: You do it by hand or take the system down for a minute, run from stage 1 -> stage 4 (which runs this migration) and bring it back up. This is how we've always done 4-phase migrations, with the explicit note to use the operator

pkg/datastore/test/storedschema.go
josephschorr: ?

```
	writtenRev := writeSchema(ctx, t, ds, sortedDefs, sortedText)

	// Read it back and verify text is preserved.
	readSchema, err := ds.SnapshotReader(writtenRev).ReadStoredSchema(ctx)
	require.NoError(err)
	require.Equal(sortedText, readSchema.Get().GetV1().SchemaText)
```

`writeSchema` writes the text and then we verify it on the line below.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - feat(query planner): add recursive direction strategies, and fix IS BFS (https://github.com/authzed/spicedb/pull/2891)
 - feat(query planner): introduce query plan outlines and canonicalization (https://github.com/authzed/spicedb/pull/2901)
 - Schema v2: introduces support for PostOrder traversal in walk.go (https://github.com/authzed/spicedb/pull/2761) and improve PostOrder walker cycle detection (https://github.com/authzed/spicedb/pull/2902)
+- Experimental: Add unified schema storage with ReadStoredSchema/WriteStoredSchema for improved schema read performance (https://github.com/authzed/spicedb/pull/2924)
```
miparnisari, line 81: please include an explanation of how to use said experimental, both with the spicedb operator and without it. I.e.:
> You do it by hand or take the system down for a minute, run from stage 1 -> stage 4 (which runs this migration) and bring it back up. This is how we've always done 4-phase migrations, with the explicit note to use the operator

pkg/datalayer/hashcache.go
```diff
@@ -0,0 +1,209 @@
...
+		schema, err := c.get(schemaHash)
+		if err != nil {
+			return nil, err
+		}
+		if schema != nil {
+			return schema, nil
+		}
+
+		// Load from datastore
+		schema, err = loader(ctx)
```
miparnisari, line 177: this is the wrong context. it is the original caller's context, which can get cancelled, and when that happens, the load fails for _all_ waiters, even if the SF context would still be alive.

pkg/datalayer/counting.go
```diff
@@ -48,9 +48,9 @@ type countingDataLayer struct {
 	counts   *MethodCounts
 }
 
-func (c *countingDataLayer) SnapshotReader(rev datastore.Revision) RevisionedReader {
+func (c *countingDataLayer) SnapshotReader(rev datastore.Revision, schemaHash SchemaHash) RevisionedReader {
```
miparnisari: what is the purpose of this `countingDataLayer`? it seems to achieve the same goal as the `observable` proxy. please remove it if we can get the same metrics from that one

pkg/datalayer/hashcache.go
josephschorr: Whoops - I missed it when I re-fixed it. Fixed

pkg/datalayer/counting.go
josephschorr: ok

pkg/datalayer/counting_test.go
```diff
@@ -0,0 +1,467 @@
...
+func (s *stubReadWriteTransaction) UnregisterCounter(_ context.Context, _ string) error {
+	return nil
+}
+
+func (s *stubReadWriteTransaction) StoreCounterValue(_ context.Context, _ string, _ int, _ datastore.Revision) error {
+	return nil
+}
+
+// stubDataLayer implements DataLayer for unit tests.
+type stubDataLayer struct {
```
miparnisari: stubDataLayer, mockDataLayer, countingDataLayer, txCountingDatalayer, defaultDataLayer,  readonlyDataLayer.

we don't need so many implementations. it genuinely makes the code harder to navigate (what is a prod implementation? what is a test implementation?)

please use `mockDataLayer` instead of `stubDatalayer` and `txCountingDatalayer`. you can write expectations for how many times a method on the mock was called like so:

```go
defer ctrl.Finish()
dl := mock_datalayer.NewMockDataLayer(ctrl)
dl.EXPECT().Somemethod(...).Times(1)
```

if you merge the PR as-is, i will likely come back and remove all these alternate mocks and replace with _the_ auto-generated mock. Please, save us both the time and do the update in this PR.

CHANGELOG.md
josephschorr: Added

pkg/datalayer/counting_test.go
josephschorr: Fine

pkg/datalayer/schema_adapter.go
```diff
@@ -48,6 +52,14 @@ func (l *legacySchemaReaderAdapter) SchemaText(ctx context.Context) (string, err
 		return "", err
 	}
 
+	// Sort the definitions by name for deterministic output
```
miparnisari, line 55: please update the godoc of `SchemaText` (and of the interface's godoc too) to say that the defs and caveats will be sorted in the response.
i'm not convinced that in the new storage implementation this is the case. i don't see where the sorting is happening in the code path that uses `WriteSchemaViaStoredSchema`

pkg/datalayer/schema_adapter.go
```diff
@@ -308,3 +320,236 @@ func writeSchemaViaLegacy(ctx context.Context, legacyWriter datastore.LegacySche
...
+	}
+	return result, nil
+}
+
+var _ SchemaReader = (*storedSchemaReaderAdapter)(nil)
+
+// WriteSchemaViaStoredSchema builds a StoredSchema proto and writes it via WriteStoredSchema.
+// If cache is nil, a no-op cache is used.
+func WriteSchemaViaStoredSchema(ctx context.Context, rwt datastore.ReadWriteTransaction,
+	definitions []datastore.SchemaDefinition, schemaString string, cache storedSchemaCache,
```
miparnisari, line 484: this function isn't doing any sorting. please amend the godoc to say what are the expectations around order. for example, must the caller sort definitions by name? must `schemaString` be sorted too? lastly, is the caller satisfying these pre-requisites?

pkg/datalayer/hashcache.go
```diff
@@ -0,0 +1,209 @@
...
+		return latest.schema, nil
+	}
+
+	// Slow path: Check cache
+	schema, ok := c.cache.Get(SchemaCacheKey(schemaHash))
+	if !ok {
+		return nil, nil
+	}
+
+	return schema, nil
```
miparnisari, line 96: you are missing a unit test that hits this line. Please add:

```go
func TestSchemaHashCache_SlowPathCacheHit(t *testing.T) {
	shc := newSchemaHashCache(newTestSchemaCache())

	schema1 := makeTestSchema("definition v1 {}")
	schema2 := makeTestSchema("definition v2 {}")

	// Set hash1, then hash2 — latest now points to hash2
	err := shc.Set(SchemaHash("hash1"), schema1)
	require.NoError(t, err)
	err = shc.Set(SchemaHash("hash2"), schema2)
	require.NoError(t, err)

	shc.cache.Wait()

	// Get hash1 — latest is hash2, so fast path misses,
	// but backing cache should have it (slow path hit, line 95)
	retrieved, err := shc.get(SchemaHash("hash1"))
	require.NoError(t, err)
	require.NotNil(t, retrieved)
	require.Equal(t, "definition v1 {}", retrieved.Get().GetV1().SchemaText)
}
```

internal/datastore/mysql/schema_chunker_test.go
```diff
@@ -0,0 +1,151 @@
...
+	require.Equal(t, "chunk_index", BaseSchemaChunkerConfig.ChunkIndexColumn)
+	require.Equal(t, "chunk_data", BaseSchemaChunkerConfig.ChunkDataColumn)
+	require.Equal(t, 64*1024, BaseSchemaChunkerConfig.MaxChunkSize)
+	require.Equal(t, sq.Question, BaseSchemaChunkerConfig.PlaceholderFormat)
+	require.Equal(t, common.WriteModeInsertWithTombstones, BaseSchemaChunkerConfig.WriteMode)
+}
+
+func TestWrite(t *testing.T) {
+	txn := &fakeTransaction{}
+	executor := &fakeExecutor{transaction: txn}
```
miparnisari: I don't understand these unit tests. run them by hand and look at coverage: they cover zero lines of the mysql chunker code. you are testing a `fakeTx` and a `fakeExecutor`. shouldn't be using `mysqlTransactionAwareTransaction` and `mysqlRevisionAwareExecutor`?

`spanner/schema_chunker_test.go` has the same issue

pkg/datastore/test/storedschema.go
miparnisari: i don't understand this test then

internal/datastore/mysql/migrations/zz_migration.0012_populate_schema_tables.go
```diff
@@ -0,0 +1,188 @@
...
+		return allDefs[i].GetName() < allDefs[j].GetName()
+	})
+
+	// Generate canonical schema text
+	canonicalSchemaText, _, err := generator.GenerateSchema(allDefs)
+	if err != nil {
+		return fmt.Errorf("failed to generate canonical schema: %w", err)
+	}
+
+	// Compute SHA256 hash
```
miparnisari: here and in every other similar migration, use `ComputeSchemaHash`. also, this line appears twice `generator.GenerateSchema(allDefs)`

actually, since `ComputeSchemaHash` is generating the schema text as well, you could return it and use it directly.

finally: please amend the godoc of `generator.GenerateSchema` and `generator.GenerateSchemaWithCaveatTypeSet` to say that it does NOT do any sorting and callers are expected to do that

pkg/datalayer/impl.go
```diff
@@ -82,10 +152,17 @@ func (d *defaultDataLayer) Close() error {
...
+	reader     datastore.Reader
+	rev        datastore.Revision
+	schemaMode SchemaMode
+	schemaHash SchemaHash
+	cache      storedSchemaCache
 }
 
 func (r *revisionedReader) ReadSchema(ctx context.Context) (SchemaReader, error) {
+	if r.schemaMode.ReadsFromNew() {
+		return newStoredSchemaReaderAdapter(r.reader, r.schemaHash, r.rev, r.cache)
```
miparnisari: pass the `ctx` to `newStoredSchemaReaderAdapter` and then use it  in here: `ctx, span := tracer.Start(context.Background(), "ReadStoredSchema")`

otherwise the span for `ReadStoredSchema` won't have the right parent:

image

(i also don't know why 1 call to CheckPermission requires 4 schema reads... i am using `read-new-write-new`)

pkg/datalayer/schema_adapter.go
josephschorr: ? They aren't sorted in the response in the future; *today* they are

josephschorr: I think this the misunderstanding: the *legacy* mode sorts them because that is what we do today and we need to continue to do so to match current output and have it be stable.

In the new single-store mode, we return the text we were given, which is also stable, but not sorted.

pkg/datalayer/schema_adapter.go
josephschorr: It shouldn't be, as discussed above

pkg/datalayer/hashcache.go
josephschorr: Added

internal/datastore/mysql/schema_chunker_test.go
josephschorr: Fixed

internal/datastore/mysql/migrations/zz_migration.0012_populate_schema_tables.go
josephschorr: Fixed

run-fdw-demo.sh
miparnisari, line 1: remove this file

internal/services/v1/permissions_queryplan.go
```diff
@@ -75,13 +75,18 @@ func (m *QueryPlanMetadata) ApplyAdvisor(co query.CanonicalOutline) (query.Canon
 // checkPermissionWithQueryPlan executes a permission check using the query plan API.
 // This builds an iterator tree from the schema and executes it against the datastore.
 func (ps *permissionServer) checkPermissionWithQueryPlan(ctx context.Context, req *v1.CheckPermissionRequest) (*v1.CheckPermissionResponse, error) {
-	atRevision, checkedAt, err := consistency.RevisionFromContext(ctx)
+	// Lazy initialization of QueryPlanMetadata if nil
```
miparnisari, line 78: this block deviates from `main`, please undo it (borked rebase)

internal/services/v1/permissions_test.go
```diff
@@ -320,6 +324,282 @@ func TestCheckPermissions(t *testing.T) {
...
+		0,
+		memdb.DisableGC,
+		true,
+		testserver.ServerConfig{
+			MaxUpdatesPerWrite:    1000,
+			MaxPreconditionsCount: 1000,
+			StreamingAPITimeout:   30 * time.Second,
+			DataLayerOpts: []datalayer.DataLayerOption{
+				datalayer.WithSchemaMode(datalayer.SchemaModeReadNewWriteBoth),
+				datalayer.WithSchemaCache(&simpleSchemaCache{items: make(map[datalayer.SchemaCacheKey]*datastore.ReadOnlyStoredSchema)}),
```
miparnisari, line 340: nit for a follow up PR: 
- this line has a lot of cognitive overhead. 
- you had to do a similar thing in `datalayer_test.go`, you defined `testSchemaCache` there as well with the exact same implementation

a refactor is due. Two ideas: either make `testSchemaCache` prod code so that whoever needs a really dumb cache can use it, or change what `WithSchemaCache`  receives.. instead of _an implementation_ make it receive a configuration object that maybe has a `testing` flag.

pkg/datalayer/schema_adapter.go
```diff
@@ -308,3 +320,236 @@ func writeSchemaViaLegacy(ctx context.Context, legacyWriter datastore.LegacySche
...
+) error {
+	if cache == nil {
+		cache = noopSchemaCache{}
+	}
```
miparnisari, line 488:
```suggestion
// WriteSchemaViaStoredSchema builds a StoredSchema proto and writes it via WriteStoredSchema.
// It also updates the cache (if not nil)
func WriteSchemaViaStoredSchema(ctx context.Context, rwt datastore.ReadWriteTransaction,
	definitions []datastore.SchemaDefinition, schemaString string, cache storedSchemaCache,
) error {
```

put the `nil` check in line 549

internal/datastore/spanner/revisions.go
```diff
@@ -3,48 +3,70 @@ package spanner
...
-	return revisions.NewForTime(now), nil
+	return datastore.RevisionWithSchemaHash{Revision: revisions.NewForTime(now), SchemaHash: schemaHash}, nil
 }
 
-func (sd *spannerDatastore) HeadRevision(ctx context.Context) (datastore.Revision, error) {
-	return sd.headRevisionInternal(ctx)
-}
-
-func (sd *spannerDatastore) now(ctx context.Context) (time.Time, error) {
+func (sd *spannerDatastore) nowWithHash(ctx context.Context) (time.Time, string, error) {
```
miparnisari, line 31: - in the method below, you are returning `""` which could lead to bugs i think. why not a sentinel?
- nit, for a followup PR: the schema hash is sometimes a `[]byte`, sometimes a `string` and sometimes a `SchemaHash`. i suggest replacing every `string` usage with `SchemaHash`

miparnisari (dismissed):
Tested locally 👍

internal/datastore/spanner/schema_chunker.go
```diff
@@ -116,8 +117,10 @@ func (t *spannerChunkedBytesTransaction) convertInsertToMutations(args []any) ([
 
 	cols := []string{colSchemaName, colSchemaChunkIndex, colSchemaChunkData, colTimestamp}
 	mutations := make([]*spanner.Mutation, 0, len(args)/argsPerRow)
-	for i := 0; i < len(args); i += argsPerRow {
-		vals := []any{args[i], args[i+1], args[i+2], spanner.CommitTimestamp}
+	for row := range slices.Chunk(args, argsPerRow) {
+		vals := make([]any, 0, 4)
+		vals = append(vals, row...)
+		vals = append(vals, spanner.CommitTimestamp)
```
tstirrat15, line 123: `gosec` was complaining about this line with concern about out-of-bounds access. I think it was a false positive, given that the logic was basically just chunking, but using `Chunk` makes things slightly more explicit and removes the possibility of an OOB error.

pkg/benchmarks/check_wide_direct.go
```diff
@@ -37,17 +36,7 @@ func setupCheckWideDirect(ctx context.Context, ds datastore.Datastore) (*QuerySe
...
-		SchemaString: schemaText,
-	}, compiler.AllowUnprefixedObjectType())
-	if err != nil {
-		return nil, err
-	}
-
-	_, err = ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
-		return rwt.LegacyWriteNamespaces(ctx, compiled.ObjectDefinitions...)
-	})
+	_, err := datalayer.WriteStoredSchemaForTest(ctx, ds, schemaText)
```
tstirrat15, line 39: I added a new helper for this and reused it in a few places. This solved an issue where attempting to reference the `revision.SchemaHash` was returning an empty string and failing the benchmark.

pkg/datalayer/testutil.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
+	"github.com/authzed/spicedb/pkg/schemadsl/input"
+)
+
+// WriteStoredSchemaForTest takes a datastore handle and a schema string
+// and writes a stored schema with hash to the datastore. Intended as
+// a convenient handle for test logic.
+func WriteStoredSchemaForTest(ctx context.Context, ds datastore.Datastore, schemaText string) (datastore.Revision, error) {
```
tstirrat15, line 14: This feels like something that can probably be used in a bunch of places, and probably has several implementations across the codebase. It'd be good to consolidate but I don't want to do that in the context of this PR.

pkg/query/benchmarks/check_benchmark_test.go
```diff
@@ -33,7 +33,7 @@ func BenchmarkCheck(b *testing.B) {
 			rawDS, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
 			require.NoError(b, err)
 			b.Cleanup(func() {
-				require.NoError(b, rawDS.Close())
```
tstirrat15, line 36: drive-by pet peeve kind of thing. `require` doesn't mean anything in Cleanup

tstirrat15 (commented):
See comments

internal/testfixtures/datastore.go
```diff
@@ -149,14 +147,14 @@ var StandardCaveatedRelationships = []string{
 }
 
 // EmptyDatastore returns an empty datastore for testing.
-func EmptyDatastore(ds datastore.Datastore, require *require.Assertions) (datastore.Datastore, datastore.Revision) {
-	rev, err := ds.HeadRevision(context.Background())
-	require.NoError(err)
+func EmptyDatastore(t testing.TB, ds datastore.Datastore) (datastore.Datastore, datastore.Revision) {
+	rev, err := ds.HeadRevision(t.Context())
```
tstirrat15, line 151: I'd say this change is what motivated most of it - there's no reason not to use `t.Context()` here except that we didn't make it available at this callsite for some reason.

tstirrat15 (commented):
See comment

tstirrat15 (dismissed):
LGMT

tstirrat15 (dismissed):
LGTM

tstirrat15 (dismissed):
LGTM

tstirrat15 (dismissed):
LGTM

internal/datastore/crdb/crdb_test.go
```diff
@@ -784,9 +790,9 @@ func TransactionMetadataMarkingTest(t *testing.T, rawDS datastore.Datastore) {
 func StreamingWatchTest(t *testing.T, rawDS datastore.Datastore) {
 	require := require.New(t)
 
-	ds, rev := testfixtures.DatastoreFromSchemaAndTestRelationships(rawDS, `
+	ds, rev := testfixtures.DatastoreFromSchemaAndTestRelationships(t, rawDS, `
 		caveat somecaveat(somecondition int) {
-			somecondition == 42
+			somecondition == 4t, 2
```
miparnisari: undo

internal/datastore/crdb/crdb_test.go
```diff
@@ -307,6 +307,10 @@ func TestWatchFeatureDetection(t *testing.T) {
 
 			tt.postInit(ctx, adminConn)
 
+			// Grant SELECT on schema_revision to unprivileged user so HeadRevision can read schema hashes.
```
miparnisari, line 310: this is weird, why is just in a test? hmm

tstirrat15 (dismissed):
LGTM

tstirrat15 (approved):
LGTM

pkg/datastore/test/namespace.go
```diff
@@ -161,7 +163,11 @@ func NamespaceDeleteTest(t *testing.T, tester DatastoreTester) {
 	require.NotNil(folderTpl)
 	tRequire.RelationshipExists(ctx, folderTpl, revision)
 
+	// TODO: it's as though the rwt is at the wrong revision?
```
miparnisari, line 166: is this comment still applicable?

PR #2927, opened 2026-02-28
test(query): Add a new benchmark for wide arrows

barakmich:

### DIFF
--- a/pkg/query/benchmarks/check_wide_arrow_benchmark_test.go
+++ b/pkg/query/benchmarks/check_wide_arrow_benchmark_test.go
@@ -1,0 +1,148 @@
+package benchmarks
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/internal/datastore/common"
+	"github.com/authzed/spicedb/internal/datastore/memdb"
+	"github.com/authzed/spicedb/pkg/datalayer"
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/query"
+	"github.com/authzed/spicedb/pkg/schema/v2"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
+	"github.com/authzed/spicedb/pkg/schemadsl/input"
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+// BenchmarkCheckWideArrow benchmarks permission checking through a wide arrow relationship.
+// This creates a scenario with:
+// - 10 files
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Added
+- testing(query planner): add a wide arrow benchmark to test query plan performance in the small (#2927)
```
miparnisari: please remove this; this change has no impact to end users. add the label `skip-changelog` to the PR

pkg/query/benchmarks/check_wide_arrow_benchmark_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// The subject we're checking: user:user15
+	// This user should have access through multiple groups
+	subject := query.NewObject("user", "user15").WithEllipses()
+
+	// Reset the timer - everything before this is setup
+	b.ResetTimer()
+
+	// Run the benchmark
+	for b.Loop() {
```
miparnisari, line 134: TIL `b.Loop`!

> // Loop resets the benchmark timer the first time it is called in a benchmark,
// so any setup performed prior to starting the benchmark loop does not count
// toward the benchmark measurement. Likewise, when it returns false, it stops
// the timer so cleanup code is not measured.

> // Prior to the introduction of Loop, benchmarks were expected to contain an
// explicit loop from 0 to b.N. Benchmarks should either use Loop or contain a
// loop to b.N, but not both. Loop offers more automatic management of the
// benchmark timer, and runs each benchmark function only once per measurement,
// whereas b.N-based benchmarks must run the benchmark function (and any
// associated setup and cleanup) several times.

pkg/query/benchmarks/check_wide_arrow_benchmark_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// The resource we're checking: file:file0
+	resources := query.NewObjects("file", "file0")
+
+	// The subject we're checking: user:user15
+	// This user should have access through multiple groups
+	subject := query.NewObject("user", "user15").WithEllipses()
+
+	// Reset the timer - everything before this is setup
+	b.ResetTimer()
```
jzelinskie, line 131: You don't need this when b.Loop is used, although maybe better safe than sorry.

CHANGELOG.md
barakmich: I went ahead and did this. But, further conversation here.

miparnisari (approved):

PR #2928, opened 2026-02-28
feat(query): add a PlanAdvisor framework in place of StatisticsSource

barakmich:

### DIFF
--- a/pkg/query/mutations.go
+++ b/pkg/query/mutations.go
@@ -7,25 +7,14 @@ type OutlineMutation func(Outline) Outline
 // MutateOutline performs a bottom-up traversal of the outline tree, applying
 // all the given transformation functions to each node after processing its children.
 func MutateOutline(outline Outline, fns []OutlineMutation) Outline {
-	// Recurse on children first (bottom-up)
-	if len(outline.SubOutlines) > 0 {
-		newSubs := make([]Outline, len(outline.SubOutlines))
-		for i, sub := range outline.SubOutlines {
-			newSubs[i] = MutateOutline(sub, fns)
+	// WalkOutlineBottomUp cannot fail when the callback never returns an error,
+	// so the error return is safe to ignore here.
+	result, _ := WalkOutlineBottomUp(outline, func(node Outline) (Outline, error) {
+		for _, fn := range fns {
+			node = fn(node)
 		}
-		outline = Outline{
-			Type:        outline.Type,
-			Args:        outline.Args,
-			SubOutlines: newSubs,
-			ID:          outline.ID,
-		}
-	}
-
-	// Then apply all mutation functions in sequence to current node
-	result := outline
-	for _, fn := range fns {
-		result = fn(result)
-	}
+		return node, nil
+	})
 	return result
 }
 
--- a/pkg/query/advisor.go
+++ b/pkg/query/advisor.go
@@ -58,64 +58,43 @@ func ApplyAdvisor(co CanonicalOutline, advisor PlanAdvisor) (CanonicalOutline, e
 	}, nil
 }
 
-// applyAdvisorMutations recursively walks the outline tree bottom-up.
+// applyAdvisorMutations walks the outline tree bottom-up via WalkOutlineBottomUp.
 // For each node it calls GetMutations on the advisor and applies the returned
 // mutations in sequence. After each mutation it verifies that the resulting
 // node's ID matches the original node's ID; a mismatch is a programmer bug.
 func applyAdvisorMutations(outline Outline, co CanonicalOutline, advisor PlanAdvisor) (Outline, error) {
-	// Recurse into children first (bottom-up).
-	if len(outline.SubOutlines) > 0 {
-		newSubs := make([]Outline, len(outline.SubOutlines))
-		for i, sub := range outline.SubOutlines {
-			mutated, err := applyAdvisorMutations(sub, co, advisor)
-			if err != nil {
-				return Outline{}, err
+	return WalkOutlineBottomUp(outline, func(node Outline) (Outline, error) {
+		mutations, err := advisor.GetMutations(node, co)
+		if err != nil {
+			return Outline{}, err
+		}
+		result := node
+		for _, fn := range mutations {
+			next := fn(result)
+			if next.ID != result.ID {
+				return Outline{}, spiceerrors.MustBugf(
+					"advisor mutation changed outline node ID from %d to %d; mutations must preserve the root node ID",
+					result.ID, next.ID,
+				)
 			}
-			newSubs[i] = mutated
+			result = next
 		}
-		outline = Outline{
-			Type:        outline.Type,
-			Args:        outline.Args,
-			SubOutlines: newSubs,
-			ID:          outline.ID,
-		}
-	}
-
-	mutations, err := advisor.GetMutations(outline, co)
-	if err != nil {
-		return Outline{}, err
-	}
-
-	result := outline
-	for _, fn := range mutations {
-		next := fn(result)
-		if next.ID != result.ID {
-			return Outline{}, spiceerrors.MustBugf(
-				"advisor mutation changed outline node ID from %d to %d; mutations must preserve the root node ID",
-				result.ID, next.ID,
-			)
-		}
-		result = next
-	}
-	return result, nil
+		return result, nil
+	})
 }
 
-// collectAdvisorHints recursively walks the outline tree and calls GetHints on
-// the advisor for each node, storing any returned hints in the hints map keyed
-// by the node's ID.
+// collectAdvisorHints walks the outline tree pre-order via WalkOutlinePreOrder,
+// calling GetHints on the advisor for each node and storing any returned hints
+// in the hints map keyed by the node's ID.
 func collectAdvisorHints(outline Outline, co CanonicalOutline, advisor PlanAdvisor, hints map[OutlineNodeID][]Hint) error {
-	nodeHints, err := advisor.GetHints(outline, co)
-	if err != nil {
-		return err
-	}
-	if len(nodeHints) > 0 {
-		hints[outline.ID] = nodeHints
-	}
-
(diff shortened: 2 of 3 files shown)


pkg/query/arrow_reversal_test.go
```diff
@@ -0,0 +1,159 @@
...
+	dsSchema, err := schema.BuildSchemaFromDefinitions(compiled.ObjectDefinitions, nil)
+	require.NoError(t, err)
+
+	canonicalOutline, err := BuildOutlineFromSchema(dsSchema, "file", "viewer")
+	require.NoError(t, err)
+
+	resources := NewObjects("file", "file0")
+	subject := NewObject("user", "user42").WithEllipses()
+
+	reader := NewQueryDatastoreReader(datalayer.NewDataLayer(rawDS).SnapshotReader(revision))
```
miparnisari: `NewQueryDatastoreReader` is undefined?

pkg/query/hint.go
```diff
@@ -0,0 +1,15 @@
+package query
+
+import "github.com/authzed/spicedb/pkg/spiceerrors"
+
+type Hint func(it Iterator) error
```
miparnisari, line 5:
```suggestion
// Hint can mutate the underlying Iterator
type Hint func(it Iterator) error
```

I asked Claude the conceptual difference between `hint` and `outlinemutation` and it gave me:

image

I wonder whether a better name for `hint` would be `AfterCompilationMutation`, and `OutlineMutation` would be a `BeforeCompilationMutation`

miparnisari (dismissed):
LGTM but please fix the compilation errors

pkg/query/advisor_combine.go
```diff
@@ -0,0 +1,38 @@
+package query
+
+// CombinePlanAdvisors creates a PlanAdvisor that combines multiple advisors
+// by running through them in order and returning the first non-nil result
```
tstirrat15, line 4: > returning the first non-nil result

Does that mean that the list implies a priority, and the highest-priority hint/mutation is applied and the rest are ignored?

pkg/query/advisor_count.go
```diff
@@ -0,0 +1,53 @@
...
+// defaultArrowFanout is a baseline fan-out to assume when there is no data for a direction.
+// Put another way, it has to be at least this bad on average to consider flipping the arrow direction.
+const defaultArrowFanout = 3.0
+
+// CountAdvisor uses observed CountStats keyed by CanonicalKey to hint whether
+// arrow iterators should be reversed. It compares the IterSubjectsResults of the
+// left subtree against the IterResourcesResults of the right subtree: if the left
+// fan-out is, on average, wider than the right fan-out, starting from the right is likely cheaper.
+type CountAdvisor struct {
+	stats map[CanonicalKey]CountStats
```
tstirrat15, line 12: Any thread safety concerns in this context?

pkg/query/advisor_count.go
```diff
@@ -0,0 +1,53 @@
...
+
+// NewCountAdvisor creates a CountAdvisor from a snapshot of observed stats.
+func NewCountAdvisor(stats map[CanonicalKey]CountStats) *CountAdvisor {
+	return &CountAdvisor{stats: stats}
+}
+
+// GetHints returns an ArrowDirectionHint for arrow nodes when observed result
+// ratios suggest reversal is beneficial. For all other node types it returns nil.
+func (a *CountAdvisor) GetHints(outline Outline, keySource CanonicalKeySource) ([]Hint, error) {
+	if outline.Type != ArrowIteratorType || len(outline.SubOutlines) != 2 {
```
tstirrat15, line 23: Oh demorgans

tstirrat15: Is there ever a case where you'd have an arrow type and the length of SubOutlines wouldn't be 2? Or is this belt-and-suspenders?

pkg/query/advisor_combine.go
tstirrat15: Or is the idea that they won't tend to point at the same nodes?

pkg/query/advisor_count_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// keySourceFromMap is a minimal CanonicalKeySource for tests, backed by an
+// explicit map from OutlineNodeID to CanonicalKey.
+type keySourceFromMap map[OutlineNodeID]CanonicalKey
```
tstirrat15, line 11: Huh, defining a named type of a primitive type and then defining methods on it is something I'm sure i've seen before but never properly grokked

pkg/query/advisor_static.go
tstirrat15, line 1: So a "mutation" is a modification to the structure of the outline, where a "hint" is a modification to the state/execution context of a particular iterator, more or less?

pkg/query/advisor_static.go
```diff
@@ -0,0 +1,429 @@
...
+		if err != nil || !shouldRotate {
+			return nil, err
+		}
+		return []OutlineMutation{mutation}, nil
+	}
+
+	return nil, nil
+}
+
+// tryRotateLeft checks if rotating (A->B)->C to A->(B->C) would reduce cost.
```
tstirrat15, line 164: Curious about "rotate" here - is there a sense in which reassociation is the same thing as rotation?

pkg/query/arrow_reversal_test.go
```diff
@@ -0,0 +1,159 @@
...
+// arrow directions to RTL) and asserts the advised plan returns the same set
+// of paths.
+func TestDoubleWideArrowAdvisedMatchesPlain(t *testing.T) {
+	t.Parallel()
+
+	const (
+		numFiles      = 5
+		numOrgs       = 29  // prime
+		numGroups     = 97  // prime
+		numUsers      = 997 // prime
```
tstirrat15, line 32: For fun?

pkg/query/arrow_reversal_test.go
```diff
@@ -0,0 +1,159 @@
...
+	`
+
+	compiled, err := compiler.Compile(compiler.InputSchema{
+		Source:       input.Source("test"),
+		SchemaString: schemaText,
+	}, compiler.AllowUnprefixedObjectType())
+	require.NoError(t, err)
+
+	_, err = rawDS.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+		return rwt.LegacyWriteNamespaces(ctx, compiled.ObjectDefinitions...)
```
tstirrat15, line 69: Is there a reason not to use the new method here?

pkg/query/hint.go
tstirrat15: Or maybe "provides context for iterator execution?"

I don't think it necessarily modifies the behavior of the iterator - I think that's up to the executor.

pkg/query/mutations.go
```diff
@@ -0,0 +1,119 @@
...
+//
+// If rotateLeft is false:
+//   - Transforms A->(B->C) into (A->B)->C
+//   - Requires: outline is ArrowIteratorType with right child also ArrowIteratorType
+//
+// Returns outline unchanged if preconditions are not met or if the outline
+// is not an arrow type.
+func RotateArrowMutation(rotateLeft bool) OutlineMutation {
+	return func(outline Outline) Outline {
+		// Must be an arrow with exactly 2 children
```
tstirrat15, line 57: When is "not 2 children" going to be true?

pkg/query/hint.go
barakmich: Correct. It's a hint on how to evaluate the iterator, not the structure of the query (both branches need to happen, it's a question on what order) -- arguably, reordering the ands and ors could be a hint as well; perhaps that's preferable in the future, but tbd, cause I kinda wanted two ways of working on these trees

pkg/query/advisor_combine.go
barakmich: Yep, highest priority first. This is just a convenience I put in so that we can do something like "follow the stats. if the stats say nothing at all, or have no opinion here, fallback to pure balance" or something

pkg/query/advisor_count.go
barakmich: Not from the advisor -- this map is immutable (I could add a comment there though)

pkg/query/advisor_count.go
barakmich: Belt-and-suspenders to be sure

pkg/query/advisor_count_test.go
barakmich: Yep; that's a trick. See also `CanonicalKey` itself, which is just a string (that knows how to hash itself, among other tricks)

pkg/query/advisor_static.go
barakmich: Yep!

pkg/query/advisor_static.go
barakmich: "Rotate" in the sense of the tree -- rotating a tree takes 

```
  A
 / \
B   C
 \ 
  D
```

If rotated right, it's C at the top (with A as left branch) and rotated left it's B at the top, so
```
  B
 / \
D   A
     \
      C 
```

(These are binary trees, but the idea generalizes (and in fact, arrows are always binary))

pkg/query/arrow_reversal_test.go
barakmich: and for steps that are other integers. can't be anything but relatively prime with a prime. If we had four orgs and step 2, then 0, 2, 0, 2 ... causing duplicate edges. With 5, (or any relative prime) you're guaranteed never to loop until you've hit everyone. (0, 2, 4, 1, 3)

pkg/query/arrow_reversal_test.go
barakmich: I suppose not. This is what happens when things are happening simultaneously. Worth a quick follow-up when I get a few more things merged; i'll plan time for that

PR #2929, opened 2026-02-28
chore(services): store observed counts in permissionsServer, add LR/LS flags

barakmich:
## Description
Add the "lr" and "ls" strings as inputs for `--experimental-query-plan`. Adds testing for the gRPC routes along with the existing consistency tests. And in so doing, found another bug: we want to combine the caveat _both_ from the wildcard, _and_ the relationship itself, which may or may not have one

### DIFF
--- a/internal/services/v1/permissions_queryplan.go
+++ b/internal/services/v1/permissions_queryplan.go
@@ -247,7 +247,10 @@ func (ps *permissionServer) lookupResourcesWithQueryPlan(req *v1.LookupResources
 		return ps.rewriteError(ctx, err)
 	}
 
-	optimized, err := queryopt.ApplyOptimizations(co, queryopt.StandardOptimzations)
+	optimized, err := queryopt.ApplyOptimizations(co, queryopt.StandardOptimzations, queryopt.RequestParams{
+		SubjectType:     req.Subject.Object.ObjectType,
+		SubjectRelation: req.Subject.OptionalRelation,
+	})
 	if err != nil {
 		return ps.rewriteError(ctx, err)
 	}
@@ -364,7 +367,10 @@ func (ps *permissionServer) lookupSubjectsWithQueryPlan(req *v1.LookupSubjectsRe
 		return ps.rewriteError(ctx, err)
 	}
 
-	optimized, err := queryopt.ApplyOptimizations(co, queryopt.StandardOptimzations)
+	optimized, err := queryopt.ApplyOptimizations(co, queryopt.StandardOptimzations, queryopt.RequestParams{
+		SubjectType:     req.SubjectObjectType,
+		SubjectRelation: req.OptionalSubjectRelation,
+	})
 	if err != nil {
 		return ps.rewriteError(ctx, err)
 	}


internal/services/v1/permissions_queryplan.go
```diff
@@ -13,9 +14,57 @@ import (
 	"github.com/authzed/spicedb/pkg/schema/v2"
 )
 
+// QueryPlanMetadata aggregates CountStats for iterator canonical keys across multiple queries.
+// This allows tracking which parts of query plans are used most frequently.
+type QueryPlanMetadata struct {
+	mu    sync.Mutex
+	stats map[query.CanonicalKey]query.CountStats // GUARDED_BY(mu)
```
miparnisari, line 24: should we put a cap on how much this map can grow?

barakmich: It'll grow only as the schema changes. For very large schemas, it may touch ~1000 entries. Rapidly changing large schemas with no overlap could make this grow... but I'm not worried about it. Plus, we're still behind the experimental flag and I think we'll have to come back to it.

pkg/query/intersection_test.go
```diff
@@ -503,14 +503,12 @@ func TestIntersectionIterSubjects(t *testing.T) {
...
-	ctx := &Context{
-		Context:  t.Context(),
-		Executor: LocalExecutor{},
-	}
-
 	t.Run("SimpleIntersection", func(t *testing.T) {
 		t.Parallel()
 
+		ctx := NewTestContext(t)
+		ctx.Context = t.Context()
```
miparnisari: nit: line 23 of `testutil.go` is already doing this

internal/services/integrationtesting/consistency_test.go
```diff
@@ -234,6 +234,180 @@ func TestConsistency(t *testing.T) {
...
+					)
+				})
+			}
+		})
+}
+
+// validateQueryPlanLookupSubjects checks that for each resource × subject-type
+// pair the query plan handler returns at least the subjects the accessibility
+// set defines as directly accessible. Wildcard exclusion checking is omitted
+// because the query plan handler does not yet populate ExcludedSubjects.
```
miparnisari, line 331: TIL. is there a ticket tracking this work?

barakmich: There is now

pkg/query/intersection_arrow_test.go
```diff
@@ -276,21 +276,18 @@ func TestIntersectionArrowIteratorCaveatCombination(t *testing.T) {
...
 		t.Parallel()
 
+		ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+		require.NoError(err)
+		revision, err := ds.ReadWriteTx(context.Background(), func(ctx context.Context, tx datastore.ReadWriteTransaction) error {
+			return nil
+		})
+		require.NoError(err)
+		ctx := NewLocalContext(context.Background(),
+			WithRevisionedReader(datalayer.NewDataLayer(ds).SnapshotReader(revision)))
```
miparnisari: Please, here and below, use a mock instead:

```go
defer ctrl.Finish()
ds := mock_datalayer.NewMockRevisionedReader(ctrl)

ctx := NewLocalContext(context.Background(), WithRevisionedReader(ds))
```

also, i'm surprised that the mock is never invoked. And i'm not just talking about this test, i'm saying that in every test here where i did this change, the mock is never invoked. Is this right?

pkg/query/datastore.go
```diff
@@ -273,37 +315,38 @@ func (r *DatastoreIterator) iterSubjectsWildcardImpl(ctx *Context, resource Obje
 				allSubjectsResource,
 				noResourceRelation,
 				subjectType,
-				r.base.Caveat() != "", r.base.Expiration(),
+				false, false, // fetch all — we override caveats below
```
miparnisari: seems like we are dropping the expiration here and in line 293?

pkg/query/intersection_arrow_test.go
barakmich: Why make a mock when we can just build a real one?

miparnisari: lots of reasons:
1. mocks are way faster
2. mocks are way easier to control and inspect (see my observation above)
3. if the signature of `NewMemdbDatastore` or of `ReadWriteTx` changes, you don't have to change anything here

barakmich: (1) In the grand scheme of things, it's a couple extra milliseconds

(3) If the signature changes, I feel I actively _want_ to change something here, to make sure I'm using it right

As for (2), a matter of taste.

barakmich: But, I wonder if I can't obviate it entirely

internal/services/integrationtesting/query_plan_consistency_test.go
```diff
@@ -78,20 +78,23 @@ func runQueryPlanConsistencyForFile(t *testing.T, filePath string) {
 
 	t.Run("lookup_resources", func(t *testing.T) {
 		if os.Getenv("TEST_QUERY_PLAN_RESOURCES") == "" {
-			t.Skip("Skipping IterResources tests: set TEST_QUERY_PLAN_RESOURCES=true to enable")
+			t.Skip("Skipping IterResources tests due to deprectation: Set TEST_QUERY_PLAN_RESOURCES=true to enable")
```
miparnisari, line 81: i'm confused; what deprecation?

pkg/query/intersection_arrow_test.go
miparnisari: > But, I wonder if I can't obviate it entirely

yes, this is what i pointed out

pkg/query/datastore.go
miparnisari: @barakmich i see you resolved the comment but nothing changed; did i misunderstand something in the code?

miparnisari: nvm i see the code changed now

PR #2931, opened 2026-03-01
chore: clean up all unused parameters

mishto:
Fixes [#2921](https://github.com/authzed/spicedb/issues/2921)

## Description
This PR removes unused parameters from functions across the codebase.

`unparam` still flags a number of method return parameters that are always nil with messages such as:
```
internal/caveats/run.go:236:84: (*CaveatRunner).runExpressionWithCaveats$1 - result 1 (error) is always nil
```
and parameters that always receive the same value with messages such as:
```
internal/datastore/proxy/indexcheck/queryshapevalidators.go:113:108: validateSubjectRelation - required always receives true
```

I am happy to make those changes here or in a followup PR if desired.

## Testing

Existing unit tests.

### DIFF
--- a/internal/services/shared/schema.go
+++ b/internal/services/shared/schema.go
@@ -30,11 +30,12 @@ type ValidatedSchemaChanges struct {
 	// additiveOnly indicates whether all operations should be applied
 	// or only those operations which are additive (i.e. not changes or deletions)
 	additiveOnly bool
+	schemaText   string
 }
 
 // ValidateSchemaChanges validates the schema found in the compiled schema and returns a
 // ValidatedSchemaChanges, if fully validated.
-func ValidateSchemaChanges(ctx context.Context, compiled *compiler.CompiledSchema, caveatTypeSet *caveattypes.TypeSet, additiveOnly bool) (*ValidatedSchemaChanges, error) {
+func ValidateSchemaChanges(ctx context.Context, compiled *compiler.CompiledSchema, caveatTypeSet *caveattypes.TypeSet, additiveOnly bool, schemaText string) (*ValidatedSchemaChanges, error) {
 	// 1) Validate the caveats defined.
 	newCaveatDefNames := mapz.NewSet[string]()
 	for _, caveatDef := range compiled.CaveatDefinitions {
@@ -67,6 +68,7 @@ func ValidateSchemaChanges(ctx context.Context, compiled *compiler.CompiledSchem
 		newCaveatDefNames:    newCaveatDefNames,
 		newObjectDefNames:    newObjectDefNames,
 		additiveOnly:         additiveOnly,
+		schemaText:           schemaText,
 	}, nil
 }
 
@@ -266,7 +268,7 @@ func ApplySchemaChangesOverExisting(
 		definitions = append(definitions, unchangedDefinitions...)
 
 		// WriteSchema will handle writing new/changed definitions and deleting removed ones
-		if err := rwt.WriteSchema(ctx, definitions); err != nil {
+		if err := rwt.WriteSchema(ctx, definitions, validated.schemaText, caveatTypeSet); err != nil {
 			return nil, err
 		}
 	}
--- a/pkg/datalayer/datalayer.go
+++ b/pkg/datalayer/datalayer.go
@@ -6,6 +6,7 @@ import (
 
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 
+	caveattypes "github.com/authzed/spicedb/pkg/caveats/types"
 	"github.com/authzed/spicedb/pkg/datastore"
 	"github.com/authzed/spicedb/pkg/datastore/options"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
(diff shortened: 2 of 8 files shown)


mishto:

tstirrat15:
If this is something that can be linted with `golangci-lint`, can you add this to the configuration?

mishto:
@tstirrat15 I added `unparam` to the linter ignoring the other types of errors this linter flags.

.golangci.yaml
```diff
@@ -95,6 +96,9 @@ linters:
       - linters:
           - "perfsprint"
         path: '(.+)_test\.go'
+      - linters:
+          - "unparam"
+        text: "result .* is always|result .* is never used|always receives"
```
mishto, line 101: This ignores linting issues such as: 
1. `result 1 (error) is always nil`
2. `result 0 (bool) is never used`
3. `expectDisconnect always receives false`

pkg/datalayer/impl.go
```diff
@@ -143,7 +143,7 @@ func (t *readWriteTransaction) BulkLoad(ctx context.Context, iter datastore.Bulk
 }
 
 func (t *readWriteTransaction) WriteSchema(ctx context.Context, definitions []datastore.SchemaDefinition, schemaString string, caveatTypeSet *caveattypes.TypeSet) error {
```
miparnisari, line 145: why did the linter not yell here?

why not remove `schemaString` and `caveatTypeSet` from line 145?  you will have to make an update to the interface too (cc @josephschorr)

mishto: Fixed, good catch. The linter ignores a "funcs that may satisfy an interface" among other exceptions to minimize false positives.

.golangci.yaml
tstirrat15: I wonder if the `always receives` are worth fixing. It doesn't have to be in the context of this PR, though.

internal/services/shared/schema.go
```diff
@@ -268,7 +268,7 @@ func ApplySchemaChangesOverExisting(
 		definitions = append(definitions, unchangedDefinitions...)
 
 		// WriteSchema will handle writing new/changed definitions and deleting removed ones
-		if err := rwt.WriteSchema(ctx, definitions, validated.schemaText, caveatTypeSet); err != nil {
+		if err := rwt.WriteSchema(ctx, definitions); err != nil {
```
miparnisari: `validated.schemaText` has no use anymore - please remove it

mishto: done. also removed the parameter even though this is a public method and it doesn't get flagged in `unparam`.

pkg/schemautil/schemautil.go
```diff
@@ -12,16 +12,15 @@ import (
...
 
 func ValidateSchemaChangesWithCaveatTypeSet(
 	ctx context.Context,
 	compiled *compiler.CompiledSchema,
 	caveatTypeSet *caveattypes.TypeSet,
 	isAdditiveOnly bool,
-	schemaText string,
 ) (*shared.ValidatedSchemaChanges, error) {
 	return shared.ValidateSchemaChanges(ctx, compiled, caveatTypeSet, isAdditiveOnly)
 }
```
mishto, line 27: These two methods are actually not used. Should I remove them while at it?

tstirrat15: Yes please!

mishto: Done!

pkg/datalayer/datalayer.go
```diff
@@ -123,7 +122,7 @@ type ReadWriteTransaction interface {
 	BulkLoad(ctx context.Context, iter datastore.BulkWriteRelationshipSource) (uint64, error)
 
 	// WriteSchema writes the full set of schema definitions.
-	WriteSchema(ctx context.Context, definitions []datastore.SchemaDefinition, schemaString string, caveatTypeSet *caveattypes.TypeSet) error
+	WriteSchema(ctx context.Context, definitions []datastore.SchemaDefinition) error
```
tstirrat15: Apologies on this point - this particular interface is written this way because it's a part of a multiphase piece of work to change how schemas are stored in the datastore. The args are currently unused but will be used under that new code. Can we roll this change and its associates back out?

mishto: No problem. I’ll get to this later today or tomorrow.

pkg/schemautil/schemautil.go
```diff
@@ -7,25 +7,8 @@ import (
...
-	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 )
 
-// ValidateSchemaChanges validates the schema found in the compiled schema and returns a
-// ValidatedSchemaChanges, if fully validated.
-func ValidateSchemaChanges(ctx context.Context, compiled *compiler.CompiledSchema, isAdditiveOnly bool, schemaText string) (*shared.ValidatedSchemaChanges, error) {
-	return ValidateSchemaChangesWithCaveatTypeSet(ctx, compiled, caveattypes.Default.TypeSet, isAdditiveOnly, schemaText)
-}
-
-func ValidateSchemaChangesWithCaveatTypeSet(
```
miparnisari, line 19: please bring these back; we have private code that uses these

pkg/datalayer/impl.go
mishto: @miparnisari FYI, I reverted this based on @tstirrat15 [request](https://github.com/authzed/spicedb/pull/2931#discussion_r2886597383)

.golangci.yaml
mishto: I also think that private functions always returning nil errors can add unnecessary (and non-testable) error handling throughout the code. There seem to be quite a few of those.

pkg/schemautil/schemautil.go
mishto: done

pkg/datalayer/datalayer.go
tstirrat15: Thank you!

tstirrat15:
I appreciate your patience in the review process too.

PR #2937, opened 2026-03-02
chore: use correct logger in import logic

tstirrat15:
## Description
I used the incorrect logger when I originally wrote this, and then I copied it when I moved import logic into `schemadsl`. This fixes it.

## Changes
* Use our internal `logging` package instead of a bare reference to zerolog
## Testing
Review.


josephschorr (approved):
LGTM

PR #2941, opened 2026-03-03
refactor: remove metrics that are not useful or not written to

miparnisari:


internal/datastore/crdb/watch.go
```diff
@@ -37,18 +36,6 @@ const (
 	queryChangefeedPreV22 = "EXPERIMENTAL CHANGEFEED FOR %s WITH updated, cursor = '%s', resolved = '%s';"
 )
 
-var retryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
```
miparnisari, line 40: FYI not written

internal/datastore/spanner/watch.go
```diff
@@ -28,18 +27,6 @@ const (
 	CombinedChangeStreamName = "combined_change_stream"
 )
 
-var retryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
```
miparnisari, line 31: FYI not written

internal/graph/check.go
```diff
@@ -39,16 +39,9 @@ var dispatchChunkCountHistogram = prometheus.NewHistogram(prometheus.HistogramOp
 	Buckets: []float64{1, 2, 3, 5, 10, 25, 100, 250},
 })
 
-var directDispatchQueryHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
```
miparnisari, line 42: FYI not useful, metric can only take 2 values

PR #2942, opened 2026-03-03
chore(datastore): expose GC interfaces, errors, and context proxy

ecordell:
## Description

This is an internal change to make testing datastores simpler.

## Testing

Confirmed that existing tests pass


internal/datastore/context.go
```diff
@@ -1,150 +1,10 @@
...
-
-func (p *ctxProxy) Statistics(ctx context.Context) (datastore.Stats, error) {
-	return p.delegate.Statistics(context.WithoutCancel(ctx))
-}
-
-func (p *ctxProxy) ReadyState(ctx context.Context) (datastore.ReadyState, error) {
-	return p.delegate.ReadyState(context.WithoutCancel(ctx))
+// NewSeparatingContextDatastoreProxy delegates to pkg/datastore.
+func NewSeparatingContextDatastoreProxy(d ds.Datastore) ds.StrictReadDatastore {
+	return ds.NewSeparatingContextDatastoreProxy(d)
```
tstirrat15: Why do we have this proxy if it's just a wrapper? Like why aren't we using the ds proxy directly?

tstirrat15 (commented):
See comment

internal/datastore/context.go
ecordell: I did this is an intermediate state to make sure the existing code continued to work; I removed the others but looks like I forgot this one. will update

pkg/datastore/gc_test.go
```diff
@@ -190,7 +160,7 @@ func alwaysErr() error {
 
 func TestGCFailureBackoff(t *testing.T) {
 	t.Cleanup(func() {
-		goleak.VerifyNone(t)
+		goleak.VerifyNone(t, goleak.IgnoreTopFunction("go.opencensus.io/stats/view.(*worker).start"))
```
miparnisari:
```suggestion
		goleak.VerifyNone(t, testutil.GoLeakIgnores()...)
```

internal/datastore/common/gc_test.go
```diff
@@ -220,41 +190,20 @@ func TestGCFailureBackoff(t *testing.T) {
 			break
 		}
 	}
-	// We expect about 5 failures; the behavior of the library means that there's some wiggle room here.
-	// (owing to the jitter in the backoff)
-	// we should see failures at 100s, 200s, 400s, 800s
-	// but depending on jitter this could end up being squished down such that we get 5 failures.
-	// Experimentally, we see 5 most often and 4 sometimes, so asserting greater than 3 works here.
```
miparnisari, line 227: why remove this and every other godoc? these tests are tricky so the comments are helpful

pkg/datastore/context.go
```diff
@@ -1 +1,153 @@
 package datastore
+
+import (
+	"context"
+
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// retryableDatastore is a local interface to avoid a circular import with pkg/datastore/test.
```
miparnisari: ooooff what? 😆

ecordell: fixed

I did spend time doing a "sentinal retryableerror", but I found that interacted poorly with the `pool.RetryableErr` - had to do a conversion, but only for CRDB.

Instead I settled on making the example retryable error part of the datastore tester itself, lmk what you think

internal/datastore/common/gc_test.go
ecordell: fixed

ecordell:
@tstirrat15 @miparnisari updated, but it looks like some unrelated test flakes

pkg/datastore/context.go
```diff
@@ -1 +1,143 @@
...
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// NewSeparatingContextDatastoreProxy severs any timeouts in the context being
+// passed to the datastore and only retains tracing metadata.
+//
+// This is useful for datastores that do not want to close connections when a
+// cancel or deadline occurs.
+func NewSeparatingContextDatastoreProxy(d Datastore) StrictReadDatastore {
```
miparnisari, line 15: i just noticed that Spanner does not use this proxy

pkg/datastore/test/transactions.go
```diff
@@ -13,6 +13,15 @@ import (
 )
 
 func RetryTest(t *testing.T, tester DatastoreTester) {
+	type retryableProvider interface {
+		RetryableError() error
+	}
+	rp, ok := tester.(retryableProvider)
```
miparnisari, line 19: not for this PR. But: we should refactor this so that the Test accepts a `RetryableDatastore` directly and then we don't have to do this ugly casting :p

pkg/datastore/test/datastore.go
```diff
@@ -32,12 +32,28 @@ func (f DatastoreTesterFunc) New(tb testing.TB, revisionQuantization, gcInterval
...
+	return &TesterFactory{retryErr: retryErr}
+}
 
-	ExampleRetryableError() error
+// NewTester wraps a DatastoreTester and adds a retryable error for use in RetryTest.
+func (f *TesterFactory) NewTester(tester DatastoreTester) DatastoreTester {
+	return &retryableTester{DatastoreTester: tester, retryErr: f.retryErr}
 }
 
+type retryableTester struct {
```
miparnisari, line 50: not for this PR. But: i'd love to refactor this. Reiterating this comment: https://github.com/authzed/spicedb/pull/2924#discussion_r2887934702, I generally dislike code that is made just for testing. In this particular case, we could make this a production-grade interface called "RetryableTransactionsDatastore"  with one method "ExampleRetryableError()"

i would also love to run tests based on interfaces implemented and not on "categories"

internal/datastore/proxy/observable_test.go
```diff
@@ -20,10 +20,8 @@ func (obs observableTest) New(tb testing.TB, revisionQuantization, _, gcWindow t
...
 
-func TestObservableProxy(t *testing.T) {
-	test.All(t, observableTest{}, true)
-}
+var observableFactory = test.NewTesterFactory(memdb.ErrSerialization)
 
-func (p *observableProxy) ExampleRetryableError() error {
-	return memdb.ErrSerialization
+func TestObservableProxy(t *testing.T) {
+	test.All(t, observableFactory.NewTester(observableTest{}), true)
```
miparnisari, line 26: not for this PR. But: why are we running the entire suite of tests for a proxy that _emits metrics_ :sob: Are we even checking that it is emitting metrics? this would be better done with unit tests.

miparnisari (approved):

pkg/datastore/test/transactions.go
ecordell: hmm...maybe I'm misremembering but I thought we had to match the signature `func(t *testing.T, tester DatastoreTester)` so that other parts of the test runner could treat them as sets i.e. `for _, test := range []func(t *testing.T, tester DatastoreTester)`

pkg/datastore/test/datastore.go
ecordell: I don't have a strong opinion but I personally don't mind code that's just for testing if it's just in test packages. 

I personally don't like when they leak into non-test packages.

pkg/datastore/context.go
ecordell: Yeah spanner manages connections differently. Killing a spanner connection is no big deal

PR #2947, opened 2026-03-04
chore: fix regression from #2912 - tracing in ReadSchema calls

miparnisari:
## Description

- Prevent missing data from trace because of `SchemaText()` losing the context information
- add more spans

Before:

before

After:

image

### DIFF
--- a/internal/middleware/memoryprotection/memory_protection.go
+++ b/internal/middleware/memoryprotection/memory_protection.go
@@ -9,12 +9,15 @@ import (
 	middleware "github.com/grpc-ecosystem/go-grpc-middleware/v2"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.opentelemetry.io/otel"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	log "github.com/authzed/spicedb/internal/logging"
 )
+
+var tracer = otel.Tracer("spicedb/internal/middleware/memory_protection")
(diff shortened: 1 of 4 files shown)


pkg/datalayer/schema_adapter.go
```diff
@@ -30,9 +30,7 @@ type legacySchemaReaderAdapter struct {
 
 // SchemaText returns the schema text at the current revision by reading all namespaces and caveats
 // and generating the schema text from them.
-func (l *legacySchemaReaderAdapter) SchemaText() (string, error) {
-	ctx := context.Background()
```
miparnisari, line 34: removing this line is the actual fix

pkg/schemadsl/generator/generator.go
```diff
@@ -18,18 +21,23 @@ import (
...
+
 // Ellipsis is the relation name for terminal subjects.
 const Ellipsis = "..."
 
 // MaxSingleLineCommentLength sets the maximum length for a comment to made single line.
 const MaxSingleLineCommentLength = 70 // 80 - the comment parts and some padding
 
 func GenerateSchema(definitions []compiler.SchemaDefinition) (string, bool, error) {
-	return GenerateSchemaWithCaveatTypeSet(definitions, caveattypes.Default.TypeSet)
+	return GenerateSchemaWithCaveatTypeSet(context.TODO(), definitions, caveattypes.Default.TypeSet)
```
tstirrat15, line 33: What does TODO mean? I've never quite understood that. Should it be a `.Background()`?

miparnisari: no idea 😆

PR #2950, opened 2026-03-05
feat: add tracing to proto validation

miparnisari:
## Description

add tracing to proto validation

## Testing

| Before | After |
| --- | --- |
| image image  | image image |
| image image  |  image image |

### DIFF
--- a/internal/middleware/handwrittenvalidation/handwrittenvalidation.go
+++ b/internal/middleware/handwrittenvalidation/handwrittenvalidation.go
@@ -2,68 +2,39 @@ package handwrittenvalidation
 
 import (
 	"context"
-	"errors"
-	"fmt"
 
-	"buf.build/go/protovalidate"
-	"go.opentelemetry.io/otel"
-	otelcodes "go.opentelemetry.io/otel/codes"
-	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"google.golang.org/protobuf/proto"
 )
-
-var tracer = otel.Tracer("spicedb/internal/middleware")
-
-// mustNewProtoValidator wraps protovalidate.New() to panic
-// if the validator can't be constructed.
-func mustNewProtoValidator(opts ...protovalidate.ValidatorOption) protovalidate.Validator {
-	validator, err := protovalidate.New(opts...)
-	if err != nil {
-		wrappedErr := fmt.Errorf("could not construct validator: %w", err)
-		panic(wrappedErr)
-	}
-	return validator
-}
 
 type handwrittenValidator interface {
 	HandwrittenValidate() error
 }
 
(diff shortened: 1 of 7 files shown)


internal/middleware/handwrittenvalidation/doc.go
```diff
@@ -1,2 +1,2 @@
-// Package handwrittenvalidation defines middleware that runs custom-made validations on incoming requests.
+// Package handwrittenvalidation defines middleware that runs validations on incoming requests.
 package handwrittenvalidation
```
miparnisari, line 2: maybe i should rename the package...

tstirrat15: We also might be able to get rid of this package? I haven't actually looked into it

internal/middleware/handwrittenvalidation/handwrittenvalidation.go
tstirrat15, line 1: I don't think folding together the validation logic makes sense to me. If the goal is just "add spans around an interceptor," I'd rather write a function that takes an interceptor and a name and returns a new interceptor that adds spans with that name around the provided interceptor, and then we can throw it around the protovalidate interceptor. I think there's a good chance that we can get rid of this handwritten logic altogether using CEL rules that reference multiple fields and I'd rather not tie them together.

miparnisari: > I don't think folding together the validation logic makes sense to me. 

why not? the responsibilties are the same: validate the incoming request.

the split can cause bugs. Look at the existing codebase today, in particular the Watch service: https://github.com/authzed/spicedb/blob/main/internal/services/v1/watch.go#L39. Notice something odd? if there were a handwritten validation defined, this code is buggy, it's not running it. 

> I'd rather write a function that takes an interceptor and a name and returns a new interceptor that adds spans with that name around the provided interceptor,

agree, i want to do this and put a span around EVERY interceptor because there are so many and i don't know where the time is being spent. but: separate PR

>  think there's a good chance that we can get rid of this handwritten logic altogether using CEL rules that reference multiple fields and I'd rather not tie them together.

agree, but: separate PR

internal/middleware/interceptorwrapper/interceptorwrapper.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"go.opentelemetry.io/otel"
+	"google.golang.org/grpc"
+)
+
+var tracer = otel.Tracer("spicedb/internal/middleware")
+
+// WrapUnaryServerInterceptorWithSpans returns a new interceptor that wraps the given interceptor
+// with a span, measuring the duration of the interceptor's pre-handler logic.
+func WrapUnaryServerInterceptorWithSpans(
```
miparnisari, line 14: @tstirrat15 is the idea to wrap more interceptors with this?

tstirrat15: That's the idea - it's here if we want it.

PR #2954, opened 2026-03-05
chore: add singleflightcheck analyzer to enforce context-aware singleflight

### LINKED ISSUE #2953: Use context-aware singleflight package across the repo
We use https://pkg.go.dev/resenje.org/singleflight in a handful of places to ensure that singleflights can't end up deadlocked. However, there's a few places where we we are using https://pkg.go.dev/golang.org/x/sync/singleflight instead, such as in schema cache logic, and this may be responsible for deadlock behavior that has been reported by a handful of users.

## TODO
- [ ] Add linter rule that says we can't use `golang.org/x/sync/singleflight`
- [ ] Replace all usages with the resenje package
- [ ] optional: add a wrapping library that forces the use of a timeout and use that across the codebase

josephschorr:
Adds a custom analyzer that detects uses of golang.org/x/sync/singleflight in functions with a context.Context parameter, requiring resenje.org/singleflight instead. The resenje variant is context-aware and supports generics, preventing context cancellation from being silently ignored in singleflighted calls.

Migrate the two remaining callsites (optimized revisions and schema caching) to resenje.org/singleflight.

Fixes #2953

### DIFF
--- a/internal/datastore/proxy/schemacaching/standardcache.go
+++ b/internal/datastore/proxy/schemacaching/standardcache.go
@@ -6,7 +6,7 @@ import (
 	"sync"
 	"unsafe"
 
-	"golang.org/x/sync/singleflight"
+	"resenje.org/singleflight"
 
 	"github.com/authzed/spicedb/pkg/cache"
 	"github.com/authzed/spicedb/pkg/datastore"
@@ -20,7 +20,7 @@ import (
 type definitionCachingProxy struct {
 	datastore.Datastore
 	c         cache.Cache[cache.StringKey, *cacheEntry]
-	readGroup singleflight.Group
+	readGroup singleflight.Group[string, *cacheEntry]
 }
 
 func (p *definitionCachingProxy) Close() error {
@@ -176,8 +176,7 @@ func readAndCache[T schemaDefinition](
 	loaded, found := r.p.c.Get(cache.StringKey(cacheRevisionKey))
 	if !found {
 		// We couldn't use the cached entry, load one
-		var err error
-		loadedRaw, err, _ := r.p.readGroup.Do(cacheRevisionKey, func() (any, error) {
+		loadedEntry, _, err := r.p.readGroup.Do(ctx, cacheRevisionKey, func(ctx context.Context) (*cacheEntry, error) {
 			// sever the context so that another branch doesn't cancel the
 			// single-flighted read
 			loaded, updatedRev, err := reader(context.WithoutCancel(ctx), name)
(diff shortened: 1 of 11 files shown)


internal/datastore/proxy/schemacaching/standardcache.go
```diff
@@ -176,8 +176,7 @@ func readAndCache[T schemaDefinition](
 	loaded, found := r.p.c.Get(cache.StringKey(cacheRevisionKey))
 	if !found {
 		// We couldn't use the cached entry, load one
-		var err error
-		loadedRaw, err, _ := r.p.readGroup.Do(cacheRevisionKey, func() (any, error) {
+		loadedEntry, _, err := r.p.readGroup.Do(ctx, cacheRevisionKey, func(ctx context.Context) (*cacheEntry, error) {
```
tstirrat15, line 179: For this and the other callsite: we're now passing the context which means it *can* be cancelled, but do we know that both of these codepaths have timeouts on them somewhere further up the chain?

miparnisari: i'm wondering why we didn't act defensively and added a context timeout anyway..

magefiles/lint.go
```diff
@@ -101,6 +101,7 @@ func (Lint) Analyzers() error {
 		"-protomarshalcheck",
 		"-telemetryconvcheck",
 		"-iferrafterrowclosecheck",
+		"-singleflightcheck",
```
miparnisari, line 104: use `depguard` please

```
settings:
    depguard:
      rules:
        main:
          deny:
            - pkg: "golang.org/x/sync/singleflight"
              desc: "use resenje.org/singleflight instead so that we can break out of deadlocks"
```

internal/datastore/proxy/schemacaching/standardcache.go
josephschorr: > i'm wondering why we didn't act defensively and added a context timeout anyway..

That was my proposal: we wrap this (and any other singleflight) calls in a timeout anyway, but since we didn't get consensus, I did this portion first

josephschorr: >  but do we know that both of these codepaths have timeouts on them somewhere further up the chain?

Yes, all requests have timeouts ultimately from the API layer that moves downward into here (except for writes, which break the context, but have their own timeouts)

magefiles/lint.go
tstirrat15: I think the idea was that there may be places where you want to singleflight and don't have access to a `context` object. Unless we're saying that we'd want to create a context with timeout at that callsite?

josephschorr: We can't use depguard: there are times when we need to use singleflight without context being available; the linter only checks the cases where context is available

PR #2955, opened 2026-03-05
test(*): remove all t.Parallel() calls in the tests

barakmich:

### DIFF
--- a/internal/services/integrationtesting/certtest/cert_test.go
+++ b/internal/services/integrationtesting/certtest/cert_test.go
@@ -1,6 +1,6 @@
 //go:build !skipintegrationtests
 
-package integrationtesting_test
+package certtest_test
 
 import (
 	"context"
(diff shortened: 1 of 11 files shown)


jzelinskie:
There have been various PRs over the years adding/removing t.Parallel everywhere.
It'd be great if we could codify it somewhere that you shouldn't add this unless you've really tested it makes things faster. Maybe an approach similar to replacing panic with our linter + MustBugf()?

tstirrat15:
I added it as a note in `CONTRIBUTING.md`. I'm disinclined to enforce anything programmatically because I think it's likely that there are going to be exceptions to our current rule.

internal/services/integrationtesting/certtest/cert_test.go
```diff
@@ -1,6 +1,6 @@
 //go:build !skipintegrationtests
 
-package integrationtesting_test
+package certtest_test
```
miparnisari, line 3: 👍

PR #2956, opened 2026-03-05
refactor(query): add QueryDatastoreReader to act as an implementation shim for datastores

barakmich:
## Description

Currently, there are very few calls in pkg/query direct to the datastore interface. We'd like to keep this interface minimal, in an effort to start to clean up datastores over time as we continue with query planner.

This has the added benefit of being a way to inject various testing scenarios directly with the query planner tests and benchmarks; for example, adding a slight delay on any call to simulate a network latency.

### DIFF
--- a/pkg/query/datastore.go
+++ b/pkg/query/datastore.go
@@ -74,11 +74,19 @@ func (r *DatastoreIterator) checkNormalImpl(ctx *Context, resources []Object, su
 }
 
 func (r *DatastoreIterator) checkWildcardImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
+	// Invariant: wildcard subjects in the datastore are always stored with the ellipsis
+	// relation. The "*" is only ever an ObjectID; "type:*#relation" is syntactically
+	// invalid and cannot be written. Any caller passing a non-ellipsis relation here
+	// would cause us to query with the wrong relation filter and return a false negative.
+	if subject.Relation != tuple.Ellipsis {
+		return nil, spiceerrors.MustBugf("checkWildcardImpl called with non-ellipsis subject relation %q for subject %s:%s; wildcard subjects are always stored with ellipsis relation", subject.Relation, subject.ObjectType, subject.ObjectID)
+	}
+
 	// Query the datastore for wildcard relationships (subject ObjectID = "*")
 	wildcardSubject := ObjectAndRelation{
 		ObjectType: subject.ObjectType,
 		ObjectID:   WildcardObjectID,
-		Relation:   subject.Relation,
+		Relation:   tuple.Ellipsis,
 	}
 
 	resourceType := ObjectType{Type: r.base.DefinitionName()}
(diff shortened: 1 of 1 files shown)


pkg/query/reader.go
```diff
@@ -0,0 +1,300 @@
...
+	resourceRelation string,
+	subject ObjectAndRelation,
+	withCaveats, withExpiration bool,
+) (PathSeq, error) {
+	resourceIDs := make([]string, len(resources))
+	for i, res := range resources {
+		resourceIDs[i] = res.ObjectID
+	}
+
+	// All resources in a DatastoreIterator share the same type.
```
miparnisari: is DatastoreIterator the only expected caller of this method?
more generally, maybe rephrase this to write an expectation of _all_ callers: "it is assumed that callers of this method pass in resources that share a type". i'm also thinking that we could be defensive on this and reuse the `for` loop above to validate this assumption, which won't cost any extra cpu cycles

```go
    resourceIDs := make([]string, len(resources))
	resourceType := resources[0].ObjectType
	for i, res := range resources {
		resourceIDs[i] = res.ObjectID
		if res.ObjectType != resourceType {
			return nil, fmt.Errorf("expected resource type %q but got %q", resourceType, res.ObjectType)
		}
	}
```

pkg/query/datastore.go
```diff
@@ -81,87 +50,52 @@ func (r *DatastoreIterator) CheckImpl(ctx *Context, resources []Object, subject
...
-			{
-				OptionalSubjectType: r.base.Type(),
-				OptionalSubjectIds:  []string{tuple.PublicWildcard}, // Look for "*" subjects
-				RelationFilter:      r.buildSubjectRelationFilter(),
-			},
-		},
+	wildcardSubject := ObjectAndRelation{
+		ObjectType: subject.ObjectType,
+		ObjectID:   WildcardObjectID,
+		Relation:   subject.Relation,
```
miparnisari: wouldn't it it be a bug if `subject.Relation != ""`?

pkg/query/datastore.go
miparnisari, line 1: for a follow up PR, can you please add more unit tests to this file? in https://app.codecov.io/gh/authzed/spicedb/blob/main/pkg%2Fquery%2Fdatastore.go i see that there are large blocks of code not covered by anything

pkg/query/datastore.go
```diff
@@ -284,153 +187,107 @@ func (r *DatastoreIterator) iterSubjectsWildcardImpl(ctx *Context, resource Obje
...
+	// by doing a CheckRelationships probe with subject ObjectID = "*".
+	subjectType := ObjectType{
+		Type:        r.base.Type(),
+		Subrelation: r.base.Subrelation(),
+	}
+
+	wildcardSubject := ObjectAndRelation{
+		ObjectType: r.base.Type(),
+		ObjectID:   WildcardObjectID,
+		Relation:   r.base.Subrelation(),
```
miparnisari, line 214: same as above, would it not be a bug to find a relationship where the subject is `something:*#somethingelse`?

pkg/query/caveat.go
```diff
@@ -219,6 +217,26 @@ func (c *CaveatIterator) SubjectTypes() ([]ObjectType, error) {
 	return c.subiterator.SubjectTypes()
 }
 
+// caveatDefinitionLookupAdapter wraps a QueryDatastoreReader to satisfy
+// caveats.CaveatDefinitionLookup (which takes a bulk name slice) by calling
+// LookupCaveatDefinition individually for each name.
+type caveatDefinitionLookupAdapter struct{ r QueryDatastoreReader }
+
+func (a caveatDefinitionLookupAdapter) LookupCaveatDefinitionsByNames(
```
tstirrat15, line 225: Is the idea with this shim that we'd eventually have a bulk method that isn't making individual requests?

barakmich: Or uses the underlying cache, but yes, that's an eventual benefit.

pkg/query/reader.go
barakmich: Yep, this is worth a signature change. Fixed.

pkg/query/datastore.go
barakmich: added invariants

pkg/query/datastore.go
barakmich: same invariant resolution

Issue #2829, opened 2026-01-16
Refactor Prometheus usage to explicitly pass a registry


tstirrat15:
We're currently implicitly using the default registry, which means that tests sometimes flake because two different tests attempt to add the same metric to the same registry, and there's functionality where you can explicitly set up and pass a registry: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-Advanced_Uses_of_the_Registry

If we moved to that approach, we could have each test provide its own registry and avoid duplicate registration.

tstirrat15:
This may mean we can't use `promauto`. Hopefully that isn't too big an issue.

miparnisari:
Slightly related: a side effect of registering metrics inside `func init()` is that if you run SpiceDB against Spanner (for example, via `docker-compose -f docker-compose.spanner.yaml up --build`), you can still see CRDB metrics in http://localhost:9090/metrics. (With zero value, but still...)

```
# HELP crdb_client_resets cockroachdb client-side tx reset distribution
# TYPE crdb_client_resets histogram
crdb_client_resets_bucket{le="0"} 0
crdb_client_resets_bucket{le="1"} 0
crdb_client_resets_bucket{le="2"} 0
crdb_client_resets_bucket{le="5"} 0
crdb_client_resets_bucket{le="10"} 0
crdb_client_resets_bucket{le="20"} 0
crdb_client_resets_bucket{le="50"} 0
crdb_client_resets_bucket{le="+Inf"} 0
crdb_client_resets_sum 0
crdb_client_resets_count 0
# HELP crdb_healthy_nodes the number of healthy crdb nodes detected by spicedb
# TYPE crdb_healthy_nodes gauge
crdb_healthy_nodes 0
```

PR #2957, opened 2026-03-06
test: prevent duplicate metrics collector registration attempted

miparnisari:
This PR fixes a test flake.

The other way to prevent this flake is to do what https://github.com/authzed/spicedb/issues/2829 says, but this is more work and riskier.


tstirrat15 (approved):
Yeah, this makes sense to me.

PR #2958, opened 2026-03-06
fix: postgres revision compare for semi-disjoint overlapping transactions

jakedt:
## Description

In certain strange cases, we may incorrectly order postgres transactions that were actually concurrent. This PR fixes those known cases.

## Testing

Unit testing

### DIFF
--- a/internal/datastore/postgres/snapshot.go
+++ b/internal/datastore/postgres/snapshot.go
@@ -191,7 +191,8 @@ func (s pgSnapshot) compare(rhs pgSnapshot) comparisonResult {
 // (in-progress) or any txid >= s.xmax (unseen). If other can see any of
 // those as committed, then other has strictly more information.
 func (s pgSnapshot) otherHasMoreInfo(other pgSnapshot) bool {
-	// Check if other sees any of s's in-progress transactions as committed
+	// If any of the in-progress transactions in this snapshot are visible (i.e. commited
+	// or rolled back) to the other snapshot, the other snapshot has more information.
 	if slices.ContainsFunc(s.xipList, other.txVisible) {
 		return true
 	}
@@ -200,12 +201,25 @@ func (s pgSnapshot) otherHasMoreInfo(other pgSnapshot) bool {
 	// Transactions in [s.xmax, other.xmax) that are NOT in other's xipList
 	// are committed from other's perspective but completely unknown to s.
 	// If the range contains more txids than other has in-progress in that
-	// range, at least one must be committed.
+	// range, at least one must be settled.
+
+	//  The following logic is functionally equivalent to iterating from
+	// `s.xmax` to `other.xmax` and asking whether each of those txids is present
+	// in `other.xipList`. if any are *not* present, that means that `other`
+	// knows that one of those txids has been settled, and therefore `other`
+	// has more information.
+
+	// Doing this naively (i.e. the iteration above) is O(n) on the size of `other.xipList`;
+	// the implementation below is equivalent but makes use of `slices.BinarySearch`
+	// to make it O(log(n)).
+
+	// This condition is only possible if the maximum transaction that the other snapshot
+	// is aware of is further along than the current snapshot.
 	if other.xmax > s.xmax {
 		rangeSize := other.xmax - s.xmax
 		lo, _ := slices.BinarySearch(other.xipList, s.xmax)
 		hi, _ := slices.BinarySearch(other.xipList, other.xmax)
-		xipInRange := uint64(hi - lo)
+		xipInRange := uint64(hi - lo) //nolint:gosec  // we already know that other.xmax is greater than s.xmax, and we know that xipList is sorted in ascending order.
 		if xipInRange < rangeSize {
 			return true
 		}


internal/datastore/postgres/snapshot.go
```diff
@@ -186,11 +186,32 @@ func (s pgSnapshot) compare(rhs pgSnapshot) comparisonResult {
...
 }
 
-func (s pgSnapshot) anyTXVisible(first uint64, others []uint64) bool {
-	if s.txVisible(first) {
+// otherHasMoreInfo returns true if other knows the disposition of any
+// transaction that s does not. s doesn't know about its own xipList entries
+// (in-progress) or any txid >= s.xmax (unseen). If other can see any of
+// those as committed, then other has strictly more information.
+func (s pgSnapshot) otherHasMoreInfo(other pgSnapshot) bool {
+	// Check if other sees any of s's in-progress transactions as committed
```
josephschorr: can you quote 'other' and 's'; its kind of hard to follow with just the words

internal/datastore/postgres/snapshot.go
```diff
@@ -171,8 +171,8 @@ func (cr comparisonResult) String() string {
 // 0:4:2   -> (1,3 visible)
 // 0:4:2,3 -> (1 visible)
 func (s pgSnapshot) compare(rhs pgSnapshot) comparisonResult {
-	rhsHasMoreInfo := rhs.anyTXVisible(s.xmax, s.xipList)
-	lhsHasMoreInfo := s.anyTXVisible(rhs.xmax, rhs.xipList)
+	rhsHasMoreInfo := s.otherHasMoreInfo(rhs)
```
josephschorr, line 174: Given the complexity of this, it might be worth having a generated long description, with some additional examples

internal/datastore/postgres/snapshot_test.go
tstirrat15, line 1: It doesn't have to be in the context of this PR, but this does seem like something that would benefit from property-based testing if we can express some general constraints about the relationships between revisions (equivalence, partial ordering, total ordering, etc)

internal/datastore/postgres/snapshot.go
miparnisari: i agree with the description (i had to use a visualization to grasp the idea: https://habrastorage.org/r/w1560/webt/5z/dx/cf/5zdxcfxgsqzukzjqgyetvvahzru.png)

but i disagree with writing examples. what better examples than _the unit tests_?

internal/datastore/postgres/snapshot.go
```diff
@@ -186,11 +186,32 @@ func (s pgSnapshot) compare(rhs pgSnapshot) comparisonResult {
...
+	// If the range contains more txids than other has in-progress in that
+	// range, at least one must be committed.
+	if other.xmax > s.xmax {
+		rangeSize := other.xmax - s.xmax
+		lo, _ := slices.BinarySearch(other.xipList, s.xmax)
+		hi, _ := slices.BinarySearch(other.xipList, other.xmax)
+		xipInRange := uint64(hi - lo)
+		if xipInRange < rangeSize {
+			return true
+		}
```
tstirrat15, line 225: If I've got this right, this is essentially masking `[s.xmax, s.xmax + 1, ..., other.xmax - 1]` against `other.xipList` and asking whether there's anyhing left, right? Is there a more direct way to express that that retains the `O(log(n))` runtime? Is performance key here, or could we use an approach that's easier to read but slightly slower, such as constructing a set of `[s.xmax, other.xmax)` and then subtracting the set of `other.xip` and asking whether the resulting set is nonempty?

internal/datastore/postgres/snapshot_test.go
```diff
@@ -101,7 +101,126 @@ func TestCompare(t *testing.T) {
...
+
+		// Both have different in-progress sets that overlap in complex ways
+		{snap(5, 12, 5, 8, 10), snap(5, 12, 6, 9, 11), concurrent},
+		{snap(5, 12, 6, 9, 11), snap(5, 12, 5, 8, 10), concurrent},
+
+		// Wide gap: LHS committed low txs, RHS committed high txs
+		{snap(50, 50), snap(40, 60, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49), concurrent},
+		{snap(40, 60, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49), snap(50, 50), concurrent},
+
+		// === Large xid gaps: must not iterate the range ===
```
tstirrat15, line 182: I guess this would be why not to use something `O(n)`, but how much do they actually differ in practice?

internal/datastore/postgres/snapshot_test.go
```diff
@@ -101,7 +101,126 @@ func TestCompare(t *testing.T) {
...
+		{snap(5, 10, 5, 6, 7), snap(5, 10, 5, 6, 7, 8, 9), gt},
+
+		// One sees everything the other does plus a higher xmax range
+		{snap(0, 5), snap(0, 10), lt},
+		{snap(0, 10), snap(0, 5), gt},
+
+		// === Concurrent: each knows something the other doesn't ===
+		// Original bug case: LHS knows tx 100 done, RHS knows tx 102 done
+		{snap(101, 101), snap(100, 104, 100, 101), concurrent},
+		{snap(100, 104, 100, 101), snap(101, 101), concurrent},
```
tstirrat15, line 156: I take it this is the test that would have failed under the old code?

internal/datastore/postgres/snapshot_test.go
jakedt: Probably not very much, but I don't think the extra clarity is worth the risk. I had an earlier implementation which iterated the range, but this one is more efficient in every way.

miparnisari (approved):

PR #2965, opened 2026-03-10
feat: update DevContext and LSP to support composable schemas

tstirrat15:
## Description

 Filesystem-aware development system — The DevContext and schema compilation now accept a filesystem  
  (fs.FS), enabling validation of schemas that use import syntax (composable schemas).               
                                                                                                       
  Key changes:                                                                                           
   
  1. `pkg/development/` — devcontext.go and schema.go updated to thread an fs.FS through schema          
  compilation. The position mapper (`schema_position_mapper.go`) was reworked significantly to
  handle multi-file schemas correctly, with new tests.                                                 
  2. `pkg/schemadsl/compiler/` — The importer and position mapper now work with filesystem-backed      
  sources. development.go gained helpers for filesystem-aware compilation. The translator     
  tracks source file information more precisely.
  3. `LSP (internal/lsp/)` — New overlay filesystem (`overlay.go`) lets the LSP validate open/unsaved files
   by layering in-memory content over the real filesystem. Handlers updated to support                 
  `textDocument/definition` (go-to-definition) and fix `requestDiagnostics`. Substantial new tests.                                                                                              

## Testing
you can test this by bringing it over to VSCode and trying it out with schemas of your own. you have to

1. On spicedb repo:  `gh pr checkout 2965 && mage build:binary `
2. On vscode repo : `gh pr checkout 51`
3. edit line 95 on `package.json` to point to the binary you created in step 1
4. hit F5 to run the extension

if you want to test in zed:

1. On zed repo: `gh pr checkout 635 && mage build:binary`
2. you can now do `./zed validate ....`


internal/lsp/handlers.go
```diff
@@ -282,6 +283,63 @@ func (s *Server) textDocHover(_ context.Context, r *jsonrpc2.Request) (*Hover, e
 	return hoverContents, nil
 }
 
+func (s *Server) textDocDefinition(_ context.Context, r *jsonrpc2.Request) (*lsp.Location, error) {
```
miparnisari, line 300: FYI this is what gets called in Vs Code when you click F12/Go to Definition

internal/lsp/handlers.go
```diff
@@ -57,16 +54,19 @@ func (s *Server) computeDiagnostics(ctx context.Context, uri lsp.DocumentURI) ([
...
 			Relationships: nil,
-		})
+		}, development.WithSourceFS(overlayFS), development.WithRootFileName(string(uri)))
 		if err != nil {
 			return err
 		}
-
 		// Get errors.
 		for _, devErr := range devErrs.GetInputErrors() {
+			if !slices.Contains(devErr.Path, string(uri)) {
```
miparnisari: FYI this code gets called in VSCode when you save a file.

`devErrs` contains errors from ALL files. However, `computeDiagnostics` is only computing errors for `uri`. So, i'm skipping here.

internal/lsp/handlers.go
```diff
@@ -197,12 +197,13 @@ func (s *Server) getCompiledContents(path lsp.DocumentURI, files *persistent.Map
...
 
-	justCompiled, derr, err := development.CompileSchema(file.contents)
+	overlayFS := newLSPOverlayFS(uriToSourceDir(path), files)
+	justCompiled, derr, err := development.CompileSchema(file.contents, development.WithSourceFS(overlayFS))
 	if err != nil {
 		return nil, err
 	}
 	if derr != nil {
-		return nil, &jsonrpc2.Error{Code: jsonrpc2.CodeInternalError, Message: derr.String()}
+		return nil, errors.New(derr.String())
```
miparnisari: FYI otherwise this get shown in the error console but not to the user

pkg/caveats/parameters.go
```diff
@@ -66,6 +66,9 @@ func ConvertContextToParameters(
 
 // ParameterTypeString returns the string form of the type reference.
 func ParameterTypeString(typeRef *core.CaveatTypeReference) string {
+	if typeRef == nil {
```
miparnisari, line 69: FYI prevents a panic in line 73. Can move to different PR if so desired

pkg/development/devcontext.go
```diff
@@ -318,25 +318,28 @@ func loadCompiled(
...
+				}
+
 				errors = append(errors, &devinterface.DeveloperError{
 					Message: terr.Error(),
 					Kind:    devinterface.DeveloperError_SCHEMA_ISSUE,
 					Source:  devinterface.DeveloperError_SCHEMA,
 					Context: errWithSource.SourceCodeString,
 					Line:    lineNumber,
 					Column:  columnPosition,
+					Path:    []string{path},
```
miparnisari: FYI `zed` uses this to show the correct content from the file that has the error

pkg/development/warnings.go
```diff
@@ -61,6 +61,9 @@ func warningForPosition(warningName string, message string, sourceCode string, s
 
 // GetWarnings returns a list of warnings for the given developer context.
 func GetWarnings(ctx context.Context, devCtx *DevContext) ([]*devinterface.DeveloperWarning, error) {
+	if devCtx == nil {
```
miparnisari, line 65: FYI prevents panic on line 68.

pkg/development/schema_position_mapper.go
```diff
@@ -106,6 +142,8 @@ func (r *SchemaPositionMapper) ReferenceAtPosition(source input.Source, position
 			return nil, err
 		}
 
+		targetSource := r.resolveTargetSource(relation.Name, source)
```
miparnisari, line 145: FYI this is what VSCode uses to know which file contains the error and display the right text content

pkg/schemadsl/compiler/development.go
```diff
@@ -55,6 +55,69 @@ func (nc *NodeChain) String() string {
 	return out.String()
 }
 
+// DefinitionNodeSource returns the input source for the AST node defining the given name.
+// For definitions compiled from imports, this will be the imported file path (e.g. "users.zed").
+// For definitions in the root schema, this will be the root source (e.g. "schema").
```
miparnisari: If there is a better way to do this please let me know

internal/lsp/lspdefs.go
```diff
@@ -53,6 +54,10 @@ type InitializeParams struct {
 }
 
 type ClientCapabilities struct {
+	Workspace WorkspaceClientCapabilities `json:"workspace"`
```
miparnisari: FYI bug fix, IIRC for this https://github.com/authzed/spicedb-vscode/issues/41

pkg/schemadsl/compiler/importer.go
```diff
@@ -7,40 +7,42 @@ import (
...
-}
-
-func importFile(fsys fs.FS, filePath string) (*dslNode, error) {
+func importFile(fsys fs.FS, filePath string, toplevel *dslNode, mapper input.PositionMapper) (*dslNode, string, error) {
 	schemaBytes, err := fs.ReadFile(fsys, filePath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read import in schema file %q: %w", filePath, err)
+		nodeSource, _ := toplevel.GetString(dslshape.NodePredicateSource)
+		errMsg := fmt.Errorf("failed to read import %q: %w", filePath, err)
+		return nil, "", toContextError(errMsg.Error(), nodeSource, toplevel, mapper)
```
miparnisari, line 22: FYI changing to `contextError` so that we don't lose line number info of where the error happened

pkg/schemadsl/generator/generator.go
```diff
@@ -82,13 +82,7 @@ func GenerateSchemaWithCaveatTypeSet(ctx context.Context, definitions []compiler
 
 // GenerateCaveatSource generates a DSL view of the given caveat definition.
 func GenerateCaveatSource(caveat *core.CaveatDefinition, caveatTypeSet *caveattypes.TypeSet) (string, bool, error) {
-	generator := &sourceGenerator{
-		indentationLevel: 0,
-		hasNewline:       true,
-		hasBlankline:     true,
-		hasNewScope:      true,
-		caveatTypeSet:    caveatTypeSet,
-	}
```
miparnisari: FYI this fixes a panic here. Notice how `flags` isn't being set.

internal/lsp/handlers.go
josephschorr: If I recall correctly, I had a map somewhere from URI <-> Path; if not, we likely should add it or helper functions, to make it clear when using one vs the other

internal/lsp/handlers.go
josephschorr: Isn't that because we marked it as `CodeInternalError`?

internal/lsp/handlers.go
josephschorr: Correct

internal/lsp/handlers.go
```diff
@@ -282,6 +287,63 @@ func (s *Server) textDocHover(_ context.Context, r *jsonrpc2.Request) (*Hover, e
...
+
+		if resolved == nil || resolved.TargetPosition == nil {
+			return nil
+		}
+
+		// Determine the target file URI from TargetSource.
+		targetURI := params.TextDocument.URI
+		if resolved.TargetSource != nil && *resolved.TargetSource != "schema" {
+			sourceDir := uriToSourceDir(params.TextDocument.URI)
+			targetURI = lsp.DocumentURI("file://" + filepath.Join(sourceDir, string(*resolved.TargetSource)))
```
josephschorr: And you'd use the new map/mapping functions here

internal/lsp/handlers.go
```diff
@@ -282,6 +287,63 @@ func (s *Server) textDocHover(_ context.Context, r *jsonrpc2.Request) (*Hover, e
...
+		}
+
+		nameStart := resolved.TargetPosition.ColumnPosition + resolved.TargetNamePositionOffset
+		location = &lsp.Location{
+			URI: targetURI,
+			Range: lsp.Range{
+				Start: lsp.Position{Line: resolved.TargetPosition.LineNumber, Character: nameStart},
```
josephschorr: Triple check this: I forget which is 0-indexed and which is 1-indexed, so best to triple check (and maybe even add a test over a conversion function)

internal/lsp/overlay.go
```diff
@@ -0,0 +1,76 @@
...
+	return &lspOverlayFS{
+		base:      os.DirFS(sourceDir),
+		sourceDir: sourceDir,
+		files:     files,
+	}
+}
+
+func (o *lspOverlayFS) Open(name string) (fs.File, error) {
+	absPath := filepath.Join(o.sourceDir, filepath.FromSlash(name))
+	uri := lsp.DocumentURI("file://" + absPath)
```
josephschorr, line 43: we need to be VERY careful here... we need should verify the name doesn't contain `..` to prevent searching outside of the working directory (for security); I suspect filepath pkg handles this, but let's add a test

pkg/caveats/parameters.go
josephschorr: When would this be `nil`?

pkg/development/devcontext.go
```diff
@@ -318,25 +318,28 @@ func loadCompiled(
...
-			lineNumber, err := safecast.Convert[uint32](errWithSource.LineNumber)
-			if err != nil {
-				log.Err(err).Msg("could not cast lineNumber to uint32")
-			}
-			columnPosition, err := safecast.Convert[uint32](errWithSource.ColumnPosition)
-			if err != nil {
-				log.Err(err).Msg("could not cast columnPosition to uint32")
-			}
 			if ok {
+				// NOTE: zeroes are fine here to mean "unknown"
```
josephschorr: Add a comment if these are 1-indexed (which I believe they are)

pkg/development/schema_position_mapper_test.go
```diff
@@ -244,7 +245,7 @@ func TestSchemaPositionMapper(t *testing.T) {
 			}
 			`,
 			line:   3,
-			column: 6,
+			column: 6, // TODO if you put 23, the mapper doesn't return anything
```
josephschorr, line 248: ???

pkg/composableschemadsl/compiler/translator.go
```diff
@@ -826,7 +826,7 @@ func translateImports(itctx importResolutionContext, root *dslNode) error {
 			// This is a new node provided by the translateImport
 			parsedImportRoot, err := importFile(itctx.sourceFS, filePath)
 			if err != nil {
-				return toContextError("failed to read import in schema file", "", topLevelNode, itctx.mapper)
+				return toContextError(err.Error(), "", topLevelNode, itctx.mapper)
```
miparnisari: i need to remove this change; the `composableschemadsl` compiler isn't used anymore :)

internal/lsp/handlers.go
miparnisari: i undid this change.

pkg/development/schema_position_mapper_test.go
miparnisari: moved to https://github.com/authzed/spicedb/pull/2972

internal/lsp/handlers.go
miparnisari: i'm not sure what map you are referring to, but i don't see why a map makes sense, can you explain or provide a code snippet with what you mean?

internal/lsp/overlay.go
```diff
@@ -74,3 +83,9 @@ func uriToSourceDir(uri lsp.DocumentURI) string {
 	path := strings.TrimPrefix(string(uri), "file://")
 	return filepath.Dir(path)
 }
+
+// resolveURI resolves a relative path against the directory of baseURI,
+// returning a new file:// DocumentURI.
+func resolveURI(baseURI lsp.DocumentURI, relativePath string) lsp.DocumentURI {
+	return lsp.DocumentURI("file://" + filepath.Join(uriToSourceDir(baseURI), relativePath))
```
tstirrat15, line 90: Could use `net/url` here by replacing `Scheme`: https://stackoverflow.com/questions/63197536/replacing-protocol-and-hostname-in-url-in-go

pkg/development/schema.go
```diff
@@ -25,6 +26,12 @@ func WithSourceFS(fsys fs.FS) CompileOption {
 	return func(cfg *compileConfig) { cfg.fsys = fsys }
 }
 
+// WithRootFileName saves the root file of the schema
+// so errors can be displayed accurately on UIs.
+func WithRootFileName(name string) CompileOption {
+	return func(cfg *compileConfig) { cfg.rootFileName = name }
+}
```
tstirrat15, line 33: Barak and I were talking about this - would it make sense to hang the filename off of the definition such that error handling can look up the filename along with the position straight off the node?

pkg/development/schema_position_mapper.go
```diff
@@ -248,14 +288,24 @@ func (r *SchemaPositionMapper) ReferenceAtPosition(source input.Source, position
...
 
+// resolveTargetSource returns the input.Source for the file where a definition or caveat
+// is defined. For imported definitions, this will be the imported file path. For definitions
+// in the root schema, this returns the fallback source.
+func (r *SchemaPositionMapper) resolveTargetSource(name string, fallback input.Source) input.Source {
+	if defSource := r.schema.GetPathToDefinitionOrPartialOrCaveat(name); defSource != "" {
+		return input.Source(defSource)
+	}
+	return fallback
+}
```
tstirrat15, line 307: Yeah, I think that would clean up this code a bit

pkg/development/schema_position_mapper_test.go
```diff
@@ -465,3 +492,113 @@ definition document {
...
+	relation somerelation: user with is_raining
+	relation oops: group#member
+	...secret
+	permission usessecret = secret
+}
+`
+	sourceFS := fstest.MapFS{
+		"path/partials.zed": &fstest.MapFile{Data: []byte("use partial\npartial secret {\nrelation secret: user\n}")},
+		"path/users.zed":    &fstest.MapFile{Data: []byte("definition user {}\ncaveat is_raining(day string) {\nday == \"saturday\"\n}")},
+	}
```
tstirrat15, line 513: Oooh I like this

pkg/schemadsl/compiler/development.go
```diff
@@ -55,6 +55,69 @@ func (nc *NodeChain) String() string {
 	return out.String()
 }
 
+// GetPathToDefinitionOrPartialOrCaveat returns the input source for the AST node defining the given name.
+// For definitions compiled from imports, this will be the imported file path (e.g. "users.zed").
+// For definitions in the root schema, this will be the root source (e.g. "schema").
+// Returns empty string if not found.
+func (cs *CompiledSchema) GetPathToDefinitionOrPartialOrCaveat(name string) string {
```
tstirrat15, line 62: Yeah, I think we should annotate the node rather than treating it as a calculated value. I can make that change today.

pkg/development/schema.go
tstirrat15: I'll do this in a follow-on PR

pkg/development/schema_position_mapper.go
tstirrat15: TODO for later

pkg/schemadsl/compiler/development.go
tstirrat15: Deferring to future refactor

pkg/schemadsl/compiler/development.go
```diff
@@ -85,11 +148,18 @@ func PositionToAstNodeChain(schema *CompiledSchema, source input.Source, positio
...
 	if !node.Has(dslshape.NodePredicateStartRune) {
 		return nil, nil
 	}
 
+	// Skip nodes from imported files whose rune positions may overlap with the root file.
+	if nodeSource, err := node.GetString(dslshape.NodePredicateSource); err == nil {
+		if nodeSource != rootSource {
+			return nil, nil
+		}
+	}
```
tstirrat15, line 161: Why is this part necessary?

pkg/schemadsl/compiler/importer.go
```diff
@@ -7,43 +7,45 @@ import (
...
-		SchemaString: string(schemaBytes),
+		SchemaString: content,
 	})
-	return parsedSchema, err
+	return parsedSchema, content, err
 }
 
 // Take a filepath and ensure that it's local to the current context.
-func validateFilepath(path string) error {
+func validateFilepath(path string, toplevel *dslNode, mapper input.PositionMapper) error {
```
tstirrat15, line 35: I think we might be able to delegate this to `DirFS`'s restrictions. I guess we'd still need to rewrite the error in that case, though.

pkg/schemadsl/compiler/development.go
miparnisari: Makes this test pass: "cursor on the import path" on `schema_position_mapper_test.go`

PR #2968, opened 2026-03-11
feat(qp): prune branches that cannot lead to the subject type of the …

miparnisari:
…request

## Description

This is an optimization for Check API when using the query planner. We prune branches of the tree that can never lead to the subject type in the request.

## Testing

Unit tests

### DIFF
--- a/internal/services/v1/permissions_queryplan.go
+++ b/internal/services/v1/permissions_queryplan.go
@@ -58,12 +58,13 @@ func (ps *permissionServer) checkPermissionWithQueryPlan(ctx context.Context, re
 		return nil, ps.rewriteError(ctx, err)
 	}
 
-	optimized, err := queryopt.ApplyOptimizations(co, queryopt.StandardOptimzations)
+	optimized, err := queryopt.ApplyOptimizations(co, queryopt.StandardOptimzations, queryopt.RequestParams{
+		SubjectType:     req.Subject.Object.ObjectType,
+		SubjectRelation: req.Subject.OptionalRelation,
+	})
 	if err != nil {
 		return nil, ps.rewriteError(ctx, err)
 	}
-
-	optimized = queryopt.ApplyReachabilityPruning(optimized, req.Subject.Object.ObjectType)
 
 	it, err := optimized.Compile()
 	if err != nil {
--- a/pkg/query/queryopt/reachability_pruning.go
+++ b/pkg/query/queryopt/reachability_pruning.go
@@ -1,129 +1,91 @@
 package queryopt
 
 import (
-	"slices"
-
 	"github.com/authzed/spicedb/pkg/query"
 )
 
-// ApplyReachabilityPruning applies subject-type reachability pruning to a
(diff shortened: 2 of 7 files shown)


internal/services/v1/permissions_queryplan.go
```diff
@@ -63,6 +63,8 @@ func (ps *permissionServer) checkPermissionWithQueryPlan(ctx context.Context, re
 		return nil, ps.rewriteError(ctx, err)
 	}
 
+	optimized = queryopt.ApplyReachabilityPruning(optimized, req.Subject.Object.ObjectType)
```
barakmich: We should make this a little more generalized (ie, this is the first request-parameterized optimization!) but this is fine for now.

pkg/query/queryopt/reachability_pruning.go
```diff
@@ -0,0 +1,129 @@
...
+// For exclusions (e.g. editor - writer), the pruning decision is based on whether the
+// main set on the left can reach the target subject type. The right set is ignored.
+//
+// For intersections (e.g. a & b & c), the pruning decision is based on whether
+// all of the branches can reach the target subject type. If at least one cannot, the
+// entire intersection is elided.
+func ApplyReachabilityPruning(co query.CanonicalOutline, targetSubjectType string) query.CanonicalOutline {
+	pruned, _ := pruneOutline(co.Root, targetSubjectType)
+
+	// TODO do i need to call FillMissingIds?
```
barakmich: Not if it's just pruning 🙂

pkg/query/queryopt/registry.go
```diff
@@ -46,7 +62,7 @@ type Optimizer struct {
 // outline via query.MutateOutline. Nodes that survive mutation unchanged keep
 // their existing IDs. Newly synthesized nodes (ID==0) receive fresh IDs via
 // query.FillMissingNodeIDs. The returned CanonicalOutline is ready to compile.
-func ApplyOptimizations(co query.CanonicalOutline, names []string) (query.CanonicalOutline, error) {
+func ApplyOptimizations(co query.CanonicalOutline, names []string, requestParams RequestParams) (query.CanonicalOutline, error) {
```
barakmich, line 65: I dig it 🙂

pkg/query/queryopt/registry.go
```diff
@@ -29,13 +29,29 @@ func GetOptimization(name string) (Optimizer, error) {
...
+
 // Optimizer describes a single named outline optimization.
 type Optimizer struct {
 	Name        string
 	Description string
-	Mutation    query.OutlineMutation
+	// NewTransform creates a whole-tree transformation for this optimizer,
+	// optionally using request-specific parameters. Each transform typically
+	// calls query.MutateOutline internally with its own mutations.
+	NewTransform func(RequestParams) OutlineTransform
```
barakmich, line 54: I would do one little trick here, which is to name this signature (eg, RequestOutlineTransform, under the above) and have a simple identity helper that wraps it (turning the caveat pushdown into something like `NewTransform: withoutRequestParams(caveatPushdown)`)

PR #2970, opened 2026-03-11
refactor: create new `queryopt` package to target query plan optimizations

barakmich:
## Description

Before, optimize.go lived in `pkg/query` and worked on iterator trees. With the introduction of Outlines, it's much more convenient to work on Outlines before compiling an iterator tree.

This creates the new package framework, registry, and ports the existing optimize files, deleting the old ones.

## Testing

Testing ported to new package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Added
+- New `pkg/query/queryopt` package for building optimizations into the query plan
+
 ### Changed
 - Updated CI so that Postgres tests run against v18 which is GA and not against v13 which is EOL (https://github.com/authzed/spicedb/pull/2926)
 - Added tracing to request validation (https://github.com/authzed/spicedb/pull/2950)


pkg/query/queryopt/caveat_pushdown.go
```diff
@@ -0,0 +1,121 @@
...
+// relocated instances of the same caveat carry the original ID.
+func caveatPushdownInner(outline query.Outline, originalID query.OutlineNodeID) query.Outline {
+	// A CaveatIterator must have exactly one child.
+	if len(outline.SubOutlines) != 1 {
+		return outline
+	}
+	child := outline.SubOutlines[0]
+
+	// Do not push through IntersectionArrow (all() semantics require post-intersection caveat eval).
+	if child.Type == query.IntersectionArrowIteratorType {
```
josephschorr, line 55: As an improvement for the future, it might be worth add an `.Is(...)` onto the child that takes in one or more types, and returns true if any match

pkg/query/queryopt/caveat_pushdown.go
```diff
@@ -0,0 +1,121 @@
...
+	// same ID as the caveat node being relocated.
+	newSubs := make([]query.Outline, len(child.SubOutlines))
+	changed := false
+	for i, sub := range child.SubOutlines {
+		if outlineContainsCaveat(sub, caveat.CaveatName) {
+			wrapped := query.Outline{
+				Type:        query.CaveatIteratorType,
+				Args:        outline.Args,
+				SubOutlines: []query.Outline{sub},
+				ID:          originalID,
```
josephschorr, line 83: is the ID the thing that is carrying the caveat+local context?

PR #2975, opened 2026-03-19
refactor: make all the signatures with PathSeq go from`Path` to `*Path`

barakmich:
## Description

We can reduce memory usage and minorly improve performance by doing fewer copies. Since we're streaming these objects with `iter.Seq2[Path, error]` every time we iterate we instantiate a new copy. This allows us to use pointers instead, eliding the copy.

### DIFF
--- a/pkg/query/arrow.go
+++ b/pkg/query/arrow.go
@@ -136,7 +136,7 @@ func (a *ArrowIterator) checkRightToLeft(ctx *Context, resources []Object, subje
 	return func(yield func(*Path, error) bool) {
 		if ctx.shouldTrace() {
 			ctx.TraceStep(a, "arrow check (right-to-left) with %d resources for subject %s:%s",
-			len(resources), subject.ObjectType, subject.ObjectID)
+				len(resources), subject.ObjectType, subject.ObjectID)
 		}
 
 		// Strategy: Start from the right side with the target subject
@@ -181,13 +181,13 @@ func (a *ArrowIterator) checkRightToLeft(ctx *Context, resources []Object, subje
 
 			if ctx.shouldTrace() {
 				ctx.TraceStep(a, "left side returned %d paths for intermediate %s:%s",
-				count, intermediateAsSubject.ObjectType, intermediateAsSubject.ObjectID)
+					count, intermediateAsSubject.ObjectType, intermediateAsSubject.ObjectID)
 			}
 		}
 
 		if ctx.shouldTrace() {
 			ctx.TraceStep(a, "arrow check (right-to-left) completed: %d right paths, %d total result paths",
-			rightPathCount, totalResultPaths)
+				rightPathCount, totalResultPaths)
 		}
 	}, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/query/intersection.go
```diff
@@ -152,45 +163,42 @@ func (i *IntersectionIterator) IterSubjectsImpl(ctx *Context, resource Object, f
...
+			newPathsByKey := make(map[string]*Path)
 
 			// First collect all paths from this iterator by subject
-			currentIterPaths := make(map[string]Path)
+			currentIterPaths := make(map[string]*Path)
 			for _, path := range paths {
 				key := ObjectAndRelationKey(path.Subject)
 				if existing, exists := currentIterPaths[key]; !exists {
-					currentIterPaths[key] = path
+					pathCopy := *path
```
miparnisari, line 184: 🤔 this isn't a full copy? `Path` has fields that are pointers

pkg/query/arrow.go
```diff
@@ -71,33 +71,39 @@ func (a *ArrowIterator) CheckImpl(ctx *Context, resources []Object, subject Obje
 // checkLeftToRight implements the left-to-right strategy:
 // For each resource, IterSubjects on left, then Check on right
 func (a *ArrowIterator) checkLeftToRight(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
-	return func(yield func(Path, error) bool) {
-		ctx.TraceStep(a, "processing %d resources", len(resources))
+	return func(yield func(*Path, error) bool) {
+		if ctx.shouldTrace() {
+			ctx.TraceStep(a, "processing %d resources", len(resources))
+		}
```
tstirrat15, line 77: Is this about reducing string allocations? I'm wondering if there's a nicer way to do this than the conditional. I wish thunks in golang were more ergonomic to write.

pkg/query/context.go
```diff
@@ -251,7 +254,10 @@ func (ctx *Context) IterSubjects(it Iterator, resource Object, filterSubjectType
 		isTopLevel = true
 	})
 
-	tracedIterator := ctx.traceEnterIfEnabled(it, iterSubjectsTraceString(resource, filterSubjectType))
+	var tracedIterator Iterator
+	if ctx.shouldTrace() {
+		tracedIterator = ctx.traceEnterIfEnabled(it, iterSubjectsTraceString(resource, filterSubjectType))
```
tstirrat15, line 259: Are these necessary if we're also doing these checks within the iterators? Or is it again about the string allocs here?

pkg/query/intersection.go
```diff
@@ -52,45 +58,42 @@ func (i *IntersectionIterator) CheckImpl(ctx *Context, resources []Object, subje
 			for _, path := range paths {
 				key := path.Resource.Key()
 				if existing, exists := pathsByKey[key]; !exists {
-					pathsByKey[key] = path
+					pathCopy := *path
```
tstirrat15, line 61: This kinda surprises me - why does dereffing make a copy? Or is it that you're assigning a concrete value and that's why it happens?

pkg/query/path.go
```diff
@@ -488,22 +479,23 @@ func PathSeqFromSlice(paths []Path) PathSeq {
...
-				yield(Path{}, err)
+				yield(nil, err)
 				return
 			}
 
 			key := path.EndpointsKey()
 			if existing, exists := seen[key]; !exists {
-				seen[key] = path
+				pathCopy := *path
+				seen[key] = &pathCopy
```
tstirrat15, line 493: Same q here about shallow vs deep copy

pkg/query/recursive.go
```diff
@@ -225,133 +239,126 @@ func (r *RecursiveIterator) breadthFirstIterSubjects(ctx *Context, resource Obje
...
-			// Enable collection mode for all matching sentinels
+			// Enable collection mode for all matching sentinels.
 			for _, sentinelID := range sentinelIDs {
 				ctx.EnableFrontierCollection(sentinelID)
 			}
 
-			// Track paths collected at this ply (keyed by endpoints for deduplication)
-			plyPaths := make(map[string]Path)
+			// Clear the ply map and reuse its backing storage.
+			clear(plyPaths)
```
tstirrat15, line 284: TIL about `clear`

PR #2977, opened 2026-03-20
Fix duplicate diagnostics in LSP server

ivanauth:
Fixes authzed/spicedb-vscode#41

The LSP server's `ClientCapabilities` struct mapped diagnostics at a non-standard JSON path, so pull diagnostics support was never detected. The server always pushed `publishDiagnostics` even when the client was already pulling via `textDocument/diagnostic`, causing duplicate errors.

Fixed by restructuring `ClientCapabilities` to match the LSP spec and detecting pull support via `textDocument.diagnostic` presence.

**Tested manually** — confirmed duplicates in VS Code before the fix and single errors after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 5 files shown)


miparnisari (approved):
Tested locally, works great!

image

 I will test this again with https://github.com/authzed/spicedb/pull/2965

PR #2979, opened 2026-03-20
fix(postgres): some requests may silently swallow errors of sort "revision not found in replica"

miparnisari:
## Description

Fixing this bug mentioned in https://github.com/authzed/spicedb/issues/2525#issuecomment-4094529592

> **Bug 2: Strict read assertion error is silently lost**
> The WHERE clause filters out all rows → 0 results (valid namespace data is dropped)
> The DO block's ASSERT fails → produces an error
> QuerierFuncs.QueryFunc in pgx.go calls rows.Err() before the deferred rows.Close() has drained the second statement's result set, so the ASSERT error is never propagated

```
6:01PM WRN error calling script step error="rpc error: code = Unknown desc = object definition `auth/organisation` not found" consistency=FullyConsistent op=CheckPermission script="spam user#view checks" step=1 worker=89
```

## Testing

[script.yaml](https://github.com/user-attachments/files/25392706/script.yaml)

1. `docker-compose -f docker-compose.postgres.yaml up --build` (with REPLICA_APPLY_DELAY=25ms)
2. `thumper migrate --insecure --no-verify-ca --endpoint localhost:50051 --token foobar ./script.yaml && thumper run --insecure --no-verify-ca --qps 100 --endpoint localhost:50051 --token foobar ./script.yaml`

wait 1 hour; you should see no errors

### DIFF
--- a/internal/datastore/crdb/pool/pool.go
+++ b/internal/datastore/crdb/pool/pool.go
@@ -177,8 +177,7 @@ func (p *RetryPool) MinConns() uint32 {
 // connection on error, or retrying on a retryable error.
 func (p *RetryPool) ExecFunc(ctx context.Context, tagFunc func(ctx context.Context, tag pgconn.CommandTag, err error) error, sql string, arguments ...any) error {
 	return p.withRetries(ctx, 0, func(conn *pgxpool.Conn) error {
-		tag, err := conn.Conn().Exec(ctx, sql, arguments...)
-		return tagFunc(ctx, tag, err)
+		return common.QuerierFuncsFor(conn.Conn()).ExecFunc(ctx, tagFunc, sql, arguments...)
 	})
 }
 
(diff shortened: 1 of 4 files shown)


internal/datastore/crdb/pool/pool.go
```diff
@@ -177,33 +177,23 @@ func (p *RetryPool) MinConns() uint32 {
...
-		if err != nil {
-			return err
-		}
-		defer rows.Close()
-		err = rowsFunc(ctx, rows)
-		if err != nil {
-			return err
-		}
-		return rows.Err()
+		return common.QuerierFuncsFor(conn.Conn()).QueryFunc(ctx, rowsFunc, sql, optionsAndArgs...)
```
miparnisari, line 188: FYI this is a refactor; the code looked the same and i didn't want to implement the fix in two places. So i'm reusing the code here.

What _is_ weird is that **CRDB** is reusing code that is located at "internal/datastore/**postgres**/common/pgx.go".

josephschorr: CRDB uses the common PG stuff because it supports the PG protocol; that's why its in the `common` package, so it can be shared with both

miparnisari: so, was it an oversight to not reuse the code from `common` in the first place? and this refactor is safe?

PR #2984, opened 2026-03-23
fix(memory): use global protovalidate instance and bump version

miparnisari:
## Description

This is to fix a memory regression with `protovalidate`. Memory consumption isn't back to pre-`protovalidate`, but it's very close

image

## References

https://github.com/bufbuild/protovalidate-go/issues/286

### DIFF
--- a/internal/dispatch/keys/hasher_ristretto.go
+++ b/internal/dispatch/keys/hasher_ristretto.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"unsafe"
 
-	"github.com/ccoveille/go-safecast/v2"
 	"github.com/cespare/xxhash/v2"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/genutil/protovalidate.go
```diff
@@ -1,18 +1,11 @@
...
 // if the validator can't be constructed.
-func MustNewProtoValidator(opts ...protovalidate.ValidatorOption) protovalidate.Validator {
-	validator, err := protovalidate.New(opts...)
-	if err != nil {
-		wrappedErr := fmt.Errorf("could not construct validator: %w", err)
-		panic(wrappedErr)
-	}
-	return validator
+func MustNewProtoValidator() protovalidate.Validator {
+	return protovalidate.GlobalValidator
```
tstirrat15: This is going to have the same cold-start issues that we were seeing previously. Can we give it options?

PR #2993, opened 2026-03-24
refactor: convert the signature of a query plan `CheckImpl` to return a single path instead of a PathSeq

barakmich:
## Description

This is a performance benefit and retroactive clarification of the interface. Before, the returned PathSeq contained only the potential paths ending with the resource/subject pair being checked or an empty sequence.

However, even for an empty sequence, we still had to build and iterate an iterator. Now, combined with the switch to `*Path` checks are cheaper and use less memory, as well as being more clear (No path, none exists. If there's an attached caveat, then it's conditional, as per other paths)

### DIFF
--- a/pkg/query/alias.go
+++ b/pkg/query/alias.go
@@ -83,19 +83,20 @@ func (a *AliasIterator) CheckImpl(ctx *Context, resource Object, subject ObjectA
 		return nil, err
 	}
 
-	// Rewrite the sub-path relation if present
 	if subPath != nil {
+		// We have a path! Even if it's caveated, rewrite it and return.
 		subPath.Relation = a.relation
+		return subPath, nil
 	}
 
-	// Check for self-edge: if resource with the alias relation matches the subject
+	// We have no sub-path. We need to check for the self edge, if resource with the alias relation matches the subject
 	resourceWithAlias := resource.WithRelation(a.relation)
 	isSelfEdge := resourceWithAlias.ObjectID == subject.ObjectID &&
 		resourceWithAlias.ObjectType == subject.ObjectType &&
 		resourceWithAlias.Relation == subject.Relation
 
 	if !isSelfEdge {
-		return subPath, nil
+		return nil, nil
 	}
 
 	// Build the synthetic self-edge path
@@ -110,16 +111,7 @@ func (a *AliasIterator) CheckImpl(ctx *Context, resource Object, subject ObjectA
 		Metadata: make(map[string]any),
 	}
 
-	if subPath == nil {
-		return selfPath, nil
-	}
-
-	// Both self-edge and sub-path match: OR-merge them
-	merged, err := selfPath.MergeOr(subPath)
-	if err != nil {
-		return nil, err
-	}
-	return merged, nil
+	return selfPath, nil
 }
 
 func (a *AliasIterator) IterSubjectsImpl(ctx *Context, resource Object, filterSubjectType ObjectType) (PathSeq, error) {
@@ -143,7 +135,7 @@ func (a *AliasIterator) IterSubjectsImpl(ctx *Context, resource Object, filterSu
 // a subject anywhere in the datastore (expired or not), and the filter allows it, we
 // include a self-edge in the results.
 func (a *AliasIterator) shouldIncludeSelfEdge(ctx *Context, resource Object, filterSubjectType ObjectType) bool {
-	if ctx.TopLevelOperation != IterSubjects {
+	if ctx.TopLevelOperation != OperationIterSubjects {
 		return false
 	}
 	// First check: does the filter allow this resource type as a subject?
--- a/pkg/query/arrow.go
+++ b/pkg/query/arrow.go
@@ -103,14 +103,13 @@ func (a *ArrowIterator) checkLeftToRight(ctx *Context, resource Object, subject 
 		}
 
 		combined := combineArrowPaths(leftPath, rightPath)
-		if result == nil {
-			result = combined
-		} else {
-			merged, err := result.MergeOr(combined)
-			if err != nil {
-				return nil, err
-			}
-			result = merged
+		if combined.Caveat == nil {
+			return combined, nil
+		}
+
+		result, err = result.MergeOr(combined)
+		if err != nil {
+			return nil, err
 		}
 	}
 
@@ -162,14 +161,13 @@ func (a *ArrowIterator) checkRightToLeft(ctx *Context, resource Object, subject 
 		}
 
 		combined := combineArrowPaths(leftPath, rightPath)
-		if result == nil {
-			result = combined
-		} else {
-			merged, err := result.MergeOr(combined)
-			if err != nil {
-				return nil, err
-			}
-			result = merged
+		if combined.Caveat == nil {
+			return combined, nil
+		}
+
+		result, err = result.MergeOr(combined)
+		if err != nil {
+			return nil, err
 		}
 
 		if ctx.shouldTrace() {
--- a/pkg/query/context.go
+++ b/pkg/query/context.go
@@ -182,8 +182,8 @@ func (ctx *Context) notifyEnterIterator(op Operation, key CanonicalKey) {
 // ObserveReturnIterator is always called afterward.
 func (ctx *Context) notifyCheckResult(key CanonicalKey, path *Path) {
 	for _, obs := range ctx.observers {
-		obs.ObservePath(Check, key, path)
-		obs.ObserveReturnIterator(Check, key)
+		obs.ObservePath(OperationCheck, key, path)
+		obs.ObserveReturnIterator(OperationCheck, key)
 	}
 }
 
@@ -220,7 +220,7 @@ func (ctx *Context) Check(it Iterator, resource Object, subject ObjectAndRelatio
 	}
 
 	if ctx.TopLevelOperation == OperationUnset {
-		ctx.TopLevelOperation = Check
+		ctx.TopLevelOperation = OperationCheck
 	}
 
 	var tracedIterator Iterator
@@ -228,7 +228,7 @@ func (ctx *Context) Check(it Iterator, resource Object, subject ObjectAndRelatio
 		tracedIterator = ctx.traceEnterIfEnabled(it, checkTraceString(resource, subject))
 	}
 	key := it.CanonicalKey()
-	ctx.notifyEnterIterator(Check, key)
+	ctx.notifyEnterIterator(OperationCheck, key)
 
 	path, err := ctx.Executor.Check(ctx, it, resource, subject)
 	if err != nil {
@@ -283,7 +283,7 @@ func (ctx *Context) IterSubjects(it Iterator, resource Object, filterSubjectType
 	ctx.topLevelOnce.Do(func() {
 		ctx.topLevelIterator = it
 		isTopLevel = true
-		ctx.TopLevelOperation = IterSubjects
+		ctx.TopLevelOperation = OperationIterSubjects
 	})
 
 	var tracedIterator Iterator
@@ -291,7 +291,7 @@ func (ctx *Context) IterSubjects(it Iterator, resource Object, filterSubjectType
 		tracedIterator = ctx.traceEnterIfEnabled(it, iterSubjectsTraceString(resource, filterSubjectType))
 	}
 	key := it.CanonicalKey()
-	ctx.notifyEnterIterator(IterSubjects, key)
+	ctx.notifyEnterIterator(OperationIterSubjects, key)
 
 	pathSeq, err := ctx.Executor.IterSubjects(ctx, it, resource, filterSubjectType)
 	if err != nil {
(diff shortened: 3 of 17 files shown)


pkg/query/types.go
```diff
@@ -11,10 +11,10 @@ import (
 // Plan is the external-facing notion of a query plan. These follow the general API for
 // querying anything in the database as well as describing the plan.
 type Plan interface {
-	// CheckImpl tests if, for the underlying set of relationships (which may be a full expression or a basic lookup, depending on the iterator)
-	// any of the `resourceIDs` are connected to `subjectID`.
-	// Returns the sequence of matching paths, if they exist, at most `len(resourceIDs)`.
-	CheckImpl(ctx *Context, resources []Object, subject ObjectAndRelation) (PathSeq, error)
+	// CheckImpl tests if the given resource is connected to subject in the underlying
+	// set of relationships (which may be a full expression or a basic lookup, depending
+	// on the iterator). Returns the matching Path if found, or nil if not found.
```
miparnisari, line 16: sorry i'm confused by this change. there could still be more than one path connecting resource X with subject Y, right? are you arbitrarily picking one of those paths and returning it? (looks like you are looking at the impl)

EDIT: ah, i just had claude explain this to me. The Path returned is really just a permission verdict with a combined caveat expression,  
not a literal graph path. The multiple graph paths that may exist between resource and subject get
folded into that one answer via the appropriate boolean operation on their caveats.

i also think it's worth documenting that this function could return `nil, nil` which can bite us if we forget to check for `nil`. or, we could return a sentinel error `path not found`. we could do this in a follow-up PR if needed.

pkg/query/arrow.go
```diff
@@ -60,136 +60,127 @@ func (a *ArrowIterator) CheckImpl(ctx *Context, resources []Object, subject Obje
...
+		combined := combineArrowPaths(leftPath, rightPath)
+		if result == nil {
+			result = combined
+		} else {
+			merged, err := result.MergeOr(combined)
+			if err != nil {
+				return nil, err
 			}
+			result = merged
 		}
```
miparnisari, line 113: if we find a `result` here and it is not caveated, can we not return it immediately? same question for `checkRightToLeft`

pkg/query/alias.go
```diff
@@ -76,25 +76,50 @@ func (a *AliasIterator) maybePrependSelfEdge(resource Object, subSeq PathSeq, sh
...
+		},
+		Metadata: make(map[string]any),
+	}
+
+	if subPath == nil {
+		return selfPath, nil
+	}
+
+	// Both self-edge and sub-path match: OR-merge them
+	merged, err := selfPath.MergeOr(subPath)
```
tstirrat15: What does this typically mean in practice? We keep the resource + subject and get rid of the relation?

pkg/query/alias.go
```diff
@@ -118,6 +143,9 @@ func (a *AliasIterator) IterSubjectsImpl(ctx *Context, resource Object, filterSu
 // a subject anywhere in the datastore (expired or not), and the filter allows it, we
 // include a self-edge in the results.
 func (a *AliasIterator) shouldIncludeSelfEdge(ctx *Context, resource Object, filterSubjectType ObjectType) bool {
+	if ctx.TopLevelOperation != IterSubjects {
+		return false
+	}
```
tstirrat15: I'm missing context here - why is this the case?

pkg/query/context.go
```diff
@@ -214,28 +212,61 @@ func (ctx *Context) wrapPathSeqWithObservers(op ObserverOperation, key Canonical
...
-func (ctx *Context) Check(it Iterator, resources []Object, subject ObjectAndRelation) (PathSeq, error) {
+// Check tests if the given resource is connected to subject in the underlying set of
+// relationships. Returns the matching Path if found, or nil if not found.
+func (ctx *Context) Check(it Iterator, resource Object, subject ObjectAndRelation) (*Path, error) {
 	if ctx.Executor == nil {
 		return nil, spiceerrors.MustBugf("no executor has been set")
 	}
 
+	if ctx.TopLevelOperation == OperationUnset {
+		ctx.TopLevelOperation = Check
```
tstirrat15: `OperationCheck`?

pkg/query/exclusion_test.go
```diff
@@ -124,38 +105,28 @@ func TestExclusionIterator(t *testing.T) {
...
 		exclusion := NewExclusionIterator(mainSet, excludedSet)
 
-		pathSeq, err := ctx.Check(exclusion, NewObjects("document", "doc1", "doc2", "doc3", "doc4"), NewObject("user", "alice").WithEllipses())
+		// doc1: not excluded → should be found
+		resultA, err := ctx.Check(exclusion, NewObject("document", "doc1"), NewObject("user", "alice").WithEllipses())
 		require.NoError(err)
+		require.NotNil(resultA, "Should contain pathA (doc1)")
 
-		rels, err := CollectAll(pathSeq)
+		// doc3: not excluded → should be found
```
tstirrat15, line 121: This is completely unrelated, but does your OS/editor do the `→` ligature automatically?

pkg/query/exclusion_test.go
```diff
@@ -313,20 +273,10 @@ func TestExclusionErrorHandling(t *testing.T) {
...
-				require.Contains(err.Error(), "faulty iterator collection error")
-				foundError = true
-				break
-			}
-		}
-		require.True(foundError, "Expected error during iteration")
+		// The error from shouldFailOnCollect propagates via CheckImpl
+		_, err := ctx.Check(exclusion, NewObject("document", "doc1"), NewObject("user", "alice").WithEllipses())
+		// Behavior depends on implementation; either error or nil result
+		_ = err
```
tstirrat15: Should this error be asserted against somehow?

I saw another place where you're doing a similar sort of `_` assignment and I can't remember why I thought it made sense in that context.

pkg/query/fixed.go
```diff
@@ -74,30 +74,25 @@ func NewFixedIterator(paths ...Path) *FixedIterator {
...
-			for _, resource := range resources {
-				if f.paths[i].Resource.Equals(resource) &&
-					GetObject(f.paths[i].Subject).Equals(GetObject(subject)) {
-					count++
-					if !yield(&f.paths[i], nil) {
-						return
-					}
-					break
-				}
+	for i := range f.paths {
```
tstirrat15: Why index iteration here rather than `_, path`?

tstirrat15: Is it an alloc thing?

pkg/query/recursive_test.go
```diff
@@ -49,16 +52,124 @@ func TestRecursiveIteratorEmptyBaseCase(t *testing.T) {
...
+	_, err = recursive.CheckImpl(ctx, Object{ObjectType: "folder", ObjectID: "folder1"}, ObjectAndRelation{ObjectType: "user", ObjectID: "tom", Relation: "..."})
+	require.Error(t, err, "Should get error from faulty iterator during execution")
+	// The error message depends on which strategy is used:
+	// - IterSubjects strategy: "IterSubjects failed... at depth" or "execution failed at ply" (from BFS)
+	// - Deepening strategy: "check failed at ply"
+	require.True(t,
+		strings.Contains(err.Error(), "execution failed at ply") ||
+			strings.Contains(err.Error(), "check failed at ply") ||
+			strings.Contains(err.Error(), "at depth"),
+		"Error should be wrapped with ply/depth info, got: %s", err.Error())
```
tstirrat15, line 160: Is the idea that it's up to the executor to determine which strategy is used, so the test is nondeterministic?

pkg/query/recursive_test.go
```diff
@@ -49,16 +52,124 @@ func TestRecursiveIteratorEmptyBaseCase(t *testing.T) {
...
-	// Execute - should terminate immediately with empty result
-	ctx := NewTestContext(t)
+	ds, err := memdb.NewMemdbDatastore(0, 0, memdb.DisableGC)
+	require.NoError(t, err)
 
-	// Test CheckImpl with a faulty iterator
-	seq, err := recursive.CheckImpl(ctx, []Object{{ObjectType: "folder", ObjectID: "folder1"}}, ObjectAndRelation{ObjectType: "user", ObjectID: "tom", Relation: "..."})
-	require.NoError(t, err, "CheckImpl should return sequence without error")
+	ctx := NewLocalContext(context.Background(),
+		WithRevisionedReader(datalayer.NewDataLayer(ds).SnapshotReader(datastore.NoRevision)))
```
miparnisari: the test passes without `ds`. here and in line 156

pkg/query/recursive_coverage_test.go
```diff
@@ -198,14 +198,12 @@ func TestIterativeDeepening_MaxDepth(t *testing.T) {
...
 	require.NoError(err)
 
-	paths, err := CollectAll(seq)
-	require.NoError(err)
-
-	// Should have run exactly maxDepth iterations
-	require.LessOrEqual(len(paths), maxDepth, "Should not exceed max depth")
+	// Should complete within maxDepth iterations (result may or may not be found)
+	_ = path
+	_ = maxDepth
```
miparnisari: what is this doing?

pkg/query/recursive.go
```diff
@@ -548,185 +548,161 @@ func (r *RecursiveIterator) breadthFirstIterResources(ctx *Context, subject Obje
...
+				if err != nil {
+					return nil, err
 				}
+				result = merged
 			}
+		}
 
+		// Early termination: if we previously found a path but this ply adds nothing new,
+		// we've reached a fixed point.
+		if plyPath == nil && foundAtPreviousPly {
```
miparnisari, line 596: this block is never hit by unit tests. please add a test.

there are also quite a few other blocks in this file that are not covered by tests.

pkg/query/recursive.go
```diff
@@ -548,185 +548,161 @@ func (r *RecursiveIterator) breadthFirstIterResources(ctx *Context, subject Obje
...
+			return nil, err
+		}
+		if GetObject(path.Subject).Equals(GetObject(subject)) {
+			if ctx.shouldTrace() {
+				ctx.TraceStep(r, "Check via IterSubjects: found matching path")
 			}
-
-			// Filter paths where subject matches input subject (compare only type and ID, not relation)
-			for path, err := range pathSeq {
+			if result == nil {
```
miparnisari: this structure:

```go
 			if result == nil {
				result = path
			} else {
				merged, err := result.MergeOr(path)
				if err != nil {
					return nil, err
				}
				result = merged
			}
```

i can see it in 6 different places. 

any chance of a refactor?

pkg/query/observer_count.go
```diff
@@ -31,41 +31,41 @@ func NewCountObserver() *CountObserver {
...
-func (c *CountObserver) ObservePath(op ObserverOperation, key CanonicalKey, _ *Path) {
+func (c *CountObserver) ObservePath(op Operation, key CanonicalKey, _ *Path) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	stats := c.stats[key]
 	switch op {
-	case CheckOperation:
+	case Check:
 		stats.CheckResults++
```
miparnisari: if path is `nil` we still increment?

pkg/query/context.go
```diff
@@ -214,28 +212,61 @@ func (ctx *Context) wrapPathSeqWithObservers(op ObserverOperation, key Canonical
...
+		tracedIterator = ctx.traceEnterIfEnabled(it, checkTraceString(resource, subject))
 	}
 	key := it.CanonicalKey()
-	ctx.notifyEnterIterator(CheckOperation, key)
+	ctx.notifyEnterIterator(Check, key)
 
-	pathSeq, err := ctx.Executor.Check(ctx, it, resources, subject)
+	path, err := ctx.Executor.Check(ctx, it, resource, subject)
 	if err != nil {
 		return nil, err
```
miparnisari, line 235: no `ctx.notifyCheckResult(key, nil)` ?

pkg/query/arrow.go
barakmich: good catch!

pkg/query/alias.go
barakmich: This was a corner case to IterSubjects that slows things down. We need to also check for the existence of the subject in other conditions if we're enumerating the subjects (the consistency tests found this). If we don't have to do this check though, because we're doing a different operation that happens to call IterSubjects, then we can skip it.

pkg/query/exclusion_test.go
barakmich: nah, I don't like ligatures

pkg/query/fixed.go
barakmich: It can be, but for fixed it shouldn't matter

pkg/query/recursive_test.go
barakmich: It's actually an opportunity for another optimization!

pkg/query/recursive.go
barakmich: yep! nice, now that they're pointers

pkg/query/context.go
barakmich: Errors should bubble up as much as possible. This may throw the stats a little, but so could notifying a nil path on error.

PR #3023, opened 2026-04-07
feat(crdb): sentinel-drain cancel handling for connection pool

ecordell:
## Description

[WIP]: requires further testing

Cancelled write contexts in the CockroachDB driver previously caused pgx to close the connection (its default behavior), depleting the write pool and requiring the `write-conn-acquisition-timeout` (30ms) as backpressure. Read contexts were severed entirely via `SeparatingContextDatastoreProxy` to avoid the same problem for reads (authzed/spicedb#953).

A prior attempt to fix this (authzed/spicedb#2294) used pgx's built-in `CancelRequestContextWatcherHandler` but was reverted: the 100ms sleep that pgx uses to avoid a post-cancel race wasn't reliable; stale `SQLSTATE 57014` errors could still leak onto the next query.

This PR fixes the race deterministically. When a context is cancelled, a cancel handler sends a PostgreSQL cancel request to CRDB (so it stops work early), then the connection is polled with `SELECT 1` before returning to the pool. `SELECT 1` either succeeds (no cancel outstanding; connection is clean) or returns `57014` (cancel in progress). After a successful `SELECT 1`, CRDB's cancel semantics should guarantee the connection is safe to re-use.

With the pool no longer depleted by cancellations:
- `write-conn-acquisition-timeout` defaults to 0 (disabled)
- `SeparatingContextDatastoreProxy` is removed: cancelled reads now also send a cancel to CRDB and drain cleanly
- The whole behavior is gated behind `--datastore-experimental-cancel-draining` (default: on) so operators can revert to the old behavior if needed; disabling it also restores the 30ms timeout

## Testing

Ran local tests but this needs more validation; hence the draft. If it's inconclusive I'll switch the default to `off`.

## References

- authzed/spicedb#953 original context-severing proxy (now removed)
- authzed/spicedb#2294 initial attempt to remove severing
- authzed/spicedb#2434 prior revert (the race this PR fixes)


internal/datastore/crdb/pool/cancel_handler.go
```diff
@@ -0,0 +1,97 @@
...
+	done   chan struct{}
+	stopFn context.CancelFunc
+}
+
+var _ ctxwatch.Handler = (*cancelHandler)(nil)
+
+// HandleCancel is called by pgx's context watcher when the context is cancelled.
+// It sends a cancel request to CRDB in a goroutine and sets a net.Conn deadline
+// so that the blocked Exec/Query returns promptly.
+func (h *cancelHandler) HandleCancel(_ context.Context) {
```
tstirrat15: Is the context in the argument the context that has already been cancelled, and that's why we're creating a new context below this?

internal/datastore/crdb/pool/cancel_handler_test.go
```diff
@@ -0,0 +1,139 @@
...
+			calls++
+			<-ctx.Done()
+			return ctx.Err()
+		},
+	}
+	ctx, cancel := context.WithTimeout(t.Context(), 10*time.Millisecond)
+	defer cancel()
+	result := drainConnWithContext(ctx, blocking)
+	require.False(t, result, "expected timed-out connection to return false")
+	require.Equal(t, 1, calls, "expected one Exec attempt before timeout")
```
tstirrat15: This should use `synctest`

internal/datastore/crdb/pool/pool.go
```diff
@@ -28,6 +31,74 @@ type pgxPool interface {
...
+		config.ConnConfig.BuildContextWatcherHandler = func(pgConn *pgconn.PgConn) ctxwatch.Handler {
+			h := &cancelHandler{
+				pgConn:        pgConn,
+				deadlineDelay: 1 * time.Second,
+			}
+			key := weak.Make(pgConn)
+			handlers.Store(key, weak.Make(h))
+			// When h is collected (connection failed; the pgConn ↔ cancelHandler
+			// cycle became unreachable), remove the stale map entry.
+			runtime.AddCleanup(h, func(key weak.Pointer[pgconn.PgConn]) {
```
tstirrat15: I haven't used `AddCleanup` before. Is there much of an overhead associated with it? Is this a hot enough path to be worried about that? Is there potential for thrashing around these values?

internal/datastore/crdb/pool/pool.go
```diff
@@ -28,6 +31,74 @@ type pgxPool interface {
...
+			}, key)
+			return h
+		}
+
+		// Use the pgconn-level AfterConnect to mark the handler ready.
+		// This fires inside pgconn.ConnectConfig, immediately after connectOne
+		// writes pid/secretKey and before the connection is surfaced to pgxpool.
+		// Using the pgconn hook avoids interfering with pgxpool.Config.AfterConnect
+		// (which NewRetryPool wraps for pool housekeeping).
+		prevPgconn := config.ConnConfig.AfterConnect
```
tstirrat15: Nit:
```suggestion
		prevAfterConnect := config.ConnConfig.AfterConnect
```
And associated changes

internal/datastore/crdb/pool/pool.go
```diff
@@ -309,6 +393,7 @@ func (p *RetryPool) withRetries(ctx context.Context, acquireTimeout time.Duratio
 		err = wrapRetryableError(ctx, fn(conn))
 		if err == nil {
 			conn.Release()
+			conn = nil // suppress the deferred release
```
tstirrat15, line 312: This seems like a potential footgun. Is there a different place that we can set the `defer` so that we don't have to do this bookkeeping?

internal/datastore/crdb/pool/pool.go
```diff
@@ -346,13 +432,25 @@ func (p *RetryPool) withRetries(ctx context.Context, acquireTimeout time.Duratio
...
+		// deterministically before pool release. Without the cancel handler
+		// there is no in-flight CancelRequest to absorb, so we skip the drain.
+		// If the connection is broken, GC it so the pool replaces it.
+		if p.cancelHandlerInstalled && (errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)) {
+			if conn != nil && !p.drainCancelledConnFn(conn) {
+				p.GC(conn.Conn())
+			}
+		} else if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
 			log.Ctx(ctx).Warn().Err(err).Uint8("retries", retries).Msg("error is not resettable or retryable")
 		}
```
tstirrat15: Could the conditionals here be reorganized to make this read more nicely? the `errors.Is` clauses are the inverse of each other.

internal/datastore/crdb/pool/pool.go
```diff
@@ -459,5 +557,68 @@ func wrapRetryableError(ctx context.Context, err error) error {
...
 	}
+
+	// A query-canceled (57014) with a live context means a stale CancelRequest
+	// from a prior context cancellation reached the connection after drainConn
+	// had already confirmed it clean (the net.Conn deadline fired and SELECT 1
+	// succeeded before the in-flight cancel arrived at CRDB). The connection is
+	// clean once CRDB returns 57014; retry on the same connection.
+	if sqlErrorCode(err) == CrdbQueryCanceledCode {
+		return &RetryableError{Err: err}
+	}
```
tstirrat15: Are we expecting to hit this branch in practice?

internal/datastore/crdb/pool/pool.go
```diff
@@ -459,5 +557,68 @@ func wrapRetryableError(ctx context.Context, err error) error {
...
+// drainConn is the testable core of drainCancelledConn. It uses a 5-second
+// timeout: generous enough to handle a briefly-unresponsive CRDB node, yet
+// bounded so a partitioned node doesn't hold a pool connection indefinitely.
+// In practice only 0–2 iterations occur (the cancel either already hit the
+// original query or arrives on the first SELECT 1).
+func drainConn(q connQueryer) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	return drainConnWithContext(ctx, q)
+}
```
tstirrat15: Curious about context propagation here - if there's a parent context available, should it be used, or are we intentionally severing the context here?

internal/datastore/crdb/pool/pool_test.go
```diff
@@ -304,3 +305,153 @@ func TestSuccessfulFunctionExecution(t *testing.T) {
...
+
+	require.ErrorIs(t, err, context.Canceled)
+}
+
+// TestWithRetriesCancelHandlerBranchEntered verifies that when the cancel
+// handler is installed and fn returns context.Canceled, withRetries enters the
+// cancel-handler branch (not the general "not resettable or retryable" path).
+// With a nil conn (unavoidable in unit tests — pgxpool.Conn requires a real
+// pool), the drain function is skipped due to the nil check, but the branch
+// selection is confirmed by the error type: context.Canceled without wrapping.
```
tstirrat15: I'm a little confused on this point - how is it distinguishable from the above behavior if the assertion in both cases is `errors.Is(context.Cancelled)`?

internal/datastore/crdb/pool/pool_test.go
```diff
@@ -304,3 +305,153 @@ func TestSuccessfulFunctionExecution(t *testing.T) {
...
+	require.ErrorIs(t, err, context.Canceled)
+	// Drain is not called because conn is nil (unit test limitation), but the
+	// cancel-handler branch was entered — confirmed by the error being unwrapped
+	// context.Canceled rather than a logged "not resettable or retryable" error.
+	require.False(t, drainCalled, "drain not called with nil conn")
+}
+
+// TestWithRetriesSkipsDrainWhenCancelHandlerNotInstalled verifies that without
+// the cancel handler, context.Canceled errors do not trigger drain logic.
+func TestWithRetriesSkipsDrainWhenCancelHandlerNotInstalled(t *testing.T) {
```
tstirrat15: I'm also not clear on how this test differs from the preceding one

internal/datastore/crdb/pool/sqlerrors_test.go
```diff
@@ -447,3 +448,7 @@ func Test_sqlErrorCode(t *testing.T) {
 		})
 	}
 }
+
+func TestCrdbQueryCanceledCode(t *testing.T) {
+	require.Equal(t, "57014", CrdbQueryCanceledCode)
+}
```
tstirrat15: Necessary?

internal/datastore/postgres/common/cancelation.go
tstirrat15, line 1: For future readers: this is the aforementioned built-in handler that has undesirable behavior around racing

internal/datastore/crdb/pool/pool.go
```diff
@@ -28,6 +31,74 @@ type pgxPool interface {
...
+		//
+		// Both key and value are weak pointers so the map does not prevent the
+		// GC from collecting a failed connection's pgConn ↔ cancelHandler cycle.
+		// On success, AfterConnect removes the entry via LoadAndDelete. On
+		// failure (connectOne returns an error before AfterConnect fires), the
+		// pgConn and cancelHandler become unreachable — the GC collects the
+		// cycle, and AddCleanup then deletes the stale map entry.
+		var handlers sync.Map // weak.Pointer[pgconn.PgConn] → weak.Pointer[cancelHandler]
+
+		config.ConnConfig.BuildContextWatcherHandler = func(pgConn *pgconn.PgConn) ctxwatch.Handler {
```
tstirrat15: What does it mean that we're overriding `BuildContextWatcherHandler`? Is that something we control, or could we potentially be whacking something here?

internal/datastore/crdb/pool/cancel_handler.go
```diff
@@ -0,0 +1,97 @@
...
+// sends a PostgreSQL cancel request to CRDB (so CRDB stops executing the query
+// early), sets a deadline on the connection's net.Conn, and returns. It does NOT
+// sleep 100ms like pgx's default CancelRequestContextWatcherHandler.
+//
+// The connection is NOT immediately safe to reuse after HandleCancel — it may
+// still receive SQLSTATE 57014 on the next query. Callers must drain the
+// connection (via drainCancelledConn in withRetries) before returning it to the
+// pool.
+type cancelHandler struct {
+	pgConn        *pgconn.PgConn
```
tstirrat15: Is this the `pgConn` whose work we want to cancel, or is this the connection that we're using to send the cancel request? Is the cancel request transparently handled by pgx?

tstirrat15 (commented):
Mostly looks good; will approve after we've done more validation.

Is this something that we'd want to extend to the PG driver as well? Would it be easy to do so with the current structure of the code?

PR #3031, opened 2026-04-10
feat: Added support for YAML-based validation files in the Language S…

miparnisari:
…erver

## Description

Support YAML in VS Code.

## Testing

https://github.com/user-attachments/assets/71da8545-73f0-4521-aeec-e1ac7cf70df7

## References

n/a

### DIFF
--- a/internal/lsp/handlers.go
+++ b/internal/lsp/handlers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"slices"
 	"strings"
 
@@ -18,6 +19,9 @@ import (
 	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/schemadsl/generator"
 	"github.com/authzed/spicedb/pkg/schemadsl/input"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+	"github.com/authzed/spicedb/pkg/validationfile"
+	"github.com/authzed/spicedb/pkg/validationfile/blocks"
 )
 
 func (s *Server) textDocDiagnostic(ctx context.Context, r *jsonrpc2.Request) (FullDocumentDiagnosticReport, error) {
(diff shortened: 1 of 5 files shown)


internal/lsp/handlers.go
```diff
@@ -335,9 +484,15 @@ func (s *Server) textDocDefinition(_ context.Context, r *jsonrpc2.Request) (*lsp
 			targetURI = resolveURI(params.TextDocument.URI, string(*resolved.TargetSource))
 		}
 
+		r := newLspRange(resolved)
+		r.Start.Line += off.lineOff
+		r.Start.Character += off.colOff
+		r.End.Line += off.lineOff
+		r.End.Character += off.colOff
```
barakmich: maybe newLspRange takes an `off`?

internal/lsp/handlers.go
```diff
@@ -268,6 +406,10 @@ func (s *Server) textDocHover(_ context.Context, r *jsonrpc2.Request) (*Hover, e
 		var lspRange *lsp.Range
 		if resolved.TargetPosition != nil {
 			lspRange = newLspRange(resolved)
+			lspRange.Start.Line += off.lineOff
+			lspRange.Start.Character += off.colOff
+			lspRange.End.Line += off.lineOff
+			lspRange.End.Character += off.colOff
```
barakmich: for here too

internal/lsp/handlers.go
```diff
@@ -54,8 +58,81 @@ func (s *Server) computeDiagnostics(ctx context.Context, uri lsp.DocumentURI) ([
...
+				}
+				for _, assertErr := range assertionErrors {
+					diag := newLspDiagnostic(assertErr, lsp.Error)
+					adjustAssertionDiagnostic(&diag, file.contents)
+					diagnostics = append(diagnostics, diag)
+				}
+			}
+
+			return nil
+		}
```
barakmich, line 150: Since this `if` always returns -- it sure seems like this block should be its own function

tstirrat15 (dismissed):
LGTM

internal/lsp/handlers.go
miparnisari: done

tstirrat15 (approved):
LGTM

PR #3038, opened 2026-04-13
chore: upgrade spanner lib

tstirrat15:
## Description
Per [this discord discussion](https://discord.com/channels/844600078504951838/1435532794041471068], v1.89.0 of the spanner lib may help prevent a deadlock that some users have encountered. Upgrading is likely something that we want to do anyway.

## Changes
* Bump spanner lib and associated packages
## Testing
Review. See that tests pass.


internal/datastore/spanner/migrations/driver.go
```diff
@@ -42,12 +42,12 @@ func NewSpannerDriver(ctx context.Context, database, credentialsFilePath, emulat
 	}
 	log.Ctx(ctx).Info().Str("spanner-emulator-host", os.Getenv(emulatorSettingKey)).Msg("spanner emulator")
 	log.Ctx(ctx).Info().Str("credentials", credentialsFilePath).Str("db", database).Msg("connecting")
-	client, err := spanner.NewClient(ctx, database, option.WithCredentialsFile(credentialsFilePath))
+	client, err := spanner.NewClient(ctx, database, option.WithCredentialsFile(credentialsFilePath)) //nolint:staticcheck  // TODO: provide authn pathway through ADC https://pkg.go.dev/cloud.google.com/go#hdr-Authentication_and_Authorization
```
tstirrat15: I added deprecations in the flags for these as well.

internal/datastore/spanner/options.go
```diff
@@ -44,8 +44,6 @@ type spannerOptions struct {
 	disableStats                 bool
 	readMaxOpen                  int
 	writeMaxOpen                 int
-	minSessions                  uint64
-	maxSessions                  uint64
```
tstirrat15, line 48: These no longer do anything in the spanner client so I'm removing them;.

internal/datastore/spanner/options.go
```diff
@@ -39,13 +39,10 @@ type spannerOptions struct {
 	followerReadDelay            time.Duration
 	maxRevisionStalenessPercent  float64
 	credentialsFilePath          string
-	credentialsJSON              []byte
```
tstirrat15, line 42: There wasn't any way to actually set this option in the flags so I removed it.

internal/datastore/spanner/spanner.go
```diff
@@ -164,12 +157,12 @@ func NewSpannerDatastore(ctx context.Context, database string, opts ...Option) (
 	)
 
 	clientConfig := spanner.ClientConfig{
-		SessionPoolConfig:    cfg,
 		DisableNativeMetrics: config.datastoreMetricsOption != DatastoreMetricsOptionNative,
 	}
 	if meterProvider != nil {
 		clientConfig.OpenTelemetryMeterProvider = meterProvider
 	}
+	// TODO: why is this context.background instead of the incoming context?
```
tstirrat15: Need to ask on this point.

internal/datastore/spanner/spanner.go
```diff
@@ -164,12 +157,12 @@ func NewSpannerDatastore(ctx context.Context, database string, opts ...Option) (
 	)
 
 	clientConfig := spanner.ClientConfig{
-		SessionPoolConfig:    cfg,
```
tstirrat15, line 167: No longer relevant

internal/gateway/gateway.go
tstirrat15, line 1: These changes are related to deprecations that had been in effect for a while but we'd silenced them and they were finally removed.

internal/gateway/gateway.go
```diff
@@ -162,22 +162,44 @@ func registerHandler(ctx context.Context, mux *runtime.ServeMux, endpoint string
...
+	propagator := otel.GetTextMapPropagator()
+	ctx = propagator.Extract(ctx, propagation.HeaderCarrier(r.Header))
+	propagator.Inject(ctx, carrier)
 	return metadataCopy
 }
 
+type metadataCarrier struct {
+	mdata *metadata.MD
+}
+
```
tstirrat15, line 184: I thought about pulling this into a separate file but this is the only place where this approach is necessary

pkg/cmd/datastore/datastore.go
```diff
@@ -295,9 +294,21 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
...
 	flagSet.StringVar(&opts.SpannerCredentialsFile, flagName("datastore-spanner-credentials"), "", "path to service account key credentials file with access to the cloud spanner instance (omit to use application default credentials)")
+	err = flagSet.MarkDeprecated(flagName("datastore-spanner-credentials"), "prefer Application Default Credentials: https://docs.cloud.google.com/docs/authentication/client-libraries#adc")
+	if err != nil {
+		return err
+	}
 	flagSet.StringVar(&opts.SpannerEmulatorHost, flagName("datastore-spanner-emulator-host"), "", "URI of spanner emulator instance used for development and testing (e.g. localhost:9010)")
 	flagSet.Uint64Var(&opts.SpannerMinSessions, flagName("datastore-spanner-min-sessions"), 100, "minimum number of sessions across all Spanner gRPC connections the client can have at a given time")
+	err = flagSet.MarkDeprecated(flagName("datastore-spanner-min-sessions"), "sessions flags are deprecated as the client no longer uses a session pool")
```
tstirrat15, line 313: These flags are now no-op

pkg/cmd/datastore/datastore.go
```diff
@@ -295,9 +294,21 @@ func RegisterDatastoreFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, opt
 	flagSet.BoolVar(&opts.EnableConnectionBalancing, flagName("datastore-connection-balancing"), defaults.EnableConnectionBalancing, "enable connection balancing between database nodes (CockroachDB driver only)")
 	flagSet.DurationVar(&opts.ConnectRate, flagName("datastore-connect-rate"), 100*time.Millisecond, "rate at which new connections are allowed to the datastore (at a rate of 1/duration) (CockroachDB driver only)")
 	flagSet.StringVar(&opts.SpannerCredentialsFile, flagName("datastore-spanner-credentials"), "", "path to service account key credentials file with access to the cloud spanner instance (omit to use application default credentials)")
+	err = flagSet.MarkDeprecated(flagName("datastore-spanner-credentials"), "prefer Application Default Credentials: https://docs.cloud.google.com/docs/authentication/client-libraries#adc")
```
tstirrat15, line 307: Application Default Credentials are provided by the environment (through one mechanism or another) so we don't need a new flag. This marks the old approach as deprecated.

tstirrat15 (commented):
See comments

internal/datastore/spanner/spanner.go
```diff
@@ -164,13 +157,12 @@ func NewSpannerDatastore(ctx context.Context, database string, opts ...Option) (
...
 		DisableNativeMetrics: config.datastoreMetricsOption != DatastoreMetricsOptionNative,
 	}
 	if meterProvider != nil {
 		clientConfig.OpenTelemetryMeterProvider = meterProvider
 	}
-	client, err := spanner.NewClientWithConfig(context.Background(), database, clientConfig, spannerOpts...)
+	client, err := spanner.NewClientWithConfig(ctx, database, clientConfig, spannerOpts...)
```
tstirrat15, line 168: I think this was an oversight - there isn't a reason that I can see to sever the context here.

tstirrat15 (commented):
See comment

internal/datastore/spanner/options.go
miparnisari: WDYM? https://github.com/authzed/spicedb/blob/fd6ca35383dd3e7dff184f32afcda33e30cb926f/pkg/cmd/datastore/datastore.go#L722

miparnisari: omg there is no

```go
flagSet.StringVar(&opts.SpannerCredentialsJSON, ...)
```

🤦‍♀️

internal/gateway/gateway.go
```diff
@@ -162,22 +162,44 @@ func registerHandler(ctx context.Context, mux *runtime.ServeMux, endpoint string
...
+func (mc metadataCarrier) Set(key, value string) {
+	mc.mdata.Set(key, value)
+}
+
+func (mc metadataCarrier) Keys() []string {
+	keys := make([]string, 0, mc.mdata.Len())
+	for k := range *mc.mdata {
+		keys = append(keys, k)
+	}
+	return keys
```
jzelinskie: slices.Collect(maps.Keys(*nc.mdata))

internal/datastore/spanner/options.go
miparnisari: actually, hold on. The user that contributed this feature said

> I am using SpiceDB library mode 

which is why they didn't need a flag.

internal/datastore/spanner/spanner.go
miparnisari: we need to add a note in the changelog about this

miparnisari (changes requested):
this PR is a breaking change in terms of the flags that we support. We should amend the changelog to say so.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Removed MySQL metrics prefixed with `go_sql_stats_connections_*` in favor of those prefixed with `go_sql_*` (https://github.com/authzed/spicedb/pull/2980)
 - Removed support for Spanner flag value `--datastore-spanner-metrics=deprecated-prometheus`; please use values `otel` or `native` (https://github.com/authzed/spicedb/pull/2980)
 - Reduced binary size (https://github.com/authzed/spicedb/pull/3005)
+- Upgraded the spanner client, which changed the internal implementation to not use a session pool. This means that the `--datastore-spanner-max-sessions` and `--datastore-spanner-min-sessions` flags are now deprecated and no-op. We also strongly recommend using [Application Default Credentials](https://docs.cloud.google.com/docs/authentication/client-libraries#adc) in favor of a credentials file.
```
```suggestion
- Upgraded the spanner client, which changed the internal implementation to not use a session pool. This means that the `--datastore-spanner-max-sessions` and `--datastore-spanner-min-sessions` flags are now deprecated and no-op. We also strongly recommend using [Application Default Credentials](https://docs.cloud.google.com/docs/authentication/client-libraries#adc) in favor of a credentials file. (https://github.com/authzed/spicedb/pull/3038)
```

miparnisari (approved):

miparnisari (changes requested):
Let's undo the JSON change as per https://github.com/authzed/spicedb/issues/1941#issuecomment-4240867726

miparnisari (approved):

PR #3039, opened 2026-04-13
fix: wildcards in query planner

barakmich:
## Description

Noticing the error `"ERROR: index \"pk_relation_tuple\" cannot be used for this query (SQLSTATE 42809)"` led to finding an inconsistency in arguments being sent from the query planner's datastore abstraction when requesting relations that matched wildcards. 

This PR cleans up the handling of wildcards within query planner by making them an expected part of the Path returns. This means `Arrow`s need to handle wildcards properly as well, and so on up the chain, matching behavior from previous implementations.

It also calls the right shape (or no shape) for AllSubjectsForResources depending on the call.

## Testing

Local testing and integration testing against CRDB


pkg/query/arrow.go
miparnisari, line 1: please add unit tests. None of the changes in this file are covered by tests.

are you able to see these annotations in Github's UI?

image

I recommend downloading the [Codecov chrome extension](https://chromewebstore.google.com/detail/codecov/gedikamndpbemklijjkncpnolildpbgo): 

which adds green & red coloring
image

pkg/query/exclusion.go
```diff
@@ -167,10 +175,49 @@ func (e *ExclusionIterator) IterSubjectsImpl(ctx *Context, resource Object, filt
 			}
 			mainCount++
 
-			// Check if this subject exists in the excluded set
+			if mainPath.Subject.ObjectID == tuple.PublicWildcard {
```
miparnisari, line 178: please add unit tests for this too.

pkg/query/wildcard_subjects_test.go
miparnisari, line 1: I feel like we could extend this file to test wildcards across Check / LR / LS and across arrows, intersection arrows, exclusion, intersection, etc...

pkg/query/intersection_arrow.go
miparnisari, line 1: need unit tests here too.

miparnisari (changes requested):

miparnisari (changes requested):
there is still uncovered code for arrows and intersection arrows

pkg/query/wildcard_arrow_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	path, err := ctx.Check(iarrow, NewObject("document", "doc1"), NewObject("user", "alice").WithEllipses())
+	require.NoError(t, err)
+	require.NotNil(t, path, "wildcard left should be satisfied via IterResources inversion")
+}
+
+// TestIntersectionArrowIterator_WildcardSkip_IterSubjects verifies the
+// parallel skip branch in IntersectionArrowIterator.IterSubjectsImpl.
+func TestIntersectionArrowIterator_WildcardSkip_IterSubjects(t *testing.T) {
+	ds, rev, dsSchema := setupTestDB(t, arrowWildcardSchema,
```
miparnisari, line 164: nit: the test is called `TestIntersectionArrow` but the schema used `arrowWildcardSchema` doesn't have intersection arrows. Same comment for the test above: `TestIntersectionArrowIterator_WildcardInversion_Check`

i guess it doesn't matter too much because we have line 180 , it's just slightly confusing 😆

miparnisari (dismissed):

miparnisari (approved):

PR #3040, opened 2026-04-14
chore: fix workflow deprecations

tstirrat15:
## Description
Fixing one of the warnings here and trying to get rid of trivy flaps: https://github.com/authzed/spicedb/actions/runs/24415833647?pr=3038

## Changes
* Bump version and let float
## Testing
Review.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -51,7 +51,7 @@ jobs:
           # trivy does not flag an old SpiceDB version
           fetch-depth: 0
       - uses: "authzed/actions/setup-go@main"
-      - uses: "docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9" # v3.7.0
+      - uses: "docker/login-action@v4" # v3.7.0
         with:
           username: "${{ env.DOCKERHUB_PUBLIC_USER }}"
           password: "${{ env.DOCKERHUB_PUBLIC_ACCESS_TOKEN }}"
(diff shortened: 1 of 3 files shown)


miparnisari (dismissed):

Dockerfile
```diff
@@ -6,18 +6,9 @@ COPY . .
...
-RUN apk update && apk add --no-cache git
-RUN git clone https://github.com/authzed/grpc-health-probe.git
-WORKDIR /go/src/app/grpc-health-probe
-RUN git checkout main
-RUN CGO_ENABLED=0 go install -a -tags netgo -ldflags=-w
-
 # use `docker buildx imagetools inspect <image>` to get the multi-platform sha256
 FROM cgr.dev/chainguard/static@sha256:99a5f826e71115aef9f63368120a6aa518323e052297718e9bf084fb84def93c
-COPY --from=health-probe-builder /go/bin/grpc-health-probe /bin/grpc_health_probe
+COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.48 /ko-app/grpc-health-probe /usr/local/bin/grpc_health_probe
```
tstirrat15, line 11: Our fork [is now in sync](https://github.com/authzed/grpc-health-probe) with the upstream. The upstream publishes a container, which means we can grab the binary directly using a remote copy. This avoids us needing to build it ourselves, which is convenient.

tstirrat15 (commented):
See comments

.github/workflows/security.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
           # trivy does not flag an old SpiceDB version
           fetch-depth: 0
       - uses: "authzed/actions/setup-go@main"
-      - uses: "docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9" # v3.7.0
+      - uses: "docker/login-action@v4" # v3.7.0
```
miparnisari, line 54:
```suggestion
      - uses: "docker/login-action@v4"
```

Dockerfile
miparnisari: does this mean we can archive `github.com/authzed/grpc-health-probe`? (i don't remember why we forked it in the first place)

miparnisari (approved):

PR #3042, opened 2026-04-14
chore: work around ghcr limitations and update goreleaser config

tstirrat15:
## Description
In using the `COPY` command to pull in the built version of `grpc-health-probe`, we were hitting ghcr egress limits for their organization. This changes the approach to use `ADD` and grab the binary directly.

I also ran `goreleaser check` on the configuration files and found some deprecations, so I migrated the config file.

## Changes
Will annotate

## Testing
```
  # Build the binary
  CGO_ENABLED=0 go build -tags memoryprotection -ldflags="-s -w -checklinkname=0" -o spicedb ./cmd/spicedb

  # Set up the directory structure goreleaser would create
  mkdir -p linux/amd64 && cp spicedb linux/amd64/spicedb

  # Build the image
  docker buildx build -f Dockerfile.release --platform linux/amd64 -t spicedb-release-test .
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -58,7 +58,7 @@ nfpms:
         dst: "/usr/share/man/man1/spicedb.1"
         file_info:
           mode: 0644
-furies:
+gemfury:
   - account: "authzed"
     secret_name: "GEMFURY_PUSH_TOKEN"
     disable: "{{ gt (len .Prerelease) 0 }}"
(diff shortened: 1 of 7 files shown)


.goreleaser.yml
```diff
@@ -58,7 +58,7 @@ nfpms:
         dst: "/usr/share/man/man1/spicedb.1"
         file_info:
           mode: 0644
-furies:
+gemfury:
```
tstirrat15, line 61: This is a name deprecation; everything else is the same

.goreleaser.yml
```diff
@@ -81,121 +81,57 @@ snapcrafts:
         plugs:
           - "network-bind"
           - "network"
-brews:
```
tstirrat15, line 84: This became `homebrew_casks` and its build got simplified pretty considerably

.goreleaser.yml
```diff
@@ -81,121 +81,57 @@ snapcrafts:
...
+  - alternative_names:
+      - "{{.ProjectName}}@{{.Version}}"
+    description: "Google Zanzibar-inspired permissions database for fine-grained access control"
+    homepage: "https://github.com/authzed/spicedb"
+    url:
+      template: "https://github.com/authzed/spicedb/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
+    completions:
+      bash: "completions/spicedb.bash"
+      zsh: "completions/spicedb.zsh"
+      fish: "completions/spicedb.fish"
```
tstirrat15, line 94: You can make this generate the completions for you, but because we're using completions across so many different keys, we might as well keep this approach.

.goreleaser.yml
```diff
@@ -81,121 +81,57 @@ snapcrafts:
...
+    repository:
+      owner: "authzed"
+      name: "homebrew-tap"
+      token: "{{ .Env.HOMEBREW_TAP_GITHUB_TOKEN }}"
+    commit_author:
+      name: "authzedbot"
+      email: "infrastructure@authzed.com"
+    commit_msg_template: "Brew cask update for {{ .ProjectName }} version {{ .Tag }}"
+    skip_upload: "auto"
+dockers_v2:
```
tstirrat15, line 106: `dockers_v2` uses `buildx`, which means that the individual builds aren't necessary. It also simplifies the configuration by a lot because it automatically generates manifests.

Dockerfile.release
```diff
@@ -3,9 +3,12 @@
...
 FROM $BASE
-# NOTE: this copies in the health probe binary from the official build of the container.
-COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.48 /ko-app/grpc-health-probe /usr/local/bin/grpc_health_probe
-COPY spicedb /usr/local/bin/spicedb
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETPLATFORM
+# NOTE: downloads the health probe binary directly from the official release.
+ADD --chmod=755 https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.48/grpc-health-probe-${TARGETOS}-${TARGETARCH} /usr/local/bin/grpc_health_probe
+COPY $TARGETPLATFORM/spicedb /usr/local/bin/spicedb
```
tstirrat15, line 11: This change reflects the update to `dockers_v2` - this is how you reference the built binary for a particular platform

Dockerfile.release
```diff
@@ -3,9 +3,12 @@
...
 
 FROM $BASE
-# NOTE: this copies in the health probe binary from the official build of the container.
-COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.48 /ko-app/grpc-health-probe /usr/local/bin/grpc_health_probe
-COPY spicedb /usr/local/bin/spicedb
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETPLATFORM
+# NOTE: downloads the health probe binary directly from the official release.
+ADD --chmod=755 https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.48/grpc-health-probe-${TARGETOS}-${TARGETARCH} /usr/local/bin/grpc_health_probe
```
tstirrat15: This is what replaces the `COPY` command - we instead grab the binary directly from the release page. This seemed easier than trying to figure out how to mirror the built image to somewhere that we could reliably pull from.

tstirrat15 (commented):
See comments

Dockerfile.release
miparnisari:
```suggestion
ADD --chmod=755 https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.48/grpc_health_probe-${TARGETOS}-${TARGETARCH} /usr/local/bin/grpc_health_probe
```

tstirrat15: lol ty i just figured this out as well

tstirrat15:
Another step would be to create a `spicedb-base` repo and image that just has chainguard + the grpc health probe and use that as the base image for `ko`. Having our own image hosted on our own org would (probably) mean we don't get throttled.

barakmich (approved):

PR #3044, opened 2026-04-15
chore: refactor TestHelpLinks

tstirrat15:
## Description
The output of this test isn't very helpful: https://github.com/authzed/spicedb/actions/runs/24448553627/job/71483554267?pr=3019

I think the failing test is because github is telling a non-browser user agent to screw off, but this should make it a bit clearer when things fail.

## Changes
Refactor test to provide more context on failure

## Testing
Review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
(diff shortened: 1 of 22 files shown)


barakmich (approved):

tstirrat15:
I also merged in the `v1.51.1` branch, which makes that tag an ancestor of the HEAD of main again. This means that goreleaser identifies that release as the most recent one, which fixes the trivy issues we were seeing.

Issue #2788, opened 2025-12-22
Stuck pgxpool connections on LookupResource calls


heissa83:
### What platforms are affected?

linux

### What architectures are affected?

amd64

### What SpiceDB version are you using?

v1.47.1

### What SpiceDB flags do you use?

GOMAXPROCS: "3"
SPICEDB_DASHBOARD_ENABLED: "false"
SPICEDB_DATASTORE_ENGINE: postgres
SPICEDB_DISPATCH_CLUSTER_ENABLED: "false"
SPICEDB_DISPATCH_UPSTREAM_ADDR: ""
SPICEDB_GRPC_SHUTDOWN_GRACE_PERIOD: 1s
SPICEDB_HTTP_ENABLED: "false"
SPICEDB_LOG_LEVEL: info
SPICEDB_OTEL_PROVIDER: otlpgrpc
SPICEDB_WRITE_RELATIONSHIPS_MAX_UPDATES_PER_CALL: "6000"

### Steps to Reproduce

We are using a very simplistic schema:
system -> organization -> folder -> project
Connected through a parent relation on the higher resource with a owner and a reader role and a modify and a read permission.
We create data for 50 organizations, each with 10 folders each with 10 projects.
We add some subjects to the roles.

Then we create a large amount of parallel LookupResource requests (500-1000) at the same time.

### Expected Result

Maybe a slow down or a timeout on the calls made to the SpiceDB. But after all the Requests have been cancelled or timed out the SpiceDB should respond to requests again.

### Actual Result

The SpiceDB stays stuck and does not respond to any further requests. Debugging showed that all available pgxpool connections are used and so all newly incoming calls wait infinitely to claim a new connection from the pool. The health check still says that the Instance is healthy but it will not recover from this state.

tstirrat15:
> wait infinitely

Why aren't connections being returned to the pool as the queries complete?

How big is your postgres instance, and how many connections are you allocating to each SpiceDB instance? It looks like you haven't configured the read and write pool values, so they'll be using the defaults:

```
      --datastore-conn-pool-read-max-open int                                 number of concurrent connections open in a remote datastore's connection pool (default 20)
      --datastore-conn-pool-read-min-open int                                 number of minimum concurrent connections open in a remote datastore's connection pool (default 20)
      --datastore-conn-pool-write-max-open int                                number of concurrent connections open in a remote datastore's connection pool (default 10)
[...]

PR #3048, opened 2026-04-15
fix: put a timeout to callers of methods that use singleflight

miparnisari:
## Description

Potential fix for https://github.com/authzed/spicedb/issues/2788

image

## Testing

```
docker-compose -f docker-compose.deadlock.yaml up -d
sh reproduce_deadlock.sh
```

[reproduce_deadlock.sh](https://github.com/user-attachments/files/26760763/reproduce_deadlock.sh)
[docker-compose.deadlock.yaml](https://github.com/user-attachments/files/26760765/docker-compose.deadlock.yaml)

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -20,7 +20,6 @@ import (
 	"github.com/shopspring/decimal"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/sync/errgroup"
-	"resenje.org/singleflight"
 
 	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/internal/datastore/crdb/migrations"
@@ -279,7 +278,6 @@ type crdbDatastore struct {
 	beginChangefeedQuery string
 	transactionNowQuery  string
 
-	featuresGroup  singleflight.Group[string, *datastore.Features]
 	cachedFeatures *datastore.Features // GUARDED_BY(featuresLock)
 	featuresLock   sync.Mutex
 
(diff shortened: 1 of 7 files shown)


internal/datastore/crdb/crdb.go
```diff
@@ -543,9 +541,7 @@ func (cds *crdbDatastore) Features(ctx context.Context) (*datastore.Features, er
 		return cached, nil
 	}
 
-	features, _, err := cds.featuresGroup.Do(ctx, "", func(ictx context.Context) (*datastore.Features, error) {
-		return cds.features(ictx)
-	})
```
miparnisari, line 586: FYI this is a refactor. Singleflight is dangerous so i want to use it in as few files as possible. I am moving this to `singleflight.go`

pkg/cmd/server/server_test.go
```diff
@@ -471,7 +471,7 @@ func TestServerGracefulTerminationOnError(t *testing.T) {
 	}, WithPresharedSecureKey("psk"), WithDatastore(ds), WithEnableMemoryProtectionMiddleware(false))
 	cancel()
 	_, err = c.Complete(ctx)
-	require.NoError(t, err)
+	require.ErrorIs(t, err, context.Canceled)
```
miparnisari, line 474: Before this PR, `Complete(ctx)` went straight to

```go
func (mdb *memdbDatastore) Features(_ context.Context) (*datastore.Features, error) {
	return mdb.OfflineFeatures()
}
```

Now it passes through the proxy which reacts to the `ctx`.

miparnisari:
superseded by https://github.com/authzed/spicedb/pull/3142

PR #3049, opened 2026-04-15
refactor: combine benchmarks into integration, and add short/long benchmarks

barakmich:
## Description

Benchmark runs have been slowing down our CI, and we have more on the way. This PR splits benchmarks into short and long, where long is defined as "involves pulling docker backends" and runs for a shorter time. The full long tests are still run on main.

This also provides comparison points for the query planner, unifies the integration tests and the query planner benchmarks (which are now more sparse), and removes the duplicate BenchmarkQueryRelationships

## Testing

The whole PR applies to testing alone


miparnisari:
Interesting! https://github.com/authzed/spicedb/actions/runs/24481601361/job/71547258089?pr=3049

for the majority of the benchmarks, the query planner performs more iterations. But for a few, it performs less. E.g.

```
BenchmarkServices/memory/WideGroups/LookupResources/dispatch        	    1184	   2920949 ns/op	  851693 B/op	   10106 allocs/op
BenchmarkServices/memory/WideGroups/LookupResources/queryplan       	     296	   8500733 ns/op	 1774830 B/op	   43522 allocs/op
```

.github/workflows/benchmark.yaml
```diff
@@ -17,7 +17,7 @@ permissions:
 
 jobs:
   benchmarks:
-    name: "Run Go benchmarks on Main"
+    name: "Run Full Go benchmarks on Main"
```
miparnisari, line 20: we should modify this workflow so that if the benchmarks on HEAD get 2x worse compared to HEAD~1, we need to be alerted: 

1. get the commit SHA for HEAD~1
2. restore benchmark from the cache
3. run the comparison with `fail-on-alert: true` 
4. `if: failure` then slack message

internal/datastore/memdb/memdb_test.go
```diff
@@ -146,35 +146,3 @@ func TestAnythingAfterCloseDoesNotPanic(t *testing.T) {
 	_, err = reader.CountRelationships(t.Context(), "blah")
 	require.ErrorIs(err, ErrMemDBIsClosed)
 }
-
-func BenchmarkQueryRelationships(b *testing.B) {
```
miparnisari, line 150: 👍 
you could also get rid of `BenchmarkPostgresQuery` since it's basically the same as `BenchmarkDatastoreDriver/TestTuple/SnapshotReadOnlyNamespace`

miparnisari (dismissed):

miparnisari (approved):

PR #3051, opened 2026-04-15
feat(queryopt): set simplification optimizer

jzelinskie:
## Description

Adds a new query-plan optimizer, set-simplification, that eliminates structurally redundant nodes from outline trees using ten set-theoretic laws. The optimizer runs as a standard pass (registered in StandardOptimizations) and requires no cardinality or schema information — all laws hold unconditionally.

Laws implemented

Union
- A ∪ A = A (idempotency)
- A ∪ (A ∩ B) = A (absorption)
- A ∪ (A − B) = A (complement-absorption)

Intersection
- A ∩ A = A (idempotency)
- A ∩ (A ∪ B) = A (absorption)
- Y ∩ (A − C) = ∅ when Y ⊆ C (complement-annihilation)

Exclusion
- X − A = ∅ when X ⊆ A (annihilation, generalized from A − A = ∅)
- (A ∪ B) − Y = B − Y when A ⊆ Y (left-pruning)
- A − ∅ = A (null-identity)

Structural prerequisite
- (A ∪ B) ∪ C = A ∪ B ∪ C and (A ∩ B) ∩ C = A ∩ B ∩ C (associativity flattening, runs first so absorption sees all
  peers)

### Implementation notes

The optimizer is structured as one function per law, all composed via MutateOutline in a single bottom-up pass. Subsumption tests reuse the existing isSubsumedBy predicate; no new structural predicates were introduced.

A − ∅ = A lives in the absorption mutation chain rather than in the shared NullPropagation utility. NullPropagation is also called by reachability-pruning, which intentionally leaves A − ∅ intact (the null subtrahend is a meaningful pruning artifact in that context).

The optimizer was initially named absorption-idempotency; it is renamed to set-simplification to reflect the full scope of rules it now covers.

## Testing

Added unit tests for everything including those missing for NullPropagation.
LMK if I need to add any other kind of tests when adding optimizers

## References

- [Wikipedia for set theory absorption](https://en.wikipedia.org/wiki/Absorption_law)

### DIFF
--- a/pkg/query/mutations.go
+++ b/pkg/query/mutations.go
@@ -22,7 +22,7 @@ func MutateOutline(outline Outline, fns []OutlineMutation) Outline {
 // upward through the tree according to each node type's semantics:
 //   - Union: null if ALL children are null
 //   - Intersection: null if ANY child is null
-//   - Arrow/IntersectionArrow: null if the right child is null
+//   - Arrow/IntersectionArrow: null if either child is null
 //   - Exclusion: null if the left child is null
 //   - Caveat/Alias/Recursive: null if the only child is null
 //
@@ -47,7 +47,7 @@ func NullPropagation(outline Outline) Outline {
 		}
 
 	case ArrowIteratorType, IntersectionArrowIteratorType:
-		if len(outline.SubOutlines) == 2 && outline.SubOutlines[1].Type == NullIteratorType {
+		if len(outline.SubOutlines) == 2 && (outline.SubOutlines[0].Type == NullIteratorType || outline.SubOutlines[1].Type == NullIteratorType) {
 			return Outline{Type: NullIteratorType, ID: outline.ID}
 		}
 
--- a/pkg/query/mutations_test.go
+++ b/pkg/query/mutations_test.go
@@ -6,6 +6,268 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestNullPropagation(t *testing.T) {
+	live := Outline{Type: FixedIteratorType}
+	null := Outline{Type: NullIteratorType}
+
+	// --- UnionIteratorType ---
+
+	t.Run("union: all null → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: UnionIteratorType, SubOutlines: []Outline{null, null}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("union: all null (3 children) → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: UnionIteratorType, SubOutlines: []Outline{null, null, null}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("union: one live child → unchanged", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: UnionIteratorType, SubOutlines: []Outline{live, null}})
+		require.Equal(t, UnionIteratorType, result.Type)
+	})
+
+	t.Run("union: mixed (3 children, 1 live) → unchanged", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: UnionIteratorType, SubOutlines: []Outline{null, live, null}})
+		require.Equal(t, UnionIteratorType, result.Type)
+	})
+
+	t.Run("union: all live → unchanged", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: UnionIteratorType, SubOutlines: []Outline{live, live}})
+		require.Equal(t, UnionIteratorType, result.Type)
+	})
+
+	t.Run("union: preserves ID when null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: UnionIteratorType, ID: 7, SubOutlines: []Outline{null, null}})
+		require.Equal(t, NullIteratorType, result.Type)
+		require.Equal(t, OutlineNodeID(7), result.ID)
+	})
+
+	// --- IntersectionIteratorType ---
+
+	t.Run("intersection: any null child → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: IntersectionIteratorType, SubOutlines: []Outline{live, null}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("intersection: null first child → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: IntersectionIteratorType, SubOutlines: []Outline{null, live}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("intersection: all null → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: IntersectionIteratorType, SubOutlines: []Outline{null, null}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("intersection: one null in 3 children → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: IntersectionIteratorType, SubOutlines: []Outline{live, null, live}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("intersection: all live → unchanged", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: IntersectionIteratorType, SubOutlines: []Outline{live, live}})
+		require.Equal(t, IntersectionIteratorType, result.Type)
+	})
+
+	t.Run("intersection: preserves ID when null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: IntersectionIteratorType, ID: 8, SubOutlines: []Outline{live, null}})
+		require.Equal(t, NullIteratorType, result.Type)
+		require.Equal(t, OutlineNodeID(8), result.ID)
+	})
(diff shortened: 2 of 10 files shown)


pkg/query/queryopt/caveat_pushdown.go
```diff
@@ -9,6 +9,7 @@ func init() {
 		Pushes caveat evalution to the lowest point in the tree.
 		Cannot push through intersection arrows
 		`,
+		Priority: 20,
```
tstirrat15, line 12: Do we have a means of constructing a list of optimizers at startup time rather than taking the registration approach? This smells like it's going to turn into `z-index` eventually.

tstirrat15 (dismissed):
LGTM but would like barak's eyes on it as well

pkg/query/queryopt/caveat_pushdown.go
jzelinskie: This is definitely a place that could use work. I think for now it'd probably help to just have a central list where optimizers are registered, similar how to we register gRPC middleware. If it ever got too confusing doing that, it'd make sense to explicitly add "before" and "after" properties to each optimizer and let the system compute the order based on those (kind like systemd startup ordering)

josephschorr: @jzelinskie I'd say we do the before/after now: its going to rapidly become untenable, so I recommend as a followup PR

pkg/query/mutations.go
```diff
@@ -22,7 +22,7 @@ func MutateOutline(outline Outline, fns []OutlineMutation) Outline {
 // upward through the tree according to each node type's semantics:
 //   - Union: null if ALL children are null
 //   - Intersection: null if ANY child is null
-//   - Arrow/IntersectionArrow: null if the right child is null
+//   - Arrow/IntersectionArrow: null if the either child is null
```
miparnisari:
```suggestion
//   - Arrow/IntersectionArrow: null if either child is null
```

pkg/query/mutations_test.go
```diff
@@ -6,6 +6,263 @@ import (
...
+	})
+
+	// --- ExclusionIteratorType ---
+
+	t.Run("exclusion: null left child (minuend) → null", func(t *testing.T) {
+		result := NullPropagation(Outline{Type: ExclusionIteratorType, SubOutlines: []Outline{null, live}})
+		require.Equal(t, NullIteratorType, result.Type)
+	})
+
+	t.Run("exclusion: null right child (subtrahend) → unchanged (A − ∅ = A)", func(t *testing.T) {
```
miparnisari, line 154: super extra nitpick

```suggestion
	t.Run("exclusion: null left child (main set) → null", func(t *testing.T) {
		result := NullPropagation(Outline{Type: ExclusionIteratorType, SubOutlines: []Outline{null, live}})
		require.Equal(t, NullIteratorType, result.Type)
	})

	t.Run("exclusion: null right child (excluded set) → unchanged (A − ∅ = A)", func(t *testing.T) {
```

we've been using `main` and `excluded` across the codebase.

jzelinskie: Because this optimizer implementation is using all set theory terminology, it might make sense to make an exception for this scenario. I'm willing to update this if you think consistency in the overall codebase is more important than consistency across the tests/implementation.

EDIT: oops, this is in the nullpropagation! i'll totally update this now

pkg/query/queryopt/set_simplification_test.go
miparnisari, line 1: could we add more tests? the coverage is quite poor: https://app.codecov.io/gh/authzed/spicedb/pull/3051/blob/pkg/query/queryopt/set_simplification.go

pkg/query/queryopt/set_simplification.go
```diff
@@ -0,0 +1,399 @@
...
+// appears among Y's direct children. Anything satisfying Y must also satisfy X, so
+// the union already covers Y via X.
+func unionAbsorption(outline query.Outline) query.Outline {
+	return eliminateRedundantChildren(outline, query.UnionIteratorType, func(y, x query.Outline) bool {
+		if y.Type != query.IntersectionIteratorType {
+			return false
+		}
+		for _, factor := range intersectionFactors(x) {
+			if !slices.ContainsFunc(y.SubOutlines, func(c query.Outline) bool {
+				return query.OutlineCompare(factor, c) == 0
```
tstirrat15, line 127: This doesn't necessarily need to be a part of this PR, but it'd be nice to have an `OutlineEquals` alias for this.

tstirrat15, line 127: Is this potentially expensive for an intersection at the top of a bunch of deep trees? Or are we expecting that the shapes we're working with here aren't going to be problematic in that way?

pkg/query/queryopt/set_simplification.go
```diff
@@ -0,0 +1,399 @@
...
+// unionComplementAbsorption is an OutlineMutation that drops exclusion children
+// from a union when they are subsumed by a sibling.
+// Law: A ∪ (A − B) = A
+//
+// A child Y is dropped when Y is an exclusion (M − S) and the minuend M is
+// subsumed by some sibling X. Since (M − S) ⊆ M ⊆ X, the union already covers Y.
+func unionComplementAbsorption(outline query.Outline) query.Outline {
+	return eliminateRedundantChildren(outline, query.UnionIteratorType, func(y, x query.Outline) bool {
+		return y.Type == query.ExclusionIteratorType &&
+			len(y.SubOutlines) == 2 &&
```
tstirrat15, line 145: Is this not satisfied by construction? Or are we being defensive here?

pkg/query/queryopt/set_simplification.go
```diff
@@ -0,0 +1,399 @@
...
+//
+// Elements of A − C are outside C by definition; elements of Y are inside C
+// (since Y ⊆ C); the two sets are disjoint so their intersection is empty.
+func intersectionComplementAnnihilation(outline query.Outline) query.Outline {
+	if outline.Type != query.IntersectionIteratorType {
+		return outline
+	}
+
+	for i, child := range outline.SubOutlines {
+		if child.Type != query.ExclusionIteratorType || len(child.SubOutlines) != 2 {
```
tstirrat15, line 187: Same question here

tstirrat15 (commented):
See comments

pkg/query/queryopt/set_simplification.go
jzelinskie: This was just being defensive in case of refactoring.

pkg/query/queryopt/set_simplification.go
jzelinskie: defensiveness

pkg/query/queryopt/set_simplification.go
jzelinskie: I didn't evaluate the cost of outline compares when drafting this. I assumed they were cheap, but that's definitely not the case.

Here's what Claude thinks about this:

>`unionAbsorption` complexity at a single Union node:
>
> `eliminateRedundantChildren` does O(n²) pair comparisons. For each `(y, x)` pair in `shouldDrop`:
>
> ```go
> for _, factor := range intersectionFactors(x) {             // O(k) factors of x
>     if !slices.ContainsFunc(y.SubOutlines, factor.Equals) { // O(m) calls per factor
>         return false
>     }
> }
> ```
>
> Each `factor.Equals` call invokes `OutlineCompare` (outline.go:428), which recursively walks the entire subtree — O(T) where T is subtree size. There is no memoization or hash shortcut.
>
>So total cost at a Union node: O(n² × k × m × T)
>
>If you have `Union[Intersection[A,B], Intersection[A,B,C], ...]` where A, B, C are deep trees, checking whether `A.Equals(A)` does a full tree walk each time — even comparing the same subtree to itself.
>
>The early exits in `OutlineCompare` (line 432: type first, then args) help when subtrees differ near the root, but for the absorption case the trees being compared will often share a common prefix all the way down.
>
> The fix: Precompute a `Serialize()` hash per node before running mutations, and use it as an O(1) equality fast path. `Serialize()` (outline.go:573) already exists and `CanonicalKey.Hash()` (line 53) wraps xxhash. The blocker is that `Outline` is passed by value with no cached hash field, and `ID` is only populated post-canonicalization.
>
>The cleanest option is adding a lazily-precomputed `serialHash uint64` field to `Outline`, set during the pre-optimization `FillMissingNodeIDs` pass, and short-circuiting `OutlineCompare` with a hash comparison first.

I think this suggestion is a way bigger change that needs feedback from @barakmich.

pkg/query/queryopt/set_simplification_test.go
jzelinskie: The file you linked has 96.93% coverage. Did you mean something else? I still added a commit to cover these remaining lines.

pkg/query/queryopt/set_simplification.go
jzelinskie: Looks like you did this elsewhere, so I adopted the method in another commit.

jzelinskie: I discussed with @barakmich a bit. This makes sense as a follow up because there could be other places that could benefit from this (e.g. cache keys).

tstirrat15 (approved):
LGTM

PR #3052, opened 2026-04-15
fix: use advise in query plan code paths

barakmich:
## Description

We were doing structural optimizations in the code path for the query plan flag, but not stats-based ones (even though we were collecting stats) -- so let's do that!


internal/services/v1/permissions_queryplan_test.go
```diff
@@ -179,19 +179,10 @@ func TestQueryPlanMetadataMergeMultipleKeys(t *testing.T) {
 }
 
 func TestPermissionServerQueryPlanMetadataInitialization(t *testing.T) {
```
miparnisari:
```suggestion
func TestQueryPlanMetadataInitialization(t *testing.T) {
```

miparnisari (dismissed):

miparnisari:
Sounds like the changelog should be updated 😉

miparnisari (approved):

PR #3053, opened 2026-04-15
chore: Revert "fix: add classic histogram buckets to perf insights metric (#3027)"

tstirrat15:
## Description
Reverts #3027. This change would have greatly increased the cardinality of the metrics and we don't want to do that.

## Changes
* Revert the change
## Testing
Review.


tstirrat15:
Dupe of #3045

PR #3055, opened 2026-04-16
build: refactor mage targets and build tags to prevent silent skipping of tests

miparnisari:
## Description

- refactor `mage` targets to NOT list test specific dirs which is error-prone.
- remove tag `skipintegrationtests` which was confusing
- remove tag `docker` which was unnecessary

## Testing

With this PR, codecov coverage MUST go up.
If we are omitting any tests, Codecov will tell us because coverage will drop on those files.

## References

pre-requisite for https://github.com/authzed/spicedb/pull/3016

### DIFF
--- a/internal/fdw/pgserver_e2e_test.go
+++ b/internal/fdw/pgserver_e2e_test.go
@@ -690,11 +690,21 @@ func runPGServer(t *testing.T, client *authzed.Client) int {
 	})
 
 	go func() {
-		_ = pgserver.Run(ctx, fmt.Sprintf("localhost:%d", port))
+		// Bind to all interfaces so the Postgres container can reach us via
+		// host.docker.internal. "localhost" binds to 127.0.0.1 only, which the
+		// container cannot reach (its packets arrive from the docker bridge IP).
+		_ = pgserver.Run(ctx, fmt.Sprintf("0.0.0.0:%d", port))
 	}()
 
-	// Give PGServer time to start
-	time.Sleep(50 * time.Millisecond)
+	// Wait until the server is actually accepting connections.
+	require.Eventually(t, func() bool {
+		conn, err := net.DialTimeout("tcp", fmt.Sprintf("localhost:%d", port), 100*time.Millisecond)
+		if err != nil {
+			return false
+		}
+		_ = conn.Close()
+		return true
+	}, 10*time.Second, 20*time.Millisecond, "PGServer did not start accepting connections")
 
 	return port
 }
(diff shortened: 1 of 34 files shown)


internal/services/integrationtesting/consistency_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build !skipintegrationtests
+//go:build integration || datastoreconsistency
```
miparnisari, line 1: TODO: send followup. I have to do this ugly thing because there are shared helpers defined here. The fix is to move the helpers to a test files that is NOT behind a tag.

internal/fdw/pgserver_e2e_test.go
```diff
@@ -690,11 +690,21 @@ func runPGServer(t *testing.T, client *authzed.Client) int {
...
-	// Give PGServer time to start
-	time.Sleep(50 * time.Millisecond)
+	// Wait until the server is actually accepting connections.
+	require.Eventually(t, func() bool {
+		conn, err := net.DialTimeout("tcp", fmt.Sprintf("localhost:%d", port), 100*time.Millisecond)
+		if err != nil {
+			return false
+		}
+		_ = conn.Close()
+		return true
```
tstirrat15: Prefer `EventuallyWithT` here so that we can see what's failing when it fails

magefiles/test.go
```diff
@@ -259,11 +274,39 @@ func (Testcons) Mysql(ctx context.Context) error {
...
 	args = append(args, coverageFlags...)
 	return goDirTestWithEnv(ctx, ".", "./internal/services/integrationtesting/...",
 		env,
 		args...)
 }
+
+// findDirsWithBuildTag returns Go package paths (e.g. "./internal/fdw") containing at least
+// one _test.go file whose //go:build constraint references tag. Lets us target the right
+// packages without hard-coding folder paths that can drift.
+func findDirsWithBuildTag(tag string) ([]string, error) {
```
tstirrat15, line 289: I'm confused on this - why is it insufficient to do `./...`?

tstirrat15 (dismissed):
LGTM, but see comment

magefiles/test.go
miparnisari: because it runs the unit tests too 🤦‍♀️ 

we probably should be adding `if testing.short() { skip }` to all integration tests and then change `test:unit` to include `-short`. but that's a bigger change i think

tstirrat15: Ugh. Okay, I agree that this PR is better than the status quo. I wish it wasn't necessary, though.

tstirrat15 (dismissed):
LGTM

magefiles/test.go
miparnisari: baby steps :)

tstirrat15 (approved):
LGTM

PR #3058, opened 2026-04-17
ci: on merge, compare benchmark and slack if regression

miparnisari:
## Description

- Follow-up of https://github.com/authzed/spicedb/pull/3049#discussion_r3090011788. Since now the full benchmark suite runs only on pushes to `main`, we want to make sure that we get alerted if they regress.
- Ignore benchmarking code from code coverage

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -440,6 +440,8 @@ jobs:
       needs.paths-filter.outputs.codechange == 'true' &&
       github.event_name == 'pull_request' &&
       github.base_ref == 'main'
+    permissions:
+      pull-requests: "write"
     steps:
       - uses: "actions/checkout@v6"
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yaml
```diff
@@ -440,6 +440,8 @@ jobs:
       needs.paths-filter.outputs.codechange == 'true' &&
       github.event_name == 'pull_request' &&
       github.base_ref == 'main'
+    permissions:
+      pull-requests: "write"
```
miparnisari, line 444: fixes a 403 error

```
 data: {
      message: 'Resource not accessible by integration',
      documentation_url: 'https://docs.github.com/rest/pulls/reviews#create-a-review-for-a-pull-request',
      status: '403'
    }
```

source: https://github.com/authzed/spicedb/actions/runs/24590845143/job/71911126559?pr=3056

tstirrat15 (approved):
LGTM

PR #3063, opened 2026-04-21
test: Improve code coverage of tests

josephschorr:
Adds unit tests for various packages that had insufficient coverage


internal/dispatch/combined/combined_test.go
```diff
@@ -58,3 +63,117 @@ func TestCombinedRecursiveCall(t *testing.T) {
...
+		MetricsEnabled(true),
+		PrometheusSubsystem("test_subsystem"),
+		DispatchChunkSize(50),
+		RelationshipChunkCacheConfig(cfg),
+		CaveatTypeSet(caveattypes.Default.TypeSet),
+		Cache(dispatchCache),
+		ConcurrencyLimits(graph.ConcurrencyLimits{Check: 10, LookupResources: 5}),
+	)
+	require.NoError(t, err)
+	require.NotNil(t, dispatcher)
```
tstirrat15: How does this assert that the options were applied? Or are we just saying that they don't cause errors when provided as inputs?

tstirrat15: This comment goes for every test in this file

internal/services/v1/errors_test.go
```diff
@@ -0,0 +1,265 @@
...
+		Operation: v1.Precondition_OPERATION_MUST_MATCH,
+		Filter: &v1.RelationshipFilter{
+			ResourceType: "document",
+		},
+	}
+	err := func() PreconditionFailedError {
+		var target PreconditionFailedError
+		_ = errors.As(NewPreconditionFailedErr(precondition), &target)
+		return target
+	}()
```
tstirrat15: Is this necessary because there isn't a `New*` constructor for this error type?

pkg/datalayer/datalayer_test.go
```diff
@@ -0,0 +1,459 @@
...
+	require.NotNil(t, errCh)
+
+	require.Equal(t, underlying, UnwrapDatastore(dl))
+	require.NoError(t, dl.Close())
+}
+
+// TestCountingDataLayer_CountsQueryAndReverseQuery ensures QueryRelationships
+// and ReverseQueryRelationships increment their counters on both the snapshot
+// reader and the RW transaction.
+func TestCountingDataLayer_CountsQueryAndReverseQuery(t *testing.T) {
```
tstirrat15: I thought we already had this test? What codepath is it landing on?

pkg/schema/v2/flatten_test.go
```diff
@@ -11,6 +11,49 @@ import (
...
+	require.NoError(t, err)
+	require.NotNil(t, flattened)
+	require.Same(t, flattened.resolvedSchema, flattened.ResolvedSchema())
+
+	visitor := &testVisitor{}
+	_, err = WalkFlattenedSchema(flattened, visitor, struct{}{})
+	require.NoError(t, err)
+
+	require.NotEmpty(t, visitor.schemas)
+	require.GreaterOrEqual(t, len(visitor.definitions), 2)
```
tstirrat15, line 41: Is this just testing that the code doesn't error, more or less?

tstirrat15 (dismissed):
See comments, otherwise LGTM

internal/datastore/proxy/checkingreplicated_test.go
```diff
@@ -16,6 +16,84 @@ import (
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
+func TestCheckingReplicatedWithNoReplicasReturnsPrimary(t *testing.T) {
```
miparnisari, line 19: there is a godoc in the prod code that says

```
// NewCheckingReplicatedDatastore (...)
// NOTE: Be *very* careful when using this function. It is not safe to use this function without
// knowledge of the layout of the underlying datastore and its replicas.
// the replicas *must* point to a *stable* instance of the datastore (not a load balancer).
```

is it possible to encode this knowledge in a test?

Similarly, for `NewStrictReplicatedDatastore`:

```
// NewStrictReplicatedDatastore (...)
// This is useful when the read pool points to a load balancer that can transparently handle the request.
```

can we encode that in a test?

(see https://github.com/authzed/spicedb/issues/2525#issuecomment-4100524433 for more context)

internal/graph/errors_test.go
```diff
@@ -0,0 +1,144 @@
...
+func TestAlwaysFailError(t *testing.T) {
+	err := NewAlwaysFailErr()
+
+	var afe AlwaysFailError
+	require.ErrorAs(t, err, &afe)
+	require.Equal(t, "always fail", err.Error())
+}
+
+func TestRelationNotFoundError(t *testing.T) {
+	err := NewRelationNotFoundErr("document", "viewer")
```
miparnisari: I cannot find any references to `NewRelationNotFoundErr`. Might as well remove it and get rid of the test too.

pkg/datalayer/datalayer_test.go
miparnisari: i don't even know why we have the `CountingDataLayer` 🫠 its only prod use has `datalayer.UnaryCountingInterceptor(nil)`. so what's the point? we're not doing anything with the counts.

is this for future use? and how does it differ from the `observableProxy`?

if we _do_ need it, can we split this file into `counting_test.go`, `impl_test.go`, `readonly_test.go`, etc?

miparnisari (changes requested):
Thanks for doing this. Just a few comments. The suggestion on the additional test can be done in a follow up because i think it's not trivial.

internal/dispatch/combined/combined_test.go
miparnisari: changed code to be able to assert on the final options

internal/services/v1/errors_test.go
miparnisari: `NewPreconditionFailedErr` returns the `error` interface, but the test needs the concrete `PreconditionFailedError` value to call `GRPCStatus()` on it.

Updated.

miparnisari (approved):

PR #3065, opened 2026-04-22
fix: avoid hash collision in caveat context

tstirrat15:
## Description
Someone pointed out to us that if you use a nested list of lists in a caveat context, it was possible under our current serialization and hashing logic to have a serialization collision and therefore a dispatch cache hash collision.

This fixes that by using deterministic proto marshalling of the caveat context Struct object.

## Changes
* Update `context_hash.go` to use deterministic marshalling
* Propagate the error signature change through the stack
* change `hashableContext` to `hashableContextString` to satisfy the interface
## Testing
Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Upgraded the spanner client, which changed the internal implementation to not use a session pool. This means that the `--datastore-spanner-max-sessions` and `--datastore-spanner-min-sessions` flags are now deprecated and no-op. We also strongly recommend using [Application Default Credentials](https://docs.cloud.google.com/docs/authentication/client-libraries#adc) in favor of a credentials file. (https://github.com/authzed/spicedb/pull/3038)
 - Query Planner: error `"ERROR: index \"pk_relation_tuple\" cannot be used for this query (SQLSTATE 42809)"` returned when using wildcards (https://github.com/authzed/spicedb/pull/3039)
 - Providing one of (`--grpc-tls-cert-path`, `--grpc-tls-key-path`) but not the other is now considered an error state, as both are necessary if you want to use TLS.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Upgraded the spanner client, which changed the internal implementation to not use a session pool. This means that the `--datastore-spanner-max-sessions` and `--datastore-spanner-min-sessions` flags are now deprecated and no-op. We also strongly recommend using [Application Default Credentials](https://docs.cloud.google.com/docs/authentication/client-libraries#adc) in favor of a credentials file. (https://github.com/authzed/spicedb/pull/3038)
 - Query Planner: error `"ERROR: index \"pk_relation_tuple\" cannot be used for this query (SQLSTATE 42809)"` returned when using wildcards (https://github.com/authzed/spicedb/pull/3039)
 - Providing one of (`--grpc-tls-cert-path`, `--grpc-tls-key-path`) but not the other is now considered an error state, as both are necessary if you want to use TLS.
+- In a caveat context that uses nested lists of lists, the hashes generated for cache keys could collide because of an issue with the serialization logic. The serialization now uses deterministic protobuf serialization which avoids this issue.
```
```suggestion
- In a caveat context that uses nested lists of lists, the hashes generated for cache keys could collide because of an issue with the serialization logic. The serialization now uses deterministic protobuf serialization which avoids this issue (https://github.com/authzed/spicedb/pull/3065)
```

internal/services/v1/permissions_test.go
```diff
@@ -2446,3 +2446,68 @@ func TestExportBulkRelationshipsWithFilter(t *testing.T) {
 		})
 	}
 }
+
+// TestBulkCheckCaveatContextCollision is a regression test for an issue with
+// caveat hash collision.
+func TestBulkCheckCaveatContextCollision(t *testing.T) {
```
miparnisari, line 2452: why do we need an end to end test for something that could easily be encoded at a unit test level - compare the hashing of `{"x": ["a", [], "b"]}` and `{"x": [["a"], "b"]}` and ensures they are not the same?

the test you wrote, if it fails, tells us very little about where the problem could be.

miparnisari (approved):

PR #3070, opened 2026-04-23
feat: Add debug output to MaxRecursionDepthError when debug header is set

tstirrat15:
## Description
Opening as an alternative/continuation to #3036. This takes a slightly different approach, which is to wait until a max depth error is encountered and then build the trace as the dispatch stack is unwound. It tracks across dispatch boundaries as well.

## Notes
This code is a no-op if the request flag isn't present or if a max recursion depth error isn't encountered.

## Changes
Will annotate.

## Testing
Review. See that tests pass. See authzed/zed#682 and its testing section for an example of code that drives this.


internal/dispatch/tracing.go
```diff
@@ -0,0 +1,118 @@
...
+const traversalTraceKey spiceerrors.MetadataKey = "dispatch_traversal_trace"
+
+// PrependTraversalFrame adds a new frame to the front of the traversal trace carried
+// by a MaxDepthExceeded error, then returns the updated error. If err is not a
+// MaxDepthExceeded condition it is returned unchanged.
+//
+// Designed to be called once per dispatchIter level as the error unwinds: each level
+// prepends its frame so the outermost caller sees the full path from root to leaf.
+// Because the trace is stored in the gRPC ErrorInfo metadata it survives a network hop.
+func PrependTraversalFrame(originalErr error, resourceType, resourceID, relation string) error {
```
tstirrat15: Note here: the traces are the reverse of a normal stack trace, in the sense that they start with the subject and continue down as maximum recursion depth is approached. I think it ends up making sense from a user experience perspective.

internal/graph/lookupresources2.go
```diff
@@ -686,6 +686,33 @@ func (crr *CursoredLookupResources2) redispatchOrReport(
 			// all found results, as no further filtering will be needed.
 			if entrypoint.IsDirectResult() {
 				stream := unfilteredLookupResourcesDispatchStreamForEntrypoint(ctx, foundResources, parentStream, ci)
+				if parentRequest.EnableDebugTrace {
+					// Debug path: per-subject dispatch so each recursive frame is annotated.
```
tstirrat15: This is a design choice - it will increase concurrency for debug requests, but it makes it easier to isolate the specific path that the LR took to hit the recursion depth limit, and it's understood that debugs are often slower.

There's a question here of whether it'd be dangerous to run against an active cluster - will it potentially consume too many resources?

internal/dispatch/tracing.go
```diff
@@ -0,0 +1,118 @@
...
+		return
+	}
+	store.mu.Lock()
+	store.trace = trace
+	store.mu.Unlock()
+}
+
+// SnapshotLookupDebugTrace returns the last trace saved via SaveTraceToContext,
+// or nil if none was saved or no tracker is present.
+func SnapshotLookupDebugTrace(ctx context.Context) *dispatchv1.LookupDebugTrace {
```
miparnisari: unused

miparnisari (commented):
Please include in the testing section of the PR description:
- the schema + relationships that would produce a cycle.
- the CURL requests we can use to test out the thing e2e for LR and LS. 

With this command and [this](https://authzed.com/docs/spicedb/modeling/recursion-and-max-depth#example-of-a-cycle) data (plus another rel `group:secondgroup#member@user:maria`), I get an error as expected (400), but I don't get any debug info. Can you look into why it doesn't work with http? 

```
curl --location --request POST 'http://localhost:8443/v1/permissions/resources' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'io.spicedb.requestdebuginfo: true' \
--header 'Authorization: Bearer super-secret-key' \
--data '{
  "resourceObjectType": "resource",
  "permission": "view",
  "subject": {
    "object": {
      "objectType": "user",
      "objectId": "maria"
    }
  }
}'
```

internal/services/v1/permissions.go
miparnisari, line 1: why did we not update LS as well??

internal/services/v1/permissions.go
```diff
@@ -529,6 +541,11 @@ func (ps *permissionServer) lookupResources3(req *v1.LookupResourcesRequest, res
 	}
 	usagemetrics.SetInContext(ctx, respMetadata)
 
+	debugEnabled := lookupDebugTraceEnabled(ctx)
```
miparnisari: this is confusing. In the incoming request you look for `io.spicedb.requestdebuginfo`, but internally you are passing this as a flag `enable_debug_trace`? why not just pick one or the other (header vs body parameter)

internal/services/v1/debug_test.go
```diff
@@ -921,3 +928,188 @@ func TestBulkCheckPermissionWithDebug(t *testing.T) {
...
+	req.Equal("someuser", trace.Frames[0].GetResourceId())
+	req.Equal("...", trace.Frames[0].GetRelation())
+
+	// Should have folder a as the third frame
+	req.Equal("folder", trace.Frames[2].GetResourceType())
+	req.Equal("a", trace.Frames[2].GetResourceId())
+	req.Equal("view", trace.Frames[2].GetRelation())
+}
+
+func TestLookupResourcesDebugTraceV2_LR2(t *testing.T) {
```
miparnisari: the test above is also testing lr2, no? what is the difference between this test and the one above?

internal/services/v1/permissions.go
tstirrat15: Because it'd be easy enough to do as a follow-on and you can already get the same behavior by doing a `check --explain` with a random subject and the same object. LS uses the same paths/direction as Check; LR goes the other direction.

If we determine it's necessary, it'll be pretty easy to add.

internal/services/v1/permissions.go
tstirrat15: Fair question; I'm happy to go with either approach but wanted to do some bike shedding on body parameter vs header.

internal/services/v1/debug_test.go
tstirrat15: No - the one above tests LR3, somewhat confusingly. If you don't explicitly disable LR3 it's on by default in tests - see https://github.com/authzed/spicedb/pull/3070/changes#diff-455357489fc60317ea646ef4d4da567f4e36e3a9a8d4e738e396381228732b93R1051.

The name of the test is from the original name of the proto in the original PR. I'll update the name.

internal/graph/lookupresources2.go
josephschorr: Yes and it will be MUCH slower and cause cache misses, etc. We cannot do this - even behind a debug flag

miparnisari: what if we do this: run a 2 phase LR?

on phase 1 we run the normal LR. 

if we hit a MaxDepthExceededError AND if `debug=true` - we initiate phase 2 which re-executes the original request but does NOT emit any results, and does this per-subject dispatch.

internal/services/v1/permissions.go
tstirrat15: Went for body

pkg/datastore/test/relationships.go
```diff
@@ -707,7 +707,7 @@ func DeleteWithInvalidPrefixTest(t *testing.T, tester DatastoreTester) {
 		return err
 	})
 	require.Error(err)
-	require.ErrorContains(err, "value does not match regex")
```
miparnisari, line 710: ?

tstirrat15: https://github.com/authzed/spicedb/actions/runs/25070206298/job/73448263566

I'm guessing this is related to protovalidate?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added support for YAML-based validation files in the Language Server (https://github.com/authzed/spicedb/pull/3024)
 - Enable statistics-based optimizations when `--experimental-query-plan` is enabled. (https://github.com/authzed/spicedb/pull/3052)
 - Added missing implementations of cursoring for LookupResource, LookupSubjects and ReadRelationships calls in FDW (https://github.com/authzed/spicedb/pull/3016)
+- Added debug behavior to LookupResources that can help identify data cyles in LR calls (https://github.com/authzed/spicedb/pull/3070)
```
```suggestion
- Support new `withDebug` flag in LookupResources calls to identify cycles (https://github.com/authzed/spicedb/pull/3070)
```

internal/services/v1/debug_test.go
```diff
@@ -921,3 +927,176 @@ func TestBulkCheckPermissionWithDebug(t *testing.T) {
...
+			}
+		}
+	}
+
+	req.NotEmpty(traceStr, "expected non-empty traversal trace in error details")
+
+	debugInfo := &dispatch.LookupDebugInfo{}
+	req.NoError(prototext.Unmarshal([]byte(traceStr), debugInfo), "trace must parse as LookupDebugTrace")
+	req.Len(debugInfo.CycleMembers, 1, "there should be one cycle member found at the final depth")
+	// NOTE: this value will change if the max recursion depth changes, since that determines
```
miparnisari, line 1012: 👍

internal/services/v1/permissions.go
```diff
@@ -625,12 +626,23 @@ func (ps *permissionServer) lookupResources3(req *v1.LookupResourcesRequest, res
...
+			EnableDebugTrace: req.WithDebug,
 		},
 		stream)
 	if err != nil {
+		if req.WithDebug && dispatchpkg.IsMaxDepthExceeded(err) {
+			if debugInfo := dispatchpkg.ExtractTraversalTrace(err); debugInfo != nil {
+				// We'll only attach debug info if cyclemembers is non-empty. Otherwise we're
+				// in a normal recursion-too-deep scenario and the debuginfo doesn't help.
+				if len(debugInfo.CycleMembers) > 0 {
+					logging.Warn().Strs("cycle-members", debugInfo.CycleMembers).Msg("The following resource/relation pairs were found in a cycle in LookupResources:")
```
miparnisari, line 641: nitpick: instead of creating a new log line can we not enrich the `finished call` log line with this information?

image

that log is very easy to get lost in the forest because it's not linked to a particular request id.

internal/graph/lookupresources2.go
miparnisari: i like the current approach using the set intersection.

miparnisari (dismissed):
LGTM but you gotta fix conflicts

internal/services/v1/permissions.go
tstirrat15: I don't think our logging is currently set up for that - we'd need to pass the event through the tree in a way that we don't currently do. It's why I opted to make it a `warn` instead of an `info`.

miparnisari: OK, in that case, i think i'd also add a span attribute or a span event to the ongoing span indicating that a cycle was found. But can be done in a followup.

miparnisari (approved):

PR #3071, opened 2026-04-23
fix: improve error when expiration is before caveat

### LINKED ISSUE #2995: Improve error messaging around expiration
If you do `relation: foo with expiration and some_caveat` and run `zed validate`, you get an error message saying "unexpected TokenTypeKeyword" or something to that effect. It'd be nice if we provided a bit more context, especially because `and expiration` is a special form and that isn't immediately apparent.

ivanauth:
## Summary
- Emit a clear error when users write `with expiration and <caveat>` instead of `with <caveat> and expiration`

Fixes #2995


josephschorr (dismissed):
LGTM

miparnisari (approved):

PR #3072, opened 2026-04-23
feat: introduce protobuf and gRPC for dispatching query plans

barakmich:
## Description

Step one in integrating Query Planner with dispatch (for distributed caching, mostly) is to stub out the implementation and provide the request structure. Implementing a dispatch-aware executor is next.

### DIFF
--- a/proto/internal/dispatch/v1/dispatch.proto
+++ b/proto/internal/dispatch/v1/dispatch.proto
@@ -251,4 +251,5 @@ message ResultPath {
   google.protobuf.Timestamp expiration = 8;
   repeated core.v1.RelationshipIntegrity integrity = 9;
   google.protobuf.Struct metadata = 10;
+  repeated ResultPath excluded_subjects = 11;
 }
--- a/proto/internal/dispatch/v1/dispatch.proto
+++ b/proto/internal/dispatch/v1/dispatch.proto
@@ -5,6 +5,7 @@ import "buf/validate/validate.proto";
 import "core/v1/core.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
 
 option go_package = "github.com/authzed/spicedb/pkg/proto/dispatch/v1";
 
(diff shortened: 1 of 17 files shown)


proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -210,3 +213,43 @@ message CheckDebugTrace {
...
   string source_id = 8;
 }
+
+enum PlanOperation {
+  PLAN_OPERATION_CHECK = 0;
+  PLAN_OPERATION_LOOKUP_RESOURCES = 1;
+  PLAN_OPERATION_LOOKUP_SUBJECTS = 2;
+}
+
+message PlanContext {
```
miparnisari, line 223: for LR/LS: no continuation token or cursor? or is that not needed?

proto/internal/dispatch/v1/dispatch.proto
```diff
@@ -210,3 +213,43 @@ message CheckDebugTrace {
...
+  PlanOperation operation = 1;
+  string canonical_key = 2;
+  core.v1.ObjectAndRelation resource = 3;
+  core.v1.ObjectAndRelation subject = 4;
+  PlanContext plan_context = 5;
+}
+
+message DispatchPlanResponse {
+  ResponseMeta metadata = 1;
+  repeated ResultPath paths = 2;
```
miparnisari, line 240: 1. why no godocs in any of the fields? this one in particular deserves a godoc. You even wrote it in the design doc :)
2. a nitpick: `DispatchPlanRequest` makes it sound that node A is asking node B to come up with the plan for the request. However, with the inclusion of `paths` in the response, now it sounds like node A is asking node B to _execute_ the plan. Please make this distinction clear. Maybe it should be `DispatchExecutionOfPlanRequest`? (i hate this name but hopefully you get the point i'm trying to make)

proto/internal/dispatch/v1/dispatch.proto
miparnisari: never mind - it's streaming

miparnisari (dismissed):

proto/internal/dispatch/v1/dispatch.proto
tstirrat15: I'm okay with the name. I agree it's a little awkward/potentially misleading, but it also follows the pattern laid out in this file, which is `Dispatch<some type><Request|Response>`

tstirrat15 (dismissed):
Looks good to me; see Maria's comments

proto/internal/dispatch/v1/dispatch.proto
tstirrat15: Hmm... I guess it could be `DispatchQueryPlanRequest` or `DispatchQueryPlannnerRequest`, at the risk of being a little verbose

tstirrat15 (dismissed):
LGTM

tstirrat15 (approved):
LGTM

PR #3079, opened 2026-04-29
feat: implement handling of dispatch and caching for DispatchQueryPlan

barakmich:


internal/dispatch/caching/caching.go
```diff
@@ -405,8 +405,57 @@ func (cd *Dispatcher) DispatchLookupSubjects(req *v1.DispatchLookupSubjectsReque
...
 
 func (cd *Dispatcher) DispatchQueryPlan(req *v1.DispatchQueryPlanRequest, stream dispatch.PlanStream) error {
-	// TODO: add caching logic
-	return cd.d.DispatchQueryPlan(req, stream)
+	switch req.Operation {
+	case v1.PlanOperation_PLAN_OPERATION_CHECK:
+		return cd.dispatchQueryPlanCheckCached(req, stream)
+	default:
+		// TODO: add caching for LookupResources and LookupSubjects
+		return cd.d.DispatchQueryPlan(req, stream)
```
tstirrat15, line 413: That'll be followup?

internal/dispatch/caching/caching.go
```diff
@@ -405,8 +405,57 @@ func (cd *Dispatcher) DispatchLookupSubjects(req *v1.DispatchLookupSubjectsReque
...
+	cd.checkTotalCounter.Inc()
+
+	requestKey, err := cd.keyHandler.PlanCheckCacheKey(stream.Context(), req)
+	if err != nil {
+		return err
+	}
+
+	if cachedResultRaw, found := cd.c.Get(requestKey); found {
+		var response v1.DispatchQueryPlanResponse
+		if err := response.UnmarshalVT(cachedResultRaw.([]byte)); err != nil {
```
tstirrat15: I know this is outside the scope of the PR, but it seems like we ought to be able to narrow this type somehow, either with generics or by modifying the interface so that we're always storing a byte array

internal/dispatch/caching/caching.go
```diff
@@ -405,8 +405,57 @@ func (cd *Dispatcher) DispatchLookupSubjects(req *v1.DispatchLookupSubjectsReque
...
+		return stream.Publish(&response)
+	}
+
+	// Cache miss — collect the streamed results to cache them.
+	collecting := dispatch.NewCollectingDispatchStream[*v1.DispatchQueryPlanResponse](stream.Context())
+	if err := cd.d.DispatchQueryPlan(req, collecting); err != nil {
+		return err
+	}
+
+	// Forward collected results to the caller and cache the first response (Check produces at most one).
```
tstirrat15: If check can produce at most one, why are we iterating?

internal/dispatch/graph/graph.go
```diff
@@ -471,12 +474,166 @@ func (ld *localDispatcher) DispatchLookupSubjects(
...
+	}
+}
+
+// findIteratorByCanonicalKey loads the schema at the given revision, compiles
+// all permissions, and returns the iterator subtree matching the canonical key.
+func (ld *localDispatcher) findIteratorByCanonicalKey(ctx context.Context, revision datastore.Revision, targetKey query.CanonicalKey) (query.Iterator, error) {
+	dl := datalayer.MustFromContext(ctx)
+	reader := dl.SnapshotReader(revision)
+
+	sr, err := reader.ReadSchema(ctx)
```
tstirrat15, line 585: The idea is that this should usually be hitting the cache?

internal/dispatch/graph/graph.go
```diff
@@ -471,12 +474,166 @@ func (ld *localDispatcher) DispatchLookupSubjects(
...
+	if err != nil {
+		return nil, fmt.Errorf("DispatchQueryPlan: failed to read schema: %w", err)
+	}
+
+	nsDefs, err := sr.ListAllTypeDefinitions(ctx)
```
tstirrat15, line 590: Is there a `ListAllDefinitions` method? Or is it that `BuildSchemaFromDefinitions` differentiates?

internal/dispatch/graph/graph.go
```diff
@@ -471,12 +474,166 @@ func (ld *localDispatcher) DispatchLookupSubjects(
...
+		datastore.DefinitionsOf(nsDefs),
+		datastore.DefinitionsOf(caveatDefs),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("DispatchQueryPlan: failed to build schema: %w", err)
+	}
+
+	for nsName, def := range fullSchema.Definitions() {
+		for permName := range def.Permissions() {
+			co, err := query.BuildOutlineFromSchema(fullSchema, nsName, permName)
```
tstirrat15, line 610: I'm curious about how much work this entails - if all of the outline optimizations are running that sounds like it could be heavy.

I also know that this is something that we intend to iterate past pretty quickly.

internal/dispatch/keys/hasher_common.go
```diff
@@ -99,3 +101,18 @@ type hashableContextString string
 func (hc hashableContextString) AppendToHash(hasher hasherInterface) {
 	hasher.WriteString(string(hc))
 }
+
+type hashableContext struct {
+	*structpb.Struct
+}
+
```
tstirrat15, line 109: I think I like this better than what I did in #3065. Is it worth refactoring the other locations?

internal/dispatch/remote/cluster.go
```diff
@@ -852,8 +853,46 @@ func (cr *clusterDispatcher) DispatchLookupSubjects(
...
+	case v1.PlanOperation_PLAN_OPERATION_LOOKUP_SUBJECTS:
+		requestKey, err = cr.keyHandler.PlanLookupSubjectsDispatchKey(stream.Context(), req)
+	default:
+		return fmt.Errorf("unknown plan operation: %v", req.Operation)
+	}
+	if err != nil {
+		return err
+	}
+
+	ctx := context.WithValue(stream.Context(), consistent.CtxKey, requestKey)
```
tstirrat15, line 874: `ctxkey` package worth using here? Or does this match the pattern in the file and we should leave it for now?

internal/services/integrationtesting/benchmark_dispatchqueryplan_test.go
tstirrat15, line 31: Lint failure?

pkg/query/queryopt/registry.go
```diff
@@ -57,31 +86,23 @@ type Optimizer struct {
...
+// synthesized nodes (ID==0) receive fresh IDs via query.FillMissingNodeIDs.
+// The returned CanonicalOutline is ready to compile.
+func ApplyOptimizations(co query.CanonicalOutline, opts []Optimizer, params RequestParams) (query.CanonicalOutline, error) {
 	// Sort by descending priority so higher-priority mutations run first.
-	slices.SortFunc(opts, func(a, b Optimizer) int {
+	sorted := make([]Optimizer, len(opts))
+	copy(sorted, opts)
+	slices.SortFunc(sorted, func(a, b Optimizer) int {
 		return cmp.Compare(b.Priority, a.Priority)
 	})
```
tstirrat15: https://pkg.go.dev/slices#Sorted

tstirrat15 (dismissed):
Nothing blocking, had some questions/comments

pkg/query/queryopt/alias_chain_collapse_test.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func TestOptimizersForRequest_IterSubjects_SpecificRelationExcludesCollapse(t *testing.T) {
+	// IterSubjects with a specific subject relation excludes alias-chain-collapse.
+	opts := OptimizersForRequest(RequestParams{
+		Operation:       query.OperationIterSubjects,
+		SubjectRelation: "owner",
+	})
+	require.False(t, hasOptimizer(opts, "alias-chain-collapse"))
+}
```
miparnisari: nit - could rewrite this as one, table-driven, test

internal/dispatch/keys/computed.go
```diff
@@ -113,3 +119,31 @@ func lookupSubjectsRequestToKey(req *v1.DispatchLookupSubjectsRequest, option di
...
 	)
 }
+
+// planCheckRequestToKey converts a plan check request into a cache key
+func planCheckRequestToKey(req *v1.DispatchQueryPlanRequest, option dispatchCacheKeyHashComputeOption) DispatchCacheKey {
+	return dispatchCacheKeyHash(planCheckPrefix, req.PlanContext.Revision, option,
+		hashableString(req.CanonicalKey),
+		hashableOnr{req.Resource},
+		hashableOnr{req.Subject},
+		hashableContext{Struct: req.PlanContext.CaveatContext},
```
miparnisari, line 129: why was adding `hashableContext` necessary? here and below can we not do 

```go
	stableContextString, err := caveats.StableContextStringForHashing(req.PlanContext.CaveatContext)
	if err != nil {
		return DispatchCacheKey{}, err
	}
	//...
	hashableContextString(stableContextString)
```

?

tstirrat15: I'd also be okay with reworking our other usages to use `hashableContext`

internal/dispatch/caching/caching.go
```diff
@@ -405,8 +405,57 @@ func (cd *Dispatcher) DispatchLookupSubjects(req *v1.DispatchLookupSubjectsReque
...
+
+func (cd *Dispatcher) dispatchQueryPlanCheckCached(req *v1.DispatchQueryPlanRequest, stream dispatch.PlanStream) error {
+	cd.checkTotalCounter.Inc()
+
+	requestKey, err := cd.keyHandler.PlanCheckCacheKey(stream.Context(), req)
+	if err != nil {
+		return err
+	}
+
+	if cachedResultRaw, found := cd.c.Get(requestKey); found {
```
miparnisari: Nit: I was comparing your implementation with the original one and i see a comment on line 158:
```
// Disable caching when debugging is enabled.
```

i assume this comment is wrong; my understanding is that having debugging on does NOT turn off caching. Can you please remove that comment?

internal/dispatch/caching/caching.go
barakmich: Yep, an exercise for the future. Not impossible; but I'm not convinced that it's as important for the moment

internal/dispatch/caching/caching.go
barakmich: Yeah; this is just using the results verbatim. I think there's a lot of history in the caching layer, for exactly this reason

internal/dispatch/caching/caching.go
barakmich: That comment is for DispatchCheck, which does look at the flag?

internal/dispatch/caching/caching.go
barakmich: Fixed!

internal/dispatch/graph/graph.go
barakmich: Yep; it's why the schema storage PR (#2924) is just as useful for QP as default

internal/dispatch/graph/graph.go
barakmich: Yeah, BuildSchemaFromDefinitions is trying to take as simple an interface (just the list of top-level things in the schema) so this is one way to do it. This could be a helper in schema/v2 though (using the new SchemaReader, which didn't exist when schema/v2 was written)

internal/dispatch/graph/graph.go
barakmich: Yeah, once we can cache schemas (especially v2 schemas) we can cache outlines as well. It's not too crazy -- it's just a local memory cache and goes along in the query plan metadata field, but it's a followup task (which I'll make a ticket for)

internal/dispatch/keys/hasher_common.go
barakmich: Ah, I see. It's maybe worth another ticket for cleanup

internal/dispatch/remote/cluster.go
barakmich: Yep, matches the pattern here, so leaving for now

tstirrat15 (approved):
LGMT

Issue #422, opened 2022-02-17
Report metrics for kuberesolver


ecordell:
It's important for an operator to know that spicedb is continually receiving updated peer lists. Emit metrics that can be used to alert if the list is not updated expediently.

This may require a contribution to the upstream kuberesolver or forking the project.

lalalalatt:
@ecordell I have make a patch for this issue to the upstream [kuberesolver](https://github.com/sercand/kuberesolver/): [kuberesolver#52](https://github.com/sercand/kuberesolver/pull/52)

ecordell:
Awesome! We'll pull it in and test it once merged.

tstirrat15:
Update on this: looks like the upstream has merged but we haven't taken the version yet. There's a breaking change around `Endpoint`s vs `EndpointSlice`s that we'll need to sort out.

PR #3081, opened 2026-04-30
chore: update kuberesolver

tstirrat15:
Part of #422 

## Description
This brings in the version of kuberesolver that publishes the associated metrics. The breaking change is around the use of `Endpoint`s vs `EndpointSlice`s.

## Changes
* Bump kuberesolver version
## Testing
Review.


tstirrat15:
Putting this on hold because there's a bug in v6: https://github.com/sercand/kuberesolver/issues/66

tstirrat15:
Gonna close this; it seems like we're unlikely to take that version, and it seems unlikely that a fix for it will go in anytime soon. We may end up vendoring that code and fixing it here.

PR #3083, opened 2026-05-01
test: prevent flaky TestCertWatcher

miparnisari:
## Description

```
--- FAIL: TestCertWatcherStart (12.99s)
    --- FAIL: TestCertWatcherStart/should_reload_currentCert_after_move_out (10.31s)
        certwatcher_test.go:182: 
            	Error Trace:	/home/runner/work/spicedb/spicedb/pkg/x509util/certwatcher_test.go:189
            	            				/opt/hostedtoolcache/go/1.26.2/arm64/src/runtime/asm_arm64.s:1447
            	Error:      	Should be false
            	Messages:   	first and second privatekeys should not be equal
        certwatcher_test.go:182: 
            	Error Trace:	/home/runner/work/spicedb/spicedb/pkg/x509util/certwatcher_test.go:190
            	            				/opt/hostedtoolcache/go/1.26.2/arm64/src/runtime/asm_arm64.s:1447
            	Error:      	Should not be zero, but was 0
            	Messages:   	the serial numbers should compare as different
        certwatcher_test.go:182: 
            	Error Trace:	/home/runner/work/spicedb/spicedb/pkg/x509util/certwatcher_test.go:182
            	Error:      	Condition never satisfied
            	Test:       	TestCertWatcherStart/should_reload_currentCert_after_move_out
FAIL
```

`setupWatcher` launched `Start()` in a goroutine and immediately let the test mutate the watched files. Under CPU contention, the `Start` goroutine could be delayed past the test's   `os.Rename(certPath, …) call `. So when `Start` finally ran `cw.watcher.Add(certPath)`, the file no longer existed at that path, `Add` returned ENOENT, and `Start` exited. The polling ticker  was never created, the cert was never reloaded, and `EventuallyWithT` timed out.

The bug was not in the certwatcher's reload logic — it was in the test's failure to wait for the watcher to be ready before changing the files it was supposed to watch. The `Start()` error was also being swallowed (`_ = ...`), which hid the real cause.

## Testing

```
[17:23:49] ~/Documents/GitHub/spicedb (flakiness-certwatcher-2) $ go test -c -race ./pkg/x509util/
[17:27:25] ~/Documents/GitHub/spicedb (flakiness-certwatcher-2) $ stress ./x509util.test -test.run TestCertWatcherStart/should_reload_currentCert_after_move_out
5s: 30 runs so far, 0 failures, 16 active
10s: 64 runs so far, 0 failures, 16 active
15s: 96 runs so far, 0 failures, 16 active
20s: 134 runs so far, 0 failures, 16 active
25s: 176 runs so far, 0 failures, 16 active
```

### DIFF
--- a/pkg/x509util/certwatcher_test.go
+++ b/pkg/x509util/certwatcher_test.go
@@ -57,7 +57,7 @@ func TestCertWatcherSequentialMetricRegistration(t *testing.T) {
 	require.NoError(t, err)
 
 	go func() {
-		_ = watcher1.Start(ctx1)
+		watcher1.Start(ctx1)
 	}()
 	cancel1()
 
@@ -67,7 +67,7 @@ func TestCertWatcherSequentialMetricRegistration(t *testing.T) {
 	require.NoError(t, err)
 
 	go func() {
-		_ = watcher2.Start(ctx2)
+		watcher2.Start(ctx2)
 	}()
 	cancel2()
 }
@@ -91,8 +91,10 @@ func setupWatcher(t *testing.T, ip string) (certPath, keyPath string, watcher *C
 
 	startWatcher = func(interval time.Duration) {
 		go func() {
-			_ = watcher.WithWatchInterval(interval).Start(t.Context())
+			watcher.WithWatchInterval(interval).Start(t.Context())
 		}()
+		err := <-watcher.Started()
+		assert.NoError(t, err)
 		assert.NoError(t, watcher.ReadCertificate())
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/x509util/certwatcher_test.go
```diff
@@ -91,8 +91,10 @@ func setupWatcher(t *testing.T, ip string) (certPath, keyPath string, watcher *C
 
 	startWatcher = func(interval time.Duration) {
 		go func() {
-			_ = watcher.WithWatchInterval(interval).Start(t.Context())
+			watcher.WithWatchInterval(interval).Start(t.Context())
 		}()
+		err := <-watcher.Started()
+		require.NoError(t, err)
```
tstirrat15: This can't be a `require` inside a goroutine. It should be an `if !assert.NoError { return }`

tstirrat15 (commented):
The overall shape of things makes sense but it needs to use `assert`

tstirrat15 (approved):
LGTM

PR #3086, opened 2026-05-04
fix: enable OTel auto-detection via standard OTEL_ env vars, deprecate --otel-provider (fixes #712)

### LINKED ISSUE #712: Configure Opentelemetry solely through OTEL_ flags
## Overview
I went to get opentelemetry tracing working with our infrastructure. We run an otel sidecar in an ECS context, so the opentelemetry receiver is at `http://localhost:4317` and uses the grpc mechanism.

This was what was necessary to get things working:

```
    { "name" : "SPICEDB_OTEL_PROVIDER", "value" : "otlpgrpc" },
    { "name" : "SPICEDB_OTEL_ENDPOINT", "value" : "localhost:4317" },
    { "name" : "OTEL_EXPORTER_OTLP_ENDPOINT", "value" : "http://localhost:4317" },
```

Specifying the provider makes sense to me; needing to specify both the `SPICEDB_` and `OTEL_` versions of the endpoint doesn't. I tested with both individually and neither was sufficient by itself.

Doing this configuration solely through `OTEL_`-flagged environment variables would make sense to me, as would having maybe a single flag that says "otel tracing is on."

I'm also game for other approaches/solutions; this was just the one that was discussed in discord.

## Other background
Discord conversation around this here: https://discord.com/channels/844600078504951838/844600078948630559/1000124099043016814

Jdepp007004:
enables tracing when only OTEL_ env vars are set
added pre-run hook to activate provider if env config exists
preserved existing flag behavior (flags override env)
respected explicit --otel-provider=none
deprecated --otel-provider flag
added tests covering env-only, flags-only, and precedence behavior

### DIFF
--- a/pkg/cmd/serve.go
+++ b/pkg/cmd/serve.go
@@ -209,6 +209,11 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	// NOTE: cobraotel.New takes service name as an arg rather than command name.
 	otel := cobraotel.New("spicedb")
 	otel.RegisterFlags(tracingFlags)
+	// Deprecate --otel-provider: the provider is now inferred automatically
+	// from standard OTEL_* environment variables when not explicitly set.
+	if err := tracingFlags.MarkDeprecated("otel-provider", "provider is inferred automatically from configuration; this flag will be removed in a future release"); err != nil {
+		return fmt.Errorf("failed to mark otel-provider as deprecated: %w", err)
+	}
 
 	loggingFlagSet := nfs.FlagSet(BoldBlue("Logging"))
 	loggingFlagSet.BoolVar(&config.EnableRequestLogs, "grpc-log-requests-enabled", false, "enable logging of API request payloads")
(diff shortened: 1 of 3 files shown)


tstirrat15:
I'm digging more into this and I don't think this is the approach we want to take.

I think [cobraotel](https://github.com/jzelinskie/cobrautil/blob/main/cobraotel/cobraotel.go), the package we use to do this configuration, is intended to abstract over the configuration of the `OTEL_` flags, such that you don't need to use those flags.

I think there may be a bug in `cobraotel` where it's not picking up or propagating those flags correctly. It seems as though the `SPICEDB_` flags should all work, and if they're set, they should override the `OTEL_` flags. It seems that it should also be possible to configure tracing through the `OTEL_` flags as long as you have `SPICEDB_OTEL_PROVIDER` set, and only that set.

My sense is that most of the work would be in `cobraotel` - writing some tests to ensure that all of the flags are correctly picked up and respected. I think the `PROVIDER` is probably necessary, because I'm not sure there's a way to set up a provider based solely on the presence or absence of the flags, at least not without hacking around things a bit.

Does all of that make sense?

Jdepp007004:
That's does make a lot of sense, thanks for the clarification.

I shall fork the upstream repo and write some detailed tests to find out the exact bug and share the findings so that we can decide upon a direction in which we may proceed.

PR #3091, opened 2026-05-05
fix: add benchmark fixes after schema cache, applying optimizations

barakmich:
## Description
Schema cache -- writing the legacy way and the new way (and reading from legacy) broke the benchmark integration tests. This remedies that. It also fixes the fact that most of the benchmarks weren't applying optimizations, so now we are.

## Testing

These are the affected tests. 

## References

Each commit in this PR is relatively stand-alone, for reviewers

### DIFF
--- a/internal/services/integrationtesting/benchmark_dispatchqueryplan_test.go
+++ b/internal/services/integrationtesting/benchmark_dispatchqueryplan_test.go
@@ -63,21 +63,68 @@ func newDispatchQueryPlanHandle(b *testing.B, fileName string) *dispatchQueryPla
 	return &dispatchQueryPlanHandle{ds: ds, revision: rev, schema: fullSchema, schemaHash: schemaHash}
 }
 
-// compileIterator builds and compiles an iterator for the given resource type and permission.
-func (h *dispatchQueryPlanHandle) compileIterator(b *testing.B, resourceType, permission, subjectType, subjectRelation string) query.Iterator {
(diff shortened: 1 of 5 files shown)


internal/services/integrationtesting/benchmark_dispatchqueryplan_test.go
```diff
@@ -265,16 +312,38 @@ func (d *localQueryPlanDispatcher) DispatchQueryPlan(req *v1.DispatchQueryPlanRe
...
+	// Map PlanOperation to query.Operation for optimizer selection.
+	var operation query.Operation
+	switch req.Operation {
+	case v1.PlanOperation_PLAN_OPERATION_CHECK:
+		operation = query.OperationCheck
+	case v1.PlanOperation_PLAN_OPERATION_LOOKUP_RESOURCES:
+		operation = query.OperationIterResources
+	case v1.PlanOperation_PLAN_OPERATION_LOOKUP_SUBJECTS:
+		operation = query.OperationIterSubjects
+	}
```
tstirrat15, line 326: This seems like it could be either a helper function or a map

tstirrat15 (approved):
Makes sense to me

internal/services/integrationtesting/benchmark_dispatchqueryplan_test.go
barakmich: I'll add that in my rebased followup 🙂

PR #3093, opened 2026-05-06
fix: wire DispatchExecutor into the GRPC path properly and fix correctness

barakmich:
## Description

This PR uses DispatchExecutor to actually send subproblems on Alias boundaries. This then showed up the final minor bugs in correctness in the integration tests, which were then eliminated, via adding the necessary fields to the proto PlanContext

This also introduces batching operations to the executors which allow the iterators to make choices about when to batch partial results (such as in an arrow, pulling from the left and sending them all to the right) 

## Testing

Integration Testing discovered much of this, notably `go test -v -tags integration ./internal/services/integrationtesting/queryconsistency -run="GRPC/"`

### DIFF
--- a/internal/dispatch/executor.go
+++ b/internal/dispatch/executor.go
@@ -6,6 +6,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"github.com/authzed/spicedb/internal/caveats"
 	"github.com/authzed/spicedb/pkg/datalayer"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	v1 "github.com/authzed/spicedb/pkg/proto/dispatch/v1"
@@ -242,6 +243,36 @@ func toProtoStruct(m map[string]any) (*structpb.Struct, error) {
 		return nil, nil
 	}
 	return structpb.NewStruct(m)
+}
+
+// NewQueryContext builds a query.Context whose Executor is a DispatchExecutor
+// driven by the given PlanContext, so that dispatched sub-requests carry the
+// same plan state. Plan-derived options (caveat context, recursion depth,
+// datastore limit) are applied first; extra opts are appended after.
+func NewQueryContext(
+	stdContext context.Context,
+	dispatcher Dispatcher,
+	planContext *v1.PlanContext,
+	reader query.QueryDatastoreReader,
+	caveatRunner *caveats.CaveatRunner,
+	opts ...query.ContextOption,
+) *query.Context {
+	executor := NewDispatchExecutor(dispatcher, planContext)
+
+	baseOpts := []query.ContextOption{
+		query.WithReader(reader),
+		query.WithCaveatContext(CaveatContextFromPlanContext(planContext)),
+		query.WithCaveatRunner(caveatRunner),
+	}
+	if d := planContext.GetMaxRecursionDepth(); d > 0 {
+		baseOpts = append(baseOpts, query.WithMaxRecursionDepth(int(d)))
+	}
+	if l := planContext.GetOptionalDatastoreLimit(); l > 0 {
+		baseOpts = append(baseOpts, query.WithPaginationLimit(l))
+	}
+	baseOpts = append(baseOpts, opts...)
+
+	return query.NewQueryContext(stdContext, executor, baseOpts...)
 }
 
 // NewPlanContext builds a PlanContext proto from query.Context fields.
--- a/internal/dispatch/executor.go
+++ b/internal/dispatch/executor.go
@@ -131,9 +131,14 @@ func (e *DispatchExecutor) dispatchIterSubjects(ctx *query.Context, it query.Ite
 	subCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	// For LookupSubjects the proto Subject field carries the *filter* subject
+	// type+relation rather than a real subject ONR — the receiver needs the
+	// filter to apply the same reachability/optimizer decisions the sender
+	// would have applied. The actual receiver-side iteration uses NoObjectFilter
+	// since the sender filters the returned paths itself.
 	req := e.buildRequest(v1.PlanOperation_PLAN_OPERATION_LOOKUP_SUBJECTS, it, resource, query.ObjectAndRelation{
-		ObjectType: resource.ObjectType,
-		ObjectID:   resource.ObjectID,
+		ObjectType: filterSubjectType.Type,
+		Relation:   filterSubjectType.Subrelation,
 	})
 	stream := NewCollectingDispatchStream[*v1.DispatchQueryPlanResponse](subCtx)
 	if err := e.dispatcher.DispatchQueryPlan(req, stream); err != nil {
@@ -162,9 +167,14 @@ func (e *DispatchExecutor) dispatchIterResources(ctx *query.Context, it query.It
 }
 
 func (e *DispatchExecutor) buildRequest(op v1.PlanOperation, it query.Iterator, resource query.Object, subject query.ObjectAndRelation) *v1.DispatchQueryPlanRequest {
+	// dispatch{Check,IterSubjects,IterResources} only enter buildRequest when
+	// it is a *query.AliasIterator, so the type assertion is safe. The dispatch
+	// boundary is always the (definition, relation) the alias represents; we
+	// encode that as "definition#relation" in the canonical_key field.
+	alias := it.(*query.AliasIterator)
 	return &v1.DispatchQueryPlanRequest{
 		Operation:    op,
-		CanonicalKey: string(it.CanonicalKey()),
+		CanonicalKey: alias.DefinitionName() + "#" + alias.Relation(),
 		Resource: &core.ObjectAndRelation{
 			Namespace: resource.ObjectType,
 			ObjectId:  resource.ObjectID,
--- a/internal/dispatch/executor.go
+++ b/internal/dispatch/executor.go
@@ -177,7 +177,7 @@ func (e *DispatchExecutor) dispatchCheck(ctx *query.Context, it query.Iterator, 
 	subCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	req := e.buildRequest(v1.PlanOperation_PLAN_OPERATION_CHECK, it, resource, subject)
+	req := e.buildRequest(ctx, v1.PlanOperation_PLAN_OPERATION_CHECK, it, resource, subject)
 	stream := NewCollectingDispatchStream[*v1.DispatchQueryPlanResponse](subCtx)
 	if err := e.dispatcher.DispatchQueryPlan(req, stream); err != nil {
 		return nil, err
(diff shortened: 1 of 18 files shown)


internal/dispatch/graph/graph.go
```diff
@@ -592,15 +591,64 @@ func (ld *localDispatcher) DispatchQueryPlan(
 		}
 		return nil
 
+	case v1.PlanOperation_PLAN_OPERATION_CHECK_MANY_RESOURCES:
+		// Iterate CheckImpl directly: dispatch boundary already crossed for this
+		// alias. qctx still uses DispatchExecutor, so nested aliases re-dispatch.
+		for _, m := range req.Many {
+			res := query.Object{ObjectType: m.Namespace, ObjectID: m.ObjectId}
```
tstirrat15: Nit: what kind of thing is `m`? Can the name reflect that? Is it a path or an object or?

internal/dispatch/graph/graph.go
```diff
@@ -592,15 +591,64 @@ func (ld *localDispatcher) DispatchQueryPlan(
...
-// the iterator subtree matching the canonical key.
-func (ld *localDispatcher) findIteratorByCanonicalKey(ctx context.Context, revision datastore.Revision, schemaHash datalayer.SchemaHash, targetKey query.CanonicalKey, queryParams queryopt.RequestParams) (query.Iterator, error) {
+// compileIteratorForDispatchKey loads the schema at the given revision, parses
+// the dispatch key (currently encoded as "definition#relation"), builds the
+// outline standalone for that pair, applies the same optimizers the sender
+// would have used, and compiles to an iterator. The returned iterator's outer
+// node is the *AliasIterator for the requested (definition, relation); callers
+// invoke its Impl method directly so the alias's rewrite + self-edge logic
+// runs without re-triggering the executor's dispatch decision at this level.
+func (ld *localDispatcher) compileIteratorForDispatchKey(ctx context.Context, revision datastore.Revision, schemaHash datalayer.SchemaHash, req *v1.DispatchQueryPlanRequest) (query.Iterator, error) {
```
tstirrat15, line 645: Is this likely to be the function whose results get cached?

internal/dispatch/graph/graph.go
```diff
@@ -627,41 +675,43 @@ func (ld *localDispatcher) findIteratorByCanonicalKey(ctx context.Context, revis
...
+	case v1.PlanOperation_PLAN_OPERATION_CHECK,
+		v1.PlanOperation_PLAN_OPERATION_CHECK_MANY_RESOURCES,
+		v1.PlanOperation_PLAN_OPERATION_CHECK_MANY_SUBJECTS:
+		return query.OperationCheck
+	case v1.PlanOperation_PLAN_OPERATION_LOOKUP_RESOURCES:
+		return query.OperationIterResources
+	case v1.PlanOperation_PLAN_OPERATION_LOOKUP_SUBJECTS:
+		return query.OperationIterSubjects
+	default:
+		return query.OperationCheck
```
tstirrat15: What would hit this case? Is this something we want to be strict about, or does it make sense to allow fallthrough?

internal/dispatch/executor.go
```diff
@@ -130,9 +196,14 @@ func (e *DispatchExecutor) dispatchIterSubjects(ctx *query.Context, it query.Ite
...
 	defer cancel()
 
-	req := e.buildRequest(v1.PlanOperation_PLAN_OPERATION_LOOKUP_SUBJECTS, it, resource, query.ObjectAndRelation{
-		ObjectType: resource.ObjectType,
-		ObjectID:   resource.ObjectID,
+	// For LookupSubjects the proto Subject field carries the *filter* subject
+	// type+relation rather than a real subject ONR — the receiver needs the
+	// filter to apply the same reachability/optimizer decisions the sender
+	// would have applied. The actual receiver-side iteration uses NoObjectFilter
+	// since the sender filters the returned paths itself.
```
tstirrat15, line 204: I wanna ask you about this but I'll do that out-of-band

internal/dispatch/executor.go
```diff
@@ -160,10 +231,121 @@ func (e *DispatchExecutor) dispatchIterResources(ctx *query.Context, it query.It
...
 }
 
-func (e *DispatchExecutor) buildRequest(op v1.PlanOperation, it query.Iterator, resource query.Object, subject query.ObjectAndRelation) *v1.DispatchQueryPlanRequest {
+func (e *DispatchExecutor) dispatchCheckManySubjects(ctx *query.Context, it query.Iterator, resource query.Object, subjects []query.ObjectAndRelation) ([]*query.Path, error) {
+	subCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	out := make([]*query.Path, len(subjects))
+	chunk := int(e.dispatchChunkSize)
+	for start := 0; start < len(subjects); start += chunk {
```
tstirrat15: Would `slices.Chunk` help here?

internal/dispatch/executor.go
```diff
@@ -160,10 +231,121 @@ func (e *DispatchExecutor) dispatchIterResources(ctx *query.Context, it query.It
...
+	return out, nil
+}
+
+func (e *DispatchExecutor) dispatchCheckManyResources(ctx *query.Context, it query.Iterator, resources []query.Object, subject query.ObjectAndRelation) ([]*query.Path, error) {
+	subCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	out := make([]*query.Path, len(resources))
+	chunk := int(e.dispatchChunkSize)
+	for start := 0; start < len(resources); start += chunk {
```
tstirrat15: Same here

internal/dispatch/executor.go
```diff
@@ -160,10 +231,121 @@ func (e *DispatchExecutor) dispatchIterResources(ctx *query.Context, it query.It
...
+	if len(manySubjects) > 0 {
+		req.Many = make([]*core.ObjectAndRelation, len(manySubjects))
+		for i, s := range manySubjects {
+			req.Many[i] = &core.ObjectAndRelation{
+				Namespace: s.ObjectType,
+				ObjectId:  s.ObjectID,
+				Relation:  s.Relation,
+			}
+		}
+	}
```
tstirrat15, line 330: I take it this is where we're wishing for a sum type in go?

internal/dispatch/executor.go
```diff
@@ -173,7 +355,56 @@ func (e *DispatchExecutor) buildRequest(op v1.PlanOperation, it query.Iterator,
...
+	if tlo == v1.PlanOperation_PLAN_OPERATION_CHECK {
+		tlo = queryOpToPlanOperation(topLevelOp)
+	}
+	return &v1.PlanContext{
+		Revision:               pc.Revision,
+		CaveatContext:          pc.CaveatContext,
+		MaxRecursionDepth:      pc.MaxRecursionDepth,
+		OptionalDatastoreLimit: pc.OptionalDatastoreLimit,
+		SchemaHash:             pc.SchemaHash,
+		InProgressKeys:         append(append([]string(nil), pc.InProgressKeys...), key),
```
tstirrat15: This syntax bent my brain a bit. I wish go had a better way of expressing this.

pkg/query/arrow.go
```diff
@@ -57,11 +57,19 @@ func (a *ArrowIterator) CheckImpl(ctx *Context, resource Object, subject ObjectA
...
 	//
-	// But for now, we cover the first two.
+	// But for now, we cover the first two. When BatchedArrows is set the
+	// per-element Check loop is replaced by a single CheckMany call so the
+	// executor (e.g. DispatchExecutor) can collapse fanout into one RPC.
 	switch a.direction {
 	case leftToRight:
+		if ctx.BatchedArrows {
+			return a.checkLeftToRightBatch(ctx, resource, subject)
+		}
```
tstirrat15, line 67: Would it eventually make sense to make the `Check` interface a list/sequence of things to check rather than a single thing, and then a single check is just a special case?

Not asking for that refactor if that is indeed the case, just wanting to check my mental model.

pkg/query/alias.go
```diff
@@ -32,6 +33,18 @@ func NewAliasIterator(relation string, subIt Iterator) *AliasIterator {
 	}
 }
 
+// NewAliasIteratorWithDefinition creates a new Alias iterator that also carries
+// the schema definition name to which the alias belongs. The definition name is
+// used by the dispatch layer to identify the (definition, relation) boundary
+// represented by this alias.
+func NewAliasIteratorWithDefinition(definitionName, relation string, subIt Iterator) *AliasIterator {
```
miparnisari: nit - this is never used?

tstirrat15 (dismissed):
LGTM

internal/dispatch/executor_test.go
```diff
@@ -634,3 +634,228 @@ func TestPaginationLimitFromPlanContext_Zero(t *testing.T) {
...
+		Subject:  query.ObjectAndRelation{ObjectType: "user", ObjectID: "alice", Relation: "..."},
+	})
+
+	resources := []query.Object{
+		{ObjectType: "document", ObjectID: "doc1"},
+		{ObjectType: "document", ObjectID: "missing"},
+	}
+	out, err := exec.CheckManyResources(ctx, fixed, resources, query.ObjectAndRelation{ObjectType: "user", ObjectID: "alice", Relation: "..."})
+	require.NoError(t, err)
+	require.Len(t, out, 2)
```
miparnisari: `require.Len(t, out, len(resources))` ?

internal/dispatch/executor_test.go
```diff
@@ -634,3 +634,228 @@ func TestPaginationLimitFromPlanContext_Zero(t *testing.T) {
...
+	require.NotNil(t, out[0])
+	require.Nil(t, out[1])
+	require.Empty(t, td.planCalls)
+}
+
+func TestDispatchExecutor_CheckManyResources_AliasDispatchesAndPairsByONR(t *testing.T) {
+	// Receiver returns paths in arbitrary order; sender must pair to inputs by
```
miparnisari, line 665: nit: who are the receiver and the sender in this test? might want to say `td` returns paths... and `exec` must pair. Or rename `td` to sender and `exec` to receiver

pkg/query/arrow_batched_test.go
```diff
@@ -0,0 +1,225 @@
...
+			if expectedSubjectID != "" {
+				req.Equal(expectedSubjectID, path.Subject.ObjectID)
+			}
+		})
+	}
+}
+
+func TestArrow_BatchedVsUnbatched_LTR(t *testing.T) {
+	factory := func() *ArrowIterator {
+		left := NewFolderHierarchyFixedIterator()
```
miparnisari, line 46: can we add wildcards to the data set? i wanna test lines 242-260 in `arrow.go`

internal/dispatch/executor.go
miparnisari: this looks like a bug fix

pkg/query/alias.go
barakmich: merged to enforce

internal/dispatch/graph/graph.go
barakmich: fixed

internal/dispatch/executor.go
barakmich: multi-lined it for clarity

pkg/query/arrow.go
barakmich: Potentially. That might be a refactor for the future if it helps (and indeed it might)

internal/dispatch/graph/graph.go
```diff
@@ -498,41 +501,25 @@ func (ld *localDispatcher) DispatchQueryPlan(
 
 	schemaHash := datalayer.SchemaHash(planCtx.GetSchemaHash())
 
-	// Load schema at the requested revision.
+	// Load schema at the requested revision and rebuild the iterator subtree
+	// for the (definition, relation) pair encoded in CanonicalKey.
 	// TODO: use cached compiled plans (Phase 7) instead of recompiling each time.
```
miparnisari: "phase 7"??

tstirrat15 (approved):
LGTM

PR #3096, opened 2026-05-07
refactor: wire QueryPlanMetadata through dispatch as well

barakmich:
## Description

DispatchQueryPlan rebuilds the query on the receiving side, but does not collect statistics while doing so. It should share the statistics with the root server metadata. This refactor moves the QueryPlanMetadata into pkg/query to avoid import cycles and then uses the same instantiated struct throughout the server layer.


internal/testserver/cluster.go
```diff
@@ -159,9 +160,16 @@ func TestClusterWithDispatch(t testing.TB, size uint, ds datastore.Datastore, ad
 	cancelFuncs := make([]func(), 0, size)
 
 	for i := range size {
+		// One QueryPlanMetadata per node, shared between the node's dispatcher
+		// (so receiver-side DispatchQueryPlan augments + advises off it) and the
+		// permissions service (so the sender's advisor and the receiver's are
+		// consulting the same store).
+		queryPlanMetadata := query.NewQueryPlanMetadata()
```
tstirrat15, line 167: I was a little surprised to see that nothing was crossing a network boundary here.

Is the idea that both the sender and receiver of a dispatch are going to make the same updates to the stats, but each one has their own copy of the stats?

Would that also mean that 1. if a dispatch request ends up going from a node to itself, the same stats object gets updated twice and 2. if a particular dispatch tree stays on only a subset of the nodes, it'll be only that subset that sees those stats?

I don't think either of these are necessarily a problem, but I think it could result in some interesting/subtle behavior and I wanna make sure my mental model is correct.

tstirrat15 (approved):
See comment

internal/testserver/cluster.go
barakmich: Yes to both (1) and (2). The stats aren't shared, yet. There are other designs on how we get statistics, like long queries to the datastore, but for now if they're in memory then, aside from gossiping or whatever, they're pretty much locked to a node

Issue #24, opened 2021-08-20
Add demo command


jzelinskie:
A CLI command that runs a demo of SpiceDB with a schema and data already that generates load would be useful for both load testing and showing users what using the database is like.

This is inspired by the same functionality in [CockroachDB](https://www.cockroachlabs.com/docs/stable/movr).

ztheile:
Is this still open for contributing?

PR #3100, opened 2026-05-07
feat: add demo command

paegun:
## Description

Add a demo command to the spicedb cli.

When spicedb is executed with the demo command and a sample dataset name, a test server is launched and the same dataset is loaded.

Sample datasets harvested from https://github.com/authzed/examples/blob/main/schemas .

## Testing

Manually ran each demo sample dataset and used zed to connect and permission check for:
[X] subject granted permission to resource
[X] subject is not granted permission to resource
[X] subject does not exist
[X] resource does not exist
[X] permission does not exist

For example:
```
spicedb demo --dataset github
```

```
zed permission check repository:authzed_go admin user:jimmy
```

## References

#24


paegun:

PR #3106, opened 2026-05-08
feat: inject prometheus registry

paegun:
## Description

Throughout, use of prometheus.Register and prometheus.Unregister lead to test isolation being compromised. This change makes the Prometheus Registerer an injectable.

As expected w/i #2829 , promauto use also needed to be removed.

I did not revert https://github.com/authzed/spicedb/pull/2957 which worked around the test isolation issue.

## Testing
[X] Execute existing unit tests.
[X] Manually test as follows:
  1.  run the test server
  2. load a dataset
  3. query the dataset
  4. check metrics.

## References

#2829


internal/datastore/crdb/pool/balancer.go
```diff
@@ -68,7 +64,18 @@ type NodeConnectionBalancer struct {
...
 // NewNodeConnectionBalancer builds a new nodeConnectionBalancer for a given connection pool and health tracker.
-func NewNodeConnectionBalancer(pool *RetryPool, healthTracker *NodeHealthTracker, interval time.Duration) *NodeConnectionBalancer {
+func NewNodeConnectionBalancer(pool *RetryPool, healthTracker *NodeHealthTracker, interval time.Duration, registerer prometheus.Registerer) *NodeConnectionBalancer {
+	if registerer == nil {
+		registerer = prometheus.DefaultRegisterer
+	}
+	for _, c := range []prometheus.Collector{connectionsPerCRDBNodeCountGauge, pruningTimeHistogram} {
+		if err := registerer.Register(c); err != nil {
+			var alreadyRegistered prometheus.AlreadyRegisteredError
+			if !errors.As(err, &alreadyRegistered) {
```
tstirrat15: Prefer `AsType` here

internal/datastore/crdb/pool/balancer.go
```diff
@@ -68,7 +64,18 @@ type NodeConnectionBalancer struct {
 }
 
 // NewNodeConnectionBalancer builds a new nodeConnectionBalancer for a given connection pool and health tracker.
-func NewNodeConnectionBalancer(pool *RetryPool, healthTracker *NodeHealthTracker, interval time.Duration) *NodeConnectionBalancer {
+func NewNodeConnectionBalancer(pool *RetryPool, healthTracker *NodeHealthTracker, interval time.Duration, registerer prometheus.Registerer) *NodeConnectionBalancer {
+	if registerer == nil {
+		registerer = prometheus.DefaultRegisterer
+	}
+	for _, c := range []prometheus.Collector{connectionsPerCRDBNodeCountGauge, pruningTimeHistogram} {
```
tstirrat15: It seems like this logic could be wrapped with a function like:

```go
func util.RegisterPrometheusMetrics(registerer prometheus.Registerer, errMessage string, metrics... prometheus.Metric) error {}
```

So that we can cut down on the duplication a bit.

Could also have it return the list of metrics for unregistration, or else have it return a closure that you can call in a `Close` function that does all of the unregistration for you.

Maybe package all that up in a struct? I'm thinking out loud here; take whatever parts of it make sense to you and work in the context.

internal/dispatch/caching/caching.go
```diff
@@ -457,12 +463,15 @@ func (cd *Dispatcher) dispatchQueryPlanCheckCached(req *v1.DispatchQueryPlanRequ
...
-	prometheus.Unregister(cd.lookupSubjectsTotalCounter)
+	if cd.registerer != nil {
+		cd.registerer.Unregister(cd.checkTotalCounter)
+		cd.registerer.Unregister(cd.checkFromCacheCounter)
+		cd.registerer.Unregister(cd.lookupResourcesTotalCounter)
+		cd.registerer.Unregister(cd.lookupResourcesFromCacheCounter)
+		cd.registerer.Unregister(cd.lookupSubjectsFromCacheCounter)
+		cd.registerer.Unregister(cd.lookupSubjectsTotalCounter)
+	}
+
```
tstirrat15: This could be cleaner if we have a struct that tracks these and has an `Unregister` method, or else the registration function returns a function for unregistering.

tstirrat15 (commented):
I've left a couple of comments. I generally like where this is going but there's a couple of things I'd want to touch up. I also want to get another couple of pairs of eyes on this - I'll poke some folks.

There's also a handful of tests failing. The unit tests are easy to run, but you can use the `mage` targets to make the e2e and integration tests easy to run. `golangci-lint` will also catch these kinds of things across the rest of the codebase.

internal/datastore/crdb/pool/balancer.go
paegun: addressed in 428b7c383f3642782b431d1fa5be474b1db1baa4

internal/datastore/crdb/pool/balancer.go
paegun: addressed in 8af3a7305ec2e20d5bbbe378b7efdec8eb8f50a7

RegisterPrometheusMetrics returning a closure.

internal/dispatch/caching/caching.go
paegun: We have need of the counters, ie to increment, so would be due to wrap all metrics interactions, not just register/unregister.

paegun: I have submitted an alternative PR #3120 which takes the more aggressive approach of removing direct Prometheus usage w/i business logic paths.

paegun:
@tstirrat15 , cc: @miparnisari I provided this PR but believe https://github.com/authzed/spicedb/pull/3120 better covers #2829 . However, 3weeks, no feedback. I leave the work here for your consideration

PR #3108, opened 2026-05-09
otel: replace cobraotel with native lifecycle management

### LINKED ISSUE #3095: Bug: OpenTelemetry traces are not flushed on shutdown, dropping spans on SIGTERM
## Description

SpiceDB never calls `Shutdown()` or `ForceFlush()` on the OpenTelemetry `TracerProvider` during process termination. As a result, spans buffered in the `BatchSpanProcessor` at the moment of SIGTERM/SIGINT are dropped. This is visible on slow traces, where the root span arrives at the collector but many of its inner spans are missing.   

### Likely root cause                                                                                                                        
                                                                                                                                        
  The `TracerProvider` is constructed inside `cobraotel.RunE()` ([cobraotel.go#L202-L206](https://github.com/jzelinskie/cobrautil/blob/f7fe73942d0f/cobraotel/cobraotel.go#L202-L206)):
                                                                                                                                        
  ```go           
  otel.SetTracerProvider(trace.NewTracerProvider(
      trace.WithSampler(trace.ParentBased(trace.TraceIDRatioBased(sampleRatio))),
      trace.WithBatcher(exporter),                                                                                                      
      trace.WithResource(res),
  ))
```
                                                                                     
The provider is registered globally and the reference is discarded, therefore SpiceDB has no handle to it. When `main` returns, the `BatchSpanProcessor`  goroutine is likely being abruptly killed and any unexported spans are lost.

### Proposed fix

Drop the `cobraotel` dependency; have SpiceDB build, own and terminate the `TracerProvider` object.

### PR DESCRIPTION
**Pull request by Jdepp007004**
## Description

Closes https://github.com/authzed/spicedb/issues/3095 and https://github.com/authzed/spicedb/issues/712

## Testing

1. `docker-compose up --build`, then send a few requests to the server and go to localhost:3000. You should be able to see traces
2. shutdown the docker containers. You should see this log: "shutting down OTel provider"

### CODE REVIEW

File: pkg/cmd/server/otel.go [hunk 1]
```diff
@@ -0,0 +1,252 @@
...
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+)
+
+// otelProviderContextKey is the unexported context key for the TracerProvider.
+type otelProviderContextKey struct{}
```
tstirrat15: Use `ctxkey` generally

pkg/cmd/server/otel.go
```diff
@@ -0,0 +1,252 @@
...
+	f.StringToString("otel-headers", nil,
+		`key=value pairs sent as headers to the OTel provider`)
+	f.Float64("otel-sample-ratio", 0.01,
+		`ratio of traces that are sampled`)
+
+	// Legacy flags
+	f.String("otel-jaeger-endpoint", "", "OpenTelemetry collector endpoint - the endpoint can also be set by using enviroment variables")
+	_ = f.MarkHidden("otel-jaeger-endpoint")
+	f.String("otel-jaeger-service-name", "spicedb", "service name for trace data")
+	_ = f.MarkHidden("otel-jaeger-service-name")
```
tstirrat15: Drop these

pkg/cmd/server/otel.go
```diff
@@ -0,0 +1,252 @@
...
+	flushCtx, flushCancel := context.WithTimeout(ctx, OTelShutdownTimeout)
+	defer flushCancel()
+	if err := provider.ForceFlush(flushCtx); err != nil {
+		// Log but continue — Shutdown must still be attempted.
+		fmt.Printf("otel: ForceFlush error (continuing to Shutdown): %v\n", err)
+	}
+
+	shutCtx, shutCancel := context.WithTimeout(ctx, OTelShutdownTimeout)
+	defer shutCancel()
+	return provider.Shutdown(shutCtx)
```
tstirrat15: Why a separate context?

pkg/cmd/server/otel.go
```diff
@@ -0,0 +1,252 @@
...
+// and stores it on the command context so the shutdown handler can retrieve it.
+func OTelPreRunE(cmd *cobra.Command, _ []string) error {
+	provider, err := InitOTelProvider(cmd)
+	if err != nil {
+		return fmt.Errorf("initializing OTel provider: %w", err)
+	}
+	parent := cmd.Context()
+	if parent == nil {
+		parent = context.Background()
+	}
```
tstirrat15: Same q here

pkg/cmd/server/otel.go
```diff
@@ -0,0 +1,252 @@
...
+	_ = f.MarkHidden("otel-jaeger-endpoint")
+	f.String("otel-jaeger-service-name", "spicedb", "service name for trace data")
+	_ = f.MarkHidden("otel-jaeger-service-name")
+}
+
+// InitOTelProvider reads otel-* flags from cmd, builds a TracerProvider, sets
+// it as the global OTel provider, and returns it for lifecycle management.
+//
+// Returns (nil, nil) when otel-provider is "none". Callers must handle nil.
+func InitOTelProvider(cmd *cobra.Command) (otelShutdowner, error) {
```
tstirrat15: `cobraotel` tied the provider to the lifecycle of the command because that was what made sense in that package, but I don't think we need or want it in this context.

The general flow for setting up components of SpiceDB is:

1. Take cobra flags that are defined in the command and put their values into a config struct
2. Take that config struct and construct the required object
3. Pass that object down and in as necessary

The [telemetry prometheus registry](https://github.com/authzed/spicedb/blob/3e8337db21884d198643a1684a90335240c5b408/pkg/cmd/server/server.go#L509-L539) is probably a decent analogue.

pkg/cmd/util/util.go
```diff
@@ -439,8 +439,19 @@ func (d *disabledHTTPServer) Close() {}
 // so that they were shared across all commands, but this
 // made it difficult to organize the flags, so we lifted them here.
 func RegisterCommonFlags(cmd *cobra.Command) {
-	otel := cobraotel.New("spicedb")
-	otel.RegisterFlags(cmd.Flags())
+	f := cmd.Flags()
```
tstirrat15: I like this

pkg/cmd/serve.go
```diff
@@ -265,14 +262,23 @@ func NewServeCommand(programName string, config *server.Config) *cobra.Command {
...
 			signalctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 			defer stop()
 
-			return server.Run(signalctx)
+			defer func() {
+				// Shutdown OTel provider to ensure all traces are flushed
+				if provider := server.OTelProviderFromContext(cmd.Context()); provider != nil {
+					if err := server.ShutdownOTelProvider(context.Background(), provider); err != nil {
+						log.Warn().Err(err).Msg("failed to cleanly shutdown OpenTelemetry provider")
+					}
```
tstirrat15: Yeah, this makes sense, but I'd rather that we configure the otel provider here as well, not in the RunE

pkg/cmd/server/otel_integration_test.go
```diff
@@ -0,0 +1,82 @@
...
+// TestOTelIntegration_FullChain_EnvToProvider simulates the full
+// DefaultPreRunE chain with OTel flags set and verifies the TracerProvider
+// is non-nil in the command context after OTelPreRunE executes.
+func TestOTelIntegration_FullChain_EnvToProvider(t *testing.T) {
+	cmd := &cobra.Command{Use: "test"}
+	RegisterOTelFlags(cmd)
+	cmd.SetContext(context.Background())
+	require.NoError(t, cmd.Flags().Set("otel-provider", "otlpgrpc"))
+	require.NoError(t, cmd.Flags().Set("otel-endpoint", "localhost:4317"))
+	require.NoError(t, cmd.Flags().Set("otel-insecure", "true"))
```
tstirrat15: I generally like the tests. Can we add a test or two that establishes 1. that you can use the otel environment variables to configure flags that aren't set and 2. which value overrides which if you declare both?

tstirrat15 (commented):
See comments

pkg/cmd/serve.go
miparnisari: why are we doing the flushing here and not putting the shutdown function of the provider inside of the `closeables` struct that the server already has?

https://github.com/authzed/spicedb/blob/ee7c9a7c93a3ae75fdc9a5d9a30d3f6511cc479a/pkg/cmd/server/server.go#L177-L178

miparnisari (changes requested):
Would it be possible to add a new struct within 

https://github.com/authzed/spicedb/blob/ee7c9a7c93a3ae75fdc9a5d9a30d3f6511cc479a/pkg/cmd/server/server.go#L63

called `otelConfig` or something like that, so that anyone doing `server.NewConfigWithOptionsAndDefaults` can set these programmatically?

pkg/cmd/server/otel.go
Jdepp007004: Done, switched to the ctxkey package to match how the rest of the codebase handles context keys.

pkg/cmd/server/otel.go
Jdepp007004: Dropped them from both RegisterOTelFlags and RegisterCommonFlags in util.go

pkg/cmd/server/otel.go
Jdepp007004: there was no real reason for two separate contexts there, it was an oversight on my part ,
fixed to use one shared context for the whole operation

pkg/cmd/server/otel.go
Jdepp007004: removed these too

pkg/cmd/server/otel.go
Jdepp007004: Restructured to follow the pattern you described. Flag values now get copied into an OTelConfig struct, InitOTelProvider takes that struct along with a context and has no dependency on Cobra anymore. The provider is constructed inside Complete and registered with closeables so shutdown is handled automatically as part of the normal server shutdown sequence.

pkg/cmd/util/util.go
Jdepp007004: Thanks, glad it reads better there

pkg/cmd/serve.go
Jdepp007004: You're right, the defer approach was the wrong place for it. I wasn't fully across the closeables pattern when I first wrote that. Moved into Complete now so OTel shuts down in the same ordered sequence as everything else rather than racing against it.

pkg/cmd/server/otel_integration_test.go
Jdepp007004: Done, added TestOTelConfig_EnvVarConfiguresUnsetFlag and TestOTelConfig_ExplicitFlagOverridesEnvVar to the integration test file. First one verifies the SDK picks up OTEL_EXPORTER_OTLP_ENDPOINT when the endpoint flag is not explicitly set, second one documents that an explicit flag value takes precedence over the env var.

Jdepp007004:
Addressing miparnisari's review — added OTelConfig as a struct in the server package with an OTel field on Config so it can be set programmatically without going through Cobra flags. InitOTelProvider now takes that struct directly along with a context and has no Cobra dependency at all.

pkg/cmd/serve.go
```diff
@@ -260,15 +257,27 @@ func NewServeCommand(programName string, config *server.Config) *cobra.Command {
...
 				return err
 			}
 			signalctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 			defer stop()
 
-			return server.Run(signalctx)
+			// OTel provider shutdown is handled by closeables inside Complete().
+			// No explicit defer needed here.
+
+			return srv.Run(signalctx)
```
miparnisari: please remove comment and undo `srv` rename

pkg/cmd/util/util.go
miparnisari: The code in `RegisterOTelFlags` looks awfully similar :) 

I think this code can be deleted. None of the other commands (other than `spicedb serve`) use the otel flags.

EDIT: i've gone ahead and deleted it.

pkg/cmd/server/otel_integration_test.go
miparnisari, line 1: Do we really need **three** different files for tests? 

EDIT: I've simplified this

miparnisari (dismissed):
Approving, but i'm biased because I made changes 😄

pkg/cmd/server/otel_integration_test.go
miparnisari: I removed these tests because they weren't really asserting anything useful

docs/spicedb.md
```diff
@@ -123,12 +123,6 @@ spicedb datastore gc [flags]
...
       --datastore-watch-connect-timeout duration                              how long the watch connection to the underlying datastore should wait before timing out (CockroachDB driver only) (default 1s)
-      --otel-endpoint string                                                  OpenTelemetry collector endpoint - the endpoint can also be set by using enviroment variables
-      --otel-insecure                                                         connect to the OpenTelemetry collector in plaintext
-      --otel-provider string                                                  OpenTelemetry provider for tracing ("none", "otlphttp", "otlpgrpc") (default "none")
-      --otel-sample-ratio float                                               ratio of traces that are sampled (default 0.01)
-      --otel-service-name string                                              service name for trace data (default "spicedb")
-      --otel-trace-propagator string                                          OpenTelemetry trace propagation format ("b3", "w3c", "ottrace"). Add multiple propagators separated by comma. (default "w3c")
```
miparnisari, line 131: FYI - `serve` still has them. All the other commands don't use these values.

pkg/cmd/server/otel.go
```diff
@@ -0,0 +1,161 @@
...
+type otelShutdowner interface {
+	Shutdown(ctx context.Context) error
+	ForceFlush(ctx context.Context) error
+}
+
+// OTelConfig holds the configuration for the OpenTelemetry provider.
+// It is bound to Cobra flags by RegisterOTelFlags and lives on server.Config
+// so the provider can be constructed during Complete().
+// Embedded callers may set it programmatically.
+type OTelConfig struct {
```
tstirrat15, line 32: May be worth adding the options build tag thing to this so that we get debug information

tstirrat15 (dismissed):
See comment, otherwise LGTM

pkg/cmd/server/otel.go
miparnisari: done

tstirrat15 (approved):
LGTM

PR #3109, opened 2026-05-12
feat: add prometheus metrics for DispatchQueryPlan

barakmich:
## Description

We weren't keeping track of Prometheus dispatch metrics for DispatchQueryPlan in our load testing. This adds them, and separable by operation.


tstirrat15 (approved):
LGTM

PR #3112, opened 2026-05-12
chore: use otter as the primary cache implementation and get rid of alternative implementations

tstirrat15:
## Description
We've had three alternative implementations for a while now, with `ristretto` being the one that is available by default, with two other implementations behind flags. We've load tested this internally and it appears that all three of the implementations are interchangeable from a performance perspective. `otter` is the most modern of the three and it supports generics, which means that we don't have to muck around with calculating our own keys.

This PR makes otter the primary cache, removes the other implementations, and simplifies where possible.

## Changes
Will annotate.

## Testing
Review. See that benchmarks are the same.


internal/datastore/proxy/schemacaching/caching.go
```diff
@@ -27,8 +27,7 @@ const (
 // DatastoreProxyTestCache returns a cache used for testing.
 func DatastoreProxyTestCache(t testing.TB) cache.Cache[cache.StringKey, CacheEntry] {
 	cache, err := cache.NewStandardCache[cache.StringKey, CacheEntry](&cache.Config{
-		NumCounters: 1000,
```
tstirrat15, line 30: This configuration value isn't used by `otter`, so I deprecated it and removed places where it's set.

internal/dispatch/keys/computed_test.go
```diff
@@ -703,20 +701,6 @@ func TestCacheKeyNoOverlap(t *testing.T) {
 	}
 }
 
-func TestComputeOnlyStableHash(t *testing.T) {
```
tstirrat15, line 706: This is related to removing ristretto - there was a process-specific hash that was a part of its key computation, but that isn't used by the other cache implementations, so I got rid of its computation and got rid of logic and tests around that.

internal/dispatch/keys/dispatchcheckkey.go
```diff
@@ -18,17 +17,9 @@ func (dck DispatchCacheKey) StableSumAsBytes() []byte {
...
-// AsUInt64s returns the cache key in the form of two uint64's. This method returns uint64s created
-// from two distinct hashing algorithms, which should make the risk of key overlap incredibly
-// unlikely.
-func (dck DispatchCacheKey) AsUInt64s() (uint64, uint64) {
-	return dck.processSpecificSum, dck.stableSum
-}
-
 func (dck DispatchCacheKey) KeyString() string {
 	firstBytes := binary.AppendUvarint(make([]byte, 0, 8), dck.stableSum)
-	secondBytes := binary.AppendUvarint(make([]byte, 0, 8), dck.processSpecificSum)
```
tstirrat15, line 30: This value is always `0` unless it's been computed in the ristretto logic.

internal/dispatch/keys/dispatchcheckkey.go
```diff
@@ -18,17 +17,9 @@ func (dck DispatchCacheKey) StableSumAsBytes() []byte {
 	return binary.AppendUvarint(make([]byte, 0, 8), dck.stableSum)
 }
 
-// AsUInt64s returns the cache key in the form of two uint64's. This method returns uint64s created
-// from two distinct hashing algorithms, which should make the risk of key overlap incredibly
-// unlikely.
-func (dck DispatchCacheKey) AsUInt64s() (uint64, uint64) {
```
tstirrat15, line 24: This was only used in the ristretto implementation.

internal/dispatch/keys/hasher_ristretto.go
```diff
@@ -1,82 +0,0 @@
...
-}
-
-//go:noescape
-//go:linkname memhash runtime.memhash
-func memhash(p unsafe.Pointer, h, s uintptr) uintptr
-
-func runMemHash(seed uint64, data []byte) uint64 {
-	ss := (*stringStruct)(unsafe.Pointer(&data))
-	return uint64(memhash(ss.str, uintptr(seed), uintptr(ss.len))) //nolint:gosec
-}
```
tstirrat15, line 74: This is the part that I removed - this is the only place where `computeOption` is actually referenced, and the `processSpecificSum` is only referenced in the key logic in the ristretto implementation.

internal/dispatch/keys/hasher.go
```diff
@@ -117,3 +119,52 @@ func (hc hashableContext) AppendToHash(hasher hasherInterface) {
 	}
 	hasher.WriteString(stable)
 }
+
+// dispatchCacheKeyHash computes a DispatchCheckKey for the given prefix and any hashable values.
```
tstirrat15, line 123: This is pulling over the `hasher_ristretto` logic that we're keeping

internal/dispatch/keys/hasher_wasm.go
tstirrat15, line 1: We shouldn't need a separate wasm target now, because we're not using features that are unsupported by the wasm target.

internal/dispatch/keys/keys.go
```diff
@@ -62,71 +62,71 @@ type Handler interface {
 type baseKeyHandler struct{}
 
 func (b baseKeyHandler) LookupResources2CacheKey(_ context.Context, req *v1.DispatchLookupResources2Request) (DispatchCacheKey, error) {
-	return lookupResourcesRequest2ToKey(req, computeBothHashes)
```
tstirrat15, line 65: I got rid of the `computeOption` argument because the behaviors are no longer different.

pkg/cache/cache.go
```diff
@@ -23,19 +23,10 @@ func (sk StringKey) KeyString() string {
...
-	// counters than the max cache capacity, as this will improve eviction
-	// accuracy and subsequent hit ratios.
-	//
-	// For example, if you expect your cache to hold 1,000,000 items when full,
-	// NumCounters should be 10,000,000 (10x). Each counter takes up roughly
-	// 3 bytes (4 bits for each counter * 4 copies plus about a byte per
-	// counter for the bloom filter). Note that the number of counters is
-	// internally rounded up to the nearest power of 2, so the space usage
-	// may be a little larger than 3 bytes * NumCounters.
+	// Deprecated: NumCounters was used to control behavior of the cache
```
tstirrat15, line 27: This option was specific to ristretto, so we deprecate it.

internal/dispatch/keys/computed.go
```diff
@@ -7,13 +7,6 @@ import (
 	"github.com/authzed/spicedb/pkg/tuple"
 )
 
-type dispatchCacheKeyHashComputeOption int
```
tstirrat15, line 10: The behavior here no longer differs, so I got rid of the options and removed their references.

pkg/cache/cache_otter.go
```diff
@@ -1,5 +1,3 @@
-//go:build !wasm
```
tstirrat15, line 1: We shouldn't need to exclude this from the wasm build anymore.

pkg/cache/cache_test.go
tstirrat15, line 1: These changes are just flattening the test and testing Otter directly.

pkg/cache/standard.go
```diff
@@ -1,12 +1,11 @@
-//go:build !wasm
-
 package cache
 
 // NewStandardCache creates a new cache with the given configuration.
 func NewStandardCache[K KeyString, V any](config *Config) (Cache[K, V], error) {
-	return NewRistrettoCache[K, V](config)
+	// TODO: check name here
+	return NewOtterCache[K, V]("", config)
```
tstirrat15, line 15: Should there be a name here? I wasn't sure.

pkg/cmd/server/cacheconfig.go
```diff
@@ -48,7 +48,7 @@ func (cc *CacheConfig) WithRevisionParameters(
 
 // CompleteCache translates the CLI cache config into a cache config.
 func CompleteCache[K cache.KeyString, V any](cc *CacheConfig) (cache.Cache[K, V], error) {
-	if !cc.Enabled || cc.MaxCost == "" || cc.MaxCost == "0%" || cc.NumCounters == 0 {
```
tstirrat15, line 51: `NumCounters` is deprecated, so we don't switch on it anymore.

pkg/cmd/server/cacheconfig.go
```diff
@@ -71,46 +71,16 @@ func CompleteCache[K cache.KeyString, V any](cc *CacheConfig) (cache.Cache[K, V]
 		return nil, errors.New("could not cast max cost to int64")
 	}
 
-	if cc.CacheKindForTesting != "" {
```
tstirrat15, line 74: Otter is now the standard, so we get rid of this switching.

pkg/cmd/server/cacheconfig.go
```diff
@@ -133,14 +103,13 @@ func RegisterCacheFlags(flags *pflag.FlagSet, flagPrefix, flagDescription string
...
+	flags.Int64Var(&config.NumCounters, flagPrefix+"-num-counters", 0, "number of counters for tracking access frequency in the cache for "+flagDescription+". A higher number means more accurate eviction decisions but more memory usage")
+	err := flags.MarkDeprecated(flagPrefix+"-num-counters", "this flag is now unused")
+	if err != nil {
+		return err
+	}
 	flags.BoolVar(&config.Metrics, flagPrefix+"-metrics", defaults.Metrics, "enable metrics for the cache for "+flagDescription)
 	flags.BoolVar(&config.Enabled, flagPrefix+"-enabled", defaults.Enabled, "enable caching of "+flagDescription)
 
-	// Hidden flags.
-	flags.StringVar(&config.CacheKindForTesting, flagPrefix+"-kind-for-testing", defaults.CacheKindForTesting, "choose a different kind of cache, for testing")
```
tstirrat15, line 141: This flag was already hidden, so I'm comfortable with deleting it.

tstirrat15 (commented):
See comments

tstirrat15:
Hmm... seems Otter is still incompatible with wasm. I'll reinstate the wasm build flags.

tstirrat15:
Okay, this is ready for review.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 - Build: strip quarantine attribute for MacOS (https://github.com/authzed/spicedb/pull/3082)
+- Cache: switch to [otter](https://maypok86.github.io/otter/) as the primary cache implementation
```
miparnisari: move to `[unreleased]`, and add `https://github.com/authzed/spicedb/pull/3112`

internal/dispatch/keys/computed_test.go
miparnisari: 🥳

internal/dispatch/keys/dispatchcheckkey.go
```diff
@@ -7,8 +7,7 @@ import (
 // DispatchCacheKey is a struct which holds the key representing a dispatch operation being
 // dispatched or cached.
 type DispatchCacheKey struct {
-	stableSum          uint64
-	processSpecificSum uint64
+	stableSum uint64
 }
```
miparnisari: nit: can we make this `type DispatchCacheKey uint64`?

pkg/cache/cache_wasm.go
```diff
@@ -1,3 +1,5 @@
+//go:build wasm
```
miparnisari, line 1: how did this work before? 😆

pkg/cmd/server/cacheconfig.go
```diff
@@ -71,46 +71,16 @@ func CompleteCache[K cache.KeyString, V any](cc *CacheConfig) (cache.Cache[K, V]
...
-				NumCounters: cc.NumCounters,
-				DefaultTTL:  cc.defaultTTL,
-			})
-
-		default:
-			return nil, fmt.Errorf("unknown cache kind: %s", cc.CacheKindForTesting)
-		}
-	}
-
 	if cc.Metrics {
```
miparnisari, line 74: why do we have this `if` here? the config that gets sent in the constructor is exactly the same, and so is the underlying code. 

Looking at the implementation, it looks like we're always emmitting metrics 🤔  😬 (which is fine by me, i don't know why we expose `--cache-metrics` as a flag?)

pkg/cache/cache_wasm.go
miparnisari: also, can we get rid of `NewRistrettoCache` and `NewTheineCache`?

pkg/cache/standard.go
miparnisari:
```go
func mustRegisterCache(name string, c withMetrics) {
	if _, loaded := caches.LoadOrStore(name, c); loaded {
		panic("two caches with the same name: " + name)
	}
}

func unregisterCache(name string) {
	caches.Delete(name)
}
```

how does that work if the key is empty? 
why do we have `_`?
 should we not panic if `_` is actually `false`? so many questions..

pkg/cache/cache_wasm.go
tstirrat15: Not sure; if it's not necessary and we don't want it to be there I can remove it.

internal/dispatch/keys/dispatchcheckkey.go
tstirrat15: Sure

tstirrat15:
An LLM note about why otter isn't compatible with WASM:

>   Here's the chain:
>
> 1. xruntime.CacheLineSize = unsafe.Sizeof(cpu.CacheLinePad{}) — defined in otter's internal package
> 2. cpu.CacheLinePad is struct{ _ [cacheLineSize]byte } from golang.org/x/sys/cpu
> 3. In cpu_wasm.go (the WASM-specific file), cacheLineSize = 0 with the comment: "Make CacheLinePad an empty struct and hope that the usual struct alignment rules are good enough."
>
> So unsafe.Sizeof(cpu.CacheLinePad{}) → size of an empty struct → 0.
> 
> This is why otter can't be built for WASM — any code that uses CacheLineSize as an array size or divisor (e.g [CacheLineSize]byte or modulo operations) would either produce a zero-size array or a division by zero at runtime.

internal/dispatch/keys/keys.go
miparnisari: 🥳

miparnisari (approved):

PR #3117, opened 2026-05-13
testing: add benchmarking to compare LocalExecutor vs DispatchExecutor

barakmich:
…and "classic" compared to query planner

## Description

Adds benchmarking comparing LocalExecutor to DispatchExecutor in the dispatch/gRPC benchmark cases, as well as comparison with the current algorithm in pkg/query/benchmarks

### DIFF
--- a/internal/services/integrationtesting/benchmark_executors_test.go
+++ b/internal/services/integrationtesting/benchmark_executors_test.go
@@ -1,0 +1,311 @@
+//go:build integration
+
+package integrationtesting_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/internal/caveats"
+	"github.com/authzed/spicedb/internal/dispatch"
+	"github.com/authzed/spicedb/internal/dispatch/graph"
+	testdatastore "github.com/authzed/spicedb/internal/testserver/datastore"
+	"github.com/authzed/spicedb/internal/testserver/datastore/config"
+	bm "github.com/authzed/spicedb/pkg/benchmarks"
+	caveattypes "github.com/authzed/spicedb/pkg/caveats/types"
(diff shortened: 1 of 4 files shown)


internal/services/integrationtesting/benchmark_executors_test.go
```diff
@@ -0,0 +1,311 @@
...
+// datastore for each engine:
+//
+//	dispatch           — graph.LocalOnlyDispatcher.DispatchCheck (today's path).
+//	queryplan_local    — query.NewLocalContext + LocalExecutor (no dispatch hop).
+//	queryplan_dispatch — dispatch.NewQueryContext + DispatchExecutor (alias
+//	                     boundaries cross back through the dispatcher).
+//
+// gRPC is intentionally bypassed so the numbers reflect the executor
+// difference, not transport overhead. Hierarchy:
+// engine → scenario → operation → variant.
```
tstirrat15, line 53: Does this mean that this is mostly about the construction and execution of the executors, rather than actually exercising the path that the dispatch would walk (if that phrasing makes sense)?

pkg/query/benchmarks/flags_test.go
```diff
@@ -7,8 +7,9 @@ import (
 )
 
 var (
-	includeDelay = flag.Bool("bench-delay", false, "include delay variants in query planner benchmarks")
-	includePlain = flag.Bool("bench-plain", false, "include plain (non-advised) benchmark variants")
+	includeDelay   = flag.Bool("bench-delay", false, "include delay variants in query planner benchmarks")
+	includePlain   = flag.Bool("bench-plain", false, "include plain (non-advised) benchmark variants")
+	includeClassic = flag.Bool("bench-classic", false, "include the classic (graph dispatcher) variant for Check; uses an in-memory, no-cache, no-gRPC local dispatcher")
```
tstirrat15, line 12: Okay yeah got it it's just about the overhead of the dispatch logic

tstirrat15 (approved):
LGTM

PR #3118, opened 2026-05-13
feat: registry for the iterators

barakmich:
## Description

Adds a registry for new iterator implementations, along with their logical symbols and instantiation.

Importantly, this allows for new iterators outside of `pkg/query` (for example, a custom dispatch iterator, to denote a dispatch boundary)

### DIFF
--- a/pkg/query/alias.go
+++ b/pkg/query/alias.go
@@ -1,6 +1,28 @@
 package query
 
-import "strings"
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+func init() {
+	MustRegisterIterator(IteratorSpec{
+		Type: AliasIteratorType,
+		Name: "Alias",
+		ConstructWithArgs: func(args *IteratorArgs, subs []Iterator, key CanonicalKey) (Iterator, error) {
+			if len(subs) != 1 {
+				return nil, fmt.Errorf("AliasIterator requires exactly 1 subiterator, got %d", len(subs))
+			}
+			if args == nil || args.RelationName == "" {
+				return nil, errors.New("AliasIterator requires RelationName in Args")
+			}
+			alias := NewAliasIteratorWithChain(args.DefinitionName, args.RelationName, args.AliasedAs, subs[0])
+			alias.canonicalKey = key
+			return alias, nil
+		},
+	})
+}
 
 // AliasIterator is an iterator that rewrites the Resource's Relation field of all paths
 // streamed from the sub-iterator.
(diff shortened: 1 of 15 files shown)


pkg/query/alias.go
```diff
@@ -1,6 +1,28 @@
...
+	MustRegisterIterator(IteratorSpec{
+		Type: AliasIteratorType,
+		Name: "Alias",
+		ConstructWithArgs: func(args *IteratorArgs, subs []Iterator, key CanonicalKey) (Iterator, error) {
+			if len(subs) != 1 {
+				return nil, fmt.Errorf("AliasIterator requires exactly 1 subiterator, got %d", len(subs))
+			}
+			if args == nil || args.RelationName == "" {
+				return nil, errors.New("AliasIterator requires RelationName in Args")
+			}
```
tstirrat15, line 19: I dunno whether this would make things better or worse, but it makes me wonder if there's a way to do declarative validation (a la Zod) of this kind of stuff - have tags on the (a?) struct that name the constraints and then you can call `something.Validate`

This is mostly me not liking manual validation logic if i can avoid it

pkg/query/intersection.go
```diff
@@ -6,6 +6,24 @@ import (
...
+		Type: IntersectionIteratorType,
+		Name: "Intersection",
+		ConstructWithArgs: func(_ *IteratorArgs, subs []Iterator, key CanonicalKey) (Iterator, error) {
+			it := NewIntersectionIterator(subs...)
+			// NewIntersectionIterator returns a FixedIterator when subs is empty.
+			switch v := it.(type) {
+			case *IntersectionIterator:
+				v.canonicalKey = key
+			case *FixedIterator:
+				v.canonicalKey = key
```
tstirrat15, line 20: Are these supposed to be the same?

pkg/query/registry.go
```diff
@@ -0,0 +1,57 @@
+package query
+
+import "github.com/authzed/spicedb/pkg/spiceerrors"
+
+// IteratorType is an enum to represent each basic type of iterator by a
+// well-known byte.
+//
+// Remember to also update  the allIteratorTypes list below when adding a new one.
```
tstirrat15, line 8: What is this comment referring to?

tstirrat15 (approved):
See comments, otherwise LGTM

pkg/query/intersection.go
barakmich: Yeah it is technically correct (the best kind) but mostly the canonical keys need a refresh before too long

barakmich:
Updated the comments in the next PR 🙂

PR #3119, opened 2026-05-13
fix: route query plan dispatch checks through singleflight

barakmich:
## Description

We have singleflight, but it doesn't have a group for query plans. Check is the most meaningful one here (same reason we don't have a group for LR or LS) so let's make checks singleflight-able

### DIFF
--- a/internal/dispatch/singleflight/singleflight.go
+++ b/internal/dispatch/singleflight/singleflight.go
@@ -39,8 +39,9 @@ type Dispatcher struct {
 	delegate   dispatch.Dispatcher
 	keyHandler keys.Handler
 
-	checkGroup  singleflight.Group[string, *v1.DispatchCheckResponse]
-	expandGroup singleflight.Group[string, *v1.DispatchExpandResponse]
+	checkGroup     singleflight.Group[string, *v1.DispatchCheckResponse]
+	expandGroup    singleflight.Group[string, *v1.DispatchExpandResponse]
+	planCheckGroup singleflight.Group[string, []*v1.DispatchQueryPlanResponse]
 }
 
 func (d *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckRequest) (*v1.DispatchCheckResponse, error) {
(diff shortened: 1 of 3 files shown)


internal/dispatch/singleflight/singleflight.go
```diff
@@ -147,11 +148,43 @@ func (d *Dispatcher) DispatchLookupSubjects(req *v1.DispatchLookupSubjectsReques
...
-	// other dispatch methods.
-	singleFlightCount.WithLabelValues("DispatchQueryPlan", "passthrough").Inc()
-	return d.delegate.DispatchQueryPlan(req, stream)
+	// Only PLAN_OPERATION_CHECK is request/response-shaped (single ResultPath
+	// per call) and therefore safe to deduplicate via singleflight. The lookup
+	// variants produce streamed multi-result outputs that don't fit the
+	// single-value singleflight model. Recursion safety is provided by the
+	// receiver-side DispatchExecutor, which refuses to dispatch any alias
+	// whose key is already in PlanContext.in_progress_keys, so a plan-check
+	// can never recurse to itself with the same dispatch key.
```
tstirrat15, line 157: With the potential issue here being a singleflight deadlock if this guarantee isn't present?

internal/dispatch/singleflight/singleflight_test.go
```diff
@@ -343,6 +345,126 @@ func TestDispatchQueryPlanRecordsSingleflightMetric(t *testing.T) {
 	assertCounterWithLabel(t, reg, 1, "spicedb_dispatch_single_flight_total", "DispatchQueryPlan")
 }
 
+func TestDispatchQueryPlanCheckCoalescesConcurrentCalls(t *testing.T) {
```
tstirrat15, line 348: If this isn't talking to a datastore, this would be a good candidate for `synctest`.

tstirrat15 (commented):
See comments

tstirrat15 (approved):
LGTM

PR #3121, opened 2026-05-15
fix(QP): short-circuit unions where applicable

miparnisari:


pkg/query/intersection_test.go
```diff
@@ -195,6 +196,10 @@ func TestIntersectionIteratorEarlyTermination(t *testing.T) {
 	path, err := ctx.Check(intersect, NewObject("document", "doc1"), NewObject("user", "alice").WithEllipses())
 	require.NoError(err)
 	require.Nil(path, "Early termination should return nil")
+
+	trace := traceLogger.DumpTrace()
+	require.Contains(trace, "sub-iterator 0 returned nil, short-circuiting",
```
miparnisari, line 201: FYI this test is just to prove that intersection is short-circuiting correctly. (I don't think there is a better way of asserting the short-circuiting)

barakmich (approved):

PR #3122, opened 2026-05-18
feat: introduce a hand-written serialize/deserialize for iterators

barakmich:
## Description

For both DispatchQueryPlan, as well as cursoring DispatchQueryPlan, we need some way to serde the iterator tree, and we'd like to do it quickly (ie, not with deep protos). This is pretty fast, as the benchmarks attest; Serialize takes about 1µs if it's small, maybe 12µs if it's particularly large. And about double that time on the Deserialize (2/24).

The follow up to this PR is twofold:
- Finish cursors, adding a flag to save intermediate state
- Unblock using it in DispatchQueryPlan
  - To send fewer dispatches
  - To send exact trees for execution (which, incidentally, might be as fast/faster than reconstructing the plan as we do today)

## Testing

Testing and benchmarks in the change


pkg/query/serde_test.go
tstirrat15, line 1: Would love some [rapid](https://github.com/flyingmutant/rapid) tests here.

tstirrat15 (dismissed):
Looks good, would love to see some property testing.

tstirrat15 (approved):
LGTM

PR #3125, opened 2026-05-19
refactor: carry schema hash inside ZedTokens

josephschorr:
Adds bytes schema_hash to `V1ZedToken`. The consistency middleware now propagates the real hash from the token on at-exact-snapshot and zedtoken-wins at-least-as-fresh paths, instead of substituting a bypass sentinel. `RevisionFromContext` emits new tokens carrying the hash.

`DecodeRevision` now returns a `DecodedRevision` struct `(Revision, SchemaHash, Status)` for readability.

`RemoteClockRevisions.CheckRevision` gains a separate `nowOnlyFunc` that skips the schema_revision subquery; CRDB and Spanner each register a timestamp-only variant. Datastores that don't register `nowOnlyFunc` fall back to nowFunc, preserving prior behavior.


internal/services/v1/schema.go
```diff
@@ -228,7 +228,7 @@ func (ss *schemaServer) ReflectSchema(ctx context.Context, req *v1.ReflectSchema
 	}
 
 	dl := datalayer.MustFromContext(ctx)
-	zedToken, err := zedtoken.NewFromRevision(ctx, atRevision, dl)
+	zedToken, err := zedtoken.NewFromRevision(ctx, atRevision, datalayer.NoSchemaHashInLegacyZedToken, dl)
```
miparnisari: why not `schemaHash`? a few lines above you are calling `loadCurrentSchema` which has `schemaHash` available, so you could return it there and then pass it here.

internal/services/v1/schema.go
```diff
@@ -175,7 +175,7 @@ func (ss *schemaServer) WriteSchema(ctx context.Context, in *v1.WriteSchemaReque
 		return nil, ss.rewriteError(ctx, err)
 	}
 
-	zedToken, err := zedtoken.NewFromRevision(ctx, revision, dl)
+	zedToken, err := zedtoken.NewFromRevision(ctx, revision, datalayer.NoSchemaHashInTransaction, dl)
```
miparnisari: this is weird. The description of this sentinel value is

> NoSchemaHashInTransaction is a sentinel value used when reading within a read-write transaction where the schema revision is not yet stable.

this is not the case here. 

why can we not compute the hash directly from the schema that was just written and return it?

internal/services/v1/watch.go
```diff
@@ -109,7 +109,7 @@ func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchS
 				}
 
 				filteredRelationshipUpdates := filterRelationshipUpdates(objectTypes, filters, update.RelationshipChanges)
-				zedToken, err := zedtoken.NewFromRevision(ctx, update.Revision, dl)
+				zedToken, err := zedtoken.NewFromRevision(ctx, update.Revision, datalayer.NoSchemaHashForWatch, dl)
```
miparnisari: in line 60 you can get the schema hash. and in line 77 you are ignoring the schema hash.

why not save it and pass it through lines 112 and 83??

pkg/zedtoken/zedtoken.go
```diff
@@ -58,41 +59,46 @@ type RevisionHolder interface {
...
-	return newFromRevision(revision, datastoreUniqueID)
+	return newFromRevision(revision, datastoreUniqueID, schemaHash)
 }
 
-// NewFromRevisionNoDatastoreID generates an encoded zedtoken from an integral revision without a datastore ID.
+// NewFromRevisionNoDatastoreID generates an encoded zedtoken without a datastore ID.
 // This is only for use in legacy scenarios. Most callers should use NewFromRevision.
-func NewFromRevisionNoDatastoreID(revision datastore.Revision) (*v1.ZedToken, error) {
-	return newFromRevision(revision, legacyEmptyDatastoreID)
+func NewFromRevisionNoDatastoreID(revision datastore.Revision, schemaHash datalayer.SchemaHash) (*v1.ZedToken, error) {
```
miparnisari, line 81: for a follow-up PR: this function should be called `NewFromRevisionForTesting`. it's doing the exact same thing as `MustNewFromRevisionForTesting` but not panicking.

pkg/zedtoken/zedtoken.go
```diff
@@ -132,48 +138,83 @@ func Decode(encoded *v1.ZedToken) (*zedtoken.DecodedZedToken, error) {
...
 		}
 
 		datastoreUniqueID, err := ds.UniqueID(context.Background())
 		if err != nil {
-			return datastore.NoRevision, StatusUnknown, fmt.Errorf(errDecodeError, err)
+			return legacyResult, fmt.Errorf(errDecodeError, err)
 		}
-
 		datastoreUniqueIDPrefix := datastoreUniqueID
 		if len(datastoreUniqueIDPrefix) > uniqueIDPrefixLength {
```
miparnisari, line 196: for a follow-up PR: this `if` appears up above too.

pkg/zedtoken/zedtoken.go
```diff
@@ -132,48 +138,83 @@ func Decode(encoded *v1.ZedToken) (*zedtoken.DecodedZedToken, error) {
 	return decoded, nil
 }
 
-// DecodeRevision converts and extracts the revision from a zedtoken or legacy zookie.
-func DecodeRevision(encoded *v1.ZedToken, ds RevisionHolder) (datastore.Revision, TokenStatus, error) {
+// DecodedRevision is the result of decoding a zedtoken. SchemaHash is
+// NoSchemaHashInLegacyZedToken for tokens that pre-date the schema_hash field,
+// or whose Status is not StatusValid.
+type DecodedRevision struct {
```
miparnisari, line 144: for a follow-up PR: small code smell. There is also

```
type revisionHandle struct {
	revision   datastore.Revision
	schemaHash datalayer.SchemaHash
}
```

in `consistency.go`.

i prefer `DecodedRevision` over `revisionHandle` 😉

pkg/zedtoken/zedtoken_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 
 	"github.com/authzed/spicedb/internal/datastore/revisions"
+	"github.com/authzed/spicedb/pkg/datalayer"
 	"github.com/authzed/spicedb/pkg/datastore"
```
miparnisari, line 14: for a follow-up PR: i would like to standardize this in _one_ place. i.e. move `Revision` into `datalayer`

internal/services/v1/schema.go
josephschorr: Fixed

internal/services/v1/watch.go
josephschorr: It would require actively reloading it on schema changes; I've changed it to use the starting one, but if there are any schema changes, we fallback to not sending it. This also required significant additional testing.

internal/services/v1/watch.go
```diff
@@ -108,8 +110,13 @@ func (ws *watchServer) Watch(req *v1.WatchRequest, stream v1.WatchService_WatchS
...
 				}
 
+				if len(update.ChangedDefinitions) > 0 || len(update.DeletedCaveats) > 0 || len(update.DeletedNamespaces) > 0 {
+					// If there are schema changes, we simply mark the schema hash as unknown, since we don't want to attempt to track it across changes.
+					currentSchemaHash = datalayer.NoSchemaHashForWatch
+				}
+
 				filteredRelationshipUpdates := filterRelationshipUpdates(objectTypes, filters, update.RelationshipChanges)
-				zedToken, err := zedtoken.NewFromRevision(ctx, update.Revision, dl)
+				zedToken, err := zedtoken.NewFromRevision(ctx, update.Revision, currentSchemaHash, dl)
```
miparnisari, line 119: huh, i'm just realizing now: one downside of putting the `schemaHash` inside of the `zedToken` is this exactly. Conceptually speaking, having Watch return a zedToken that returns a `schemaHash` doesn't make sense. We're overloading the meaning of a `zedtoken`.

I wonder if Watch should return a different struct.

pkg/middleware/consistency/consistency_test.go
```diff
@@ -477,6 +477,92 @@ func TestAtLeastAsFreshWithMismatchedTokenExpectFullConsistency(t *testing.T) {
...
+	handle.schemaHash = expectedHash
+
+	_, gotHash, token, err := RevisionFromContext(ctx)
+	require.NoError(err)
+	require.Equal(expectedHash, gotHash)
+
+	// Round-trip the issued token and confirm the hash made it onto the wire.
+	decoded, err := zedtoken.DecodeRevision(token, dl)
+	require.NoError(err)
+	require.Equal(expectedHash, decoded.SchemaHash)
```
miparnisari, line 563: missing: `ds.AssertExpectations(t)`

internal/services/v1/watch_test.go
```diff
@@ -434,6 +435,138 @@ definition document {
...
+		case resp := <-watchResponses:
+			received = append(received, resp)
+		case <-time.After(3 * time.Second):
+			require.FailNow("timed out waiting for watch responses")
+		}
+	}
+
+	// schemaHashOf decodes the schema hash carried by a watch response's ZedToken.
+	schemaHashOf := func(resp *v1.WatchResponse) []byte {
+		decoded, err := zedtoken.Decode(resp.GetChangesThrough())
```
miparnisari, line 548: yeah, this is the part that looks odd to me. why would `watch` return a schema hash? who's going to use this value?

miparnisari (approved):

PR #3127, opened 2026-05-20
fix: postgres implementation of HeadRevision

miparnisari:
Fixes a regression wherein a call to `postgres.HeadRevision` was inadvertently moving forward the transaction ID, which means that calling `HeadRevision` on a loop returns different revisions every time.


tstirrat15 (approved):
LGTM

Issue #3124, opened 2026-05-19
upgrade v1.53.0 failed


RyouZhang:
### What platforms are affected?

linux

### What architectures are affected?

amd64

### What SpiceDB version are you using?

v1.52.0 

### What SpiceDB flags do you use?

--datastore-engine=mysql
--enable-experimental-watchcache-schema-cache=true

### Steps to Reproduce

upgrade to v1.53.0, it's crashed

Server logs:
{"level":"info","timeout":1,"time":"2026-05-19T08:00:40.057774712Z","message":"overriding innodb_lock_wait_timeout"}
{"level":"error","error":"failed to create datastore: unable to load schema hash: Error 1146 (42S02): Table 'spicedb.stored_schema_revision' doesn't exist","time":"2026-05-19T08:00:40.066609925Z","message":"terminated with errors"}

### Expected Result

upgrade v1.53.0 failed, can't start

### Actual Result

{"level":"info","timeout":1,"time":"2026-05-19T08:00:40.057774712Z","message":"overriding innodb_lock_wait_timeout"}
{"level":"error","error":"failed to create datastore: unable to load schema hash: Error 1146 (42S02): Table 'spicedb.stored_schema_revision' doesn't exist","time":"2026-05-19T08:00:40.066609925Z","message":"terminated with errors"}

ebmeierj:
Please note, using the newer version appears to break our existing (postgresql) install even though we _don't_ use any of the experimental settings. The error appears to be the same as is being reported here. This smells like a regression.

miparnisari:
Hi folks, did you run `spicedb datastore migrate`? Ref: https://authzed.com/docs/spicedb/concepts/commands#reference-spicedb-datastore-migrate

RyouZhang:
> Hi folks, did you run `spicedb datastore migrate`? Ref: https://authzed.com/docs/spicedb/concepts/commands#reference-spicedb-datastore-migrate

yes, i have do `spicedb datastore migrate` before

kubectl exec -it spicedb-api-55b9fcb676-2582t -n authzed -- spicedb datastore migrate head --datastore-engine=mysql

2:25AM INF configured logging async=false format=auto log_level=info provider=zerolog
2:25AM INF GOMEMLIMIT is updated GOMEMLIMIT=1799999078 package=github.com/KimMachineGun/automemlimit/memlimit previous=9223372036854775807
[...]

PR #3128, opened 2026-05-20
test: include migration scripts for coverage

miparnisari:
I think we're not testing migrations properly 😭 ref: https://github.com/authzed/spicedb/issues/3124#issuecomment-4503419524

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -3,7 +3,7 @@ ignore:
   # not production code
   - "pkg/benchmarks/"
   - "pkg/query/benchmarks/"
-  - "**/zz_*.go"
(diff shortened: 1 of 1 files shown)


codecov.yaml
```diff
@@ -3,7 +3,7 @@ ignore:
   # not production code
   - "pkg/benchmarks/"
   - "pkg/query/benchmarks/"
-  - "**/zz_*.go"
+  - "**/zz_generated*.go"
```
miparnisari, line 6: we want to measure coverage for `zz_migration`

tstirrat15 (approved):
LGTM

PR #3134, opened 2026-05-25
fix(schemacache): supervisor recovers from transient watch errors

vroldanbet:
## Summary

Replaces the one-shot watch goroutine in the watching schema cache (`--enable-experimental-watchable-schema-cache`) with a supervisor that classifies errors via a deny-list and applies bounded exponential backoff. The cache now recovers from transient watch errors without a process restart.

The previous permanent-fallback path was datastore-agnostic — reachable from every backend — so the fix lives entirely on the cache side.

## Root cause

`internal/datastore/proxy/schemacaching/watchingcache.go` only retried `datastore.WatchRetryableError`. Any other error type — `WatchDisconnectedError`, bare unwrapped errors from datastore `default:` branches, unknown types — called `setFallbackMode()` and `return`ed. The retry counter was a lifetime cap (`maximumRetryCount = 10`), never reset on success, so a long-running pod that accumulated 11 transient blips would also fall back permanently.

## Fix

- **Deny-list classification (`isTerminalWatchError`).** Only `WatchCanceledError`, `WatchDisabledError`, and direct context cancellation are terminal. Watch is a read-only API, so retrying everything else is safe.
- **`setFallbackMode` on every cycle error**, terminal or not. Reads during the backoff window route through the standard caching proxy → datastore rather than being served from potentially-stale entries left over from the previous cycle.
- **Exponential backoff** via `github.com/cenkalti/backoff/v5` (`InitialInterval=100ms`, `Multiplier=2.0`, `RandomizationFactor=0.5`, `MaxInterval=30s`). `ExponentialBackOff` used directly — never `backoff.Retry()`, whose default 15-minute `MaxElapsedTime` would silently stop the supervisor.
- **Reset on success.** `bo.Reset()` fires when a cycle consumes any event so transient blips don't accumulate.
- **Cycle extraction.** Restart-loop body moved into `runWatchOnce` returning `(error, consumedEvent bool)`. Supervisor calls it in a loop, refreshing `HEAD` per cycle so long outages don't expire the revision.
- **`startSync` is now async.** The `wg.Wait()` barrier is gone — only the test suite observed it.

## New metrics

- `spicedb_datastore_watching_schema_cache_cycle_restarts_total` — counter, incremented at the start of every cycle after the first. A growing rate flags a flapping watch.
- `spicedb_datastore_watching_schema_cache_last_event_timestamp_seconds` — gauge, set to the wall-clock timestamp every time an event is consumed. Dashboards build "time since last event" as `time() - <metric>`.

## Tests

In `internal/datastore/proxy/schemacaching/watchingcache_test.go`:

- `TestIsTerminalWatchError`, `TestSleepHelper`, `TestNewSupervisorBackoff` — unit coverage for the new helpers.
- `TestWatchingCacheRecoversFromTransientError` — repro flipped, now asserts the full round-trip: `not-in-fallback` → error pushed → `in-fallback` → recovery event pushed → `not-in-fallback`. The intermediate assertion proves the error actually broke the cache.
- `TestWatchingCacheRecoversAcrossMultipleErrors` — 5 error→event cycles finish under 2 s, proving `bo.Reset()` works (without it the 5th iteration's backoff alone is ~1.6 s).
- `TestWatchingCacheTerminatesOnContextCancellation` — asserts fallback after `cancel()` but before `Close()` to isolate the cancellation path (`Close()` would otherwise mask a broken cancellation).
- `TestWatchingCacheTerminatesOnWatchDisabled` — `require.Never` (not `time.Sleep + check`) asserts the cycle counter doesn't advance after a terminal error.
- `TestWatchingCacheIncrementsCycleRestartsOnRecovery`, `TestWatchingCacheUpdatesLastEventTimestamp` — metric coverage.

10+ existing fallback-mode `require.Eventually` blocks were consolidated into an `assertEventuallyFallback(t, wcache, want bool)` helper.

`TestPostgresTransientErrorsAreResettable` was removed — supervisor correctness no longer depends on per-datastore classification.

## Behavior summary

| Error class | Before | After |
|---|---|---|
| Bare unwrapped datastore error | Permanent fallback | Brief fallback, supervisor restarts; cache recovers |
| `WatchRetryableError` | Retry up to 10 lifetime, then permanent fallback | Brief fallback, supervisor restarts; cache recovers |
| `WatchDisconnectedError` | Permanent fallback | Brief fallback, supervisor restarts; cache recovers |
| `WatchCanceledError` | Permanent fallback | Permanent fallback (shutdown — intended) |
| `WatchDisabledError` | Permanent fallback | Permanent fallback (datastore unsupported — intended) |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Cache: switch to [otter](https://maypok86.github.io/otter/) as the primary cache implementation (https://github.com/authzed/spicedb/pull/3112)
 
 ### Fixed
+- The watching schema cache (`--enable-experimental-watchable-schema-cache`) no longer enters permanent fallback on transient watch errors. A new supervisor restarts the watch cycle with bounded exponential backoff and only treats caller-driven cancellation or unsupported-watch as terminal (https://github.com/authzed/spicedb/pull/3134)
 - Watch consumers that request `WatchCheckpoints` now eventually observe every revision returned by `WriteRelationships` as a checkpoint. MemDB regressed this in https://github.com/authzed/spicedb/pull/2578 for no-op writes and MySQL never emitted checkpoints at all prior to now. Both now emit a checkpoint at the new revision. (https://github.com/authzed/spicedb/pull/3114)
 - When Query Planner evaluates a union, short-circuit if one of the branches yields a positive un-caveated result (https://github.com/authzed/spicedb/pull/3120)
 - DispatchQueryPlan previously did not try to use the singleflight middleware for check calls. (https://github.com/authzed/spicedb/pull/3119)
--- a/internal/datastore/proxy/schemacaching/watchingcache.go
+++ b/internal/datastore/proxy/schemacaching/watchingcache.go
@@ -6,9 +6,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v5"
 	"github.com/prometheus/client_golang/prometheus"
 
-	pgxcommon "github.com/authzed/spicedb/internal/datastore/postgres/common"
 	"github.com/authzed/spicedb/internal/datastore/revisions"
 	log "github.com/authzed/spicedb/internal/logging"
 	"github.com/authzed/spicedb/pkg/cache"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Query plan contexts are written to during recursive calls -- for now, disble dispatch inside recursive calls (https://github.com/authzed/spicedb/pull/3078)
+- The watching schema cache (`--enable-experimental-watchable-schema-cache`) no longer enters permanent fallback on transient watch errors. A new supervisor restarts the watch cycle with bounded exponential backoff and only treats caller-driven cancellation or unsupported-watch as terminal (https://github.com/authzed/spicedb/pull/3134)
```
miparnisari: Please move to `[unreleased]`

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -182,178 +242,189 @@ func (p *watchingCachingProxy) startSync(ctx context.Context) error {
...
-			caveats, err := reader.LegacyListAllCaveats(ctx)
-			if err != nil {
+			cycleErr, consumedEvent := p.runWatchOnce(ctx, headRevWithHash.Revision)
+			if cycleErr != nil {
 				p.namespaceCache.setFallbackMode()
 				p.caveatCache.setFallbackMode()
-				log.Warn().Err(err).Msg("received error in schema watch")
-				wg.Done()
+			}
+			if isTerminalWatchError(cycleErr) {
```
miparnisari, line 277: shouldn't this block be inside the block above? similar to 261-265

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -182,178 +242,189 @@ func (p *watchingCachingProxy) startSync(ctx context.Context) error {
...
+	reader := p.Datastore.SnapshotReader(headRev)
 
-				case <-p.closed:
-					log.Debug().Msg("schema watch closed")
-					return
+	log.Info().Str("revision", headRev.String()).Msg("prepopulating namespace watching cache")
+	namespaces, err := reader.LegacyListAllNamespaces(ctx)
+	if err != nil {
+		p.namespaceCache.setFallbackMode()
+		p.caveatCache.setFallbackMode()
```
miparnisari, line 310: nit - can we have a helper that does this? i see it in too many places:

```go
if err != nil {
	p.namespaceCache.setFallbackMode()
	p.caveatCache.setFallbackMode()
}
```

internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -54,10 +54,80 @@ var definitionsReadTotalCounter = prometheus.NewCounterVec(prometheus.CounterOpt
 	Help:      "total number of definitions read from the watching cache",
 }, []string{"definition_kind"})
 
-const maximumRetryCount = 10
+var cycleRestartsCounter = prometheus.NewCounter(prometheus.CounterOpts{
+	Namespace: "spicedb",
+	Subsystem: "datastore",
+	Name:      "watching_schema_cache_cycle_restarts_total",
+	Help:      "Times the watching schema cache restarted its watch cycle (initial cycle excluded). A high or growing rate indicates a flapping watch.",
```
miparnisari, line 61: nit - can we update the help text in lines 26 and 33:

> "Whether the watching schema cache for <REPLACE> definitions is in fallback mode (1) or normal mode (0). Fallback is triggered when the underlying datastore's schema watch becomes unavailable; in this state lookups bypass the watch-backed cache and are served from a standard caching proxy that reads from the datastore on miss.",

miparnisari (approved):

internal/datastore/proxy/schemacaching/watchingcache.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/3138

internal/datastore/proxy/schemacaching/watchingcache.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/3138

internal/datastore/proxy/schemacaching/watchingcache.go
vroldanbet: addressed in https://github.com/authzed/spicedb/pull/3138

PR #3138, opened 2026-05-26
refactor(schemacache): post-#3134 cleanup and race fix

vroldanbet:
## What

Follow-up to #3134 covering review feedback from Maria and two bug fixes uncovered during the round-trip:

- Correct the fallback-mode gauge help text — drop the CRDB-specific wording and the inaccurate "every lookup hits the datastore directly" claim.
- Nest cycle-error handling inside the `cycleErr != nil` block so it mirrors the HEAD-fetch path, eliminating a duplicate `if cycleErr != nil` guard.
- Extract `enterFallback()` for the eight sites that called `p.namespaceCache.setFallbackMode(); p.caveatCache.setFallbackMode()` in sequence. The four single-cache sites stay as-is.
- Add `isInFallback()` accessor that takes `lock.RLock()`; route all test reads (and `assertEventuallyFallback`) through it.
- Have `reset()` set `inFallbackMode = true` (previously `false`). Add `TestWatchingCacheStaysInFallbackDuringPrepopulate` as a deterministic repro.

## Why

- The gauge help missed cache is datastore-agnostic and overstated the fallback behavior (reads still go through the standard definition-caching proxy on miss, not the datastore directly).
- The cycle-error and HEAD-fetch error branches do the same thing on different shapes; consolidating the structure makes the policy easier to spot and removes a duplicate guard.
- `enterFallback()` is a single touchpoint for "the cache is no longer up to date" — useful if we ever want to attach metrics or logging to that state transition.
- The race detector caught unsynchronized test reads of `inFallbackMode`. The supervisor writes that field under `swc.lock`; tests were reading it directly, which had been masked by the synchronous `wg.Wait()` barrier in `startSync` before #3134 made `startSync` async.
- `reset()` was setting `inFallbackMode = false` *before* prepopulate ran, so `isInFallback()` could report ready while `entries` was still empty. Concurrent reads in that window missed the empty tracker, fell through to `readDefinition`, and produced spurious `NotFound` errors. The cache genuinely isn't ready between cycles — only `startAtRevision` should mark it ready. Side benefit: the fallback gauge no longer flickers `1 → 0 → 1` during recovery cycles, which is what operators looking at the metric actually want.

Both bug fixes verified with `go test ./internal/datastore/proxy/schemacaching/ -count=30 -race`.

## References

- Original PR: #3134

### DIFF
--- a/internal/datastore/proxy/schemacaching/watchingcache.go
+++ b/internal/datastore/proxy/schemacaching/watchingcache.go
@@ -150,6 +150,14 @@ func (p *watchingCachingProxy) Unwrap() datastore.Datastore {
 	return p.Datastore
 }
 
+// enterFallback puts both schema caches (namespaces and caveats) into fallback
+// mode so reads route through the standard definition-caching proxy until the
+// next successful watch cycle.
+func (p *watchingCachingProxy) enterFallback() {
+	p.namespaceCache.setFallbackMode()
+	p.caveatCache.setFallbackMode()
+}
+
 // createWatchingCacheProxy creates and returns a watching cache proxy.
 func createWatchingCacheProxy(delegate datastore.Datastore, c cache.Cache[cache.StringKey, *cacheEntry], gcWindow time.Duration, watchHeartbeat time.Duration) *watchingCachingProxy {
 	fallbackCache := &definitionCachingProxy{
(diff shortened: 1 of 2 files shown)


internal/datastore/proxy/schemacaching/watchingcache.go
```diff
@@ -550,12 +548,27 @@ func (swc *schemaWatchCache[T]) setFallbackMode() {
 	swc.fallbackGauge.Set(1)
 }
 
+// isInFallback returns the current fallback flag under the read lock.
+// Callers outside this type MUST use this rather than reading
+// inFallbackMode directly; the supervisor goroutine writes it under
```
miparnisari, line 553: nit - this comment is better placed as a comment to the field directly

miparnisari (approved):

PR #3140, opened 2026-05-26
fix: three correctness bugs surfaced by GRPC consistency cross-check

barakmich:
## Description

- recursive: inline the reflexive self-edge in IterSubjects so RecursiveIterator's
  Check-via-IterSubjects path honors `R#rel @ R#rel` identity.
- arrow: stop hard-coding Ellipsis for the intermediate's subject relation
  in checkRightToLeft/checkRightToLeftBatch. Derive it from the left's
  SubjectTypes so subrelation arrows (e.g. `maintainer: user | team#member`)
  pass the stored relation. Bug only manifested after the count advisor
  flipped arrow direction, hiding behind LR→LS sequencing.
- intersection_arrow: IterSubjectsImpl now intersects right-side subjects
  across all left subjects with caveats AND'd, instead of unioning them —
  `team.all(member)` requires the subject in EVERY team's member set.

## Testing

Added to the consistency tests to check the individual members of the set returned by LS

### DIFF
--- a/pkg/query/arrow.go
+++ b/pkg/query/arrow.go
@@ -196,12 +196,14 @@ func (a *ArrowIterator) checkRightToLeft(ctx *Context, resource Object, subject 
 
 		// rightPath.Resource is an intermediate object from the right side.
 		// Check if our input resource connects to this intermediate via the left side.
-		// Use tuple.Ellipsis as the relation since the left side stores subjects with "..."
-		// (the canonical relation for direct membership with no subrelation).
+		// The subject relation must match what the left side stores, which differs
+		// between schema arrows (left stores subjects with Ellipsis) and subrelation
+		// arrows (left stores subjects with a specific subject_relation, e.g. `member`).
+		// Derive it from the left's SubjectTypes for the intermediate's type.
 		intermediateAsSubject := ObjectAndRelation{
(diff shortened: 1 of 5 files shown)


pkg/query/arrow.go
tstirrat15, line 1: Are there unit tests for these modes that would help explain them? Not necessary but would be nice.

pkg/query/arrow.go
```diff
@@ -234,6 +236,32 @@ func (a *ArrowIterator) checkRightToLeft(ctx *Context, resource Object, subject
...
+// team:foo#member). Passing the wrong relation silently misses datastore rows.
+//
+// Implementation derives the expected relation from a.left.SubjectTypes(),
+// falling back to Ellipsis if the type isn't reflected there (e.g. constructed
+// iterator without metadata).
+func (a *ArrowIterator) intermediateSubjectRelation(intermediateType string) string {
+	subjectTypes, err := a.left.SubjectTypes()
+	if err != nil {
+		return tuple.Ellipsis
+	}
```
tstirrat15, line 252: What would an error indicate here? Like why do we not propagate this upwards?

tstirrat15 (dismissed):
LGTM

tstirrat15 (approved):
LGTM

PR #3141, opened 2026-05-28
fix(dispatch): close secondary grpc connections on shutdown

immanuwell:
SecondaryDispatch only stored the client interface, not the underlying conn, so Close() left secondary grpc connections dangling on shutdown.

Added a Conn field to SecondaryDispatch, stored it in combined.go, and updated Close() to drain secondaries before the primary. Field is nil-safe so test doubles are unaffected.

To reproduce: start with --dispatch-secondary-upstream-addr, shut down cleanly, secondary grpc goroutines (keep-alive etc) outlive the shutdown sequence.

### DIFF
--- a/internal/dispatch/remote/cluster.go
+++ b/internal/dispatch/remote/cluster.go
@@ -111,6 +111,10 @@ type SecondaryDispatch struct {
 	// Client is the client to use for dispatching to the secondary.
 	Client ClusterClient
 
+	// Conn is the underlying gRPC connection for the secondary dispatcher, used for closing.
+	// May be nil if the client is a mock or test double.
+	Conn *grpc.ClientConn
+
 	// MaximumHedgingDelay is the maximum delay that the primary will wait before dispatching
 	// when this secondary is active.
 	MaximumPrimaryHedgingDelay time.Duration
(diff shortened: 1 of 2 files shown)


immanuwell:

internal/dispatch/remote/cluster.go
```diff
@@ -914,10 +918,20 @@ func (cr *clusterDispatcher) DispatchQueryPlan(req *v1.DispatchQueryPlanRequest,
 }
 
 func (cr *clusterDispatcher) Close() error {
+	var firstErr error
```
tstirrat15, line 921: Prefer `errors.Join` here

tstirrat15 (commented):
The overall gist makes sense to me. Is it possible to write a unit test that captures this behavior, perhaps using a `goleak` assertion?

miparnisari (commented):
Thanks for your contribution! This LGTM, but I have a few questions:
- can you update CHANGELOG.md under `[unreleased]`?
- how did you notice this behavior?
- (for the maintainers): we use the `goleak` detector right now, I wonder why this wasn't caught
- (for the maintainers): this code has zero coverage 😱 https://app.codecov.io/gh/authzed/spicedb/blob/main/internal%2Fdispatch%2Fcombined%2Fcombined.go#L302

PR #3142, opened 2026-05-29
fix(datastore): prevent optimized revision singleflight from wedging the server

josephschorr:
The optimized revision computation runs under singleflight, which replaces the caller's context with one whose deadline has been stripped (context.WithoutCancel) and is only cancelled once every awaiting caller has gone away. As a result, a hung datastore call on this path (e.g. a half-open TCP connection silently dropped by a load balancer) could block the in-flight computation indefinitely. Because every API request needs an optimized revision and they all de-duplicate onto that single call, one dead connection could wedge the entire server.

This makes three complementary changes:

- Bound the shared, singleflighted computation with a low timeout so a wedged call cannot pin the latency of every waiting caller and inflate the system-wide P99. The healthy computation takes single-digit ms, so the bound is generous for the happy path.

- On a failed shared attempt, retry the computation directly (outside of singleflight) on the caller's own context, bounded by the caller's deadline (with a fallback for deadline-less callers). This keeps the aggressive shared bound from turning a transient wedge into a failed request; the retry is naturally throttled by the connection pool size.

- Bound the pgxpool acquire-time liveness ping via PingTimeout, exposed as a configurable connection-pool setting (default 5s) for both Postgres and CockroachDB read and write pools, so a half-open connection cannot hang acquisition. This is the only liveness ping pgxpool performs and it was previously unbounded.


internal/datastore/revisions/optimized.go
```diff
@@ -19,6 +19,33 @@ import (
...
+// singleflight) on the caller's own context, so the aggressive bound does not turn
+// a transient slow/wedged shared attempt into a failed request.
+const defaultOptimizedRevisionTimeout = 2 * time.Second
+
+// defaultOptimizedRevisionFallbackTimeout bounds the *direct* retry performed
+// after a failed shared attempt. The retry runs on the caller's own context, so
+// its deadline applies when present; this fallback only takes effect for callers
+// with no deadline of their own, preserving the guarantee that a revision
+// computation can never block forever.
+const defaultOptimizedRevisionFallbackTimeout = 30 * time.Second
```
tstirrat15: 30 seems like a lot? Or is it reasonable for the use case?

internal/datastore/revisions/optimized.go
```diff
@@ -64,48 +108,89 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
...
-		span.AddEvent(otelconv.EventDatastoreRevisionsComputed)
-		log.Ctx(ctx).Debug().Time("now", localNow).Time("valid", rvt).Stringer("validFor", validFor).Msg("setting valid through")
-		return datastore.RevisionWithSchemaHash{Revision: optimized, SchemaHash: schemaHash}, nil
-	})
+	// The shared, aggressively-bounded attempt failed (e.g. a wedged connection
+	// tripped the timeout, or the datastore returned a transient error). Give this
+	// request one more chance, directly and outside of singleflight, on its own
+	// context: this is not capped by the low shared bound and cannot re-attach to a
+	// poisoned in-flight call, so a brief wedge does not fail the request. The
+	// retry is naturally throttled by the connection pool's maximum size, so even
```
tstirrat15: Is this true for all datastores? I thought we don't use a connection pool for e.g. Spanner anymore

internal/datastore/revisions/optimized_test.go
```diff
@@ -227,3 +230,79 @@ func TestSingleFlightError(t *testing.T) {
...
+	req.True(one.Equal(res.Revision), "expected the direct retry to succeed")
+	req.Equal(int32(2), calls.Load(), "expected one shared attempt and one direct retry")
+}
+
+// TestOptimizedRevisionTimeout ensures that a datastore revision call that hangs
+// cannot wedge OptimizedRevision indefinitely. The shared attempt is bounded by
+// the (low) optimized-revision timeout, and the direct retry is bounded by the
+// caller's deadline (or the fallback timeout for deadline-less callers), so the
+// call always returns rather than blocking forever.
+func TestOptimizedRevisionTimeout(t *testing.T) {
```
tstirrat15, line 266: This test is a good candidate for `synctest.Test`

tstirrat15 (commented):
Mostly LGTM, see comments

jzelinskie:
It sounds like there are two orthogonal issues to address:
1. Single point of failure of the singleflight
2. Bounding the blast damage of a failed/slow singleflight

I think this PR addresses #2 quite well, although it seems like the timeout durations are orders of magnitude larger than they should be.

For #1, it probably makes sense shard requests across a few keys to singleflight and not exactly one. A simple strawman implementation would be to pick a random number between 1-N and append that to the singleflight key, so that way this issue would only impact 1/N of requests when it occurs. I suspect there's a sweet spot for managing blast damage to p99-impact that could be a constant in the codebase.

josephschorr:
> For #1, it probably makes sense shard requests across a few keys to singleflight and not exactly one. A simple strawman implementation would be to pick a random number between 1-N and append that to the singleflight key, so that way this issue would only impact 1/N of requests when it occurs. I suspect there's a sweet spot for managing blast damage to p99-impact that could be a constant in the codebase.

In the common scenario where the connection is not hanging, we don't want N calls: we do, indeed, want a single one. That's why I have the fallback: if it times out quickly, then we can retry.

> although it seems like the timeout durations are orders of magnitude larger than they should be.

We could vastly lower them... but then there is a risk of canceling a *blip* of a somewhat-longer call.

@tstirrat15 thoughts on making the defaults much lower?

pkg/cmd/datastore/datastore.go
```diff
@@ -91,6 +93,7 @@ func RegisterConnPoolFlagsWithPrefix(flagSet *pflag.FlagSet, prefix string, defa
 	flagSet.DurationVar(&opts.MaxLifetimeJitter, flagName("max-lifetime-jitter"), defaults.MaxLifetimeJitter, "waits rand(0, jitter) after a connection is open for max lifetime to actually close the connection (default: 20% of max lifetime, 30m for CockroachDB)")
 	flagSet.DurationVar(&opts.MaxIdleTime, flagName("max-idletime"), defaults.MaxIdleTime, "maximum amount of time a connection can idle in a remote datastore's connection pool")
 	flagSet.DurationVar(&opts.HealthCheckInterval, flagName("healthcheck-interval"), defaults.HealthCheckInterval, "amount of time between connection health checks in a remote datastore's connection pool")
+	flagSet.DurationVar(&opts.PingTimeout, flagName("ping-timeout"), defaults.PingTimeout, "amount of time to wait for a liveness ping against an acquired idle connection before discarding it in a remote datastore's connection pool")
```
miparnisari, line 96: i can't fathom a user understanding what this means, let alone setting the right value. Can we not just use a sensible default?

internal/datastore/revisions/optimized.go
```diff
@@ -64,48 +108,89 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
...
+	// defaultOptimizedRevisionTimeout) so a wedged computation cannot pin the
+	// latency of every waiting caller.
+	result, _, err := cor.updateGroup.Do(ctx, "", func(sfCtx context.Context) (datastore.RevisionWithSchemaHash, error) {
+		// NOTE: singleflight hands this function a context with the caller's
+		// deadline stripped. Re-impose a (low) deadline so a hung datastore call
+		// cannot block this in-flight call, and therefore every caller waiting on
+		// it, beyond the bound.
+		if cor.optimizedRevisionTimeout > 0 {
+			var cancel context.CancelFunc
+			sfCtx, cancel = context.WithTimeout(sfCtx, cor.optimizedRevisionTimeout)
```
miparnisari: ugh singleflight 😭 

i'm curious why you put the timeout inside of `do` and not outside. I don't have a good argument in favor of one or the other, I just want to understand your reasoning. For context, in https://github.com/authzed/spicedb/pull/3048 I put it for the callers.

tstirrat15:
re: default setting: I suppose we could load test around it?

internal/datastore/revisions/optimized.go
josephschorr: Lowered to 10s. Note this bound only takes effect on the direct retry, and only for callers that have no deadline of their own — anyone with a gRPC deadline is bounded by that deadline instead. It's a backstop guaranteeing a revision computation can never block forever, not latency the normal path observes. The number that actually controls blast radius is the shared 2s bound, per the load-testing discussion in the main thread.

internal/datastore/revisions/optimized.go
josephschorr: Correct — Spanner's client does its own session/channel pooling rather than using our pool config. Reworded the comment to scope the claim to the pooled SQL datastores and note that Spanner's client provides equivalent bounding.

internal/datastore/revisions/optimized_test.go
josephschorr: Converted to `synctest.Test` — the timeouts now run on the bubble's virtual clock, so the test is deterministic and no longer needs the goroutine + 5s guard scaffolding (synctest's deadlock detection covers the wedge case; verified the test still fails when the fix is reverted).

pkg/cmd/datastore/datastore.go
josephschorr: There is a sensible default — 5s, applied at the pgx layer even when the flag isn't set. The flag is just an override matching the rest of the pool-tuning flags (same audience that sets `healthcheck-interval`): the right value trades failure-detection speed against spuriously discarding healthy connections on slow links, which is deployment-specific. Happy to drop the flag if you'd rather not expose it, though.

internal/datastore/revisions/optimized.go
josephschorr: It follows from how resenje/singleflight handles contexts: the shared fn runs on a context detached from every caller (`WithoutCancel`), cancelled only once *all* waiters have gone away. A timeout outside `Do` bounds each caller's wait, but under sustained load new callers keep attaching before old ones detach — so a wedged shared call is never cancelled and the key stays poisoned indefinitely, which is exactly the wedge this PR fixes. Putting the timeout inside bounds the shared computation itself, so it fails promptly and the key is released for a fresh attempt. The two placements protect different things: outside (as in #3048) protects the caller; inside protects the key.

pkg/cmd/datastore/datastore.go
miparnisari: Yeah i'd rather not expose it (make it hidden). Then we can load test around it and find the best default value.

tstirrat15:
Do we have observability around the conditions where a lower `PingTimeout` would help?

My current thought is that we shouldn't expose it as a knob unless we know the circumstances where it would be desirable to have one value vs another; otherwise if a 5s default would work for most users let's just encode it.

Otherwise this is looking good to me and I want to get it through.

miparnisari (commented):
This LGTM.
- a few nits below
- can we update CHANGELOG.md please? Include what datastores it applies to (there are different fixes for different datastores)

internal/datastore/revisions/optimized.go
```diff
@@ -65,48 +110,90 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
...
-	})
+	// The shared, aggressively-bounded attempt failed (e.g. a wedged connection
+	// tripped the timeout, or the datastore returned a transient error). Give this
+	// request one more chance, directly and outside of singleflight, on its own
+	// context: this is not capped by the low shared bound and cannot re-attach to a
+	// poisoned in-flight call, so a brief wedge does not fail the request. For the
+	// pooled SQL datastores the retry is naturally throttled by the pool's maximum
+	// size; Spanner's client similarly bounds concurrent calls via its own session
+	// pool. Even when many waiters retry at once, the datastore is not overwhelmed.
+	log.Ctx(ctx).Warn().Err(err).Msg("shared optimized revision computation failed; retrying directly")
```
miparnisari, line 149:
```suggestion
	span.AddEvent("singleflight_caller_failed_retrying")
```

since logs are hard to see

internal/datastore/revisions/optimized.go
```diff
@@ -65,48 +110,90 @@ func (cor *CachedOptimizedRevisions) OptimizedRevision(ctx context.Context) (dat
...
+		return result, nil
+	}
 
-		cor.candidates = cor.candidates[numToDrop:]
-		cor.candidates = append(cor.candidates, validRevision{optimized, rvt, schemaHash})
-		cor.Unlock()
+	// If this caller's own context is already done, the shared call returned its
+	// error; there is nothing to retry.
+	if ctx.Err() != nil {
+		return datastore.RevisionWithSchemaHash{}, err
```
miparnisari:
```suggestion
		return datastore.RevisionWithSchemaHash{}, ctx.Err()
```

internal/datastore/revisions/optimized.go
```diff
@@ -27,8 +55,10 @@ type OptimizedRevisionFunction func(context.Context) (rev datastore.Revision, va
 // NewCachedOptimizedRevisions returns a CachedOptimizedRevisions for the given configuration
 func NewCachedOptimizedRevisions(maxRevisionStaleness time.Duration) *CachedOptimizedRevisions {
 	return &CachedOptimizedRevisions{
-		maxRevisionStaleness: maxRevisionStaleness,
-		clockFn:              clock.New(),
+		maxRevisionStaleness:             maxRevisionStaleness,
+		optimizedRevisionTimeout:         defaultOptimizedRevisionTimeout,
+		optimizedRevisionFallbackTimeout: defaultOptimizedRevisionFallbackTimeout,
```
miparnisari:
```suggestion
		optimizedRevisionProducerTimeout:         defaultOptimizedRevisionTimeout,
		optimizedRevisionConsumersTimeout: defaultOptimizedRevisionFallbackTimeout,
```

just for more clarity. Here and everywhere else

pkg/cmd/datastore/datastore.go
josephschorr: Hidden the flag (kept the 5s default wired in at the pgx layer) so we can load-test toward a default without exposing a knob users can't reason about. Easy to bake in a fixed default or un-hide once we have numbers.

internal/datastore/revisions/optimized.go
josephschorr: Added a span event (`spicedb.internal.datastore.revisions.shared_failed_retrying`) on the retry path so it's visible in traces. Kept the warn log too, since it still carries the underlying error for log-based diagnostics.

internal/datastore/revisions/optimized.go
josephschorr: Done. Note resenje/singleflight's `wait` already returns `ctx.Err()` as the error when the caller's context is what unblocked it, so in the common case this matches what we had — but it makes the select-race case deterministic and the intent explicit.

internal/datastore/revisions/optimized.go
josephschorr: Renamed the shared bound to `optimizedRevisionSharedTimeout` and kept `optimizedRevisionFallbackTimeout`. I held off on `consumersTimeout`: the shared bound is what every *waiting* caller actually observes, while the fallback only bounds the per-caller direct retry — so `consumers` read ambiguously to me. Happy to switch to `producer`/`consumers` if you'd still prefer that framing.

josephschorr:
Pushed the review-feedback round (46312d02d):

- **CHANGELOG** — added two entries under Unreleased/Fixed, scoped by datastore: the singleflight wedge fix (all datastores) and the connection ping-timeout (Postgres & CockroachDB).
- **`ping-timeout` observability** (@tstirrat15) — we do not currently have metrics isolating when a lower value would help, which is why I hid the flag rather than removing it: the 5s default ships, and we keep the override available so we can load-test toward a real default before exposing (or hard-coding) anything.
- Also: span event on the direct-retry path, `ctx.Err()` on the already-cancelled path, and renamed the shared bound to `optimizedRevisionSharedTimeout` (see inline replies).

pkg/cmd/datastore/datastore.go
```diff
@@ -58,6 +58,7 @@ type ConnPoolConfig struct {
 	MaxOpenConns        int           `debugmap:"visible"`
 	MinOpenConns        int           `debugmap:"visible"`
 	HealthCheckInterval time.Duration `debugmap:"visible" default:"30s"`
+	PingTimeout         time.Duration `debugmap:"visible"`
```
miparnisari:
```suggestion
	PingTimeout         time.Duration `debugmap:"visible" default:"5s"`
```

miparnisari (approved):

PR #3143, opened 2026-05-29
fix(streamtimeout): bound streams by inactivity, exempt bulk export

ivanauth:
The streaming-timeout interceptor cancels a stream if the server is silent for `StreamingAPITimeout` (default 30s), assuming a hung client. Bulk export legitimately goes silent for long stretches between batches on large datasets, so the interceptor was cancelling healthy streams (clients saw `RST_STREAM NO_ERROR`).

Add `WithExemptMethods` to the interceptor and exempt the two bulk **export** methods:
- `PermissionsService/ExportBulkRelationships`
- `ExperimentalService/BulkExportRelationships`

Bulk **import** is deliberately left bounded: it holds an open write transaction for the life of the stream, and each batch received from the client counts as activity, so the timeout still correctly catches a client that stops sending.

All other streaming methods keep the timeout.


internal/services/v1/experimental.go
```diff
@@ -112,7 +112,17 @@ func NewExperimentalServer(dispatch dispatch.Dispatcher, permServerConfig Permis
 				grpcvalidate.StreamServerInterceptor(validator),
 				handwrittenvalidation.StreamServerInterceptor,
 				usagemetrics.StreamServerInterceptor(),
-				streamtimeout.MustStreamServerInterceptor(config.StreamReadTimeout),
+				streamtimeout.MustStreamServerInterceptor(
+					config.StreamReadTimeout,
+					// Bulk export/import are designed to run for arbitrarily long
+					// periods on large datasets; their inter-batch gaps regularly
```
miparnisari: i'm curious about these inter-batch gaps? why do they happen?

pkg/cmd/serve.go
```diff
@@ -116,7 +116,7 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	apiFlags.Uint16Var(&config.MaximumUpdatesPerWrite, "write-relationships-max-updates-per-call", 1000, "maximum number of updates allowed for WriteRelationships calls")
 	apiFlags.IntVar(&config.MaxCaveatContextSize, "max-caveat-context-size", 4096, "maximum allowed size of request caveat context in bytes. A value of zero or less means no limit")
 	apiFlags.IntVar(&config.MaxRelationshipContextSize, "max-relationship-context-size", 25000, "maximum allowed size of the context to be stored in a relationship")
-	apiFlags.DurationVar(&config.StreamingAPITimeout, "streaming-api-response-delay-timeout", 30*time.Second, "maximum time that streaming APIs (LookupSubjects, LookupResources, ReadRelationships and ExportBulkRelationships) can be allowed to run but no response be sent to the client before the stream times out")
+	apiFlags.DurationVar(&config.StreamingAPITimeout, "streaming-api-response-delay-timeout", 30*time.Second, "maximum time that streaming APIs (LookupSubjects, LookupResources, ReadRelationships and ImportBulkRelationships) can be allowed to run but no data be sent to or received from the client before the stream times out; ExportBulkRelationships is exempt")
```
miparnisari, line 114: PR title: `exempt bulk export/import methods`

This description: `exempt bulk export`

🤔

PR #3147, opened 2026-06-02
chore: rework handles for gRPC tests

tstirrat15:
## Description
This started out as an adventure to get rid of deprecation warnings around `WithBlock`, as these [are discouraged](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md) by the gRPC docs. It turned into more of a rabbit hole when I followed the call stack up and found interfaces that didn't make sense to me.

The main idea is to split out the testing functions from the `RunnableServer` interface and change how the buffcon Listeners are stood up and passed around.

It also changes how clients are created, which potentially introduces an issue around a race between servers and clients, but a client configuration that makes the client `waitForReady` should avoid this.

## Changes
Will annotate.

## Testing
Review. See that tests pass.


e2e/spice/spicedb.go
```diff
@@ -114,8 +114,7 @@ func (s *Node) Connect(ctx context.Context, out io.Writer) error {
 	addr := net.JoinHostPort("localhost", strconv.Itoa(s.GrpcPort))
 	e2e.WaitForServerReady(addr, out)
 
-	conn, err := grpchelpers.DialAndWait(
-		ctx,
+	conn, err := grpchelpers.Dial(
```
tstirrat15: This is potentially a misnomer - what it actually does is call `NewClient`. I can change this to something else if desired.

internal/datastore/crdb/keys_test.go
```diff
@@ -142,14 +140,7 @@ func TestOverlapKeysFromContext(t *testing.T) {
...
 		}()
 
-		conn, err := grpchelpers.DialAndWait(
-			t.Context(),
-			"",
-			grpc.WithContextDialer(func(context.Context, string) (net.Conn, error) {
-				return listener.Dial()
-			}),
-			grpc.WithTransportCredentials(insecure.NewCredentials()),
-		)
```
tstirrat15, line 152: All of this is now pushed down into the helper.

internal/dispatch/combined/combined.go
```diff
@@ -275,14 +274,14 @@ func NewDispatcher(options ...Option) (dispatch.Dispatcher, error) {
...
 
-		conn, err := grpchelpers.Dial(context.Background(), opts.upstreamAddr, opts.grpcDialOpts...)
+		conn, err := grpchelpers.Dial(opts.upstreamAddr, opts.grpcDialOpts...)
 		if err != nil {
 			return nil, err
 		}
 
 		secondaryClients := make(map[string]remote.SecondaryDispatch, len(opts.secondaryUpstreamAddrs))
 		for name, addr := range opts.secondaryUpstreamAddrs {
-			secondaryConn, err := grpchelpers.Dial(context.Background(), addr, opts.grpcDialOpts...)
```
tstirrat15, line 285: This no longer takes `Context` as an arg because `grpc.NewClient` doesn't take context as an arg.

internal/grpchelpers/grpchelpers.go
```diff
@@ -2,19 +2,32 @@ package grpchelpers
...
+}
+
+// DialBuffered creates a client connection to an in-memory bufconn listener. Insecure
+// transport credentials are used by default; pass grpc.WithTransportCredentials to override.
+// RPCs will wait for the connection to be ready, handling the startup race between server
+// and client in tests. Additional opts are applied after the defaults, so caller-provided
+// options take precedence.
+func DialBuffered(listener *bufconn.Listener, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
+	opts = append([]grpc.DialOption{
+		grpc.WithDefaultServiceConfig(waitForReadyConfig),
```
tstirrat15, line 21: This is what should mean that we don't need `WithBlock` - this should functionally do the same thing.

e2e/spice/spicedb.go
```diff
@@ -114,8 +113,7 @@ func (s *Node) Connect(ctx context.Context, out io.Writer) error {
 	addr := net.JoinHostPort("localhost", strconv.Itoa(s.GrpcPort))
 	e2e.WaitForServerReady(addr, out)
 
-	conn, err := grpchelpers.DialAndWait(
-		ctx,
+	conn, err := grpc.NewClient(
```
tstirrat15, line 116: `DialAndWait` was an antipattern; `NewClient` is the preferred approach.

internal/grpchelpers/grpchelpers.go
```diff
@@ -2,19 +2,27 @@ package grpchelpers
...
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/test/bufconn"
 )
 
-// DialAndWait creates a new client connection to the target and blocks until the connection is ready.
-func DialAndWait(ctx context.Context, target string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	// TODO: move to NewClient
-	opts = append(opts, grpc.WithBlock())         // nolint: staticcheck
```
tstirrat15, line 12: This is an antipattern according to the docs, so we remove it and unalias things.

internal/services/integrationtesting/certtest/cert_test.go
```diff
@@ -182,7 +184,7 @@ func TestCertRotation(t *testing.T) {
 				},
 			},
 		}),
-	).Complete(ctx)
+	).CompleteForTesting(ctx)
```
tstirrat15: This is the new interface and the new method - it returns the listeners.

internal/services/integrationtesting/certtest/cert_test.go
```diff
@@ -191,11 +193,16 @@ func TestCertRotation(t *testing.T) {
 		wait <- err
 	}()
 
+	tlsCreds, err := grpcutil.WithCustomCerts(grpcutil.SkipVerifyCA, caFile.Name())
+	require.NoError(t, err)
+
 	// If previous code takes more than initialValidDuration*2 to execute, the cert
 	// would have expired, and Dial would retry indefinitely, hence the context timeout
-	dialCtx, cancelDial := context.WithTimeout(ctx, initialValidDuration*2)
```
tstirrat15, line 196: No longer necessary because `NewBufferedClient` doesn't take context as an arg.

pkg/cmd/server/server.go
```diff
@@ -779,24 +778,7 @@ type completedServerConfig struct {
 	telemetryReporter  telemetry.Reporter
 	healthManager      health.Manager
 
-	presharedKeys []string
-	closeFunc     func() error
-}
-
-func (c *completedServerConfig) GRPCDialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
```
tstirrat15, line 786: This is now wrapped as `NewClient`

pkg/cmd/server/server.go
```diff
@@ -178,6 +176,10 @@ type Config struct {
 // if there is no error, a completedServerConfig (with limited options for
 // mutation) is returned.
 func (c *Config) Complete(ctx context.Context) (RunnableServer, error) {
+	return c.complete(ctx)
+}
+
+func (c *Config) complete(ctx context.Context) (*completedServerConfig, error) {
```
tstirrat15, line 183: This is also invoked by `CompleteForTesting`

tstirrat15 (commented):
See comments

e2e/spice/spicedb.go
tstirrat15: (I changed this)

internal/testserver/cluster.go
```diff
@@ -95,6 +94,9 @@ func (b *SafeManualResolverBuilder) Build(target resolver.Target, cc resolver.Cl
 		addrs:  addrs,
 	}
 	b.resolvers.Store(target.URL.Hostname(), r)
+	// Push addresses immediately so the gRPC channel can proceed without
+	// requiring a separate ResolveNow call after the resolver is built.
+	r.ResolveNow(resolver.ResolveNowOptions{})
```
tstirrat15, line 99: There were resolution failures in tests without this.

internal/testserver/cluster.go
```diff
@@ -171,6 +173,7 @@ func TestClusterWithDispatch(t testing.TB, size uint, ds datastore.Datastore, ad
 			combineddispatch.PrometheusSubsystem(fmt.Sprintf("%s_%d_client_dispatch", prefix, i)),
 			combineddispatch.QueryPlanMetadata(queryPlanMetadata),
 			combineddispatch.GrpcDialOpts(
+				grpc.WithDefaultCallOptions(grpc.WaitForReady(true)),
```
tstirrat15, line 176: This also seems to reduce test failures.

tstirrat15 (commented):
Add more comments

pkg/cmd/util/util.go
```diff
@@ -234,15 +208,11 @@ func (c *completedGRPCServer) Listen(ctx context.Context) error {
...
-// DialContext starts a connection to grpc server
-func (c *completedGRPCServer) DialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	opts = append(opts, grpc.WithTransportCredentials(c.creds))
-	return c.dial(ctx, opts...)
-}
-
-// NetDialContext returns a low level net.Conn connection to the server
-func (c *completedGRPCServer) NetDialContext(ctx context.Context, s string) (net.Conn, error) {
-	return c.netDial(ctx, s)
+// BufferedListener returns the in-memory listener if the server is running in BufferedNetwork
```
miparnisari: nit: since `Complete` returns an interface and not the concrete object (i dislike this, but not the point of this comment), i'd prefer if we put all the godocs on the interface methods and not on the concrete methods. It's a better dev experience

pkg/cmd/server/middleware_test.go
```diff
@@ -376,10 +378,10 @@ func TestMiddlewareOrdering(t *testing.T) {
 			Enabled: true,
 		}),
 	)
-	rs, err := c.Complete(ctx)
+	rs, listeners, err := c.CompleteForTesting(ctx)
 	require.NoError(t, err)
 
-	clientConn, err := rs.GRPCDialContext(ctx)
+	clientConn, err := grpchelpers.NewBufferedClient(listeners.GRPC, grpcutil.WithInsecureBearerToken("psk"))
```
miparnisari: why `grpcutil.WithInsecureBearerToken("psk")`? that wasn't there before

pkg/cmd/server/server.go
```diff
@@ -759,11 +760,9 @@ func (c *Config) initializeGateway(ctx context.Context) (util.RunnableHTTPServer
 	return gatewayServer, closeableGatewayHandler, nil
 }
 
-// RunnableServer is a spicedb service set ready to run
+// RunnableServer is a spicedb service set ready to run.
 type RunnableServer interface {
 	Run(ctx context.Context) error
-	GRPCDialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error)
-	DispatchNetDialContext(ctx context.Context, s string) (net.Conn, error)
```
miparnisari, line 766: this is a breaking change for consumers. what should they migrate to?

please udpate the godoc of `Run` to include this

pkg/cmd/server/middleware_test.go
tstirrat15: It was previously down in the `GRPCDialContext` implementation for this `RunnableServer`. It's now provided as an input.

pkg/cmd/util/util.go
tstirrat15: Werks for me

internal/grpchelpers/grpchelpers.go
```diff
@@ -2,19 +2,27 @@ package grpchelpers
...
-	return grpc.DialContext(ctx, target, opts...) // nolint: staticcheck
-}
+const waitForReadyConfig = `{"methodConfig": [{"name": [{}], "waitForReady": true}]}`
```
miparnisari, line 12: for a follow up PR: is there some kind of validation that we can run on this json? for example: what if you had a typo and wrote `waitorReady`, would something break loudly today?

pkg/cmd/util/util.go
```diff
@@ -201,20 +178,23 @@ func (c *GRPCServerConfig) clientCreds() (credentials.TransportCredentials, erro
 }
 
 type RunnableGRPCServer interface {
+	// Listen runs a configured server
 	Listen(ctx context.Context) error
```
miparnisari: Why isn't this called `Run` 🤣

pkg/cmd/util/util.go
```diff
@@ -201,20 +178,23 @@ func (c *GRPCServerConfig) clientCreds() (credentials.TransportCredentials, erro
...
-	DialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error)
-	NetDialContext(ctx context.Context, s string) (net.Conn, error)
+	// BufferedListener returns the in-memory listener if the server is running in BufferedNetwork
+	// mode, or nil for real network servers.
```
miparnisari, line 185: re: "or nil for real network servers"

is this true? look at line 120

internal/services/integrationtesting/certtest/cert_test.go
```diff
@@ -191,11 +193,16 @@ func TestCertRotation(t *testing.T) {
 		wait <- err
 	}()
 
+	tlsCreds, err := grpcutil.WithCustomCerts(grpcutil.SkipVerifyCA, caFile.Name())
```
miparnisari: 1. this test is asserting that certificate rotation works, why add `SkipVerifyCA`? 
2. for a follow up PR: this entire test is wrong. It's not proving that the rotated cert is getting served to clients. To really prove that cert rotation works, we should wait until the cert gets rotated, then create a brand new connection (to force a new tls handshake) and then issue a CheckPermission.

You can prove (2) by temporarily disabling the hot reloading in `updateCachedCertificate`. The test still passes.

internal/grpchelpers/grpchelpers.go
tstirrat15: Not that I know of - I think the gRPC stack will emit a log to that effect, but it's the opposite of loud.

internal/services/integrationtesting/certtest/cert_test.go
tstirrat15: Good catch; that was definitely wrong. I changed it to `VerifyCA` and the test still passes. I'll make a note to add an issue for #2.

pkg/cmd/server/server.go
tstirrat15: I added `NewClient`; does that work?

pkg/cmd/server/server.go
```diff
@@ -759,18 +762,22 @@ func (c *Config) initializeGateway(ctx context.Context) (util.RunnableHTTPServer
...
 }
 
 // completedServerConfig holds the full configuration to run a spicedb server,
 // but is assumed have already been validated via `Complete()` on Config.
 // It offers limited options for mutation before Run() starts the services.
 type completedServerConfig struct {
 	ds datastore.Datastore
+	// grpcAddr carries the address for the gRPC server
+	// so that `NewClient` can reference it
+	grpcAddr string
```
tstirrat15: Using this for passthrough feels a little janky. Lmk if you have a better way of doing this.

pkg/cmd/util/util.go
tstirrat15: Right - see https://github.com/authzed/spicedb/pull/3147/changes#diff-a6dd8837b02e468bf9c13ffff426f27ff542b8823c8664747aa664dcdfe0a735R218. If that cast fails it'll be nil.

tstirrat15 (commented):
See comments

pkg/cmd/testserver/testserver.go
```diff
@@ -250,10 +255,9 @@ func (c *completedTestServer) Run(ctx context.Context) error {
...
 
-func (c *completedTestServer) GRPCDialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	return c.gRPCServer.DialContext(ctx, opts...)
-}
-
-func (c *completedTestServer) ReadOnlyGRPCDialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
-	return c.readOnlyGRPCServer.DialContext(ctx, opts...)
+// NewClient as implemented by TestServer just provides a handle on the address,
+// under the assumption that most folks won't be running this in a context where
+// running through a buffcon is necessary.
```
miparnisari: nit - this godoc is a little weird. Imagine reading this without knowing about the existence of this PR. 😄

pkg/cmd/server/server.go
miparnisari: you could add a new method `address()` to `util.RunnableGRPCServer` and then replace `grpc.NewClient(c.grpcAddr` with `grpc.NewClient(c.gRPCServer.address()`

then you can get rid of `grpcAddr` in `completedServerConfig` and in `completedTestServer`

pkg/cmd/server/server.go
```diff
@@ -759,18 +762,22 @@ func (c *Config) initializeGateway(ctx context.Context) (util.RunnableHTTPServer
...
 }
 
-// RunnableServer is a spicedb service set ready to run
+// RunnableServer is a spicedb service set ready to run.
 type RunnableServer interface {
+	// Run takes the runnable server configuration and starts the server.
 	Run(ctx context.Context) error
-	GRPCDialContext(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error)
-	DispatchNetDialContext(ctx context.Context, s string) (net.Conn, error)
+	// NewClient returns a grpc client set up to connect to the server.
```
miparnisari:
```suggestion
	// NewClient returns a grpc connection to this server.
	// It is up to the caller to close the connection when done using it.
```

pkg/cmd/server/server.go
```diff
@@ -759,18 +762,22 @@ func (c *Config) initializeGateway(ctx context.Context) (util.RunnableHTTPServer
...
+	NewClient(opts ...grpc.DialOption) (*grpc.ClientConn, error)
```
miparnisari, line 770:
```suggestion
	NewClientConnection(opts ...grpc.DialOption) (*grpc.ClientConn, error)
```

pkg/cmd/util/util.go
miparnisari: oh lol i was looking at the wrong method

miparnisari (approved):

miparnisari (dismissed):
very nice!

miparnisari (approved):

PR #3148, opened 2026-06-02
refactor: cleanup in-progress keys and readvising

barakmich:
## Description

This PR cleans up the now-unneccessary InProgressKeys field in the request proto, as the serialized subplan is now being sent in the proto.

Then, a performance gain, to re-advise on dispatch (perhaps flipping arrows based on local statistics) and also choosing whether to dispatch based on operation (for now, Check benefits the most)

👋


barakmich:
I probably won't be cleaning this one up, but I wanted to pass on the last things I had in the queue. Take care, y'all 🖖

PR #3156, opened 2026-06-05
fix: set library-friendly defaults on server Config

### LINKED ISSUE #2502: Missing defaults in configs when using SpiceDB as a library
To build `options` objects, this repo uses `defaults.MustSet(options)` quite a lot. However, the `defaults` library needs the fields of `options` to be annotated with the `default` tag, and I can't see that being the case in many of the config structs in this repo.

For example, here is the config to create a new server: https://github.com/authzed/spicedb/blob/a0e2a476bb4101d86040bf82a1e534aaff71f728/pkg/cmd/server/server.go#L56-L153. 

If anyone tries to use SpiceDB as a library (as shown in this example: https://github.com/authzed/examples/blob/main/spicedb-as-library/main.go#L107-L122), they will get whatever defaults we have specified above, or the zero value:
- ShutdownGracePeriod will be 0
- NamespaceCacheEnabled will be `false`
- StreamingAPITimeout will be 0
- etc

So I propose:
1. We should add the defaults with the values shown in https://authzed.com/docs/spicedb/concepts/commands#reference-spicedb-serve, here and in any other `struct` that is potentially user-facing.
   - Follow-up: can we enforce via tests that every field has to have a `default`?
   - Follow-up: can we somehow ensure that there is no drift between defaults in the various places where defaults are set?  E.g.
         - https://github.com/authzed/spicedb/blob/c2b3b59d138f041f9d8e990801ec074f1e2932c0/pkg/cmd/serve.go#L106
         - https://github.com/authzed/spicedb/blob/b784d2af3e45ba3d75cd856c441a6f0a3fd039f4/internal/services/v1/relationships.go#L123
1. We should update the docs to say, for example, what a `0` value means. Eg. if `MaxCaveatContextSize=0`, does this mean that the limit is infinite, or that you cannot use caveats?

ivanauth:
## Changes

* Add `default:"..."` tags for scalar `Config` fields whose CLI flag
  default in `pkg/cmd/serve.go` is non-zero.

Closes #2502

### DIFF
--- a/internal/services/integrationtesting/certtest/cert_test.go
+++ b/internal/services/integrationtesting/certtest/cert_test.go
@@ -184,12 +184,7 @@ func TestCertRotation(t *testing.T) {
 			},
 		}),
 	)
-	// Disable all caching and metrics so this test server retains its
-	// pre-PR behavior. The new library defaults enable real caches
-	// (matching CLI), which (a) break parallel subtests by registering
-	// duplicate metrics with prometheus.DefaultRegisterer and (b) can
-	// affect test semantics by introducing caching where there was
-	// previously a NoopCache.
+	// Disable caches and their metrics to avoid "duplicate metrics" errors
 	cfg.DispatchClusterMetricsEnabled = false
 	cfg.DispatchClientMetricsEnabled = false
 	cfg.DatastoreConfig.EnableDatastoreMetrics = false
--- a/internal/services/integrationtesting/certtest/cert_test.go
+++ b/internal/services/integrationtesting/certtest/cert_test.go
@@ -126,7 +126,7 @@ func TestCertRotation(t *testing.T) {
 	require.NoError(t, err)
 
 	ctx, cancel := context.WithCancel(t.Context())
-	srv, err := server.NewConfigWithOptionsAndDefaults(
+	cfg := server.NewConfigWithOptionsAndDefaults(
 		server.WithDatastore(ds),
 		server.WithDispatcher(dispatcher),
 		server.WithDispatchMaxDepth(50),
@@ -183,7 +183,22 @@ func TestCertRotation(t *testing.T) {
 				},
 			},
 		}),
-	).Complete(ctx)
+	)
+	// Disable all caching and metrics so this test server retains its
+	// pre-PR behavior. The new library defaults enable real caches
+	// (matching CLI), which (a) break parallel subtests by registering
+	// duplicate metrics with prometheus.DefaultRegisterer and (b) can
+	// affect test semantics by introducing caching where there was
+	// previously a NoopCache.
+	cfg.DispatchClusterMetricsEnabled = false
+	cfg.DispatchClientMetricsEnabled = false
+	cfg.DatastoreConfig.EnableDatastoreMetrics = false
+	cfg.NamespaceCacheConfig = server.CacheConfig{}
+	cfg.DispatchCacheConfig = server.CacheConfig{}
+	cfg.ClusterDispatchCacheConfig = server.CacheConfig{}
+	cfg.LR3ResourceChunkCacheConfig = server.CacheConfig{}
+	cfg.StoredSchemaCacheConfig = server.CacheConfig{}
+	srv, err := cfg.Complete(ctx)
 	require.NoError(t, err)
 
 	wait := make(chan error, 1)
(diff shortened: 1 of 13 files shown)


miparnisari (approved):

pkg/cmd/server/server.go
```diff
@@ -65,7 +65,7 @@ type Config struct {
 	GRPCServer             util.GRPCServerConfig `debugmap:"visible"`
```
miparnisari, line 64: Can we also add defaults to every field of 
- `util.GRPCServerConfig`
- `datastorecfg.Config`
- `ConnPoolConfig`
- `CacheConfig`
- `OTelConfig`

and every other struct embedded in the server's config?

pkg/cmd/server/server_test.go
```diff
@@ -81,7 +81,35 @@ func TestServerGracefulTermination(t *testing.T) {
 func TestServerDefaultOptions(t *testing.T) {
 	cfg := NewConfigWithOptionsAndDefaults()
 
+	require.Equal(t, 5*time.Second, cfg.ShutdownGracePeriod)
+	require.Equal(t, 4096, cfg.MaxCaveatContextSize)
+	require.Equal(t, 25_000, cfg.MaxRelationshipContextSize)
+	require.Equal(t, time.Second, cfg.SchemaWatchHeartbeat)
+	require.Equal(t, "read-legacy-write-legacy", cfg.ExperimentalSchemaMode)
```
miparnisari, line 88: please extend this test to check every embedded struct defaults. They should match existing defaults.

miparnisari (changes requested):

miparnisari:
> Closes https://github.com/authzed/spicedb/issues/2502

To fully close that issue, i would expect to see a test (or an architectural change) that ensures that the defaults added via a `default:...` tag can never drift from the defaults specified in https://github.com/authzed/spicedb/blob/6f3a955a1f858cdc05a3097bdbda3db245b35873/internal/services/v1/relationships.go#L151-L163 and the defaults specified in `RegisterServeFlags`, for example:

https://github.com/authzed/spicedb/blob/af55756f28557ecc5059e8724aff98abfe2a939b/pkg/cmd/serve.go#L120

miparnisari (dismissed):
Love it! Thanks for sending this

miparnisari:
uh-oh

```
Run go run mage.go testcons:postgresver 18
NO RESOLVER YET
--- FAIL: TestConsistencyPerDatastore (22.14s)
    --- FAIL: TestConsistencyPerDatastore/postgres (22.14s)
panic: two caches with the same name: namespace [recovered, repanicked]

goroutine 79 [running]:
testing.tRunner.func1.2({0x2b6d720, 0xc0023a6730})
	/opt/hostedtoolcache/go/1.26.4/arm64/src/testing/testing.go:1974 +0x2b8
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.26.4/arm64/src/testing/testing.go:1977 +0x460
panic({0x2b6d720?, 0xc0023a6730?})
```

miparnisari (dismissed):

internal/services/integrationtesting/certtest/cert_test.go
```diff
@@ -183,7 +183,22 @@ func TestCertRotation(t *testing.T) {
...
 			},
 		}),
-	).Complete(ctx)
+	)
+	// Disable all caching and metrics so this test server retains its
+	// pre-PR behavior. The new library defaults enable real caches
+	// (matching CLI), which (a) break parallel subtests by registering
+	// duplicate metrics with prometheus.DefaultRegisterer and (b) can
+	// affect test semantics by introducing caching where there was
+	// previously a NoopCache.
```
miparnisari:
```suggestion
	// Disable caches and their metrics to avoid "duplicate metrics" errors
```

internal/testserver/cluster.go
```diff
@@ -229,6 +229,20 @@ func TestClusterWithDispatch(t testing.TB, size uint, ds datastore.Datastore, ad
 
 		ctx, cancel := context.WithCancel(t.Context())
 		cfg := server.NewConfigWithOptionsAndDefaults(serverOptions...)
+		// Disable all caching and metrics so test clusters retain their
+		// pre-PR behavior. The new library defaults enable real caches
+		// (matching CLI), which (a) breaks parallel subtests by registering
+		// duplicate metrics with prometheus.DefaultRegisterer and (b) can
+		// affect test semantics by introducing caching where there was
+		// previously a NoopCache.
```
miparnisari:
```suggestion
		// Disable caches and their metrics to avoid "duplicate metrics" errors
```

miparnisari (approved):

PR #3160, opened 2026-06-08
feat(datalayer): schema caching for write transactions

ecordell:
## Description

With the experimental schema storage enabled, I was noticing a significant load on `ReadStoredSchema` due to the write path.

This PR addresses it by storing a "last seen schema hash". When performing writes, the last schema hash is included in a guard in the transaction, and if it differs from what is stored on the server, the write tx is aborted. Basically this adds optimistic concurrency around schema changes.

I ran into several related issues when validating this:

- `WriteStoredSchema` updated the cached version before committing; which means a failed schema write would poison the cache with a bad entry.
- `ErrSchemaNotFound` was swallowed and replaced with an "empty" schema in some cases; I couldn't find a motivation for this and it seems to make error messages strictly worse so I changed this.
- The "old" schema caching proxy was unconditionally installed; I updated to avoid installing it if in new schema mode.

## Testing

Added unit tests for the issues I listed above, and built and ran this in staging and watched the ReadStoredSchema requests drop to ~0 under write load.

### DIFF
--- a/internal/datastore/crdb/storedschema.go
+++ b/internal/datastore/crdb/storedschema.go
@@ -34,25 +34,31 @@ func (cr *crdbReader) ReadStoredSchema(ctx context.Context) (*datastore.ReadOnly
 	return rw.ReadStoredSchema(ctx)
 }
 
-// ReadStoredSchemaHash reads only the schema hash from the schema_revision table.
-func (rwt *crdbReadWriteTXN) ReadStoredSchemaHash(ctx context.Context) (string, error) {
-	sql, args, err := psql.Select("hash").
+// assertSchemaHash verifies the schema_revision row matches expectedHash.
+// Under CRDB's default SERIALIZABLE isolation, both FOR UPDATE and FOR SHARE are unreplicated
+// best-effort locks that cannot be relied upon for correctness. Correctness for concurrent
+// schema and relationship writes is instead guaranteed by CRDB's serializable conflict
+// detection at commit time. The exclusive flag is ignored.
+func assertSchemaHash(ctx context.Context, tx pgx.Tx, expectedHash string) error {
+	q := psql.Select("hash").
 		From("schema_revision").
-		Where(sq.Eq{"name": "current"}).
-		Limit(1).
-		ToSql()
+		Where(sq.Eq{"name": "current"})
+	sqlStr, args, err := q.ToSql()
 	if err != nil {
-		return "", fmt.Errorf("failed to build schema hash query: %w", err)
+		return fmt.Errorf("failed to build schema hash precondition query: %w", err)
 	}
 
-	var hash []byte
-	if err := rwt.tx.QueryRow(ctx, sql, args...).Scan(&hash); err != nil {
+	var storedHash []byte
+	if err := tx.QueryRow(ctx, sqlStr, args...).Scan(&storedHash); err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
-			return "", datastore.ErrSchemaNotFound
+			return datastore.ErrSchemaNotFound
 		}
-		return "", fmt.Errorf("failed to read schema hash: %w", err)
+		return fmt.Errorf("failed to check schema hash precondition: %w", err)
 	}
-	return string(hash), nil
+	if string(storedHash) != expectedHash {
+		return datastore.ErrSchemaHashPreconditionFailed
+	}
+	return nil
 }
 
 // WriteStoredSchema writes the unified stored schema to the CRDB schema table.
--- a/internal/datastore/crdb/storedschema.go
+++ b/internal/datastore/crdb/storedschema.go
@@ -2,9 +2,11 @@ package crdb
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	sq "github.com/Masterminds/squirrel"
+	"github.com/jackc/pgx/v5"
 
 	"github.com/authzed/spicedb/internal/datastore/common"
 	"github.com/authzed/spicedb/pkg/datastore"
@@ -30,6 +32,27 @@ func (cr *crdbReader) ReadStoredSchema(ctx context.Context) (*datastore.ReadOnly
 
 	rw := common.NewSQLSingleStoreSchemaReaderWriterWithBuiltInMVCC(chunker)
 	return rw.ReadStoredSchema(ctx)
+}
+
+// ReadStoredSchemaHash reads only the schema hash from the schema_revision table.
+func (rwt *crdbReadWriteTXN) ReadStoredSchemaHash(ctx context.Context) (string, error) {
+	sql, args, err := psql.Select("hash").
+		From("schema_revision").
+		Where(sq.Eq{"name": "current"}).
+		Limit(1).
+		ToSql()
+	if err != nil {
+		return "", fmt.Errorf("failed to build schema hash query: %w", err)
+	}
+
+	var hash []byte
+	if err := rwt.tx.QueryRow(ctx, sql, args...).Scan(&hash); err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return "", datastore.ErrSchemaNotFound
+		}
+		return "", fmt.Errorf("failed to read schema hash: %w", err)
+	}
+	return string(hash), nil
 }
 
 // WriteStoredSchema writes the unified stored schema to the CRDB schema table.
--- a/internal/datastore/memdb/memdb.go
+++ b/internal/datastore/memdb/memdb.go
@@ -62,7 +62,7 @@ func NewMemdbDatastore(
 	}
 
 	uniqueID := uuid.NewString()
-	return &memdbDatastore{
+	mdb := &memdbDatastore{
 		CommonDecoder: revisions.CommonDecoder{
 			Kind: revisions.Timestamp,
 		},
@@ -80,7 +80,9 @@ func NewMemdbDatastore(
 		watchBufferLength:       watchBufferLength,
 		watchBufferWriteTimeout: 100 * time.Millisecond,
 		uniqueID:                uniqueID,
-	}, nil
+	}
+	mdb.writeTxReady = sync.NewCond(&mdb.RWMutex)
+	return mdb, nil
 }
 
 type memdbDatastore struct {
@@ -91,6 +93,7 @@ type memdbDatastore struct {
 	db             *memdb.MemDB // GUARDED_BY(RWMutex)
 	revisions      []snapshot   // GUARDED_BY(RWMutex)
 	activeWriteTxn *memdb.Txn   // GUARDED_BY(RWMutex)
+	writeTxReady   *sync.Cond   // broadcast when activeWriteTxn becomes nil
 
 	negativeGCWindow        int64
 	quantizationPeriod      int64
@@ -192,9 +195,10 @@ func (mdb *memdbDatastore) ReadWriteTx(
 				mdb.Lock()
 				defer mdb.Unlock()
 
-				if mdb.activeWriteTxn != nil {
-					err = ErrSerialization
-					return
+				// Block until any active write transaction finishes rather than
+				// returning ErrSerialization and busy-retrying with sleeps.
+				for mdb.activeWriteTxn != nil {
+					mdb.writeTxReady.Wait()
 				}
 
 				if err = mdb.checkNotClosed(); err != nil {
@@ -211,25 +215,34 @@ func (mdb *memdbDatastore) ReadWriteTx(
 
 		newRevision := mdb.newRevisionID()
 		rwt := &memdbReadWriteTx{memdbReader{&sync.Mutex{}, txSrc, nil, time.Now()}, newRevision}
+		if config.SchemaHashPrecondition != "" {
+			if err := assertSchemaHash(ctx, rwt, config.SchemaHashPrecondition); err != nil {
+				mdb.Lock()
+				if tx != nil {
+					tx.Abort()
+					mdb.activeWriteTxn = nil
+					mdb.writeTxReady.Signal()
+				}
+				mdb.Unlock()
+				return datastore.NoRevision, err
+			}
+		}
 		if err := f(ctx, rwt); err != nil {
 			mdb.Lock()
 			if tx != nil {
 				tx.Abort()
 				mdb.activeWriteTxn = nil
+				mdb.writeTxReady.Signal()
 			}
 
-			// If the error was a serialization error, retry the transaction
+			// If the error was a serialization error, retry the transaction.
+			// We *must* return the inner error unmodified in case it's not an error type
+			// that supports unwrapping (e.g. gRPC errors)
 			if errors.Is(err, ErrSerialization) {
 				mdb.Unlock()
-
-				// If we don't sleep here, we run out of retries instantaneously
-				time.Sleep(1 * time.Millisecond)
 				continue
 			}
 			defer mdb.Unlock()
-
-			// We *must* return the inner error unmodified in case it's not an error type
-			// that supports unwrapping (e.g. gRPC errors)
 			return datastore.NoRevision, err
 		}
 
@@ -352,6 +365,7 @@ func (mdb *memdbDatastore) ReadWriteTx(
 			tx.Commit()
 		}
 		mdb.activeWriteTxn = nil
+		mdb.writeTxReady.Signal()
 
 		if err := mdb.checkNotClosed(); err != nil {
 			return datastore.NoRevision, err
--- a/internal/datastore/memdb/memdb_test.go
+++ b/internal/datastore/memdb/memdb_test.go
@@ -28,7 +28,9 @@ func (memDBTest) New(_ testing.TB, revisionQuantization, _, gcWindow time.Durati
 }
 
 func TestMemdbDatastore(t *testing.T) {
-	test.All(t, memdbFactory.NewTester(memDBTest{}))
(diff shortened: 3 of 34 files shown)


pkg/datalayer/schema_adapter.go
```diff
@@ -541,15 +527,8 @@ func WriteSchemaViaStoredSchema(ctx context.Context, rwt datastore.ReadWriteTran
 	}
 
 	if err := rwt.WriteStoredSchema(ctx, storedSchema); err != nil {
-		return "", err
-	}
-
-	// Update cache after successful write
```
ecordell, line 547: misleading comment here - the "write" was successful but the commit hasn't happened yet

internal/datastore/memdb/storedschema.go
```diff
@@ -42,6 +42,37 @@ func (r *memdbReader) ReadStoredSchema(_ context.Context) (*datastore.ReadOnlySt
 	return datastore.NewReadOnlyStoredSchema(storedSchema), nil
 }
 
+// ReadStoredSchemaHash reads only the schema hash from the in-memory schema revision table.
+func (rwt *memdbReadWriteTx) ReadStoredSchemaHash(_ context.Context) (string, error) {
+	rwt.mustLock()
+	defer rwt.Unlock()
```
tstirrat15, line 49: Is there a read-only mutex? Would that be worth using here?

internal/datastore/postgres/storedschema.go
```diff
@@ -28,6 +29,29 @@ func (r *pgReader) ReadStoredSchema(ctx context.Context) (*datastore.ReadOnlySto
...
 }
 
+// ReadStoredSchemaHash reads only the schema hash from the schema_revision table.
+func (rwt *pgReadWriteTXN) ReadStoredSchemaHash(ctx context.Context) (string, error) {
+	sql, args, err := psql.Select("hash").
+		From("schema_revision").
+		Where(sq.Eq{"name": "current"}).
+		Where(sq.Eq{"deleted_xid": liveDeletedTxnID}).
+		OrderBy("created_xid DESC").
+		Limit(1).
```
tstirrat15: Are we ever expecting there to be more than one live hash, or is this defensiveness?

internal/services/shared/schema.go
```diff
@@ -109,7 +110,10 @@ func ApplySchemaChanges(ctx context.Context, rwt datalayer.ReadWriteTransaction,
 	defer span.End()
 	sr, err := rwt.ReadSchema(ctx)
 	if err != nil {
-		return nil, err
+		if !errors.Is(err, datastore.ErrSchemaNotFound) {
+			return nil, err
+		}
+		return ApplySchemaChangesOverExisting(ctx, rwt, caveatTypeSet, validated, nil, nil)
```
tstirrat15, line 116: Which issue did this change address?

pkg/datalayer/impl.go
```diff
@@ -85,9 +85,25 @@ func (d *defaultDataLayer) SnapshotReader(rev datastore.Revision, schemaHash Sch
 }
 
 func (d *defaultDataLayer) ReadWriteTx(ctx context.Context, fn TxUserFunc, opts ...options.RWTOptionsOption) (datastore.Revision, error) {
-	return d.ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
-		return fn(ctx, &readWriteTransaction{rwt: rwt, schemaMode: d.schemaMode, cache: d.cache})
+	var pendingHash SchemaHash
+	var pendingSchema *datastore.ReadOnlyStoredSchema
```
tstirrat15: Sanity check: are transactions always handled by a single thread? i.e. are there thread safety concerns in this context?

tstirrat15: Imagining something where you race two goroutines within a transaction, both of which are executing a schema write. I realize that sounds like it would be insane calling code.

tstirrat15 (dismissed):
LGTM; see comments

internal/datastore/memdb/storedschema.go
ecordell: this lock is for serializing access to the rwt itself, it's not a lock on the data being read

internal/datastore/postgres/storedschema.go
ecordell: just defensiveness for now

internal/services/shared/schema.go
ecordell: When testing I discovered that this would fail on a fresh DB without a schema written - which didn't make sense to do for a WriteSchema call

pkg/datalayer/impl.go
ecordell: yeah in general the transactions are not thread safe, though I'm not sure how good a job we do of documenting that

internal/services/shared/schema.go
tstirrat15: Oh yeah that's not ideal.

pkg/datalayer/datalayer.go
```diff
@@ -106,6 +106,12 @@ type RevisionedReader interface {
 }
 
 // TxUserFunc is a type for the function that users supply when they invoke a read-write transaction.
+//
+// When WithSchemaHashPrecondition is passed to ReadWriteTx, the datalayer may transparently retry
+// fn with a refreshed schema hash if the stored schema changed since the precondition was captured.
+// This is only safe when ALL schema-sensitive validation (ReadSchema, ValidateRelationshipUpdates,
```
tstirrat15, line 112: This scares me a little... Do we have a way to check/assert this? Would an analyzer do it?

internal/services/shared/errors.go
```diff
@@ -171,6 +171,10 @@ func rewriteError(ctx context.Context, err error, _ *ConfigForErrors) error {
 		return status.Errorf(codes.OutOfRange, "invalid zedtoken: %s", err)
 	case errors.As(err, &datastore.CaveatNameNotFoundError{}):
 		return spiceerrors.WithCodeAndReason(err, codes.FailedPrecondition, v1.ErrorReason_ERROR_REASON_UNKNOWN_CAVEAT)
+	case errors.Is(err, datastore.ErrSchemaHashPreconditionFailed):
+		return status.Errorf(codes.FailedPrecondition, "%s", err)
+	case errors.Is(err, datastore.ErrSchemaNotFound):
+		return status.Errorf(codes.NotFound, "%s", err)
```
tstirrat15, line 177: Does this error get returned on things other than `ReadSchema`, e.g. `WriteRels`? If so, it seems like it should be `NotFound` in some cases and `FailedPrecondition` in others. Not sure how difficult this would be to implement, though.

internal/datastore/crdb/storedschema.go
```diff
@@ -32,6 +34,31 @@ func (cr *crdbReader) ReadStoredSchema(ctx context.Context) (*datastore.ReadOnly
 	return rw.ReadStoredSchema(ctx)
 }
 
+// assertSchemaHash acquires a FOR UPDATE lock on the schema_revision row and verifies
+// it matches expectedHash. Returns ErrSchemaHashPreconditionFailed if not found or mismatched.
+func assertSchemaHash(ctx context.Context, tx pgx.Tx, expectedHash string) error {
```
tstirrat15, line 42: Just to make sure I understand the mental model - these functions are more or less acquiring some kind of lock on the datastore, and then a subsequent call that invokes this same function will fail (e.g. because of the locking behavior of `SELECT FOR UPDATE`), and that subsequent call will be retried?

internal/datastore/memdb/memdb.go
```diff
@@ -211,6 +211,22 @@ func (mdb *memdbDatastore) ReadWriteTx(
...
+		if config.SchemaHashPrecondition != "" {
+			if err := assertSchemaHash(ctx, rwt, config.SchemaHashPrecondition); err != nil {
+				mdb.Lock()
+				if tx != nil {
+					tx.Abort()
+					mdb.activeWriteTxn = nil
+				}
+				if errors.Is(err, ErrSerialization) {
+					mdb.Unlock()
+					time.Sleep(1 * time.Millisecond)
```
tstirrat15: Is this necessary/is there another way to get the same behavior?

pkg/datalayer/impl.go
```diff
@@ -85,9 +109,76 @@ func (d *defaultDataLayer) SnapshotReader(rev datastore.Revision, schemaHash Sch
...
+	if preconditionHash == "" && d.schemaMode.ReadsFromNew() {
+		if h := d.loadLastSchemaHash(); h != "" {
+			preconditionHash = h
+		} else if result, err := d.ds.HeadRevision(ctx); err == nil && result.SchemaHash != "" {
+			preconditionHash = SchemaHash(result.SchemaHash)
+			d.observeSchemaHash(preconditionHash)
+		}
+	}
+
+	for attempt := 0; ; attempt++ {
```
tstirrat15: Is there a reason this isn't `range maxSchemaHashRetries`?

pkg/datalayer/impl.go
```diff
@@ -85,9 +109,76 @@ func (d *defaultDataLayer) SnapshotReader(rev datastore.Revision, schemaHash Sch
...
+
+		// If preconditionHash was refreshed on a prior retry, override the option so
+		// the underlying datastore's assertSchemaHash runs against the new hash.
+		dsOpts := opts
+		if SchemaHash(rwtOpts.SchemaHashPrecondition) != preconditionHash {
+			overridden := make([]options.RWTOptionsOption, len(opts)+1)
+			copy(overridden, opts)
+			overridden[len(opts)] = options.WithSchemaHashPrecondition(string(preconditionHash))
+			dsOpts = overridden
+		}
```
tstirrat15: Is this the "standard" way to do this kind of logic? I'm kind of surprised that we're not just mutating the ds, or else appending without copying, since the last option will win.

tstirrat15 (commented):
See comments, otherwise LGTM

internal/datastore/crdb/storedschema.go
ecordell: yes, I updated this to do `FOR UPDATE` only for schema writes, not for rel writes

Rel writes use `FOR SHARE` which just serializes them around an in flight `FOR UPDATE`. The only reason this is needed at all is because we have a flag that lets you use repeatable read, otherwise the fact that the DBs are serializable should make this work with just a read.

I added tests to confirm these behaviors.

pkg/datalayer/datalayer.go
ecordell: I can't think of an easy way to validate this, maybe an analyzer. 

It's not really different semantics than before - this schema validation step still happened before my change - just documenting it for future uses.

Maybe I can make a followup?

internal/services/shared/errors.go
ecordell: Yeah it would get returned from writerels if there hasn't been a schema write yet, but to me that doesn't sound too unreasonable

On main, `newStoredSchemaReaderAdapter` catches `ErrSchemaNotFound` and returns an empty schema adapter with no error. That's behavior I changed in this PR (mentioned in the PR desc) because I couldn't understand a reason for it. So on main, a rel write with no schema would succeed at the `ReadSchema` call, then fail validation with `UnknownNamespaceError` because the namespace doesn't exist in the empty schema...IMO this is better but I can see how you'd want a preconditionfailed here.

internal/datastore/memdb/memdb.go
ecordell: refactored

pkg/datalayer/impl.go
ecordell: updated

pkg/datalayer/impl.go
ecordell: this works around a specific gocritic check that doesn't like when you append but reassign to a different variable

internal/services/shared/errors.go
tstirrat15: Yeah, I agree that it's an improvement over the previous behavior. This works for now.

internal/datastore/memdb/memdb_test.go
```diff
@@ -82,15 +84,16 @@ func TestConcurrentWritePanic(t *testing.T) {
...
 	require := require.New(t)
 
 	ds, err := NewMemdbDatastore(0, 1*time.Hour, 1*time.Hour)
 	require.NoError(err)
 
 	ctx := t.Context()
 
-	// Kick off a number of writes to ensure at least one hits an error.
+	// With sync.Cond serialization, concurrent writes block until the active write
+	// finishes rather than failing immediately, so all goroutines succeed.
```
tstirrat15, line 96: Is this a new behavior of the datastore impl, or is this test-specific behavior? If it's the latter, I don't think this test is relevant anymore - it was meant to watch/catch the behavior where sufficient write pressure would cause writes to fail.

pkg/datastore/test/storedschema.go
```diff
@@ -777,6 +782,337 @@ func OptimizedRevisionSchemaHashTest(t *testing.T, tester DatastoreTester) {
...
+	req.NoError(err)
+
+	// relWriteInside is closed when the rel write has acquired FOR SHARE and is inside fn.
+	// releaseRelWrite is closed to let the rel write commit and release FOR SHARE.
+	relWriteInside := make(chan struct{})
+	releaseRelWrite := make(chan struct{})
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
```
tstirrat15: Prefer `wg.Go`

pkg/datastore/test/storedschema.go
```diff
@@ -777,6 +782,337 @@ func OptimizedRevisionSchemaHashTest(t *testing.T, tester DatastoreTester) {
...
+			return nil
+		}, options.WithSchemaHashPrecondition(hashV1), options.WithDisableRetries(true))
+	}()
+
+	// Wait for the rel write to be inside its transaction (FOR SHARE held).
+	select {
+	case <-relWriteInside:
+	case <-time.After(10 * time.Second):
+		t.Fatal("timed out waiting for relationship write to enter its transaction")
+	}
```
tstirrat15: Could we use a context with timeout instead of `t.Fatal` here? Or are we concerned that we'd deadlock?

pkg/datastore/test/storedschema.go
```diff
@@ -777,6 +782,337 @@ func OptimizedRevisionSchemaHashTest(t *testing.T, tester DatastoreTester) {
...
+	// Release the rel write, then wait for the schema write to complete.
+	close(releaseRelWrite)
+	wg.Wait()
+
+	select {
+	case err := <-schemaWriteDone:
+		req.NoError(err, "schema write should succeed")
+	case <-time.After(10 * time.Second):
+		t.Fatal("schema write did not complete within 10s")
+	}
```
tstirrat15: Same

pkg/datastore/test/storedschema.go
```diff
@@ -777,6 +782,337 @@ func OptimizedRevisionSchemaHashTest(t *testing.T, tester DatastoreTester) {
...
+	case <-relWriteInside:
+	case <-time.After(10 * time.Second):
+		t.Fatal("timed out waiting for relationship write to enter its transaction")
+	}
+
+	// Start the schema write. On Postgres/MySQL it blocks on FOR UPDATE until the rel write
+	// releases FOR SHARE. On CRDB it proceeds immediately — correctness is guaranteed by
+	// CRDB's serializable isolation detecting the read-write conflict at commit time instead.
+	schemaWriteDone := make(chan error, 1)
+	go func() {
```
tstirrat15: `wg.Go` here too

pkg/datastore/test/storedschema.go
```diff
@@ -777,6 +782,337 @@ func OptimizedRevisionSchemaHashTest(t *testing.T, tester DatastoreTester) {
...
+	schemaTextV2, _, err := generator.GenerateSchema(ctx, defsV2)
+	req.NoError(err)
+
+	// Launch constant rel write traffic. Each goroutine writes unique relationships in a
+	// tight loop, holding FOR SHARE on the schema row for the duration of each write.
+	var attempted atomic.Int64
+	const writers = 5
+	var wg sync.WaitGroup
+	for i := range writers {
+		wg.Add(1)
```
tstirrat15: Go

pkg/datastore/test/storedschema.go
```diff
@@ -777,6 +782,337 @@ func OptimizedRevisionSchemaHashTest(t *testing.T, tester DatastoreTester) {
...
+		}, options.WithSchemaHashPrecondition(hashV1), options.WithSchemaHashPreconditionExclusive(true))
+		schemaWriteDone <- err
+	}()
+
+	select {
+	case err := <-schemaWriteDone:
+		req.NoError(err)
+	case <-time.After(10 * time.Second):
+		t.Fatal("schema write did not complete within 10s under constant rel write traffic — possible lock starvation")
+	}
```
tstirrat15: Similar here

tstirrat15 (commented):
See comments

internal/datastore/memdb/memdb_test.go
ecordell: Yeah I just switched it to test the opposite - now the parallel writes succeed because they serialize via the Cond

tstirrat15 (dismissed):
LGTM

tstirrat15 (dismissed):
LGTM

tstirrat15 (approved):
LGTM

PR #3163, opened 2026-06-08
test: address flake in HeadRevisionDoesNotConsumeXID

tstirrat15:
## Description
This test has been flaking: https://github.com/authzed/spicedb/actions/runs/27164166509/job/80187393099?pr=3154

Relaxing the assertion slightly should help.

## Changes
* Use a leq assertion instead of an equality assertion
## Testing
Review.

### DIFF
--- a/internal/datastore/postgres/postgres_shared_test.go
+++ b/internal/datastore/postgres/postgres_shared_test.go
@@ -1162,7 +1162,11 @@ func HeadRevisionDoesNotConsumeXIDTest(t *testing.T, ds datastore.Datastore) {
 	}
 	after := nextXID()
 
-	require.Equal(t, before, after, "HeadRevision burned %d xid(s) over %d calls; expected 0", after-before, iterations)
+	// NOTE: we're expecting <= 1 here because there's a heartbeat goroutine that increments
+	// the xid over time. this may or may not fire once during the test.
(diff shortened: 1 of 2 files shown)


internal/datastore/postgres/postgres_shared_test.go
```diff
@@ -1162,7 +1162,11 @@ func HeadRevisionDoesNotConsumeXIDTest(t *testing.T, ds datastore.Datastore) {
 	}
 	after := nextXID()
 
-	require.Equal(t, before, after, "HeadRevision burned %d xid(s) over %d calls; expected 0", after-before, iterations)
+	// NOTE: we're expecting <= 1 here because there's a heartbeat goroutine that increments
```
miparnisari: before we do this relaxation, can we try something else? add `cmd = append(cmd, "-c", "autovacuum=off")` in line 66 inside `RunPostgresForTestingWithCommitTimestamps`. I assume that none of the existing tests depend on this vacuum process (@josephschorr ?)

my suspicion is that the vacuum processing is what is burning this extra XID.

josephschorr: Let's try it - see if it does remove the flake

miparnisari (approved):

PR #3166, opened 2026-06-09
feat: in-process embedded permissions and per-schema-version caching

josephschorr:
## Summary

Adds `pkg/embedded`, an in-process (no gRPC) permissions library, and makes checks cheaper by caching schema-derived artifacts once per schema version.

### Changes

- **`datastore`: schema-derived caches on the stored schema.** `ReadOnlyStoredSchema` now hosts lazily-built derived caches that live exactly as long as their schema version and are discarded with it. Cache kinds self-register at init via `RegisterDerivedCache` (avoiding an import cycle back into the owning packages) and are fetched with `GetDerivedCache`. Compiled caveats today; type systems/reachability planned.
- **`caveats`: compiled caveats cached per schema version.** Deserializing a caveat rebuilds its CEL environment, previously paid on every check because each check constructs a fresh `CaveatRunner`. Now cached on the shared stored schema, keyed by caveat name; falls back to per-runner caching when the reader isn't backed by a unified stored schema.
- **`datastore`: stored-schema cache costed by bytes.** Entries were admitted at cost 1, so the byte-denominated `MaxCost` (server default 32MiB; `SchemaCacheMaxCostBytes` in embedded) bounded entry *count*, not memory. Entries are now costed by `EstimatedSize()`: schema bytes (exact serialized size when the reader has it) plus each registered derived-cache kind's estimate once populated.
- **`datastore`: bootstrap honors schema mode.** Bootstrap always wrote legacy per-definition schema, so unified-schema servers found none. New `BootstrapSchemaMode` option; zero value preserves prior behavior.
- **`embedded`: new `pkg/embedded`.** Permission checks in-process against a datastore via the dispatch engine; caveat context passed as native Go values. Includes README and tests.

### Testing

`mage lint:all` and the unit suite pass. New tests: derived-cache registry and cost estimation (`pkg/datastore`), compiled-caveat cache (`internal/caveats`), embedded checks across both schema modes (`pkg/embedded`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Updated the Prometheus buckets for `grpc_server_handling_seconds` and `spicedb_datastore_query_latency` to be able to correlate them (https://github.com/authzed/spicedb/pull/3188)
 - Use `testcontainers` instead of `ory/dockertest` for running containers in integration tests (https://github.com/authzed/spicedb/pull/2782)
 - Embedded: add `pkg/embedded`, an in-process library for running permission checks against a datastore via the dispatch engine, without standing up a gRPC server (https://github.com/authzed/spicedb/pull/3166)
-- Caveats: compiled caveats (and their CEL environments) are now cached per schema version rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
-- Datastore: schema-derived artifacts (e.g. compiled caveats) are now cached on the stored schema (`ReadOnlyStoredSchema`) and share its lifetime, so they are rebuilt only when the schema changes (https://github.com/authzed/spicedb/pull/3166)
+- Caveats: compiled caveats (and their CEL environments) are now cached per schema version — hung off the stored schema (`ReadOnlyStoredSchema`) and rebuilt only when the schema changes — rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
 
 ### Fixed
 - Fixed a nil pointer dereference panic in `CheckBulkPermissions` that could occur under concurrent load when a tracing-enabled check shared a singleflight dispatch with a non-tracing bulk check. Debug-enabled checks are no longer singleflighted together with non-debug checks. (https://github.com/authzed/spicedb/pull/3174)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Schema: reads inside write transactions now use a cheap hash-only lookup (`schema_revision`) to check the cache before loading the full schema blob, reducing DB round-trips on cache hits (https://github.com/authzed/spicedb/pull/3160)
 - Updated the Prometheus buckets for `grpc_server_handling_seconds` and `spicedb_datastore_query_latency` to be able to correlate them (https://github.com/authzed/spicedb/pull/3188)
 - Use `testcontainers` instead of `ory/dockertest` for running containers in integration tests (https://github.com/authzed/spicedb/pull/2782)
+- Embedded: add `pkg/embedded`, an in-process library for running permission checks against a datastore via the dispatch engine, without standing up a gRPC server (https://github.com/authzed/spicedb/pull/3166)
+- Caveats: compiled caveats (and their CEL environments) are now cached per schema version rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
+- Datastore: schema-derived artifacts (e.g. compiled caveats) are now cached on the stored schema (`ReadOnlyStoredSchema`) and share its lifetime, so they are rebuilt only when the schema changes (https://github.com/authzed/spicedb/pull/3166)
 
 ### Fixed
 - Fixed a nil pointer dereference panic in `CheckBulkPermissions` that could occur under concurrent load when a tracing-enabled check shared a singleflight dispatch with a non-tracing bulk check. Debug-enabled checks are no longer singleflighted together with non-debug checks. (https://github.com/authzed/spicedb/pull/3174)
--- a/internal/caveats/run.go
+++ b/internal/caveats/run.go
@@ -51,11 +51,24 @@ func RunSingleCaveatExpression(
 	return runner.RunCaveatExpression(ctx, expr, context, reader, debugOption)
 }
 
+// cachedSchemaProvider is satisfied (structurally) by SchemaReaders backed by a unified
+// stored schema. It exposes the shared, per-schema-version stored schema, which hosts
+// schema-derived caches such as the compiled-caveat cache.
+type cachedSchemaProvider interface {
+	StoredSchema() *datastore.ReadOnlyStoredSchema
+}
+
 // CaveatRunner is a helper for running caveats, providing a cache for deserialized caveats.
 type CaveatRunner struct {
 	caveatTypeSet       *caveattypes.TypeSet
 	caveatDefs          map[string]*core.CaveatDefinition
 	deserializedCaveats map[string]*caveats.CompiledCaveat
+
+	// schemaCache, when non-nil, is a compiled-caveat cache tied to the stored schema
+	// (and thus shared across checks and invalidated on schema change). It is discovered
+	// from the reader on first use. When nil, deserializedCaveats provides per-runner
+	// caching only (the legacy behavior).
+	schemaCache *CompiledCaveatCache
 }
 
 // NewCaveatRunner creates a new CaveatRunner.
@@ -90,6 +103,21 @@ func (cr *CaveatRunner) RunCaveatExpression(
 func (cr *CaveatRunner) PopulateCaveatDefinitionsForExpr(ctx context.Context, expr *core.CaveatExpression, reader CaveatDefinitionLookup) error {
 	ctx, span := tracer.Start(ctx, "PopulateCaveatDefinitions")
 	defer span.End()
+
+	// If the reader is backed by a unified stored schema, use the compiled-caveat cache
+	// tied to that schema version so deserialization (which rebuilds the CEL environment)
+	// is paid once per schema rather than once per check.
+	if cr.schemaCache == nil {
+		if provider, ok := reader.(cachedSchemaProvider); ok {
+			if stored := provider.StoredSchema(); stored != nil {
+				schemaCache, err := CompiledCaveatCacheFor(stored)
+				if err != nil {
+					return err
+				}
+				cr.schemaCache = schemaCache
+			}
+		}
+	}
 
 	// Collect all referenced caveat definitions in the expression.
 	caveatNames := mapz.NewSet[string]()
@@ -138,12 +166,23 @@ func (cr *CaveatRunner) get(caveatDefName string) (*core.CaveatDefinition, *cave
 		return caveat, deserialized, nil
 	}
 
-	parameterTypes, err := caveattypes.DecodeParameterTypes(cr.caveatTypeSet, caveat.ParameterTypes)
-	if err != nil {
-		return nil, nil, err
+	compile := func() (*caveats.CompiledCaveat, error) {
+		parameterTypes, err := caveattypes.DecodeParameterTypes(cr.caveatTypeSet, caveat.ParameterTypes)
+		if err != nil {
+			return nil, err
+		}
+		return caveats.DeserializeCaveatWithTypeSet(cr.caveatTypeSet, caveat.SerializedExpression, parameterTypes)
 	}
 
-	justDeserialized, err := caveats.DeserializeCaveatWithTypeSet(cr.caveatTypeSet, caveat.SerializedExpression, parameterTypes)
+	// Prefer the schema-tied cache (shared across checks) when available; fall back to
+	// per-runner compilation otherwise.
+	var justDeserialized *caveats.CompiledCaveat
+	var err error
+	if cr.schemaCache != nil {
+		justDeserialized, err = cr.schemaCache.GetOrCompile(caveatDefName, compile)
+	} else {
+		justDeserialized, err = compile()
+	}
 	if err != nil {
 		return caveat, nil, err
 	}
--- a/internal/caveats/run.go
+++ b/internal/caveats/run.go
@@ -64,11 +64,11 @@ type CaveatRunner struct {
 	caveatDefs          map[string]*core.CaveatDefinition
 	deserializedCaveats map[string]*caveats.CompiledCaveat
 
-	// schemaCache, when non-nil, is a compiled-caveat cache tied to the stored schema
+	// compiledCaveatCache, when non-nil, is a compiled-caveat cache tied to the stored schema
 	// (and thus shared across checks and invalidated on schema change). It is discovered
 	// from the reader on first use. When nil, deserializedCaveats provides per-runner
 	// caching only (the legacy behavior).
-	schemaCache *CompiledCaveatCache
+	compiledCaveatCache *CompiledCaveatCache
 }
 
 // NewCaveatRunner creates a new CaveatRunner.
@@ -107,14 +107,14 @@ func (cr *CaveatRunner) PopulateCaveatDefinitionsForExpr(ctx context.Context, ex
 	// If the reader is backed by a unified stored schema, use the compiled-caveat cache
 	// tied to that schema version so deserialization (which rebuilds the CEL environment)
 	// is paid once per schema rather than once per check.
-	if cr.schemaCache == nil {
+	if cr.compiledCaveatCache == nil {
 		if provider, ok := reader.(cachedSchemaProvider); ok {
 			if stored := provider.StoredSchema(); stored != nil {
-				schemaCache, err := CompiledCaveatCacheFor(stored)
+				compiledCaveatCache, err := CompiledCaveatCacheFor(stored)
 				if err != nil {
 					return err
 				}
-				cr.schemaCache = schemaCache
+				cr.compiledCaveatCache = compiledCaveatCache
 			}
 		}
 	}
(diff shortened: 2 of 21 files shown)


internal/caveats/schemacache.go
```diff
@@ -0,0 +1,49 @@
...
+
+// compiledCaveatCacheKey identifies the schema-derived cache of compiled (deserialized)
+// caveats. The cache is tied to a single stored-schema version (via datalayer.CachedSchema)
+// and is discarded when the schema changes.
+var compiledCaveatCacheKey = datalayer.NewDerivedCacheKey("caveats.compiled")
+
+func init() {
+	if err := datalayer.RegisterDerivedCache(compiledCaveatCacheKey, func() any { return &CompiledCaveatCache{} }); err != nil {
+		spiceerrors.MustPanicf("failed to register compiled caveat cache: %v", err)
+	}
```
tstirrat15: Are we expecting to have many of these? Like do we need a registry, or can this be singleton?

pkg/datalayer/derivedcache.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+	"sync"
+
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/spiceerrors"
+)
+
+// CachedSchema bundles a read-only stored schema with lazily-built, schema-derived caches
+// (for example compiled caveats, type systems, or reachability). Instances are produced and
```
tstirrat15: These being the other examples of things that we would register?

pkg/datalayer/derivedcache.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return &CachedSchema{schema: schema}
+}
+
+// Schema returns the underlying read-only stored schema.
+func (c *CachedSchema) Schema() *datastore.ReadOnlyStoredSchema { return c.schema }
+
+// DerivedCacheKey identifies a kind of schema-derived cache (e.g. compiled caveats). Create
+// one per kind, typically as a package-level var, via NewDerivedCacheKey.
+type DerivedCacheKey struct{ name string }
```
tstirrat15: For my own edification: why is this a struct with a single member, rather than a subtype of `string` directly?

pkg/datalayer/derivedcache.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	built := factory.(func() any)()
+	actual, _ := c.derived.LoadOrStore(key, built)
+	return actual
+}
+
+// GetDerivedCache returns the schema-derived cache of type T registered under key for the
+// given cached schema, building it lazily on first access.
+func GetDerivedCache[T any](c *CachedSchema, key DerivedCacheKey) T {
+	return c.derivedCache(key).(T)
```
tstirrat15: `MustBugF` here? Or are we not worried about a panic here?

pkg/embedded/permissions.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// SchemaMode controls how schema is read. The zero value reads legacy per-definition
+	// schema. Use datalayer.SchemaModeReadNewWriteNew (or *Both) for the unified schema.
+	SchemaMode datalayer.SchemaMode
+
+	// SchemaCacheMaxCostBytes, when > 0, enables an in-memory stored-schema cache of the
+	// given size in bytes. Caching the stored schema across checks is what allows
+	// schema-derived caches (e.g. compiled caveats) to persist; strongly recommended when
+	// SchemaMode reads from the unified schema.
+	SchemaCacheMaxCostBytes int64
```
tstirrat15, line 49: curious - why do we need a max cost here but not on the other schemas?

pkg/embedded/permissions.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// DispatchConcurrencyLimit, DispatchChunkSize, and MaxDepth use sane defaults if zero.
+	DispatchConcurrencyLimit uint16
+	DispatchChunkSize        uint16
+	MaxDepth                 uint32
+}
+
+// Permissions issues in-process permission checks against a datastore.
+type Permissions struct {
```
tstirrat15, line 58: `PermissionsChecker`? Or is the idea that this is `PermissionsService` minus the `Service`?

pkg/embedded/permissions.go
```diff
@@ -0,0 +1,185 @@
...
+		dispatcher: dispatcher,
+		cts:        cts,
+		chunkSize:  chunkSize,
+		maxDepth:   maxDepth,
+	}, nil
+}
+
+// CheckRequest is a single, fully-consistent permission check. The caveat context is passed
+// as native Go values (no structpb / base64 round-trip).
+type CheckRequest struct {
```
tstirrat15: Maybe mark that this should track the protobuf API shape? Is there a way to get a derived golang shape from the protobuf shape? Like I'm assuming this is because we don't want to pay the serde cost on `Struct`s?

pkg/embedded/permissions.go
```diff
@@ -0,0 +1,185 @@
...
+	case dispatchv1.ResourceCheckResult_MEMBER:
+		return CheckResult{HasPermission: true}, nil
+	case dispatchv1.ResourceCheckResult_CAVEATED_MEMBER:
+		return CheckResult{IsConditional: true, MissingContext: cr.MissingExprFields}, nil
+	default:
+		return CheckResult{}, nil
+	}
+}
+
+// Close releases resources held by the checker. It does not close the datastore.
```
tstirrat15, line 191: With the datastore, is that because we don't want to have to do dependency checking on the datastore?

tstirrat15 (commented):
See comments

miparnisari (changes requested):
I'm not a fan of the approach taken in this PR. You're constructing things that we already have (e.g you wrote a new `CheckRequest` object and `embedded.NewPermissions`) and we don't offer the entire API surface. We are also losing request validation I think?

Can we do something like this instead? Reuse our existing objects: 

```go

type server struct {
    dl  datalayer
    v1svc.NewPermissionsServer
    v1svc.NewSchemaServer
    v1svc.NewWatchServer
}

// server inherits the entire API surface, plus:

func  New(dl datalayer, options ...) *server {
    // configure everything
    // including eg the protovalidate object
}

func (s *server) Close() {
}
```
 
It's probably more work, but i think it's better for the long run since we don't duplicate code. 

If you look at our existing `devcontext` code, it looks very similar to what I suggested 😉 we could replace all of it with this `server` object.

https://github.com/authzed/spicedb/blob/95522074bcc63066c9fb618718cbe4598258ed66/pkg/development/devcontext.go#L169-L217

josephschorr:
> I'm not a fan of the approach taken in this PR. You're constructing things that we already have (e.g you wrote a new `CheckRequest` object and `embedded.NewPermissions`) and we don't offer the entire API surface. We are also losing request validation I think?
> 
> Can we do something like this instead? Reuse our existing objects:

Not without re-introducing a requirement to encode/decode obejcts as structpb's, which is a loss of performance we should not need to pay, especially for this simplified use case

miparnisari:
> which is a loss of performance we should not need to pay

The penalty comes only in caveated checks. IMO it doesn't justify the maintenance burden that comes with the duplication and the poor DX that comes from only offering one API (`CheckPermission`).

internal/caveats/schemacache.go
josephschorr: Registry, to avoid an import cycle: a typed field per cache kind would make `pkg/datastore` depend on the packages that own those caches (caveats now, type systems/reachability later) — which already depend on it. Registration inverts that. We only expect a handful.

pkg/datalayer/derivedcache.go
josephschorr: Yes — compiled caveats today; type systems and reachability later.

pkg/datalayer/derivedcache.go
josephschorr: So the key is opaque — it can only come from `NewDerivedCacheKey`, not a bare string literal at some call site. Can make it a `string` subtype if you'd rather.

pkg/datalayer/derivedcache.go
josephschorr: Yeah — it now returns an error instead of relying on the bare type assertion.

pkg/embedded/permissions.go
josephschorr: It's the memory bound for the one cache the embedded library stands up. It's a real byte budget now — each entry is costed by the schema size plus the derived caches hanging off it.

pkg/embedded/permissions.go
josephschorr: Holding the name until the API shape settles (your thread with miparnisari).

pkg/embedded/permissions.go
josephschorr: Exactly — passing native Go values avoids building a `structpb.Struct` just for the server to convert it straight back. Added a note that it mirrors `v1.CheckPermissionRequest` and should be kept in sync.

pkg/embedded/permissions.go
josephschorr: The caller owns the datastore — it may be shared and outlive the checker, so closing it here would be wrong. `Close` only releases what this object created.

josephschorr:
> > which is a loss of performance we should not need to pay
> 
> The penalty comes only in caveated checks. IMO it doesn't justify the maintenance burden that comes with the duplication and the poor DX that comes from only offering one API (`CheckPermission`).

It is a non-trivial amount of overhead and many "in memory" systems do use caveats for dynamic or real-time behavior, so I do think it is worth it

pkg/embedded/permissions.go
```diff
@@ -0,0 +1,187 @@
...
+	// Datastore is the datastore to check against. Required. The caller owns its lifecycle.
+	Datastore datastore.Datastore
+
+	// CaveatTypeSet is the caveat type set used to compile and evaluate caveats.
+	// Defaults to caveattypes.Default.
+	CaveatTypeSet *caveattypes.TypeSet
+
+	// SchemaMode controls how schema is read. The zero value reads legacy per-definition
+	// schema. Use datalayer.SchemaModeReadNewWriteNew (or *Both) for the unified schema.
+	SchemaMode datalayer.SchemaMode
```
miparnisari, line 43: why give this as an option? why not use the new mode always?

pkg/embedded/permissions.go
miparnisari: why give this as an option at all? just turn it on always, no?

pkg/embedded/permissions.go
```diff
@@ -0,0 +1,187 @@
...
+	case dispatchv1.ResourceCheckResult_CAVEATED_MEMBER:
+		return CheckResult{IsConditional: true, MissingContext: cr.MissingExprFields}, nil
+	default:
+		return CheckResult{}, nil
+	}
+}
+
+// Close releases resources held by the checker. It does not close the datastore.
+func (p *Permissions) Close() error {
+	return p.dispatcher.Close()
```
miparnisari: close the cache too

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Schema: reads inside write transactions now use a cheap hash-only lookup (`schema_revision`) to check the cache before loading the full schema blob, reducing DB round-trips on cache hits (https://github.com/authzed/spicedb/pull/3160)
 - Updated the Prometheus buckets for `grpc_server_handling_seconds` and `spicedb_datastore_query_latency` to be able to correlate them (https://github.com/authzed/spicedb/pull/3188)
+- Embedded: add `pkg/embedded`, an in-process library for running permission checks against a datastore via the dispatch engine, without standing up a gRPC server (https://github.com/authzed/spicedb/pull/3166)
+- Caveats: compiled caveats (and their CEL environments) are now cached per schema version rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
+- Datastore: schema-derived artifacts (e.g. compiled caveats) are now cached on the stored schema (`ReadOnlyStoredSchema`) and share its lifetime, so they are rebuilt only when the schema changes (https://github.com/authzed/spicedb/pull/3166)
```
miparnisari: these two entries seem to be saying the same thing: "the schema cache now includes the compiled caveats"

internal/caveats/run.go
```diff
@@ -51,11 +51,24 @@ func RunSingleCaveatExpression(
...
 type CaveatRunner struct {
 	caveatTypeSet       *caveattypes.TypeSet
 	caveatDefs          map[string]*core.CaveatDefinition
 	deserializedCaveats map[string]*caveats.CompiledCaveat
+
+	// schemaCache, when non-nil, is a compiled-caveat cache tied to the stored schema
+	// (and thus shared across checks and invalidated on schema change). It is discovered
+	// from the reader on first use. When nil, deserializedCaveats provides per-runner
+	// caching only (the legacy behavior).
+	schemaCache *CompiledCaveatCache
```
miparnisari:
```suggestion
	compiledCaveatCache *CompiledCaveatCache
```

internal/caveats/schemacache.go
```diff
@@ -0,0 +1,70 @@
...
+	actual, _ := c.m.LoadOrStore(name, compiled)
+	return actual.(*caveats.CompiledCaveat), nil
+}
+
+// CompiledCaveatCacheFor returns the compiled-caveat cache tied to the given stored schema,
+// building it lazily on first access.
+func CompiledCaveatCacheFor(s *datastore.ReadOnlyStoredSchema) (*CompiledCaveatCache, error) {
+	return datastore.GetDerivedCache[*CompiledCaveatCache](s, compiledCaveatCacheKey)
+}
```
miparnisari, line 44: What is a "derived" cache / cache key? Why are you using `init()`s? This code strikes me as way too complex for what you are trying to accomplish.  

Why not stick the compiled caveat cache straight into the `datastore.ReadOnlyStoredSchema` object?

pkg/embedded/permissions.go
miparnisari: I don't mind either way

pkg/embedded/permissions.go
josephschorr: We need to support both schema modes here — the new unified schema is still experimental, so we can't use it unconditionally.

pkg/embedded/permissions.go
josephschorr: The schema cache is only consulted in the unified (new) schema mode — in legacy mode the schema hash is a bypass sentinel, so the cache does nothing. Since new mode is still experimental and this is memory the caller owns, we make it an explicit opt-in for now. Once the new schema mode becomes the default, we'll flip both this and `SchemaMode` to sensible defaults — added a TODO to that effect.

pkg/embedded/permissions.go
josephschorr: Good catch; fixed

CHANGELOG.md
josephschorr: Merged into a single entry — the datastore line was really just the mechanism behind the caveat caching.

internal/caveats/run.go
josephschorr: Renamed the field and its references to `compiledCaveatCache` — agreed, it isn't caching schemas. Also added a test (`TestCaveatRunnerUsesSchemaTiedCompiledCaveatCache`) covering the schema-tied-cache path.

internal/caveats/schemacache.go
josephschorr: The direct approach is an import cycle: `ReadOnlyStoredSchema` is in `pkg/datastore`, and the compiled-caveat cache is in `internal/caveats`, which imports `pkg/datastore`. The registry lets caveats — and other schema-derived caches — hang off the stored schema without datastore depending on them. It's a bit heavy for a single kind today, but I'm planning to add more kinds in the near future, so this is building for that.

pkg/embedded/permissions.go
josephschorr: Keeping `Permissions` for now; can revisit the name if the API shape shifts.

internal/caveats/schemacache.go
miparnisari: Two questions:
- can we break the import cycle?
- how important are the other caches that you are planning to add? I am concerned that we are, maybe, needlessly future-proofing this code without a strong need. If the other caches are not as important, I would push towards simplifying this code. I have a local stash with the simplified version. Let me know and I can push it.

miparnisari (dismissed):
I approve this PR on the promise that it will improve performance, but i'm still concerned on the amount of tech debt that we've introduced :/ Specifically these two threads:

- https://github.com/authzed/spicedb/pull/3166#pullrequestreview-4508614082
- https://github.com/authzed/spicedb/pull/3166#discussion_r3555676029

internal/caveats/schemacache.go
josephschorr: Addressed in `6a4f57b`.

On the cycle — there isn't one for *this* cache. `CompiledCaveatCache` stores `*pkg/caveats.CompiledCaveat`, and `pkg/caveats` doesn't import `pkg/datastore`, so the value can hang off `ReadOnlyStoredSchema` directly. The registry/`init()` was pre-building for the *other* planned caches (type systems, reachability), whose types live in packages that do import `pkg/datastore` — that's the future-proofing you flagged, and it wasn't earning its keep for a single cache.

So I dropped the whole mechanism — no `RegisterDerivedCache`/`GetDerivedCache`, no factory/estimator callbacks, no `init()`. In its place is one generic on the stored schema, used identically for every cache kind:

```go
type DerivedCacheKey[T any] struct{ /* typed handle */ }
func NewDerivedCacheKey[T any]() DerivedCacheKey[T]
func LoadOrStoreDerived[T any](r *ReadOnlyStoredSchema, key DerivedCacheKey[T], build func() T) (T, error)
```

The caveat cache is now just:

```go
var compiledCaveatCacheKey = datastore.NewDerivedCacheKey[*CompiledCaveatCache]()

func CompiledCaveatCacheFor(s *datastore.ReadOnlyStoredSchema) (*CompiledCaveatCache, error) {
	return datastore.LoadOrStoreDerived(s, compiledCaveatCacheKey, func() *CompiledCaveatCache { return &CompiledCaveatCache{} })
}
```

Keying is by the typed handle — no string keys, no reflection, one `sync.Map` lookup — so `T` comes from the key and a wrong return type is a compile error rather than a runtime one. When the type-system/reachability caches land they attach the same way with their own `NewDerivedCacheKey[T]()`, and `pkg/datastore` never depends on the owning packages.

miparnisari (dismissed):

miparnisari (approved):

PR #3167, opened 2026-06-09
refactor: remove global `caches` object

miparnisari:
Solves https://github.com/authzed/spicedb/pull/3156#issuecomment-4661748374 by simplifying how our caching works.

Right now, each cache is registered in a global `caches sync.Map`, which means that if you try to spin up 2 spicedb servers within the same process (as our tests do) and both have caching turned on, a panic will occur. This PR fixes that.


pkg/cache/metrics.go
```diff
@@ -46,32 +42,30 @@ var (
 	)
 )
 
-var caches sync.Map
```
miparnisari, line 49: The goal of this PR is to get rid of this global var.

pkg/cache/cache_otter.go
```diff
@@ -62,6 +78,49 @@ type otterCache[K KeyString, V any] struct {
...
+	registerer prometheus.Registerer
+	collectors []prometheus.Collector
+}
+
+// registerMetrics registers this cache's metrics, labeled by its name, with the
+// given registerer. The metrics are read from the cache at scrape time. On any
+// registration failure, already-registered metrics are rolled back.
+func (wtc *otterCache[K, V]) registerMetrics(registerer prometheus.Registerer) error {
+	labels := prometheus.Labels{"cache": wtc.name}
+	collectors := []prometheus.Collector{
```
miparnisari, line 93: i moved `cache/metrics` to here because i disliked the indirection. This does mean, though, that if we ever decide to add a new cache impl we will have to bring this too. I don't think we'll add a cache impl any time soon so that's why i did this

pkg/cache/cache_otter.go
```diff
@@ -50,10 +66,10 @@ func NewOtterCache[K KeyString, V any](name string, config *Config) (Cache[K, V]
...
 	cache, err := otter.New(opts)
 	return &otterCache[K, V]{
-		name,
-		cache,
-		otterMetrics{atomic.Uint64{}, counter},
-		config.DefaultTTL,
+		name:    name,
+		cache:   cache,
+		metrics: otterMetrics{atomic.Uint64{}, counter},
+		ttl:     config.DefaultTTL,
```
tstirrat15, line 72: 🙏

tstirrat15 (approved):
LGTM

PR #3173, opened 2026-06-11
perf(mysql): force per-query-shape indexes for relationship queries

josephschorr:
## Summary

MySQL's query optimizer would sometimes choose a suboptimal index for relationship reads, and the prior index-to-query-shape mapping was inaccurate — several shapes pointed at indexes that don't serve them well (with `TODO`s in the code noting exactly that). This PR forces the index that actually matches each query shape and uses the sort order that index provides, bringing MySQL in line with the explicit index-forcing already done for CRDB, Postgres, and Spanner.

`relation_tuple` is clustered on the synthetic `id` primary key, so the resource-keyed `UNIQUE` index (`uq_relation_tuple_living`) plays the role that the primary key plays in CRDB/Postgres for forward query shapes.

## Mapping

| Index | Sort order | Query shapes |
|---|---|---|
| `uq_relation_tuple_living` | `ByResource` | `CheckPermissionSelectDirectSubjects`, `CheckPermissionSelectIndirectSubjects`, `AllSubjectsForResources`, `FindResourceOfType` |
| `ix_relation_tuple_by_subject` | `BySubject` | `MatchingResourcesForSubject` |
| `ix_relation_tuple_by_subject_relation` | `BySubject` | `FindResourceAndSubjectWithRelations`, `FindSubjectOfTypeAndRelation`, `FindResourceRelationForSubjectRelation` |

`IndexingHintForQueryShape` returns a `forcedIndex` (emitting MySQL `FORCE INDEX (...)`) for these shapes and no hint otherwise. `QueryRelationships` and `ReverseQueryRelationships` apply the hint based on the request's query shape. `uq_relation_tuple_namespace` is now documented as a uniqueness-only constraint — it is not used to serve any read shape.

## Testing

- New unit test `TestIndexingHintForQueryShape` covering the shape → index/sort mapping.
- New `BenchmarkMySQLRelQueries` benchmark exercising the affected forward and reverse query shapes (requires Docker).
- `mage testds:mysql` passes locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Cache: switch to [otter](https://maypok86.github.io/otter/) as the primary cache implementation (https://github.com/authzed/spicedb/pull/3112)
 - Server handles: `GRPCDialContext` as a handle on the server used deprecated gRPC methods. We modernized it and renamed it to `NewClient` (https://github.com/authzed/spicedb/pull/3147)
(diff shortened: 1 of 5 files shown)


tstirrat15 (approved):
LGTM

PR #3174, opened 2026-06-11
fix: guard nil debug request in bulk check

### LINKED ISSUE #3159: CheckBulkPermissions panics with nil pointer dereference in bulkcheck.go:189 under concurrent load

linux

amd64

v1.53.0

--datastore-engine=postgres
--datastore-conn-pool-read-max-open=20
--datastore-conn-pool-read-min-open=5
--datastore-conn-pool-write-max-open=6
--datastore-conn-pool-write-min-open=2
--enable-experimental-watchable-schema-cache=true
--http-enabled=true

### Steps to Reproduce

1. Run SpiceDB v1.53.0 with a PostgreSQL datastore containing a large dataset.
2. Trigger concurrent load (in our case, a QA load test reproduced this reliably):
  -  Send multiple concurrent CheckBulkPermissions requests simultaneously
  - Each request contains a small number of items (under 100) with a caveat context on every item

Additional datastore context at time of incident; Relationship tuple counts by definition type:
- primary resource type:   ~1,000,000 tuples
- role type:               ~300,000 tuples
- tenant type:          ~13,000 tuples

Per-namespace resource count at query time was under 100 items,
Confirming this is not a large payload issue, but a concurrency issue.

### Expected Result

CheckBulkPermissions handles concurrent requests without crashing.
All requests return results or errors gracefully.

### Actual Result

SpiceDB panics with a nil pointer dereference inside the PreloadedTaskRunner goroutine, 
crashing the process. All replicas eventually crash.

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1bd6475]

goroutine 61457 [running]:
github.com/authzed/spicedb/internal/services/v1.(*bulkChecker).checkBulkPermissions.func4(...)
	/home/runner/work/spicedb/spicedb/internal/services/v1/bulkcheck.go:189 +0x5f5
github.com/authzed/spicedb/internal/services/v1.(*bulkChecker).checkBulkPermissions.func5.1(...)
	/home/runner/work/spicedb/spicedb/internal/services/v1/bulkcheck.go:294 +0x796
github.com/authzed/spicedb/internal/taskrunner.(*PreloadedTaskRunner).runner(...)
	/home/runner/work/spicedb/spicedb/internal/taskrunner/preloadedtaskrunner.go:109 +0x76
created by github.com/authzed/spicedb/internal/taskrunner.(*PreloadedTaskRunner).spawnIfAvailable
	/home/runner/work/spicedb/spicedb/internal/taskrunner/preloadedtaskrunner.go:81 +0xb8

Line 189 in bulkcheck.go is:
    dl := datalayer.MustFromContext(ctx).SnapshotReader(atRevision, schemaHash)

Environment:
- 3 replicas on Kubernetes
- PostgreSQL datastore
- amd64
- Panic occurs only under concurrent load, not during normal traffic
- Per-namespace resource count is small (under 100 items per bulk check call)
- All replicas crashed with the same stack trace

ivanauth:
CheckBulkPermissions panics when a DebugInformation entry with a nil
Check.Request reaches the resource-ID matching loop. This happens when a
tracing-enabled request shares a singleflight dispatch with a non-tracing
bulk check call — the combined metadata node produced by
combineResponseMetadata never sets Check.Request.

Add a nil guard to skip such entries.

Closes #3159


josephschorr (dismissed):
LGTM

miparnisari (changes requested):
Can you explain this solution? The bug report said

> Our CheckPermission calls have with_tracing: true but CheckBulkPermissions has with_tracing: false

how does this translate to `if di.Check == nil || di.Check.Request == nil`? if i understood correctly, in this particular issue, they would NOT be nil?

You also said 

>  the combined metadata node produced by combineResponseMetadata never sets Check.Request.

is this not the root cause then? Why are we not fixing that?

ivanauth:
Thank you for the feedback!

`di.Check.Request` actually is nil in this case. The node reaching that loop isn't the original per-resource trace node (those do have `Request` set). It's the wrapper node created by `combineResponseMetadata`, which intentionally sets fields like `TraceId`, `SourceId`, and `SubProblems` but leaves `Request` nil. The real trace entries are nested under `SubProblems`. The panic is consistent with that: `slices.Contains(di.Check.Request.ResourceIds, ...)` can only nil-deref if `di.Check.Request` is nil.

I don't think `combineResponseMetadata` is the bug. A combined node represents multiple sub-checks, so there isn't a single request to attach to it. The code also relies on `Request == nil` when deciding whether to flatten subproblems, so populating it would change that behavior.

The actual issue is one layer up: the singleflight key doesn't include the debug flag. That allows a `with_tracing=true` `CheckPermission` request and a `with_tracing=false` `CheckBulkPermissions` request to share the same flight, which can leak debug metadata into the bulk response.

I've reworked the fix so debug-enabled checks no longer share a singleflight flight and instead dispatch directly. That prevents debug metadata from leaking into non-debug requests and fixes the panic at the source. It also fixes the reverse case, where a tracing request could lose trace data because it shared a flight with a non-tracing request.

I kept the nil check as a defensive guard since `Request == nil` is still a valid shape for combined metadata nodes.

 Happy to adjust the approach if you'd prefer it handled differently or close this PR completely since it's getting too complex.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Server handles: `GRPCDialContext` as a handle on the server used deprecated gRPC methods. We modernized it and renamed it to `NewClient` (https://github.com/authzed/spicedb/pull/3147)
 
 ### Fixed
+- Fixed a nil pointer dereference panic in `CheckBulkPermissions` that could occur under concurrent load when a tracing-enabled check shared a singleflight dispatch with a non-tracing bulk check. Debug-enabled checks are no longer singleflighted together with non-debug checks. (https://github.com/authzed/spicedb/pull/3174)
```
miparnisari: Move to unreleased section

internal/dispatch/singleflight/singleflight.go
```diff
@@ -80,6 +80,18 @@ func (d *Dispatcher) DispatchCheck(ctx context.Context, req *v1.DispatchCheckReq
...
 
+	// Debug-enabled checks must not share a flight with non-debug checks: the
+	// dispatch key does not depend on req.Debug, so an identical concurrent
+	// request with a different debug level would otherwise collapse into the
+	// same flight. A non-debug leader would deny a tracing follower its trace,
+	// and a debug leader would leak its trace to a non-debug follower (the
+	// latter has panicked CheckBulkPermissions, see #3159). Dispatch directly
+	// so the response carries exactly the debug info this caller asked for.
+	if req.Debug != v1.DispatchCheckRequest_NO_DEBUG {
+		singleFlightCount.WithLabelValues("DispatchCheck", "debug").Inc()
```
miparnisari, line 91: Why are we incrementing the sf counter if we didn't use sf?

internal/services/v1/bulkcheck_test.go
```diff
@@ -0,0 +1,102 @@
...
+				Check: &dispatchv1.CheckDebugTrace{
+					SubProblems: []*dispatchv1.CheckDebugTrace{{IsCachedResult: true}},
+				},
+			},
+		},
+	}, nil
+}
+
+// TestCheckBulkPermissionsWithNilDebugRequest ensures that a debug information entry
+// with a nil Check.Request does not panic CheckBulkPermissions when tracing was not
```
miparnisari: Is this supposed to test the "continue" line? If so, it's not getting hit according to codecov

miparnisari (changes requested):
Lgtm, just a couple nits

internal/dispatch/singleflight/singleflight.go
ivanauth: I was following the existing pattern in this same function, where we already bump this counter for other cases that skip singleflight, like missing (empty bloom filter) and loop (possible recursion), plus passthrough in the query plan path. So the shared label is really tracking the dispatch reason rather than just whether it was shared, and debug slots in as another intentional bypass we can see in the metric.

Happy to drop it though if you'd rather this only count requests that actually go through singleflight. Let me know what you prefer.

internal/services/v1/bulkcheck.go
```diff
@@ -186,6 +186,9 @@ func (bc *bulkChecker) checkBulkPermissions(ctx context.Context, req *v1.CheckBu
 				// Find the debug info that matches the resource ID.
 				var debugInfo *dispatchv1.DebugInformation
 				for _, di := range debugInfos {
+					if di.Check == nil || di.Check.Request == nil {
+						continue
+					}
 					if slices.Contains(di.Check.Request.ResourceIds, resourceID) {
```
miparnisari, line 192: Nit: Instead of having the `continue` above, we could also do `if slices.Contains(di.GetCheck().GetRequest().GetResourceIds(), resourceID) {`  which is enforced by the `protogetter` linter https://github.com/authzed/spicedb/pull/2638

miparnisari (approved):

internal/services/v1/bulkcheck_test.go
miparnisari: nvm, it is getting hit

PR #3175, opened 2026-06-11
test: add more goleak checks

miparnisari:
Followup to https://github.com/authzed/spicedb/pull/3171 since there are more goroutine leaks.

I am pretty sure that the leak is that we made telemetry ON by default 

https://github.com/authzed/spicedb/blame/87482ed4fbeabd57ae4e896e1b4674958e6b16ba/pkg/cmd/server/server.go#L162

and now we are leaking its connection. The fix is in `internal/telemetry/reporter.go`.


josephschorr (approved):
LGTM

PR #3176, opened 2026-06-12
feat(crdb): experimental crdb query cancellation

ecordell:
When a request context is canceled mid-query on the CRDB datastore, pgx's default behavior is to destroy the connection. Under load this drains the write pool and triggers the metastable death spiral described in #2576. A previous attempt to fix this using pgx's built-in `CancelRequestContextWatcherHandler` (pgwire cancel protocol) was reverted in #2434 because CRDB applies pgwire cancels asynchronously; a late-arriving cancel could kill the *next* query on the connection.

This PR implements cancellation using CockroachDB's `CANCEL QUERIES` statement instead. A new `pool.Canceler` owns a small dedicated connection pool and a registry mapping each pooled write connection to its CRDB `session_id` (captured via `SHOW session_id` at connect). A custom `ctxwatch.Handler` fires on context cancellation and issues `CANCEL QUERIES IF EXISTS (SELECT query_id FROM [SHOW CLUSTER STATEMENTS] WHERE session_id = '...')` on a sibling connection, then **blocks until that statement completes** before releasing the original connection back to the pool. This sequencing eliminates the wrong-query race by construction: no cancellation can be in flight when the next query starts on the same session.

The feature is gated behind `--datastore-experimental-crdb-query-cancellation` (default off). When enabled, set `--write-conn-acquisition-timeout=0`; with connections no longer being destroyed on cancel, pool exhaustion under load is far less likely and the acquisition-timeout backpressure is no longer needed.


internal/datastore/crdb/pool/cancel_handler.go
```diff
@@ -0,0 +1,101 @@
...
+		pgConn:        pgConn,
+		canceler:      canceler,
+		deadlineDelay: deadlineDelay,
+	}
+}
+
+// HandleCancel implements ctxwatch.Handler.
+func (h *CancelQueryContextWatcherHandler) HandleCancel(_ context.Context) {
+	h.cancelFinished = make(chan struct{})
+	h.cancelSucceeded.Store(false)
```
tstirrat15, line 69: Is the idea here that there are potentially multiple requests using this same handler? I'm wondering why this isn't in the constructor

internal/datastore/crdb/pool/cancel_handler.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Fallback: if in-band cancellation fails to resolve the in-flight
+	// operation, this deadline unblocks the pending read so the connection
+	// errors out and is destroyed.
+	_ = h.conn.SetDeadline(time.Now().Add(h.deadlineDelay))
+
+	go func() {
+		defer close(h.cancelFinished)
+		if err := h.canceler.CancelSessionQueries(h.pgConn); err != nil {
+			log.Warn().Err(err).Msg("in-band query cancellation failed; connection will be discarded")
```
tstirrat15, line 79: At the risk of Yet Another Metric, successful and failed cancellations seem like they'd be good things to count

tstirrat15: Nvm I see you beat me to it

tstirrat15:
> When enabled, set --write-conn-acquisition-timeout=0; with connections no longer being destroyed on cancel, pool exhaustion under load is far less likely and the acquisition-timeout backpressure is no longer needed.

If the fallback behavior when the cancelpool isn't available is the old connection cancellation behavior, couldn't you still hit the degenerate condition?

internal/datastore/crdb/pool/canceler.go
```diff
@@ -0,0 +1,186 @@
...
+func (c *Canceler) InstallOn(config *pgxpool.Config) {
+	afterConnect := config.AfterConnect
+	config.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
+		if afterConnect != nil {
+			if err := afterConnect(ctx, conn); err != nil {
+				return err
+			}
+		}
+		return c.RegisterSession(ctx, conn)
+	}
```
tstirrat15, line 168: It feels like there ought to be a nicer way to do this but I know that's not on this code.

internal/datastore/crdb/pool/canceler.go
```diff
@@ -0,0 +1,186 @@
...
+	config.BeforeClose = func(conn *pgx.Conn) {
+		if beforeClose != nil {
+			beforeClose(conn)
+		}
+		c.UnregisterSession(conn.PgConn())
+	}
+
+	config.ConnConfig.BuildContextWatcherHandler = func(pgConn *pgconn.PgConn) ctxwatch.Handler {
+		return NewCancelQueryContextWatcherHandler(pgConn, c, defaultCancelDeadlineDelay)
+	}
```
tstirrat15, line 180: Do we need to do the same wrapping behavior here?

internal/datastore/crdb/pool/pgx.go
```diff
@@ -33,19 +33,22 @@ import (
...
-// beginFuncExec is copied directly from pgx
+// beginFuncExec is adapted from pgx
 // src: https://github.com/jackc/pgx/blob/f59e8bf5551f403e6b7ec0912097bce85ea21351/tx.go#L410-L425
+// It diverges from upstream by severing the context for rollbacks: the
+// surrounding context is often already canceled when rollback runs, and the
+// rollback must still reach the server to keep the connection reusable.
 func beginFuncExec(ctx context.Context, tx pgx.Tx, fn func(pgx.Tx) error) (err error) {
 	defer func() {
-		rollbackErr := tx.Rollback(ctx)
+		rollbackErr := tx.Rollback(context.WithoutCancel(ctx))
```
tstirrat15, line 43: Was this a latent bug or is this necessitated by the new approach?

tstirrat15 (commented):
See comments, otherwise LGTM

internal/datastore/crdb/pool/pgx.go
ecordell: latent bug! in cases where we would need to rollback due to a context cancel the rollback would always fail too.

PR #3184, opened 2026-06-17
fix: match IPv4-mapped IPv6 addresses in the in_cidr caveat

kanywst:
## Description

`in_cidr` does not match an IPv4-mapped IPv6 address (`::ffff:10.1.2.3`) against an IPv4 CIDR, while the dotted form (`10.1.2.3`) matches. They are the same host (RFC 4291 section 2.5.5.2). The parsed address is not unmapped, so `netip.Prefix.Contains` compares across families.

Unmap at parse time so both forms compare equal.

## Testing

`TestIPAddressIPv4Mapped` fails before the change and passes after; the rest of `pkg/caveats` stays green.


tstirrat15 (dismissed):
This looks good to me. Thank you for putting this up!

tstirrat15:
Would you be willing to sign the CLA?

tstirrat15:
Can you also add an entry to CHANGELOG.md? I'll help with getting it through CI - we've had some trouble with flaky jobs recently.

kanywst:

miparnisari (approved):

PR #3187, opened 2026-06-18
fix(mysql): exponential backoff on write retries (#3172)

### LINKED ISSUE #3172: MySQL deadlocks on WriteRelationships

linux

amd64

v1.52.0

_No response_

### Steps to Reproduce

**Steps to reproduce**

1. Deploy SpiceDB with the MySQL datastore (`datastoreEngine: mysql`)
2. Configure with `SPICEDB_DATASTORE_MAX_TX_RETRIES=25` and `SPICEDB_DATASTORE_LOCK_WAIT_TIMEOUT=1`
3. Send 20+ concurrent `WriteRelationships` requests where multiple requests write TOUCH updates targeting the same resource object (e.g., `venue:X#member@user:Y1`, `venue:X#member@user:Y2`, ...)
4. Observe `max retries exceeded: unable to write relationships: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction`

### Expected Result

All 20+ concurrent `WriteRelationships` calls eventually succeed.

### Actual Result

Observe `max retries exceeded: unable to write relationships: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction`

ivanauth:
Concurrent WriteRelationships on the same resource could deadlock (MySQL 1213) and then fail with "max retries exceeded" - the retry loop retried with no delay, re-colliding on the same contention. Adds jittered exponential backoff between retries (matching Postgres), skipping the sleep after the final attempt.

Fixes #3172.


internal/datastore/mysql/datastore.go
```diff
@@ -338,6 +339,18 @@ func (mds *mysqlDatastore) SnapshotReader(rev datastore.Revision) datastore.Read
 
 func noCleanup() error { return nil }
 
+// sleepOnErr sleeps for an exponentially increasing duration with jitter after a
+// retryable error, matching the backoff used by the Postgres datastore.
+func sleepOnErr(ctx context.Context, err error, retries uint8) {
```
miparnisari: instead of copying this function can we move the existing one to a common file that all datastores can refer?

miparnisari (changes requested):
> Concurrent WriteRelationships on the same resource

could we add a datastore test for this scenario? and please due the due dilligence of checking that it fails on `main` for mysql and then bring your fix in :)

I see that there are some tests written already that could be checking this?

- `ConcurrentWriteSerializationTest`.
- `RetryTest`

Why aren't they failing 🤔

ivanauth:
Both addressed:

1. Moved `SleepOnErr` into `internal/datastore/common` and updated the Postgres, CRDB, and MySQL implementations to use it, removing the duplicate implementation.

2. Added `TestMySQLConcurrentWriteDeadlock`, which runs 50 concurrent writes against the same resource with a low `MaxRetries` setting. The test reproduces the exact failure from the issue (`Error 1213 ... max retries exceeded`) on `main` and passes with this fix.

The existing tests don't catch this scenario:

* `ConcurrentWriteSerializationTest` writes to different resources, so it doesn't create lock contention.
* `RetryTest` uses an injected mock error rather than triggering a real database deadlock.

miparnisari (approved):

PR #3189, opened 2026-06-19
fix(postgres): stop logging a harmless(*this is not correct*) GC lock release as an error

### LINKED ISSUE #2847: Failure to release lock reported in GC process on Postgres
```
2026-01-23 14:54:31.844	
{
  "level": "error",
  "error": "failed to release lock 1",
  "time": "2026-01-23T19:54:31Z",
  "message": "error unlocking after gc run"
}
```

This isn't a huge problem but it shouldn't be happening

ivanauth:
We log an ERROR ("failed to release lock 1 / error unlocking after gc run") whenever pg_advisory_unlock says we don't hold the lock. That happens when the GC connection gets reset mid-run — the new session never took the lock, so there's nothing to release. It's harmless (the lock just avoids redundant GC runs, and GC deletes are idempotent), but it shows up as a scary error.

Now we recognize that specific case and skip the error. Real release failures still get logged.

Fixes #2847.

### DIFF
--- a/internal/datastore/postgres/gc.go
+++ b/internal/datastore/postgres/gc.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


ivanauth:
Closing this - #3059 handles it properly.

PR #3196, opened 2026-06-26
refactor: new proxy for singleflighting ds.OptimizedRevision

miparnisari:
## Summary

Moves optimized-revision caching out of the individual datastores and into a single `proxy.NewOptimizedRevisionProxy`.

## Before

Every SQL datastore embedded a `revisions.CachedOptimizedRevisions` struct (CRDB and Spanner via a further `RemoteClockRevisions` wrapper). Each datastore constructor computed `maxRevisionStaleness` itself, built the cache, then had to remember to call `SetOptimizedRevisionFunc` / `SetNowFunc` / `SetNowOnlyFunc` after construction to plug in its own query. `Datastore.OptimizedRevision` returned an already-cached `RevisionWithSchemaHash`, so caching, `singleflight`, jitter and the 2s/10s timeout logic all lived inside each datastore implementation, and the proxy chain's `singleflightProxy` explicitly skipped optimized revisions because "the datastore already does it".

## After

- The interface changes to `OptimizedRevision(ctx context.Context) (RevisionWithSchemaHashAndValidity, error)` and every datastore (CRDB, Spanner, Postgres, MySQL, memdb) returns the uncached answer: one round-trip, plus how long it stays valid.
- Caching, deduplication, jitter, and the bounded shared/fallback timeouts move into a single new `proxy.NewOptimizedRevisionProxy`, wired once in `pkg/cmd/datastore/datastore.go` alongside the other proxies.
- The old `CachedOptimizedRevisions` and `RemoteClockRevisions` are deleted; what remains in `internal/datastore/revisions` is two pure helpers: `Quantize` (used by CRDB, Spanner and memdb) and `CheckGCWindow` (used by CRDB and Spanner). `CheckRevision` is implemented directly on those two datastores.
- The per-engine `MaxRevisionStalenessPercent` option, field and default are removed from CRDB, Postgres, MySQL and Spanner. Nothing read them once the cache moved out. The `--datastore-revision-quantization-max-staleness-percent` flag and `dsconfig.Config.MaxRevisionStalenessPercent` are unchanged and now feed the proxy directly.

## Benefits

- Less lines of code overall and IDE navigation is easier
- Consistent with the existing architecture. SpiceDB already layers observability, singleflight, schema caching, etc. as **proxies**; the optimized-revision cache was the odd one baked into the storage layer.
- Fixes `maxRevisionStaleness` inconsistency. Each datastore computed it separately; now it is computed once, including the GC-window cap that previously lived only in `RemoteClockRevisions`.


internal/datastore/revisions/hlc.go
```diff
@@ -0,0 +1,48 @@
...
+
+// QuantizeHLC rounds an HLC "now" revision down to a quantization boundary,
+// after subtracting the follower-read delay, and reports how long the resulting
+// quantized revision remains valid (i.e. until the next quantization boundary).
+//
+// This is the in-Go quantization used by datastores whose only revision
+// primitive is a raw HLC clock (CockroachDB, Spanner). Datastores that quantize
+// in SQL (Postgres, MySQL) compute validFor directly in their query and do not
+// use this helper.
+func QuantizeHLC(now WithTimestampRevision, followerReadDelay, quantization time.Duration) (datastore.Revision, time.Duration) {
```
josephschorr, line 17: why not put this on the revision itself?

pkg/datastore/datastore.go
```diff
@@ -754,7 +754,13 @@ type ReadOnlyDatastore interface {
...
 	// OptimizedRevision gets a revision that will likely already be replicated
 	// and will likely be shared amongst many queries.
-	OptimizedRevision(ctx context.Context) (RevisionWithSchemaHash, error)
+	//
+	// Implementations return the *uncached* answer: one datastore round-trip per
+	// call, yielding the revision, how long it remains valid, and the schema hash
+	// visible at that revision (or "" if not provided on this path). Caching,
+	// deduplication, and jitter are layered on by proxy.NewOptimizedRevisionProxy.
+	// Callers that do not need validFor may discard it with _.
+	OptimizedRevision(ctx context.Context) (rev Revision, validFor time.Duration, schemaHash string, err error)
```
josephschorr, line 763: This should probably return a struct + error, since it has 4 returns now

miparnisari: done

internal/datastore/revisions/hlc.go
miparnisari: i'm not sure what you mean - make this a method of `WithTimestampRevision`? but it's an interface, not a struct

josephschorr: Ah - I forgot it was on the interface. Ok

internal/datastore/proxy/optimized_revision.go
```diff
@@ -0,0 +1,216 @@
...
+	rvt := localNow.Add(fresh.ValidFor)
+
+	// Prune the candidates that have definitely expired.
+	p.mu.Lock()
+	var numToDrop uint
+	for _, candidate := range p.candidates {
+		if candidate.validThrough.Add(p.maxStaleness).Before(localNow) {
+			numToDrop++
+		} else {
+			break
```
josephschorr, line 163: remove the else branch and put a continue above

josephschorr (approved):
LGTM

Issue #3198, opened 2026-06-30
Proposal: per-operation timing + datastore-read attribution in the Check debug trace


poucet:
## Problem

When debugging the latency of a single `CheckPermission` / `CheckBulkPermissions`
call via the debug trace (`--with-tracing`), it is currently very hard to tell
**where the time actually goes**. Two questions in particular cannot be answered
from a single response today:

1. **Are sub-problems traversed in parallel or serially?** `ConcurrentChecker`
   resolves `sub_problems` concurrently (`union`/`all`/`difference` →
   `dispatchAllAsync` → the semaphore-bounded taskrunner). A parent's `duration`
   is neither the sum nor the max of its children's durations, so from the trace
   you cannot tell whether siblings overlapped, serialized behind the
   concurrency limit, or where the idle gaps are — i.e. you cannot reconstruct
   the critical path.

2. **Where are we blocked on datastore reads?** The only DB signal derivable
   from the trace today is per-node *self-time* (node span minus children
   spans), which lumps datastore blocking and CPU together. For a bulk check
   suspected of being read-bound, that blob is exactly what you can't decompose.

The goal is to answer both from the response itself, on demand, **without**
standing up a sampled OpenTelemetry → Tempo pipeline.

## Key observation

Both pieces of timing already exist on the server and are thrown away:

- `internal/graph/check.go` captures an absolute `startTime` when
  `Debug != NO_DEBUG`, uses it only to compute `duration`, then drops it.
- `internal/datastore/proxy/observable.go` (`QueryRelationships` /
  `ReverseQueryRelationships`) already times every relationship query from issue
  until the result iterator is fully drained — i.e. real blocking/streaming time
  — but it lands only in a Prometheus histogram (aggregate, by `query_shape`)
  and an OTel span, never in the per-check debug trace.

So this is plumbing of already-computed values, not new measurement.

## Proposal

Add to `CheckDebugTrace` (public `authzed.api.v1` and internal
`dispatch.v1`):

- `start_time` (`google.protobuf.Timestamp`) per node — combined with the
  existing authoritative `duration`, places every dispatch node on one absolute
  time axis. (`end_time = start_time + duration` is derived at the public
  boundary; included for convenience, droppable if a minimal surface is
  preferred.)
- `repeated DatastoreQuery datastore_queries` per node — each carrying
  `{query_shape, start_time, duration, relationship_count}` for the relationship
  reads that node issued itself. Per-query events (not one summed number)
[...]

PR #3199, opened 2026-06-30
feat(graph): per-node timing + datastore-read attribution in the check debug trace (internal proto)

poucet:
## What

Adds two pieces of timing to the **internal** dispatch `CheckDebugTrace` so a
single `Check` / `CheckBulk` debug trace (`--with-tracing`) can answer two
questions that `duration` alone cannot:

1. **Were sub-problems traversed concurrently?** Each node now carries a
   wall-clock `start_time`; combined with the existing authoritative `duration`,
   every dispatch node lands on one absolute time axis, so sibling overlap (or
   serialization behind the concurrency limit) is reconstructable.
2. **Where was time spent blocked on datastore reads?** Each node now carries
   `repeated DatastoreQuery datastore_queries` — `{query_shape, start_time,
   duration, relationship_count}` for the relationship reads that node issued
   itself. Per-query events (not one summed number) because a node can issue
   concurrent reads, so the events show overlap instead of hiding it.

Implements the internal-proto slice of authzed/spicedb#3198.

## How

- `proto/internal/dispatch/v1/dispatch.proto`: `CheckDebugTrace.start_time` (9),
  `CheckDebugTrace.datastore_queries` (10), and a new `DatastoreQuery` message.
- `internal/dstrace`: a small leaf package with a context-scoped,
  mutex-guarded `Collector` (a node may issue concurrent reads).
- `internal/graph/check.go`: when `Debug != NO_DEBUG`, install a fresh collector
  per node (dispatched children re-enter `Check` and install their own, so reads
  attribute to the node that issued them), and stamp `start_time` +
  `datastore_queries` next to the existing `duration`.
- `internal/datastore/proxy/observable.go`: the proxy already times each query
  from issue until the iterator is fully drained; it now also records that
  observation into the collector when one is present. This is a single
  conditional branch and a no-op when debug tracing is off.

## Scope / follow-up

Internal-proto only. The public `authzed.api.v1.CheckDebugTrace` fields
(`start_time`/`end_time`/`datastore_queries`) live in `authzed/api` and must
land + be published there before the buf dep can be bumped and
`internal/services/v1/debug.go` `convertCheckTrace` can surface them. The
internal trace already round-trips through `DebugInfo`, so the approach is fully
exercised end-to-end here.

## Tests

- `internal/dstrace`: collector unit tests (recording, concurrency safety,
  nil-safety, nested independence).
- `internal/datastore/proxy/observable_test.go`: proxy records query events
  (forward + reverse) into a context collector; no-op without one.
- `internal/graph/computed`: an integration test wrapping memdb in the
  observable proxy and running a multi-hop check with trace debugging, then
  walking the dispatch trace to assert every executed node has a `start_time`
  and that datastore reads are attributed as `DatastoreQuery` events.

## Compatibility

New fields only (9/10 internal); `buf breaking ... use: WIRE` passes. `duration`
is unchanged and remains authoritative. The collector and the proxy branch only
activate under debug tracing, so the non-debug hot path is untouched.


poucet:
- - -
   - - -

Issue #3200, opened 2026-06-30
1.54.0 excessive memory usage causing production outage


mcgrawia:
### What platforms are affected?

linux

### What architectures are affected?

amd64

### What SpiceDB version are you using?

v1.54.0

### What SpiceDB flags do you use?

ENV SPICEDB_DATASTORE_ENGINE=postgres

### Steps to Reproduce

1. use spicedb version 1.53.0
2. run migration and upgrade to v1.54.0

### Expected Result

Memory utilization should remain the same.

### Actual Result

**IMMEDIATE ASK: What steps are required to downgrade to v1.53.0? This is causing an immediate production degradation for our platform.**

Issue overview: 

Memory utilization climbs until 90%, at which point spiceDB starts rejecting permission check calls, causing upstream service outages.

```
{"level":"warn","protocol":"grpc","grpc.component":"server","grpc.service":"authzed.api.v1.PermissionsService","grpc.method":"CheckPermission","grpc.method_type":"unary","requestID":"d91tlc16e243iepo54j0","peer.address":"10.8.23.44:16036","grpc.start_time":"2026-06-30T15:03:12.634861755Z","grpc.code":"ResourceExhausted","grpc.error":"rpc error: code = ResourceExhausted desc = server rejected the request because memory usage is too high","grpc.time_ms":0,"time":"2026-06-30T15:03:12.6349217Z","message":"finished call"}
```

Here's a 2-week snapshot of our metrics, you can see the restart event mid-way through when the v1.54.0 deployed and the memory utilization has grown since:

<img width="1384" height="345" alt="Image" src="https://github.com/user-attachments/assets/b1c2eb1b-9028-41f3-94b4-c4e75fddc31d" />

tstirrat15:
SpiceDB exposes a `pprof` profiling endpoint. Are y'all able to attach to it and look at a memory profile? We'll be doing our own debugging on our side but it could speed things up.

tstirrat15:
And are there other flags that you're using? Any experimental flags?

mcgrawia:
Hi @tstirrat15 thanks for the help. I'm checking to see if we can connect to the `pprof` port. We're running in ECS so I need to speak to our infra folks to see what our options are.

I don't believe we're setting any other flags. We have the following configs for our ECS task:

ENV:
SPICEDB_DATASTORE_ENGINE=postgres
[...]

PR #3201, opened 2026-06-30
chore: change how memory limit is set

tstirrat15:
## Description
See #3200. When SpiceDB couldn't determine the memory available to it, such as can happen in AWS ECS, it was assuming that it had much more memory than it did. This is a set of improvements to that behavior.

## Changes
* Choose a conservative limit when memory can't be determined and emit a warning message
* Log the memory limit on startup together with how the limit was determined for debugging purposes
* Improve cost estimations for cache entries for the general cache
* Improve cost estimations for the schema hash cache using the size of the serialized schema
* Add tests for the new behaviors
## Testing
Review. See that tests pass.


josephschorr (approved):
LGTM

PR #3202, opened 2026-07-01
fix(datastore): return descriptive, engine-agnostic errors instead of raw driver errors

josephschorr:
## What

Two changes so the v1 API returns descriptive, engine-agnostic datastore errors rather than surfacing raw datastore-driver internals (SQLSTATE codes, `COPY from stdin` text, engine identity) to clients.

### 1. Surface the real error when a bulk COPY is aborted

`pgxcommon.BulkLoad` (used by the PostgreSQL and CRDB datastores) streams relationships via pgx `CopyFrom`. When the relationship source aborts the copy, pgx reports an opaque driver-level COPY-failure error that hides the real cause. `BulkLoad` now returns the source's own error — which is already a clean gRPC status / validation error — when the source aborted the copy, and leaves genuine DB-side failures untouched.

### 2. Sanitize unhandled raw driver errors at the shared error boundary

`rewriteError` is the single error-conversion choke point every v1 service funnels through. Its `default` case previously returned unhandled errors verbatim. It now routes unrecognized raw `*pgconn.PgError` (postgres/crdb) and `*mysql.MySQLError` (mysql) through a new `sanitizeDatastoreDriverError`, which logs the full error server-side and returns a descriptive, engine-agnostic gRPC status:

- query canceled → `Canceled`
- serialization / deadlock / lock-wait timeout → `Aborted` ("please retry")
- anything else → `Internal`

Existing typed error conversions are unaffected — they are matched before the `default` case.

## Testing

- New unit tests for both paths: bulk-load source-error recovery, and driver-error sanitization asserting no engine internals appear in the client-facing message.
- Existing `TestRewriteError` unchanged and passing.
- `go build ./...` clean and the affected package suites are green. The full integration suite (which requires database containers) was not run locally.


tstirrat15 (approved):
LGTM

PR #3210, opened 2026-07-02
fix(crdb): suppress row-level TTL deletes from the Watch API at the source

josephschorr:
## Problem

CockroachDB's row-level TTL job hard-deletes expired relationships, and those deletes appear in the changefeed backing the Watch API as ordinary DELETEs — CRDB provides no per-event origin marker (see cockroachdb/cockroach#106694).

The existing watch-side heuristic that was meant to filter them never actually fired: the changefeed is created without the `diff` option, so DELETE events have `after: null` and the deleted relationship's expiration is always `nil`, which forced `hasNonTTLEvent = true` on every delete. This was verified empirically against CRDB v26.1: a TTL-style delete (delete-only transaction, no metadata row, expired relationship) leaked through the Watch API. Consequently, the write-side transaction-metadata marker rows written to support that heuristic (for every delete-only transaction and every transaction touching expiring relationships) were pure write amplification.

## Change

- **Suppress at the source**: at datastore startup (watch enabled, CRDB >= 24.1), check `SHOW CREATE TABLE` for the `ttl_disable_changefeed_replication` storage parameter on `relation_tuple`/`relation_tuple_with_integrity` and set it if absent. This is the same parameter migration 0008 already applies to `transaction_metadata`. The check-before-set makes fleet-wide concurrent startups benign: after the first successful `ALTER`, every other pod sees the param and skips DDL entirely (steady state is one cheap read per table per start).
- **Fail-open**: if the runtime credentials lack `ALTER TABLE` privileges (or any other error), a warning is logged with the manual statement to run, and startup continues.
- **Remove the dead heuristic** in `processChanges` and all conditional marker writes: a `transaction_metadata` row is now written only when the caller supplies transaction metadata. The `IncludesExpiredAt` RWTOption is removed.
- **Rolling-upgrade safe**: the watch-side strip of the legacy `$spicedb_transaction_key` marker is retained (older writers still inject it within the GC window); stripped-to-empty metadata is not emitted.

## Behavior notes

- **CRDB < 24.1** (minimum supported remains 23.1.30): the parameter does not exist, so a startup warning is logged and TTL deletes continue to be emitted — identical to the (broken-heuristic) behavior before this change. v24.1.0 GA is the floor for suppression because it includes the fix for cockroachdb/cockroach#121589 (TTL deletes leaking past the filter on transaction retry).
- **Operator-set values are respected**: if the parameter is already present in the table DDL (any value), startup does not override it.
- **External changefeeds**: the storage parameter affects *any* changefeed over the relationship tables, not just SpiceDB's. Operators running their own changefeeds who want TTL deletes can opt back in per-changefeed with `ignore_disable_changefeed_replication`.
- A future SpiceDB changefeed wishing to report row GCs as events can likewise use `ignore_disable_changefeed_replication`.

## Testing

- New integration test asserting both relationship tables carry the parameter after startup.
- New end-to-end regression test: a delete performed in a session with `disable_changefeed_replication = true` (which sets the same `OmitInRangefeeds` transaction flag the TTL job sets under the storage parameter) is **not** emitted by Watch, while a subsequent normal delete **is** — proving the suppression machinery end-to-end and that SpiceDB's changefeed does not set `ignore_disable_changefeed_replication`.
- Updated `TransactionMetadataMarkingTest`: delete-only transactions no longer produce a metadata row; user-supplied metadata still does.
- Full `-tags datastore` CRDB suite green against v26.1.0.


internal/datastore/crdb/crdb_test.go
```diff
@@ -758,33 +776,109 @@ func TransactionMetadataMarkingTest(t *testing.T, rawDS datastore.Datastore) {
...
+	}
+}
 
-	err = cds.readPool.QueryFunc(ctx, func(ctx context.Context, rows pgx.Rows) error {
-		for rows.Next() {
-			var count int
-			err := rows.Scan(&count)
-			require.NoError(err)
-			require.Equal(3, count)
+func TTLChangefeedSuppressionWatchTest(t *testing.T, rawDS datastore.Datastore) {
```
miparnisari, line 806: why is this test in `crdb_test.go`? should it not be a generic datastore test to ensure we don't regress this in all the other datastores (present and future)?

josephschorr: Ideally, but this specific test relies on CRDB-specific calls, so I put it here

miparnisari: Discussed offline; not trivial to do

miparnisari (dismissed):

miparnisari (approved):

PR #3211, opened 2026-07-02
Experimental: changelog table for crdb

ecordell:
Note: for reference, hasn't been validated

## Description

CockroachDB changefeeds stop emitting `resolved` timestamps during bulk loads, which stalls SpiceDB's CRDB `Watch` (it relies on changefeed `resolved` checkpoints to know a revision is complete). This adds an **experimental, opt-in, CRDB-only** Watch path that doesn't depend on the changefeed for correctness, so Watch keeps advancing under bulk load.

Behind `--datastore-experimental-crdb-changelog-watch` (default **off** — when off, behavior is unchanged):

- Creates an append-only **`relationship_changelog`** table (hash-sharded PK on the HLC commit timestamp, row-level TTL) at init.
- Relationship, schema, and bulk-load writes **dual-write** into it in the same transaction.
- `Watch` **polls the table at present time** instead of consuming a changefeed. The table is append-only, so a present read sees every committed row; completeness comes from the cluster's **max clock offset** (checkpoint/cursor advances only to `clusterNow − maxOffset`), not changefeed `resolved`, so no write is skipped. A lightweight changefeed on the table is used only as a low-latency nudge, with the poll interval as the correctness backstop.

Notes:
- Checkpoint floor ≈ `maxOffset` (`--datastore-experimental-crdb-changelog-watch-max-offset`, default 250ms). Must be ≥ the cluster's `--max-offset` (CRDB default 500ms) or Watch can miss commits.
- Prototype scope: no cutover backfill, no integrity metadata in the changelog.

## Testing

Extensive unit and integration tests, but the impact to write throughput has not been measured, that could make this a dealbreaker.

TODO: run with `--datastore-engine cockroachdb --datastore-experimental-crdb-changelog-watch` and exercise Watch during a bulk load.


internal/datastore/common/changes.go
```diff
@@ -270,7 +270,7 @@ func (ch *Changes[R, K]) AddChangedDefinition(
 	case *core.CaveatDefinition:
 		delete(record.caveatsDeleted, t.Name)
 
-		if existing, ok := record.definitionsChanged[nsPrefix+t.Name]; ok {
+		if existing, ok := record.definitionsChanged[caveatPrefix+t.Name]; ok {
```
ecordell, line 273: this looks like a latent bug to me

docs/spicedb.md
```diff
@@ -91,6 +91,8 @@ spicedb datastore gc [flags]
       --datastore-disable-watch-support                                       disable watch support (only enable if you absolutely do not need watch)
       --datastore-engine string                                               type of datastore to initialize ("cockroachdb", "mysql", "postgres", "spanner") (default "memory")
       --datastore-experimental-column-optimization                            enable experimental column optimization (default true)
+      --datastore-experimental-crdb-changelog-watch                           EXPERIMENTAL: use a dual-written changelog table polled at a closed timestamp for Watch instead of a changefeed (CockroachDB driver only). Keeps Watch advancing during bulk loads at the cost of write latency.
```
miparnisari, line 94: > at the cost of write latency.

we should verify this before merge

docs/spicedb.md
```diff
@@ -414,6 +418,8 @@ spicedb serve [flags]
       --datastore-disable-watch-support                                                 disable watch support (only enable if you absolutely do not need watch)
       --datastore-engine string                                                         type of datastore to initialize ("cockroachdb", "mysql", "postgres", "spanner") (default "memory")
       --datastore-experimental-column-optimization                                      enable experimental column optimization (default true)
+      --datastore-experimental-crdb-changelog-watch                                     EXPERIMENTAL: use a dual-written changelog table polled at a closed timestamp for Watch instead of a changefeed (CockroachDB driver only). Keeps Watch advancing during bulk loads at the cost of write latency.
+      --datastore-experimental-crdb-changelog-watch-max-offset duration                 EXPERIMENTAL: safety margin subtracted from the cluster clock to derive the changelog Watch completeness cursor/checkpoint; MUST be >= the cluster's --max-offset (CockroachDB default 500ms) or Watch can miss commits (CockroachDB driver only) (default 250ms)
```
miparnisari, line 422: nit: these are the only flags that we added (to `serve`); why do they also appear under `gc` and `repair`?

internal/datastore/common/changes.go
miparnisari: 😱 yes, please send a different PR

internal/datastore/crdb/changelog.go
```diff
@@ -0,0 +1,309 @@
...
+	}, createChangelogTableSQL(gcWindow))
+}
+
+// operationToChangelog maps a tuple update operation to its stored string.
+func operationToChangelog(op tuple.UpdateOperation) (string, error) {
+	switch op {
+	case tuple.UpdateOperationCreate:
+		return "create", nil
+	case tuple.UpdateOperationTouch:
+		return "touch", nil
```
miparnisari, line 83: if i send a touch but the relationship already existed and nothing got written, will this writes an event to the changelog? if the answer is yes, this seems useless 🤔  what consumer would care about this?

internal/datastore/crdb/schema/schema.go
```diff
@@ -7,13 +7,14 @@ import (
...
-	TableRelationshipCounter = "relationship_counter"
-	TableTransactionMetadata = "transaction_metadata"
+	TableNamespace             = "namespace_config"
+	TableTuple                 = "relation_tuple"
+	TableTupleWithIntegrity    = "relation_tuple_with_integrity"
+	TableTransactions          = "transactions"
+	TableCaveat                = "caveat"
+	TableRelationshipCounter   = "relationship_counter"
+	TableTransactionMetadata   = "transaction_metadata"
+	TableRelationshipChangelog = "relationship_changelog"
```
miparnisari, line 17:
```suggestion
	TableChangelog = "changelog"
```

internal/datastore/crdb/watch_changelog.go
```diff
@@ -0,0 +1,671 @@
...
+	if err != nil {
+		return err
+	}
+	return sendChange(datastore.RevisionChanges{Revision: rev, IsCheckpoint: true})
+}
+
+// changelogRowFields holds the reconstructed non-key columns of a changelog
+// row, shared by the buffered and immediate scan paths.
+type changelogRowFields struct {
+	kind                         string
```
miparnisari, line 500: enum

internal/datastore/crdb/watch_changelog.go
```diff
@@ -0,0 +1,671 @@
...
+
+// feedChangelogRow reconstructs one changelog row and feeds it to the tracker.
+//
+// When normalizeCreate is true, a "create" operation is reported as Touch. The
+// shared change tracker (internal/datastore/common.Changes, used by the
+// buffered path and by the Postgres/MySQL watch implementations) only
+// distinguishes Touch vs Delete — at the row level a physical insert is
+// indistinguishable from a touch of a previously nonexistent row — so those
+// backends report inserts as Touch. The immediate path emits through
+// streamingChangeProvider, which does represent Create distinctly, so it passes
```
miparnisari, line 520: > The immediate path emits through streamingChangeProvider, which does represent Create distinctly

what's the reason for this and do we want to preserve it? seems weird that each "mode" returns different things

PR #3218, opened 2026-07-08
test: add test for cache's ttl behavior

tstirrat15:
## Description
We had questions about whether TTL behavior was working as expected in the new caching library so I wrote a test.

## Changes
* Add TTL test
* Remove `Wait()` from interface, since the implementation that required it (ristretto) no longer exists.
## Testing
Review.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -3,9 +3,11 @@ package cache
 import (
 	"math"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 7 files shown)


miparnisari (approved):

miparnisari (dismissed):

pkg/cache/cache_test.go
```diff
@@ -151,4 +147,38 @@ func TestCacheWithMetrics(t *testing.T) {
...
+			// wait for a bit
+			time.Sleep(3 * time.Minute)
+
+			// Retrieve the original key again; we're expecting not to find it.
+			_, found = cache.Get(StringKey("a key"))
+			assert.False(t, found, "expected key %s to be found", "a key")
+
+			cache.Close()
+
+			synctest.Wait()
```
ecordell: I would've expected this above the cache.Get?

ecordell:
```suggestion
```

Looked at this more and I think it can just be removed. The `synctest.Test` wrapper already waits for all goroutines to exit so this is redundant.

I thought moving it after the `time.Sleep` would make sense so that any goroutines that fire in otter machinery for TTLs have a chance to run before the `cache.Get`, but it looks like the TTL check fires syncronously when you call `cache.Get` so clock advance is enough.

miparnisari (dismissed):

miparnisari (approved):

PR #3219, opened 2026-07-08
feat: add query plan cache

jzelinskie:
## Description

Introduces a queryPlanCache that caches parsed schemas and compiled outlines behind singleflight
groups, replacing the per-request schema load and outline build in checkPermission, lookupResources, and lookupSubjects query plan paths.

Key correctness considerations:
- Bypass-sentinel schema hashes (transactions, dev mode, tests) skip both cache and singleflight to avoid coalescing requests that may read from different datastore revisions.
- MaxCost accounts for the entryWeight overhead (key string length) so entries are not immediately evicted.
- ApplyAdvisor clones CanonicalKeys before mutation so cached outlines are not corrupted by downstream advisor passes.

## Testing

Full consistency tests should handle this.
Just a thought now is maybe we'd want metrics on the cache hits for this.

### DIFF
--- a/internal/services/v1/permissions_queryplan.go
+++ b/internal/services/v1/permissions_queryplan.go
@@ -9,11 +9,9 @@ import (
 	caveatsimpl "github.com/authzed/spicedb/internal/caveats"
 	"github.com/authzed/spicedb/internal/dispatch"
 	"github.com/authzed/spicedb/pkg/datalayer"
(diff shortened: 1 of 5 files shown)


jzelinskie:
I'm not convinced that creating the outline is a bottleneck worth addressing just yet and this approach is quite naive.

PR #3220, opened 2026-07-08
fix(query): sound caveat resolution for recursive relations in the query planner

josephschorr:
## Summary

Fixes a set of correctness bugs in the experimental query planner's recursive relation handling (`pkg/query`), plus the test-infrastructure gaps that were hiding them. Scoped to the query planner; the legacy dispatch engine is untouched.

All changes are opt-in-path only: the query planner is gated behind `--experimental-query-plan` (default off), so production behavior is unchanged unless that flag is set.

## What was broken

**Caveat soundness on recursive relations (the important one).** For a schema like `relation member: user | group#member | group#member with somecaveat`, the planner tracked reachable objects with a *global* visited set and kept the caveat of the *first* path by which each object was reached. When the same object was later reached by a weaker (less-caveated) path, its descendants kept the stale, over-restrictive caveat. Result: a subject reachable by an unconditional path could be reported as only *conditionally* reachable — a wrong authorization decision. No cycle is required; a caveated diamond triggers it.

**Silent truncation at max recursion depth.** When a recursive traversal exceeded `MaxRecursionDepth`, the planner fell off the end of its loop and returned a *partial* result (or `NOT_MEMBER`) with no error — a "deny" that actually means "I gave up." The legacy engine errors here; the planner now does too.

**Missing query shapes.** The existence-probe query used for recursive self-edges specified no query shape, which panics under the validating datastore — the reason the caveat bug was never caught in tests. A second query-shape descriptor (`MatchingResourcesForSubject`) was missing from the forward validator, breaking index checking on SQL datastores.

**Inert consistency suite.** The query-plan consistency test skipped *every* assertion because a gate on the parent test node called `runtime.Goexit`. Check and LookupResources consistency now run by default on every test config.

## How it's fixed

- A canonical caveat-condition type (`internal/caveats/canonical.go`) represents a caveat expression as a DNF with idempotent `AND`, absorbing "unconditional", and subsumption. This makes it possible to detect when a newly-discovered path *weakens* the condition under which an object is reachable, and guarantees the traversal terminates.
- Both recursive BFS loops (`IterSubjects`/Check and `IterResources`) now run a semi-naive fixpoint: an object is re-expanded when its reaching condition weakens, and results are buffered and flushed at convergence so a caveat is never emitted before it is final. When no recursive edge is caveated, this reduces to exactly the previous BFS — no change to the common path.
- Depth exhaustion now yields a typed `MaxRecursionDepthError`.

## Testing

- New unit tests for the caveated-diamond soundness bug (Check, IterSubjects, IterResources), caveated-cycle termination, deep-chain depth errors, and the canonicalizer.
- Query-shape validator and existence-probe tests.
- `go test ./pkg/query/... ./internal/caveats/...` and the query-plan consistency integration suite pass; `-race` clean.

## Follow-ups (not in this PR)

- LookupSubjects diverges from the dispatcher on wildcard-with-exclusion schemas; left gated behind `TEST_QUERY_PLAN_SUBJECTS` with a pointer to the issue.
- The planner still carries per-path expiration through the graph, which the datastore already enforces on read; that merge is a separate cleanup.
- Gating cyclic-data support behind an explicit schema opt-in is deferred to a follow-up.


internal/caveats/canonical.go
```diff
@@ -0,0 +1,309 @@
...
+//
+// Two properties make that guarantee hold:
+//   - AND is idempotent and commutative: c1 ∧ c2 ∧ c1 collapses to {c1, c2}, so
+//     the set of distinct conjuncts is finite.
+//   - ⊤ (unconditional) is absorbing under OR: once an object is reachable
+//     unconditionally, no further path can weaken it.
+//
+// The zero value is the "false" condition (no disjuncts); use Top() for the
+// unconditional condition and FromExpression to derive one from a caveat.
+type Condition struct {
```
jzelinskie, line 27: This seems like a type that probably already existed for the non-query-planner code path. We should see if we can share this new code across both implementations.

internal/services/integrationtesting/queryconsistency/query_plan_consistency_test.go
```diff
@@ -81,24 +81,23 @@ func runQueryPlanConsistencyForFile(t *testing.T, filePath string) {
...
 
 	t.Run("lookup_resources", func(t *testing.T) {
-		if os.Getenv("TEST_QUERY_PLAN_RESOURCES") == "" {
-			t.Skip("Skipping IterResources tests due to deprectation: Set TEST_QUERY_PLAN_RESOURCES=true to enable")
-		}
 		runQueryPlanLookupResources(t, handle)
 	})
 
 	t.Run("lookup_subjects", func(t *testing.T) {
+		// LookupSubjects consistency is still gated: the query planner diverges from
```
jzelinskie, line 88: It's a nitpick but slapping a TODO in here just so it's obvious using tools like grep to see that this is remaining work would be nice quality of life change.

internal/caveats/canonical.go
```diff
@@ -0,0 +1,309 @@
...
+// an opaque (e.g. negated) subexpression. key is the canonical identity used for
+// dedup, sorting and comparison; expr is retained so the original expression can
+// be rebuilt, since the key alone cannot recover a caveat's context.
+type atom struct {
+	key  string
+	expr *core.CaveatExpression
+}
+
+// Top returns the unconditional (always-true) condition.
+func Top() Condition {
```
jzelinskie, line 42: Do we ever plan to modify this or are we using it as a constant? Might not be the worst to just make this be a variable so there's only one instance in memory.

Another nitpick might be to just call this `Unconditional` because it's not immediately obvious what "top" means when you're reading this in the context of the rest of the codebase.

jzelinskie (changes requested):

internal/caveats/canonical.go
josephschorr: I checked the dispatcher path. It tracks expressions in MembershipSet and combines them with the shared caveat builders, but it does not have an equivalent canonical condition type. Replacing that path with Condition would change expression representation and require compatibility tests, so I kept this PR scoped to the query planner. The canonical type is in internal/caveats for future reuse.

internal/services/integrationtesting/queryconsistency/query_plan_consistency_test.go
josephschorr: Added a TODO at the LookupSubjects consistency gate so the remaining wildcard exclusion issue is easy to find.

internal/caveats/canonical.go
josephschorr: Renamed Top/IsTop to Unconditional/IsUnconditional for clarity. The returned Condition has only a boolean and nil slice, so I kept value construction instead of introducing a mutable shared variable.

jzelinskie (approved):

PR #3221, opened 2026-07-09
ci: update benchmark mage targets

miparnisari:
Two goals:
- attempt to prevent these alarms from firing on every PR 
image

- attempt to prevent these random spikes that may be due to CPU contention 
image


tstirrat15 (approved):
This makes sense to me

PR #3222, opened 2026-07-10
chore: refactor migration code and tests

miparnisari:
## What this changes
  
  **Every datastore engine now gets a migration test.** Migratable engines   expose their   migrations in order plus the earliest one the current code can run against.   

The suite's   new `TestMigration` walks an empty database one migration at a time,   serving SpiceDB's   schema service against the half-migrated database from that earliest   migration onward —   writing a schema before each remaining migration and re-reading it after,   proving each   migration preserves what the last one wrote. Out-of-repo engines get it by   registering   themselves.

  **To enable it, engines register their own builders.** Each gains an   `enginebuilder.go`   registering with the new leaf package `pkg/cmd/datastore/dsconfig`, which   now owns   `Config`. `pkg/cmd/datastore` re-exports it and still links every in-repo   engine: **no   caller changes imports, no behavior changes.** `Config` had to move so the
  test suite can   build a datastore by engine name without depending on the command layer.
  
  Migration plumbing also moved out of `pkg/cmd/migrate.go` into   the exported   `pkg/datastore/migration`. 

**Important files to review:** 
- `pkg/datastore/{test/migration.go,migration/migration.go}`,   
- `pkg/cmd/datastore/{dsconfig/config.go,engines.go}`, 
- one   `enginebuilder.go`.


ecordell (dismissed):

internal/datastore/crdb/crdb_test.go
```diff
@@ -209,7 +209,7 @@ func TestCRDBDatastoreWithIntegrity(t *testing.T) { //nolint:tparallel
 	t.Parallel()
 	b := testdatastore.RunCRDBForTesting(t, crdbTestVersion())
 
-	test.All(t, crdbFactory.NewTester(test.DatastoreTesterFunc(func(t testing.TB, revisionParameters test.RevisionParameters, watchBufferLength uint16) (datastore.Datastore, error) {
+	test.AllWithExceptions(t, crdbFactory.NewTester(test.DatastoreTesterFunc(func(t testing.TB, revisionParameters test.RevisionParameters, watchBufferLength uint16) (datastore.Datastore, error) {
```
miparnisari, line 212: because this is the integrity variant of the CRDB tests. we don't need them both to run

internal/datastore/crdb/enginebuilder.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, errors.New("max-retries could not be cast to uint8")
+	}
+
+	watchChangeBufferMaximumSize, err := common.WatchBufferSize(opts.WatchChangeBufferMaximumSize)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewCRDBDatastore(
```
miparnisari, line 40: FYI this entire file is just moving code around, no changes (except for the `init` above). Same goes for every other engine.

internal/datastore/proxy/revision_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/authzed/spicedb/internal/datastore/revisions"
+	"github.com/authzed/spicedb/pkg/datastore"
+)
+
+// mustParseRevisionForTest parses a test revision string. It mirrors
+// revisionparsing.MustParseRevisionForTest, which cannot be used here: that
+// package imports the datastore engines, which import this package.
+func mustParseRevisionForTest(revisionStr string) datastore.Revision {
```
tstirrat15, line 11: Is there a reason we don't put this in its own package and have both import it?

pkg/datastore/test/migration.go
```diff
@@ -0,0 +1,217 @@
...
+// against an empty, unmigrated database:
+//  1. Each migration is applied one at a time, verifying that it applies
+//     cleanly and that the recorded version advances to it.
+//  2. Once the datastore reaches the engine's verifiable migration (the
+//     earliest one the current code can operate against), SpiceDB's schema
+//     service is served against it and a schema is written and read back
+//     through the WriteSchema and ReadSchema APIs.
+//  3. Before each remaining migration a new schema is written; after the
+//     migration that schema must still be readable, verifying that each
+//     migration preserves the data written through the API before it.
```
tstirrat15, line 51: I'm not opposed to us asserting this, but is this in response to something that we've observed, or is it anticipatory?

pkg/datastore/test/migration.go
```diff
@@ -0,0 +1,217 @@
...
+		),
+	)
+	v1.RegisterSchemaServiceServer(srv, v1svc.NewSchemaServer(v1svc.SchemaServerConfig{
+		CaveatTypeSet:       caveattypes.Default.TypeSet,
+		ExpiringRelsEnabled: true,
+	}, validator))
+
+	go func() {
+		if err := srv.Serve(listener); err != nil {
+			log.Err(err).Msg("migration test schema service stopped serving")
```
tstirrat15, line 167: Could use `assert.NoError` here

tstirrat15 (approved):
LGTM, see comments

internal/datastore/proxy/revision_test.go
miparnisari: i can do as a follow!

pkg/datastore/test/migration.go
miparnisari: Neither. It's a test that we never had, making sure that data written in migration X is still readable at migration X+ 1

PR #3233, opened 2026-07-13
chore: fix broken test

tstirrat15:
## Description
This test was [failing reliably](https://github.com/authzed/spicedb/actions/runs/29281782343/job/86924624271?pr=3218) prior to this.

## Changes
* Use an atomic for `loadCalls`
* Use an errgroup for the spawned goroutines
* Use `Go()` instead of `Add()`
## Testing
Review. See that tests pass.


josephschorr (approved):
LGTM

PR #3237, opened 2026-07-15
docs(fdw): fix has_permission filter examples

### LINKED ISSUE #3236: postgres-fdw docs: `has_permission = true` examples do not work
The Lookup Resources / Lookup Subjects examples in `internal/fdw/README.md`
(and a few in `Architecture.md`) filter on `AND has_permission = true`, but
running them as documented fails:

```
ERROR:  unsupported feature or query type: unsupported select statement: operation not supported: column_ref:{...sval:"has_permission"...}
```
```
ERROR:  unsupported feature or query type: unsupported select statement: operation not supported: column_ref:{...sval:"has_permission"...}
```

Postgres folds `= true` into a bare column reference before pushdown, and the
proxy rejects `has_permission` in WHERE clauses.

Writing the filter as `has_permission IS true` works through postgres_fdw
(it's evaluated locally as a `Filter` instead of being pushed down), so the
README examples should use that. The `Architecture.md` examples describe SQL
as received by the proxy, where no local filtering exists — those should just
drop the condition.

capybara-alt:
Fixes #3236

The documented `AND has_permission = true` examples fail when run: Postgres
folds `= true` into a bare column reference before pushdown, and the proxy
rejects `has_permission` in WHERE clauses.

## Changes

* `README.md`: use `has_permission IS true` in the Lookup Resources / Lookup
  Subjects examples — `IS TRUE` is not pushed down by postgres_fdw and gets
  evaluated locally, so the query works as intended.
* `Architecture.md`: drop the `has_permission` condition from the
  handler-dispatch and cursor examples entirely — these show SQL as received
  by the proxy, where no local filtering exists and neither form is supported.

## Testing

Verified against a live setup (SpiceDB + FDW proxy + Postgres 17 with
postgres_fdw): the original examples fail with `operation not supported:
column_ref`, the updated README examples return the expected rows
(`EXPLAIN VERBOSE` confirms `IS TRUE` becomes a local `Filter` node), and the
updated Architecture.md queries dispatch to LookupResources/LookupSubjects.

### DIFF
--- a/internal/fdw/Architecture.md
+++ b/internal/fdw/Architecture.md
@@ -254,8 +254,7 @@ The permissions table supports three distinct query patterns, each mapping to a 
    ```sql
    SELECT resource_id FROM permissions
    WHERE resource_type = 'doc' AND permission = 'view'
-     AND subject_type = 'user' AND subject_id = 'alice'
-     AND has_permission = true;
+     AND subject_type = 'user' AND subject_id = 'alice';
    ```
    → Calls `LookupResources(user:alice, view, doc)`
 
@@ -264,8 +263,7 @@ The permissions table supports three distinct query patterns, each mapping to a 
    SELECT subject_id FROM permissions
    WHERE resource_type = 'doc' AND resource_id = '1'
      AND permission = 'view'
-     AND subject_type = 'user'
-     AND has_permission = true;
+     AND subject_type = 'user';
    ```
    → Calls `LookupSubjects(doc:1, view, user)`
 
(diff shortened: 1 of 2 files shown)


josephschorr:
> * `README.md`: use `has_permission IS true` in the Lookup Resources / Lookup
>   Subjects examples — `IS TRUE` is not pushed down by postgres_fdw and gets
>   evaluated locally, so the query works as intended.

Is this being added solely to make sure the "missing caveat context" results are not included?

PR #3239, opened 2026-07-19
fix: Handle concurrent MemDB writes

### LINKED ISSUE #3212: memdb: concurrent WriteRelationships can be silently lost
Disclaimer: This report is mostly AI-generated but it describes a real issue we are encountering in CI and I have manually verified the reproduction instructions. 

macos, linux (race is platform-independent; observed on both)

arm64, amd64

v1.52.0 (`spicedb serve-testing`); by code inspection the race is still present on `main`.

### Steps to Reproduce

1. Start `spicedb serve-testing`.
2. In a loop: issue **two concurrent** `WriteRelationships` requests (2 TOUCHes each, disjoint resources), then immediately do a `fully_consistent` `ReadRelationships` for both resources.
3. Repeat ~10–20k rounds.

Python repro script attached below. On an M-series MacBook it hits ~1 loss per 4,000 rounds:

```
round 1190: doc b-1190 acked but read back []
round 6263: doc a-6263 acked but read back []
...
done: 20000 rounds, 5 lost-write events
```

This seems to depend a bit on the machine utilization, so running the repro script multiple times in parallel helps.

### Expected Result

Every acked `WriteRelationships` is visible to a subsequent `fully_consistent` read.

### Actual Result

Occasionally one of the two writes is acked OK but **all of its relationships are invisible** to an immediately following `fully_consistent` read. The loss is transaction-granular and self-heals on the next commit to the datastore, so it is only observable in the window between the two commits and whatever write comes next.

<details><summary>Minimum reproducible example</summary>
<p>

```Python
"""Sharp probe for the memdb revision-order inversion race.

Each round: 2 concurrent WriteRelationships (2 TOUCHes each), then an
immediate fully-consistent read of all 4 rels. If commit order inverted
revision order, the later-committed write's rels are invisible at head.
"""

import asyncio
import sys
import uuid

import grpc
from authzed.api.v1 import (
    AsyncClient,
    Consistency,
    ObjectReference,
    ReadRelationshipsRequest,
    Relationship,
    RelationshipFilter,
    RelationshipUpdate,
    SubjectReference,
    WriteRelationshipsRequest,
    WriteSchemaRequest,
)
from grpcutil import insecure_bearer_token_credentials

SCHEMA = """
definition user {}
definition document {
    relation viewer: user
    relation editor: user
}
"""

ENDPOINT = "localhost:30051"


def touch(doc: str, relation: str) -> RelationshipUpdate:
    return RelationshipUpdate(
        operation=RelationshipUpdate.OPERATION_TOUCH,
        relationship=Relationship(
            resource=ObjectReference(object_type="document", object_id=doc),
            relation=relation,
            subject=SubjectReference(
                object=ObjectReference(object_type="user", object_id="tom")
            ),
        ),
    )


async def main() -> None:
    key = f"repro-inv-{uuid.uuid4().hex[:8]}"
    client = AsyncClient(ENDPOINT, insecure_bearer_token_credentials(key))
    await client.WriteSchema(WriteSchemaRequest(schema=SCHEMA))
    rounds = int(sys.argv[1]) if len(sys.argv) > 1 else 20000
    lost_events = 0

    for r in range(rounds):
        doc_a = f"a-{r}"
        doc_b = f"b-{r}"

        async def write(doc: str) -> bool:
            try:
                await client.WriteRelationships(
                    WriteRelationshipsRequest(
                        updates=[touch(doc, "viewer"), touch(doc, "editor")]
                    )
                )
                return True
            except grpc.aio.AioRpcError as e:
                print(f"round {r}: write {doc} error {e.code().name}")
                return False

        ok_a, ok_b = await asyncio.gather(write(doc_a), write(doc_b))

        present: set[tuple[str, str]] = set()
        for doc in (doc_a, doc_b):
            async for resp in client.ReadRelationships(
                ReadRelationshipsRequest(
                    relationship_filter=RelationshipFilter(
                        resource_type="document", optional_resource_id=doc
                    ),
                    consistency=Consistency(fully_consistent=True),
                )
            ):
                present.add((doc, resp.relationship.relation))

        for doc, ok in ((doc_a, ok_a), (doc_b, ok_b)):
            if not ok:
                continue
            got = {rel for d, rel in present if d == doc}
            if got != {"viewer", "editor"}:
                lost_events += 1
                print(f"round {r}: doc {doc} acked but read back {sorted(got)}")

        if r % 5000 == 0 and r > 0:
            print(f"...{r} rounds, {lost_events} losses so far")

    print(f"done: {rounds} rounds, {lost_events} lost-write events")


asyncio.run(main())

```

</p>
</details> 

<details><summary>Analysis</summary>
<p>
In `internal/datastore/memdb/memdb.go`, `ReadWriteTx`:

- stamps `newRevision := mdb.newRevisionID()` at **transaction entry**, *before* the user function runs and before the write lock is acquired (the `memdb.Txn` is created lazily via `txSrc` on first use);
- on success appends `snapshot{newRevision, snap}` to `mdb.revisions` in **commit order**.

Interleaving:

1. Tx A enters, stamps revision **100**, gets preempted before creating its txn.
2. Tx B enters, stamps revision **101**, acquires the write txn first, commits → `revisions = [..., {101, snapB}]`.
3. Tx A acquires the write txn, commits → `revisions = [..., {101, snapB}, {100, snapA}]` — out of order.

Now `headRevisionNoLock()` returns the *last* element's revision (**100**), and `SnapshotReader`'s `sort.Search` for the first entry `>= 100` lands on `{101, snapB}` — a snapshot taken **before A committed**. A fully-consistent read therefore misses all of A's acked writes. Any later commit appends a genuinely newer revision and the head read resolves to a complete snapshot again, which is why the loss is transient.

This is distinct from #1545 / #1547: that path (serialization retries exhausted) correctly returns `DEADLINE_EXCEEDED` to the caller — verified in the same repro under heavier load (230 errored writes, none silently lost). The inversion above is the genuinely silent path, and likely what the original #1545 reporter observed ("responded OK, but actually not written").

Possible fixes: stamp the revision at commit time under the same lock that appends to `mdb.revisions`, or insert the snapshot in revision order and make `headRevisionNoLock` return the max.

A Go-level reproduction in the style of `TestConcurrentWriteRelsError` (two goroutines, read-back after both `ReadWriteTx` calls return) should hit it as well.
</p>
</details>

Naganathan05:
## Description

  Fixes #3212: concurrent writes to the `memdb` datastore could acknowledge a `WriteRelationships` call whose data was then invisible to an immediately following fully-consistent read.

**Root cause:** within a single `ReadWriteTx`, `mdb.Lock()` is acquired and released in three separate windows (assigning the revision number, claiming the write-transaction slot, and committing + recording the snapshot) rather than held for the whole transaction. 
Between those windows, another transaction can fully commit, so two transactions can be recorded in a different order than their revision numbers were assigned in. Snapshots were recorded via a plain `append`, so `mdb.revisions` could end up out of order, and `SnapshotReader`'s binary search assumes it's sorted, so it could resolve to a stale snapshot that predates the write being requested.

**Fix:** Holding one lock for a transaction's entire duration would also fix the ordering, but at the cost of serializing all transactions. 
Instead, `insertRevisionSnapshot` inserts each snapshot at its correct sorted position based on revision ID, regardless of the order transactions actually commit in.

  ## Testing
  
  Added `TestConcurrentWriteRevisionInversionLosesRead`, which forces the scenario that triggers the bug and asserts a read at a transaction's own revision sees its write.

Test result without adding the fix for insertion:

  image

  ## References

  - Fixes #3212


Naganathan05:

Naganathan05:
Requires review @miparnisari

internal/datastore/memdb/memdb_test.go
```diff
@@ -113,6 +113,75 @@ func TestConcurrentWriteRelsSucceed(t *testing.T) {
 	require.NoError(g.Wait())
 }
 
+// TestConcurrentWriteRevisionInversionLosesRead reproduces concurrent writes to MemDB
+func TestConcurrentWriteRevisionInversionLosesRead(t *testing.T) {
```
miparnisari: does this test reproduce the error 100% of the time? if not, can we make it so? or at least make it so that it fails a large percentage of the runs?

e.g. run a `for` loop 1000 times where in each iteration we spin up two concurrent goroutines?

internal/datastore/memdb/memdb_test.go
```diff
@@ -113,6 +113,75 @@ func TestConcurrentWriteRelsSucceed(t *testing.T) {
...
+
+	revB, err := ds.ReadWriteTx(ctx, func(ctx context.Context, rwt datastore.ReadWriteTransaction) error {
+		return rwt.WriteRelationships(ctx, []tuple.RelationshipUpdate{
+			tuple.Touch(tuple.MustParse("document:doc-b#viewer@user:tom")),
+		})
+	}, options.WithDisableRetries(true))
+	require.NoError(err)
+
+	close(releaseA)
+	<-aDone
```
miparnisari: why do we need all this synchronization? can we not use a `waitgroup` to spin up both goroutines?

internal/datastore/memdb/revisions.go
```diff
@@ -38,6 +42,13 @@ func (mdb *memdbDatastore) newRevisionID() revisions.TimestampRevision {
 	return created
 }
 
+func (mdb *memdbDatastore) insertRevisionSnapshot(newRevision revisions.TimestampRevision, schemaHash string, snap *memdb.MemDB) {
+	insertAt := sort.Search(len(mdb.revisions), func(i int) bool {
+		return mdb.revisions[i].revision.GreaterThan(newRevision)
+	})
+	mdb.revisions = slices.Insert(mdb.revisions, insertAt, snapshot{newRevision, schemaHash, snap})
```
miparnisari: nit: document this function to say that it's assumed that the caller already acquired the mutex

internal/datastore/memdb/memdb.go
```diff
@@ -374,7 +374,7 @@ func (mdb *memdbDatastore) ReadWriteTx(
 		// Create a snapshot and add it to the revisions slice
 		schemaHash := mdb.getCurrentSchemaHashNoLock()
 		snap := mdb.db.Snapshot()
-		mdb.revisions = append(mdb.revisions, snapshot{newRevision, schemaHash, snap})
+		mdb.insertRevisionSnapshot(newRevision, schemaHash, snap)
```
miparnisari: i'm confused... you said

> Root cause: a transaction's revision number is stamped before it acquires the write lock

isn't the solution to acquire the lock? and which lock are you talking about? i see we're acquiring on line 249 already.

internal/datastore/memdb/memdb_test.go
```diff
@@ -113,6 +113,75 @@ func TestConcurrentWriteRelsSucceed(t *testing.T) {
...
+	indexOf := func(r datastore.Revision) int {
+		mds.RLock()
+		defer mds.RUnlock()
+		for i, snap := range mds.revisions {
+			if snap.revision.Equal(r) {
+				return i
+			}
+		}
+		return -1
+	}
```
miparnisari: nit: move this to `revisions.go`. we could reuse this in the future.

internal/datastore/memdb/memdb_test.go
Naganathan05: Yes, its deterministic. The test fails 100% of the time if the code fix introduced is removed.
The test manually simluates goroutine B which has the latest revision number to complete its execution before goroutine A which has the earlier revision number. 
This is done using channels and its the same reason why wait groups are not used as they will not ensure the order of execution that goroutine B completes the execution before A is unblocked and proceeds to commit..

Added docstrings mentioning the test scenario

internal/datastore/memdb/memdb_test.go
Naganathan05: For the same reason mentioned in previous comment. Channels are mainly used here to make goroutine B complete its execution before A and using wait group can relatively complicate things as it can only ensure the completion of all goroutine execution at a point.

internal/datastore/memdb/revisions.go
Naganathan05: Made the changes

internal/datastore/memdb/memdb_test.go
Naganathan05: Moved it.

internal/datastore/memdb/memdb.go
Naganathan05: Yes the lock `mdb.Lock()` is already existing as you mentioned, but this is released and acquired thrice in `ReadWriteTx` by a single goroutine/transaction. 
In between the release and acquire actions, another transaction B is able to run its own ReadWriteTx to completion, including its commit and snapshot recording before A resumes.

A single global lock can be done per transaction for `ReadWriteTx`, but this would affect the concurrency and serialize all the transaction in this operation.
Instead we can ensure that the snapshots are inserted in the sorted order as per their revision ID regardless of the transaction's commit order. This is the fix that is implemented in this PR.

There is a little gap in the PR description, will update it mentioning this.

internal/datastore/memdb/memdb_test.go
miparnisari: Your test didn't have 100% pass rate when ran many times (there was another bug - the collision guard in `newRevisionID`).

I've updated it to account for this, plus i've added more assertions.

internal/datastore/memdb/memdb.go
miparnisari: Your fix addresses the symptom but not the root cause. I've updated the code to fix the root cause.

Naganathan05:
@miparnisari 
 Went over the code changes, it makes sense that you moved the task of assigning revision number inside `activeWriteTxn`. This should address the issue. 
Any additional test addition required ?

Naganathan05:
@miparnisari
Anything pending to be done ?

miparnisari:
Since I made changes to your PR, i'm waiting for someone else in the maintainer team to review. @tstirrat15 can you review?

internal/datastore/memdb/memdb.go
```diff
@@ -247,123 +256,130 @@ func (mdb *memdbDatastore) ReadWriteTx(
 		}
 
 		mdb.Lock()
-		defer mdb.Unlock()
+		defer mdb.Unlock() // TODO is this defer correct? it runs at the end of the function, not at the end of the for loop's iteration
```
tstirrat15, line 259: I imagine that it's suboptimal but also correct. Otherwise we'd need to have it at every function exit.

pkg/datastore/test/relationships.go
```diff
@@ -2263,6 +2263,136 @@ func ConcurrentWriteDeadlockTest(t *testing.T, tester DatastoreTester) {
...
+//   - when the two revisions are concurrent (e.g. Postgres snapshots of
+//     overlapping transactions), neither write may be visible at the other's
+//     revision
+//   - the head revision must see every committed write
+//
+// This test is safe even for datastores that serialize write transactions with
+// a global lock (e.g. memdb): transaction A performs no reads or writes until
+// transaction B has fully committed, so the two never contend for the write
+// lock.
+func ConcurrentWriteRevisionVisibilityTest(t *testing.T, tester DatastoreTester) {
```
tstirrat15, line 2282: I wonder if this would be better as a fuzz test than a unit test. Running a thousand iterations of DB read + write sounds pretty heavy.

tstirrat15 (commented):
See comments, otherwise LGTM

tstirrat15 (dismissed):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Caveats: compiled caveats (and their CEL environments) are now cached per schema version — hung off the stored schema (`ReadOnlyStoredSchema`) and rebuilt only when the schema changes — rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
 
 ### Fixed
+- MemDB: overlapping write transactions could violate every snapshot-consistency invariant of the datastore — a committed write could be invisible at its own returned revision (breaking read-your-writes, e.g. an at-exact-snapshot `Check` right after `WriteRelationships`), a later commit could leak into reads at an earlier revision, a write visible at one revision could be missing at a later one (including head), and two concurrent transactions could be assigned the same revision. Revisions are now assigned at write-transaction acquisition, where writer serialization makes the two orders identical. (https://github.com/authzed/spicedb/pull/3239)
```
miparnisari: move to under [unreleased] please

miparnisari (approved):

internal/datastore/memdb/memdb.go
miparnisari: how is it correct? if we re-attempt the function, the lock will not be unlocked before the subsequent attempts no?

PR #3243, opened 2026-07-20
fix(postgres): guard strict reads against replica revision gaps

### LINKED ISSUE #2525: Using Read Replica on AWS Aurora Postgres causing errors

josephschorr:
## Problem

Fixes #2525.

With Postgres read replicas enabled, SpiceDB intermittently returned `object definition <ns> not found` (surfaced as `Unknown`/`FailedPrecondition`) even when the configured follower delay exceeded the replication latency. It reproduced only under load and went away when replicas were removed.

## Root cause

Reads filter rows with `pg_visible_in_snapshot(created_xid, revision.snapshot)` — a pure function of the revision's snapshot. A snapshot makes visible any committed transaction with an xid `< xmax` that is not in its in-progress (`xip`) list, so visible data can come from transactions with an xid **at or above `xmin`**.

The strict read-replica guard, however, only asserted that transaction `xmin - 1` had resolved on the replica. When a concurrent, still-open transaction holds `xmin` below a freshly committed write, a lagging replica that has applied `xmin-1` but not that write passes the guard while the row is absent → zero rows, no error, no fallback → not-found. At production write QPS there is almost always an in-progress write with a lower xid than the most recently committed ones, which is why it only showed up under load.

## Fix

Replace the single-transaction check with an inline snapshot-**domination** check, evaluated on the replica's own connection so it stays valid behind a load balancer (which is why the check must be inline rather than a separate `CheckRevision` round-trip):

- the replica's current snapshot `xmax` must have reached the revision's `xmax`; and
- the replica must not still consider in-progress any transaction below the revision's `xmax` that the revision treats as committed.

This is the inline equivalent of the comparison `CheckRevision` performs. A single visible xid is **not** sufficient, because commit order does not follow xid order: a lower-xid transaction the revision sees as committed can still be replaying on the replica after higher xids have been applied.

## Testing

New datastore tests (`//go:build datastore && postgres`) stand up a real primary + streaming hot-standby replica in containers and drive replication lag deterministically with `pg_wal_replay_pause()`:

- `TestStrictReadGuardDetectsVisibleDataGapOnReplica` — drives the production guard SQL directly; it raises `replica missing revision` while behind and succeeds after catch-up.
- `TestReplicaLagStrictReadFallsBackForNamespace` — end-to-end through the strict replicated datastore; the read now falls back to the primary and finds the namespace instead of reporting it not-found.
- `TestReplicaServesCaughtUpRevisionWithoutFallback` — verifies a caught-up (min-latency) revision is still served **by the replica** with no fallback, asserted via the fallback metric.

Verified against Postgres 14, 17, and 18; `mage testds:postgres` is green.

Routing `FullyConsistent` reads to the primary is a separate optimization and is intentionally out of scope here.


AlexeyRaga:
Hey, great work, can it be merged now?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Caveats: compiled caveats (and their CEL environments) are now cached per schema version — hung off the stored schema (`ReadOnlyStoredSchema`) and rebuilt only when the schema changes — rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
 
 ### Fixed
+- Postgres: read replicas no longer intermittently return `object definition not found` under load. The strict read-replica guard now verifies that the replica's snapshot has caught up to the revision being read (snapshot domination) instead of checking a single transaction id, so it correctly falls back to the primary when the replica is missing a visible revision. (https://github.com/authzed/spicedb/pull/3243)
```
miparnisari: move to under [unreleased]

miparnisari (changes requested):
I tested this PR locally using the commands described in https://github.com/authzed/spicedb/issues/2525#issuecomment-3921159372 and I can still reproduce the error.

internal/datastore/postgres/strictreader.go
```diff
@@ -52,23 +52,55 @@ func (srqf strictReaderQueryFuncs) addAssertToSelectSQL(sql string) string {
...
+func (s pgSnapshot) replicaContainsRevisionPredicate() string {
+	revisionXip := make([]string, 0, len(s.xipList))
+	for _, xip := range s.xipList {
+		revisionXip = append(revisionXip, fmt.Sprintf("'%d'::xid8", xip))
+	}
+
+	return fmt.Sprintf(
+		`pg_snapshot_xmax(pg_current_snapshot()) >= '%d'::xid8 `+
+			`AND NOT EXISTS (`+
+			`SELECT 1 FROM pg_snapshot_xip(pg_current_snapshot()) AS replica_xip `+
```
miparnisari, line 148: the value of `pg_current_snapshot()` can change between 100 and 102, right? the tests you wrote don't account for this scenario. I think this is what is causing me to not see the issue fixed. And i think _one_ fix would be to run _this specific query_ with `repeatable read` isolation level. I'm been running the repro script for 5 hours with this change and i wasn't able to reproduce it

josephschorr:
@miparnisari thank you — you were exactly right, and your repro instructions on #2525 are what made this findable at all. Without your review I'd have merged a fix that still dropped reads under load.

## You were right about `pg_current_snapshot()`

Under READ COMMITTED, every statement of a multi-statement simple query takes its **own** snapshot. Demonstrable with nothing but psql and a concurrent writer:

```sql
-- one batch, one implicit transaction, a commit lands during the sleep
SELECT pg_current_snapshot()::text, pg_sleep(3);  --> 740:740:
SELECT pg_current_snapshot()::text;               --> 741:741:
```

So the guard raced exactly where you pointed: the `WHERE` filter and the trailing `DO ... ASSERT` are evaluated against different snapshots. When the replica catches up in between, the filter drops every row (its snapshot was behind) while the assertion passes (its snapshot is not) — **zero rows, no error**, i.e. `object definition not found`. In production that window is microseconds wide, which is why it only appears under sustained load.

## Reproducing it deterministically

New test `TestStrictReadGuardHoldsAcrossStatementBoundary`: 16 goroutines hold guarded reads in flight against a replica with WAL replay paused, then replay is resumed underneath them, so some of those reads straddle the instant the replica catches up. Any read that returns zero rows *and* no error is the bug. Against the previous version of this PR it hits within ~10 seconds (cycle 1 on PG 18, cycle 2 on PG 14).

The tests I had written missed it because they pause replay for the duration of the read — the replica never catches up mid-query, which is the only moment that matters.

## Why not REPEATABLE READ

Your suggestion is the right instinct and it genuinely does fix the race — both `BEGIN ISOLATION LEVEL REPEATABLE READ` and a session-level `default_transaction_isolation` pin one snapshot across both statements (verified). Neither is usable here, though:

- Making `SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;` the first statement of the batch breaks the read outright: pgx hands back only the **first** result set, so the SELECT's rows are silently discarded. Verified — zero rows, no error, the same failure by a different route.
- Setting it at the session/connection level isn't reliable behind a transaction-pooling load balancer, which is the exact deployment this strict path exists to serve.

## What it does instead

The guard now **raises from within the SELECT itself**, so its verdict is tied to the snapshot the rows are actually read under. There is no second statement left to disagree with it. The `DO` block stays purely as a backstop: it can turn a read that slipped through into an error, never the reverse, because the condition is monotonic as snapshots advance.

Two subtleties, both verified against live Postgres:

- **The guard expression has to stay non-volatile.** A volatile expression is pushed into the scan's per-row filter, where it never runs for a query that matches nothing — precisely the lagging-replica case, since the rows are missing *because* of the lag. Built only from stable functions it becomes a gating `One-Time Filter`, evaluated exactly once before the scan regardless of how many rows match. It also keeps parallel plans available; a `set_config`-based variant I tried made every replica read parallel-unsafe.
- **It casts to `boolean`, not `xid8`.** Before PG 15, `xid8`'s input function silently parses unparseable text as `0` — no error, guard disabled entirely. My first attempt passed on 18 and failed on 14, which is how this surfaced.

## Verification

The race test is green on Postgres **14, 15, 16, 17 and 18** (~30k guarded reads per run, zero silent empties), alongside the existing strict-read tests and the full `TestPostgresDatastore` suite. Also added: the guard must raise on a lagging replica even when the query matches no rows, and a unit test pinning the raise's error message to `IsReplicationLagError`.

Would you re-run your repro from #2525 against this? I'd like your confirmation before merging. One thing to expect: `FullyConsistent` reads still get routed to a replica and then fall back to the primary, so on that workload you should see fallbacks and correct answers rather than errors — making that path skip the replica entirely is a separate optimization, still out of scope here.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Caveats: compiled caveats (and their CEL environments) are now cached per schema version — hung off the stored schema (`ReadOnlyStoredSchema`) and rebuilt only when the schema changes — rather than rebuilt on every check, reducing check cost for schemas with many caveats (https://github.com/authzed/spicedb/pull/3166)
 
 ### Fixed
+- Postgres: read replicas no longer intermittently return `object definition not found` under load. The strict read-replica guard now verifies that the replica's snapshot has caught up to the revision being read (snapshot domination) instead of checking a single transaction id, and raises from within the read itself rather than from a trailing assertion, so a replica that catches up mid-query can no longer let an incomplete read through. In both cases the read correctly falls back to the primary. (https://github.com/authzed/spicedb/pull/3243)
```
miparnisari: move to [unreleased]

miparnisari (approved):
been running the repro script for an hour and no errors 🎉 

and verified that 3 out of 4 tests that pass now, fail on `main`

PR #3247, opened 2026-07-24
test: add test for catching cache memory leaks

miparnisari:
Would have caught https://github.com/authzed/spicedb/issues/3200

## Testing

`git checkout v1.54 && git cherry-pick da30538993fcd0699da6c4cc6747f11220cb19a7 && mage test:image`

```
found 3 packages with //go:build image:
  ./cmd/spicedb/integration
  ./cmd/spicedb/memoryprotection
  ./pkg/cache
ok      github.com/authzed/spicedb/cmd/spicedb/integration      74.872s
?       github.com/authzed/spicedb/cmd/spicedb/memoryprotection [no test files]
--- FAIL: TestCacheFillingDoesNotTripMemoryLimiter (81.44s)
    memory_limiter_integration_test.go:98: round 1/20: all 10000 checks returned HAS_PERMISSION
    memory_limiter_integration_test.go:98: round 2/20: all 10000 checks returned HAS_PERMISSION
    memory_limiter_integration_test.go:98: round 3/20: all 10000 checks returned HAS_PERMISSION
    memory_limiter_integration_test.go:98: round 4/20: all 10000 checks returned HAS_PERMISSION
    memory_limiter_integration_test.go:98: round 5/20: all 10000 checks returned HAS_PERMISSION
    memory_limiter_integration_test.go:98: round 6/20: all 10000 checks returned HAS_PERMISSION
    memory_limiter_integration_test.go:97: 
                Error Trace:    /Users/miparnisari/Documents/GitHub/spicedb/pkg/cache/memory_limiter_integration_test.go:362
                                                        /Users/miparnisari/Documents/GitHub/spicedb/pkg/cache/memory_limiter_integration_test.go:97
                Error:          Received unexpected error:
                                round 7: check for doc-6548 rejected with RESOURCE_EXHAUSTED (rpc error: code = ResourceExhausted desc = server rejected the request because memory usage is too high) — the server's heap crossed GOMEMLIMIT under cache-filling load
                Test:           TestCacheFillingDoesNotTripMemoryLimiter
    memory_limiter_integration_test.go:205: last spicedb logs:
```


magefiles/test.go
```diff
@@ -55,7 +55,14 @@ func (Test) unit(ctx context.Context, coverage bool) error {
 // Image Run tests that run the built image
 func (Test) Image(ctx context.Context) error {
 	mg.Deps(Build{}.Testimage)
-	return goDirTest(ctx, "./cmd/spicedb", "./...", "-tags", "image")
+	dirs, err := findDirsWithBuildTag("image")
```
ecordell, line 58: is there a reason to do this vs. just `goTest(ctx, "./...", "-tags", "image")`? why find them ourselves instead of letting `go test` find them?

pkg/cache/memory_limiter_integration_test.go
```diff
@@ -0,0 +1,397 @@
...
+	rounds := envIntOr(t, "CACHE_FILL_ROUNDS", 20)
+
+	ctx, cancel := context.WithTimeout(t.Context(), 20*time.Minute)
+	defer cancel()
+
+	net, err := network.New(ctx)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = net.Remove(t.Context()) })
+
+	startPostgres(ctx, t, net.Name)
```
ecordell: is there a reason this has all its own helpers defined instead of re-using the datastore and spicedb test helpers we use for other integration tests?

magefiles/test.go
miparnisari: `goTest(ctx, "./...", "-tags", "image")`

runs the image tests AND the unit tests, which takes forever 😓  if the unit tests were fast this wouldn't be an issue, but alas

pkg/cache/memory_limiter_integration_test.go
miparnisari: updated

ecordell (approved):

PR #3253, opened 2026-07-30
fix: ReadRels doing useless work after hitting the page size

tstirrat15:
## Description
A user reported an issue where they received `optional_limit` values from `ReadRelationships` but then the query hung open until it timed out. I found a place in the readrels logic where when we were iterating over the paginated datastore values, we were always fetching at least one additional rel, which in the worst case could produce the hang the user saw. This fixes that particular failure mode.

## Changes
* Move iteration break check to the bottom of the fetch loop to prevent overfetching from the datastore iterator
## Testing
Review.

### DIFF
--- a/internal/services/v1/relationships_test.go
+++ b/internal/services/v1/relationships_test.go
@@ -8,6 +8,7 @@ import (
 	"maps"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -28,6 +29,7 @@ import (
 	"github.com/authzed/spicedb/internal/testserver"
 	"github.com/authzed/spicedb/pkg/datalayer"
 	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/datastore/options"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/testutil"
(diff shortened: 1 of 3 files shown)


internal/services/v1/relationships_test.go
```diff
@@ -1370,6 +1372,76 @@ func TestReadRelationshipsBeyondAllowedLimit(t *testing.T) {
...
+	queryCount := &atomic.Uint64{}
+	conn, _, _ := testserver.NewTestServerWithConfig(t, 0, memdb.DisableGC, true,
+		testserver.DefaultTestServerConfig,
+		func(t testing.TB, ds datastore.Datastore) (datastore.Datastore, datastore.Revision) {
+			ds, rev := tf.StandardDatastoreWithData(t, ds)
+			return queryCountingDatastore{ds, queryCount}, rev
+		})
+	client := v1.NewPermissionsServiceClient(conn)
+
+	// Reset the count so that only the queries made by ReadRelationships are counted.
```
miparnisari: i'm confused by this comment. by this point of execution would any reads have happened during server setup or something?

miparnisari (dismissed):
LGTM but please change the title of the PR to "fix: ReadRels doing useless work after hitting the page size" or something like that, and add it to changelog :)

internal/services/v1/relationships_test.go
tstirrat15: Potentially during WriteSchema?

tstirrat15: It wasn't actually necessary. Good catch.

miparnisari (approved):

PR #3264, opened 2026-08-05
chore: downgrade cel-go

tstirrat15:
## Description
v0.30.0 of cel-go indirectly [introduced a performance regression](https://github.com/cel-expr/cel-go/issues/1403) because of `protovalidate` misusing its API. This downgrades cel-go until we can get a version of `protovalidate` in that doesn't exercise this bad behavior.

## Changes
* Downgrade cel-go
## Testing
Review.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -192,7 +192,7 @@ require (
 	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/go-viper/mapstructure/v2 v2.5.0 // indirect
 	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8 // indirect
-	github.com/google/cel-go v0.30.0 // indirect
+	github.com/google/cel-go v0.29.2 // indirect
 	github.com/google/pprof v0.0.0-20260115054156-294ebfa9ad83 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -192,7 +192,7 @@ require (
 	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/go-viper/mapstructure/v2 v2.5.0 // indirect
 	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8 // indirect
-	github.com/google/cel-go v0.30.0 // indirect
+	github.com/google/cel-go v0.29.2 // indirect
```
miparnisari, line 195: i didn't know that you could do this? i thought you had to do `replace github.com/google/cel-go => github.com/google/cel-go v0.29.2` to ensure it doesn't get overwritten by accident 🤔

tstirrat15: Hmm... yeah, that would actually pin it, where this is just requesting it. I don't have a strong opinion here - this isn't that much different from an application that hasn't been updated yet. It theoretically runs the risk of accidental overwrite, but I'm not expecting us to be in this state very long. Famous last words?

miparnisari (approved):

PR #3266, opened 2026-08-06
perf: reduce memory usage in schema compilation with caveats

miparnisari:
Before

image

After

image

### DIFF
--- a/pkg/caveats/env.go
+++ b/pkg/caveats/env.go
@@ -2,6 +2,7 @@ package caveats
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/authzed/cel-go/cel"
 
@@ -13,6 +14,9 @@ import (
 type Environment struct {
 	ts        *types.TypeSet
 	variables map[string]types.VariableType
+
+	celEnvMu sync.Mutex
+	celEnv   *cel.Env // GUARDED_BY(celEnvMu)
 }
 
 // NewEnvironmentWithDefaultTypeSet creates and returns a new environment for compiling a caveat,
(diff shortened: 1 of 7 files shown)


pkg/schemadsl/compiler/translator.go
```diff
@@ -160,7 +162,6 @@ func translateCaveatDefinition(tctx *translationContext, defNode *dslNode) (*cor
 		return nil, defNode.WithSourceErrorf(definitionName, "caveat `%s` must have at least one parameter defined", definitionName)
 	}
 
-	env := caveats.NewEnvironmentWithTypeSet(tctx.caveatTypeSet)
```
miparnisari, line 163: this line was costing us a lot

tstirrat15 (dismissed):
LGTM

pkg/caveats/env.go
```diff
@@ -77,6 +81,13 @@ func (e *Environment) EncodedParametersTypes() map[string]*core.CaveatTypeRefere
 
 // asCelEnvironment converts the exported Environment into an internal CEL environment.
 func (e *Environment) asCelEnvironment(extraOptions ...cel.EnvOption) (*cel.Env, error) {
+	e.celEnvMu.Lock()
+	defer e.celEnvMu.Unlock()
+
+	if e.celEnv != nil {
```
josephschorr, line 100: This is only safe if the extraOptions are not different between calls - please add a unit test to reproduce that potential issue, and then fix; I don't *think* we change the extra options, but best to have coverage

pkg/schemadsl/compiler/translator.go
```diff
@@ -39,6 +39,8 @@ type translationContext struct {
 	// A mapping of partial name -> partial DSL nodes whose resolution depends on
 	// the resolution of the named partial
 	unresolvedPartials *mapz.MultiMap[string, *dslNode]
+
+	env *caveats.Environment
```
josephschorr: doc comment

pkg/caveats/env.go
miparnisari: Good point. I've removed the `extraOptions` param because it's never used.

josephschorr (dismissed):
LGTM

tstirrat15 (approved):
LGTM

PR #3273, opened 2026-08-13
fix: allow prefixed definition names in permission type annotations

### LINKED ISSUE #3272: Permission type annotations do not parse prefixed definition names
### What happened

With `use typechecking`, a permission type annotation that references a prefixed definition name fails to parse:

```zed
use typechecking

definition example/user {}

definition document {
  relation viewer: example/user

  permission view: example/user = viewer
}
```

```
error: parse error in `schema.zed`, line 8, column 27: Expected one of: [TokenTypeEquals], found: TokenTypeDiv
  5 | definition document {
  6 |   relation viewer: example/user
  7 |
  8 >   permission view: example/user = viewer
    >                           ^
  9 | }
 10 |
```

https://authzed.com/docs/spicedb/concepts/schema#object-type-definitions

ivanauth:
Fixes #3272

Permission type annotations (`use typechecking`) failed to parse prefixed definition names (`permission view: example/user = viewer`) because the annotation parser consumed single identifiers instead of type paths. Relation type references already accept them.

Changes:
- Parser: `consumeTypeAnnotation` uses `consumeTypePath()` at both sites (first type and each `|` continuation).
- Translator: annotations are normalized with the object type prefix like relation type references, since validation compares them against fully-qualified terminal types by exact string match. Skipped when typechecking is off (annotations are stripped then, and rejecting unprefixed ones would break previously-valid schemas).
- `use` flags are registered before partial/import translation, so annotations in partials and imported files are normalized correctly.


tstirrat15 (approved):
This looks good to me.

PR #3275, opened 2026-08-17
chore: embed consistency test files

miparnisari:


pkg/validationfile/loader.go
```diff
@@ -106,21 +131,41 @@ func PopulateFromFilesContents(ctx context.Context, dl datalayer.DataLayer, cave
...
-			}
-			data, err := io.ReadAll(file)
-			if err != nil {
-				return nil, datastore.NoRevision, fmt.Errorf("error when reading schema file %s: %w", schemaPath, err)
+			var schemaPath string
+			var data []byte
+
+			if fsys != nil {
+				// io/fs paths are always slash-separated and rooted at the
+				// filesystem, so fs.ValidPath is the escape check filepath.IsLocal
```
tstirrat15, line 139: Minor nit would be to break this down into a couple of functions so that this function isn't switching on `fsys` but rather calling different logic depending on the context

tstirrat15 (approved):
LGTM

PR #3278, opened 2026-08-18
perf(postgres): add index for resource ID prefix filter

### LINKED ISSUE #3274: Postgres - ReadRelationships with optional_resource_id_prefix filter cannot use index
## Description
`ReadRelationships` with a prefix filter causes a query like this to be generated (captured from the datastore's query builder, with `pg_current_snapshot()` standing in for a specific revision):

```sql
SELECT object_id, relation, userset_namespace, userset_object_id, userset_relation, caveat_name, caveat_context, expiration
FROM relation_tuple
WHERE pg_visible_in_snapshot(created_xid, pg_current_snapshot()) = true
  AND pg_visible_in_snapshot(deleted_xid, pg_current_snapshot()) = false
  AND namespace = '<namespace>'
  AND object_id LIKE '<prefix>\_%'
  AND (expiration IS NULL OR expiration > NOW());
```

Which with `en_US.UTF-8` collation (default in RDS) does a seq scan:

```
Gather  (cost=1000.00..410914.91 rows=21 width=154) (actual time=82.851..1220.622 rows=11 loops=1)
  Workers Planned: 2
  Workers Launched: 2
  Buffers: shared hit=327354 read=32287
  I/O Timings: shared read=2045.173
  ->  Parallel Seq Scan on relation_tuple  (cost=0.00..409912.81 rows=9 width=154) (actual time=446.095..1202.032 rows=4 loops=3)
"        Filter: (((object_id)::text ~~ '<prefix>\_%'::text) AND ((namespace)::text = '<namespace>'::text) AND pg_visible_in_snapshot(created_xid, pg_current_snapshot()) AND (NOT pg_visible_in_snapshot(deleted_xid, pg_current_snapshot())) AND ((expiration IS NULL) OR (expiration > now())))"
        Rows Removed by Filter: 1367386
        Buffers: shared hit=327354 read=32287
        I/O Timings: shared read=2045.173
Planning:
  Buffers: shared hit=1
Planning Time: 0.161 ms
Execution Time: 1220.659 ms
```

I was hoping using the prefix filter would be an optimization for something which is currently making a lot of DB calls/load but without it being able to use the index I suspect it will make it worse. 

That a filter may not be able to use indexes is called out in the filter docs 

>NOTE: The performance of the API will be affected by the selection of fields on which to filter. If a field is not indexed, the performance of the API can be significantly slower.

But this seems like a reasonable case for an index.

## Possible Solution
Adding this index allowed it to do an index scan instead:
```sql
CREATE INDEX CONCURRENTLY ix_relation_tuple_by_resource_id_prefix ON relation_tuple (namespace, object_id varchar_pattern_ops);
```

```
Index Scan using ix_relation_tuple_by_resource_id_prefix on relation_tuple  (cost=0.56..2.79 rows=22 width=154) (actual time=0.035..0.052 rows=11 loops=1)
  Index Cond: (((namespace)::text = '<namespace>'::text) AND ((object_id)::text ~>=~ '<prefix>_'::text) AND ((object_id)::text ~<~ '<prefix>`'::text))
"  Filter: (((object_id)::text ~~ '<prefix>\_%'::text) AND pg_visible_in_snapshot(created_xid, pg_current_snapshot()) AND (NOT pg_visible_in_snapshot(deleted_xid, pg_current_snapshot())) AND ((expiration IS NULL) OR (expiration > now())))"
  Buffers: shared hit=15
Planning:
  Buffers: shared hit=49 read=1
  I/O Timings: shared read=0.613
Planning Time: 0.988 ms
Execution Time: 0.635 ms
```

### Claude's summary
`ReadRelationships` with `optional_resource_id_prefix` compiles to `object_id LIKE 'prefix%'` (`BuildLikePrefixClause`, internal/datastore/common/sql.go). Postgres can only serve a `LIKE 'prefix%'` range scan from a btree under C collation or a pattern opclass[0][1], but `relation_tuple.object_id` is plain `VARCHAR` in the database's default collation and none of the shipped indexes use `varchar_pattern_ops` — so on typical deployments (docker/RDS default locales) every prefix-filtered read degrades to a full table scan. Adding `CREATE INDEX ... ON relation_tuple (namespace, object_id varchar_pattern_ops)` restores an index range scan (~2000× fewer buffers in our 4M-row test). Suggestion: ship such an index in a migration, or document that prefix filters require C collation or a manual pattern-ops index.

### Sources
[0] https://www.postgresql.org/docs/current/indexes-types.html#INDEXES-TYPES-BTREE
[1] https://www.postgresql.org/docs/current/indexes-opclass.html

epbensimpson:
## Description

A prefix filter in `ReadRelationships` cannot currently use any of the existing indexes due to using a `LIKE` in the query, unless the database is configured to use `C` collation.

This PR adds an index with `object_id varchar_pattern_ops` which allows it to be used for `LIKE` queries regardless of collation.

Obviously there is some tradeoff in terms of writes with adding an additional index which needs some consideration. But without the index or appropriate collation the filter is basically unusable on any deployment using postgres with a significant number of tuples.

## Testing

I couldn't find any existing tests specifically around index usage so testing was all manual, and was done by capturing the output from the query builder, replacing the specific revision with `pg_current_snapshot()`, and manually running the following query before and after adding the index manually, on a deployment with ~4.1 million tuples:

```sql
SELECT object_id, relation, userset_namespace, userset_object_id, userset_relation, caveat_name, caveat_context, expiration
FROM relation_tuple
WHERE pg_visible_in_snapshot(created_xid, pg_current_snapshot()) = true
  AND pg_visible_in_snapshot(deleted_xid, pg_current_snapshot()) = false
  AND namespace = '<namespace>'
  AND object_id LIKE '<prefix>\_%'
  AND (expiration IS NULL OR expiration > NOW());
```

### Before

```
Gather  (cost=1000.00..410914.91 rows=21 width=154) (actual time=82.851..1220.622 rows=11 loops=1)
  Workers Planned: 2
  Workers Launched: 2
  Buffers: shared hit=327354 read=32287
  I/O Timings: shared read=2045.173
  ->  Parallel Seq Scan on relation_tuple  (cost=0.00..409912.81 rows=9 width=154) (actual time=446.095..1202.032 rows=4 loops=3)
"        Filter: (((object_id)::text ~~ '<prefix>\_%'::text) AND ((namespace)::text = '<namespace>'::text) AND pg_visible_in_snapshot(created_xid, pg_current_snapshot()) AND (NOT pg_visible_in_snapshot(deleted_xid, pg_current_snapshot())) AND ((expiration IS NULL) OR (expiration > now())))"
        Rows Removed by Filter: 1367386
        Buffers: shared hit=327354 read=32287
        I/O Timings: shared read=2045.173
Planning:
  Buffers: shared hit=1
Planning Time: 0.161 ms
Execution Time: 1220.659 ms
```

### After

```
Index Scan using ix_relation_tuple_by_resource_id_prefix on relation_tuple  (cost=0.56..2.79 rows=22 width=154) (actual time=0.035..0.052 rows=11 loops=1)
  Index Cond: (((namespace)::text = '<namespace>'::text) AND ((object_id)::text ~>=~ '<prefix>_'::text) AND ((object_id)::text ~<~ '<prefix>`'::text))
"  Filter: (((object_id)::text ~~ '<prefix>\_%'::text) AND pg_visible_in_snapshot(created_xid, pg_current_snapshot()) AND (NOT pg_visible_in_snapshot(deleted_xid, pg_current_snapshot())) AND ((expiration IS NULL) OR (expiration > now())))"
  Buffers: shared hit=15
Planning:
  Buffers: shared hit=49 read=1
  I/O Timings: shared read=0.613
Planning Time: 0.988 ms
Execution Time: 0.635 ms
```

## References

Fixes #3274

### DIFF
--- a/internal/datastore/postgres/migrations/zz_migration.0026_add_resource_id_prefix_index.go
+++ b/internal/datastore/postgres/migrations/zz_migration.0026_add_resource_id_prefix_index.go
@@ -1,0 +1,13 @@
+package migrations
+
+import (
+	"github.com/authzed/spicedb/internal/datastore/postgres/schema"
+)
+
+func init() {
+	registerIndexMigration(
+		schema.IndexRelationshipByResourceIDPrefix,
+		"add-resource-id-prefix-index",
+		"populate-schema-tables",
+	)
+}
--- a/internal/datastore/postgres/schema/indexes.go
+++ b/internal/datastore/postgres/schema/indexes.go
@@ -104,6 +104,14 @@ var IndexGCDeadRelationships = common.IndexDefinition{
 	PreferredSortOrder: options.Unsorted,
 }
 
+// IndexRelationshipByResourceIDPrefix is an index that supports ReadRelationships queries
+// which filter on optional_resource_id_prefix.
+var IndexRelationshipByResourceIDPrefix = common.IndexDefinition{
+	Name:               `ix_relation_tuple_by_resource_id_prefix`,
+	ColumnsSQL:         `relation_tuple (namespace, object_id varchar_pattern_ops)`,
+	PreferredSortOrder: options.Unsorted,
+}
+
 var pgIndexes = []common.IndexDefinition{
 	UniqueLivingRelationshipIndex,
 	IndexRelationshipBySubject,
@@ -113,6 +121,7 @@ var pgIndexes = []common.IndexDefinition{
 	IndexExpiringRelationships,
 	IndexSortedRelationTupleTransaction,
 	IndexGCDeadRelationships,
+	IndexRelationshipByResourceIDPrefix,
 }
 
 var NoIndexingHint common.IndexingHint = nil


epbensimpson:

josephschorr:
Adding another index will significantly impact both write performance limits and storage requirements - Can we change the existing forward index to support prefixes somehow?

epbensimpson:
I'm not an expert on postgres so take this with a grain of salt, but it doesn't _seem_ feasible to me. 

Assuming replacing the existing index with a new one is fine (would require a brief exclusive lock on the table which could be a problem under load), using `varchar_pattern_ops` we'd end up with an index that does not match the collation of the underlying column and would have significant read performance implications.

Using `object_id COLLATE "C"` in the existing index (rather, a new index that replaces it) is an option but has a large blast radius. Queries would need to be rewritten to also explicitly specify `COLLATE "C"` and wouldn't be able to use the index if they weren't, which seems like a potential footgun. It would also change result ordering to be byte-ordered which may not be desirable.

epbensimpson:
We've manually deployed this index to our production instance in the meantime, here's some stats about the index creation:

**DB Type:** AWS Aurora Postgres `db.r6g.2xlarge`
**Relations:** ~159,363,120
**Elapsed Time:** 19 m 44 s 185 ms
**Index Sizes:**

```
pk_relation_tuple,56 GB
uq_relation_tuple_living_xid,53 GB
ix_relation_tuple_alive_by_resource_rel_subject_covering,26 GB
ix_relation_tuple_by_subject,7121 MB
ix_watch_index,2760 MB
ix_relation_tuple_by_resource_id_prefix,2758 MB
ix_relation_tuple_by_subject_relation,2176 MB
ix_gc_index,235 MB
ix_relation_tuple_expired,8192 bytes
```

kchenery:
> Adding another index will significantly impact both write performance limits and storage requirements - Can we change the existing forward index to support prefixes somehow?

If this is a big concern, I think you can get a bigger win there by swapping the `created_xid` and `deleted_xid` columns in the PK and then removing the index `uq_relation_tuple_living_xid` as this effectively becomes a duplicate then.

I didnt spot any queries that would be negatively impacted by this change as everytime you reference `created_xid` where an index here is useful, you also reference `deleted_xid`. Cant be 100% certain of that though as I'm not too familiar with the access patterns. Also I know this is far easier said than done.

As you'll see from Ben's data above that secondary unique index is almost half the table volume.

josephschorr:
> If this is a big concern, I think you can get a bigger win there by swapping the `created_xid` and `deleted_xid` columns in the PK and then removing the index `uq_relation_tuple_living_xid` as this effectively becomes a duplicate then.

That's an interesting idea. Would you guys be willing to try it? If not, I can run a test

PR #3286, opened 2026-08-25
perf(crdb): use pgx exec mode for reads

josephschorr:
CockroachDB reads inline the revision into the SQL text — `crdbReader.addFromToQuery` concatenates `AS OF SYSTEM TIME <rev>` into the FROM clause — so every new revision produces statement text that will never be looked up again. pgx keys its statement cache on exact text, so the read pool accumulates prepared statements in CockroachDB without bound.

Writes are different: they run with no `AS OF SYSTEM TIME` inlined, and their text varies only with the number of relationships in a batch. Those statements are genuinely reused, so the write pool keeps pgx's default `cache_statement` mode.

Postgres has defaulted to `exec` since #1994; CRDB was never switched over.

### Numbers

3-node CockroachDB cluster, blended check / bulk-check / write load, 15 minutes per arm. CockroachDB's `sql_mem_sql_session_prepared_current`, summed across nodes:

| | prepared statement memory |
| --- | --- |
| before | 61.75 MB, growing linearly, no plateau |
| after | 1.18 MB, plateaus early |

Check latency is unchanged.

### Also included

- Registers `map[string]any` as `jsonb`. Exec mode infers parameter types from Go types rather than asking the server, and would otherwise reject caveat context as ambiguous.
- Forces binary transfer for the schema chunk queries. `bytea` in text format is hex-encoded, and since every chunk of a schema goes into a single multi-row INSERT, a large schema would exceed CockroachDB's 16MiB pgwire message limit.

Deployments can revert with `?default_query_exec_mode=cache_statement` in the datastore URI.

### DIFF
--- a/internal/datastore/crdb/crdb.go
+++ b/internal/datastore/crdb/crdb.go
@@ -80,6 +80,12 @@ func newCRDBDatastore(ctx context.Context, url string, options ...Option) (datas
 	if err != nil {
 		return nil, common.RedactAndLogSensitiveConnString(ctx, errUnableToInstantiate, err, url)
 	}
+	// Reads only. See the note on the write pool below for why the two differ.
+	pgxcommon.ConfigureDefaultQueryExecMode(readPoolConfig.ConnConfig)
+	readPoolConfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
+		RegisterTypes(conn.TypeMap())
+		return nil
+	}
 
 	writePoolConfig, err := pgxpool.ParseConfig(url)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/datastore/crdb/crdb.go
```diff
@@ -89,6 +95,28 @@ func newCRDBDatastore(ctx context.Context, url string, options ...Option) (datas
...
+	// switched to exec. The two paths generate SQL text with very different lifetimes:
+	//
+	//   Reads inline the revision -- crdbReader.addFromToQuery concatenates
+	//   "AS OF SYSTEM TIME <rev>" into the FROM clause -- so every new revision produces a fresh
+	//   generation of statement text that will never be looked up again. pgx keys its cache on exact
+	//   text, so a read pool accumulates prepared statements without bound: measured at 62-68MB
+	//   across a 3-node cluster and still climbing linearly when the test ended.
+	//
+	//   Writes run with atSpecificRevision == "" and inline nothing. Their text varies only with the
+	//   number of relationships in the batch (squirrel appends one VALUES tuple per row), which is
```
tstirrat15, line 108: squirrel?

tstirrat15 (approved):
Nice find

internal/datastore/crdb/crdb.go
josephschorr: Its the go lib we use for building SQL

internal/datastore/crdb/schema_chunker.go
```diff
@@ -17,6 +17,39 @@ const (
...
+// which in effect halves the largest schema SpiceDB can store on CRDB. Note that
+// shrinking crdbMaxChunkSize does not help, because the limit is hit by the
+// combined message rather than by any individual chunk.
+//
+// DescribeExec is used rather than CacheStatement deliberately. Both restore the
+// binary format, but CacheStatement would leave named prepared statements on the
+// server, which is the very cost the exec-mode default was introduced to avoid;
+// and because the INSERT's SQL text varies with the number of chunks, caching it
+// would churn rather than be reused. DescribeExec costs one extra round trip, on
+// queries that run only when a schema is written or when the schema cache misses.
```
miparnisari, line 50: can we simplify this godoc please?

also, in a different PR, can you add a claude.md file to never write such large godocs?

internal/datastore/crdb/pool/health.go
```diff
@@ -46,6 +46,7 @@ func NewNodeHealthChecker(url string) (*NodeHealthTracker, error) {
 	if err != nil {
 		return nil, err
 	}
+	pgxcommon.ConfigureDefaultQueryExecMode(connConfig)
```
miparnisari, line 49: why does the health tracker need this?

internal/datastore/crdb/migrations/driver.go
```diff
@@ -36,6 +36,7 @@ func NewCRDBDriver(ctx context.Context, url string) (*CRDBDriver, error) {
 	}
 	pgxcommon.ConfigurePGXLogger(connConfig)
 	pgxcommon.ConfigureOTELTracer(connConfig, false)
+	pgxcommon.ConfigureDefaultQueryExecMode(connConfig)
```
miparnisari, line 39: why does the migration driver need this?

internal/datastore/crdb/schema_chunker.go
josephschorr: I actually asked it to be explicitly descriptive for this one comment

internal/datastore/crdb/pool/health.go
josephschorr: It doesn't *need* it, but the goal was to have all reads except those explicitly carved out using the exec mode - I suppose we could undo it for health and migrations, but I wanted to minimize the number of prepared queries we used

internal/datastore/crdb/migrations/driver.go
josephschorr: See above

miparnisari (approved):

PR #3292, opened 2026-09-02
fix: prevent namespace cache exhaustion and fix hitrate metrics

### LINKED ISSUE #3290: Namespace cache Breaking with Otter Cache implementation
### Summary
Since upgrading to `v1.56.0`, we are experiencing severe namespace cache hit rate degradation over time, which significantly increases request latency.

### Observed Behavior
- **Initial state:** Cache hit rate starts strong at **98%+**.
- **Degraded state:** Hit rate drops significantly to **~10%** after running for a while.
- **Cache scaling observation:**
  - **32 MB (default):** Hit rate degrades after ~2 hours.
  - **128 MB:** Hit rate degrades after ~24 hours.
  - **256 MB:** Currently in production as a temporary mitigation, accompanied by daily rolling restarts.

### Environment & Workload Profile
- **Namespace Count:** 25 total namespaces.
  - ~15 namespaces are >200 KB (database size).
  - A few namespaces are ~500 KB (database size).
- **Traffic Profile:**
  - `CheckPermission`: ~400 req/sec
  - `WriteRelationships`: ~0.7 req/sec

### Isolation & Troubleshooting Steps
- We experienced no cache degradation prior to the cache engine change (when using Ristretto).
- We downgraded to `v1.53.0` and explicitly switched between cache implementations via CLI flags; the degradation occurred consistently when using the new cache engine implementation.

### Additional Context / Logs
Image

tstirrat15:
Fixes #3290 
Written with LLM assistance

## Description

In #3112 we changed to using otter as the cache implementation. It has a different set of behaviors around admission, retention, and eviction of values. A combination of that difference in behavior and us failing to align the namespace cache TTL with revision window meant that cache entries slowly accumulated until Sets began to fail, reducing cache hit rate.

We were also miscounting cache hit rates because of how we were reading entries before setting, which effectively double-counted certain events.

This adds tests to ensure that namespace cache entries get properly evicted over time, aligns the TTL of the cache with the revision window (namespace entry cache keys include the revision), and puts some safeguards around a long TTL getting set and the behavior recurring.

## Changes
See annotations.

## Testing
Review. See that tests pass.

### DIFF
--- a/internal/datastore/proxy/schemacaching/hitrate_test.go
+++ b/internal/datastore/proxy/schemacaching/hitrate_test.go
@@ -1,0 +1,207 @@
+package schemacaching
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/ccoveille/go-safecast/v2"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+
+	"github.com/authzed/spicedb/internal/datastore/proxy/proxy_test"
+	"github.com/authzed/spicedb/internal/datastore/revisions"
+	"github.com/authzed/spicedb/pkg/cache"
+	"github.com/authzed/spicedb/pkg/datastore"
+	ns "github.com/authzed/spicedb/pkg/namespace"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// countingReader is a datastore.Reader that serves a fixed schema and counts
+// how many times a definition was actually loaded from "the datastore" without
+// hitting the cache.
+type countingReader struct {
+	proxy_test.MockReader
+
+	defs  map[string]*core.NamespaceDefinition
+	reads atomic.Int64
+}
+
+func (cr *countingReader) LegacyReadNamespaceByName(_ context.Context, name string) (*core.NamespaceDefinition, datastore.Revision, error) {
+	cr.reads.Add(1)
+	def, ok := cr.defs[name]
+	if !ok {
+		return nil, datastore.NoRevision, datastore.NewNamespaceNotFoundErr(name)
+	}
+	return def, revisions.NewForTransactionID(1), nil
+}
+
+// countingDatastore hands out the same countingReader for every revision. Only
+// SnapshotReader and Close are exercised, so the embedded interface is left nil.
+type countingDatastore struct {
+	datastore.Datastore
+
+	reader *countingReader
+}
+
+func (cd *countingDatastore) SnapshotReader(datastore.Revision) datastore.Reader {
+	return cd.reader
+}
+
+func (cd *countingDatastore) Close() error { return nil }
+
+// schemaForHitRateTest builds a schema of the given size, with definitions
+// shaped roughly like real ones so the size estimator produces realistic costs.
+func schemaForHitRateTest(numDefs int) map[string]*core.NamespaceDefinition {
+	defs := make(map[string]*core.NamespaceDefinition, numDefs)
+	for i := range numDefs {
+		name := fmt.Sprintf("resource_definition_number_%d", i)
+		defs[name] = ns.Namespace(
+			name,
+			ns.MustRelation("owner", nil, ns.AllowedRelation("user", "...")),
+			ns.MustRelation("editor", nil, ns.AllowedRelation("user", "..."), ns.AllowedRelation("group", "member")),
+			ns.MustRelation("viewer", nil, ns.AllowedRelation("user", "..."), ns.AllowedRelation("group", "member")),
+			ns.MustRelation("view", ns.Union(
+				ns.ComputedUserset("viewer"),
+				ns.ComputedUserset("editor"),
+				ns.ComputedUserset("owner"),
+			)),
+		)
+	}
+	return defs
+}
+
+// TestNamespaceCacheHitsUnderStandardUsage exercises the namespace cache the
+// way a running SpiceDB exercises it: a stable schema is read over and over,
+// while the revision at which it is read rotates as the quantization window
+// advances.
+//
+// Because cache keys are (definition name, revision) pairs, every rotation
+// produces a whole new set of keys and strands the previous set: no request
+// will ever ask for the old revision again. The working set, though, is only
+// ever one revision's worth of definitions, which fits with room to spare.
+//
+// Under those conditions the cache must keep serving the current revision's
+// definitions from memory, so each (definition, revision) pair should reach the
+// datastore exactly once no matter how many revisions have rolled by. This test
+// runs long enough for the cache to fill several times over, because that is
+// when the failure appears: a cache that lets stranded entries accumulate ends
+// up permanently at capacity, and an eviction policy that prefers incumbents
+// then starts turning away the definitions that are actually in use. Hit rate
+// falls and the extra reads land on the datastore.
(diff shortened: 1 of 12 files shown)


pkg/cache/cache_otter.go
```diff
@@ -179,13 +179,7 @@ func (wtc *otterCache[K, V]) Set(key K, value V, cost int64) bool {
 	weight := entryWeight(keyStr, uintCost)
 
 	wtc.metrics.costAdded.Add(uint64(weight))
-	if _, ok := wtc.cache.GetIfPresent(keyStr); ok {
-		wtc.cache.Invalidate(keyStr)
-	}
```
tstirrat15, line 184: Querying and invalidating isn't necessary because `Set` does this work anyway. This was a part of the issue - otter has its own logic around where a key that is `Set` gets put in terms of eviction queues, and doing this manual get/invalidate messed with its internal statistics and also skewed the reported hit/miss statistics.

pkg/cache/cache_otter.go
```diff
@@ -179,13 +179,7 @@ func (wtc *otterCache[K, V]) Set(key K, value V, cost int64) bool {
...
 	wtc.cache.Set(keyStr, valueAndCost[V]{value, weight})
-	if wtc.ttl > 0 {
-		wtc.cache.SetExpiresAfter(keyStr, wtc.ttl)
-	}
```
tstirrat15, line 188: This is already covered by the `ExpiryCalculator`.

pkg/cache/cache.go
```diff
@@ -59,11 +59,6 @@ type Cache[K KeyString, V any] interface {
 	// otherwise it returns the zero value of V and false.
 	Get(key K) (V, bool)
 
-	// GetTTL returns the TTL configured for entries in this cache.
-	// If zero, entries never expire. The current Otter-backed
-	// implementation refreshes the TTL on access.
-	GetTTL() time.Duration
```
tstirrat15, line 65: This isn't actually invoked anywhere; dropping it from the interface should be safe.

pkg/cache/cache_otter.go
```diff
@@ -92,15 +92,13 @@ func newOtterCache[K KeyString, V any](name string, config *Config) (*otterCache
 		name:    name,
 		cache:   cache,
 		metrics: otterMetrics{atomic.Uint64{}, counter},
-		ttl:     config.DefaultTTL,
```
tstirrat15, line 95: This was only used by `GetTTL` after removing the `SetExpiry` call in `Set` so we can remove it.

pkg/cmd/server/server.go
```diff
@@ -227,7 +227,15 @@ func (c *Config) complete(ctx context.Context) (*completedServerConfig, error) {
...
 
-	nscc, err := CompleteCache[cache.StringKey, schemacaching.CacheEntry](c.OTel.PrometheusRegistry, &c.NamespaceCacheConfig)
+	nscc, err := CompleteCache[cache.StringKey, schemacaching.CacheEntry](c.OTel.PrometheusRegistry,
+		// the namespace cache has revisions in its keys, so it needs
+		// this configuration to ensure that those entries expire along
+		// with the revisions.
+		c.NamespaceCacheConfig.WithRevisionParameters(
+			c.DatastoreConfig.RevisionQuantization,
+			c.DatastoreConfig.FollowerReadDelay,
+			c.DatastoreConfig.MaxRevisionStalenessPercent,
```
tstirrat15, line 237: This is the crux of the fix - without this, the cache was set up without a real TTL and entries accumulated until they started getting rejected.

tstirrat15 (commented):
See comments

internal/datastore/proxy/schemacaching/hitrate_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/authzed/spicedb/internal/datastore/proxy/proxy_test"
+	"github.com/authzed/spicedb/internal/datastore/revisions"
+	"github.com/authzed/spicedb/pkg/cache"
+	"github.com/authzed/spicedb/pkg/datastore"
+	ns "github.com/authzed/spicedb/pkg/namespace"
+	core "github.com/authzed/spicedb/pkg/proto/core/v1"
+)
+
+// countingReader is a datastore.Reader that serves a fixed schema and counts
+// how many times a definition was actually loaded from "the datastore".
```
josephschorr: loaded without hitting the cache

internal/datastore/proxy/schemacaching/standardcache.go
```diff
@@ -24,14 +25,36 @@ type definitionCachingProxy struct {
...
+		// by tests; otherwise a cache is constructed by the server startup logic and
+		// passed to this function.
+		built, err := cache.NewStandardCache[cache.StringKey, *cacheEntry](&cache.Config{
+			MaxCost:    fallbackMaxCost,
+			DefaultTTL: fallbackTTL,
 		})
+		// The error is only possible for a MaxCost that cannot be represented as
+		// a uint64, which a constant budget cannot hit; fall back to no caching
+		// rather than to a nil cache that would panic on first use.
+		if err != nil {
```
josephschorr: It might be better to simply panic here and change the function to MustNewDefinition...

pkg/cmd/server/server.go
miparnisari: how did Ristretto do this?

pkg/cmd/server/cacheconfig_test.go
```diff
@@ -132,3 +137,42 @@ func TestWithRevisionParameters(t *testing.T) {
...
+		WithGRPCServer(util.GRPCServerConfig{Network: util.BufferedNetwork, Enabled: true}),
+		// The cluster dispatch cache is only built when the dispatch server is on.
+		WithDispatchServer(util.GRPCServerConfig{Network: util.BufferedNetwork, Enabled: true}),
+		WithNamespaceCacheConfig(CacheConfig{Name: "namespace", Enabled: true, MaxCost: "32MiB"}),
+		WithDispatchCacheConfig(CacheConfig{Name: "dispatch", Enabled: true, MaxCost: "1MiB"}),
+		WithClusterDispatchCacheConfig(CacheConfig{Name: "cluster_dispatch", Enabled: true, MaxCost: "1MiB"}),
+		WithEnableMemoryProtectionMiddleware(false),
+	)
+
+	completed, err := c.complete(t.Context())
```
miparnisari, line 165: is `c.complete` modifying `c` in place? 

i'm wondering why we need to call this code at all. You're never using `completed` after this setup

pkg/cmd/server/cacheconfig_test.go
```diff
@@ -60,11 +63,13 @@ func TestWithRevisionParameters(t *testing.T) {
 		expectedTTL          time.Duration
 	}{
 		{
+			// A zero TTL means "never expire", which is not a safe answer for a
+			// revision-keyed cache, so the fallback is used instead.
```
miparnisari: for a revision-keyed cache or for any cache?

miparnisari (dismissed):

pkg/cmd/server/cacheconfig_test.go
tstirrat15: Yeah, it's modifying things in place - applying defaults and doing validation and such.

miparnisari: so then why do we need `completed`?

tstirrat15: See the `t.Cleanup` right below this

miparnisari (approved):

PR #3295, opened 2026-09-04
fix: when spicedb shuts down, set health to NOT_SERVING

miparnisari:
⚠️ Pre-requisite: https://github.com/authzed/consistent/pull/4 ⚠️

## What this fixes

On SIGINT or SIGTERM, SpiceDB closed its gRPC listeners immediately while its health service still reported `SERVING`. Load balancers and Kubernetes readiness probes kept routing to the instance, so requests failed with `Unavailable` (connection refused, "error reading server preface: EOF") until the endpoint was
removed. Meanwhile the surviving replicas kept dispatching to the dead peer, turning most of their own requests into `DeadlineExceeded` or `Unavailable`.

With this PR, shutdown first flips the health service to `NOT_SERVING`, keeps every listener open for 2 seconds so routers can notice, and only then drains. The compose stack and Envoy config are updated to match: Envoy polls the gRPC health service, and dispatch resolves a shared DNS alias so the hashring holds every node.

## How it was verified

Chaos run against the docker-compose stack (CockroachDB, two SpiceDB nodes, Envoy), once on `main` and once on this branch, with the same seeded sequence of 100 kills. Each iteration: one `zed permission check` as a readiness gate (always `true`), thumper at 100 qps (LookupResources returning 100 results, plus checks and writes), then `docker kill` of a random target with a random signal, then a 10 second observation window. Response codes were read from thumper's gRPC client metrics.

## Final comparison (success % in 10s window after kill; errors in parentheses)

The rows that matter for the PR   are the four spicedb-1/spicedb-2 rows, where main is ~7% and not-serving is 100%. 

The other two are killing Envoy, and SpiceDB can't salvage that.

| experiment | n | main | not-serving |
|---|---|---|---|
| spicedb-1 SIGTERM | 11 | 6.9% (DeadlineExceeded=6952, Unavailable=3020) | 100.0% (-) |
| envoy SIGINT | 16 | 0.2% (Unavailable=16075) | 0.4% (Unavailable=16065) |
| spicedb-2 SIGINT | 20 | 7.6% (DeadlineExceeded=12907, Unavailable=5166) | 100.0% (-) |
| spicedb-1 SIGINT | 17 | 7.0% (DeadlineExceeded=11105, Unavailable=4367) | 100.0% (-) |
| spicedb-2 SIGTERM | 21 | 8.5% (DeadlineExceeded=12108, Unavailable=6705) | 100.0% (-) |
| envoy SIGTERM | 15 | 0.3% (Unavailable=15058) | 0.3% (Unavailable=15051) |
| ALL  | 100 | 5.3% (Unavailable=50391, DeadlineExceeded=43072) | 69.1% (Unavailable=31116) |


pkg/cmd/server/server.go
```diff
@@ -610,6 +615,23 @@ func (c *Config) complete(ctx context.Context) (*completedServerConfig, error) {
...
+		drainDelay = 0
+	}
+	closeables.AddWithoutError(func() {
+		healthManager.Close()
+		if drainDelay == 0 {
+			return
+		}
+		log.Info().Dur("drain-delay", drainDelay).Msg("health set to NOT_SERVING, waiting before closing listeners")
+		time.Sleep(drainDelay)
+	})
```
tstirrat15, line 633: Only Q here: is there something deterministic that we can use that's not a sleep? It'd be nice if we didn't have to wait longer than we need to, or else make sure that we wait long enough that we don't trigger error conditions.

I don't understand enough of the context to know what signal we'd use, though.

tstirrat15 (dismissed):
LGTM, see question

pkg/cmd/server/server.go
miparnisari: >  is there something deterministic that we can use that's not a sleep

i don't think so? maybe a cancellable `ctx`?

tstirrat15 (approved):
LGTM

PR #3296, opened 2026-09-06
fix(tls): reload certificate authorities when they rotate on disk

### LINKED ISSUE #1448: Support monitoring CA rotation
We are currently deploying our SpiceDB cluster using the K8s operator and have it setup with a CA/TLS certificate for our clients to use when accessing it. Our CA's currently rotate every 90 days per our internal Ops standards. When they do, our internal services that talk to SpiceDB will pick up the change and reinitialize their client; however, SpiceDB does not currently reinitialize and will hold the value of the last, expired, certificate.

Without directly monitoring, this issue is only made apparent with an outage as our calls will start failing and erroring due to expired certificates, though the clients are using the new valid one.

Can SpiceDB support monitoring the CA cert as it does with the TLS certs? This would allow us to keep the rotation automated without worrying about kicking over pods ahead of each expiry or worrying about downtime during the rotation.

Thanks!

oleksii-udovychenko:
Closes #1448, where @miparnisari invited a PR.

## Problem

SpiceDB reads its CA bundle once at startup and pins it for the life of the process. Once peers present leaves signed by a new authority, every handshake fails with `x509: certificate signed by unknown authority`. In a dispatch cluster each pod pins its own copy at its own boot, so they fail together and the only recovery is a fleet restart.

## Background

Two things get called "certificate rotation" and only one of them works today.

### Leaf rotation, which already works

A leaf re-issued before it expires, signed by the same authority. [cert-manager](https://cert-manager.io/docs/) does this unattended at two thirds of the certificate's lifetime ([renewal docs](https://cert-manager.io/docs/usage/certificate/#renewal)), so on a 90 day certificate it replaces every pod's keypair roughly every 60 days.

Both TLS servers survive it because neither stores the serving certificate on the `tls.Config`. The gRPC server [builds an `x509util.CertWatcher`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L136) over the cert and key paths and sets [`GetCertificate: watcher.GetCertificate`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L141) rather than `Certificates`; the HTTP server [does the same](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L294-L300). `crypto/tls` calls [`GetCertificate`](https://pkg.go.dev/crypto/tls#Config) on every inbound handshake, so the server presents whatever the watcher read most recently. No config is rebuilt and no listener is recreated.

The watcher keeps that current two ways: an [fsnotify](https://github.com/fsnotify/fsnotify) watch on both paths, and a poll ticker as a backstop, [`defaultWatchInterval`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/x509util/certwatcher.go#L46) at 10s. [`completedGRPCServer.Run`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L210) [starts it](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L213) and [waits on `Started()`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L215) before serving, so a watcher that failed to start is a startup error. [`TestCertRotation`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/internal/services/integrationtesting/certtest/cert_test.go#L66) covers it.

### Authority rotation, which does not

Rotating the authority changes the bytes of the bundle on disk. It cannot be one swap, because trust for the new authority has to be in place on every peer before any leaf chains to it, or the peers that have not caught up reject the new leaves. So it runs as a staged overlap, for a bundle mounted from a Secret:

1. **The new authority is added to the bundle.** cert-manager or an operator writes `old + new` into the Secret. No pod's filesystem has changed yet.

2. **kubelet projects the new revision onto disk,** about a minute later. It refreshes mounted volumes on its [periodic pod sync](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) rather than when the Secret changes, and [`syncFrequency`](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) defaults to 1m. It never rewrites `ca.crt`: [AtomicWriter](https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/volume/util/atomic_writer.go) writes the whole revision into a fresh timestamped directory, symlinks `..data_tmp` at it, then renames that over `..data`. The rename is the atomic step. The visible file stays a symlink through `..data`, so the path keeps resolving, and each revision points it at a different inode, now holding `old + new`. kubelet reaps the retired revision directory afterwards.

   ```
   ca.crt  -> ..data/ca.crt
   ..data  -> ..2026_09_06_10_05_00.222222222/
   ```

3. **Every peer re-reads the bundle from disk.** A process that read the file once at boot still holds `old` alone, and waiting does not change that. This is the only step that needs anything from SpiceDB.

4. **Leaves are re-issued under the new authority.** Peers that completed step 3 verify them; peers that skipped it fail every handshake from here on. Nothing looks wrong before this point.

5. **The old authority is dropped.** It comes out of the Secret and kubelet publishes another revision the same way. `ca.crt` then resolves to a third inode holding `new` alone.

## Current state

A handshake reads two fields of `tls.Config` differently, so a long-lived config freezes one of them and not the other:

```
handshake -> GetCertificate()      -> callback, run per handshake -> returns what CertWatcher last read   (fresh)
handshake -> RootCAs / ClientCAs   -> *x509.CertPool, read as-is  -> pool built at container start        (frozen)
```

Four sites build a pool from an operator-supplied path and hold the result, two through [`x509util.CustomCertPool`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/x509util/x509util.go#L14) and two through `grpcutil.WithCustomCerts`:

| Site | Reached by | Builds the pool via | Today |
|---|---|---|---|
| [`clientCreds()`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L159) | `ClientCAPath`, programmatic only | [`x509util.CustomCertPool`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L167) | pinned at startup |
| [`telemetry/reporter.go`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/internal/telemetry/reporter.go#L146) | [`--telemetry-ca-override-path`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/serve.go#L226) | `x509util.CustomCertPool` | pinned at startup |
| [`dispatch/combined/combined.go`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/internal/dispatch/combined/combined.go#L252) | [`--dispatch-upstream-ca-path`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/serve.go#L159) | `grpcutil.WithCustomCerts(VerifyCA, ...)` | pinned at startup |
| [`gateway/gateway.go`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/internal/gateway/gateway.go#L58) | [`--http-upstream-override-tls-cert-path`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/serve.go#L89) | `grpcutil.WithCustomCerts(SkipVerifyCA, ...)` | pool built but never used, `SkipVerifyCA` sets `InsecureSkipVerify` |

## Proposed solution

The frozen row becomes a callback, so it resolves per handshake the way `GetCertificate` already does:

```
handshake -> GetCertificate()      -> callback, run per handshake -> returns what CertWatcher last read      (fresh)
handshake -> CAPool.CertPool()     -> callback, run per handshake -> re-reads the path, at most once per 10s (fresh)
```

Per site:

| Site | Today | With this change |
|---|---|---|
| `clientCreds()` | `x509util.CustomCertPool`, pinned at startup | `NewCAPool` behind `NewReloadingTLSCreds`, so `ClientCAs` is resolved per handshake |
| `telemetry/reporter.go` | `x509util.CustomCertPool`, pinned at startup | `NewCAPool` behind `DialTLSContext`, so `RootCAs` is resolved per connection |
| `dispatch/combined/combined.go` | `grpcutil.WithCustomCerts(VerifyCA, ...)`, pinned at startup | as `clientCreds()`, trading `grpcutil.WithCustomCerts` for `NewReloadingTLSCreds` |
| `gateway/gateway.go` | `grpcutil.WithCustomCerts(SkipVerifyCA, ...)`, pool built but never used | unchanged, because that pool is dead code behind `InsecureSkipVerify`; making it reload would change behaviour beyond this fix |

`x509util.CAPool` owns the CA path instead of a pool built from it once. Behind one mutex it holds the path, the bytes last read from it, the authorities parsed out of those bytes, and the time of that read. Four entry points, in the order a site uses them:

- **`NewCAPool(caPath)`** reads the path eagerly, so a missing or malformed bundle fails at construction rather than at the first handshake. `WithReloadInterval` sets the minimum gap between reads, default 10s. There is no background goroutine and no `Start` step to forget.

- **`CertPool()`** does the reloading, in this order:
  - inside the interval, return the cached pool without touching the filesystem
  - otherwise re-read the path, which may be a single PEM file or a directory of them
  - if the bytes are identical to the last read, return the same pool, so a reconcile loop rewriting the file unchanged costs nothing
  - if the read or the parse fails, log a warning and keep the authorities already loaded
  - otherwise swap in the new pool and log at info

- **`NewReloadingTLSCreds(ca, base)`** for the gRPC sites. It resolves the trust store per handshake, then delegates the handshake itself to `credentials.NewTLS`.

- **`CAPool.DialTLSContext(base)`** for the `http.Transport` site. `http.Transport` reads `TLSClientConfig` once per client but calls the dial hook once per connection, so the telemetry reporter goes through this adapter.

Both adapters supply the pool and nothing else: neither sets `InsecureSkipVerify` and neither builds a chain of its own.

## Compatibility

This PR replaces `CustomCertPool` with `NewCAPool` instead of keeping both, so there is one way to read a CA path and it reloads. `pkg/x509util` is outside `internal/`, so that changes a public package. I converted both in-tree callers, and a code search found no external importers.

Would you rather I keep `CustomCertPool` as a one-shot wrapper, so the exported API stays as it is?

## Design decisions

- **Why on-demand rather than a watcher.** [`combined.NewDispatcher`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/internal/dispatch/combined/combined.go#L193), the site that most needs this, has no context to scope a goroutine to and nowhere to report a watcher that failed to start. `Dispatcher` does have [`Close`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/internal/dispatch/dispatch.go#L34), so a goroutine could be scoped to that, but then every test that builds a dispatcher has to close it or `goleak` fails. A watcher API also fails silently when nobody starts it: [`NewTLSCertWatcher`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/x509util/certwatcher.go#L77) reads eagerly, and I ran one unstarted, which kept returning the original bytes indefinitely after the files rotated underneath it, with no error and no log line. The serving path avoids that because it starts the watcher and blocks on `Started()`. `NewDispatcher` can do neither. `CAPool` has no start step.

- **Why not fsnotify, given that it works here.** A rotation does wake an inotify watch on the visible `ca.crt`, because reaping the retired directory changes the watched inode's link count, which Linux reports as `CHMOD` rather than `REMOVE` ([documented here](https://github.com/fsnotify/fsnotify#watching-a-file-doesnt-work-well)), and [`CertWatcher.handleEvent`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/x509util/certwatcher.go#L259) already re-adds the watch and re-reads on that. I drove a real `CertWatcher` against a real atomic-writer layout with its poll interval set to 10 minutes, so only fsnotify could explain a pickup, and it caught the rotation in 50ms over four rotations. Detection works; the lifecycle above is what rules it out. `CertWatcher` also keeps its 10s poll ticker alongside the watcher, so even there events only speed up a poll that still runs.

- **Why not extend `CertWatcher` to cover CA bundles.** The types do not line up. It is a keypair watcher: `certPath` plus `keyPath`, `tls.X509KeyPair`, `GetCertificate(*tls.ClientHelloInfo)`. A CA bundle has no private key, needs `AppendCertsFromPEM` into an `*x509.CertPool` for `RootCAs`/`ClientCAs`, and its path may be a directory of PEMs. A key-less mode leaves half the struct nil per branch, makes `RegisterCallback(func(tls.Certificate))` untypeable for a pool, and breaks [`TestCertWatcherNew`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/x509util/certwatcher_test.go#L40), which asserts that empty paths must error. One "watched path" primitive underneath both would be cleaner. It means refactoring the serving-certificate path, which is more than a bug fix should touch.

- **Why a new type rather than a fix inside `grpcutil`.** The dispatch site already goes through `grpcutil.WithCustomCerts`, so patching that one helper looks like the smaller diff. [`WithCustomCertBytes`](https://github.com/authzed/grpcutil/blob/2ea1e3d2d98b/dial_opts.go#L109) parses the bundle and passes the finished pool to `credentials.NewTLS`, so the pool is pinned inside credentials this repo does not own and there is nothing left for a caller to re-resolve. It would also miss the telemetry reporter, an `http.Client` that never touches `grpcutil`.

- **Why not `k8s.io/apiserver`'s [`dynamiccertificates`](https://pkg.go.dev/k8s.io/apiserver/pkg/server/dynamiccertificates).** [`DynamicFileCAContent`](https://github.com/kubernetes/kubernetes/blob/v1.34.0/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L52) reloads correctly, but it returns [`VerifyOptions()`](https://github.com/kubernetes/kubernetes/blob/v1.34.0/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L258) for the apiserver's client-cert authenticator rather than a `tls.Config` hook, so both adapters here would still have to exist. The dependency cost is high: SpiceDB has one direct `k8s.io` dependency today, and that package needs `apimachinery`, `client-go` and `klog`, out of a module that also carries etcd, cel-go and the rest of the apiserver stack. It has the same lifecycle trap, since [`Run(ctx, workers)`](https://github.com/kubernetes/kubernetes/blob/v1.34.0/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L157) drives fsnotify through a workqueue while the constructor loads eagerly, so an unstarted one is frozen, and `RunOnce` avoids that but re-reads on every call with no interval. And it is one `os.ReadFile` on a single filename, where a CA path here may be a directory of PEMs.

- **Why 10s.** On-demand costs one `os.ReadFile` on the first connection after the interval lapses, and up to 10s of staleness, the same bound `CertWatcher` already accepts with its 10s poll. Neither term dominates: kubelet has already taken up to a minute to write the bytes to disk, and an established dispatch connection does not re-verify at all, so a new pool takes effect when the connection is replaced, at `--dispatch-cluster-max-conn-age`, [30s by default](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/cmd/util/util.go#L64).

- **Why keep stale authorities on a failed read.** During an update the bundle can be momentarily absent or half written. Failing closed there turns a successful rotation into an outage.

## Testing

- **`capool_test.go` covers the reload rules:** `CertPool()` picks up a rotated authority, the retired one stops verifying, nothing is re-read inside the interval, and a missing or truncated bundle keeps the previous pool.

- **`TestCAPoolFollowsKubernetesSecretRotation` builds the kubelet layout above,** timestamped directory and `..data` rename included.

- **Every converted site has a test that rotates certificates on disk against a real handshake.** Each one fails if I revert only that site's fix:

  - dispatch, `TestDispatchFollowsCARotation`: never recovers, `x509: certificate signed by unknown authority`
  - telemetry, `TestRemoteReporterFollowsCARotation`, HTTPS sink on loopback: reports never resume
  - `clientCreds()`, `TestClientCredsFollowCARotation`, real gRPC server: `credentials never picked up the rotated CA`

- **The dispatch test runs in a real container** under `mage test:image`. The node dispatches to itself over TLS, so one container covers both ends of the handshake, with the certificate and the bundle mounted the way an operator would mount them. It breaks dispatch first, by publishing a leaf from an authority that is not in the bundle, then publishes that authority and waits for recovery. Without that first step the recovery check would pass even with the fix reverted. It takes 13 to 62s, depending on how the probes line up with the reload interval, and runs on memdb, so there is no datastore container.

- **`ClientCAPath` has no flag bound to it,** so that site can only be driven in-process.

## Future work

- **An event-driven fast path.** The fsnotify result above makes this cheap to add on top of the interval, the way `CertWatcher` already combines both. Do you want it in this PR or a follow-up?

- **Metrics.** `CAPool` has no lifecycle to scope a collector registration to, so reloads and failures only log. Counters would need a `Register`/`Close` pair.

- **A pre-existing bug I did not fix.** Pointing a CA path at a mounted Secret *directory* rather than at the file inside it does not work at all: [`dirContents`](https://github.com/authzed/spicedb/blob/9b5cdc1e849249ecd31cfeca82149767dbb8760f/pkg/x509util/x509util.go#L44) walks the mount root, treats the `..data` symlink as a regular file, and fails with `read ..data: is a directory`. That surfaces at config time, so the process fails to start rather than degrading. It predates this PR and applies equally to the old `CustomCertPool`, so I left it alone. Should it go in here?

- **Smaller things.** There is no `--grpc-client-ca-path` flag, which is why the `clientCreds()` site can only be tested in-process, and adding one is more than a fix needs. The three new test files each define their own small CA helper, the way `certtest` and `certwatcher_test` already do. And `changelog-enforcer` stays red until someone applies `Skip-Changelog`. Should I pull the CA helper out into one place, and add a changelog entry?


oleksii-udovychenko:

PR #3298, opened 2026-09-08
fix(crdb): expire unused overlay keys with 24h row-level TTL

### LINKED ISSUE #2507: CockroachDB overlay key table needs GC
Right now, no rows are ever removed from the CRDB driver's `transactions` table, which is used for overlay key tracking. We should probably set a default expiration of, say, 24h on the rows in that table

siddiqueirshad:
Fixes #2507

## Description

The CockroachDB driver's `transactions` table is used for overlay-key tracking so related writes overlap and the new-enemy problem cannot occur. Rows were inserted/upserted on write (`ON CONFLICT DO UPDATE SET timestamp = now()`) and never removed.

That is unbounded growth: a small, stable key set for `static`/`prefix` overlap, and an unbounded set for the `request` overlap strategy, where clients can introduce arbitrary keys.

Overlay keys only need to exist long enough to collide with concurrent (or recently completed) writes. This migration enables CockroachDB row-level TTL so a key expires 24 hours after last touch. Because `ttl_expire_after` refreshes `crdb_internal_expiration` on UPDATE, an active key is kept alive by the existing upsert.

The ALTER TABLE is version-gated the same way as the existing transaction-metadata and relationship-expiration migrations:

- CRDB v22.1: `ttl_expire_after` only
- CRDB v22.2–v23: also `ttl_job_cron = '@hourly'`
- CRDB v24+: also `ttl_disable_changefeed_replication` so operator changefeeds do not see TTL deletes (Watch does not subscribe to this table)

## Testing

- Unit tests for migration head/chain and version-specific SQL
- Datastore integration test (`TestTransactionsTableTTL`) against CRDB v26.2.5: `SHOW CREATE TABLE transactions` includes `ttl_expire_after`, and a static-overlap write still upserts overlay keys

Made with [Cursor](https://cursor.com)


siddiqueirshad:
Hi @vroldanbet, when you get a chance, could you please take a look at this PR? Thanks!

Issue #3300, opened 2026-09-08
Improve handling of `CANCELLED` errors in the dispatch layer


tstirrat15:
`CANCELLED` at the server level should mean "the client cancelled the request and went away." Currently, with the way that our dispatching layer works, a dispatching gRPC client can return with a `CANCELLED` error at its layer and then that bubbles up to the caller, which obscures what's going on at the server level.

A couple things here:

1. Wrap the dispatcher remote boundary to catch `CANCELLED` errors and turn them into `UNAVAILABLE`s if the caller's context isn't already cancelled. This happens in `internal/dispatch/remote/cluster.go` at lines 206, 780, 731, and 744. Make sure the context check is done on the dispatcher's inbound context rather than the hedging context.
2. Make hedging cancels more visible with `context.WithCancelCause` in `cluster.go` at lines 552 and 557. We use this pattern at `internal/graph/cursors.go:451`, and we work around the lack of this pattern at `internal/graph/lookupsubjects.go:354`.

breken-ai:
Fix up matching the spec - PR: https://github.com/authzed/spicedb/pull/3301

Someone filed this with an exact fix spec; we implemented it at the four named sites (DispatchCheck, DispatchExpand, both hedged-streaming returns), switched hedged contexts to context.WithCancelCause, and added unit + bufconn integration coverage. Full internal/dispatch/remote suite passes with -race. The non-hedged streaming setup paths are called out in the PR as deliberate follow-up scope.

Built by breken, your AI support engineer - breken.ai finds the bug, proves the fix with unmatched RCA, answers customers in your voice, and keeps your code and support docs clean - this one's on us.

PR #3301, opened 2026-09-09
fix(dispatch): reshape internal CANCELED to UNAVAILABLE at the remote boundary

breken-ai:
Implements #3300.

- Adds reshapeDispatchError, which converts CANCELED from the remote dispatch boundary into UNAVAILABLE when the dispatcher's inbound context is not itself canceled. Server-level CANCELED once again means only "the client went away"; internal cancels (hedging races, losing dispatches) no longer masquerade as client cancels to the caller. The context check is done on the inbound context, not the hedging/timeout context.
- Applied at the four remote-boundary returns named in the issue: DispatchCheck, DispatchExpand, and the two hedged-streaming returns in dispatchStreamingRequest.
- Hedged contexts now use context.WithCancelCause, with the winning dispatcher canceling the others with an explicit cause ("canceled because dispatcher %q was the first to return results"), matching the pattern in internal/graph/cursors.go.

Tests: unit table for reshapeDispatchError plus bufconn integration tests for both the sync (Check) and hedged streaming (LookupSubjects) paths - a dispatcher returning CANCELED against a live caller context now surfaces UNAVAILABLE. Full internal/dispatch/remote suite passes (`go test -race -count=1 ./internal/dispatch/remote/...` - ok, 3.99s). TestDispatchTimeout's accepted-error assertion was widened to allow Unavailable, since reshaping now applies on that path.

Follow-up, deliberately out of scope here: the non-hedged streaming setup paths (the direct handler() error returns) can still leak CANCELED and could get the same treatment.

> Built by breken, your AI support engineer - breken.ai - this one's on us.


breken-ai:

PR #2157, opened 2024-12-07
Remove dead-code

IvanCherepov:
Addresses the most of dead-code from https://github.com/authzed/spicedb/issues/1719

➜  spicedb git:(main) deadcode -test ./... | grep -v pkg
```
internal/datasets/subjectsetbytype.go:93:28: unreachable func: SubjectByTypeSet.Len
internal/datastore/crdb/options.go:329:6: unreachable func: DebugAnalyzeBeforeStatistics
internal/datastore/mysql/options.go:218:6: unreachable func: DebugAnalyzeBeforeStatistics
internal/datastore/postgres/log_tracer.go:13:25: unreachable func: tracingLogger.Log
internal/datastore/postgres/options.go:341:6: unreachable func: DebugAnalyzeBeforeStatistics
internal/datastore/postgres/options.go:348:6: unreachable func: WithQueryInterceptor
internal/datastore/postgres/common/pgx.go:166:6: unreachable func: ConnectWithInstrumentation
internal/developmentmembership/trackingsubjectset.go:59:32: unreachable func: TrackingSubjectSet.RemoveFrom
internal/dispatch/combined/combined.go:129:6: unreachable func: RemoteDispatchTimeout
internal/graph/errors.go:59:6: unreachable func: NewAlwaysFailErr
internal/graph/errors.go:73:34: unreachable func: RelationNotFoundError.NamespaceName
internal/graph/errors.go:78:34: unreachable func: RelationNotFoundError.NotFoundRelationName
internal/graph/errors.go:82:34: unreachable func: RelationNotFoundError.MarshalZerologObject
internal/graph/errors.go:87:34: unreachable func: RelationNotFoundError.DetailsMetadata
internal/graph/errors.go:95:6: unreachable func: NewRelationNotFoundErr
internal/graph/errors.go:136:6: unreachable func: NewRelationMissingTypeInfoErr
internal/graph/errors.go:179:6: unreachable func: NewUnimplementedErr
internal/graph/graph.go:30:24: unreachable func: ExpandResult.ResultError
internal/graph/graph.go:38:6: unreachable func: AlwaysFailExpand
internal/logging/logger.go:39:6: unreachable func: Log
internal/logging/logger.go:43:6: unreachable func: Print
internal/logging/logger.go:47:6: unreachable func: Printf
internal/middleware/dispatcher/dispatcher.go:28:6: unreachable func: FromContext
internal/middleware/dispatcher/dispatcher.go:38:6: unreachable func: MustFromContext
[...]

Issue #1719, opened 2024-01-23
Remove statically analyzable dead-code


jzelinskie:
### What platforms are affected?

linux, macos, windows, others

### What architectures are affected?

amd64, arm64, others

### What SpiceDB version are you using?

main

### Steps to Reproduce

1. `go install golang.org/x/tools/cmd/deadcode@latest`
2. `deadcode -test ./...`

### Expected Result

No output

### Actual Result

```
internal/datasets/subjectset.go:39:22: unreachable func: SubjectSet.UnionWithSet
internal/datasets/subjectsetbytype.go:86:28: unreachable func: SubjectByTypeSet.Len
internal/datastore/common/jitter.go:12:6: unreachable func: WithJitter
internal/datastore/mysql/options.go:204:6: unreachable func: DebugAnalyzeBeforeStatistics
internal/datastore/postgres/options.go:315:6: unreachable func: DebugAnalyzeBeforeStatistics
internal/datastore/postgres/options.go:322:6: unreachable func: WithQueryInterceptor
internal/developmentmembership/trackingsubjectset.go:59:32: unreachable func: TrackingSubjectSet.RemoveFrom
internal/dispatch/combined/combined.go:119:6: unreachable func: RemoteDispatchTimeout
internal/graph/errors.go:59:6: unreachable func: NewAlwaysFailErr
internal/graph/errors.go:73:32: unreachable func: ErrRelationNotFound.NamespaceName
internal/graph/errors.go:78:32: unreachable func: ErrRelationNotFound.NotFoundRelationName
internal/graph/errors.go:82:32: unreachable func: ErrRelationNotFound.MarshalZerologObject
internal/graph/errors.go:87:32: unreachable func: ErrRelationNotFound.DetailsMetadata
internal/graph/errors.go:95:6: unreachable func: NewRelationNotFoundErr
internal/graph/errors.go:136:6: unreachable func: NewRelationMissingTypeInfoErr
internal/graph/errors.go:166:6: unreachable func: NewUnimplementedErr
internal/graph/graph.go:51:6: unreachable func: AlwaysFailExpand
internal/logging/logger.go:39:6: unreachable func: Log
internal/logging/logger.go:43:6: unreachable func: Print
internal/logging/logger.go:47:6: unreachable func: Printf
internal/middleware/dispatcher/dispatcher.go:28:6: unreachable func: FromContext
internal/middleware/dispatcher/dispatcher.go:38:6: unreachable func: MustFromContext
[...]

PR #3304, opened 2026-09-09
fix: apply dispatch-upstream-timeout to outbound dispatches

### LINKED ISSUE #3303: dispatch-upstream-timeout is never applied; outbound dispatches always use the 60s default
### Summary

`--dispatch-upstream-timeout` (`SPICEDB_DISPATCH_UPSTREAM_TIMEOUT`) has no effect. The configured value never reaches the outbound dispatch client, which falls back to the hardcoded 60s default. Setting the flag to any value is silently a no-op.

Affects v1.56.0, v1.56.1, and `main`.

### Details

The flag is parsed correctly into `Config.DispatchUpstreamTimeout`, but there are two option setters named `RemoteDispatchTimeout` and neither delivers it to the code that applies the wire deadline.

**1. The combined (outbound) dispatcher never receives it.**

`combineddispatch.NewDispatcher` at [`pkg/cmd/server/server.go#L345-L369`](https://github.com/authzed/spicedb/blob/v1.56.0/pkg/cmd/server/server.go#L345-L369) is passed 14 options; `combineddispatch.RemoteDispatchTimeout` is not among them. That option has no non-test caller anywhere in the tree:

```
$ grep -rn "combineddispatch.RemoteDispatchTimeout\|combined\.RemoteDispatchTimeout" --include='*.go' .
(only internal/dispatch/combined/combined_test.go:149)
```

So `opts.remoteDispatchTimeout` keeps its zero value, is passed as `DispatchOverallTimeout: 0` at [`combined.go#L308-L311`](https://github.com/authzed/spicedb/blob/v1.56.0/internal/dispatch/combined/combined.go#L308-L311), and is replaced by the 60s fallback at [`remote/cluster.go#L127-L130`](https://github.com/authzed/spicedb/blob/v1.56.0/internal/dispatch/remote/cluster.go#L127-L130):

```go
dispatchOverallTimeout := config.DispatchOverallTimeout
if dispatchOverallTimeout <= 0 {
	dispatchOverallTimeout = 60 * time.Second
}
```

That value is then stamped onto every outbound dispatch at [`remote/cluster.go#L280`](https://github.com/authzed/spicedb/blob/v1.56.0/internal/dispatch/remote/cluster.go#L280).

**2. The one call site that does pass it targets a field that is never read.**

[`server.go#L400`](https://github.com/authzed/spicedb/blob/v1.56.0/pkg/cmd/server/server.go#L400) calls `clusterdispatch.RemoteDispatchTimeout(c.DispatchUpstreamTimeout)`. That assigns `optionState.remoteDispatchTimeout` at [`internal/dispatch/cluster/cluster.go#L71-L76`](https://github.com/authzed/spicedb/blob/v1.56.0/internal/dispatch/cluster/cluster.go#L71-L76), but the field (declared at `:25`) has no reader — `cluster.NewClusterDispatcher` builds only `graph.NewDispatcher` and `caching.NewCachingDispatcher`, no remote client, so it has nothing to apply a wire timeout to. This call site appears to be the reason the omission has gone unnoticed: it looks like the flag is wired.

### Why the deadline is observable as 60s rather than `min(caller, 60s)`

`resenje.org/singleflight` v0.4.3 runs the leader's function under `context.WithoutCancel`, so the caller's remaining deadline is stripped before `remote/cluster.go:280` applies its own `WithTimeout`. Singleflighted unary `DispatchCheck` / `DispatchExpand` / plan-check therefore advertise ~60s on the wire regardless of what the caller had left. Streaming and the singleflight bypass paths (empty `TraversalBloom`, loop detection, `Debug != NO_DEBUG`) correctly carry `min(caller, 60s)`.

### Impact

Deployments that set a low `--dispatch-upstream-timeout` to bound peer failures do not get it. When one peer stops responding but its TCP connection stays open — a frozen node rather than a crash — dispatches to it hold a 60s budget instead of the configured one. Callers with shorter deadlines give up first, but the peer-side work and the singleflight leader persist, so a single unresponsive node degrades checks fleet-wide for far longer than the configuration implies.

### Suggested fix

Add the missing option:

```diff
 dispatcher, err = combineddispatch.NewDispatcher(
 	combineddispatch.UpstreamAddr(c.DispatchUpstreamAddr),
 	combineddispatch.UpstreamCAPath(c.DispatchUpstreamCAPath),
+	combineddispatch.RemoteDispatchTimeout(c.DispatchUpstreamTimeout),
 	combineddispatch.SecondaryUpstreamAddrs(c.DispatchSecondaryUpstreamAddrs),
```

and separately either remove `cluster.RemoteDispatchTimeout` or document it as inert.

### Prior art

[#1719](https://github.com/authzed/spicedb/issues/1719) lists `internal/dispatch/combined/combined.go:119:6: unreachable func: RemoteDispatchTimeout` in its dead-code inventory, and [#2157](https://github.com/authzed/spicedb/pull/2157) proposed deleting the option — closed unmerged. Worth noting the right resolution is wiring it rather than removing it.

One reason the signal went quiet: `combined_test.go:149` (added between v1.53.0 and v1.54.0) now calls `RemoteDispatchTimeout`, which makes the symbol statically reachable, so `deadcode` no longer flags it while production wiring is still absent.

### Environment

- SpiceDB v1.56.0 (also verified present in v1.56.1 and `main`)
- Kubernetes, 8 replicas, `--dispatch-cluster-enabled`, `--dispatch-upstream-addr=kubernetes:///...`

bryantvolk:
Fixes #3303.

`--dispatch-upstream-timeout` / `SPICEDB_DISPATCH_UPSTREAM_TIMEOUT` is parsed into `Config.DispatchUpstreamTimeout` but never reaches the dispatcher that applies the wire deadline, so setting it has no effect. Outbound dispatches always use the hardcoded 60s fallback.

`combineddispatch.NewDispatcher` is passed 14 options and `RemoteDispatchTimeout` is not among them, so `opts.remoteDispatchTimeout` keeps its zero value, reaches `remote.NewClusterDispatcher` as `DispatchOverallTimeout: 0`, and is replaced by `60 * time.Second` at `internal/dispatch/remote/cluster.go:127-130`.

`combineddispatch.RemoteDispatchTimeout` currently has no non-test caller. The one call site that looks like it wires the flag — `clusterdispatch.RemoteDispatchTimeout(c.DispatchUpstreamTimeout)` at `server.go:408` — assigns `optionState.remoteDispatchTimeout` in `internal/dispatch/cluster`, a field that has no reader: `cluster.NewClusterDispatcher` builds only `graph` and `caching` dispatchers, so it has no remote client to apply a timeout to. That is likely why the omission went unnoticed. This PR does not touch that call site; whether to remove it or document it as inert seems worth a maintainer's call.

Worth noting for triage: #1719 already lists `internal/dispatch/combined/combined.go:119:6: unreachable func: RemoteDispatchTimeout` in its dead-code inventory, and #2157 proposed deleting the option (closed unmerged). The symbol stopped being flagged once `combined_test.go:149` began calling it, which made it statically reachable while production wiring remained absent.

## Impact

Deployments that lower `--dispatch-upstream-timeout` to bound peer failures do not get it. When a peer stops responding while its TCP connection stays open — a frozen node rather than a crash — dispatches to it hold a 60s budget. The caller's own deadline does not clip this on singleflighted paths, because `resenje.org/singleflight` runs the leader under `context.WithoutCancel`, so the ~60s is what actually goes out as `grpc-timeout`.

We hit this on v1.56.0 with `dispatchUpstreamTimeout: 1s` configured: a node froze for ~78s and roughly 75% of the resulting check failures were on healthy replicas blocked on sub-dispatches to the frozen one.

## Testing

`go test ./internal/dispatch/... ./pkg/cmd/...` passes locally (go1.27.1, darwin/arm64); `gofmt` and `go vet` clean on the changed package. `TestNewDispatcher_WithUpstream` in `internal/dispatch/combined/combined_test.go` already covers the option itself.

I did not add a regression test, since asserting the wire deadline reaches the remote dispatcher would mean exporting or otherwise reaching into `clusterDispatcher.dispatchOverallTimeout`. Happy to add one in whatever shape you'd prefer.


bryantvolk:

miparnisari (changes requested):
LGTM. Can you add an entry to the CHANGELOG.md? Include an explanation in terms of customer-facing value. Thanks!

bryantvolk:
@miparnisari added the CHANGELOG entry, plz review when you have a sec

PR #3309, opened 2026-09-10
perf(postgres): key limited relationship deletes on (tableoid, ctid)

### LINKED ISSUE #3306: Postgres: `DeleteRelationships` with limit using inefficient query
While trying to bulk delete relationships in preparation for a schema migration, we ran into an issue where batches of 1000 were taking significant amounts of time to delete. My colleague noted that the query could probably make use of `ctid` to improve performance and after digging into it with Claude that does seem to be the case.

While the below is Claude's summary, I have validated its assertions manually to the best of my ability and it doesn't seem to have hallucinated anything 🙃 

# Problem

`deleteRelationshipsWithLimit` in `internal/datastore/postgres/readwrite.go` (v1.56.1) soft-deletes in two steps: a CTE selects up to `limit` rows by their seven key columns, then the `UPDATE` re-finds each row by that key.

```sql
WITH found_tuples AS (SELECT namespace, object_id, relation, userset_namespace, userset_object_id, userset_relation, created_xid
                      FROM relation_tuple WHERE <filter> AND deleted_xid = <live> LIMIT 1000)
UPDATE relation_tuple SET deleted_xid = $1
WHERE (namespace, object_id, relation, userset_namespace, userset_object_id, userset_relation, created_xid) IN (SELECT * FROM found_tuples)
```

The select half uses the alive covering index and is fast. For the re-find, Postgres picks `ix_relation_tuple_by_subject`, matches on the five subject-leading columns, and applies `object_id` and `created_xid` as a post-filter:

```
Update on relation_tuple  (cost=117.53..411.59 rows=0 width=0)
  ->  Nested Loop  (cost=117.53..411.59 rows=103 width=209)
        ->  HashAggregate  (cost=116.84..117.84 rows=100 width=366)
              Group Key: (found_tuples.namespace)::text, (found_tuples.object_id)::text, (found_tuples.relation)::text, (found_tuples.userset_namespace)::text, (found_tuples.userset_object_id)::text, (found_tuples.userset_relation)::text, found_tuples.created_xid
              ->  Subquery Scan on found_tuples  (cost=0.70..115.09 rows=100 width=366)
                    ->  Limit  (cost=0.70..114.09 rows=100 width=171)
                          ->  Index Scan using ix_relation_tuple_alive_by_resource_rel_subject_covering on relation_tuple relation_tuple_1  (cost=0.70..56180.78 rows=49544 width=171)
                                Index Cond: (((namespace)::text = '<redacted>'::text) AND ((relation)::text = '<redacted>'::text) AND ((userset_namespace)::text = '<redacted>'::text))
        ->  Index Scan using ix_relation_tuple_by_subject on relation_tuple  (cost=0.70..2.93 rows=1 width=177)
              Index Cond: (((userset_object_id)::text = (found_tuples.userset_object_id)::text) AND ((userset_namespace)::text = (found_tuples.userset_namespace)::text) AND ((userset_relation)::text = (found_tuples.userset_relation)::text) AND ((namespace)::text = (found_tuples.namespace)::text) AND ((relation)::text = (found_tuples.relation)::text))
              Filter: (((found_tuples.object_id)::text = (object_id)::text) AND (found_tuples.created_xid = created_xid))
```

The planner estimates that inner lookup at one row. The real cost is the number of relationships sharing that subject on that relation, because `object_id` is only checked after the walk. Observed on a 157M-row table:

| Subject fan-in | Per-row cost |
| --- | --- |
| 1 (subject unique to the resource) | ~1ms; 1000-row batch in ~1s |
| ~26k | 15 to 20ms; 1000-row batch in 15 to 20s |
| whole relation (wildcard subject `*`) | 1000-row batch exceeds a 60s deadline |

Batch size makes no difference, since the cost is per row. `OptionalLimit` exists to bound the cost of a delete, but today that bound only holds when subjects are not shared.

# Proposed Solution

Keying the `UPDATE` on the physical row address avoids the re-lookup entirely:

```sql
WITH found_tuples AS (SELECT tableoid, ctid FROM relation_tuple WHERE <filter> AND deleted_xid = <live> LIMIT 1000)
UPDATE relation_tuple SET deleted_xid = $1
WHERE (tableoid, ctid) IN (SELECT tableoid, ctid FROM found_tuples)
```

```
Update on relation_tuple  (cost=115.59..229.09 rows=0 width=0)
  ->  Nested Loop  (cost=115.59..229.09 rows=100 width=48)
        ->  HashAggregate  (cost=115.59..116.59 rows=100 width=44)
              Group Key: found_tuples.tableoid, found_tuples.ctid
              ->  Subquery Scan on found_tuples  (cost=0.70..115.09 rows=100 width=44)
                    ->  Limit  (cost=0.70..114.09 rows=100 width=10)
                          ->  Index Scan using ix_relation_tuple_alive_by_resource_rel_subject_covering on relation_tuple relation_tuple_1  (cost=0.70..56180.78 rows=49544 width=10)
                                Index Cond: (((namespace)::text = '<redacted>'::text) AND ((relation)::text = '<redacted>'::text) AND ((userset_namespace)::text = '<redacted>'::text))
        ->  Tid Scan on relation_tuple  (cost=0.00..1.11 rows=1 width=10)
              TID Cond: (found_tuples.ctid = ctid)
              Filter: (found_tuples.tableoid = tableoid)
```

The inner side becomes a `Tid Scan`: one heap fetch per row, independent of fan-in. Verified on Aurora PostgreSQL and on vanilla PostgreSQL 16.4. Semantics are unchanged: the CTE and the `UPDATE` share one statement snapshot, and a row version replaced by a concurrent writer surfaces as the same serialization failure the current `created_xid` match produces, which the datastore already retries.

This is the pattern the GC already uses. `batchDelete` in `internal/datastore/postgres/gc.go` selects `gcPKCols = {tableoid, ctid}` and deletes with `WHERE (tableoid, ctid) IN (SELECT tableoid, ctid FROM rows)`, including `tableoid` for partitioned tables. The change to `deleteRelationshipsWithLimit` is the `selectForDelete` column list and the `IN` list in the CTE format string.

### Considerations

- **Isolation level.** The `ctid` join matches only the row version the CTE saw. Under SERIALIZABLE (the default) and REPEATABLE READ, a concurrent modification of that row raises a serialization failure in both the current and proposed forms, so behaviour is identical. Under READ COMMITTED they would differ: the seven-column key follows the update chain and still soft-deletes the new version, while the `ctid` form skips it for that round. SpiceDB does not run write transactions at READ COMMITTED, so this is theoretical, but it is the one semantic difference. Address reuse is not a concern: vacuum cannot remove a tuple visible to a live snapshot, so a `ctid` returned by the CTE cannot be recycled within the same statement.
- **Partitioned tables.** `ctid` is only unique per physical table, hence the `tableoid` pairing, which mirrors `gcPKCols`. The datastore tests should be run against the partitioned configuration.
- **Plan stability.** The fast plan is a nested loop with a `Tid Scan` inner side. If the planner ever chose a hash semi-join instead, the inner side would become a sequential scan, which is worse than today. At the default `--max-delete-relationships-limit` of 1000 the nested loop wins by a wide margin. For the non-partitioned case, `WHERE ctid = ANY (ARRAY(SELECT ctid FROM …))` plans as a `Tid Scan` with `TID Cond: (ctid = ANY ($0))` and has no join at all, at the cost of dropping `tableoid`.

epbensimpson:
## Description

Fixes #3306

Improves performance of `DeleteRelationships` when using a limit, by using the `ctid` to avoid an expensive crawl of a second index in order to `UPDATE` the identified tuples in some query plans.

Query planners are fun so it took a bit to come up with a benchmark that successfully triggered the bad plan, but it got there in the end. The good news is that tables which never hit the bad plan see no regression from this change - the `Tid Scan` re-find is at worst equivalent to an index lookup on the primary key, and measured slightly faster.

## Benchmarks

Adds `BenchmarkPostgresDeleteWithLimit`, which seeds rows shaped like a large production table and deletes 100 or 1000 rows for three subject fan-in scenarios. On this table `EXPLAIN` shows the previous query re-finding rows through `ix_relation_tuple_by_subject` (`Rows Removed by Filter: 20999` per row in the shared and wildcard scenarios); the new query uses a `Tid Scan`.

Results on my MacBook Pro M5 24gb against Postgres 18 using `-count 6 -benchtime 5x`: 

| Scenario | Limit | Before | After | Change |
| --- | --- | --- | --- | --- |
| Unique subject | 100 | 4.16 ms ± 15% | 3.62 ms ± 20% | −13% (p=0.041) |
| Unique subject | 1000 | 20.8 ms ± 13% | 17.8 ms ± 6% | −15% (p=0.009) |
| Shared subject (fan-in 21k) | 100 | 120.6 ms ± 2% | 3.65 ms ± 26% | −97% (p=0.002) |
| Shared subject (fan-in 21k) | 1000 | 1208.6 ms ± 2% | 18.2 ms ± 6% | −98% (p=0.002) |
| Wildcard subject (fan-in 21k) | 100 | 120.0 ms ± 4% | 3.47 ms ± 17% | −97% (p=0.002) |
| Wildcard subject (fan-in 21k) | 1000 | 1187.0 ms ± 4% | 18.1 ms ± 4% | −98% (p=0.002) |
| geomean | | 110.3 ms | 8.03 ms | −93% |

## Testing

Existing Postgres datastore tests cover the limited delete path and no regressions were found running the tests locally.

### DIFF
--- a/internal/datastore/benchmark/postgres_delete_bench_test.go
+++ b/internal/datastore/benchmark/postgres_delete_bench_test.go
@@ -1,0 +1,207 @@
+package benchmark
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+
+	"github.com/authzed/spicedb/internal/datastore/postgres"
+	"github.com/authzed/spicedb/internal/testfixtures"
+	testdatastore "github.com/authzed/spicedb/internal/testserver/datastore"
+	"github.com/authzed/spicedb/internal/testserver/datastore/config"
+	dsconfig "github.com/authzed/spicedb/pkg/cmd/datastore"
+	"github.com/authzed/spicedb/pkg/datastore"
+	"github.com/authzed/spicedb/pkg/datastore/options"
+	"github.com/authzed/spicedb/pkg/tuple"
+)
+
+// The seeded data mimics the statistical shape of a large production
+// relation_tuple table rather than its absolute size, since the shape is what
+// determines the query plans Postgres chooses:
+//
+//   - A large background population with a modest number of distinct subjects
+//     (production tables commonly have on the order of 10k) spread evenly over
+//     many (namespace, relation) pairs. This gives the columns of relation_tuple
+//     production-like selectivity estimates, and gives the subject-leading
+//     indexes the same size ratio to the primary key as on a large table.
+//   - Small per-scenario namespaces that differ only in subject fan-in: the
+//     number of relationships sharing one subject on one relation. Each scenario
+//     is kept to a small fraction of the table so that its own subjects do not
+//     dominate the column statistics, and the plan stays the same across
+//     scenarios while the fan-in varies.
+const (
+	// Background population: evenly spread subjects, as on a production table.
+	pgDeleteBenchBackgroundRows       = 2_000_000
+	pgDeleteBenchBackgroundNamespaces = 20
+	pgDeleteBenchBackgroundRelations  = 50
+	pgDeleteBenchBackgroundSubjects   = 13_000
+
+	// Per scenario: one namespace with this many relations and rows per relation.
+	// The rows per relation is also the fan-in of the shared and wildcard scenarios,
+	// and must exceed the largest delete limit below.
+	pgDeleteBenchScenarioRelations       = 2
+	pgDeleteBenchScenarioRowsPerRelation = 21_000
+)
+
+var pgDeleteBenchLimits = []uint64{100, 1000}
+
+// errRollbackBenchmarkTx is returned from the ReadWriteTx callback to roll back
+// the deletion, so every iteration operates on the same set of live rows.
+var errRollbackBenchmarkTx = errors.New("rollback benchmark transaction")
+
+// pgDeleteBenchScenario describes a set of relationships living in their own
+// namespace that share a subject fan-in pattern.
+type pgDeleteBenchScenario struct {
+	name      string
+	namespace string
+	// subjectIDSQL is a SQL expression yielding the subject object ID for the
+	// row numbered g in generate_series; r is the row's relation index.
+	subjectIDSQL string
+}
+
+var pgDeleteBenchScenarios = []pgDeleteBenchScenario{
+	{
+		// Every relationship has a subject unique to its resource (fan-in of 1).
+		name:         "UniqueSubject",
+		namespace:    "unique",
+		subjectIDSQL: "'unique-' || g",
+	},
+	{
+		// Every relationship on a relation shares one subject, so each subject has
+		// a fan-in of pgDeleteBenchScenarioRowsPerRelation on that relation.
+		name:         "SharedSubject",
+		namespace:    "shared",
+		subjectIDSQL: "'shared-' || r",
+	},
+	{
+		// Every relationship is granted to the wildcard subject.
+		name:         "WildcardSubject",
+		namespace:    "wild",
+		subjectIDSQL: "'" + tuple.PublicWildcard + "'",
+	},
+}
+
+// BenchmarkPostgresDeleteWithLimit benchmarks DeleteRelationships with a delete
+// limit against Postgres, on a table shaped like a large production deployment,
+// for subjects with differing fan-in. Requires Docker.
+//
+// Run with:
+//
+//	go test ./internal/datastore/benchmark/ -bench BenchmarkPostgresDeleteWithLimit \
+//	    -benchmem -run '^$' -timeout 60m
+func BenchmarkPostgresDeleteWithLimit(b *testing.B) {
+	if testing.Short() {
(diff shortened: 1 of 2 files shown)


epbensimpson:
Was just thinking a couple of things about using direct SQL to do the table seeding in the benchmark:

* It's a bit brittle to inline the SQL, if the DB changes then the benchmark would drift
* Might be worth extracting to a helper function for reuse if it's kept

Not sure how relevant either of these is given it's a benchmark, but happy to either replace it with `WriteRelationships` or do the helper extraction if desired.

josephschorr:
> Was just thinking a couple of things about using direct SQL to do the table seeding in the benchmark:
> 
> * It's a bit brittle to inline the SQL, if the DB changes then the benchmark would drift
> * Might be worth extracting to a helper function for reuse if it's kept

Yeah, if you don't mind :)

epbensimpson:
Added a `seedRelationships` helper for the benchmarks which uses `ReadWriteTransaction.BulkLoad` instead of direct SQL and should be datastore agnostic.

josephschorr:
@epbensimpson can you add an entry to the CHANGELOG.md as well? I also rebased to test on latest - you may need to do that for your branch as well

PR #3314, opened 2026-09-11
chore: make lame duck delay configurable

tstirrat15:
## Description
Follow-on to #3295. The approach of adding a delay was correct, but the length of the delay depends on the configuration of the load balancer in question, which means that the delay should be configurable. This implements that.

See https://sre.google/sre-book/load-balancing-datacenter/#robust_approach_lame_duck for more information.

## Changes
* Add `--grpc-shutdown-drain-delay` as a configurable flag
* Update code and tests accordingly
* Add delay to compose files
## Testing
Review. See that tests pass. Stand up and shut down the compose file and see that it behaves as expected.


pkg/cmd/serve.go
```diff
@@ -74,10 +74,11 @@ func RegisterServeFlags(cmd *cobra.Command, config *server.Config) error {
 	// Flags for the gRPC API server
 	util.RegisterGRPCServerFlags(grpcFlagSet, &config.GRPCServer, "grpc", "gRPC", ":50051", true)
 	grpcFlagSet.StringSliceVar(&config.PresharedSecureKey, PresharedKeyFlag, []string{}, "(required) preshared key(s) that must be provided by clients to authenticate requests")
-	grpcFlagSet.DurationVar(&config.ShutdownGracePeriod, "grpc-shutdown-grace-period", 5*time.Second, "time limit given to the server to shutdown gracefully after it receives SIGINT or SIGTERM. A value of zero means no limit")
 	if err := cobra.MarkFlagRequired(grpcFlagSet, PresharedKeyFlag); err != nil {
 		return fmt.Errorf("failed to mark flag as required: %w", err)
 	}
+	grpcFlagSet.DurationVar(&config.ShutdownGracePeriod, "grpc-shutdown-grace-period", 5*time.Second, "time limit given to the server to shutdown gracefully after it receives SIGINT or SIGTERM. A value of zero means no limit")
```
tstirrat15, line 80: Just moving this to after the `MarkFlagRequired` that interacts with PresharedSecureKey

ecordell: the previous value that worked for us was 15s right?

tstirrat15: See #3295 - the value hasn't changed over these two PRs.

ecordell (approved):

PR #3316, opened 2026-09-14
feat: cursored bulk relationship deletion

josephschorr:
## Summary

Adds cursor-advanced batched relationship deletion, from the datastore layer up through a new CLI command and the `DeleteRelationships` API.

Deleting a large set of relationships from CockroachDB was impractical: an unbounded `DELETE` does not complete on a table with billions of rows, and a repeated `DELETE ... LIMIT n` loop degrades with every batch because each statement rescans the MVCC tombstones its predecessors left behind. The fix advances a primary-key cursor across batches so each statement starts where the previous one stopped, committing every batch separately.

## Datastore layer

- `DeleteRelationships` returns a `DeleteRelationshipsResult` (count, limit-reached, and a resume cursor) instead of `(uint64, bool, error)`.
- New delete options `WithCursoredDelete` / `WithDeleteAfter` opt into the cursored path; the `CursoredDeleteDatastore` capability interface advertises support. Engines that cannot resume a cursored deletion return a typed `CursoredDeleteNotSupportedError` rather than silently ignoring the cursor.
- CockroachDB implements it as a `WITH deleted AS (DELETE ... ORDER BY ... LIMIT n RETURNING ...) SELECT ..., count(*) OVER ()` statement over a forced index, choosing the sort order from an allowlist of indexes whose physical key order it can actually serve. The cursor-column helpers are shared with the existing select-pagination path.

## Bulk helper and CLI

- `datastore.BulkDeleteRelationships` drives the batch loop with one transaction per batch, repeating sweeps until a pass deletes nothing (to catch rows inserted behind the cursor), and falls back to limited-loop deletes on datastores without support.
- New `spicedb datastore delete-relationships` command. Every filter flag is required, with `<any>` as the explicit match-anything token, so a forgotten flag cannot widen a deletion.

## API

`DeleteRelationships` now supports resumable, cursored deletion. With `optional_limit` and `optional_allow_partial_deletions` set, the response returns an `after_result_cursor`; passing it back as `optional_cursor` on the next call resumes after the last relationship removed. Delete cursors are namespaced to the `DeleteRelationships` call (via the per-API request hash), so a cursor from `ReadRelationships` or `LookupResources` is rejected, and vice versa. Datastores without cursored-delete support return `Unimplemented` when a cursor is supplied, and otherwise fall back transparently to the existing non-cursored partial deletion.

## Testing

- Datastore conformance for cursored delete on CockroachDB (cursor advances strictly, nothing skipped or double-deleted, non-matching relationships survive).
- Unit + memdb integration for the API: cursor request-hash namespacing (cross-API rejection), foreign-cursor rejection, limit/partial validation, unsupported-datastore → `Unimplemented`, and preservation of the existing non-cursored partial-deletion behavior.
- End-to-end API integration against real CockroachDB: a batched deletion resumes across calls via `after_result_cursor`, deletes everything, and the resulting cursor is rejected by `ReadRelationships`.
- Datastore-tagged benchmarks measuring the per-batch KV scan-time degradation of the naive loop against the flat cost of the cursored path.

## Dependency note

The proto fields land via `authzed/api` and the Go bindings via `authzed/authzed-go`. This branch pins `authzed-go` to a `main` pseudo-version containing those bindings; it should be re-pinned to a tagged `authzed-go` release before merge.


internal/datastore/crdb/cursoreddelete.go
```diff
@@ -0,0 +1,215 @@
...
+	candidate, err := schema.IndexForFilter(schemaInfo, dsFilter)
+	if err != nil {
+		return "", nil, options.Unsorted, fmt.Errorf("unable to determine index for filter: %w", err)
+	}
+
+	order, index := cursorOrderForIndex(candidate)
+
+	// AVOID_FULL_SCAN is a soft hint: CockroachDB falls back to a full scan
+	// when no other plan exists, so it cannot make the statement fail. It is
+	// applied only on this path, never to ordinary deletes.
```
tstirrat15: If it doesn't actually prevent a full scan, what's it doing?

pkg/cmd/deleterelationships.go
```diff
@@ -0,0 +1,445 @@
...
+	return &cobra.Command{
+		Use:   "delete-relationships",
+		Short: "bulk deletes relationships matching a filter",
+		Long: `Deletes every relationship matching a filter, in committed batches.
+
+Every filter flag is required. Pass ` + anyToken + ` to leave a component
+unconstrained; an omitted flag is an error, never a wildcard, so a forgotten
+flag cannot widen the deletion. Note that ` + anyToken + ` is not "*": "*" is a
+legitimate subject object id (the wildcard subject), so passing it to
+--subject-id matches only wildcard relationships. Pass ` + ellipsisToken + ` to
```
tstirrat15: The note doesn't feel relevant - we're telling people positively what the thing is.

pkg/cmd/deleterelationships.go
```diff
@@ -0,0 +1,445 @@
...
+		return err
+	}
+
+	if flags.batchSize == 0 {
+		return errors.New("--batch-size must be positive")
+	}
+
+	isTTY := term.IsTerminal(int(os.Stdin.Fd()))
+	target := redactedTarget(cfg.Engine, cfg.URI)
+	if err := confirmDeletion(os.Stdin, cmd.OutOrStdout(), target, describeFilter(filter), flags.skipConfirmation, isTTY); err != nil {
```
tstirrat15: TIL about `OutOrStdout` - that'll probably be useful in `zed`

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
...
+- The `DeleteRelationships` API now supports resumable, cursored batch deletion. When `optional_limit` and `optional_allow_partial_deletions` are set, the response returns an `after_result_cursor`; passing it back as `optional_cursor` on the next call resumes the deletion after the last relationship removed, which on CockroachDB avoids rescanning already-deleted relationships. Datastores that cannot resume a cursored deletion return an error when a cursor is supplied; without a cursor they transparently fall back to the existing non-cursored partial deletion. Delete cursors are namespaced to the `DeleteRelationships` call, so a cursor from `ReadRelationships` or `LookupResources` (or vice versa) is rejected.
```
```suggestion
- The `DeleteRelationships` API now supports resumable, cursored batch deletion. When `optional_limit` and `optional_allow_partial_deletions` are set, the response returns an `after_result_cursor`; passing it back as `optional_cursor` on the next call resumes the deletion after the last relationship removed, which on CockroachDB avoids rescanning already-deleted relationships. Datastores that cannot resume a cursored deletion return an error when a cursor is supplied; without a cursor they transparently fall back to the existing non-cursored partial deletion.
```

tstirrat15 (commented):
Minor things

internal/datastore/crdb/cursoreddelete.go
josephschorr: It's the soft counterpart of `NO_FULL_SCAN`: it penalizes full scans so the optimizer prefers the constrained span from the equality/cursor bounds, but won't error the legitimately-unconstrained case (delete-everything, or the first batch before a cursor exists) the way `NO_FULL_SCAN` would. Reworded the comment to say that.

pkg/cmd/deleterelationships.go
josephschorr: Reframed positively — dropped the "`<any>` is not `*`" contrast and just state what `*` is (a literal wildcard-subject id, so `--subject-id *` matches only wildcard relationships).

CHANGELOG.md
josephschorr: Applied — dropped the cursor-namespacing sentence.

tstirrat15 (approved):
LGTM

PR #3319, opened 2026-09-17
feat(datastore): add SortKeyRevision for order-preserving revision keys

vroldanbet:
## What

Introduces a new interface for revisions, offering a sortable representation

## Why

`Revision.String()` cannot serve as a key in an ordered key/value store. For SpiceDB revision implementations, it is variable-width, so one revision's string **can be a byte prefix of another's**, and its byte order may not match its revision order

For example, `HLCRevision` renders as <nanoseconds>.<10 digits>, so "9.0000000000" sorts above "10.0000000000" even though 9 precedes 10. 

`ByteSortable()` was introduced to report that a type of revision is totally ordered and could therefore be rendered as sortable bytes. The implicit contract was that `String()` representation would be safe, but it was not.

## How

`SortKeyRevision` is an optional capability: a revision that implements it renders itself as bytes that sort in revision order and that no other revision's bytes can be a prefix of. It is a separate interface rather than a method on `Revision` so that adding it breaks no existing implementation.

Encodings are fixed-width per type, which is what delivers _prefix-freeness_, since equal-length keys cannot prefix one another:

```
  HLCRevision            8-byte ordered wall time, then 4-byte logical clock
  TimestampRevision      8-byte ordered int64
  TransactionIDRevision  8-byte big-endian uint64
```

Signed values have their top bit flipped so negatives sort below non-negatives. 

`postgresRevision` **deliberately does not implement the interface**: its revisions are transaction snapshots and only partially ordered, so no order-preserving encoding of them exists.


internal/datastore/revisions/hlcrevision.go
```diff
@@ -102,6 +106,18 @@ func (hlc HLCRevision) ByteSortable() bool {
 	return true
 }
 
+// AppendSortKey implements datastore.SortKeyRevision. It writes the wall time and then the logical
+// clock: the same two fields, in the same order, that LessThan and GreaterThan compare. Every key
+// is hlcSortKeyLength bytes, so no key can be a prefix of another.
+//
+// The logical clock goes in as stored, logicalClockOffset included, and is compared as the uint32
+// it is - exactly what LessThan does - so key order and comparator order agree for every value the
+// field can hold.
```
vroldanbet: I think this is explaining the implementation, and if at all, should be in the method, not in the public doc

internal/datastore/postgres/revisions.go
```diff
@@ -361,6 +361,11 @@ func (pr postgresRevision) ByteSortable() bool {
 	return false
 }
 
+// NOTE: postgresRevision deliberately does not implement datastore.SortKeyRevision. These revisions
```
ecordell, line 364: it might be useful to define an (arbitrary but deterministic) total order, just so all revisions are sortable?

ecordell (dismissed):
should we mark `ByteSortable()` as deprecated?

internal/datastore/postgres/revisions.go
vroldanbet: We explored it but ended up abandoning the idea; it's just too much complexity. @mazdakb can tell you about his experience trying to use pg WAL LSN as the total order. It's full of landmines.

vroldanbet:
>should we mark ByteSortable() as deprecated?

I considered it. I think it's now _correct_ to w.r.t I intended with it initially. What I did was reframe the comment: do not use `String` for this. 

But since one still has to assert the new interface, we may as well deprecate `ByteSortable()`. Does it work if I do it in a follow-up?

ostafen (approved):

PR #3325, opened 2026-09-19
fix(testing): remove four sources of test flakiness

vroldanbet:
## What

Fixes four tests that failed for reasons unrelated to what they were testing.

## Why

**Database setup gave up too early.** Before each test the harness creates a database and runs migrations, retrying if the database isn't ready. The retry budget was 5 seconds — but it also had to cover the migrations, which are the slow part. On a busy machine they simply take longer, and the test reported a failure instead of waiting.

**Unit tests were calling home.** Four tests that start a server built their config with defaults, and the default telemetry endpoint is a real URL. Starting the server started the reporter, so a unit test made a live network call to our telemetry endpoint. It fires in roughly 3% of runs and leaves a connection open, which the leak detector then reports.

**A garbage-collection test slept instead of waiting.** It slept 600ms for a 500ms background task, then asserted the task had run. Miss that 100ms margin once on a loaded machine and it reports the task never ran.

**Two tests failed only on developer machines.** The CRDB tests trust whatever certificate file happens to be in your home directory, and the memory-protection test can't start a server because its deliberately tiny memory limit rounds the cache budget down to zero.

## How

- Size the setup budget for the slowest case observed, not the typical one
- Turn telemetry off in those four tests
- Poll for the background task instead of sleeping past it — also faster, since it stops as soon as the work is done
- Pin the certificate setting so the environment can't change it, and disable the caches the memory-protection test doesn't need

## References

- The database timeout reproduced 3/3 and passed 3/3 after the change
- The CRDB fix matters more than it looks: because the suite stops at the first failure, this one test was hiding roughly 70% of the CRDB tests from anyone affected


internal/datastore/crdb/crdb_test.go
```diff
@@ -383,9 +383,20 @@ func newCRDBWithUser(t *testing.T) (adminConn *pgx.Conn, connStrings map[provisi
...
+	// mere existence of that file as a request to verify against it. A developer
+	// who happens to have one (assorted Postgres tooling creates it, and it need
+	// have nothing to do with CockroachDB) gets "x509: certificate signed by
+	// unknown authority" here - and because the suite runs with -failfast, loses
+	// roughly 70% of the CRDB tests behind this single failure. Setting it empty
+	// makes "require" mean require regardless of what is in the home directory.
 	connStrings = map[provisionedUser]string{
-		testuser:     fmt.Sprintf("postgresql://testuser:testpass@%[1]s/testspicedb?sslmode=require", hostAndPort),
-		unprivileged: fmt.Sprintf("postgresql://unprivileged:testpass2@%[1]s/testspicedb?sslmode=require", hostAndPort),
+		testuser:     fmt.Sprintf("postgresql://testuser:testpass@%[1]s/testspicedb?sslmode=require&sslrootcert=", hostAndPort),
```
ecordell, line 398: what a weird config option

ecordell (approved):

internal/datastore/crdb/crdb_test.go
vroldanbet: It fixes a rather obscure phenomenon that happened locally, where I had a `./postgres` with a cert, which caused CRDB tests to fail locally. This is a dev improvement so folks don't hit the same issue I did.

PR #3326, opened 2026-09-19
fix(testing): stop revision tests racing the database clock

vroldanbet:
## What

These tests compared two different readings of the database clock and failed when time moved between them.

## Why

The test asks the database for the current time, then inserts rows at offsets from it — two seconds ago, four seconds ago, and so on.

The code being tested asks the database for the time *again* when it runs, and rounds that reading down to the nearest half-second or second. If the clock crosses one of those rounding points between the two readings, the code picks a different starting row than the test set up, and the expected count no longer matches:

```
Not equal: expected: 2, actual: 3
incorrect number of revisions visible to snapshot
```

How much slack there is depends on where in the current interval the test happens to start — anywhere from zero to a full interval. Against several round-trips to a database container, that's close to a coin flip on a loaded machine.

The query itself is correct. This is purely a test problem.

## How

Start the setup just after a rounding point, so it has a full interval of slack rather than a random fraction of one. Then check whether the clock crossed a rounding point anyway, and if it did, discard that attempt and retry rather than reporting a failure about data that no longer matches. The final attempt always checks its result, so a real regression still fails.

MySQL's copy of the test has the same flaw and gets the same fix.

## References

- A second case in the same table failed under repetition, which is what showed this was a problem with the whole table rather than one bad row
- Verified with 3 repeated runs and the race detector: postgres 27/27, mysql 21/21, zero retries triggered


internal/datastore/mysql/datastore_test.go
```diff
@@ -659,68 +653,166 @@ func QuantizedRevisionTest(t *testing.T, b testdatastore.RunningEngineForTest) {
...
+					tc.testName, attempt, tc.quantization, tc.followerReadDelay)
+			}
+		})
+	}
+}
 
-				sql, args, err := bulkWrite.ToSql()
-				require.NoError(err)
+// quantizedRevisionCase is one row of QuantizedRevisionTest's table.
+type quantizedRevisionCase struct {
```
ecordell, line 680: nit: a little weird to define this below the usage

ecordell (approved):

PR #3328, opened 2026-09-21
perf(testing): run the test suites in parallel, and fix what that uncovered

vroldanbet:
## What

Runs the test suites concurrently instead of one test at a time, and fixes the bugs that doing so uncovered.

Two suites were serial. The consistency suite ran its 73 fixtures one after another; the shared datastore suite ran its hundred-odd subtests one after another, on every engine. In both cases the work is independent — each test builds its own database — so it can simply run at once.

## Why

The datastore jobs are the longest thing in CI and set the floor for how fast any PR can go green. They were also using a fraction of the machine: a sequential test binary cannot spread across cores, so runner size was not a useful lever until now.

Running tests side by side also turned out to expose real bugs that serial execution had been hiding. Those are fixed here rather than separately, because until they are fixed the parallel suites cannot be green, and because a fix that lives in a sibling branch never reaches the branches that are failing.

## How

- Run consistency fixtures, and the shared datastore suite's subtests, in parallel
- Add a per-engine opt-out (`test.RunSubtestsSerially`) for suites that cannot yet share a container
- Update the Spanner emulator to 1.5.58 — the old one hangs when fixtures run concurrently
- Move the datastore jobs to a 16-core runner, now that they can use it

**Two production fixes**, both found by the parallel suites and both described in `CHANGELOG.md`:

- **Postgres connection pools.** Creating a datastore handed the pool the same deadline it used for its own startup checks, then cancelled it on return. Filling the pool continues on a background goroutine, so the cancellation landed mid-connect and those half-broken connections still entered the pool, failing later with `timeout: read tcp ...: i/o timeout` against a healthy database.
- **CockroachDB watch cancellation.** Cancelling a watch returned `WatchCanceledError` if the connection was already open, and a raw pgx connect error if it was not — a race the caller cannot see.

Plus test-only fixes for assertions that were over-specified or racing a clock.

## Results

| job | before | after |
|---|---|---|
| crdb 25.2.0 | 1143s | **758s** |
| crdb 26.2.5 | 900s | **649s** |
| postgres | 108.8s | **58.7s** |
| mysql | 371s | **259s** |
| consistency (postgres) | 304.0s | **165.5s** |

CockroachDB 25.2.0 is the slowest job in the matrix, so that is 6.4 minutes off the critical path.

The larger runner was measured against an otherwise identical 8 vCPU run: 1.57x faster for 2x the per-minute price, so **+$0.43 per run**. Worth paying for the faster feedback.

## Status per engine

CockroachDB, MySQL, memdb, Postgres and Spanner all run their subtests in parallel. The Postgres GC suites stay serial, because those tests take exclusive locks.

## References

Depends on the connection fix in #3330, which addressed the other cause of the same `i/o timeout` failure.


internal/datastore/spanner/spanner_test.go
```diff
@@ -28,7 +28,20 @@ func TestSpannerDatastore(t *testing.T) {
...
+	// all read it - so a second emulator cannot coexist with the one every other
+	// subtest is using. It also has to be set with t.Setenv, which panics in a
+	// test that has called t.Parallel. The emulator already running here hands
+	// out unmigrated databases just as well, so use it.
+	test.RegisterMigrationTestConfig(Engine, func(t *testing.T) string {
+		return b.NewDatabase(t)
+	})
+
 	// Transaction tests are excluded because, for reasons unknown, one cannot read its own write in one transaction in the Spanner emulator.
+	//
```
ecordell: the tests below run without `test.RunSubtestsSerially()` which is maybe fine but calling out since the PR desc said spanner was opted out too

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Fixed
+- Postgres: fixed connections being handed to queries already broken, which surfaced as `timeout: read tcp ...: i/o timeout` against a perfectly healthy database, most often under load and shortly after startup. Creating a datastore gave the connection pool the same context it used to bound its own startup checks, and then cancelled it on return - but filling the pool happens on a background goroutine that keeps using that context, so the cancellation landed part-way through opening connections. pgx responds to a cancelled context by setting an immediate deadline on the connection it is working on, and those connections still entered the pool. Pool warm-up now runs on a context that keeps the startup timeout but is not cancelled when startup finishes. (https://github.com/authzed/spicedb/pull/3333)
```
ecordell: > keeps the startup timeout

doesn't the code above strip the timeout entirely?

pkg/datastore/test/datastore.go
```diff
@@ -203,16 +203,68 @@ func WithCategories(cats ...string) Categories {
...
+// RunSubtestsSerially makes an engine run the suite's subtests one at a time.
+// It is the opt-out for an engine whose tests cannot share their container
+// concurrently.
+func RunSubtestsSerially() SuiteOption {
+	return func(o *suiteOptions) { o.runner = serial }
+}
+
+// RunSubtestsInParallel makes an engine run the suite's subtests concurrently.
+// It is only needed to override a serial default.
+func RunSubtestsInParallel() SuiteOption {
```
ecordell: nit: this is the default so it's never used

internal/datastore/spanner/spanner_test.go
vroldanbet: yeah the PR body wasn't updated accordingly. I started incrementally chipping away datastores that could be enabled in parallel, and eventually fixes where identified to get all datastores running in parallel again

CHANGELOG.md
vroldanbet: >Pool warm-up now runs on a context that keeps the startup timeout but is not cancelled when startup finishes. (https://github.com/authzed/spicedb/pull/3333)

It's referring to the follow-up PR that introduces the warm-up functionality, which didn't remove it either because pgx really recommends not passing context cancellation there anyway. I'll remove tha last paragraph.

internal/datastore/spanner/spanner_test.go
vroldanbet: ah no, this refers to

```
if !except.Transaction() {
    t.Run("TestWriteAndReadInRWT", runner(tester, WriteAndReadInRWT))
}
```

This is still excluded for Spanner

ecordell (approved):

PR #3329, opened 2026-09-21
fix(runtime): correct the memory fallback from 256KiB to 256MiB

vroldanbet:
## What

The fallback memory figure was 256 KiB, not the 256 MB its comment claims — out by a factor of 1024.

## Why

```go
// 256mb is tiny, but it should comfortably fit in most runtimes.
const fallbackMemoryLimit = 256 * 1024
```

This is the figure used when available memory can't be determined at all, which the surrounding code calls out as a real situation: an AWS ECS task whose memory limit never reaches the cgroup. Cache budgets are taken as a percentage of it, so a cache asking for 70% of available memory got about 183 KiB instead of about 180 MiB.

Nothing reports this. There's no error path — the budget is simply small, the process runs on with almost no cache, and it looks like a performance problem rather than a misconfiguration.

## How

Correct the constant, and add a test that pins the value.

The five existing tests all missed this because each only asserts the fallback is greater than zero, which 256 KiB satisfies perfectly well. The new test also adds a guard that doesn't depend on getting the units right: 1% of the fallback must still exceed a mebibyte.

## References

- Introduced in #3201, with the comment already saying 256mb — so the intent was clear from the start and the value was wrong from the start
- Both new assertions were confirmed to fail against the old value before the fix was applied


squat (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 - Datastore: the wait between transaction retries is now capped at 10 seconds. It previously doubled without limit (25ms, 50ms, 100ms, …), which made `--datastore-max-tx-retries` an exponential wall-clock budget rather than a count of attempts: at the default of 10 the waits total about 25 seconds, but at 15 they total 14 minutes and at 20 over seven hours. (https://github.com/authzed/spicedb/pull/3353)
+- Caches: when the amount of available memory could not be determined - for example an AWS ECS task whose memory limit never reaches the cgroup - the fallback used for percent-based budgets was 256KiB rather than the intended 256MiB. A cache configured for 70% of available memory was sized at roughly 183KiB instead of roughly 180MiB. There was no error in this case; the caches were simply almost useless. (https://github.com/authzed/spicedb/pull/3329)
```
```suggestion
- Caches: when the amount of available memory could not be determined - for example an AWS ECS task whose memory limit never reaches the cgroup - the fallback used for percent-based budgets was 256KiB rather than the intended 256MiB. A cache configured for 70% of available memory was sized at roughly 183KiB instead of roughly 180MiB. (https://github.com/authzed/spicedb/pull/3329)
```

tstirrat15 (dismissed):
See suggestion

tstirrat15 (approved):

squat (approved):

PR #3330, opened 2026-09-21
fix(testing): stop cancelled requests breaking shared database connections

vroldanbet:
## What

The index-checking test wrapper threw away the datastore it was asked to wrap, discarding a safeguard that keeps abandoned requests from breaking shared database connections.

## Why

Tests were failing with errors like:

```
unable to read namespace config:
timeout: read tcp 127.0.0.1:60636->127.0.0.1:32769: i/o timeout
```

The wrapper searched down through the datastore for the SQL layer and then talked to that directly, dropping everything layered on top. One of those layers stops a cancelled request from killing a query that is already running — which matters because the database driver reacts to a cancelled request by immediately expiring the connection it was using, and connections are shared between requests.

Without that protection, the dispatcher abandoning work it no longer needs — normal behaviour, roughly 2,600 times in a single run — kept expiring connections other requests were about to use. The failure surfaced on whichever query picked up the connection next, which is why it looked like a network fault and never pointed at the real cause.

This is test-only. Production keeps the safeguard, because only tests use this wrapper.

## How

Talk to the datastore as given, preserving everything wrapped around it, and pull out only the one capability index checking actually needs.

## References

Verified on the postgres consistency suite with the machine deliberately loaded, twelve fixtures at once, six runs:

| | failures | connections expired | runtime |
|---|---|---|---|
| before | 1 in 4 | ~2,600 per run | 335–366s |
| after | 0 in 6 | 0 | 295–336s |

Runs are also ~10% faster without the churn.


internal/datastore/proxy/indexcheck/indexcheck.go
```diff
@@ -17,25 +17,40 @@ import (
...
+// Delegating to the unwrapped datastore instead would silently discard every proxy
+// wrapped around it - in particular NewSeparatingContextDatastoreProxy, which the SQL
+// datastores apply to themselves. That proxy severs cancellation on reads so that a
+// caller going away cannot kill an in-flight query; pgx reacts to a cancelled context
+// by setting an immediate deadline on the connection. Without it, the dispatcher
+// cancelling losing subproblems - thousands of times in a run - churns pooled
+// connections, and a read can come back with "timeout: read tcp ...: i/o timeout"
+// having nothing to do with the query that reported it.
 func WrapWithIndexCheckingDatastoreProxyIfApplicable(ds datastore.Datastore) datastore.Datastore {
 	uds := datastore.UnwrapAs[datastore.SQLDatastore](ds)
```
ecordell, line 39: makes me wonder if there are other uses of this we need to audit for the same bug.

vroldanbet: That's a fair point, I'll look into it

ecordell (commented):
It would be nice to have a test in one of the generic DB test suites to confirm the cancellation severing so that this bug can't creep back in unnoticed. (as a followup)

ecordell (approved):

PR #3332, opened 2026-09-21
experiment: consistency suites on 8 vCPU for comparison (do not merge)

vroldanbet:
## What

Not for merge. A measurement PR: identical to #3328 beneath it, except the three consistency jobs run on **8 vCPU** instead of 16.

## Why

#3328 moves the consistency jobs to a 16-core runner, which costs twice as much per minute. That's only worth it if the jobs get proportionally faster. Comparing this PR's job durations against #3328's answers that directly, with everything else held constant.

## How

One change: `depot-ubuntu-24.04-arm-16` → `depot-ubuntu-24.04-arm-8` on the three consistency jobs. Fixtures still run in parallel, same emulator version, no parallelism cap — so any difference in duration is the runner size and nothing else.

## References

Compare the `Datastore Consistency Tests` durations here against the same jobs on #3328.

Depot ARM pricing is roughly $0.024/min for arm-8 and $0.048/min for arm-16. The upgrade is cost-neutral only if the jobs run about twice as fast; below that it buys wall-clock time at a higher price, which is a judgement call rather than a free win.


vroldanbet:
Closing: this was a measurement, and it gave us the answer.

Running the consistency suites on 8 vCPU against an otherwise identical 16 vCPU run (#3328) showed 1.57x more total runner time on the smaller machine, for half the per-minute price — so 16 vCPU costs about $0.43 more per run (+27%). The 19 jobs unchanged between the two runs had a median ratio of 0.97, so the comparison was not skewed by run-to-run noise.

Keeping 16 vCPU. It does not shorten time-to-green today, because the critical path is the CRDB integration job at ~19.5 min either way, but the faster feedback on the consistency suites is worth the modest increase. The measured numbers are recorded in #3328's description.

PR #3334, opened 2026-09-21
fix(testing): repair two tests in the shared datastore suite

vroldanbet:
## What

Two fixes in the suite that every datastore engine runs:

- `TestUniqueID` has never actually run — in any engine, ever.
- `StatsTest` does its work eleven times over instead of once.

## Why

**`TestUniqueID`** is written differently from its 110 neighbours. They pass the test function to `t.Run`; this one wraps it in a closure that builds the test function and then throws it away. `t.Run` gets handed a body that does nothing, so the subtest reports a pass in 0.00s without executing a line of the test. It has been green and empty for as long as it has existed. It passes once actually run.

**`StatsTest`** loops so it can wait for the statistics tables to catch up. The problem is that the success path doesn't leave the loop — it falls off the bottom and the loop goes round again. So it always runs the full eleven passes and asks for statistics twice on each one. On CockroachDB every one of those triggers an `ANALYZE`, which is why this test is disproportionately slow there.

## How

Pass the test function to `t.Run` the same way every other entry does, and `break` out of the retry loop once the assertions have passed. The retry behaviour is unchanged — it still retries when statistics aren't ready yet, and still fails properly if they never are.

## References

- Found while investigating why the CRDB datastore job is the slowest thing in CI.


squat (approved):
Ahh sneakyy

PR #3335, opened 2026-09-21
perf(crdb): set the TTL changefeed parameter during migration

vroldanbet:
## What

Move one table setting from "fixed up every time SpiceDB starts" to "applied once, by the migration that should have applied it".

## Why

CockroachDB's row-level TTL job hard-deletes expired relationships. Those deletes would otherwise show up in the changefeed behind the Watch API, so both relationship tables need `ttl_disable_changefeed_replication`.

The migration that adds the TTL policy sets the other two TTL settings but not this one. Startup notices it missing and adds it with two `ALTER TABLE`s. Because the migration never learns, this repeats for every database, on every start, forever.

It's mostly invisible in production — two statements once per process. It is not invisible in the test suites, which build a database per test: those two statements cost roughly 0.2s of every datastore construction, and there are hundreds per CI job.

## How

Set the parameter as part of the TTL policy in the migration, so a freshly migrated database already has it and the startup check finds nothing to do.

The startup check stays. Databases migrated before this change still need it, and an operator who has deliberately set the parameter themselves should keep their choice — the check already respects that.

The parameter arrived in CockroachDB v24.1, so the migration gates on the same version the startup check does. Older clusters get the policy without it and are fixed up at startup exactly as before.

## References

- Existing coverage in `TestTTLChangefeedSuppressionParam` and `TestTTLChangefeedSuppressionWatch` asserts the end state, so it holds whichever way the parameter gets set. Both pass.


vroldanbet:
Closing in favour of fixing the underlying issue rather than this symptom.

This edited migration 0009, which has already shipped. That means two databases with identical migration history end up with different schemas, and it reaches no existing deployment — the startup fix-up in `ttl_changefeed.go` still has to stay, so nothing gets deleted. The entire payoff was a couple of hundred milliseconds per datastore construction in the test suites.

The real problem is that nothing asserts that migrating to head produces the schema the runtime expects. `MigrationTest` checks that each migration applies cleanly and preserves data, but never checks the result, which is why migration 0009 setting two of three TTL settings went unnoticed. A follow-up PR adds that check.

PR #3337, opened 2026-09-21
perf(testing): run the shared datastore suite's subtests in parallel

vroldanbet:
## What

The shared datastore test suite now runs its subtests in parallel, and each engine chooses whether to.

`AllWithExceptions`, `All` and `OnlyGCTests` take `SuiteOption`s. The default is parallel; `RunSubtestsSerially()` opts an engine out, `RunSubtestsInParallel()` opts one back in if the default ever flips. No test was rewritten, moved or excluded.

| Engine | Runs | CI |
| --- | --- | --- |
| CockroachDB | parallel | green |
| MySQL | parallel | green |
| memdb | parallel | green |
| Postgres + pgbouncer | serial | opted out, see below |
| Spanner | serial | opted out, see below |

## Why

The suite runs 111 subtests back to back, and every one of them creates its own database and runs the migrations on it before it does any work. That setup is the bulk of what most subtests cost.

The top-level test for each engine is already marked parallel, but there is only one of it in the package, so it has nothing to run alongside. All the available concurrency was one level down, going unused.

CI job times, this branch against its base. CockroachDB is the critical path of the datastore matrix, which is why it is the one to care about. The engines that opted out are unchanged, as expected, and are listed to show the change did not cost them anything.

| Job | Before | After |
| --- | --- | --- |
| crdb, 26.2.5 | 900s | 649s |
| crdb, 25.2.0 | 1143s | 758s |
| mysql | 371s | 259s |
| postgres 14-18 | 297-380s | 307-343s |
| pgbouncer 14-18 | 274-423s | 304-348s |
| spanner | 511s | 542s |

## How

A subtest is wrapped by a `runner` before it goes to `t.Run`: `serial`, which is what was there before, or `parallel`, which calls `t.Parallel()` first. The suite picks one from its options.

Subtests do not share data, since each builds its own database, but all the subtests of an engine share one container. That is what the two opted-out engines trip over.

**Spanner** failed as infrastructure. `TestMigration` and `TestStoredSchemaUpdate` failed first, then every remaining subtest died at 0.00s with `grpc: the client connection is closing` — a shared client torn down while siblings were still using it.

**Postgres** failed on its own terms, which is the more interesting one: `TestNamespaceWrite` asserted and failed on all five Postgres versions and all five pgbouncer versions. That is a test that is not parallel-safe, not a container being shared badly.

Both are known and deferred. Fixing them belongs in follow-ups, one engine at a time, and each is a one-line removal here once its cause is understood.

## References

Based on `fix/datastore-suite-dead-and-looping-tests`, which touches the same lines.


vroldanbet:
Cherry-picked into #3328, which is the same concern — making the test suites run in parallel — so the two now travel together.

Commits `1983f2e8` and `3cb604a8` applied cleanly onto `perf/parallel-consistency-fixtures`. The measured results carry over: crdb 25.2.0 1143s to 758s, crdb 26.2.5 900s to 649s, mysql 371s to 259s, with Postgres and Spanner opted out via `RunSubtestsSerially()`.

PR #3338, opened 2026-09-21
fix(testing): stop the FDW e2e test losing its port before it binds

vroldanbet:
## What

Makes the FDW end-to-end test hold the port it is going to use, instead of
picking one, letting it go, and hoping it is still free when the server starts.
Also makes both test servers report why they stopped instead of dropping the
error.

## Why

`TestEndToEnd/simple` fails occasionally in the Integration Tests job, always
the same way: ten seconds of `connect: connection refused` followed by
`PGServer did not start accepting connections`. Example:
[run 35624116825](https://github.com/authzed/spicedb/actions/runs/35624116825/job/106414205833).

The test chose its port with a helper that opens a listener on port 0, reads
the port back, closes the listener and returns the number. The server was then
started on that port from a goroutine. Between the close and the bind the port
belongs to nobody, and the Integration Tests job runs six packages plus a
docker daemon at once. Anything can take it - a published container port, or
just an ordinary outgoing connection, because Linux hands a freshly released
ephemeral port to `connect()` as readily as to `bind()`. When that happens the
server's bind fails, the error is thrown away by `_ = pgserver.Run(...)`, and
all the test can say afterwards is that nothing ever answered.

Two smaller things fell out of the same investigation. The reported address was
`[::1]` while the server binds `0.0.0.0`, which reads like an IPv6 problem but
is not - it is just the first address `localhost` resolves to, and Go tries
IPv4 straight after. And the SpiceDB server in the same file discarded its
startup error too.

## How

`PgBackend` gets a `Serve(ctx, listener)` method. `Run(ctx, endpoint)` keeps its
signature and now binds the listener itself and calls `Serve`, so the production
path is unchanged.

The test binds `0.0.0.0:0` directly, takes the port from the listener and hands
the listener to `Serve`. The port is held from the moment it is chosen, so
nothing can take it. `net.Listen` also returns with the socket already
listening, so the ten-second readiness poll is gone rather than merely made
longer - the port accepts connections before the helper returns, whenever the
server goroutine happens to be scheduled.

Both goroutines now send what they returned down a channel. The PGServer one is
checked in cleanup, after the server has been closed; the SpiceDB one is checked
on each readiness attempt. Either way the next failure names its cause.

## References

Reproduced on Linux by taking the port in exactly that window - an outgoing
connection whose local port is the one the helper just returned:

```
before (main):        0 passed, 10 failed out of 10
after  (this branch): 10 passed, 0 failed out of 10
```

The failure matched CI line for line, including the ten-second duration and the
`Condition never satisfied` / `PGServer did not start accepting connections`
wording.

No-regression runs on this branch:

- `TestEndToEnd/simple`, `-race -count=50`: 50 passed, 0 failed
- whole `internal/fdw` package with `-race`: pass
- whole package on Linux, 3 runs: 3 passed, 0 failed

The bind conflict only happens on Linux; macOS lets the wildcard bind through,
which is why this has never shown up on a laptop.


squat (approved):
Ugh yes, classic race :')

PR #3341, opened 2026-09-21
fix(testing): give Spanner's migration test the emulator it already has

vroldanbet:
## What

Registers the Spanner emulator that the datastore suite is already running as `MigrationTest`'s config provider, so the test stops starting a second emulator. One emulator container per `mage testds:spanner` run instead of two.

This is the prerequisite for taking Spanner off `test.RunSubtestsSerially()` in #3334. It does not remove the opt-out itself — see **How** for the follow-up.

## Why

#3334 turns the shared suite's subtests parallel and Spanner went red, which is why `750cc265` opted it out. The reported symptom was a cascade of `rpc error: code = Canceled desc = grpc: the client connection is closing` at `internal/testserver/datastore/spanner.go:107`, which reads like a shared client being torn down. That is the cascade, not the cause. The cause is at the bottom of the CI log:

```
panic: testing: test using t.Setenv, t.Chdir, or cryptotest.SetGlobalRandom can not use t.Parallel [recovered, repanicked]
testing.(*T).Setenv(0xc000153b08, {0x2e2384c, 0x15}, {0xc0021381f1, 0xf})
github.com/authzed/spicedb/internal/testserver/datastore.RunSpannerForTesting(...)
	internal/testserver/datastore/spanner.go:57
github.com/authzed/spicedb/internal/testserver/datastore.RunDatastoreEngine(...)
	internal/testserver/datastore/datastore.go:78
github.com/authzed/spicedb/pkg/datastore/test.MigrationTest(...)
	pkg/datastore/test/migration.go:75
github.com/authzed/spicedb/pkg/datastore/test.parallel.func1(...)
	pkg/datastore/test/datastore.go:228
```

`MigrationTest` needs an empty, unmigrated database. With no registered provider it calls `testdatastore.RunDatastoreEngine(t, engineKey)`, which for Spanner starts a whole second emulator container and points `SPANNER_EMULATOR_HOST` at it.

That variable is process-wide: the Spanner client libraries read it, and both `migrations.NewSpannerDriver` and `NewSpannerDatastore` `os.Setenv` it themselves. So a second emulator can never coexist with the one the rest of the suite is using — whoever writes the variable last owns every client created after. Serially that is invisible, because nothing else is running. Under `t.Parallel` it is fatal twice over:

1. `t.Setenv` panics outright in a test that has called `t.Parallel`, and
2. Go runs the panicking test's cleanups *and its parents'* before aborting, which terminates the shared emulator and closes the shared `InstanceAdminClient`. Every sibling subtest that had not yet started then fails at 0.00s with `the client connection is closing`, and the ones that had started fail with `unable to instantiate spanner client`. The later victims even log `using spanner emulator, host: ` with an empty host, because `t.Setenv`'s own cleanup had restored the variable to unset.

That is the whole cascade from a single root.

## How

`test.RegisterMigrationTestConfig` already exists for exactly this: it lets an engine hand `MigrationTest` an unmigrated database instead of having it build one. `spannerTest.NewDatabase` creates a fresh instance and database on the running emulator and does not migrate it, which is precisely what the test wants, so it is registered as the provider.

Verified locally against the emulator (1.5.58, arm64, `-race`, `-ldflags=-checklinkname=0`):

* With the suite forced parallel (`test.RunSubtestsSerially()` removed locally), **10/10 runs green** at `-test.parallel 8` — CI's `depot-ubuntu-24.04-arm-8` core count — 313 subtests each, no data races. Wall clock 113-135s, against 492s at `-test.parallel 1` on the same machine.
* Before the change, the same configuration panics on the first run, every time.
* On this branch as it stands (Spanner still serial), `TestSpannerDatastore/TestMigration` passes and the run creates one emulator container instead of two.

Caveat on the timings: five other agents were running containers on the same machine throughout, so treat the wall-clock numbers as a ratio, not an absolute.

**Follow-up, not done here:** dropping `test.RunSubtestsSerially()` from `internal/datastore/spanner/spanner_test.go` is a one-line change, and the comment above `test.AllWithExceptions` describing the cause as "the emulator's shared client is torn down while sibling subtests are still using it" should go with it. Full local tally, all with this fix applied:

| configuration | result |
| --- | --- |
| `-test.parallel 8`, `-race` (CI-like) | 10/10 green |
| `-test.parallel 12`, no race | 6/8 green |
| `-test.parallel 1`, `-race` (serial-equivalent) | 1/2 green |

The serial-equivalent failure is `TestWatchBasic/256-false` — `expected changes missing` after 20.2s — which is nothing to do with concurrency. The Spanner watch tests already sit right against `waitForChangesTimeout`, a 20s constant carrying the comment "FIXME, for reasons unknown, tests against Spanner emulator time out if this is lower". They are flaky at any level of parallelism, and concurrency makes an existing problem more likely rather than creating a new one.

Two parallel-specific hazards did show up at 12-way, neither Spanner-specific and neither addressed here:

* `TestReadyStateWhenPaused` `docker pause`s the **shared** container, which freezes it for every concurrently running sibling — `readystate.go` says as much in its own doc comment. Observed once in 3 runs: three watch tests timed out inside the pause window, between the `datastore paused` and `unpausing container` log lines. This applies equally to CockroachDB and MySQL, which stay parallel in #3334 and also wrap `PausableTester` around their shared container. Running that one subtest outside the parallel group would fix it for all of them.
* Session pressure on one emulator. Observed once in 5 runs: `spanner: code = "DeadlineExceeded", desc = "timeout / context canceled during getting session."` Not seen at all in 10 runs at CI's 8-way.

## References

#3334


vroldanbet:
Cherry-picked into #3328 as `71b02d66` and `0245e507`, alongside the rest of the parallelism work.

The root cause you found — `MigrationTest` starting a second emulator and calling `t.Setenv`, which panics once `t.Parallel` has been called — is now fixed there, and the misleading comment blaming a torn-down client is corrected. Spanner's `RunSubtestsSerially()` opt-out stays for the moment; removing it is sequenced behind the `TestReadyStateWhenPaused` fix.

PR #3345, opened 2026-09-21
fix(crdb): report a canceled watch as canceled while still connecting

vroldanbet:
## What

Cancelling a CockroachDB watch gives the caller a different error depending on whether the watch had finished opening its connection. It should always be the same one.

## Why

Watch opens its own connection, separate from the pool. Every error path in that function turns a cancelled context into `WatchCanceledError` — except the connect itself, which sends whatever pgx produced straight to the error channel:

```
expected: datastore.WatchCanceledError
in chain:  "failed to connect to `user=root database=...`:
              failed to receive message: context canceled
              dial error: ... operation was canceled" (*pgconn.ConnectError)
```

So a caller cancelling a watch gets `WatchCanceledError` if the connection was already up, and a raw pgx connect error if it wasn't — a race they cannot see and cannot sensibly handle.

It went unnoticed because the window is small: the connection normally opens well before anyone cancels. Running the datastore suite's tests side by side made the machine busy enough to land in that window reliably.

## How

The translation already existed, in `sendError`. The connect was the only path not using it, because `sendError` was declared thirty lines further down. Moving the declaration above the connect and sending the error through it makes the two paths agree.

Nothing changes for a watch that connects successfully.

## References

- Postgres already handles this case; this brings CockroachDB in line.
- Surfaced by the parallel datastore suite in #3328.


vroldanbet:
Cherry-picked into #3328 as `c7b1151b`. This flake was firing across the whole stack, and as a sibling PR the fix reached none of it — folding it into the parallelism PR that exposed it means every branch above inherits it.

PR #3346, opened 2026-09-21
fix(testing): stop the revision serialization test racing a one-second clock

vroldanbet:
## What

A test that checks revisions survive serialization was failing because of an unrelated one-second deadline on its setup.

## Why

`RevisionSerializationTest` writes a namespace, then checks the resulting revision can be serialized and sent through the dispatch layer. The write is setup; the serialization is the point.

That write was given a hard one-second budget, so a slow write failed the test with `context deadline exceeded` before it reached anything it was meant to check:

```
--- FAIL: TestCRDBDatastoreWithoutIntegrity/TestRevisionSerialization (2.79s)
    revisions.go:169: Received unexpected error: context deadline exceeded
```

One second was ample when the suite ran one test at a time. It is not now that a hundred of them share a database server — the failure appeared in two of three CockroachDB runs on a loaded machine.

## How

Raise the deadline to thirty seconds and say in a comment what it is for. It still exists, so a genuinely hung write fails the test rather than hanging it, but it is now far enough out that expiry means something is actually wrong rather than that the machine was busy.

## References

- The deadline dates from the test's original commit in 2022 and was never a latency assertion.
- Same shape as the fix for `TestGCProcessRuns`, which replaced a fixed sleep with polling for the same reason.


vroldanbet:
Cherry-picked into #3328 as `cc5df29e`, for the same reason as #3345: the flake was firing on branches that a sibling PR could never reach.

PR #3348, opened 2026-09-21
test(postgres): run the shared datastore suite's subtests in parallel

vroldanbet:
## What

Removes the `test.RunSubtestsSerially()` opt-out from the two `AllWithExceptions` calls in `internal/datastore/postgres/postgres_shared_test.go`, so Postgres runs the shared datastore suite's subtests in parallel like the other engines. The GC calls stay serial — those tests take exclusive locks.

To make that possible, the shared suite stops asserting an ordering between a definition's `LastWrittenRevision` and the revision it was read at.

`Skip-Changelog` — test-only.

## Why

The parallel suite failed deterministically on all five Postgres versions and all five pgbouncer variants:

```
--- FAIL: TestPostgresDatastore/postgres-18-/TestNamespaceWrite
    namespace.go:100: Should be false        // require.False(createdRev.GreaterThan(secondWritten))
--- FAIL: TestPostgresDatastore/postgres-18-/TestNamespaceDeleteNoRelationships
    namespace.go:226: Should be true         // require.True(nsCreatedRev.LessThan(deletedRev))
--- FAIL: TestPostgresDatastore/postgres-18-/TestNamespaceDelete
```

Instrumenting the failing assertion with the real snapshots shows exactly why:

```
createdRev    = pgSnapshot{xmin:0x7985, xmax:0x7985, xipList:nil}
secondWritten = pgSnapshot{xmin:0x797a, xmax:0x7985,
                           xipList:[0x797a 0x797d 0x797e 0x7980 0x7981 0x7983]}
                optionalTxID: 0x7984
```

A Postgres revision is a PostgreSQL snapshot, and `pgSnapshot.compare` is a partial order over *how much a snapshot knows*. But the last-written revision is synthesized from the row's `created_xid` alone — `revisionForVersion` builds `xid+1:xid+1:`, which in that encoding asserts "every transaction below `xid+1` is settled". Here it claims to know the disposition of 0x797a and five others that `secondWritten` still had in flight, so it carries strictly more information and compares as **greater** than the snapshot it was read at.

**This is not an artefact of parallel subtests sharing one cluster.** One concurrent transaction in the *same* database is enough. Measured against a live Postgres 18 with a single extra open transaction:

```
concurrent in-flight xid in the SAME database: 853
writtenRev = pgSnapshot{xmin:0x355, xmax:0x357, xipList:[0x355]}  (own txid 0x356)
createdRev = pgSnapshot{xmin:0x357, xmax:0x357, xipList:nil}
createdRev.GreaterThan(writtenRev) = true
```

Serial subtests simply never have a concurrent transaction in flight, which is why the suite only noticed once it went parallel.

**This is not a production bug.** Every consumer of `LastWrittenRevision` — `internal/dispatch/graph`, `internal/graph/check`, `internal/namespace/util`, and both schema caches — takes only `.Definition` and discards the revision. Nothing in the server compares it against anything, and it is never returned to a client as a ZedToken (`ReadSchema`'s `read_at` comes from the request revision). The over-claim is latent, not live, and it is inherent to the encoding: `pgSnapshot` has no way to say "I know nothing about transactions below this one". Making it faithful would mean joining `relation_tuple_transaction` for the writing transaction's stored snapshot on every schema read — real cost, no production benefit, purely to satisfy a test. Not done here.

CockroachDB cannot satisfy the assertion precisely either, in the opposite direction: its last-written revision is `revisions.NewHLCForTime` over the row's MVCC timestamp, which drops the logical counter, so it can land strictly *before* the write it describes. (Verified: an intermediate version of this change asserted the definition is readable at its own last-written revision, and CRDB failed it with `object definition 'foo/bar' not found`.) The existing assertion passes on CRDB only because the imprecision happens to err in the tolerated direction.

Counters are the contrast that proves the point: `RelationshipCounter.ComputedAtRevision` stores the *real* `pg_current_snapshot()` from the `snapshot` column, and `counters.go`'s `updatedRev.Equal(...)` assertion passes in parallel without change.

## How

- `pkg/datastore/test/namespace.go`: drop the five assertions that compare a `LastWrittenRevision` against the revision a read was performed at (three `require.False(createdRev.GreaterThan(...))` in `NamespaceWriteTest`, two `require.True(nsCreatedRev.LessThan(deletedRev))` in the delete tests), with a comment explaining what the value does and does not guarantee. Kept: the returned definition matches what was written, and `createdRev.GreaterThan(startRevision)` — that one compares against a revision taken before anything happened and holds on every engine.
- `internal/datastore/postgres/revisions_test.go`: a unit test pinning the documented `revisionForVersion` behaviour, built from the numbers measured above, so the reason the suite no longer makes that comparison does not get lost.
- `internal/datastore/postgres/postgres_shared_test.go`: drop the two opt-outs, replace the stale comment.

## References

Verification, all on this branch:

| suite | runs | result |
|---|---|---|
| `TestPostgresDatastore` parallel (pg 18) | 5 | 5/5 pass, 58–65s each |
| `./internal/datastore/postgres/...` full, `postgres` tags | 1 | pass |
| `./internal/datastore/postgres/...` full, `pgbouncer` tags | 3 | 3/3 pass, 102–113s each |
| `./internal/datastore/crdb/...` full | 1 | pass |
| `./internal/datastore/memdb/...` full | 1 | pass |
| `golangci-lint` on both changed packages | 1 | 0 issues |

Before the change, for comparison: 3 parallel runs, 3 failures each time (`TestNamespaceWrite` every run; `TestNamespaceDelete` and `TestNamespaceDeleteNoRelationships` intermittently, since those additionally need a concurrent transaction older than the namespace's own creation).

MySQL and Spanner are unaffected by construction — the shared-suite change only removes assertions.

Speedup on `TestPostgresDatastore` locally, Postgres 18, same machine:

```
serial:   108.8s, 104.9s
parallel:  59.4s, 58.7s, 58.1s, 65.2s, 58.2s
```

roughly 45% off that target. Stacked on `perf/parallel-consistency-fixtures`.


vroldanbet:
Folded into #3328 as `15422b11`, alongside the rest of the parallelism work — a sibling PR could not reach the branches that were failing.

Two edits on the way in: the code comment attributing the cause to cluster-wide transaction ids was corrected, since your own experiment showed a single concurrent writer in the same database is enough; and the note about CockroachDB's opposite-direction imprecision was moved into the commit message so it is not lost.

PR #3366, opened 2026-09-22
fix(memdb): never hide the first write behind the creation snapshot

vroldanbet:
## What

MemDB could serve reads from the empty snapshot it takes when the datastore is created, even after data had been written. Requests then failed with `object definition not found` or came back empty.

## Why

This showed up as a CI flake in `TestMiddlewareOrdering`, but the underlying bug is not test-only. Starting SpiceDB with `--datastore-engine=memory` and `--datastore-bootstrap-files` can, for up to one quantization interval after startup (5 seconds by default), answer requests as if the bootstrap data were not there.

`OptimizedRevision` — the revision used by the default `minimize_latency` consistency — rounds the current time down to the nearest quantization boundary. MemDB keeps a snapshot of the still-empty database from the moment the datastore was created, and reads at any revision before the first write are served from it.

There was already a guard for this, but it only caught boundaries that landed *before* the creation snapshot. That is the common case, which is why this almost always worked. When the boundary instead landed in the small gap *between* the creation snapshot and the first write, the empty snapshot was advertised and the data went missing.

Measured on `main`: 6 failures in 2400 runs of `TestMiddlewareOrdering` (~0.25%), matching the ~1ms bootstrap gap against a 5s interval. 2400/2400 pass with this change.

## How

MemDB now advertises its head revision whenever the quantized revision falls before the first write, not just before the creation snapshot. Normal quantization is unchanged once real writes exist, so reads still get the stale-but-stable revision the interval is there to provide.

`TestOptimizedRevisionSeesFirstWrite` forces a boundary into that window on purpose; it fails 20/20 without the fix and passes 20/20 with it.

`TestMiddlewareOrdering` also now reads its bootstrapped setup data at full consistency. That test is about middleware ordering and should not depend on where a quantization boundary happens to fall; the memdb behaviour itself is covered by the new datastore test.

## References

- Flaky job: https://github.com/authzed/spicedb/actions/runs/35717349722/job/106713414978


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
 ### Fixed
+- MemDB: requests made shortly after startup could fail with `object definition not found`, or return no results, even though the schema and relationships had already been written. Reads at `minimize_latency` (the default) use a revision rounded down to the nearest quantization interval, and when that boundary happened to fall between the moment the datastore was created and the first write to it, the read was served from the empty snapshot taken at creation. This was most visible when starting SpiceDB with `--datastore-engine=memory` and `--datastore-bootstrap-files`, where the window lasted until the next interval boundary (up to 5 seconds by default). MemDB now advertises its head revision in that case, so committed data is never hidden behind the datastore's own creation. (https://github.com/authzed/spicedb/pull/3366)
```
```suggestion
- MemDB: requests made shortly after startup could fail with `object definition not found`, or return no results, even though the schema and relationships had already been written. MemDB now advertises its head revision in that case, so committed data is never hidden behind the datastore's own creation. (https://github.com/authzed/spicedb/pull/3366)
```

tstirrat15 (dismissed):
LGTM

tstirrat15 (approved):
LGTM

